PR #51, opened 2024-11-21
Updating readme.md link for Application Criteria & Current Capabilities

OcWebb:
Straightforward link update. I do hope to contribute more meaningfully in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Before contributing, please sign our Contributor License Agreement (CLA). You ca
 This is not production-ready code. It is an **early beta version** that is under active
 development. This early beta version is functional but evolving. Expect frequent changes as we develop core features.
 
-You can find a list of the systems **current capabilities and application criteria** [here](https://thousandbrainsproject.readme.io/docs/application-criteria).
+You can find a list of the systems **current capabilities and application criteria** [here](https://thousandbrainsproject.readme.io/docs/application-criteria-current-capabilities).
 
 You can find our **project road map** and details on the next features we are working on [here](https://thousandbrainsproject.readme.io/docs/project-roadmap).
 


vkakerbeck (approved):
Good catch, thank you!

I'm a bit surprised by this since I had renamed this page in [this PR](https://github.com/thousandbrainsproject/tbp.monty/pull/21) and the [file](https://github.com/thousandbrainsproject/tbp.monty/blob/main/docs/overview/application-criteria.md) is currently just named application criteria. @codeallthethingz is the URL not changed because I didn't update the title in the meta data at the top of the file? I can make a follow-up PR with that.

OcWebb:
Thank you for linking that prior PR, based on that it seems the desire was to go with the shorter slug.
In that case, I could update my pull request later. I suspect the title metadata field is used as the default slug when not otherwise defined, but this document indicates we could add a [slug field to the metadata](https://docs.readme.com/main/docs/page-metadata-and-seo#:~:text=%2D%2D%2D%0Atitle%3A%20%22Welcome!%22-,slug%3A%20%22welcome%22,-metadata%3A%0A%20title%3A%20%22My). Either solution may work.

codeallthethingz:
Yes, because readme.com's create-doc API is a bit strange.  When you create a doc you can only use the title in the create request and readme.com tells you the slug that was created.  When you're running a sync, the tool uses the slugs in the hierarchy.md to check if there is an existing page and update it, if it doesn't find the slug, it creates a new page based on the title.
  
This shouldn't happen too often because we're not changing slugs now we're live.  I did a test in this PR #54

vkakerbeck:
Thanks for the explanation @codeallthethingz Does that mean to change a slug I just need to change the metadata at the top of the page? 
@OcWebb yes, exactly, that was my original intent to use the shorter title and slug. Thanks for offering to update the PR.  I think for practical reasons (since we are still waiting on your CLA to be processed) I would merge Will's PR with the slug fix. I really appreciate you finding this error and also offering a fix for it so quickly!

OcWebb:
It makes sense to clean this up quickly with the other PR. Thanks for all the responses. I do intend to parse through the codebase and hope to contribute something more substantial in the future. Very excited for this project!

I will close this PR.

vkakerbeck:
thanks for understanding! I didn't want to dampen your enthusiasm by not merging this PR. Definitely looking forward to hearing more from you :)

PR #53, opened 2024-11-21
Add vectorized figures

nielsleadholm:
Updates several of the core figures with vectorized versions, and addresses low resolutions in several of them. Some small fixes to different figures (e.g. feature representations, new labels).

Also updates the color scheme of figures based on a new approach which combines the TBP colors with a more color-blind friendly mix. I will open a separate PR that adds this scheme to the style guide, as well as gives a high level "Legend" for our figures. 

Note that despite the resolutions, all the figures are around 300-600kb, with the largest are around 1-1.5MB (e.g. the Monty Overview diagram - same as before), so shouldn't give any performance issues.


docs/figures/how-monty-works/possible_matches.png
vkakerbeck, line 1: Can we just make the arrows solid black? The pink is a bit on an eye sore with the reds and oranges 😬

docs/figures/how-monty-works/step_episode_epoch.png
vkakerbeck, line 1: Why does the epoch 1 bracket not start at step 1 of episode 1?

docs/figures/overview/cortical_columns_lm.png
vkakerbeck, line 1: Should "Observation" maybe be "Feature"?

docs/figures/overview/reference_frames.png
vkakerbeck, line 1: Maybe here we also change observation to feature?

vkakerbeck (commented):
Beautiful 😍 Just a few minor comments

docs/figures/how-monty-works/possible_matches.png
nielsleadholm: Yes sure 😬 I was struggling to find a choice that is easy to visually parse. Somehow with the background dots and the dark red dot, any type of arrow gets visually lost a bit in at least one of them. What do you think of the white-filled one? I think maybe that's a bit more visible than solid black, without being as garish as pink?

Screenshot 2024-11-22 at 10 49 57 AM

Screenshot 2024-11-22 at 10 50 06 AM

vkakerbeck: Good point! How about the yellow/gold since this is what I tried to use the yellow dot for?
Screenshot 2024-11-22 at 11 58 56 AM

nielsleadholm: Adding that now

docs/figures/how-monty-works/overview_diagram.png
nielsleadholm, line 1: @vkakerbeck I thought it would be worth making the surface-agent models explicitly grey-scale in this to help avoid confusion.

vkakerbeck: oh that's a great idea!

vkakerbeck (approved):
Great, thanks for the updates!

PR #55, opened 2024-11-21
docs: conda init instructions

tristanls:
This pull request is a follow-up to #48, where @d-led identified a problem with the installation instructions.

The intent is to (hopefully) result in error-free installation while highlighting the option to prevent `conda` from interfering with every terminal window in the future.


docs/how-to-use-monty/getting-started.md
```diff
@@ -50,10 +50,26 @@ If you have miniconda installed, you can create the environment with the followi
 
 ```shell Intel
 conda env create
+# If you are using the zsh shell, run:
+conda init zsh
+# Or, if you are using a different shell, run:
+conda init
+# After init, if you do not want conda to change your global shell when
+# you open a new terminal, run:
+conda config --set auto_activate_base false
```
vkakerbeck, line 59: This involves a lot of reading and deciding whether to use optional commands. Usually when I go to an install guide I like to just be able to quickly copy paste all the lines in order without much complication. Could we maybe add the custom zsh and global shell instructions as a separate tab (like we currently have Intel and ARM64)? I know its not another type of architecture so alternatively we could also add a note below the command saying something like "note that if you are using the zsh shell you should be running `conda init zsh` instead of just `conda init`"
and "Hint: After init, if you do not want conda to change your global shell when you open a new terminal, run:
conda config --set auto_activate_base false"

That would make it a bit simpler to parse for the standard user but still provide the extra information for whoever needs it.

vkakerbeck (commented):
Thanks for updating! Just a comment on making it a bit easier to parse for the standard user.

docs/how-to-use-monty/getting-started.md
tristanls: I think you're right, and that would be more readable. Updated.

vkakerbeck (approved):
great, thanks for the update! Looks nice :)

docs/how-to-use-monty/getting-started.md
d-led: @vkakerbeck great! `conda config --set auto_activate_base false` did the trick along with `conda init zsh`

tristanls: @d-led, thank you for the confirmation!

PR #58, opened 2024-11-21
docs: clarify conda is required for installation

tristanls:
Current docs hint that there may be other ways to install Monty besides Conda. This pull request clarifies that Conda is the only supported way of setting up Monty.


codeallthethingz (approved):
LGTM

PR #59, opened 2024-11-21
ci: deploy api docs to github pages

tristanls:
This is a troublesome CI change to test on a fork as it requires specifically configured runners, etc.

I want to merge this to the `main` branch to test the Monty API Docs workflow against our actual environment. This seems "safe", in that we don't have any GitHub Pages currently, and all the workflow does is 1) checkout tbp.monty, 2) git diff tbp.monty, 3) deploy pages from an artifact.

Unknowns:
- Will the `deploy_sphinx_monty_api_docs` job be able to locate the GitHub Pages artifact by name alone? Most artifacts are coupled to `github.run_id`, however, by being triggered via `workflow_run`, the Monty API Docs workflow should inherit Monty's `github.run_id`.


tristanls:
Ready for review once again.

.github/workflows/monty_api_docs.yml
```diff
@@ -0,0 +1,64 @@
...
+            if [[ $changed_file != .github/ISSUE_TEMPLATE/* ]] &&
+               [[ $changed_file != .vscode/* ]] &&
+               [[ $changed_file != docs/* ]] &&
+               [[ $changed_file != rfcs/* ]] &&
+               [[ $changed_file != tools/* ]] &&
+               [[ $changed_file != CODE_OF_CONDUCT.md ]] &&
+               [[ $changed_file != CONTRIBUTING.md ]] &&
+               [[ $changed_file != LICENSE ]] &&
+               [[ $changed_file != MAINTAINERS.md ]] &&
+               [[ $changed_file != README.md ]]; then
```
codeallthethingz, line 54: I think an allow list would be better aligned for maintenance?  If you add something to the root folder that affects the API docs you should edit this file, as opposed to adding something to the root folder that doesn't effect the API docs meaning you should edit this file.

```
changed_file == src/* || 
changed_file == tools/generate-api-docs/* ||
changed_file == .github/etc...  
```

Also, I don't think this action needs to run if the project.toml changes.  The API docs won't change, even though the compilation process would, the output API docs should be identical.

codeallthethingz (approved):
Approved but left a comment, that's probably an indication of me not knowing stuff. :)

.github/workflows/monty_api_docs.yml
tristanls: I merged to test whether the code works. But to address your comment, here's my thinking.

I do want to have the consequence that adding something to the root folder that doesn't affect the API docs means editing this file.

The Monty API Docs are tied to the changing code. So, I was thinking that, at a high level, if we deploy the docs using the same trigger the code uses, then we will always have up-to-date API docs for the latest code version. That is the primary reason for the trigger criteria to be the same as Monty. While I agree that changing `pyproject.toml` probably won't affect API Docs, I thought it would be easier for any maintenance action to be to copy and paste the Monty trigger criteria than to think through this trigger separately.

The rest of my argument assumes we are using the Monty code testing trigger criteria, which use an exception list to be "safe to fail" (in this context, "safe to fail" meaning forgetting to update trigger criteria while adding things to the root directory). If we add code to a new folder, for example, `potatoes/`, and forget to update the Monty criteria, the job will continue to trigger and run. Having said that, this is just a general heuristic I followed, as the specificity of our checks (they specify which folders to run on) somewhat negates my argument here. Having said that, I would rather make individual checks more generic, e.g., `ruff check` instead of `ruff check src test benchmarks` following the "safe to fail" heuristic.

PR #63, opened 2024-11-23
ci: github pages deployment

tristanls:
This pull request is another attempt to deploy Monty API Docs to GitHub Pages.

In #59, I learned that each workflow receives its own `run_id`; therefore, the artifacts uploaded by the `Monty` workflow were not directly available to the `Monty API Docs` workflow. The change here now downloads the artifact from the `Monty` workflow run, uploads it in the context of the `Monty API Docs` workflow, and invokes the deploy-pages action.

This pull request also incorporates @codeallthethingz's feedback about the ease of maintaining the criteria for determining whether a workflow should run and refactors it into a single composite action reused between the `Monty` and `Monty API Docs` workflows.

As the new `Monty API Docs` workflow is triggered by the `workflow_run` event, it will only execute once merged to `main`.

I ask that this Pull Request be merged to `main` (after existing checks pass) so that the `Monty API Docs` can again be tested.


codeallthethingz (approved):
LGTM

PR #64, opened 2024-11-23
Tools: Resize images in markdown 

codeallthethingz:
This PR adds the functionality to add style attributes to images by using the fragment part of a markdown image URL.

For example

```
![](image.png#width=20px&height=20px)
```

Will result in 

```
<img src="image.png" style="width:20px; height:20px">
```


tristanls:
**issue** As we are adding a fragment to the URL, I think it should follow the usual URL convention of `key=value&key=value`, instead of `key:value`. This way, it can be easily parsed by a library like `urllib.parse.parse_qs`.

tools/github_readme_sync/readme.py
```diff
@@ -258,16 +274,24 @@ def replace_image(match):
             if any(ignore_image in match.groups()[1] for ignore_image in IGNORE_IMAGES):
                 return match.group(0)
             alt_text, image_src = match.groups()
+
+            # Split image source and fragment
+            src_parts = image_src.split("#")
+            clean_src = src_parts[0]
+            style = "border-radius: 8px;"
+            if len(src_parts) > 1:
+                style = f"{style}{src_parts[1]}"
```
tristanls, line 283: **issue** I think we should specify key value pairs in Markdown, and not, as this suggest, inline CSS.

docs/overview/welcome-to-the-thousand-brains-project-documentation.md
```diff
@@ -45,6 +45,5 @@ We are excited to have you here!  Our intention for making the project open-sour
...
-|<a href="https://x.com/1000brainsproj" style="display: flex; align-items: center;"><img src="../figures/overview/twitter.png" alt="Twitter" height="15" style="opacity: 1; transition: opacity 0.2s; &:hover { opacity: 0.8; }" pointer-events="none">&nbsp;Twitter/X</a> | Get the latest news and announcements, and engage with our community |
+| <a href="https://bsky.app/profile/1000brainsproj.bsky.social" style="display: flex; align-items: center;"><img src="../figures/overview/bluesky.png" alt="Bluesky" height="15" style="opacity: 1; transition: opacity 0.2s; &:hover { opacity: 0.8; }" pointer-events="none">&nbsp;Bluesky</a> <a href="https://x.com/1000brainsproj" style="display: flex; align-items: center;"><img src="../figures/overview/twitter.png" alt="Twitter" height="15" style="opacity: 1; transition: opacity 0.2s; &:hover { opacity: 0.8; }" pointer-events="none">&nbsp;Twitter</a> | Get the latest news and announcements, and engage with our community |
```
tristanls, line 48: **suggestion (non-blocking)**: I think it should be called X here.

tools/github_readme_sync/readme.py
```diff
@@ -253,21 +270,43 @@ def get_stable_version(self) -> str:
...
             if any(ignore_image in match.groups()[1] for ignore_image in IGNORE_IMAGES):
                 return match.group(0)
             alt_text, image_src = match.groups()
+
+            # Split image source and fragment
+            src_parts = image_src.split("#")
+            clean_src = src_parts[0]
+            style = "border-radius: 8px;"
+
+            # Parse and sanitize style parameters from fragment
```
tristanls, line 284: **issue (blocking)**: I would like to see two things that reinforce each other: a sanitization library and an allowlist of supported keys.

Please look at the `nh3` sanitization library to sanitize input: https://nh3.readthedocs.io/en/latest/#. I don't want us to maintain our own sanitization code. I would rather rely on the expertise of people who care about this sort of thing.

After the CSS keys are extracted, they should be checked against an allowlist of keys, in this case, consisting of a single one: `width`. However, I noticed `height` in the tests. Whatever the list, I think we should enumerate an allowlist to vastly reduce what can be put into the HTML.

One way to accomplish the above would be to first filter `key, values` for the allowlist of keys, i.e., `width`. Then, second, naively and unsafely construct the final HTML `<figure><img...` and then put the final HTML through the `nh3.clean()` before returning it.

tools/github_readme_sync/readme.py
```diff
@@ -13,6 +13,7 @@
 import re
 from collections import OrderedDict
 from typing import Any, List, Tuple
+from urllib.parse import parse_qs
```
tristanls, line 16: **note**: 🙌

tristanls (commented):
Thank you for the fragment format update 🙌 .

codeallthethingz:
Because this has been dormant for a long time - quick update on the changes

Imported and used the NH3 lib for sanitizing and added an allowlist of style keys that can be used.

tools/github_readme_sync/readme.py
```diff
@@ -254,23 +275,48 @@ def get_stable_version(self) -> str:
...
+            if len(src_parts) > 1:
+                try:
+                    params = parse_qs(src_parts[1])
+                    allowed_styles = []
+                    for key, values in params.items():
+                        if key in ALLOWED_CSS_PROPERTIES:
+                            # Sanitize both key and value
+                            safe_key = nh3.clean(key)
+                            safe_value = nh3.clean(values[0])
+                            allowed_styles.append(f"{safe_key}: {safe_value}")
```
tristanls, line 299: **note (non-blocking)**: It may be helpful to offer an error or a warning here to provide feedback on typos or trying to use unsupported properties.

tristanls (approved):
Looks great, thank you for making these improvements

PR #65, opened 2024-11-23
Docs: Resize images using new markdown syntax

codeallthethingz:
- [ ] Requires #64 to be merged before this will work / make a smaller PR

This PR uses the new fragment syntax added to the doc sync tool to resize some images with the style attribute.

Maintainers can see a preview here: https://thousandbrainsproject.readme.io/v0.0-image-attributes/docs/sensor-modules


codeallthethingz:
Tagging all the maintainers as this is a useful thing to know about.

tools/github_readme_sync/tests/readme_test.py
```diff
@@ -511,6 +511,23 @@ def test_convert_cloudinary_videos(self):
             block_str = f"[block:html]\n{json.dumps(block, indent=2)}\n[/block]"
             self.assertIn(block_str, result)
 
+    def test_caption_markdown_images_multiple_per_line(self):
```
vkakerbeck, line 514: Should this be part of this PR: https://github.com/thousandbrainsproject/tbp.monty/pull/64

nielsleadholm (commented):
Looks great, thanks for adding this! Only comment is that `full_graph.png` is still very big (see attached screenshot), not sure if it's possible to update it as part of this PR?

Screenshot 2024-11-25 at 11 15 14 AM

PR #66, opened 2024-11-24
Actions: Deploy preview docs

codeallthethingz:
This PR creates a set of hidden docs on readme.com with a version matching the current version of the docs up to the minor version + the branch name made safe for metadata on a semver version.

For example, this branch: 0.0-deploy-preview-docs

While building it puts a sticky note on the PR's comments

Building:
image

Finished:
image

Failure:
Similar to building but with a ❌

This PR needs some action secrets / vars to be created.

- [x] README_API_KEY
- [x] IMAGE_PATH=thousandbrainsproject/tbp.monty/refs/heads/main/docs/figures
- [x] DOCS_URL_PREFIX=https://thousandbrainsproject.readme.io/

> [!NOTE]
> Currently preview docs are not deleted when they are merged to main, this will be done in a separate PR.


codeallthethingz:
@tristanls I'm flummoxed.  None of the docs.yml actions are running and I need help to figure out why.

tristanls:
@codeallthethingz actions will not pass secrets on `pull_request` trigger event. It'll need to be `pull_request_target` and in `main` before it passes secrets.  This is what happened with CLA check.

codeallthethingz:
Ah, right!  Thanks.

codeallthethingz:
I tried to add a trigger of both pull_request and pull_reqeust_target the docs.yml, but I feel like the entire file is skipped if there is a reference to a secret in it, which I think means I need to split the docs.yml into two - one for the secretless check-docs and one for the secretfull preview-docs.  Do I have that right?

codeallthethingz:
@tristanls I split the file into two, and the action that need a secret, triggers on pull_request_target.  Let me know if this is the right approach overall. 

If the above is correct:
- There is now a repeating chunk of yaml should_run_docs that could be extracted into a common file.
- Also, I wasn't sure of the naming convention now we have two "docs" type actions.  Let me know what might be a good naming strategy.


Also, this can all wait till Monday.

.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,140 @@
...
+jobs:
+  deploy_preview_docs:
+    name: deploy-preview-docs
+    runs-on: ubuntu-latest
+    needs:
+      - should_run_docs
+    steps:
+      - name: Create initial PR comment
+        if: ${{ needs.should_run_docs.outputs.should_run_docs == 'true' }}
+        uses: marocchino/sticky-pull-request-comment@v2
```
tristanls, line 21: **issue (blocking)** Our security policy does not permit the execution of actions by unverified creators.

.github/workflows/docs-check.yml
```diff
@@ -1,4 +1,4 @@
-name: Docs
+name: Docs Check
```
tristanls, line 1: **issue**: Changing this calls for changing the job names per our naming convention. 

I've referenced our naming convention in other PRs, but having it in the style guide is probably good. While it is not there yet, this PR adds it and can be a reference: #69.

.github/workflows/docs-check.yml
```diff
@@ -18,6 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - should_run_docs
+    if: github.event_name == 'pull_request'
```
tristanls, line 21: **question**: Why would we not check the docs when merging to `main`?

.github/workflows/docs-check.yml
```diff
@@ -90,7 +91,7 @@ jobs:
             echo $changed_file
             if [[ $changed_file == docs/* ]] ||
                [[ $changed_file == tools/github_readme_sync/* ]] ||
-               [[ $changed_file == .github/workflows/docs.yml ]]; then
+               [[ $changed_file == .github/workflows/docs* ]]; then
```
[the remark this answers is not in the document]
tristanls: **issue**: The rationale for including the workflow file path here is to ensure the workflow is triggered when the workflow file changes, even if nothing else changes. I don't think we should trigger this workflow if some unrelated workflow file changes.

.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,140 @@
...
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+on:
+  pull_request_target:
+    branches:
+      - main
+
+jobs:
+  deploy_preview_docs:
```
tristanls, line 13: **issue**: Should follow the naming convention described in #69.

.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,140 @@
...
+          conda env create
+          source activate tbp.monty
+          pip install -e .[dev,github_readme_sync_tool,print_version_tool]
+      - name: Get version and branch
+        if: ${{ needs.should_run_docs.outputs.should_run_docs == 'true' }}
+        working-directory: tbp.monty
+        run: |
+          export PATH="$HOME/miniconda/bin:$PATH"
+          source activate tbp.monty
+          echo "VERSION=$(python -m tools.print_version.cli minor)" >> $GITHUB_ENV
```
tristanls, line 72: **suggestion (non-blocking)**: It may be safer to namespace when using a shared environment, e.g., `MONTY_VERSION`.

.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,140 @@
...
+          # Make branch name safe for semver by replacing invalid chars with dash
+          # Remove trailing dash if present
+          echo "BRANCH_NAME=$(echo ${GITHUB_HEAD_REF} | tr -c '[:alnum:]' '-' | sed 's/-*$//')" >> $GITHUB_ENV
+          # Get current date in ISO format
+          echo "DATE=$(date -u +'%Y-%m-%d %H:%M UTC')" >> $GITHUB_ENV
```
tristanls, line 77: **suggestion (non-blocking)**: It may be safer to namespace when using a shared environment, e.g., `MONTY_DATE`.

.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,140 @@
...
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          message: |
+            📚 **Documentation Preview**
+
+            ❌ Failed to build documentation preview.
+
+            Check the [workflow logs](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}) for details.
+
+  should_run_docs:
```
tristanls, line 109: **issue**: Should follow the naming convention described in #69.

tristanls, line 109: **suggestion**: For non-required checks, we do not need to go through an artisanal setup of "should run," as seen in other workflows.

We can use the default paths filtering mechanism when a check is not required. See https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore.

The artisanal "should run" code is necessary elsewhere because when a workflow is skipped using the paths mechanism above, it registers as "skipped" instead of "success." However, a required check must "succeed" and cannot be "skipped."

.github/workflows/docs-preview.yml
codeallthethingz: I created a pin_comment reusable action.

.github/workflows/docs-check.yml
codeallthethingz: We should.

codeallthethingz:
Update as this has been fallow for a while. 

I stopped using the pin-comment third party tool and wrote an action that does the same thing.

I've tested the preview docs as a pull_request triggered build, but have put it back as a pull_request_target for the code review.  Which means it won't successfully run until we merge and open a new PR.

.github/actions/pin_comment/action.yml
```diff
@@ -0,0 +1,83 @@
...
+        COMMENT="<!--pin-comment-marker-${COMMENT_ID}-->
+        ${MESSAGE}"
+
+        # Properly escape the comment for JSON
+        ESCAPED_COMMENT=$(echo "$COMMENT" | jq -R -s '.')
+
+        # Find and delete old comment with same ID using GitHub token
+        echo "Searching for existing pinned comment with ID: ${COMMENT_ID}"
+        response=$(curl -s -w "%{http_code}" -H "Authorization: token ${GITHUB_TOKEN}" \
+          "https://api.github.com/repos/${GITHUB_REPOSITORY}/issues/${ISSUE_NUMBER}/comments")
```
tristanls, line 46: **issue (non-blocking)**: This will fail to find the pinned comment if it is beyond the first page of results. I made this comment non-blocking since making new comments over and over is not the worst.

.github/workflows/docs-preview.yml
tristanls, line 1: **issue (blocking)**: Please rename to `docs_preview.yml` to match the existing `monty_api_docs.yml`.

.github/workflows/docs.yml
```diff
@@ -90,7 +90,8 @@ jobs:
             echo $changed_file
             if [[ $changed_file == docs/* ]] ||
                [[ $changed_file == tools/github_readme_sync/* ]] ||
-               [[ $changed_file == .github/workflows/docs.yml ]]; then
+               [[ $changed_file == .github/workflows/docs.yml ]] ||
+               [[ $changed_file == .github/workflows/docs-preview.yml ]]; then
```
tristanls, line 94: **question**: Why run the Docs workflow when `docs_preview.yml` changes?

tristanls (commented):
Looks great! I added some comments, but nothing exciting.

.github/actions/pin_comment/action.yml
codeallthethingz: Agreed.  Not the worst. I think I'll increase the page size to 100, but if it's bigger than that I'd rather create a new comment than spend a bunch of action credits looking for a comment.

tristanls (approved):
Looks good.

PR #67, opened 2024-11-25
Tools: Make external-link checker faster

codeallthethingz:
This PR makes the external link checker in the doc sync tool faster.

- sets threads to 10
- runs a head request check before a get to prevent downloading huge files

Because of advanced crawling prevention, it now ignores openai.com and science.org urls.
Added some logging that shows the timing of URLs that take over a second to query.


tools/github_readme_sync/hierarchy.py
```diff
@@ -278,11 +286,28 @@ def check_readme_link(url, rdme):
...
     try:
         headers = request_headers()
-        response = requests.get(url, timeout=5, headers=headers)
+        time = timeit.default_timer()
+
+        try:
+            response = requests.head(url, timeout=5, headers=headers)
+            if response.status_code < 200 or response.status_code > 299:
+                response = requests.get(url, timeout=5, headers=headers)
+        except requests.RequestException:
```
tristanls, line 300: **issue**: This will catch `requests.RequestException` thrown from `requests.get` on the line above.

tristanls (commented):
Thank you, good thinking. Looks like an improvement.

**issue**: The `timeit` functionality is unused. For timing in this pull request, use `time.perf_counter()` instead.

README.md
```diff
@@ -49,21 +49,21 @@ If you want to use this code, contribute to it, ask questions or propose ideas,
 
 [![](docs/figures/overview/discourse_screenshot.png)](https://thousandbrains.discourse.group/)
 
-If you would like to receive updates, follow us on [Bluesky](https://bluesky.social/thousandbrainsproject) or [Twitter](https://x.com/1000brainsproj) or [LinkedIn](https://www.linkedin.com/showcase/thousand-brains-project/).
+If you would like to receive updates, follow us on [Bluesky](https://bsky.app/thousandbrainsproject) or [Twitter](https://x.com/1000brainsproj) or [LinkedIn](https://www.linkedin.com/company/thousand-brains-project/).
```
tristanls, line 52: **issue (non-blocking)**: I believe the Bluesky link is: https://bsky.app/profile/1000brainsproj.bsky.social

tools/github_readme_sync/hierarchy.py
```diff
@@ -320,6 +320,35 @@ def check_external_link(url):
...
+        requests.RequestException: If either HEAD or GET request fails
+    """
+    headers = request_headers()
+    try:
+        response = requests.head(url, timeout=5, headers=headers)
+    except requests.RequestException as e:
+        try:
+            response = requests.get(url, timeout=5, headers=headers)
+        except requests.RequestException as e:
+            raise requests.RequestException(f"Failed to connect to {url}") from e
```
tristanls, line 342: **question (non-blocking)**: Why this additional error wrapping?

tools/github_readme_sync/hierarchy.py
```diff
@@ -283,18 +298,57 @@ def check_readme_link(url, rdme):
...
+    """Make HTTP request to URL, trying HEAD first then falling back to GET.
+
+    Args:
+        url: The URL to request
+
+    Returns:
+        requests.Response: The response from the server
+
+    Raises:
+        requests.RequestException: If either HEAD or GET request fails
```
tristanls, line 333: **issue (non-blocking)**: I don't think this is accurate. The raise will only happen if the GET request fails.

tools/github_readme_sync/hierarchy.py
```diff
@@ -283,18 +298,57 @@ def check_readme_link(url, rdme):
...
-            logging.debug(f"{WHITE}  {url} ({response.status_code}){RESET}")
             return [f"  broken link: {url} ({response.status_code})"]
     except requests.RequestException as e:
         return [f"  {url}: {str(e)}"]
 
     return []
 
 
+def check_url(url):
+    """Make HTTP request to URL, trying HEAD first then falling back to GET.
```
tristanls, line 324: **issue (non-blocking)**: This code was difficult to understand, and the description didn't help me. I feel something along these lines would help: "Check if the URL exists. First, try HEAD. If that fails or returns a non-200 response, try GET. If GET fails, fail the check."

tristanls (approved):
**note (non-blocking)**: Looks good to me. I raised some non-blocking documentation issues.

README.md
codeallthethingz: oh, I think this was a bad merge.  Good catch.

PR #69, opened 2024-11-25
docs: GitHub Actions naming convention in style guide

tristanls:
This pull request adds the current GitHub Actions naming convention to the Style Guide for reference.

### DIFF
--- a/docs/contributing/style-guide.md
+++ b/docs/contributing/style-guide.md
@@ -53,6 +53,58 @@ In other words, if you are creating a new file, add the copyright and license he
 > [!NOTE]
 > While we deeply value and appreciate every contribution, the source code file header is reserved for essential copyright and license information and will not be used for contributor acknowledgments.
 
+# GitHub Actions
+
+We use GitHub Actions to run our continuous integration workflows.
+
+## GitHub Actions Naming Convention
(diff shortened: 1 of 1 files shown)


docs/contributing/style-guide.md
```diff
@@ -53,6 +53,58 @@ In other words, if you are creating a new file, add the copyright and license he
...
+
+```yml
+name: Potato Stuff
+```
+
+### Job Name
+
+The job name when in position of a key in a `jobs:` dictionary is a human-readable snake_case ending with `_<workflow_name>`.
+
+When used as a value for the `name:` property, the job name is human-readable kebab-case ending with `-<workflow-name>`, e.g.,
```
codeallthethingz, line 85: Why kebab-case?  Why don't we just use snake_case everywhere?

codeallthethingz (approved):
Approved with question

docs/contributing/style-guide.md
tristanls: This is one of those path-dependent arbitrary choices. In our previous CI setup, the tasks were named in kebab-case, so they remained in kebab-case by default. I'm not opposed to renaming them.

PR #71, opened 2024-11-26
ci: always try to download api docs artifact

tristanls:
Another attempt at Monty API Docs workflow.

In the last attempt, I learned that Monty API Docs workflow could not use `git diff --name-only HEAD^1 > changed_files.txt` to detect changed files because the latest commit in `main` might not be the one that generated the docs.

Screenshot 2024-11-26 at 09 20 42

So, instead of relying on commit diff criteria, this workflow attempts to download the artifact generated by any workflow that should run. If the artifact is missing, the workflow will exit without error. It is valid to use the presence of the artifact as "should run" criteria since we check that the previous workflow conclusion was `success`.

As with previous pull requests for this workflow, it needs to be merged to `main` before it will run.

### DIFF
--- a/.github/workflows/monty_api_docs.yml
+++ b/.github/workflows/monty_api_docs.yml
@@ -34,28 +34,31 @@ jobs:
           fetch-depth: 0
           lfs: true
           path: tbp.monty
-      # We want to deploy the docs using the same criteria for deciding if Monty code changed.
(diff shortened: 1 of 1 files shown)


tristanls:
After additional thought, I am going to rework the `git diff` to always work, as the problematic scenario could occur for existing uses of "should run".

PR #72, opened 2024-11-26
ci: use commit sha instead of HEAD to check changes on push

tristanls:
This pull request improves the robustness of "should run" checks. As noted in #71, if there are two closely timed `push` events, the `git diff HEAD^1` will always diff against the latest commit instead of the one for which the check runs. Now, the check compares the desired commit against the previous one.


codeallthethingz (approved):
Solid!

PR #73, opened 2024-11-26
ci: should_run_monty action explicit inputs

tristanls:
When resuing the `should_run_monty` action, the context it is running in can change. For example, when triggered by a `workflow_run` event, there is no `github.sha` in context. This pull request makes all inputs explicit so that the action will work independently of context.

### DIFF
--- a/.github/actions/should_run_monty/action.yml
+++ b/.github/actions/should_run_monty/action.yml
@@ -2,6 +2,13 @@ name: "Should Run Monty"
 description: "Determines if the Monty workflow should run based on the changed files."
 
 inputs:
+  event_name:
+    description: "The event name."
+    required: true
+  git_base_ref:
+    description: "The git base ref to compare against."
+  git_sha:
+    description: "The git sha to compare against."
   working_directory:
     description: "The directory to run the command in."
     required: true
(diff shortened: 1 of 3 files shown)


.github/actions/should_run_monty/action.yml
```diff
@@ -15,20 +22,20 @@ runs:
   using: "composite"
   steps:
     - name: Filter by path for a pull request
-      if: ${{ github.event_name == 'pull_request' }}
+      if: ${{ inputs.event_name == 'pull_request' }}
```
codeallthethingz, line 25: should this be inputs.git_event_name to follow the convention of the other github passthroughs?

codeallthethingz (approved):
Approved with suggestion.

.github/actions/should_run_monty/action.yml
tristanls: I don't think so. I started with `github_sha` and `github_base_ref`, but then renamed to `git_sha` and `git_base_ref` because `sha` and `base_ref` are `git` concepts (not GitHub). A pull request and an event name are GitHub concepts. I could rename it to `github_event_name`.

tristanls: Renamed to `github_event_name`.

PR #74, opened 2024-11-26
ci: generate pages artifact in Monty API Docs workflow

tristanls:
After #73, I learned that the `actions/download-artifact@v4` cannot find an artifact uploaded with `actions/upload-pages-artifact@v3` (or at least I haven't found a way to configure it correctly).

This pull request uploads a "normal" Monty API Docs artifact via `actions/upload-artifact@v4`, so that `actions/download-artifact@v4` can find it. This "normal" artifact is then downloaded and the GitHub Pages artifact is created and uploaded in the Monty API Docs workflow and then deployed.

### DIFF
--- a/.github/workflows/monty.yml
+++ b/.github/workflows/monty.yml
@@ -56,27 +56,6 @@ jobs:
         with:
           name: sphinx-html-${{ github.sha }}
           path: tbp.monty/tools/generate_api_docs/build/html
-      - name: Debug - Delete generated API docs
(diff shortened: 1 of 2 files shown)


codeallthethingz (approved):

codeallthethingz (approved):
This seemed like such a simple thing last week and has turned into a marathon. :)

tristanls:
😄 I still don't know if it'll work 🤞.

PR #76, opened 2024-11-26
Docs: Glossary Addition for Model

codeallthethingz:
While reviewing a research meeting a point of misunderstanding came up around the use of the word Model, which has a distinct meaning in TBP that contrasts with deep learning.  This glossary item hopes to clarify this.


mleies (commented):
Hey Will, 
Looks good.   

Not sure if it's the appropriate place to add more about the differences vs deep learning as an example  i'd be tempted to mention that in monty there are/will be many models, and that learning in one model is independent of learning in another (probably want someone more dialed in than me to sanity check that statement.)   In deep learning a single model learns representations for all objects, but there is no independence, learning to recognize and new object may alter previously learnt objects.

mleies (approved):

vkakerbeck (approved):
Nice, thanks for adding this! Another point you could add is that Monty may have multiple models of the same object (if there is more than 1 LM). Also, that the models are structured by reference frames (not just a bag of features) and that they represent complete objects (not just basic features like edges).

vkakerbeck:
You could also add a link to here for further reading: https://thousandbrainsproject.readme.io/docs/how-learning-modules-work#object-models
And to this FAQ question: https://thousandbrainsproject.readme.io/docs/faq-monty#do-cortical-columns-in-the-brain-really-model-whole-objects-like-a-coffee-mug-in-v1

nielsleadholm (approved):
Great addition, thanks Will.

docs/overview/glossary.md
```diff
@@ -32,7 +32,17 @@ A computational unit that takes features at poses as input and uses this informa
...
+In Monty, a model (sometimes referred to as [Object Model](../how-monty-works/how-learning-modules-work.md#object-models)), is a representation of an object stored entirely within the boundaries of a learning module. The notion of a model in Monty differs from the concept of a deep learning neural network model in several ways:
+
+- A single learning module **stores multiple object models** in memory, simultaneously.
+- The Monty system may have **multiple models of the same object** if there are multiple learning modules - this is a desired behavior.
+- Learning modules **update models independently** of each other.
+- Models are structured as [reference frames](#reference-frame), not just a bag of features.
```
vkakerbeck, line 40:
```suggestion
- Models are structured using [reference frames](#reference-frame), not just a bag of features.
```

docs/overview/glossary.md
```diff
@@ -32,7 +32,17 @@ A computational unit that takes features at poses as input and uses this informa
...
+- Models represent **complete objects**, not just parts of objects.
```
vkakerbeck, line 41:
```suggestion
- Models represent **complete objects**, not just parts of objects. These objects can still become subcomponents of compositional objects but are also objects themselves (like the light bulb in a lamp).
```

codeallthethingz:
@vkakerbeck I made some changes that I think require a re-review.

vkakerbeck (approved):
Nice changes :) I just added some minor suggestions but feel free to ignore, especially the second one, if you think it adds more confusion than clarification.

PR #77, opened 2024-11-26
fix: export habitat actuator symbols that are used as types

tristanls:
I noticed that not all [HabitatActuator](https://github.com/thousandbrainsproject/tbp.monty/blob/d5631219e86f1fbaa69a8b96dcacd118b023c493/src/tbp/monty/simulators/habitat/actuator.py#L49) symbols appear in the API documentation. This is because they were not exported as part of `__all__`. This pull request exports the symbols that are used as types.


vkakerbeck (approved):
Good catch, thanks!

PR #80, opened 2024-11-27
Docs: Align images

codeallthethingz:
My markdown previewer lied to me about how the images would display.  This forces the images onto a new line.

From this:

image

to this:

image


vkakerbeck (approved):
nice :) Can we also make them centered like it was in your preview?

codeallthethingz:
> nice :) Can we also make them centered like it was in your preview?

Oh yes.  I'll have to convert them to image tags.  That happens automatically for our docs but not for the readme.md in the root.

vkakerbeck (approved):
nice, thanks!

codeallthethingz:
> nice :) Can we also make them centered like it was in your preview?

We cannot.  If we put HTML around the images to center them, we'd have to put absolute URLs in for the images, if we do that, the PR view of the page will no longer render the images.  As far as I know there is no good way to deal with this in Github.  But I could be wrong.

vkakerbeck:
Okay, no big deal, was just wondering

PR #84, opened 2024-11-29
ci: grant pull-requests write to cla workflow

tristanls:
I am trying to understand why the CLA workflow fails to write comments on a pull request. This is an attempt to see if it requires `pull-request: write` permission despite using the Issues write endpoint.


codeallthethingz (approved):
LGTM

PR #87, opened 2024-12-02
Add future work documentation

nielsleadholm:
Adds descriptions for many of the outstanding "future-work" areas of our documentation. This includes a few new sections:
- simple-cross-modal-policy
- improve-handling-of-symmetry
- reuse-hypothesis-testing-points
- support-scale-invariance

"Bottom-up distant agent policies" was removed because it was a duplicate, while a few others appear "removed" because of changes to their names/fixes to typos.

A few of the ones that I haven't done as I was a bit unsure what we had previously discussed, are:
- Improve Bounded Evidence Performance
- Make it Possible to Store Multiple Feature Maps on one Graph
- Re-Anchor Hypotheses
- Less Dependency on First Observation
- Deal with Incomplete Models

@scottcanoe and @hlee9212 tagging you for any thoughts you want to add and as this also gives an overview of some of the things we can work on soon.


docs/future-work/learning-module-improvements/use-models-with-less-points.md
```diff
@@ -1,3 +1,5 @@
 ---
 title: Use Models with Less Points
```
scottcanoe, line 2: I hate to be that guy, but "Fewer" is also an option here.

scottcanoe (approved):
Looks great, I'm excited to get to work on these!

docs/future-work/learning-module-improvements/use-models-with-less-points.md
nielsleadholm: Haha fair point, thanks

docs/future-work/motor-system-improvements/reuse-hypothesis-testing-points.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Reuse Hypothesis-Testing Policy Target Points
+---
+
+The hypothesis-testing policy is able to generate candidate points on an object that, when observed, should rapidly disambiguate between similar objects, or between similar poses of the same object.
+
+Generating these points requires a model-based policy that simulates the overlap in the graphs between the two most likely objects (or the two most likely poses of the same object). This is a relatively expensive operation, and so one approach would be to store these points in long-term memory, reusing them in future episodes.
+
+For example, when we have first learn about the concept of a mug, we might need to deliberately think about the fact that its handle is what distinguishes it from many other cylindrical objects. However, once we have experienced recognizing mugs a few times, we could quickly recall that testing the handle is a good way to confirm whether we are sensing a mug, or some other object.
```
vkakerbeck, line 9: Another way to determine ROIs on an object could be to use the evidence update as a signal. So basically locations on an object that maximally disambiguate the object from others would become hot-spots to move to next time that object is hypothesized.

docs/future-work/sensor-module-improvements/detect-local-and-global-flow.md
```diff
@@ -1,3 +1,9 @@
 ---
 title: Detect Local and Global Flow
 ---
+
+Our general view is that there are two sources of flow processed by cortical columns. A larger receptive field sensor helps to estimate global flow, where flow here will be particularly pronounced if the whole object is moving, or the sensor itself is moving. A small receptive-field sensor patch corresponds to the channel by which the primary sensory features (e.g. point-normal, color) arrive. If flow is detected here, but not in the more global channel, then it is likely that just part of the object is moving.
```
vkakerbeck, line 5: I would make the distinction more clear:
local flow - object is moving
global flow - sensor is moving

We should also mention that these may not be detectable with the same sensor (small patch can't distinguish between object and sensor movement since for the patch both of it would be global flow.

docs/future-work/sensor-module-improvements/extract-better-features.md
```diff
@@ -1,3 +1,9 @@
 ---
 title: Extract Better Features
 ---
+
+Currently non-morphological features are very simple, such as extracting the RGB or hue value at the center of the sensor patch.
+
+In the short term, we would like to extract richer features, such as using HTM's spatial-pooler or Local Binary Patterns for visual features, or processing depth information within a patch to approximate tactile texture.
+
+In the longer-term, given the "sub-cortical" nature of this sensory processing, we might also consider neural-network based feature extraction, such as shallow convolutional neural networks, however please see [our FAQ on why Monty does not currently use deep learning](../../how-monty-works/faq-monty.md#why-does-monty-not-make-use-of-deep-learning).
```
vkakerbeck, line 9: It would also be worth mentioning that extracted features should be rotation invariant. So if we look at the same location on an object from different angles, the extracted feature should be the same. This is not a given with neural networks or many other approaches.

docs/future-work/voting-improvements/can-we-change-the-cmp-to-use-displacements-instead-of-locations.md
```diff
@@ -1,3 +1,9 @@
 ---
 title: Can we change the CMP to use displacements instead of locations?
 ---
+
+Movement is core to how LMs process and model the world. Currently, an LM receives an observation encoded with a body-centric location, and then infers a displacement in object-centric coordinates. Similarly, goal-states are specified as a target location in body-centric coordinates, which are then acted upon.
+
+However, a more general formulation might be to use displacements as the core spatial information in the CMP, such that a specific location (in body-centric coordinates or otherwise) is never communicated outside of an LM or sensor module.
+
+Such an approach might align well with adding information about flow (see [Detect Local and Global Flow](../sensor-module-improvements/detect-local-and-global-flow.md)), modeling moving objects (see [Deal With Moving Objects](../learning-module-improvements/deal-with-moving-objects.md)), and supporting abstract movements like the transition from grandchild to grandparent.
```
vkakerbeck, line 9: This approach will be very tricky since if we don't get the location of the sensor relative to the body it is almost impossible to output anything in a common reference frame and therefor vote or send other outputs. We could do basic voting on object ID and rely on colocation of receptive fields in the hierarchy but also motor commands will be pretty much impossible this way. We could mention that this could allude to the difference between the where and what pathway. But also that this is not something we plan to implement but merely a possibility we want to investigate further.

docs/future-work/voting-improvements/generalize-voting-to-associative-connections.md
```diff
@@ -1,4 +1,8 @@
 ---
 title: Generalize Voting to Associative Connections
 ---
-Needed when each LM learns different representations of the same object (object_id in the code) which is the case in the brain.
\ No newline at end of file
+Currently, voting relies on all learning modules sharing the same object ID for any given object, as a form of supervised learning signal. Thanks to this, they can vote on this particular ID when communicating with one-another.
+
+However, in the setting of unsupervised learning, the object ID that is associated with any given model is unique to the parent LM. As such, we need to organically learn the mapping between the object IDs that occur together across different LMs, such that voting can function without any supervised learning signal.
```
vkakerbeck, line 6: You could also mention that the brain has to do the same thing. It could not use a globally consistent SDR representation for each object. The neurons just do associative learning and a cortical column has no idea what the incoming spikes mean.

docs/future-work/voting-improvements/use-pose-for-voting.md
```diff
@@ -1,3 +1,9 @@
 ---
 title: Use Pose for Voting
 ---
+
+Currently we do not use the pose of the objects when we are voting, however we believe it will be an important signal to use. One complication is that the poses stored for any given LM's object models are arbitrary with respect to other LM's models, as each uses an object-centric coordinate system.
```
vkakerbeck, line 5:
```suggestion
Currently we do not send out pose hypotheses when we are voting, however we believe it will be an important signal to use. One complication is that the poses stored for any given LM's object models are arbitrary with respect to other LM's models, as each uses an object-centric coordinate system.
```

vkakerbeck:
Oh, lastly, it would be good if you can update this sheet https://docs.google.com/spreadsheets/d/10b0FR9YdFYqfhIiGMpZsjmN2OAbNAjp4m_hLBCV161I/edit?gid=0#gid=0 with the topics you added/renamed so they match exactly

docs/future-work/learning-module-improvements/use-models-with-fewer-points.md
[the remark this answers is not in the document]
vkakerbeck: Do you mean like we will do when we have hierarchy? Basically in the higher-level LM, the objects modeled in the lower-level LM become features in that graph? So the switch, light bulb and lambp shade modeled in the LLLM become features in the lamp model in the HLLM?
One thing I just want to make sure you are not forgetting (you probably aren't but since you didn't mention it in your comment) is that we always model features AT LOCATIONS. And the relative locations of features are actually more important than the features themselves (think of how you can easily recognize a face made of fruits but wouldn't call a random assortment of noses and eyes a face). We always use reference frame, not just a bag of features. 

As for using less points: Yes, this would be an efficiency and generalization point. At the very beginning we would store every point in our models which would give us perfect accuracy but it was slow. So over time we figured several ways to use less points (feature change SM, graph_delta_thresholds, ...) These give us significant efficiency gains with little loss in accuracy and there are still a few more ways we can make our models more efficient by using even less points. This also relates a bit to hierarchy since in the larger objects composed of sub objects we would likely want to store less points.

vkakerbeck:
We have all of them listed out here https://docs.google.com/spreadsheets/d/10b0FR9YdFYqfhIiGMpZsjmN2OAbNAjp4m_hLBCV161I/edit?gid=0#gid=0 and some are already grouped into our next milestones (below the table). We will likely create 1-2 new, intermediate milestones to prepare for the Heterarchy pt. 2 milestone (and eventually hierarchical goal policies).

docs/future-work/cmp-hierarchy-improvements/add-associative-connections.md
```diff
@@ -1,4 +1,9 @@
 ---
 title: Add Associative Connections
```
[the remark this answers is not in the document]
nielsleadholm: Ah ok that makes sense, I was confused by the potential duplication (I think because I focused on the term "hierarchy" in the `cmp-hierarchy` grouping).

With that cleared up, I wonder if it's a bit of a duplication of "Generalize Voting to Associative Connections" --> my temptation would be to keep that one, and add the point that this should enable associating e.g. sound objects with physical objects (i.e. where their models may not both be 3D), and get rid of "Add Associative Connections" under cmp-hierarchy. What do you think @vkakerbeck ?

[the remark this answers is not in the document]
vkakerbeck: wow yes! That only now clicked for me that those two are basically the same. Its kind of cool that we can solve both these with the same solution. I think I had added this one under hierarchy because the first time I thought about these was in the context of modeling language and grounding it in physical models of objects. But I think we should just remove this one and expand on the one under voting like you suggest. Maybe add the "abstract" or "num_steps" label to it

docs/future-work/cmp-hierarchy-improvements/figure-out-performance-measure-and-supervision-in-heterarchy.md
```diff
@@ -1,3 +1,9 @@
 ---
-title: Figure out Performance Measure and Supervision in Heterarchy
+title: Figure out Performance Measures and Supervision in Heterarchy
 ---
+
+As we introduce hierarchy and leverage more unsupervised learning, representations will emerge at different levels of the system that may not correspond to any labels present in our datasets. For example, handles, or the head of a spoon, may emerge as object-representations in low-level LMs, even though the dataset only recognizes labels like "mug" and "spoon".
+
+One approach to measure the "correctness" of representations in this setting might be how well a predicted representation aligns with the outside world. For example, while LMs are not designed to be used as generative models, we could visualize how well an inferred object graph maps onto the object actually present in the world. Quantifying such alignment might leverage measures such as differences in point-clouds. This would provide some evidence of how well the learned decomposition of objects corresponds to the actual objects present in the world.
```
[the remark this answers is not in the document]
nielsleadholm: Good point, I've added those items to the start.

docs/future-work/cmp-hierarchy-improvements/send-similarity-encoding-object-id-to-next-level-test.md
```diff
@@ -1,3 +1,11 @@
 ---
 title: Send Similarity Encoding Object ID to Next Level & Test
 ---
+
+We have implemented the ability to encode object IDs using sparse-distributed representations (SDRs), and in particular can use this as a way of capturing similarity and disimlarity between objects. Using such encodings in learned [Associative Connections](add-associative-connections.md), we should observe a degree of natural generalization when recognizing compositional objects.
```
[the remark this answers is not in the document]
nielsleadholm: Yeah I've changed this to Hierarchical Connections, per the earlier discussion.

docs/future-work/cmp-hierarchy-improvements/send-similarity-encoding-object-id-to-next-level-test.md
```diff
@@ -1,3 +1,11 @@
...
+
+For example, assume a Monty system learns a dinner table setting with normal cuttlery and plates. Separately, the system learns about medieval instances of cuttlery and plates, but never sees them arranged in a dinner table setting. Based on the similarity of the medieval cutterly objects to their modern counterparts, the objects should have considerable overlap in their SDR encodings.
+
+If the system was to then see a medieval dinner table setting for the first time, it should be able to recognize the arrangement as a dinner-table setting with reasonable confidence, even if the constituent objects are somewhat different from those present when the compositional object was first learned.
```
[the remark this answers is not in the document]
nielsleadholm: Good point! Adding

docs/future-work/learning-module-improvements/bias-rotation-hypotheses.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Bias Rotation Hypotheses by Common Object Poses
```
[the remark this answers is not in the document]
nielsleadholm: Good idea, and thanks for catching that

docs/future-work/learning-module-improvements/deal-with-moving-objects.md
```diff
@@ -1,3 +1,9 @@
 ---
 title: Deal with Moving Objects
 ---
+
+This work relates to first being able to [Detect Local and Global Flow](../../future-work/sensor-module-improvements/detect-local-and-global-flow.md).
+
+Our current idea is to then use this information to model the state of the object, such that beyond its current pose, we also capture how it is moving as a function of time. This information can then be made available to other learning modules for voting and hierarchical processing.
```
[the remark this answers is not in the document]
nielsleadholm: Yeah I guess it just feels like velocity could be something more fundamental, i.e. that might be part of all CMP messages once we have flow etc. working? I.e. that things like color are optional, but whether something is moving or not is as central as its pose? But might not end up being so.

docs/future-work/learning-module-improvements/test-particle-filter-like-resampling-of-hypothesis-space.md
```diff
@@ -1,3 +1,9 @@
 ---
 title: Test Particle-Filter-Like Resampling of Hypothesis Space
 ---
+
+When the evidence values for a point in an LM's graph falls below a certain threshold, we generally stop testing it. Furthermore, the initial feature pose detected when the object was first sensed determines the pose hypotheses that are initialized.
+
+We would like to implement a method to randomly initialize a subset of rejected hypotheses, and then test these. This relates to [Less Dependency on First Observation](less-dependency-on-first-observation.md).
+
+This work could also tie in with the ability to [Bias Rotation Hypotheses by Common Object Poses](bias-rotation-hypotheses.md), as these might be poses that an LM would resample more than others.
```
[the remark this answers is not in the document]
nielsleadholm: Good point, I've updated this.

docs/future-work/motor-system-improvements.md
```diff
@@ -5,13 +5,14 @@ description: Improvements we would like to add to the motor system.
...
+- [Calculate saliency map in view finder and use it to saccade to region of interest ](motor-system-improvements/implement-efficient-saccades-driven-by-model-free-and-model-based-signals.md)#numsteps #multiobj
 - [Learn policy using RL](motor-system-improvements/learn-policy-using-rl.md) (using simplified action space) #numsteps
 - [Decompose goals into subgoals & communicate](motor-system-improvements/decompose-goals-into-subgoals-communicate.md) #goalpolicy
-- [Interprete goal states in motor system & switch policies](motor-system-improvements/interprete-goal-states-in-motor-system-switch-policies.md) #goalpolicy
+- [Interprete goal states in motor system & switch policies](motor-system-improvements/interpret-goal-states-in-motor-system-switch-policies.md) #goalpolicy
+- [Reuse hypothesis testing points](motor-system-improvements/reuse-hypothesis-testing-points.md) #goalpolicy #numsteps
+- [Implement a simple cross-modal policy](motor-system-improvements/simple-cross-modal-policy.md) #crossmodal
```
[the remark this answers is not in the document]
nielsleadholm: Oh good catch, that was a Cursor hallucination. I've changed it to `#learning #multiobj #goalpolicy #numsteps` - I feel learning in the sense that we can support learning a detailed model that we might only be able to see one angle, and this could help in learning in multi-object settings. `#goalpolicy` because of the policy element, and maybe `numsteps` more than speed since it's more about convergence than computational optimization?

docs/future-work/motor-system-improvements/reuse-hypothesis-testing-points.md
nielsleadholm: Yeah good idea!

docs/future-work/sensor-module-improvements/detect-local-and-global-flow.md
nielsleadholm: Ok yeah that's what I was trying to get at (the uncertainty depending on the size), will try rewording it.

nielsleadholm:
Thanks for the helpful comments @vkakerbeck and @hlee9212 ! Those should all be addressed now but let me know if there are any further changes you want. I'll also now double check the overview spreadsheet and make sure all the cells are updated to match here.

nielsleadholm:
@vkakerbeck I've also updated some of the hashtags where I felt some were missing.

nielsleadholm:
Lastly I've gone through and made sure the names for future-work sections are consistent across the individual articles, the header .md files, and [the overview sheet](https://docs.google.com/spreadsheets/d/10b0FR9YdFYqfhIiGMpZsjmN2OAbNAjp4m_hLBCV161I/edit?gid=0#gid=0).

docs/future-work/learning-module-improvements/deal-with-moving-objects.md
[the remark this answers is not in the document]
vkakerbeck: Hm yeah interesting question. I'm not sure how velocity would apply to more abstract objects/concepts but maybe it can be. I was just thinking if it is one of many optional features we get maximum flexibility but maybe that's not needed.

docs/future-work/cmp-hierarchy-improvements/add-top-down-connections.md
```diff
@@ -2,4 +2,8 @@
 title: Add Top-Down Connections
 ---
 
-One of the main roles of top-down connections is the associative recall and prediction outlined in [Associative Connections](add-associative-connections.md). However, top-down projections can also support decomposing goal-states into specific sub-goals, as discussed in [Decomposing Goal States](../motor-system-improvements/decompose-goals-into-subgoals-communicate.md).
\ No newline at end of file
+In Monty systems, low-level LMs project to high-level LMs, where this projection occurs if their sensory receptive fields are co-aligned. Hierarchical connections should be able to learn a mapping between objects represented at these low-level LMs, and objects represented in the high-level LMs that frequently co-occur. Such learning would be similar to that required for [Generalizing Voting To Associative Connections](../voting-improvements/generalize-voting-to-associative-connections.md).
```
vkakerbeck, line 5: I wouldn't compare it to the associative voting connections. The voting connections are between LMs that are modeling the same object. The hierarchical connections are between LMs that model a compositional object and its sub-components. So in the bottom-up connection the lower LMs object ID becomes a feature in the higher level LM. In the top down connection, we associate a feature in the higher level LM with an object in a lower level LM (+ a location). I understand where you are coming from in the sense that this is also just a learned association but maybe its still a bit confusing to read for someone who doesn't have all this background. Maybe we could just remove the reference to the voting plan since it won't be the exact same mechanism.

docs/future-work/voting-improvements/can-we-change-the-cmp-to-use-displacements-instead-of-locations.md
```diff
@@ -6,4 +6,6 @@ Movement is core to how LMs process and model the world. Currently, an LM receiv
...
\ No newline at end of file
+Such an approach might align well with adding information about flow (see [Detect Local and Global Flow](../sensor-module-improvements/detect-local-and-global-flow.md)), modeling moving objects (see [Deal With Moving Objects](../learning-module-improvements/deal-with-moving-objects.md)), and supporting abstract movements like the transition from grandchild to grandparent. It would also result in a reformulation of "goal-states" to "goal-displacements".
+
+Note that whatever approach is taken, we would need to have information about the location or displacement of the sensor to ensure that communicated displacements are still in a shared coordinate system. This may relate to the division of "what" and "where" pathways in the brain, although this is not yet clear and requires further investigation.
```
vkakerbeck, line 11: I don't see how displacements could be in a shared coordinate system. If we only communicate displacements, there is no way of knowing where the movements are relative to each other. I still think some information about location in a common reference frame needs to be present. maybe like you mention the difference between where and what pathways

vkakerbeck (approved):
Nice, thanks for the updates! Just two minor comments but overall looks great!

docs/future-work/cmp-hierarchy-improvements/add-top-down-connections.md
nielsleadholm: Yeah fair point, thanks

docs/future-work/voting-improvements/can-we-change-the-cmp-to-use-displacements-instead-of-locations.md
vkakerbeck: yeah for a single LM recognizing objects that works but I don't see how that works when multiple LMs start communicating or when we want to send goal states. How would you do voting or goal states without a shared reference frame of where sensors are relative to some common reference point?

nielsleadholm: Hmm yeah that's a good point, now that I think about it more, it does feel like location is required for to some degree, I've updated the wording.

PR #91, opened 2024-12-05
Updated to fix "No module named 'lark'"

bobcowher:
Encountered the error below when running pytest after the initial setup. Added "lark" to the pyproject.toml to resolve. 

ModuleNotFoundError: No module named 'lark'

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -26,7 +26,6 @@ dependencies:
   - cmake>=3.14.0
   - pyg::pyg
   - wget
-  - lark=1.2.2
 
   - aihabitat::habitat-sim=0.2.2
   - pytorch::pytorch=1.11.0
--- a/environment.yml
+++ b/environment.yml
@@ -26,6 +26,7 @@ dependencies:
   - cmake>=3.14.0
   - pyg::pyg
   - wget
+  - lark=1.2.2
 
   - aihabitat::habitat-sim=0.2.2
   - pytorch::pytorch=1.11.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
     'torch-geometric',
     'tqdm',
     'wandb',
+    'lark',
 ]
 description = 'Numenta Monty'
 dynamic = ['version']


tristanls:
**issue (blocking)**: I would like to first understand what happened before introducing a dependency.

Hi @bobcowher,

What commands did you run to setup Monty? What operating system are you on? What command did you run that resulted in the error?

bobcowher:
@tristanls  Great question. I'm running on PopOS 22.04 LTS, which is closely related to Ubuntu. As for steps, I cloned the project and ran the following in the workspace - 

conda env create
conda init zsh
conda activate tbp.monty

pytest

Basically, I'm just running through the default setup instructions. 

I'm curious - If you run a pip list, do you see lark installed? I'm wondering if there's something about my setup that requires lark OR if it's that everyone else already has lark and I don't.

tristanls:
Hi @bobcowher,

I'm sorry that the instructions didn't work out of the box.

I was surprised that something would require `lark`.

On my Mac:

```
(tbp.monty) tbp.monty % pip list | grep lark
(tbp.monty) tbp.monty % conda list | grep lark
```

We also list dependencies on Ubuntu. We cache heavily, but here's a run that listed the dependencies when installing tbp.monty on Ubuntu: https://github.com/thousandbrainsproject/tbp.monty/actions/runs/11903517066/job/33170672192. It also has no `lark`.

I'm at a loss as to what requires it.

Can you run `pip show lark` and see what's in the `Required-by: ` field?

bobcowher:
I figured it out, and it's a PBCAK error. If you check the lark.txt file here, with the full output, it looks like that error is actually being hit by a Python script under my ROS Humble installation, and this error happens anywhere I run pytest. It's just that I'm not using pytest anywhere else, so I assumed it had something to do with this project. 

Sorry for the time waste. I'm leaving this out there so you can see the results, but please feel free to close this PR after you do. Thanks for the help.  

[lark.txt](https://github.com/user-attachments/files/18046022/lark.txt)

tristanls:
Thank you for figuring this out, @bobcowher. As an interface for simulators, [Monty may want to support ROS in the future](https://thousandbrains.discourse.group/t/the-next-monty-3d-environment-simulator/288), so this is still helpful information.

PR #94, opened 2024-12-06
Utilities for generating `monty_config` and `dataset_args` for an arbitrary number of LMs.

scottcanoe:
This adds a few functions to `config_args.py` and `make_dataset_configs.py` that allow users to create multi-LM experiment configs without writing and using custom config data classes.

Added to `config_args.py`
 - `make_multi_lm_monty_config`: The top-level utility function. Builds a `MontyConfig` object given some the number of LMs and Monty/SM/LM classes with their associated arguments.
 - `make_multi_lm_connectivity`: generates `sm_to_agent_dict`, `sm_to_lm_matrix`, `lm_to_lm_matrix`, and `lm_to_lm_vote_matrix` for "typical" multi-LM setups (i.e., 1 agent, 1:1 SM -> LM connectivity, all-to-all voting, and no hierarchy). Used by `make_multi_lm_monty_config` and may be useful if you write your `MontyConfig` generator.

Added to `make_dataset_configs.py`
 - `make_multi_lm_habitat_dataset_args`: The top-level utility function. Builds a `MultiLMMountHabitatDatasetArgs` object given the number of SMs and optionally some mount parameters (such as sensor positions, zooms, etc.).
 - `make_multi_lm_mount_config`: Builds a dictionary of mount parameters (positions, resolutions, etc.).
 - `make_multi_lm_sensor_positions`: Generates sensor positions. See my comment for more info.

Here's a usage example:
```python

# Set up arguments for `make_multi_lm_monty_config`.
monty_config_args = [
    MontyForGraphMatching,  # monty_class
    DisplacementGraphLM,  # learning_module_class
    dict(k=5, match_attribute="displacement"),  # learning_module_args
    HabitatDistantPatchSM,  # sensor_module_class
    dict(features=["pose_vectors, "on_object, ...]),  # sensor_module_args
    NaiveScanPolicy,  # motor_system_class
    make_naive_scan_policy_config(step_size=5),  # motor_system_args
    dict(num_exploratory_steps=500),  # monty_args
]

pretrain_dist_agent_16lm = dict(
    experiment_class=MontySupervisedObjectPretrainingExperiment,
    experiment_args=ExperimentArgs(
        n_train_epochs=len(TRAIN_ROTATIONS),
        do_eval=False,
    ),
    logging_config=PretrainLoggingConfig(run_name="dist_agent_16lm"),
    monty_config=make_multi_lm_monty_config(16, *monty_config_args),  # <----------
    # Set up environment and agent.
    dataset_class=ED.EnvironmentDataset,
    dataset_args=make_multi_lm_habitat_dataset_args(16),  # <----------
    # Set up training dataloader.
    train_dataloader_class=ED.InformedEnvironmentDataLoader,
    train_dataloader_args=EnvironmentDataloaderPerObjectArgs(
        object_names=SHUFFLED_YCB_OBJECTS,
        object_init_sampler=PredefinedObjectInitializer(rotations=TRAIN_ROTATIONS),
    ),
)
```

I'm currently using a pattern like this in `dmc_pretraining_experiments.py` and `dmc_eval_experiments.py` (though I haven't submitted a monty_lab PR yet). It made it easy to add 4, 8, and 16 LM experiments and removed the need for a bunch of hand-made dataclasses (farewell to`TenLMMountHabitatDatasetArgs`, `EnvInitArgsTenLMMount`, and `TenLMMountConfig`).

Testing: I re-wrote DMC experiments to use these config generators. I then ran`dist_agent_5lm` and `dist_agent_5lm_randrot_noise`, and the results were identical. I've also pretrained models with 2, 4, 5, 8, 9, 10, and 16 LMs, and the object models look good.

16LMs


scottcanoe:
A note on generating sensor module positions: I've implemented two choices for determining how to place sensor modules. The function `make_multi_lm_sensor_positions` has the parameters `n_sms` (number of sensor modules), `delta` (grid spacing, default is 1 cm), and `order_by`. `ordery_by` can be either `"distance"` or `"spiral"`.
 - `"spiral"` is pretty straightforward. It starts at the center of a grid and spirals outward.
 - `"distance"` also spirals, but it prioritizes sensor module positions that are closest to the center.
 
For example, here are arrangements generated for 5 sensor modules:
n_sms_5_distance

Numbers indicate sensor module ID, and black squares don't have sensor modules (`"view_finder"` just indicates the last generated position which is always the same as sensor module 0).

Here's another for 16 sensor modules:
n_sms_16_distance

Though the `"distance"` method is a bit more jagged or aesthetically weird, I think it's the better choice since, in principal, it should reduce the occurrence of off-object observations for peripheral sensor modules. So I let `"distance"` by the default argument for `make_multi_lm_sensor_positions`.

I should note that you can always supply your own positions to `make_multi_lm_habitat_dataset_args`. For example, if you wanted to use the `"spiral"` method or shorter spacing between sensor modules, you can generate those with `make_multi_lm_sensor_positions` or create them manually and pass those positions into `make_multi_lm_habitat_dataset_args`.

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -1153,6 +1154,189 @@ class FiveLMMontySOTAConfig(FiveLMMontyConfig):
...
+            "sensor_module_id": "view_finder",
+            "save_raw_obs": True,
+        },
+    }
+
+    # Make motor system config.
+    if motor_system_args is None:
+        motor_system_args = {}
+    else:
+        motor_system_args = copy.deepcopy(motor_system_args)
```
vkakerbeck, line 1306: is this line needed? What does it do?

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -1153,6 +1154,189 @@ class FiveLMMontySOTAConfig(FiveLMMontyConfig):
...
+    motor_system_config = {
+        "motor_system_class": motor_system_class,
+        "motor_system_args": motor_system_args,
+    }
+
+    conn = make_multi_lm_connectivity(n_lms)
```
vkakerbeck, line 1312: I would just call this variable `connectivity` or even `connectivity_matrices`. There is enough space to spell variables out for clarity

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -1102,3 +1104,276 @@ class FiveLMMountConfig:
...
         default_factory=lambda: [10.0, 10.0, 10.0, 10.0, 10.0, 1.0]
     )
+
+
+"""
+Utilities for generating arbitrary multi-LM dataset args.
+"""
+
+
+def make_multi_lm_sensor_positions(
```
vkakerbeck, line 1114: Again, a more descriptive name could be useful if we have other generator functions later. Maybe something that expresses that they are all attached to one agent and they evenly cover a surface? Also, should it maybe be `multi_sensor_positions`?

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -1102,3 +1104,276 @@ class FiveLMMountConfig:
...
+            identical to row 0. Otherwise, the array has `n_sms` rows. row 0 is
+            always centered at (0, 0, 0), and all other rows are offset relative to it.
+
+    """
+    assert n_sms > 0, "n_lms must be greater than 0"
+    assert delta > 0, "delta must be greater than 0"
+    assert order_by in ["spiral", "distance"], "order_by must be 'spiral' or 'distance'"
+
+    # Find smallest square grid size that can fit n_lms with odd-length sides.
+    n = 1
```
vkakerbeck, line 1161: Can you give this variable a more descriptive name? What is n?

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -1102,3 +1104,276 @@ class FiveLMMountConfig:
...
+    while n_sms > n**2:
+        n += 2
+
+    # Make coordinate grids, where the center is (0, 0).
+    pts = np.arange(-n // 2 + 1, n // 2 + 1)
```
vkakerbeck, line 1166: also here I would not abbreviate. i assume it stands for "points"?

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -1102,3 +1104,276 @@ class FiveLMMountConfig:
...
+    # Should be (0, 0, 0).
+    if add_view_finder:
+        positions = np.vstack([positions, positions[0].reshape(1, -1)])
+
+    return positions
+
+
+def make_multi_lm_mount_config(
+    n_sms: int,
+    *,
```
vkakerbeck, line 1235: what is this for?

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -1102,3 +1104,276 @@ class FiveLMMountConfig:
...
+
+    # Optionally append entry for a view finder which is a duplicate of row zero.
+    # Should be (0, 0, 0).
+    if add_view_finder:
+        positions = np.vstack([positions, positions[0].reshape(1, -1)])
+
+    return positions
+
+
+def make_multi_lm_mount_config(
```
vkakerbeck, line 1233: maybe here again we make it a bit more specific and call it multi_sensor instead of multi_lm since the dataset args have nothing to do with lms

vkakerbeck (commented):
Nice, this looks very useful! I left a few comments, mostly around variable and function names. 
On a more general note, it would be nice if you can include the usage example from your PR description either in the documentation or in the docstring. In the usage example it would be good to specify the `monty_config_args` as a dictionary with keys instead of a list where it matters in which order they are specified and it is a lot less verbose (hence why you probably added comments)

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
tristanls: `/` and `*` are special symbols

Screenshot 2024-12-09 at 13 49 56

vkakerbeck: Why does it come before most of the other keyword args here? Everywhere else we use **args and **kwargs, is there a specific need to use just * here?

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -1153,6 +1154,189 @@ class FiveLMMontySOTAConfig(FiveLMMontyConfig):
     )
 
 
+"""
+Multi-LM Config Utils.
+"""
+
+
+def make_multi_lm_connectivity(n_lms: int) -> Dict:
```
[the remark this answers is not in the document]
scottcanoe: OK, I changed the name to `make_multi_lm_flat_dense_connectivity`.

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -1153,6 +1154,189 @@ class FiveLMMontySOTAConfig(FiveLMMontyConfig):
...
+    Helper function for creating monty configs for experiments having multiple sensor
+    and learning modules, and learning modules are not hierarchically connected.
+    Also assumes each LM is connected to a single upstream sensor module in 1:1 fashion,
+    and all sensor modules are mounted on a single agent.
+
+    The results of this function are primarily intended to be included in a
+    monty config dataclass/dictionary.
+
+    Args:
+        n_lms: Number of LMs. It is assumed that the number of sensor modules is equal
```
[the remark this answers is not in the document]
scottcanoe: I've updated the docstring to say "n_lms: Number of LMs. It is assumed that the number of sensor modules (not including a view finder) is equal to the number of LMs."

src/tbp/monty/frameworks/config_utils/config_args.py
scottcanoe: If we have an empty dictionary in the function signature (i., `motor_system_args={}`), the empty dictionary can be modified, and those changes will persist for future function calls. That's why I use `None` in the signature and replace it with an empty dict within the function ([example](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python)). Is that what you're referring to?

src/tbp/monty/frameworks/config_utils/config_args.py
scottcanoe: Done.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
scottcanoe: Yep, changed to `points`.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
scottcanoe: It's the number of rows/columns in the grid. I've changed it to `grid_size`.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
scottcanoe: What do you think of `multi_sm...` for these names?

In general, I went with `multi_lm` for everything across `make_dataset_args` and `config_args` since they all kinda basically belong together in a practical sense. That said, I went back and forth on it using `multi_lm` in `config_args` and `multi_sm` for `make_dataset_args`. It sounds like the preference is for either `multi_sensor` (or `multi_sm`, which I think is more consistent with `multi_lm` naming).

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
scottcanoe: So I think I'm going to remove this here. I tend to use this pattern when I want to have keyword arguments that must be given, like in a case where there are no sensible defaults. However, the defaults are fine, so no reason to keep it. On the other hand, it may be useful in `make_multi_lm_monty_config` since, as you pointed out, it'd be preferable to use a dictionary for those arguments, and I'm not sure defaults should be specified there.

src/tbp/monty/frameworks/config_utils/config_args.py
vkakerbeck: I understand why you have 
```
    if motor_system_args is None:
        motor_system_args = {}
```
my question was just whether `motor_system_args = copy.deepcopy(motor_system_args)` is necessary. I assume its so that if `motor_system_args` is use somewhere else the changes we make here won't affect it? I'm just wondering in what  context this would happen?

scottcanoe: Ah OK, sorry, I misunderstood. I thought it would probably be safer to copy any configs passed in so modifying `motor_system_args` after an experiment dict has been created wouldn't unexpectedly affect the experiment. For example, this could happen if, after you've defined one experiment config, you change something in a `motor_system_config`for use in the next experiment config (e.g., `NaiveScanSpiral`'s `step_size`), and it ends up affecting the previously defined experiment config. To me that would violate my implicit expectation that code lower in a file doesn't affect code higher up in the file. That said, a situation like that seems like not a great usage pattern, and I'd probably expect (hope?) someone would manually copy and configs that will be changed and used in next experiment config as in benchmark experiment.

I guess the other argument for copying would be to be consistent with the copying of sensor and learning module config args. It's not really a 1:1 analogy though since SM and LM configs actually have to be copied, whereas we only have one agent/motor system.

In general, I suppose I've been compartmentalizing individual experiment configs and protect them from being altered in ways I wouldn't expect, and I think that tendency just expressed itself here. That might just be a personal style thing though. Once I define an experiment config, I expect that if I want to change anything within it, I'll have to reach into the experiment config rather than changing an object that was used to initialize the experiment config.

I'm glad you brought this up though since it's revealing some unconscious patterns I've been using. Happy to go either way with this kind of stuff.

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -1153,6 +1154,189 @@ class FiveLMMontySOTAConfig(FiveLMMontyConfig):
...
+        sm_args_i["sensor_module_id"] = f"patch_{i}"
+        sensor_module_configs[f"sensor_module_{i}"] = {
+            "sensor_module_class": sensor_module_class,
+            "sensor_module_args": sm_args_i,
+        }
+    sensor_module_configs["view_finder"] = {
+        "sensor_module_class": DetailedLoggingSM,
+        "sensor_module_args": {
+            "sensor_module_id": "view_finder",
+            "save_raw_obs": True,
```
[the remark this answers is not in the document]
scottcanoe: OK, it's now using `"save_raw_obs": sensor_module_args.get("save_raw_obs", False)`.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
scottcanoe: Interesting, I hadn't considered multiple sensors connected to a single sensor module.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
scottcanoe: How about `make_multi_sensor_positions_on_grid`?

I think `on_grid` expresses that they cover a surface, though I'm not sure they do evenly cover it when the ordering method is `distance`. Those arrangements can be kind of jagged on the outermost sensors.

I'm inclined to think `on_grid` also implies that they exist in the same agent space, but let me know if not.

scottcanoe:
I'm currently scrubbing references to learning modules in my `make_dataset_configs.py` functions in favor of `multi_sensor...`, but it occurs to me that the existing code uses multi-LM language. For example, `MultiLMMountConfig` and `FiveLMMountConfig`. Maybe we want to settle on the language and change these names as well.

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -1153,6 +1154,189 @@ class FiveLMMontySOTAConfig(FiveLMMontyConfig):
...
+
+    return {
+        "sm_to_agent_dict": sm_to_agent_dict,
+        "sm_to_lm_matrix": sm_to_lm_matrix,
+        "lm_to_lm_matrix": lm_to_lm_matrix,
+        "lm_to_lm_vote_matrix": lm_to_lm_vote_matrix,
+    }
+
+
+def make_multi_lm_monty_config(
```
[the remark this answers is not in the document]
scottcanoe: I've added a parameter `connectivity_func` that is used to generate connectivity matrices. By default, it uses the function that makes flat, dense connectivity.

scottcanoe:
I'm ready for another round of review. Here's a brief summary of changes:
 - Some minor changes to variable names within functions to make them more descriptive.
 - I've replaced language related to learning modules and sensor modules with multi-sensor language in `make_dataset_configs.py`. This includes function names, parameters, and documentation. This is meant to prepare for future development where we may have more complex setups, but as I mentioned in a previous comment, this does clash somewhat with the multi-LM naming conventions used elsewhere in `make_dataset_configs.py`. 
 - I've removed the required keyword-only pattern in `make_multi_sensor_mount_config`'s signature.
 - In `config_args.py`, `make_multi_lm_connectivity` has been renamed to `make_multi_lm_flat_dense_connectivity`.
 - I've added a `connectivity_func` parameter in `make_lm_monty_config` to support the generation of non-flat, dense connectivity types.
 - I've also added the keyword-only pattern in `make_multi_lm_monty_config` to support function use with (unpacked) dictionaries instead of lists. However, I don't think many of the arguments have obvious default values, so I have made most of them **required** keyword arguments.

I haven't added a usage example to a docstring yet. I'm undecided as to where that should go (either `make_multi_sensor_habitat_dataset_args`, `make_multi_lm_monty_config`, or both?).

In general, I'm finding it a bit tricky to find the balance between a more generic approach that can support future use cases and functions that are useful now. For example, some functions (e.g., `make_sensor_positions_on_grid`) may be useful down the road in a multi-agent setting. However, pretty much every other function has certain (documented) constraints and assumptions baked in, such as 1:1 sensor -> sensor module relationships, 1:1 sensor module -> learning module relationships, a single agent, and the inclusion of a view finder.

I expect there will be some more back and forth about further changes.

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -1153,6 +1154,197 @@ class FiveLMMontySOTAConfig(FiveLMMontyConfig):
...
+    sensor_module_class: type,
+    sensor_module_args: Optional[Mapping],
+    motor_system_class: type,
+    motor_system_args: Optional[Mapping],
+    monty_args: Optional[Union[Mapping, MontyArgs]],
+    connectivity_func: Callable[[int], Mapping] = make_multi_lm_flat_dense_connectivity,
+) -> MontyConfig:
+    """Create a monty config for multi-LM experiments.
+
+    Creates a complete monty config for a multi-LM experiment. It is assumed that
```
nielsleadholm, line 1221: Is this still true now that you pass the `connectivity_func` as an argument?

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -1153,6 +1154,197 @@ class FiveLMMontySOTAConfig(FiveLMMontyConfig):
...
+        sensor_module_args = {}
+    sensor_module_configs = {}
+    for i in range(n_lms):
+        sm_args_i = copy.deepcopy(sensor_module_args)
+        sm_args_i["sensor_module_id"] = f"patch_{i}"
+        sensor_module_configs[f"sensor_module_{i}"] = {
+            "sensor_module_class": sensor_module_class,
+            "sensor_module_args": sm_args_i,
+        }
+    sensor_module_configs["view_finder"] = {
```
nielsleadholm, line 1302: Maybe it's worth being able to specify these, i.e. with a separate "view_finder" args? (which otherwise default to these?)

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -1153,6 +1154,197 @@ class FiveLMMontySOTAConfig(FiveLMMontyConfig):
...
+        sm_to_lm_matrix=connectivity["sm_to_lm_matrix"],
+        lm_to_lm_matrix=connectivity["lm_to_lm_matrix"],
+        lm_to_lm_vote_matrix=connectivity["lm_to_lm_vote_matrix"],
+        monty_args=monty_args,
+    )
+    return monty_config
+
+
+"""
+etc.
```
nielsleadholm, line 1344: What's this for?

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -1102,3 +1104,279 @@ class FiveLMMountConfig:
...
+             - "distance": sensors are ordered by their distance from the center.
+                This can result in a more jagged pattern along the edges but
+                results in sensors generally more packed towards the center.
+                Positions that are equidistant from the center are ordered
+                counterclockwise starting at 3 o'clock.
+        add_view_finder (bool, optional): Whether to include an extra position module
+            at the origin to serve as a view finder. Defaults to `True`.
+
+    Returns:
+        np.ndarray: A 2D array of sensor positions where each row has positions
```
nielsleadholm, line 1150: Maybe instead of "row has" say "row encodes"? The former to me makes it sound like there's lots of information in each row, i.e. the xyz coordinates as well as other information.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -1102,3 +1104,279 @@ class FiveLMMountConfig:
...
+
+    # Find smallest square grid size that can fit n_lms with odd-length sides.
+    grid_size = 1
+    while n_sensors > grid_size**2:
+        grid_size += 2
+
+    # Make coordinate grids, where the center is (0, 0).
+    points = np.arange(-grid_size // 2 + 1, grid_size // 2 + 1)
+    x, y = np.meshgrid(points, points)
+    y = np.flipud(y)
```
nielsleadholm, line 1169: What does this line do `y = np.flipud(y)`? Sorry just struggling to visualize it.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -1102,3 +1104,279 @@ class FiveLMMountConfig:
...
+    """Generate a multi-sensor mount configuration.
+
+    Creates a multi-sensor, single-agent mount config. Its primary use is in generating
+    a `MultiLMMountHabitatDatasetArgs` config. This function assumes that a view finder
+    will be included which entails that the number of sensor_ids, resolution, etc. will
+    be of length `n_sensors + 1`. Defaults are reasonable and reflect current common
+    practices.
+
+    Args:
+        n_sensors: Number of sensors. Count should not include an optional view finder.
```
nielsleadholm, line 1255: This suggests the viewfinder is option, but further below `add_view_finder=True` appears to be hard-coded, is it worth passing in a parameter for this? Similarly this should presumably influence the integer value of `arr_len`?

Alternatively maybe just don't state that the view-finder is optional.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -1102,3 +1104,279 @@ class FiveLMMountConfig:
...
+
+    # sensor resolutions
+    if resolutions is None:
+        resolutions = np.full([n_sensors + 1, 2], 64)
+    else:
+        resolutions = np.asarray(resolutions)
+    assert resolutions.shape == (
+        arr_len,
+        2,
+    ), "`resolutions` must have shape (n_sensors + 1, 2)"
```
nielsleadholm, line 1301: Depending on how you address the default view-finder thing, you can change the message in these asserts to use an "f-string" formatting that reflects the parameters.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -1102,3 +1104,279 @@ class FiveLMMountConfig:
...
+        sensor_ids: IDs of the sensor modules. Defaults to
+            `["patch_0", "patch_1", ... "patch_{n_sms - 1}", "view_finder"]`.
+        height: Height of the agent. Defaults to 0.0.
+        position: Position of the agent. Defaults to [0, 1.5, 0.2].
+        resolutions: Resolutions of the sensors. Defaults to (64, 64) for all
+            sensors.
+        positions: Positions of the sensors. If not provided, calls
+            `make_sensor_positions_on_grid` with its default arguments.
+        rotations: Rotations of the sensors. Defaults to [1, 0, 0, 0] for all sensors.
+        semantics: Defaults to `True` for all sensors.
```
nielsleadholm, line 1266: Given the other work you're doing, worth making sure we also set the default for this to False.

src/tbp/monty/frameworks/config_utils/config_args.py
vkakerbeck: That all makes sense, lets leave it how you have it

nielsleadholm (commented):
Really nice PR, thanks for the super clear documentation. Just added a few comments.

Re. the names of the functions, I agree it's a hard balance to strike - personally I'm ok with how they are now, i.e. avoiding making the names super specific, but then also very long-winded.

vkakerbeck (approved):
Thanks for all the updates! Looks very nice :) Happy to merge after you go through Niels comments.

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -1153,6 +1154,197 @@ class FiveLMMontySOTAConfig(FiveLMMontyConfig):
...
+"""
+Multi-LM Config Utils.
+"""
+
+
+def make_multi_lm_flat_dense_connectivity(n_lms: int) -> Dict:
+    """Create default connectivity matrices for a multi-LM experiment.
+
+    Helper function for creating monty configs for experiments having multiple sensor
+    and learning modules, and learning modules are not hierarchically connected.
```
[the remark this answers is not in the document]
scottcanoe: Good point, I updated the docstring to be more specific.

src/tbp/monty/frameworks/config_utils/config_args.py
scottcanoe: Ah good point. I'm deleting this bit.

src/tbp/monty/frameworks/config_utils/config_args.py
scottcanoe: Just a habit. I tend to break files into sections separated by double-commented section descriptions and this was not-very-helpful one. Banished!

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
scottcanoe: How about "each row is..."? "Encodes" sounds to me as if the (x, y, z) positions are encoded in an unexpected way (like a bunch of bits or something).

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
scottcanoe: `meshgrid` orders y-values from smallest (and negative) to largest when going top-to-bottom in the matrix. I flipped it so "up" is the positive y-direction to match habitat's coordinate system.

nielsleadholm: Ah nice thanks, maybe worth adding a short comment for that.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
scottcanoe: Got it.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
scottcanoe: Yeah, I toyed with making the view finder optional throughout but it got complicated. I think I'd like to keep the `add_view_finder` argument in `make_sensor_positions_on_grid` in case we have a use for it in the future and it's not hurting anything to leave with `True` as the default. I've updated the docstring though to clarify the relationship between `n_sensors` and the view finder.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
scottcanoe: Good point, will do.

src/tbp/monty/frameworks/config_utils/config_args.py
scottcanoe: Done

nielsleadholm (commented):
Thanks for making those changes, looks great.

And thanks for looking into that, sounds like a good plan. Let me know if you want to meet to discuss the semantic sensor in the view-finder - I would have thought that it can also use the estimate of semantic IDs (`get_semantic_from_depth` within `DepthTo3DLocations`), so I'm definitely curious to understand why that is failing.

PR #109, opened 2024-12-11
Fix division by zero when running the tutorial experiment

d-led:
## Problem

when running

```shell
% python benchmarks/run.py -e randrot_noise_10distinctobj_surf_agent
# or
% python benchmarks/run_parallel.py -e randrot_noise_10distinctobj_surf_agent
```

the following run-time errors can be observed:

```shell
.../tbp/monty/frameworks/models/motor_policies.py:1259: RuntimeWarning: divide by zero encountered in scalar divide
  return -np.degrees(np.arctan(x / z))
.../numpy/core/fromnumeric.py:3464: RuntimeWarning: Mean of empty slice.
  return _methods._mean(a, axis=axis, dtype=dtype,
.../numpy/core/_methods.py:192: RuntimeWarning: invalid value encountered in scalar divide
  ret = ret.dtype.type(ret / rcount)
```

as these might affect numeric results, an analysis is necessary. This has implications for published benchmarks that were affected.

[Initial discussion](https://thousandbrains.discourse.group/t/three-runtimewarning-s-when-running-the-tutorial-experiment/296?u=dled)

## Analysis

```python
>>> import numpy as np
>>> x=-1
>>> z=0
>>> -np.degrees(np.arctan(x / z))
  File "<stdin>", line 1, in <module>
ZeroDivisionError: division by zero
```

interestingly in the shell context this is a run-time error. Is this somehow (mis-)configured to be ignored?

## Fix

Fix `orienting_angle_from_normal` to not be affected by division by zero

## Result

```shell
> python benchmarks/run.py -e randrot_noise_10distinctobj_surf_agent
...
 overall/percent_correct 100
```

### DIFF
--- a/src/tbp/monty/frameworks/models/motor_policies.py
+++ b/src/tbp/monty/frameworks/models/motor_policies.py
@@ -1251,9 +1251,9 @@ class SurfacePolicy(InformedPolicy):
         x, y, z = rotated_point_normal
 
         if "horizontal" == orienting:
-            return -np.degrees(np.arctan(x / z))
+            return -np.degrees(np.arctan(x / z)) if z != 0 else -np.sign(x)*90.0
         if "vertical" == orienting:
-            return -np.degrees(np.arctan(y / z))
+            return -np.degrees(np.arctan(y / z)) if z != 0 else -np.sign(y)*90.0
 
 
 ###


tristanls:
FYI: @nielsleadholm, @vkakerbeck, @hlee9212, @scottcanoe

vkakerbeck:
Wow, thanks for taking a look at this and finding this fix! 
We will validate this against some of our longer benchmark experiments. 
- If it all looks good and it indeed improves the benchmark results we will open a corresponding PR to update the results table (since currently we don't have a way for external people to run on comparable infrastructure to have the timing column make sense). 
- If it doesn't affect performance it will be a nice fix anyways and we can just merge without updating the results table. 
- If it impacts performance negatively we'll let you know. 

(If you just want to get an impression yourself of how accuracy is affected on more challenging experiments you could also try running one of these benchmarks yourself and just ignore the runtimes when comparing https://thousandbrainsproject.readme.io/docs/benchmark-experiments#results-1)

d-led:
```
> python benchmarks/run.py -e base_77obj_dist_agent
...
overall/percent_correct 95.671
```

vs `93.51%` in the table. Eager to see complete results

scottcanoe:
Nice find @d-led! Can you post rotation error along with accuracy?

d-led:
> Nice find @d-led! Can you post rotation error along with accuracy?

|experiment|percent_correct|avg_rotation_error|
|--|--|--|
|base_77obj_surf_agent|98.7013|0.18305|
|base_77obj_dist_agent| 95.671|0.277|

I hope to see the numbers from the runs on your infrastructure 😄

scottcanoe:
@d-led So I think I know what's happening here: In `orienting_angle_from_normal`, `x`, `y`, and `z` are numpy scalars since they were obtained via numpy operators. Division by zero with numpy scalars results in a `RunTimeWarning` rather than the `ZeroDivisionError` you showed above when using built-in types. For example:
```python
>>> import numpy as np
>>> numerators = np.ones(3)
>>> denominators = np.zeros(3)
>>> numerators / denominators
<stdin>:1: RuntimeWarning: divide by zero encountered in divide
array([inf, inf, inf])
```
Or similarly
```python
>>> numerators.mean() / denominators.mean()
<stdin>:1: RuntimeWarning: divide by zero encountered in scalar divide
inf
```
It turns out that `np.arctan` handles infinities in a well-defined way, so when we end up with `np.inf` or `-np.inf` as the argument to `np.arctan`, we get a valid and expected result.
```python
>>> import numpy as np
>>> numerators = np.ones(3)
>>> denominators = np.zeros(3)
>>> np.arctan(numerators.mean() / denominators.mean())
<stdin>:1: RuntimeWarning: divide by zero encountered in scalar divide
1.5707963267948966
```
This makes sense because $\lim_{w \to \inf} \arctan(w) = \pi/2$.

So I'm pretty sure the current implementation is causing any problems numerically. On the other hand, it's straightforward to test. I'll be running benchmarks over the next few days, so I'll give your fix a try and report back. I wouldn't be surprised though if there are very small deviations in results due to rounding.

d-led:
@scottcanoe indeed, thinking of staying within numpy without conditionals, one could use `numpy.divide(x,z)` instead of `x/z`.

Shall we try this way? The branch is open for edits from maintainers

d-led:
```python
>>> np.degrees(np.arctan(np.divide(1,0)))
90.0
>>> np.degrees(np.arctan(np.divide(-1,0)))
-90.0
```

this probably makes more sense than the conditional

nielsleadholm:
Thanks @scottcanoe for elaborating those details, and @d-led for proposing a nice approach. We will still confirm with some internal tests, but it sounds like we won't expect a significant change in results, but it would still be nice to eliminate this runtime warning with a principled solution.

scottcanoe:
@d-led Do you get the same `RunTimeWarning` when doing `np.divide(1, 0)`? I get it, but only the first time it's used.
```python
>>> import numpy as np
>>> np.divide(3, 0)
<stdin>:1: RuntimeWarning: divide by zero encountered in divide
inf
>>> np.divide(3, 0)
inf
```
It also triggers the warning again when I do division by zero in another context.
```python
>>> def divide(a, b):
...     return np.divide(a, b)
...
>>> divide(1, 0)
<stdin>:2: RuntimeWarning: divide by zero encountered in divide
inf
>>> divide(1, 0)
inf
```
(Both of these blocks were run in the same interactive session.)

This is new to me. Apparently this behavior is enabled by Python's [warning filter system](https://docs.python.org/3/library/warnings.html#the-warnings-filter). Warnings are sent to a filter along with the the filename and line number where the warning was issued, and the filter decides whether to emit the warning or not. I hadn't come across this before.

In short, I believe we'll still get a zero-division warning with the `numpy.divide` change. (I actually get the warning >> 1 time when I run Monty experiments for some reason). However, I still think the `numpy.divide` switch is still worth doing since it might be possible that non-python scalars values for `x`, `y`, and `z` show up in the future due to an implementation change, and those may fail entirely with `x / z`.

Another option is the use a context manager to suppress the warnings like so:
```python
with np.errstate(divide='ignore'):
        if "horizontal" == orienting:
            return -np.degrees(np.arctan(np.divide(x, z)))
        if "vertical" == orienting:
            return -np.degrees(np.arctan(np.divide(y, z)))
```
I'm guessing `np.errstate` is pretty lightweight, so hopefully this wouldn't add much drag. Thoughts on this @tristanls?

Sidenote: It'd also be really nice to suppress `RuntimeWarning: Mean of empty slice` and a couple of others. The easiest way to do this would be to suppress all numpy warnings via [np.seterr](https://numpy.org/doc/2.1/reference/generated/numpy.seterr.html), but I'm not sure if that's a good idea. A clever way to deal with warnings that isn't too aggressive or perhaps can be switched on or off based on context (e.g. development vs regular usage) might be a good subject for an [RFC](https://thousandbrainsproject.readme.io/docs/request-for-comments-rfc).

I realize this is all getting away from the original intention behind your PR though, so no worries if you don't want to turn this into an all-out campaign against warnings :)

d-led:
> Do you get the same `RunTimeWarning` when doing `np.divide(1, 0)`? I get it, but only the first time it's used.

yes

Originally, indeed, I'd get rid of specific warnings. I'd not whole categories of warnings, as these could hide nasty bugs, as is the case with the division by zero, even though it happens rarely.

Perhaps, after the benchmark runs, this PR could be finalized - either with the branch version or the `numpy.divide` version.
Both result in the same improvements in accuracy (same results for non-random experiments)

nielsleadholm:
From my understanding, @d-led your original suggestion of a conditional (`if z=0`) would actually remove the warning, while providing a mathematically correct solution. That therefore seems like a good option if all of the `np.divide` approaches etc. are going to continue producing a warning.

d-led:
@nielsleadholm  exactly

d-led:
have the benchmarks run by any chance, @vkakerbeck ?

scottcanoe:
@d-led I'm in the process of updating benchmarks for another PR, so I've been a little delayed in checking this out. This is next on my list, so hopefully tomorrow I'll have an update.

d-led:
then, perhaps, before the benchmark run, I'll try out if inverting the conditional improves at least local performance. Also, probably, staying with the conditional that removes the warning is more sensible, and more explicit than the "mathematical" `np.divide` version

d-led:
Screenshot 2024-12-17 at 23 16 03

as inverting the conditional does improve the run-time a bit, I'd keep it (pushed). It also improves comprehending the change.

The only runtime warning left from within monty is:

```python
buffer.py:574: RuntimeWarning: invalid value encountered in multiply
  new_vals = np.empty((len(self) + 1, new_val_len)) * np.nan
```

according to a benchmark on [SO](https://stackoverflow.com/a/44921249/847349) `np.empty(n) * np.nan` can be replaced with `np.full(n, np.nan)`. Let's see if this brings an improvement as well.

nielsleadholm:
Thanks for checking that @d-led , as Scott said we've just had a couple other benchmarks we've needed to re-run, not to mention people off work due to the time of year, however we'll hopefully have an update on those soon.

Re. the `np.empty` change you made, how much of a runtime difference did you see on your end?

d-led:
> Re. the `np.empty` change you made, how much of a runtime difference did you see on your end?

Screenshot 2024-12-18 at 14 45 27

d-led:
I've rechecked while trying to keep the computer from sleeping. The total run-time time seems too volatile. So, I'd not count on my benchmarks on run-time at all. 

`overall/avg_episode_run_time` and `overall/run_time` decreased significantly in the latest attempt but `episode/run_time` stays rather stable. So, perhaps, one warning fix at a time, one could try `np.full` but in the next MR to measure it separately on your reference machines

nielsleadholm:
Great thanks for that data-point @d-led , we'll keep that in mind if we look at that warning in a future PR.

scottcanoe:
@d-led I've run a handful of benchmarks to make sure everything works as expected. They're almost identical to our [current benchmarks](https://thousandbrainsproject.readme.io/docs/benchmark-experiments), so we're good to go.

Some short benchmark experiments:
| Experiment                                  | % Correct | % Used MLH | Num Matching Steps | Rotation Error (radians) | Run Time | Episode Run Time (s) |
|---------------------------------------------|-----------|------------|--------------------|--------------------------|----------|----------------------|
| base_config_10distinctobj_dist_agent        | 98.57%    | 5.71%      | 36                 | 0.31                     | 12m      | 31s                  |
| base_config_10distinctobj_surf_agent        | 100.00%   | 0.00%      | 28                 | 0.21                     | 6m       | 28s                  |
| randrot_noise_10distinctobj_dist_agent      | 99.00%    | 7.00%      | 51                 | 0.50                     | 10m      | 57s                  |
| randrot_noise_10distinctobj_surf_agent      | 100.00%   | 0.00%      | 31                 | 0.34                     | 7m       | 53s                  | 


A longer benchmark experiment:
| Experiment                          | % Correct | % Used MLH | Num Matching Steps | Rotation Error (radians) | Run Time | Episode Run Time (s) |
|-------------------------------------|-----------|------------|--------------------|--------------------------|----------|----------------------|
| base_77obj_dist_agent               | 93.51%    | 14.29%     | 90                 | 0.31                     | 1h37m    | 297s                 | 

Thanks for looking into this, and welcome to the community! Tagging @nielsleadholm for final review.

scottcanoe (approved):

nielsleadholm (approved):
Looks great, thanks for checking those results @scottcanoe and thanks again @d-led for the nice fix! As Scott said, welcome to the community!

d-led:
@nielsleadholm @scottcanoe thank you for the warm welcome!

PR #110, opened 2024-12-12
Adding conceptual illustrations on vison of tbp page

dmsstern:
Added the illustrations previously discussed - The brain balloons illustration should switch to a night scene when user chooses dark mode (just a fun easter-egg and visual pun). Set the others as half banners because they were otherwise too large for the page.


codeallthethingz (changes requested):
Thank you for these amazing illustrations.  They're beautiful and I think add a lot to the page.


Some feedback (blocking): 

- HTML - We'd prefer that the images use the markdown format so that people don't have to also learn HTML when editing the docs.  
- Dark Mode: I think we should abandon dark mode for now as getting it to work with readme.com will be a time sink.  But we should keep the image for when we do figure it out. :) 

Other notes:
- Image Widths / Floating: There is a PR (https://github.com/thousandbrainsproject/tbp.monty/pull/64) to allow us to change the width of images and once it is merged we can change the image widths that way.  For now, having them full screen is probably fine.  As that PR is not merged, I may add in the float parameter too which is handy.

codeallthethingz:
Here is what this page looks like on the preview <3

image

codeallthethingz:
none blocking comment: The human-like-learning image could be more illustrative of that point.  Maybe a baby grasping something rather than the finger touching?

tristanls:
Hi @dmsstern, we updated our CLA signing process, hence the new comment.

codeallthethingz:
Moving this to a different PR - https://github.com/thousandbrainsproject/tbp.monty/pull/141

PR #114, opened 2024-12-14
WIP: Fix `patch_off_object`

scottcanoe:
In a small percentage of runs (<1%), Monty gets stuck bouncing between off-object observations early in the run. These experiments timeout after `max_monty_steps` with no matching steps performed. These timeout occur mostly, but not exclusively, in experiments with distant agents. So far I've focused on distant agents. Here's a quick overview of what I've observed and done.

In distant agents experiments, `get_good_view` is called at the beginning of an episode. Using the view finder, it moves the agent to a reasonable distance from the object and orients the agent so that the central pixel in the view finder is on-object. However, the view finder central area is not a perfect proxy for sensor module 0, and this method does not guarantee that sensor module 0 ("patch") is will be on-object even if the view finder is.

 1. Because of the differences in resolutions/zooms of the patch and view finder, the central view finder's pixel is covered by about 10 pixels in the patch's image. If the view finder is aimed directly on an object's edge, the patch's central pixel can end up slightly off-object. If the agent goes the wrong way to try and find the object and ends up even further off-object, then the system gets stuck in a loop, bouncing back and forth between these two off-object observations, repeatedly calling `fixme_undo_last_action`.
 2. In one case, there was a small hole in the object's semantic map. The view finder happened to oriented onto this hole which was off-object for the patch. It then jumped to an off-object location and got stuck in a `fixme_undo_last_action` loop. This is an exceedingly rare situation and is more related to a some quirk in the generation of the semantic map.

I've tried a few things, and none of have completely solved the problem. As of this writing, I've added one fix that first uses `get_good_view` with the view finder (as usual) and then performs a second pass using sensor module 0 to refine the agent's orientation. This seems to fix the issue caused by bullet point 1. 

I also experimented with adding some randomness to `fixme_undo_last_action` so that the previous action isn't undone exactly. That seemed to help in some cases, but I didn't test it widely as it doesn't guarantee we won't keep drifting off-object.

I haven't completely solved the problem yet, but including a second pass of `good_good_view` using sensor module 0 seems like a principled, albeit partial, solution. I've also added a fix that resets an LM's `current_mlh` between episodes which wasn't affecting the system's behavior, but it did cause incorrect logging of these `patch_off_object` episodes.


scottcanoe:
Tagging @hlee9212 for visibility.

scottcanoe:
Also noting here that I don't know what's causing the spurious semantic holes to sometimes show up (like in the "knife" example above), but I'll plan to take a closer look at that while working on the upcoming semantic sensor PR.

nielsleadholm (approved):
Really nice, thanks for the fix and the great documentation. Just added a couple clarifying comments/suggestions.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -715,6 +735,10 @@ def handle_successful_jump(self):
 
         else:
             self.get_good_view("view_finder")
```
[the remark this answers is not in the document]
scottcanoe: How about `get_good_view_with_patch_refinement`? It's a long name, but multi-step to me sounds like it could refer to multiple steps taken with the same sensor (as is partly done already with move forward).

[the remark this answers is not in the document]
nielsleadholm: Yeah good point, I like it!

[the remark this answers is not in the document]
scottcanoe: Updated.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -724,7 +725,7 @@ def find_location_to_look_at(self, sem3d_obs, image_shape, target_semantic_id):
         # as expected, which can otherwise break if e.g. on_object_image is passed
         # as an int or boolean rather than float
         smoothed_on_object_image = scipy.ndimage.gaussian_filter(
-            on_object_image, sem3d_obs.shape[0] / 10, mode="constant"
+            on_object_image, 2, mode="constant"
```
vkakerbeck, line 728: Why is this number hardcoded now?

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -366,6 +366,12 @@ def reset(self):
         self.symmetry_evidence = 0
         self.last_possible_hypotheses = None
 
+        self.current_mlh["graph_id"] = "no_observations_yet"
+        self.current_mlh["location"] = [0, 0, 0]
+        self.current_mlh["rotation"] = Rotation.from_euler("xyz", [0, 0, 0])
+        self.current_mlh["scale"] = 1
+        self.current_mlh["evidence"] = 0
```
vkakerbeck, line 373: Wha was the reason these needed to be added now?

docs/overview/benchmark-experiments.md
```diff
@@ -75,11 +75,11 @@ The following results are obtained from experiments on the entire YCB dataset (7
...
-| randrot_noise_77obj_dist_agent      | 87.01%    | 28.57%     | 155                | 0.64                     | 2h14m    | 479s                 |
-| randrot_noise_77obj_surf_agent      | 94.93%    | 19.48%     | 102                | 0.62                     | 1h23m    | 304s                 |
-| randrot_noise_77obj_5lms_dist_agent | 93.51%    | 3.90%      | 71                 | 0.92                     | 54m      | 1398s                |
+| base_77obj_dist_agent               | 92.21%    | 16.02%     | 88                 | 0.30                     | 1h38m    | 301s                 |
+| base_77obj_surf_agent               | 98.27%    | 4.33%      | 52                 | 0.18                     | 42m      | 123s                 |
+| randrot_noise_77obj_dist_agent      | 87.01%    | 29.00%     | 151                | 0.63                     | 2h10m    | 468s                 |
+| randrot_noise_77obj_surf_agent      | 94.37%    | 21.65%     | 113                | 0.61                     | 1h31m    | 339s                 |
+| randrot_noise_77obj_5lms_dist_agent | 90.91%    | 5.19%      | 70                 | 1.01                     | 1h7m     | 1439s                |
```
vkakerbeck, line 82: Do you have an idea why performance is worse now in terms of % correct and % used mlh?

vkakerbeck (approved):
LGTM, thanks for the fix! Just a few questions so I understand all the changes.

src/tbp/monty/frameworks/models/motor_policies.py
scottcanoe: I just landed on a kernel size of 2 because it made the smoothed images look "reasonably" smooth, and solved the issue with holes on the object. I can see why we would want to kernel size to scale with the resolution, but then I thought what we really want to scale it by is the physical size on the object between pixels since presumably holes in the object are usually pretty small (in centimeters). This is something I can come back to, but I just left it since it seemed to be giving a good result when I looked at all the images during debugging.

src/tbp/monty/frameworks/models/evidence_matching.py
scottcanoe: Since not resetting `current_mlh` was causing incorrect logging for `patch_off_object` episodes, I thought it made sense to include it here. More specifically, an episode could end up with `patch_off_object` but it would report that the most likely object was whatever the MLH was for the previous episode.

vkakerbeck: ah okay, good catch!

docs/overview/benchmark-experiments.md
scottcanoe: I was hoping to come up with a good answer for this, but ultimately I don't really know. I reran with a handful of different random seeds, sometimes recovering the roughly the same accuracy as before. Somehow the very slightly different starting conditions ended up having a noticeable change on accuracy, but I don't have much to say beyond that. It's something I'd like to investigate further in the near future. I'll be redoing all benchmarks for the semantic sensor PR, so maybe there's more I can learn then.

PR #131, opened 2025-01-12
Added Vale linter for spelling and APA title case to the docs

codeallthethingz:
This PR adds the vale documentation linter to the docs setup. 
It enforces spelling and the APA title case for headings.  It ignores python snake_case words.


tristanls:
**question**: How do I use this?

docs/future-work/project-roadmap.md
```diff
@@ -19,15 +19,15 @@ Tasks are categorized in two ways:
 
 Tasks that are done have a check mark next to them and are shaded in green. When a task gets checked off, it will add progress to the corresponding capabilities on the right.
 
-# What the TBP Team @Numenta is Working on
+# What the TBP Team is Working On
```
vkakerbeck, line 22: Wouldn't "on" qualify as a short preposition?

vkakerbeck (commented):
Lot's of good catches, I love it! I think this is the beginning of a beautiful friendship :)

Maybe we could add a sentence in the docs on editing the docs about how to use it. Also, should it be included in the conda env setup and as a github action on PRs?

codeallthethingz:
> Lot's of good catches, I love it! I think this is the beginning of a beautiful friendship :)
> 
> Maybe we could add a sentence in the docs on editing the docs about how to use it. Also, should it be included in the conda env setup and as a github action on PRs?

I added this: https://github.com/thousandbrainsproject/tbp.monty/pull/131/files#diff-3e5a50dac2843b05121cb8bc1803ec704405e0e38b006e1ef35a9819d37c583dL120-R120

Do you think we need more?

docs/future-work/project-roadmap.md
codeallthethingz: Not according to Vale.

image

codeallthethingz: I think this instance it's a linking verb, which should be capitalized.

codeallthethingz:
> Also, should it be included in the conda env setup and as a github action on PRs?

Yes, but that will be a separate PR.

docs/future-work/project-roadmap.md
vkakerbeck: Ah okay, good to know :)

vkakerbeck (approved):
Okay yeah doing that in a separate PR sounds good.

I regards to documentation of this I would make a separate paragraph for vale since it doesn't only check APA, it also checks spelling etc. and it kind of gets buried in the APA section. But we can also do this once the automation is added and this is actually being enforced. At that point I'd like to mention it in the same places in the docs where we also talk about running `ruff`.

PR #134, opened 2025-01-15
Allowing users to change the data dir

ramyamounir:
The getting started document shows how to [set the environment variables](https://thousandbrainsproject.readme.io/docs/getting-started#optional-set-environment-variables) for logs, models and wandb. I'm assuming that the goal is to allow users to move the whole tbp folder somewhere else (other than the standard `~/tbp`).

When I tried to move `~/tbp` to a different location, I realized that it's looking for the data directory in `~/tbp/data`, this path is hard coded in a couple of files. I propose this quick fix:

- Add environment variable parsing for `MONTY_DATA` in the [`run_env`](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/run_env.py) file.
- Fix all instances of hard-coded `~/tbp/data`, such as [here](https://github.com/thousandbrainsproject/tbp.monty/blob/14a423558970b1ca79a0c20f5e4255e08a8a9ddb/src/tbp/monty/frameworks/config_utils/make_dataset_configs.py#L90).
- update the getting started documentation.

I have tested running the experiment `randrot_noise_10distinctobj_surf_agent` without issues.

@tristanls Moving the tbp folder would probably break tbp.cli rsync.


codeallthethingz:
This looks great, but I'm going to remove myself from this review process as I'm not experienced in this area.

tristanls:
I assigned myself, @vkakerbeck, and @nielsleadholm as reviewers to have more sets of eyes on this possibly breaking change.

docs/how-to-use-monty/getting-started.md
```diff
@@ -151,6 +151,17 @@ export MONTY_MODELS=/path/to/your/pretrained/models/dir
...
+
+### MONTY_DATA
+
+If you did not save the data (e.g., YCB objects) in the `~/tbp/data` folder, you will need to set the **MONTY_DATA** environment variable.
+
+```shell
+export MONTY_DATA=/path/to/your/data/dir
+```
+
+This path should point to the `data` folder that contains the `habitat` folders in the case of YCB objects.
```
nielsleadholm, line 163: This is nice documentation, might be worth changing the example above to

```
export MONTY_DATA=/path/to/your/data
```

to more closely reflect the fact that the directory should end with `/data`

vkakerbeck, line 163:
```suggestion
This path should point to the `data` folder, which contains data used for your experiments. Examples of data stored in this folder are the `habitat` folder containing YCB objects, the `worldimages` folder containing camera images for the 'Monty Meets Worlds' experiments, and the `omniglot' folder containing the Omniglot dataset.
```

vkakerbeck (approved):
Looks good! Thanks for the change :) I would still suggest that everyone internal in our team uses the `~/tbp/` directory so we don't get unnecessary complications with our cli scripts.

docs/how-to-use-monty/getting-started.md
ramyamounir: Thanks Niels, I've updated the PR.

ramyamounir: Great, thanks! I agree about keeping `~/tbp` internally. Documentation fixed.

tristanls (approved):
Thank you. Looks good.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -31,6 +31,13 @@ def setup_env(monty_logs_dir_default: str = "~/tbp/results/monty/"):
         os.environ["MONTY_MODELS"] = monty_models_dir
         print(f"MONTY_MODELS not set. Using default directory: {monty_models_dir}")
 
+    monty_data_dir = os.getenv("MONTY_DATA")
+
+    if monty_data_dir is None:
+        monty_data_dir = os.path.expanduser("~/tbp/data/")
```
nielsleadholm, line 37: I noticed that in this instance you call `os.path.expanduser` once, whereas for the other directories, we don't, requiring us to call it every time e.g. `monty_logs_dir` is used. 

I think the way you've done it here looks a bit cleaner, something we can think about.

nielsleadholm (approved):
Looks great! Just added a (positive) comment about a motif you used to consider in future work but will merge.

PR #139, opened 2025-01-16
Docs: How to use Vale

codeallthethingz:
Update the docs' documentation page about how to install and use Vale.


vkakerbeck (approved):
Nice, thanks for adding this!

PR #145, opened 2025-01-17
Make Vale Dictionary Case-Sensitive

codeallthethingz:
This PR makes the vale custom dictionary case sensitive.  The vale documentation says this file is case sensitive, but that is only the case if you're including the Vale Style in the `.vale.ini` config file.


vkakerbeck (approved):
Nice, that's even better!

PR #146, opened 2025-01-20
Patch off object (again)

scottcanoe:
### Problem Statement

Note: This PR has been tested against the long benchmark experiments, but I haven't run the whole suite. It has also been tested against with the experiments that revealed the problem. I've left it as a draft since changes like these usually require a complete re-benchmarking.

My previous attempt at fixing `patch_off_object` episodes was flawed. Those modifications were sufficient to prevent `patch_off_object` episodes for all of the benchmark experiments, but I started (rarely) getting `patch_off_object` when running longer experiments with more random rotations.

The primary change to Monty in the last `patch_off_object` PR was to add `get_good_view_with_patch_refinement`. The idea is the take a first pass at finding a good view using the view finder and then, if the central patch wasn't centered on the target object, do another pass of `get_good_view` using the central patch. The idea was to make finer adjustments that would make sure the central patch was on-object at the start of the episode. Here's where it failed:

- The function that computes how much the agent should in order to have the sensor land on the targeted location did not account for the current agent/sensor rotation. It assumed the agent and sensor have not yet been rotated, and so it always works fine for the first call to `orient_to_object`. The fact that the last PR still "worked" on benchmark experiments without having fixed this is pretty surprising, but the incorrect behavior is clear once you collect all the pre-episode observations and actions visualize them (which is a little tricky since Monty's built-in loggers aren't set up for this).

- Looking up/down *and then* left/right is not the same as pitching and yawing to a target simultaneously. This is partly why attempting to orient through a multi-action procedure (down-then-left) didn't always land the sensor on the object the first time.

### Solutions

- Fix `InformedPolicy.compute_look_amounts` so that it returns movements relative to the current agent/sensor rotation. 
- Modify `InformedEnvironmentDataLoader.get_good_view` to allow multiple orienting attempts, similar to how it can perform multiple move-forward actions.
- Some functions, especially in `InformedPolicy`, needed some refactoring to make the multiple-reorienting steps doable in a relatively clean way.

### Smaller Changes
- I changed the name of parameter `view_sensor_id` to `sensor_id` in several `InformedPolicy` functions. The naming of `view_sensor_id` sounds like a holdover from when the view finder was the only sensor to use these functions. 
- The view finder was also hard-coded into `find_location_to_look_at`, and that's been changed as well.


src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -580,40 +586,36 @@ def get_good_view(
...
-        actions, on_target_object = self.motor_system.orient_to_object(
-            self._observation,
-            view_sensor_id,
-            target_semantic_id=self.primary_target["semantic_id"],
-            multiple_objects_present=multiple_objects_present,
-        )
-        if not on_target_object:
+        on_target_object = self.is_on_target_object(sensor_id)
+        num_attempts = 0
+        while not on_target_object and num_attempts < max_orientation_attempts:
```
vkakerbeck, line 596: Should we raise and error or warning if max_orientation_attempts is reached?

vkakerbeck (approved):
Wow very nice! You really dug deep there. This solution is nice and hopefully it all works now. Also nice additional refactors and renaming :)

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -540,13 +543,16 @@ def get_good_view(
         performed prior to the translational movement step.
 
         Args:
-            view_sensor_id: The name of the sensor used to inform movements.
+            sensor_id: The name of the sensor used to inform movements.
             allow_translation: Whether to allow movement toward the object via
                 the motor systems's `move_close_enough` method. If `False`, only
                 orientienting movements are performed. Default is `True`.
+            max_orientation_attempts: The maximum number of orientation attempts
+                allowed before giving up and return `False` indicating that the
```
nielsleadholm, line 551: *return --> returning

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -663,58 +663,88 @@ def orient_to_object(
             us onto the target object.
         """
         # Reconstruct 2D semantic map.
```
nielsleadholm, line 665: Could the below code call `is_on_target_object`?

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -625,10 +627,36 @@ def get_good_view_with_patch_refinement(self) -> bool:
...
-                on_target_object = self.get_good_view(patch_id, allow_translation=False)
+                on_target_object = self.get_good_view(
+                    patch_id,
+                    allow_translation=False,  # only orientation movements
+                    max_orientation_attempts=10,  # allow 10 reorientation attempts
+                )
                 break
         return on_target_object
 
+    def is_on_target_object(self, sensor_id: str) -> bool:
```
nielsleadholm, line 638: I think it might be better if this is in the motor system, rather than embodied_data, given the effort to pull as much policy (and sensory processing) out of embodied data as possible.

nielsleadholm (commented):
Wow, super thorough and nicely documented, thanks for digging into this!

Just some small suggestions on the refactor.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -663,58 +663,88 @@ def orient_to_object(
...
+                to the new target.
+            sensor_id: the ID of the sensor used to produce the relative location
 
         Returns:
-            down_amount: Amount to look down.
-            left_amount: Amount to look left.
+            rel_down_amount: Amount to look down.
+            rel_left_amount: Amount to look left.
         """
+        # The are the amounts we'd look down and left if the agnet and sensors
```
tristanls, line 718: **nit**: `# These are`

tristanls:
Hi @scottcanoe. I removed the `bug` label during triage. I've been putting a `bug` label on Issues to identify potential bugs. However, I feel that this being a PR, it is a fix or solution to a bug, not a bug itself. If you'd rather keep the label on PRs marking them as a bug fix, please re-add it, and I'll leave it alone.

src/tbp/monty/frameworks/environments/embodied_data.py
scottcanoe: The calling function (DataLoader.pre_episode) will raise the error when `on_target_object` is returned as `False`.

src/tbp/monty/frameworks/environments/embodied_data.py
scottcanoe: Done.

src/tbp/monty/frameworks/models/motor_policies.py
scottcanoe: I've actually removed the check in this function. I'm certain it isn't needed since the calling function does the on-object checking prior to calling this function.

scottcanoe:
Thanks @vkakerbeck @nielsleadholm and @tristanls for the review, great comments and suggestions as always.

Since you last saw it, I reworked `compute_look_amounts`, and it is now completely dead-on. Between the benchmarks and DMC experiments, I ran close to 10k episodes, and not a single one actually required multiple orientation steps with the patch. So I guess the question is, should we leave in the new multiple attempts machinery? I'm fine leaving it in just in case some super rare edge case comes up where its necessary, but other than that, I'm good to merge.

At some point in the future, I plan to contribute some documentation about the coordinate system(s) in use. I learned a lot working on this.

docs/overview/benchmark-experiments.md
```diff
@@ -112,8 +112,8 @@ An object is classified as detected correctly if the detected object ID is in th
...
 |---------------------------------------------|----------------------|------------------------|------------------------|------------------------|----------|----------------------|
 | surf_agent_unsupervised_10distinctobj       | 70.00%               | 83.33%                 | 1.43                   | 1.11                   | 20m      | 12s                  |
-| surf_agent_unsupervised_10distinctobj_noise | 70.00%               | 67.78%                 | 1.19                   | 2.11                   | 24m      | 15s                  |
-| surf_agent_unsupervised_10simobj            | 40.00%               | 86.67%                 | 2.60                   | 1.30                   | 28m      | 17s                  |
+| surf_agent_unsupervised_10distinctobj_noise | 70.00%               | 67.78%                 | 1.19                   | 2.11                   | 25m      | 15s                  |
+| surf_agent_unsupervised_10simobj            | 40.00%               | 86.67%                 | 2.60                   | 1.30                   | 28m      | 17s                  |16s                  |
```
vkakerbeck, line 116: Is this a formatting artifact? (should 17s be replaced by 16s instead of appending another column?)

vkakerbeck (commented):
Wow, this is awesome! I like how you rewrote the `compute_look_amounts` function. Looks like you really got a deep understanding of all the spatial transformations and different reference frames (which I've been banging my head against numerous times so I know its not an easy feat!) A writeup (in another PR) of some of the insights you had while doing that would be great! We started writing up a couple of conventions that helped us debug before here: https://thousandbrainsproject.readme.io/docs/common-issues-and-how-to-fix-them but feel free to start a new page just focused on this.

Just added one note on the benchmark table where it looks like the format may be a bit off but otherwise I'm happy with merging this.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -703,40 +695,42 @@ def compute_look_amounts(
         """Compute the amount to look down and left given a relative location.
 
         This function computes the amount needed to look down and left in order
-        for the sensor to be aimed at the object. The returned amounts are relative
+        for the sensor to be aimed at the target. The returned amounts are relative
         to the agent's current position and rotation.
 
+        TODO: Test whether this function works when the agent is facing in the
```
nielsleadholm, line 701: Is it possible to give a bit of context for this TODO, e.g. why/when it is expected things might fail?

nielsleadholm (approved):
Yeah awesome work, thanks for going even deeper on this problem. Just a clarifying comment.

docs/overview/benchmark-experiments.md
scottcanoe: Copy/paste error, fixed.

src/tbp/monty/frameworks/models/motor_policies.py
scottcanoe: Sure, the thing I have in mind is in regards to arctan2 and the discontinuity at 180 degrees. It may work totally fine when looking down positive z, but since it took some trial and error to work out the conversion to angles in body coordinates with the positive z-axis pointing opposite from the body's rotation, I wanted to leave a note in case there's some debugging in the future.

I'll make a plan to add some unit tests for this function. I'll have to work out how to initialize the agent and environment to test different scenarios, but it's pretty doable.

scottcanoe: Also adding what I just said to the docstring.

PR #147, opened 2025-01-21
refactor!: rename num_cpus to num_parallel for run_parallel

tristanls:
I noticed that when we run an experiment with `run_parallel` it outputs:
```
-------- Running evaluation experiment randrot_10distinctobj_surf_agent with 16 cpus --------
```
16 is the number of episodes that will run in parallel. However, this does not guarantee 1 CPU per episode as an episode may launch multi-CPU computations.

This pull request updates the output to read `with 16 episodes in parallel` and renames the `--num_cpus` option and `num_cpus` parameter to `--num_parallel` and `num_parallel` respectively.

As this is a breaking change, I request a broader review from recent experiment runners.

@hlee9212, this impacts floppy copies of run_parallel.

### DIFF
--- a/src/tbp/monty/frameworks/config_utils/cmd_parser.py
+++ b/src/tbp/monty/frameworks/config_utils/cmd_parser.py
@@ -121,7 +121,11 @@ def create_cmd_parser_parallel(all_configs):
         choices=list(all_configs.keys()),
     )
     parser.add_argument(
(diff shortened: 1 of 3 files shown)


nielsleadholm (approved):
Looks great, thanks Tristan.

scottcanoe (approved):
Makes sense, LGTM.

PR #149, opened 2025-01-23
refactor: compartmentalize habitat simulator dependent code

tristanls:
This pull request contributes to [RFC 6 Compartmentalize Simulator Code Implementation](https://github.com/thousandbrainsproject/tbp.monty/issues/154) and advances the code toward the goal of decoupling the Monty runtime from HabitatSim.

Screenshot 2025-01-24 at 18 34 34

## Summary of changes

Habitat-specific dependencies are removed from `frameworks/config_utils/make_dataset_configs.py` and put in the newly created `simulators/habitat/configs.py`. Habitat-specific environment and environment utils files are likewise moved to the `simulators/habitat` folder.

@ramyamounir, I attempted to incorporate [your advice of moving `AgentConfig` into the newly created `habitat/configs.py`](https://github.com/thousandbrainsproject/tbp.monty/pull/150#pullrequestreview-2570803703). However, I ended up with a circular import when `habitat/environment.py` imports `habitat/configs.py`. This is because `habitat/configs.py` imports `frameworks/config_utils/make_dataset_configs.py`, which imports `habitat/environment.py`, which imports `habitat/configs.py`. So, I left it off for now. I think the word "configs" is overloaded, which makes it seem like `AgentConfig` should be in `configs.py`. However, I believe `configs.py` is a configuration for arranging environments, whereas `AgentConfig` is a configuration for running stuff inside the environment. I think the two are distinct, as suggested by the circular import error when attempting to make them the same.

FYI: @hlee9212, @nielsleadholm, @scottcanoe


ramyamounir:
I see, nevermind then. I thought it would make things cleaner to keep all the dataclasses in the configs file, similar to how `make_dataset_configs` is now. But looks like there could be more to it. Thanks for looking into it anyway :+1:

vkakerbeck (approved):
Very nice, I like the new structure.

I assume you tested that the habitat and other benchmarks still run fine?

@scottcanoe and @hlee9212 once this PR is merged you will need to update imports in the dmc configs correspondingly https://github.com/thousandbrainsproject/monty_lab/tree/main/dmc/configs

@tristanls if it is not too much extra trouble, could you also update the imports in these monty_lab configs? https://github.com/thousandbrainsproject/monty_lab/tree/main/experiments/configs

tristanls:
@vkakerbeck, I'm not seeing any errors when running a random sampling of benchmarks.

tristanls:
> @tristanls if it is not too much extra trouble, could you also update the imports in these monty_lab configs? https://github.com/thousandbrainsproject/monty_lab/tree/main/experiments/configs

@vkakerbeck , https://github.com/thousandbrainsproject/monty_lab/pull/24

PR #151, opened 2025-01-23
refactor: remove habitat dependency from embodied_data

tristanls:
This pull request removes Habitat dependency from `frameworks/environments/embodied_data.py` as part of the work to decouple Monty runtime from HabitatSim.

Screenshot 2025-01-23 at 17 12 24

Previous planning called for writing a replacement `get_bounding_corners` function. However, after some investigation, it became clear that the use of `get_bounding_corners` within `embodied_data.py` is done only with the expectation that the Habitat environment is being used. As such, removing the dependency directly is a better refactor.

This refactoring moves the use of Habitat-specific `get_bounding_corners` into Habitat-specific `frameworks/environment/habitat.py`. To accomplish this, the `EmbodiedEnvironment` interface is expanded with `add_object` and `remove_all_objects` methods promoted directly from the `HabitatSim` implementation. Essentially, `embodied_data.py` can now call `EmbodiedEnvironment.add_object(...)`, which calls `HabitatEnvironment.add_object(...)`, which calls `HabitatSim.add_object(...)`. The call to `get_bounding_corners` is now inside `HabitatEnvironment.add_object(...)`. Previously, `embodied_data.py` called `HabitatSim.add_object(...)` directly via `dataset.env._env.add_object(...)`. This achieves our goal: if the runtime is not configured with a `HabitatEnvironment`, then `embodied_data.py` can still execute without importing a Habitat-specific dependency.

Due to the added interface in `EmbodiedEnvironment`, other environments now must implement the `add_object(...)` and `remove_all_objects` methods. This pull request adds these implementations by raising `NotImplementedError`. TODOs are added to highlight that extending existing environments with unimplemented methods highlights a mistake in how our environment hierarchy is defined and used. The work to evaluate and refactor the environment hierarchy is deferred. Environments will be revisited in the future in the scope of other planned work.

### DIFF
--- a/src/tbp/monty/frameworks/environments/embodied_environment.py
+++ b/src/tbp/monty/frameworks/environments/embodied_environment.py
@@ -10,12 +10,14 @@
 
 import abc
 import collections.abc
-from typing import Any, Dict
+from typing import Any, Dict, Optional, Tuple
 
 from tbp.monty.frameworks.actions.actions import Action
 
 __all__ = ["EmbodiedEnvironment", "ActionSpace"]
 
+VectorXYZ = Tuple[float, float, float]
+QuaternionWXYZ = Tuple[float, float, float, float]
 
 class ActionSpace(collections.abc.Container):
     """Represents the environment action space."""
@@ -34,6 +36,44 @@ class EmbodiedEnvironment(abc.ABC):
         pass
 
     @abc.abstractmethod
+    def add_object(
+        self,
+        name: str,
+        position: VectorXYZ = (0.0, 0.0, 0.0),
+        rotation: QuaternionWXYZ = (1.0, 0.0, 0.0, 0.0),
+        scale: VectorXYZ = (1.0, 1.0, 1.0),
+        semantic_id: Optional[str] = None,
+        enable_physics: Optional[bool] = False,
+        object_to_avoid=False,
+        primary_target_object=None,
+    ):
+        """Add an object to the environment.
+
+        Args:
+            name: The name of the object to add.
+            position: The initial absolute position of the object.
+            rotation: The initial rotation WXYZ quaternion of the object. Defaults to
+                (1,0,0,0).
+            scale: The scale of the object to add. Defaults to (1,1,1).
+            semantic_id: Optional override for the object semantic ID.
+            enable_physics: Whether to enable physics on the object. Defaults to False.
+            object_to_avoid: If True, run collision checks to ensure the object will not
+                collide with any other objects in the scene. If collision is detected,
+                the object will be moved. Defaults to False.
+            primary_target_object: If not None, the added object will be positioned to
+                obscure the initial view of the primary target object. Used when adding
+                multiple objects. Defaults to None.
+
+        Returns:
+            The newly added object.
+
+        TODO: This add_object interface is elevated from HabitatSim.add_object and is
+              quite specific to HabitatSim implementation. We should consider
+              refactoring this to be more generic.
+        """
+        pass
+
+    @abc.abstractmethod
     def step(self, action: Action) -> Dict[Any, Dict]:
         """Apply the given action to the environment.
 
@@ -45,6 +85,16 @@ class EmbodiedEnvironment(abc.ABC):
     @abc.abstractmethod
     def get_state(self):
         """Return the state of the environment (and agent)."""
+        pass
+
+    @abc.abstractmethod
+    def remove_all_objects(self):
+        """Remove all objects from the environment.
+
+        TODO: This remove_all_objects interface is elevated from
+              HabitatSim.remove_all_objects and is quite specific to HabitatSim
+              implementation. We should consider refactoring this to be more generic.
+        """
         pass
 
     @abc.abstractmethod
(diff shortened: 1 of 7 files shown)


tristanls:
I'm requesting a review from @vkakerbeck and @nielsleadholm due to your historical context with the code and Habitat. I also request @scottcanoe's review to highlight this change and ensure it does not interfere with your work. FYI, @hlee9212, I don't think this currently overlaps with anything you're doing.

src/tbp/monty/frameworks/environments/embodied_environment.py
```diff
@@ -33,6 +35,44 @@ def action_space(self):
...
+            position: The initial absolute position of the object.
+            rotation: The initial rotation WXYZ quaternion of the object. Defaults to
+                (1,0,0,0).
+            scale: The scale of the object to add. Defaults to (1,1,1).
+            semantic_id: Optional override for the object semantic ID.
+            enable_physics: Whether to enable physics on the object. Defaults to False.
+            object_to_avoid: If True, run collision checks to ensure the object will not
+                collide with any other objects in the scene. If collision is detected,
+                the object will be moved. Defaults to False.
+            primary_target_object: If not None, the added object will be positioned to
```
nielsleadholm, line 63: When this is passed, then the added object should *not* obscure the view of the primary target object. The understandable confusion is probably a result of the typo that exists in the following, and which would be worth also fixing:

```python
            primary_target_bb: If not None, the bounding box of the primary target
                object; passed when we're adding multiple objects, such that we ensure
                that the added object obscures the initial view of the primary target
                object (which avoiding collision alone cannot guarantee); defined by a
                list of the min and max corners
```

It should say, "that the added object does *not* obscure"

More description is provided under `find_non_colliding_positions`, bullet (ii).

nielsleadholm (commented):
Looks great, just added a clarifying comment about the use of bounding boxes.

vkakerbeck (commented):
Thanks for looking into this!

I may be missing something but could we simplify this so we don't need to add the abstract functions? As far as I understand:
- the abstract functions are added because these functions are called in the `change_object_by_idx` function in the `EnvironmentDataLoaderPerObject` class.
- The other experiments (omniglot, monty meets world,...) still continue to work (don't raise NotImplemented) because each of their Dataloaders overwrites the `change_object_by_idx` function

So what if we changed the class hierarchy is shown in the screenshot, making it more specific which dataloaders are meant to be used with Habitat? Then the `EnvironmentDataLoaderPerObject` class would have an abstract `change_object_by_idx` function which each of its subclasses need to implement (specific to whatever `EmbodiedEnvironment` it interacts with).

Screenshot 2025-01-24 at 3 19 01 PM

What do you think? Am I missing something? To me this would make things more clear in regards to which classes are expected to interact with Habitat and avoid adding `add_object` and `remove_all_objects` placeholders in the other Environment classes.

src/tbp/monty/frameworks/environments/embodied_environment.py
tristanls: Thank you for catching this. I updated both argument descriptions.

tristanls:
@vkakerbeck

Here's how I've been thinking about it and why I don't think adding a level to the class inheritance hierarchy will improve things.

The problem in the code is that the data loader reaches inside the dataset to access the environment and then calls a method on it.
It used to be:
```python
self.dataset.env._env.remove_all_objects()
```
This PR changes it to:
```python
self.dataset.env.remove_all_objects()
```
We still have the problem that the data loader reaches inside the dataset and that the data loader knows what kind of environment is contained in the dataset's `env` property.

Making the data loader more specific does not remove the problem that it should not need to know the kind of `dataset.env` in order to know what methods to call on it.

This problem can be surfaced in two ways. I chose to surface it by adding an abstract method that raises `NotImplementedError`. An alternate way of surfacing this problem is to do something like:
```python
if isinstance(self.datasets.env, HabitatEnvironment):
    self.datasets.env.remove_all_objects()
```

By adding a `HabitatDataLoader` we still have the problem that `HabitatDataLoader` reaches into `dataset.env` and knows what methods to call on it. So even with `HabitatDataLoader`, we'd still need either a `NotImplementedError` or an `isinstance` check to point out that we shouldn't be calling methods that might not exist on `dataset.env`.

If we were to update an inheritance hierarchy to get rid of reaching into `dataset.env`, we would probably need to specialize the `EnvironmentDataset` into `HabitatDataset` instead. This would then go with `HabitatDataLoader`. At least, I think that's the case at first glance. I aimed to highlight the problem but avoided solving it in this pull request. I do think that we'll need to make changes, but I wanted to defer that when we work on refactoring DataLoader and Dataset into EnvironmentInterface later on.
Screenshot 2025-01-24 at 10 11 24

vkakerbeck (approved):
Good points. Thanks for the detailed explanation. Given that we are already planning a big refactor of DataLoader and Dataset anyways I'm happy with just highlighting the issue here like you do and fixing it then.

scottcanoe (approved):
Looks good to me. No conflicts with what I've been doing.

nielsleadholm (approved):
Great stuff, thanks for the doc-string updates.

PR #153, opened 2025-01-24
refactor: remove habitat coupling from monty world benchmarks

tristanls:
This pull request advances toward the goal of decoupling the Monty runtime from HabitatSim. When this pull request is merged, people should be able to run non-HabitatSim benchmarks/experiments (not tests) without having HabitatSim installed.
Screenshot 2025-01-21 at 09 40 48

## Summary of changes

The main problem being solved is that we can no longer import all of the configurations when running an experiment. Instead, we want to import only the configurations specific to the experiment being run.

The framework `run` and `run_parallel` functions are updated to use command parsers that no longer expect fully loaded configurations but instead accept a list of strings enumerating the allowed experiment names to run.

The benchmark `run` and `run_parallel` functions have been updated not to import all configurations. Instead, they now only import configuration groups for the experiments to run.

Another change concerns the management, coordination, and synchronization of experiment names. We need access to experiment names without importing any configurations. So, `benchmarks/configs/names.py` gives us that access.

One consideration is that if the experiment names are defined in two different places, they will drift apart, and the code may fail in non-obvious ways. So, the goal was to write code that would immediately inform us of drift and force a correction. This is the idea behind using dataclasses in `benchmarks/configs/names.py` for experiment names. If we forget to declare a name in `benchmarks/configs/names.py` and we attempt to run an experiment from that configuration group, we will see:
```
(tbp.monty) tslominski@Mac tbp.monty % python benchmarks/run.py -e base_77obj_dist_agent
...
  File "benchmarks/run.py", line 67, in <module>
    from configs.ycb_experiments import CONFIGS as YCB
  File "/Users/tslominski/tbp/tbp.monty/benchmarks/configs/ycb_experiments.py", line 630, in <module>
    experiments = YcbExperiments(
TypeError: __init__() got an unexpected keyword argument 'base_77obj_surf_agent'
```
On the other hand, if we declare the name but forget to provide a configuration for it, we will see:
```
(tbp.monty) tslominski@Mac tbp.monty % python benchmarks/run.py -e base_77obj_dist_agent
...
  File "benchmarks/run.py", line 67, in <module>
    from configs.ycb_experiments import CONFIGS as YCB
  File "/Users/tslominski/tbp/tbp.monty/benchmarks/configs/ycb_experiments.py", line 630, in <module>
    experiments = YcbExperiments(
TypeError: __init__() missing 1 required positional argument: 'base_77obj_surf_agent'
```
This way, the code ensures that we don't have unconfigured experiment names hanging around.

The last bit of change is the need to separate the one Monty World Habitat experiment from the other Monty World experiments and not have Monty World experiments import HabitatSim-dependent YCB configs for some defaults. While this is the most straightforward change in this pull request, it looks the most involved as some common defaults are extracted into the `benchmarks/configs/defaults.py` file, and some inheritance is removed since the config groups are separate.

Documentation is updated regarding the Monty World Habitat experiment and links to hosted required data is provided as well. (Maintainer) preview here: https://thousandbrainsproject.readme.io/v0.0-tristanls-remove-benchmark-habitat-coupling/docs/benchmark-experiments#monty-meets-world

(Jan 29, 2025 Update): To import configurations from the `benchmarks` folder outside of Monty (e.g., monty_lab), I needed to turn it into a module structure and use absolute imports like `benchmarks.configs.defaults` instead of relative imports like `.defaults`. However, to make the absolute imports work, I needed to add the root directory to the system path, which explains `sys.path.insert(...)` in `run.py` and `run_parallel.py`. See this commit https://github.com/thousandbrainsproject/tbp.monty/commit/c06790d269c9c015fb3ea4d78e720de3b67070d5

(Jan 29, 2025 Update 2): While considering tutorials, it seemed beneficial to add `MyExperiments` boilerplate code in `benchmark/configs` so that users don't have to figure it out in order to add their own experiments. This addition is covered by [this commit changset](https://github.com/thousandbrainsproject/tbp.monty/compare/70f7aa0b6fd7c83e998b7809740a557d2d06fd6b..4aa406744cb8f06c3e841c4ebd40c459c6ce40df).


tristanls:
FYI: @hlee9212, @scottcanoe, @ramyamounir

tristanls:
@vkakerbeck 

I ran the two benchmarks that were moved around and separated:
| Experiment | % Correct | % Used MLH | Num Matching Steps | Run Time | Episode Run Time (s)
|---|---|---|---|---|---
| randrot_noise_sim_on_scan_monty_world | 80.00% | 80.83% | 412 | 38m36s | 19s
| world_image_on_scanned_model | 72.92% | 83.33% | 442 | 8m11s | 10.2s

These results are exactly the same as [the ones currently on file](https://thousandbrainsproject.readme.io/docs/benchmark-experiments#monty-meets-world) and within the expected variation in the runtime.

Regarding the monty_lab function in https://github.com/thousandbrainsproject/monty_lab/blob/dc3b78e962b47e4b54820f5b8703c85617786b55/experiments/configs/config_utils.py#L15 working, it does not. I'm looking into what to do there. I saw only one usage of that function in `profiled_runs.py`.

tristanls:
@vkakerbeck, to import configurations from the `benchmarks` folder, I needed to turn it into a module structure and use absolute imports like `benchmarks.configs.defaults` as opposed to relative imports like `.defaults`. However, to make the absolute imports work, I needed to add the root directory to the system path, which explains `sys.path.insert(...)` in `run.py` and `run_parallel.py`. See this commit https://github.com/thousandbrainsproject/tbp.monty/pull/153/commits/c06790d269c9c015fb3ea4d78e720de3b67070d5

With these changes, we can update https://github.com/thousandbrainsproject/monty_lab/blob/dc3b78e962b47e4b54820f5b8703c85617786b55/experiments/configs/config_utils.py#L15 by adding `sys.path.insert(0, os.path.expanduser("~/tbp/tbp.monty"))`.

tristanls:
@scottcanoe, I had to update the tutorials to accommodate the new way of setting up experiment configurations where names are separate from the config. The relevant changes are in [this commit diff](https://github.com/thousandbrainsproject/tbp.monty/compare/70f7aa0b6fd7c83e998b7809740a557d2d06fd6b..4aa406744cb8f06c3e841c4ebd40c459c6ce40df). Please take a look to check that I didn't mess up the tutorial. Please comment on the file in this pull request if you have comments.

@vkakerbeck, similarly, I had to update the tutorials and develop a way for people to specify their experiments without doing too much work. The [latest commits since your last review](https://github.com/thousandbrainsproject/tbp.monty/compare/70f7aa0b6fd7c83e998b7809740a557d2d06fd6b..4aa406744cb8f06c3e841c4ebd40c459c6ce40df) cover these updates.

vkakerbeck:
Nice! Thanks for remembering the tutorial updates (I keep forgetting about those!) and for looking into the config import from monty_lab.

benchmarks/run.py
```diff
@@ -8,14 +8,83 @@
...
+        field.name for field in fields(MontyWorldHabitatExperiments)
+    ]
+    pretraining_experiment_names = [
+        field.name for field in fields(PretrainingExperiments)
+    ]
+    ycb_experiment_names = [field.name for field in fields(YcbExperiments)]
+    my_experiment_names = [field.name for field in fields(MyExperiments)]
+
+    CONFIGS = dict()
+    for experiment in experiments:
```
nielsleadholm, line 62: Is it possible to re-work most of this code both here and in `run_parallel.py` as 1-2 reusable functions? To my eye the main difference is just the presence of the for loop. 

This might help to highlight that this config building code exists, and that if you create a new set of configs (other than `MY_EXPERIMENTS`), then this code will need to be changed (but only once, not both in run and run_paralllel).

This separate file could also include some short documentation outlining this config approach ("Import only the required CONFIGS to limit extraneous imports and therefore dependency on various libraries such as Habitat").

benchmarks/configs/names.py
```diff
@@ -0,0 +1,83 @@
...
+#
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+
+from dataclasses import dataclass, fields
+
```
nielsleadholm, line 11: Maybe worth adding some documentation here for why we have this approach using dataclasses ("Prevent drift in names etc.")? Just imaging someone coming back to this in 6 months might be a bit confused by this decision.

benchmarks/configs/defaults.py
```diff
@@ -0,0 +1,124 @@
...
+min_eval_steps = 20
+
+monty_models_dir = os.getenv("MONTY_MODELS")
+
+# v6 : Using TLS for point-normal estimation
+# v7 : Updated for State class support + using new feature names like pose_vectors
+# v8 : Using separate graph per input channel
+# v9 : Using models trained on 14 unique rotations
+# v10 : Using models trained without the semantic sensor
+pretrained_dir = os.path.expanduser(
```
nielsleadholm, line 122: I'm 50/50 about this and maybe this is the most practical spot, but not sure if this default is a bit out of place relative to the others, given the specific connection to YCB?

nielsleadholm (commented):
Really nice, makes a lot of sense and I like e.g. the separation out of the typical defaults. Just some minor clarifications/suggestions around things like documentation.

benchmarks/configs/names.py
tristanls: Added, now reads:
```python
"""The names of declared experiments grouped by category.

This module keeps experiment names separate from the configuration for the
experiments. The reason for doing this is so that we can import the configurations
selectively to avoid importing uninstalled dependencies (e.g., not installing a
specific simulator).

The use of dataclasses assists in raising early errors when experiment names defined
here and the corresponding experiment configurations drift apart. For additional
discussion, see: https://github.com/thousandbrainsproject/tbp.monty/pull/153.
"""
```

benchmarks/run.py
tristanls: Commits changed after rebase; the code was coalesced into `benchmarks/configs/load.py`. The note I was referencing is:
```python
def select_config(experiment: str) -> dict:
    """Select and import the configuration group for the given experiment.
    Args:
        experiment: The name of the experiment to import the configuration for.
    Returns:
        The imported configuration group for the given experiment.
    Note:
        We import the configurations selectively to avoid importing uninstalled
        dependencies. For example, if someone does not install a specific simulator,
        then do not import the corresponding configuration group.
    """
```

PR #155, opened 2025-01-27
Update model version numbers in docs

nielsleadholm:
I noticed a few of the references to the models in the documentation reference the older (v9) models, this updates these. I double checked and the curl command still works as expected.


vkakerbeck (approved):
Great, thanks for updating these!

Issue #152, opened 2025-01-24
RFC 5 Easier RFCs Implementation


tristanls:
This issue tracks the implementation of [RFC 5 Easier RFCs](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0005_easier_rfcs.md)

PR #156, opened 2025-01-27
docs: add minimal rfc template and update docs to accommodate

tristanls:
This pull request is part of [RFC 5 Easier RFCs Implementation](https://github.com/thousandbrainsproject/tbp.monty/issues/152).

@hlee9212, @scottcanoe, [y'all suggested to prompt to add images, etc in the minimal template](https://github.com/thousandbrainsproject/tbp.monty/pull/148#discussion_r1924651484). To keep the minimal template minimal, I added that prompt to RFC overall workflow instructions instead. See Step 3, "Fill in the RFC". Does that work for you, or does that specific information need to be included in the minimal template?

In the comprehensive template, the Markdown header link syntax is removed, as it does not work, from what I can tell. Extra whitespace is also removed. Otherwise, there are no changes.

FYI: @vkakerbeck, @nielsleadholm, @codeallthethingz, @ramyamounir


scottcanoe (approved):
Looks good and quite clear. The links to example RFCs are a good addition as well.

Issue #123, opened 2024-12-31
`run_parallel.py` sometimes hangs


scottcanoe:
### Describe the bug

When using `run_parallel.py,` threads/processes sometimes fail to close after the experiment has run. This occurs just before post-parallel cleanup. Whether the program hangs at this point seems random, and it occurs for me roughly half of the time.

I'm on an Apple M3 Max, Sequoia 15.1.1, using miniconda in accordance with Monty's [installation instructions](https://thousandbrainsproject.readme.io/docs/getting-started). This issue only occurs when I run experiments locally.

Strangely, this only started happening a couple of weeks ago, so perhaps it's tied to a software update. I've tried closing basically every app on my computer with the thinking that some resource isn't being released that's preventing threads/processes in a waiting state, but so I haven't found any factors that affect this behavior.

I'll post any updates here as I learn more about the problem.

PR #158, opened 2025-01-28
Don't create unused parallel processes.

scottcanoe:
We currently create a multiprocessing pool of size `num_parallel` (typically 16) regardless of whether `num_parallel` processes are needed. For example, if we run `python run_parallel.py -e only_surf_agent_training_10obj`, we create a pool of size 16 but only 10 processes are used (one per object). This leaves 6 processes doing nothing.

This PR modifies `run_episodes_parallel` so that a smaller pool will be created if the number of parallel configs to run is fewer than the number of parallel processes requested. More specifically, the pool is of size `min(len(configs), num_parallel)`.

The remaining changes in this PR are due to added docstrings and type annotation.

This change was motivated by trying to find out why `run_parallel.py` [sometimes hangs](https://github.com/thousandbrainsproject/tbp.monty/issues/123) on mac. The problem is non-deterministic, but some reading indicated that processes doing nothing could be an issue. Given the non-deterministic nature of the error, I can't know whether this PR will solve that problem, but it seems like a reasonable improvement regardless.


src/tbp/monty/frameworks/run_parallel.py
```diff
@@ -513,14 +563,55 @@ def generate_parallel_eval_configs(exp, experiment_name):
...
+    is_unittest: bool = False,
 ):
+    """Run an experiment in parallel.
+
+    This function runs an experiment in parallel by breaking a config into a set
+    of configs that can be run in parallel, executing them, and performing cleanup
+    to consolidate logs across parallelized configs. How configs are split up into
+    parallelizable configs depends on the type of experiment. For supervised
+    pre-training, a config is created for each object, and that config will run all
+    object rotations in serial. For evaluation experiments, a config is created for
```
nielsleadholm, line 581: It might be worth just adding that the reason we don't parallelize across rotations during learning is so that we don't run into issues with merging the learned graphs obtained from different rotations. This is not a problem during evaluation where we are only concerned with the output object ID and rotation, and evaluation is therefore "embarrassingly parallelizable".

nielsleadholm (approved):
Nice, thanks for making this change, makes sense. 

And thanks for the clear documentation you added, just added a single suggestion.

vkakerbeck (approved):
Nice! Good catch.

A random side thought (nothing that needs to be incorporated in this PR) I had when reading this: I wonder if it would be faster if we also just created one process per object during evaluation. Creating these processes is expensive and if inference is fast (like in the 10 object experiments) then it adds significant overhead (leading to the plateauing speedup when adding more processes). Could be worth a try at some point.

src/tbp/monty/frameworks/run_parallel.py
scottcanoe: Good idea, thanks!

scottcanoe:
Thanks @vkakerbeck and @nielsleadholm! @vkakerbeck using one process per object during evaluation definitely sounds promising. I'll tuck that away and give it a try!

tristanls (approved):

PR #159, opened 2025-01-28
RFC 8 Reorganization of Repositories

nielsleadholm:
This RFC proposes how we might reorganize our repositories to be comprehensible and maintain code quality going forward.


rfcs/0007_reorganize_repositories.md
```diff
@@ -0,0 +1,74 @@
...
+2. `monty_lab` - This is what `nupic.monty/projects` used to be, and is currently a catch-all for code that is not part of the main Monty framework.
+
+The high-level proposal is to break-up `monty_lab` where appropriate, and have a lower threshold for creating new, independent repositories.
+
+More specifically, the proposed structure and description of each repository would be:
+
+`tbp.monty`
+- Core code-base for the Monty framework.
+- All code has undergone thorough review, and Circle-CI runs unit-tests and style checks on any new code.
+- Should not become bloated with experimental code that is unlikely to be re-used in future work.
```
vkakerbeck, line 20: I would make it even more explicit. Something like "should not contain experimental code. The only experiment configs in this repository are those used for our benchmark experiments"

rfcs/0007_reorganize_repositories.md
```diff
@@ -0,0 +1,74 @@
...
+
+`monty_lab`
+- A repository to house discontinued or paused research projects like  `high_dim_coincidence_detection` and `grid_cells`.
+- Local forks of `monty_lab` can be used by contributors to track in-progress work. However, this does not mean that such code, once mature enough to be shared, should necessarily be merged into the main `monty_lab` repository.
+- In general, if the code is likely to be re-used in the future, or forms part of a paper, it should *not* go in `monty_lab` (more on this below).
+- Circle-CI does not include style or unit-test checks on new code. 
+- When merging a discontinued project, PR review does not need to be as thorough as for code pushed to the main `tbp.monty` repository.
+- Not intended as a code-base that community members can contribute to. As such, issues like bugs are less likely to be noticed and fixed. 
+
+We could then have two additional types of repositories, with no predefined number of these. These repository *types* would be:
```
vkakerbeck, line 30: I would suggest to just have one type of repository (structure could be defined by a template repository). There will inevitable be more usecases for repositories that these two (like the plotting repo that Will made, an exploratory analysis repo, a repo for experiments on a specific topic that was not published as a paper,...) The name and readme of the repo should be sufficient to inform people of the content.

rfcs/0007_reorganize_repositories.md
```diff
@@ -0,0 +1,74 @@
...
+
+`tbp_package.name_of_package`
+- E.g. `tbp_package.floppy`
+- A package that is a collection of modules that are intended to be re-used in other projects. This does not imply that it is a package that will be maintained long-term for a broader community, but is rather something that we see ourselves re-using at the TBP.
+- Should be well-documented, and have some unit-test coverage and style checks. PR reviews should be of a similar standard to `tbp.monty`.
+- Open to contributions from the community.
+
+`tbp_paper.name_of_paper`
+- E.g. `tbp_paper.dmc_paper`
```
vkakerbeck, line 40: minor side note: I don't think we should call the dmc paper. It's not an abbreviation anyone outside of TBP will understand. Why not just monty_capabilities_paper or whatever title we give the paper in the end

rfcs/0007_reorganize_repositories.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Motivation
+
+As the Thousand Brains Project grows, it will be important to have a better way of organizing the codebase. This RFC proposes we establish a sustainable structure that ensures code remains high quality and accessible, while minimizing overhead for contributors.
+
+# Open Questions
+
+- If a given paper is separated into multiple repositories, will this create unecessary challenges for someone who wants to replicate the results (e.g. generating a figure)? We can think about how to structure results folders to make this as easy as possible.
+- Should we be open to merging code into `monty_lab`'s `main` as a form of staging before it is eventually moved to a final repository? This would be akin to what has organically happened with the DMC work.
+    - An advantage of this would be to make it easier for maintainers to review provisional code that is not yet ready for a final repository.
```
vkakerbeck, line 67: I would tend not to do that. There is no problem with starting an empty repository and adding slowly to it over time. If it is a separate repo we can add style and unit test check from the start to make sure quality standards are adhered to.

rfcs/0007_reorganize_repositories.md
```diff
@@ -0,0 +1,74 @@
...
+
```
vkakerbeck, line 68: I would add another open question (or a proposal for this somewhere else in this RFC) about whether we should have a place where we list all of our repositories with a description and links to them. Similar to the README on monty_lab.

vkakerbeck (commented):
Nice, thanks for outlining this! I left a few comments, let me know what you think.

rfcs/0007_reorganize_repositories.md
nielsleadholm: Thanks yeah good point, I'll leave this open for now during the discussion.

rfcs/0007_reorganize_repositories.md
nielsleadholm: Thanks, added.

rfcs/0007_reorganize_repositories.md
nielsleadholm: Yeah good point, I've updated it to what I think is a reasonable abbreviation for the working title of "Thousand Brains Systems for Rapid, Robust Learning and Inference".

rfcs/0007_reorganize_repositories.md
nielsleadholm: Thanks, I've split this up and clarified it a bit more.

rfcs/0007_reorganize_repositories.md
nielsleadholm: Ok yeah, this is maybe the one point I'm on the fence about. Couldn't most of those fit well within `monty_lab`? Personally I feel like having some prefix in the name to clarify it's purpose is useful, but maybe that's just me.

nielsleadholm:
Thanks for the useful comments @vkakerbeck , have made some adjustments.

rfcs/0007_reorganize_repositories.md
```diff
@@ -0,0 +1,77 @@
...
+1. `tbp.monty` - This is the main repository for the Monty framework.
+2. `monty_lab` - This is what `nupic.monty/projects` used to be, and is currently a catch-all for code that is not part of the main Monty framework.
+
+The high-level proposal is to break-up `monty_lab` where appropriate, and have a lower threshold for creating new, independent repositories.
+
+More specifically, the proposed structure and description of each repository would be:
+
+`tbp.monty`
+- Core code-base for the Monty framework.
+- All code has undergone thorough review, and Circle-CI runs unit-tests and style checks on any new code.
```
tristanls, line 19: **issue**: We don't use Circle-CI. I think this is intended to read "CI" for "Continuous Integration?"

rfcs/0007_reorganize_repositories.md
```diff
@@ -0,0 +1,77 @@
...
+
+`monty_lab`
+- A repository to house discontinued or paused research projects like  `high_dim_coincidence_detection` and `grid_cells`.
+- Local forks of `monty_lab` can be used by contributors to track in-progress work. However, this does not mean that such code, once mature enough to be shared, should necessarily be merged into the main `monty_lab` repository.
+- In general, if the code is likely to be re-used in the future, or forms part of a paper, it should *not* go in `monty_lab` (more on this below).
+- Circle-CI does not include style or unit-test checks on new code. 
+- When merging a discontinued project, PR review does not need to be as thorough as for code pushed to the main `tbp.monty` repository.
+- Not intended as a code-base that community members can contribute to. As such, issues like bugs are less likely to be noticed and fixed. 
+
+We could then have two additional types of repositories, with no predefined number of these. These repository *types* would be:
```
tristanls, line 31: **thought**: I think selecting an additional namespacing within our namespacing is too much. `tbp` is our namespace, and we can use that. For example: `tbp.floppy` or `tbp.tbs_for_rapid_robust_learning_and_inference`. The `tbp.` prefix was mainly used to avoid namespace collision with `monty` or `floppy`. If we have a paper named the same as a package, we could resolve that conflict without additional prefixes.

rfcs/0007_reorganize_repositories.md
```diff
@@ -0,0 +1,77 @@
...
+- All code has undergone thorough review, and Circle-CI runs unit-tests and style checks on any new code.
+- Should not become bloated with code that is unlikely to be re-used in future work, or is of sufficient complexity that it can be better understood as a separate package.
+- Should not contain configs for various experiments; the only configs in tbp.monty are those used for our benchmark experiments.
+
+`monty_lab`
+- A repository to house discontinued or paused research projects like  `high_dim_coincidence_detection` and `grid_cells`.
+- Local forks of `monty_lab` can be used by contributors to track in-progress work. However, this does not mean that such code, once mature enough to be shared, should necessarily be merged into the main `monty_lab` repository.
+- In general, if the code is likely to be re-used in the future, or forms part of a paper, it should *not* go in `monty_lab` (more on this below).
+- Circle-CI does not include style or unit-test checks on new code. 
+- When merging a discontinued project, PR review does not need to be as thorough as for code pushed to the main `tbp.monty` repository.
```
tristanls, line 28: **thought**: I think having a lesser standard for some repositories over others will be toilsome in the long run. The code standards are not there to make the code look pretty but to reduce the cognitive load of contributors and speed up our ability to change the code without breaking existing functionality. After two and a half months of its existence, I was already asked to improve the code quality of `monty_lab`.

rfcs/0007_reorganize_repositories.md
```diff
@@ -0,0 +1,77 @@
...
+- Not intended as a code-base that community members can contribute to. As such, issues like bugs are less likely to be noticed and fixed. 
+
+We could then have two additional types of repositories, with no predefined number of these. These repository *types* would be:
+
+`tbp_package.name_of_package`
+- E.g. `tbp_package.floppy`
+- A package that is a collection of modules that are intended to be re-used in other projects. This does not imply that it is a package that will be maintained long-term for a broader community, but is rather something that we see ourselves re-using at the TBP.
+- Should be well-documented, and have some unit-test coverage and style checks. PR reviews should be of a similar standard to `tbp.monty`.
+- The code should be of high enough quality that it can be used with confidence with the current `tbp.monty` codebase at the time of creation. However, there is not an expectation that it will be usable "as-is" in the future if `tbp.monty` undergoes significant changes. This is to reduce the burden on contributors of such package repositories, and given the unknowability of how often a package will be used in the future.
```
tristanls, line 37: **thought**: Regarding usability "as-is," I think we should use the well-recognized and adopted https://semver.org/ standard to communicate the level of stability, usability, and compatibility of our code.

rfcs/0007_reorganize_repositories.md
```diff
@@ -0,0 +1,77 @@
...
+# Summary
+
+The proposal is to establish a better way of organizing Thousand Brains Project repositories.
+
+Currently there are two repositories:
+
+1. `tbp.monty` - This is the main repository for the Monty framework.
+2. `monty_lab` - This is what `nupic.monty/projects` used to be, and is currently a catch-all for code that is not part of the main Monty framework.
+
+The high-level proposal is to break-up `monty_lab` where appropriate, and have a lower threshold for creating new, independent repositories.
```
tristanls, line 13: **thought**: I agree we should have a lower threshold for creating repositories. I have 191 public repositories on GitHub, and it's fine for me.

rfcs/0007_reorganize_repositories.md
```diff
@@ -0,0 +1,77 @@
...
+# Motivation
+
+As the Thousand Brains Project grows, it will be important to have a better way of organizing the codebase. This RFC proposes we establish a sustainable structure that ensures code remains high quality and accessible, while minimizing overhead for contributors.
+
+# Open Questions
+
+- If a given paper is separated into multiple repositories, will this create unecessary challenges for someone who wants to replicate the results (e.g. generating a figure)? We can think about how to structure results folders to make this as easy as possible.
+- Should we be open to merging code into `monty_lab`'s `main` as a form of staging before it is eventually moved to a final repository? This would be akin to what has organically happened with the recent paper work.
+    - An advantage of this would be to make it easier for maintainers to review provisional code that is not yet ready for a final repository.
+- What would be a good place for providing an overview of all the repositories in the codebase, with descriptions and links to them? Perhaps on https://thousandbrainsproject.readme.io/docs/ ?
```
tristanls, line 69: **suggestion**: With helpful descriptions and tags, the built-in GitHub page can serve this purpose: https://github.com/orgs/thousandbrainsproject/repositories

rfcs/0007_reorganize_repositories.md
nielsleadholm: Do you mean you would prefer if we didn't have some code which we say can be of a lesser standard?

Are you referring to the review of floppy? That was specifically highlighted because it does not fit in with the other code found in `monty_lab`, which is why we are having this discussion to presumably move it into its own package-like repository.

The idea for having a different standard in `monty_lab` is that it would essentially be an archive of discontinued work.

nielsleadholm:
Thank you for the ongoing discussion, @tristanls I have responded to your comments. 

@ramyamounir I have added you. Please note that as a Maintainer, you can also choose to Abstain from reviewing an RFC.

rfcs/0000_reorganize_repositories.md
```diff
@@ -0,0 +1,75 @@
...
+- We should have a low threshold for creating new repositories where appropriate.
+- It is encouraged to add a description and tags to repositories such that the [overview page of the Thousand Brains Project GitHub organization](https://github.com/orgs/thousandbrainsproject/repositories) provides helpful context.
+
+# Motivation
+
+As the Thousand Brains Project grows, it will be important to have a better way of organizing the codebase. This RFC proposes we establish a sustainable structure that ensures code remains high quality and accessible, while minimizing overhead for contributors.
+
+# Open Questions
+
+- If a given paper is separated into multiple repositories, will this create unecessary challenges for someone who wants to replicate the results (e.g. generating a figure)? We can think about how to structure results folders to make this as easy as possible.
```
ramyamounir, line 67: Good point, I would probably vote against splitting one paper onto multiple repositories. 

I like the idea of lowering the threshold to creating new repositories. It's better than having a single repo for all the papers e.g., [htmpapers](https://github.com/numenta/htmpapers), but I also think that grouping everything about a single paper into one repo is a better structure and a reasonable level of abstraction. If we find that some part (e.g., ViT implementation) is being repeated in multiple papers, we can make it a package like `tbp.floppy` and import it when needed.

ramyamounir (commented):
I agree to everything here, but just left one comment on the splitting of papers onto multiple repositories.

rfcs/0000_reorganize_repositories.md
```diff
@@ -0,0 +1,75 @@
...
+- Not intended as a code-base that community members can contribute to. As such, issues like bugs are less likely to be noticed and fixed. 
+
+Additional repositories can then be created as needed. Two typical examples would be:
+
+`tbp.name_of_package`
+- E.g. `tbp.floppy`
+- A package that is a collection of modules that are intended to be re-used in other projects. This does not imply that it is a package that will be maintained long-term for a broader community, but is rather something that we see ourselves re-using at the TBP.
+- Should be well-documented, and have some unit-test coverage and style checks. PR reviews should be of a similar standard to `tbp.monty`.
+- The code should be of high enough quality that it can be used with confidence with the current `tbp.monty` codebase at the time of creation. However, there is not an expectation that it will be usable as-is if `tbp.monty` undergoes MAJOR changes (per [semver](https://semver.org/)). This is to reduce the burden on contributors of such package repositories, and given the unknowability of how often a package will be used in the future.
```
scottcanoe, line 37: Once a paper is published, we want its repo to work forever. I think this means adopting some kind of versioning scheme for tbp packages, including `tbp.monty`. In my mind, we would ideally have project specifications in a paper repo that would require, for example, `tbp.monty=1.2` or `floppy=1.0`. I don't know enough about package versioning to know how to do this, but this would keep the paper repo stable and remove the burden on package contributors.

scottcanoe (commented):
Thanks @nielsleadholm! I left a comment about versioning packages, but I'd also add that I think its preferable to avoid splitting a paper's repo into multiple repos if possible.

rfcs/0007_reorganize_repositories.md
tristanls: 🤔, I think I was responding to `monty_lab` as it is currently being used and did not imagine it in the state described in this RFC. 

Going back through and thinking of `monty_lab` as outlined in the RFC, I would say that we should probably make it an archive and no longer merge code to it. If we have some future code that we want to archive, we could archive the repo that the code is in and not go through the effort of moving it over.

rfcs/0007_reorganize_repositories.md
tristanls: The organization's README page is a good idea. I think that meets the intent better than the repositories page.

rfcs/0000_reorganize_repositories.md
codeallthethingz: I would stick with a one big repo per paper.  I've used git-submodules in the past, its painful in lots of unexpected ways.

If you have a monorepo for a paper, you can still have directories and wildly different libraries / environments at different levels in the folder structure.  It allows PR merges to be atomic units of work, the same reason the docs are co-located with the code.

codeallthethingz: @hlee9212 yeah, exactly.  Though you may want a root `.gitignore` as it is a bit simpler to think about.  I don't have experience with placing them in nested folder structures.

rfcs/0007_reorganize_repositories.md
nielsleadholm: Ok yeah I liked this idea, so essentially archive the current `monty_lab` and discontinue its use going forward. For future code that needs archiving, we can just use the [archive functionality](https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories). @vkakerbeck @scottcanoe @hlee9212 @ramyamounir any objects to this?

rfcs/0000_reorganize_repositories.md
nielsleadholm: Yeah I think this is a broader point which will always be a challenge. E.g., I'm not sure we can 100% guarantee that someone in 5 years could replicate the results, given the nature of bit-rot. However we would at least want to have a clearly defined version of tbp.monty that was used for generating the code, such that as long as Monty dependencies like Habitat and Python 3.8 are still running on someones system, then they should be able to recreate the results.

@tristanls do you have any specific suggestions on this?

codeallthethingz (approved):
I have used the package template suggested here: https://github.com/thousandbrainsproject/tbp.python_package a couple of times and it's great.

rfcs/0000_reorganize_repositories.md
tristanls: **suggestion**: I think we should version `tbp.monty` so that a paper can reference a specific version.

**suggestion**: I think we should publish `tbp.monty` versions so that the paper repository can use it as a dependency. Since `tbp.monty` relies on `conda`, then the [anaconda repository seems like the appropriate place to publish](https://docs.anaconda.com/anacondaorg/user-guide/packages/conda-packages/).

rfcs/0007_reorganize_repositories.md
tristanls: One issue with namespacing repository names is that if they become "real" and earn a `tbp.*` prefix, all the links break because the repository name changes. An alternative method that wouldn't break links would be to use repository tags.

nielsleadholm: That's a good point Ramy about how to name these. So if we are using repository tags as you suggest @tristanls , would archived repositories (and therefore effectively all repositories) have the `tbp.*` prefix? If there's no issue with this, then I'll make sure this is part of the description.

tristanls: @nielsleadholm, I have no issues with `tbp.*` prefix for all repos, as they're likely to be Python, and we're using that to avoid PyPi name collisions.

**note**: There are currently exceptions, for example https://github.com/thousandbrainsproject/cla, which does not have a `tbp.*` prefix.

**suggestion**: Maybe highlight why we put the `tbp.*` prefix in front of things? Again, as far as I understand, we originally put the `tbp.*` prefix in `tbp.monty` because [a `monty` project already existed on PyPi](https://pypi.org/project/monty/). So, this was to deconflict the name of a published package. By having our Python repositories and packages retain the `tbp.*` prefix, we are carving out our own namespace in the PyPi global namespace. In other contexts, this may not be necessary. For example, `conda` has the concept of organizations, where our prefix is `thousandbrainsproject::`. Other languages offer namespacing solutions as well.

nielsleadholm: Great thanks Tristan, I'll make sure to highlight that. I believe that was while I was on PTO, so just double checking with @vkakerbeck that you have the same memory as Tristan as to the reason for the tbp.* prefix? I can update the guidance to advise that this should be used for all Python repositories.

vkakerbeck: Yes, avoiding name collisions was the reason we added the tbp. prefix to monty. I like the idea of keeping that as a convention for other repositories, as it seems like a nice branding opportunity to clearly associate our repos and Pypi packages with this project. However, if there is a strong argument against doing this for repos without naming conflicts (for example to avoid unnecessary nesting of folders) I also don't mind not doing this.

nielsleadholm:
Thank you everyone for your feedback on this, I'd like to now motion for final comment period (FCP). All Maintainers please indicate your final views on the RFC.

To summarize the changes I have made following feedback:
- Clarified that a paper will have a single repository for it, with sub-folders used to separate experiments / code using diverse frameworks (e.g. Monty vs. Pytorch). 
- Clarified that `monty_labs` will be discontinued. Future projects should create their own repository, all of which should use the `tbp.` prefix. 
- We will add an organization README to help someone coming to the main TBP github page learn about the different repositories that exist.

ramyamounir (approved):
All are great changes, thank you! I vote for merging.

rfcs/0000_reorganize_repositories.md
```diff
@@ -0,0 +1,68 @@
...
+- All code has undergone thorough review, and Continuous Integration runs unit-tests and style checks on any new code.
+- Should not become bloated with code that is unlikely to be re-used in future work, or is of sufficient complexity that it can be better understood as a separate package.
+- Should not contain configs for various experiments; the only configs in tbp.monty are those used for our benchmark experiments.
+
+`monty_lab`
+- This would be archived in its entirety, along with its discontinued / paused research projects like  `high_dim_coincidence_detection` and `grid_cells`.
+
+New repositories can then be created as needed for future projects that do not belong in `tbp.monty`. Two typical examples are below, although a repository need not be a package or contain code for a paper:
+
+`tbp.name_of_package`
```
tristanls, line 28: **suggestion (non-blocking)**: It may be worth highlighting the template repositories we have currently available: https://github.com/thousandbrainsproject/tbp.python_package and https://github.com/thousandbrainsproject/tbp.python_library.

tristanls (approved):
Thank you for putting this together. I left one non-blocking suggestion. I think this is good to merge 👍.

vkakerbeck (approved):
Looks great, thank you for writing this up and all the updates!

rfcs/0000_reorganize_repositories.md
```diff
@@ -0,0 +1,68 @@
...
+# Summary
+
+The proposal is to establish a better way of organizing Thousand Brains Project repositories.
+
+Currently there are two repositories:
+
+1. `tbp.monty` - This is the main repository for the Monty framework.
+2. `monty_lab` - This is what `nupic.monty/projects` used to be, and is currently a catch-all for code that is not part of the main Monty framework.
+
+The high-level proposal is to discontinue the use of `monty_lab` (archiving the existing repository), and instead have a very low threshold for creating new, independent repositories for different projects.
```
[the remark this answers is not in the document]
nielsleadholm: Yeah good point, I've updated the guidance on this. Let me know if it still isn't clear.

codeallthethingz (approved):

scottcanoe (approved):
Looks great, much appreciated!

PR #160, opened 2025-01-28
chore: clarifications in pre-training tutorial

codeallthethingz:
changed model -> config, which I think is correct.
Moved the placement suggestion of a script to before the script contents.


scottcanoe (approved):
LGTM, thanks for the changes!

PR #166, opened 2025-01-31
Bug: tool was not using the correct relative path of csv files.

codeallthethingz:
Fixes a bug for relative imports for csv files in md files.
Also adds an ignore table string so that tables can be talked about in the docs without getting replaced with an actual table.


tristanls (approved):
LGTM

PR #167, opened 2025-01-31
Docs: Inserting tables

codeallthethingz:
Adds documentation in the style.md that gives an example for inlining CSV files.

https://thousandbrainsproject.readme.io/v0.0-codeallthethingz-docs-4-tables/docs/style-guide#csv-data-tables

### DIFF
--- a/docs/contributing/style-guide.csv
+++ b/docs/contributing/style-guide.csv
@@ -1,0 +1,7 @@
+Year,   Avg Global Temp. (°C),  Pirates|align right|hover Pirate Count
+1800,   14.3,                   50 000
+1850,   14.4,                   15 000
+1900,   14.6,                    5 000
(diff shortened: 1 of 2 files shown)


vkakerbeck (approved):
beautiful :) That will come in handy!

scottcanoe (approved):
Awesome! So do the columns widths in the markdown tables come directly from the csv file then? (excluding bar-separated parts)

nielsleadholm (approved):
Love it, thanks for adding this!

codeallthethingz:
> Awesome! So do the columns widths in the markdown tables come directly from the csv file then? (excluding bar-separated parts)

@scottcanoe there are no column widths.  They just expand normally like HTML tables do.

PR #177, opened 2025-02-13
Fix: incorrect header row in CSV for Unsupervised Benchmarks

codeallthethingz:
Thanks to @scottcanoe for catching this.

Table is fixed here: https://thousandbrainsproject.readme.io/v0.0-codeallthethingz-incorrect-unsupervised-csv-header/docs/benchmark-experiments#results-2


scottcanoe (approved):
LGTM.

scottcanoe:
Thanks for the fix!

PR #178, opened 2025-02-13
Chore: Build the docs when benchmarks/results/** change

codeallthethingz:
This PR fixes a problem where the docs are not getting rebuilt by the the preview system when the benchmarks/results change as they were not included in the build path trigger.


tristanls (approved):
LGTM

PR #185, opened 2025-02-21
Fix Evidence threshold bug

vkakerbeck:
There was a bug in the `_get_evidence_update_threshold` function in the EvidenceLM where instead of setting the threshold for upding a hypothesis to `x_percent_threshold` it was set to `1 - x_percent_threshold`.

Results are as expected:
- After the change is applied, accuracy and run time go down (green vs. grey) because we test way fewer hypotheses at every step.
- If we set the evidence update threshold manually back to 80% instead of 20%, we get exactly the same results as before the fix (turquoise vs. grey). This is just a sanity check.

Screenshot 2025-02-21 at 5 06 34 PM

Given that the accuracy decreases so much after the fix, it seems like it would be good to create a separate variable for the evidence update threshold (if it is a percentage) instead of using the x_percent_threshold variable defined for the terminal condition. This PR includes a proposed way to add an option to specify `evidence_update_threshold` as a percentage.

I ran some hyperparameter tests for different values:
Screenshot 2025-02-21 at 5 08 39 PM

Remaining tasks before this can become a full PR:

- [ ] Test this parameter on 77 object experiment
- [ ] Test whether there is a different optimal `x_percent_threshold` value, given that these two parameters as disentangled now
- [ ] Rerun all benchmarks & update results with the new values (if we set new values. Currently results should be exactly the same).

### DIFF
--- a/src/tbp/monty/frameworks/models/evidence_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching.py
@@ -1777,7 +1777,7 @@ class EvidenceGraphLM(GraphLM):
         elif self.evidence_update_threshold == "x_percent_threshold":
             max_global_evidence = self.current_mlh["evidence"]
             x_percent_of_max = max_global_evidence / 100 * self.x_percent_threshold
-            return x_percent_of_max
+            return max_global_evidence - x_percent_of_max
         elif self.evidence_update_threshold == "all":
             return np.min(self.evidence[graph_id])
         else:
--- a/src/tbp/monty/frameworks/models/evidence_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching.py
@@ -176,7 +176,7 @@ class EvidenceGraphLM(GraphLM):
         initial_possible_poses: initial possible poses that should be tested for.
             In ["uniform", "informed", list]. default = "informed".
         evidence_update_threshold: How to decide which hypotheses should be updated.
-            In [int, float, 'mean', 'median', 'all', 'x_percent_threshold'].
+            In [int, float, '[int]%', 'mean', 'median', 'all', 'x_percent_threshold'].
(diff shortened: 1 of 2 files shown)


nielsleadholm (commented):
The change looks good! Will review properly when it's a full PR but don't see any issue with the updated code.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1774,16 +1774,27 @@ def _get_evidence_update_threshold(self, graph_id):
...
+        ) and self.evidence_update_threshold.endswith("%"):
+            percentage_str = self.evidence_update_threshold.strip("%")
+            percentage = float(percentage_str)
+            assert (
+                percentage >= 0 and percentage <= 100
+            ), "Percentage must be between 0 and 100"
+            max_global_evidence = self.current_mlh["evidence"]
+            x_percent_of_max = max_global_evidence * (percentage / 100)
+            return max_global_evidence - x_percent_of_max
         elif self.evidence_update_threshold == "x_percent_threshold":
```
tristanls, line 1788: **suggestion**: Perhaps rename `"x_percent_threshold"` and the variable to something else? Now that `evidence_update_threshold` can take an "x percent" threshold value (e.g., "80%"), this name becomes quite confusing to parse.

vkakerbeck: I was planning to do this in a separate PR (as it would touch a lot of the other code). Is that ok?

vkakerbeck:
Just turned this into a real PR as it doesn’t affect the benchmark results. It just fixes the bug. Then we can do a follow-up PR with updating to optimal parameters and updating the benchmark results with those.

nielsleadholm (approved):
Looks great, thanks for fixing this so quickly!

PR #186, opened 2025-02-21
RFC 7 Monty Versioning

tristanls:
For easier reading, see the RFC on my fork: https://github.com/tristanls/tbp.monty/blob/rfc_monty_versioning/rfcs/0000_monty_versioning.md

### DIFF
--- a/rfcs/0000_monty_versioning.md
+++ b/rfcs/0000_monty_versioning.md
@@ -116,6 +116,14 @@ jobs:
           git push origin --delete ${{ steps.branch.outputs.name }}
 ```
 
+## Accessing specific versions
+
+The most detailed and specific way to access a specific version of `tbp.monty` is to clone the repository and checkout the specific commit.
+
+If we tag the version commit with the corresponding version number, then the version can be accessed by checking out the tag. GitHub has [releases](https://github.com/thousandbrainsproject/tbp.monty/releases) and [tags](https://github.com/thousandbrainsproject/tbp.monty/tags) pages that list the tagged versions.
+
+If we publish Monty, then the specific published version could be installed via a package manager by specifying the published version.
+
 # References
 
 * [Semantic Versioning](https://semver.org/) is a standard for versioning software.
--- a/rfcs/0000_monty_versioning.md
+++ b/rfcs/0000_monty_versioning.md
@@ -24,6 +24,8 @@ In summary, the version number is `MAJOR.MINOR.PATCH`.
 
 It is worth noting that the `MAJOR` version number `0` is special. When `MAJOR` is `0`, then incrementing the `MINOR` version number ***can include backwards-incompatible changes***. `MAJOR` version number `0` is the usual way of indicating that the software is still in development and not ready for production use.
 
+We will use `MAJOR` version number `0` for the ongoing development of Monty. Incrementing the `MAJOR` version number beyond `0` should be a future RFC, as it will be a significant event.
+
 The current version can be found in [`src/tbp/monty/__init__.py`](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/__init__.py) as the `__version__` variable.
 
 ## Updating the version
--- a/rfcs/0000_monty_versioning.md
+++ b/rfcs/0000_monty_versioning.md
@@ -28,7 +28,7 @@ The current version can be found in [`src/tbp/monty/__init__.py`](https://github
 
 ## Updating the version
 
-### Constraints
+### Proposed constraints
 
 Here are some constraints related to updating the version:
 
@@ -115,6 +115,8 @@ jobs:
         run: |
           git push origin --delete ${{ steps.branch.outputs.name }}
 ```
+
+Release branches are deferred to possible future work. The way release branches would work is that we would never update `__version__` on the main branch. Instead, every time we want to make a new version, we would create a new branch, update `__version__`, and merge it into the release branch. This has some nice properties for automation, however, it interferes with some of the proposed constraints above.
 
 ## Accessing specific versions
 
--- a/rfcs/0000_monty_versioning.md
+++ b/rfcs/0000_monty_versioning.md
@@ -143,7 +143,7 @@ The most detailed and specific way to access a specific version of `tbp.monty` i
 
 If we tag the version commit with the corresponding version number, then the version can be accessed by checking out the tag. GitHub has [releases](https://github.com/thousandbrainsproject/tbp.monty/releases) and [tags](https://github.com/thousandbrainsproject/tbp.monty/tags) pages that list the tagged versions.
 
-If we publish Monty, then the specific published version could be installed via a package manager by specifying the published version.
+If we publish Monty to [PyPI](https://pypi.org/) or [Anaconda](https://anaconda.org/), then the specific published version could be installed via a package manager by specifying the published version.
 
 # References
 
--- a/rfcs/0000_monty_versioning.md
+++ b/rfcs/0000_monty_versioning.md
@@ -1,0 +1,120 @@
+- Start Date: 2025-02-20
+- RFC PR: (leave this empty, it will be filled in after RFC is merged)
+
+# Summary
+
+Begin versioning Monty using [semantic versioning](https://semver.org/).
+
+# Motivation
+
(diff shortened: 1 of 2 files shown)


rfcs/0000_monty_versioning.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Updating the version
+
+### Constraints
+
+Here are some constraints related to updating the version:
+
+* We want the current version to be in the source code.
+* The `main` branch is protected and changes can be pushed only via Pull Requests.
+
```
codeallthethingz, line 35: I would add a constraint that it needs to be programmatically accessible for things like the print_version tool. https://github.com/thousandbrainsproject/tbp.monty/tree/main/tools/print_version

rfcs/0000_monty_versioning.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+# References
+
+* [Semantic Versioning](https://semver.org/) is a standard for versioning software.
+* [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) is a standard for commit messages.
+
+# Future possibilities
+
+If we were to adopt the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#specification) then the version number updates could be automated based on the commit message.
```
codeallthethingz, line 120: We could also put this in the `tbp.cli` tool which would take care of most of the manual steps.

```
tbp bump-minor
```

codeallthethingz (approved):
LGTM, optional suggestions.

rfcs/0000_monty_versioning.md
```diff
@@ -0,0 +1,120 @@
...
+### Constraints
+
+Here are some constraints related to updating the version:
+
+* We want the current version to be in the source code.
+* The `main` branch is protected and changes can be pushed only via Pull Requests.
+
+### Process
+
+1. Create a Pull Request that ***only*** updates the `__version__` variable in [`src/tbp/monty/__init__.py`](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/__init__.py).
```
vkakerbeck, line 38: Can you elaborate why we would want a PR with only that change instead if including the version number update in the PR with the change that triggers the version update? If we have a PR with a breaking change + a PR with the version number update, which PR would be merged first? I would assume we don't want to add the breaking change into main under the old version. So then we would increment the version before introducing the change? Why couldn't we increment the version in the same PR as the change and have the discussion on whether the change warrants a version update there?

vkakerbeck (commented):
Nice, thanks for outlining this! I like the server proposal. Just one question on the process for updating the version.

nielsleadholm (commented):
Looks great, thanks for putting this together!

Just two high level comments/questions
- Is it possible to elaborate a bit how historical versions will be accessed (e.g. if someone wants to run something from the DMC paper, and so access v0.0.1)? Would this require some use of PyPi or the like? 
- I don't have a good suggestion, and maybe there's no solution, but do you have any thoughts on how we can help us to remember to update the version number when merging PRs into `tbp.monty`? E.g. an automated (but non-blocking) email / message whenever a PR is opened for `tbp.monty`? We might eventually get in the habit, but it feels like we may merge several PRs before suddenly remembering that we need to increment the version number.

rfcs/0000_monty_versioning.md
tristanls: Added a constraint "We want the current version to be programmatically accessible, e.g., by the `tools.print_version.cli` tool." in https://github.com/thousandbrainsproject/tbp.monty/pull/186/commits/fb95a0c214f8f2ee028a21192cb40d4703ef0e55

tristanls:
@nielsleadholm 
> Is it possible to elaborate a bit how historical versions will be accessed (e.g. if someone wants to run something from the DMC paper, and so access v0.0.1)? Would this require some use of PyPi or the like?

I'm adding the following to the RFC in https://github.com/thousandbrainsproject/tbp.monty/pull/186/commits/bc351c6177771652c5fe62d3081ce882ef3078ca, which is intended to answer the above.

## Accessing specific versions

The most detailed and specific way to access a specific version of `tbp.monty` is to clone the repository and checkout the specific commit.

If we tag the version commit with the corresponding version number, then the version can be accessed by checking out the tag. GitHub has [releases](https://github.com/thousandbrainsproject/tbp.monty/releases) and [tags](https://github.com/thousandbrainsproject/tbp.monty/tags) pages that list the tagged versions.

If we publish Monty, then the specific published version could be installed via a package manager by specifying the published version.

tristanls:
@nielsleadholm 

> I don't have a good suggestion, and maybe there's no solution, but do you have any thoughts on how we can help us to remember to update the version number when merging PRs into tbp.monty? E.g. an automated (but non-blocking) email / message whenever a PR is opened for tbp.monty? We might eventually get in the habit, but it feels like we may merge several PRs before suddenly remembering that we need to increment the version number.

I don't think we should update the version with every commit. It will be OK and even desirable to merge several PRs before updating the version number. [I commented above](https://github.com/thousandbrainsproject/tbp.monty/pull/186#discussion_r1968189856) regarding what I expect the workflow to be.

rfcs/0000_monty_versioning.md
tristanls: I also added this future possibility in https://github.com/thousandbrainsproject/tbp.monty/pull/186/commits/7ff85b63a3873e5294e5be74bcdeb2792cdfc609

In the future, integrating into the `main` branch may become cumbersome if the pace of commits to `main` is high. In that case, we will update our versioning strategy. One way to do this is to create a `release` branch that is used for releasing new versions. This note is intended to highlight the possibility, but if we get there, we will get into the details then.

tristanls:
FYI, I added some more context around `MAJOR` version number `0`:

---

It is worth noting that the `MAJOR` version number `0` is special. When `MAJOR` is `0`, then incrementing the `MINOR` version number ***can include backwards-incompatible changes***. `MAJOR` version number `0` is the usual way of indicating that the software is still in development and not ready for production use.

We will use `MAJOR` version number `0` for the ongoing development of Monty. Incrementing the `MAJOR` version number beyond `0` should be a future RFC, as it will be a significant event.

tristanls:
@codeallthethingz, I want to explicitly ask if what I proposed here will work with how we currently use readme.io. I imagine that if we go from `0.0.1` to `0.0.2`, we wouldn't want to create a `v0.0.2` documentation version but instead update the `v0.0` documentation version.

codeallthethingz:
> @codeallthethingz, I want to explicitly ask if what I proposed here will work with how we currently use readme.io. I imagine that if we go from `0.0.1` to `0.0.2`, we wouldn't want to create a `v0.0.2` documentation version but instead update the `v0.0` documentation version.

That is the current behaviour.  We truncate to the minor version for the docs.

vkakerbeck (approved):
Sounds good :) Thanks for the clarifications & updates!

nielsleadholm (approved):
Thanks for the clarifications, looks great!

ramyamounir (approved):
Thanks @tristanls, looks great!

tristanls:
FYI, added expected workflow section from the comment to the RFC

---

### Expected Workflow

By adopting this versioning scheme and process, the expected workflow is as follows:


- commit
- commit
- commit
- commit
- review all commits above and update the version accordingly (in the future this will also publish Monty)
- commit
- commit
- review all commits above and update the version accordingly (in the future this will also publish Monty)
- commit
- ...

tristanls:

I interpret @codeallthethingz, @hlee9212, @vkakerbeck, @nielsleadholm, and @ramyamounir current PR acceptance as **approval**.
@scottcanoe please indicate your decision on the disposition to **merge**:
- approve: via approving this PR or commenting that you approve to merge
- abstain: by commenting that you abstain
- other: by commenting

nielsleadholm (approved):

scottcanoe (approved):
LGTM. I've not worked with GitHub releases and tags, but it sounds clean. Can you clarify what is meant by "if we publish Monty..."? I'm assuming that means putting on PyPi or conda-forge.

tristanls:
@scottcanoe, correct. Currently, I'm thinking conda. I will update the RFC to be more specific.

tristanls:

PR #187, opened 2025-02-25
add DetailedWandbTableStatsHandler to reduce unnesscessary wandb logging

vkakerbeck:
Since we are accumulating a lot of artifacts in wandb (when using run.py instead of run_parallel.py) eating up storage, I separated out the action logging. We currently don't use these tables (if we do, we use the .txt files from the `ReproduceEpisodeHandler`) and also, they don't get logged when using run_paralle.py. Here I made a separate wandb logger for them so if needed, they can be logged, but they won't be by default.

### DIFF
--- a/src/tbp/monty/frameworks/loggers/wandb_handlers.py
+++ b/src/tbp/monty/frameworks/loggers/wandb_handlers.py
@@ -146,8 +146,8 @@ class BasicWandbTableStatsHandler(WandbHandler):
 class DetailedWandbTableStatsHandler(BasicWandbTableStatsHandler):
     """Log LM stats and actions to wandb as tables.
 
-    This is a modified version of BasicWandbTableStatsHandler that, in additiob to the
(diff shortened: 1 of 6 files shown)


src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -143,6 +143,25 @@ def report_episode(self, data, output_dir, episode, mode="train", **kwargs):
         wandb.log({stats_table: table}, commit=False)
         self.report_count += 1
 
+class DetailedWandbTableStatsHandler(BasicWandbTableStatsHandler):
+    """Log LM stats and actions to wandb as tables.
+
+    This is a modified version of BasicWandbTableStatsHandler that, in additiob to the
```
nielsleadholm, line 149: additiob --> addition

stats logs --> stats, logs

nielsleadholm (approved):
Nice, looks like a useful change! Just some suggestions for typing/punctuation.

PR #188, opened 2025-02-25
docs: additional context for evidence_update_threshold

tristanls:
@vkakerbeck gave this explanation for `evidence_update_threshold` in a video. This explanation is what clicked for me and apparently @scottcanoe. This pull request adds it for context for others who try to understand the purpose of this parameter.

### DIFF
--- a/src/tbp/monty/frameworks/models/evidence_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching.py
@@ -176,13 +176,13 @@ class EvidenceGraphLM(GraphLM):
         initial_possible_poses: initial possible poses that should be tested for.
             In ["uniform", "informed", list]. default = "informed".
         evidence_update_threshold: How to decide which hypotheses should be updated.
-            When this parameter is either '[int]%' or 'x_percent_threshold`, then
-            this parameter is applied to the Most Likely Hypothesis (MLH) to
-            determine a minimum evidence threshold in order for other hypotheses to be
-            updated. Any hypotheses falling below the resulting evidence threshold do
-            not get updated. The other options set a fixed threshold that does not
-            take MLH into account.In [int, float, '[int]%', 'mean', 'median', 'all',
-            'x_percent_threshold'].
+            When this parameter is either '[int]%' or 'x_percent_threshold', then
+            this parameter is applied to the evidence for the Most Likely Hypothesis
+            (MLH) to determine a minimum evidence threshold in order for other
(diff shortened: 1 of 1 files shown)


vkakerbeck (commented):
Thanks for adding the clarification! Just one note: Only the '[int]%' and 'x_percent_threshold' options look at the evidence of the MLH. The other options would set a fixed threshold that doesn't take the evidence of the MLH into account. (or a varying threshold in the case of 'mean' and 'median' which take all evidences for all hypotheses into account)

tristanls:
Thank you @vkakerbeck, updated.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -176,7 +176,13 @@ class EvidenceGraphLM(GraphLM):
...
             In ["uniform", "informed", list]. default = "informed".
         evidence_update_threshold: How to decide which hypotheses should be updated.
-            In [int, float, '[int]%', 'mean', 'median', 'all', 'x_percent_threshold'].
+            When this parameter is either '[int]%' or 'x_percent_threshold`, then
+            this parameter is applied to the Most Likely Hypothesis (MLH) to
+            determine a minimum evidence threshold in order for other hypotheses to be
+            updated. Any hypotheses falling below the resulting evidence threshold do
+            not get updated. The other options set a fixed threshold that does not
+            take MLH into account.In [int, float, '[int]%', 'mean', 'median', 'all',
+            'x_percent_threshold'].
```
vkakerbeck, line 185:
```suggestion
            When this parameter is either '[int]%' or 'x_percent_threshold`, then
            this parameter is applied to the evidence for the Most Likely Hypothesis (MLH)
            to determine a minimum evidence threshold in order for other hypotheses to be
            updated. Any hypotheses falling below the resulting evidence threshold do
            not get updated. The other options set a fixed threshold that does not
            take MLH evidence into account. In [int, float, '[int]%', 'mean', 'median', 'all',
            'x_percent_threshold'].
```

vkakerbeck (approved):
Looks good, thank you! I just added a suggested edit to make it clear that we look at the evidence for the mlh, not the mlh itself (object ID and pose)

src/tbp/monty/frameworks/models/evidence_matching.py
tristanls: Thank you, I added this as well 👍.

Issue #86, opened 2024-11-30
Test failure: worker crashes during setup


crowogenesis:
### Describe the bug

Hello,

I was following the installation guide on 
https://thousandbrainsproject.readme.io/docs/getting-started
, and for some reason all the workers crashed. 
I'm running a very basic Debian 12 environment.

```
maximum crashed workers reached: 8

=================================================================================== FAILURES ====================================================================================
________________________________________________________________________ tests/unit/base_config_test.py _________________________________________________________________________
[gw1] linux -- Python 3.8.20 /home/defaultuser/miniconda3/envs/tbp.monty/bin/python
worker 'gw1' crashed while running 'tests/unit/base_config_test.py::BaseConfigTest::test_can_run_eval_epoch'
________________________________________________________________________ tests/unit/base_config_test.py _________________________________________________________________________
[gw0] linux -- Python 3.8.20 /home/defaultuser/miniconda3/envs/tbp.monty/bin/python
worker 'gw0' crashed while running 'tests/unit/base_config_test.py::BaseConfigTest::test_can_run_episode'
________________________________________________________________________ tests/unit/base_config_test.py _________________________________________________________________________
[gw3] linux -- Python 3.8.20 /home/defaultuser/miniconda3/envs/tbp.monty/bin/python
worker 'gw3' crashed while running 'tests/unit/base_config_test.py::BaseConfigTest::test_can_save_and_load'
________________________________________________________________________ tests/unit/evidence_lm_test.py _________________________________________________________________________
[gw2] linux -- Python 3.8.20 /home/defaultuser/miniconda3/envs/tbp.monty/bin/python
worker 'gw2' crashed while running 'tests/unit/evidence_lm_test.py::EvidenceLMTest::test_evidence_time_out'
_______________________________________________________________________ tests/unit/graph_learning_test.py _______________________________________________________________________
[gw4] linux -- Python 3.8.20 /home/defaultuser/miniconda3/envs/tbp.monty/bin/python
worker 'gw4' crashed while running 'tests/unit/graph_learning_test.py::GraphLearningTest::test_reproduce_multiple_episodes'
___________________________________________________________________________ tests/unit/policy_test.py ___________________________________________________________________________
[gw5] linux -- Python 3.8.20 /home/defaultuser/miniconda3/envs/tbp.monty/bin/python
worker 'gw5' crashed while running 'tests/unit/policy_test.py::PolicyTest::test_can_run_curv_informed_policy'
[...]

PR #191, opened 2025-02-25
Upgrade habitat-sim to 0.3.1 and Python to 3.9

codingkoi:
Newer versions of the habitat-sim library require Python 3.9 so this upgrades to the newest version that doesn't seem to break everything.

Python 3.8 is past end of life, so this also gets us on a supported version.

### DIFF
--- a/tests/unit/graph_learning_test.py
+++ b/tests/unit/graph_learning_test.py
@@ -659,6 +659,19 @@ class GraphLearningTest(BaseGraphTestCases.BaseGraphTest):
         self.exp.evaluate()
         self.exp.dataset.close()
 
+    def test_expected_observations(self):
+        pprint("...parsing experiment...")
+        config = copy.deepcopy(self.disp_config)
+        self.exp = MontyObjectRecognitionExperiment()
+        self.exp.setup_experiment(config)
+        observations = self.exp.dataset.env._env.get_observations()
+        pprint(observations)
+        assert len(observations) == 1
+        # TODO - this is going to fail but the dict I'm comparing to is large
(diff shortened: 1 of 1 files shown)


codingkoi:
This is going to need additional work, as at least two of the tests are failing when I run them locally, but I lack the knowledge to know how they should be fixed or if they're serious breaks.

These are the results I get locally:
```
FAILED tests/unit/evidence_lm_test.py::EvidenceLMTest::test_symmetry_recognition - AssertionError: Lists differ: [-180.0, 0.0, 180.0] != [180.0, 0.0, 180.0]
FAILED tests/unit/evidence_lm_test.py::EvidenceLMTest::test_two_lm_heterarchy_experiment - AssertionError: 'no_match' != 'correct'
```

codingkoi:
For context, this is an issue that @tristanls suggested as a place to start helping out.

tristanls:
Thank you for opening the pull request @codingkoi. An upgrade of sim and Python will be a helpful addition.

I noticed that the test workers crashed. Crashed test workers indicate that likely it is HabitatSim crashing. You will find some troubleshooting hints here [Test failure: worker crashes during setup · Issue #86 · thousandbrainsproject/tbp.monty · GitHub](https://github.com/thousandbrainsproject/tbp.monty/issues/86#issuecomment-2508982655)

codingkoi:
Per discussions with @tristanls, I added a small test to see if the observations in the HabitatSim environment match when run in GH actions vs my local Linux machine, since I'm getting vastly different test results between the two test environments.

jeremyshoemaker:
Closing in favor of #228

PR #194, opened 2025-02-27
build: set correct working-directory for version_updated action

tristanls:
I forgot to use `working-directory` in the `version_updated` action, [resulting in job failures](https://github.com/thousandbrainsproject/tbp.monty/actions/runs/13572685337/job/37941404176#step:3:142).


codeallthethingz (approved):
👍

PR #196, opened 2025-02-27
RFC 9 Hypotheses Resampling

ramyamounir:
This RFC is for resampling hypotheses inspired by particle filters.

View here: https://github.com/ramyamounir/tbp.monty/blob/rfc_hypotheses_resampling/rfcs/0000_hypotheses_resampling.md


rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
+- Start Date: 2025-02-27
+- RFC PR: (leave this empty, it will be filled in after RFC is merged)
+
+# Summary
+
+Resample hypotheses at every step in a manner inspired by particle-filters. This is the first step for Monty to interact with multiple objects and recognize compositional objects.
```
codeallthethingz, line 6: A link to some background reading on particle filters would be great.

ramyamounir: Ah right. @vkakerbeck shared this video a week ago. I thought it was really helpful.

https://www.youtube.com/watch?v=NrzmH_yerBU

I can add it to the RFC if needed.

ramyamounir: Also maybe check out this (more specific to Monty): https://thousandbrainsproject.readme.io/docs/test-particle-filter-like-resampling-of-hypothesis-space

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+
+# High-Level Motivation
+
+In an unsupervised learning experiment setup, Monty may be presented with multiple objects in a single episode. Ideally, we would like to move away from the traditional data loading setup of machine learning where there is a strict definition of an epoch, episode and step. As Monty starts to interact with the real world, the definition of epoch and episode will start to fade away and we'll be left with simple time discretization (i.e., step). The current definitions are:
+* Epoch: Used by the experiment class to denote one full pass through all the objects
```
nielsleadholm, line 11: *through all the objects at a specified rotation

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+* Episode: Denotes a change in object
+* Step:  Denotes a single sensation and action in the sensorimotor framework.
+
+Real world interactions do not have epochs or episodes (these are only used for performance benchmarks), instead we could imagine the agent wondering around in a multi-object dynamic environment. The objects can be occluded, moving, or even disappearing behind new objects.
```
nielsleadholm, line 15: *wondering --> wandering

Maybe add here one of the key motivations for this work - "Many objects are compositional, such as a logo on a coffee mug, where an agent can move between these."

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+* Epoch: Used by the experiment class to denote one full pass through all the objects
+* Episode: Denotes a change in object
+* Step:  Denotes a single sensation and action in the sensorimotor framework.
+
+Real world interactions do not have epochs or episodes (these are only used for performance benchmarks), instead we could imagine the agent wondering around in a multi-object dynamic environment. The objects can be occluded, moving, or even disappearing behind new objects. 
+
+**We want Monty to handle dynamic environments by seamlessly switching from one object to another as it's sensors move around on the different objects.**
+
+# The Problem
+Monty is designed to receive a weak supervision signal when the episode ends and a new episode begins (change of object). This signal performs a full reset of all states within Monty. This reset includes counters, buffer, goal state generators, learning modules and sensory modules. Additionally, this reset sets Monty back into Matching mode. The below figure shows where this resetting is done. Most resetting happens in the `pre_episode` functions of the Monty and SMs and LMs classes.
```
nielsleadholm, line 20: *a weak supervision signal --> a weak supervision signal at inference time

Clarifying because this weak supervision will still exist at learning after this RFC.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+## Assumptions and constraints:
+* Number of hypotheses should not scale up with steps, if anything they should decrease. For now, any sampled hypothesis must replace an old "unlikely" hypothesis.
+* Sampling of likely and unlikely hypotheses is based on evidence change instead of absolute evidence.
+* Terminal state calculation is still based on accumulated evidence and `x_percent_threshold`.
+* The resampling procedure only occurs if principal curvature is defined.
+
+## The Resampling Procedure
+
+1) For every object, at every new step, get initial hypotheses based on the observed pose.
+2) If principal curvature is not defined, skip
```
nielsleadholm, line 59: Shouldn't (2) happen first to avoid the cost of getting new hypotheses?

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+3) Uniformly sample M percent from these new hypotheses (e.g., M = 20% = 1000 hyp)
```
nielsleadholm, line 60: I would mention further up in the RFC that we would also do some uniform sampling (which I think is reasonable, but I think this is the first time it is mentioned).

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+	* Detecting the object change will likely be based on evidence decay and can be problematic. If an object was added and quickly changed before accumulating enough evidence, it will be difficult to detect change based on evidence decay. This will still result in incorrect initial hypotheses. Resampling without a hard reset is more general.
+
+# Future possibilities
+* This only works in inference assuming that we have good object models learned in isolation. It likely breaks learning from scratch experiments.
+* I manually force Monty back into the matching phase and reset it's counters between episodes. 
+	* Maybe we should brainstorm about what a terminal state should be in a multi-object environment.
+	* Maybe also brainstorm about ways to more seamlessly switch between matching and exploration? I don't like forcing Monty to always be in matching mode, by doing this it feels that we are making the gap between learning and inference wider.
+* Start designing environments with multiple objects aiming towards compositional models.
+	* Will require motor policy change to allow for moving between objects
```
nielsleadholm, line 75: Missing `.`

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+3) Uniformly sample M percent from these new hypotheses (e.g., M = 20% = 1000 hyp)
+4) Sample N percent from the existing hypotheses distribution of highest evidence slope (e.g., N = 10% = 500 hyp). These sampled hypotheses should be "close" to the most likely hypotheses.
+5) Replace unlikely M+N percent of existing hypotheses (e.g., 30% = 1500 hyp) with these newly sampled hypotheses. The unlikely hypotheses are chosen based on evidence change.
+
+# Alternatives Considered:
+* Detect change with evidence decay and hard reset?
+	* Here we would focus on detecting the change that triggers full reset of Monty states. 
+	* Detecting the object change will likely be based on evidence decay and can be problematic. If an object was added and quickly changed before accumulating enough evidence, it will be difficult to detect change based on evidence decay. This will still result in incorrect initial hypotheses. Resampling without a hard reset is more general.
+
+# Future possibilities
```
nielsleadholm, line 69: One thing I've been thinking about in terms of the artificial nature of this task (which is also something we've discussed before), is this current setup is a bit like a "magic trick" where someone puts their hand in a box to feel an object with one finger, then after a couple minutes, someone discretely replaces the object without their finger really feeling a major change. In other words, it's not very naturalistic. What we are actually trying to build towards is being able to seamlessly move from one sub-object to another if we have e.g. a logo on a mug, or a car composed of parts, and for the LM representation to update without needing a supervisory signal that it's on a new object. In the compositional case however, if the LM was 100% confident about where it was on a learned reference frame (e.g. on the "N" in the child object logo), after performing a movement onto the parent object (e.g. mug), it's hypothesized location in the child object's reference frame would now be in empty space, far from any learned features that exist for that object. In that case, the evidence associated with this "off-object" location should essentially be zero by definition, and it should be straightforward to implement this.

In the above case, resampling would still be important, but I think it just emphasizes that the evidence scores could be a lot less adversarial than in the current experiment setup. 

For what it's worth in the above situation, if you moved back onto the object, I think it would be the responsibility of a higher level LM that remembers e.g. logo at location x to bias the low-level LM, rather than requiring the low-level LM maintain a memory of the previous object. This would tie in with resampling, i.e. we wouldn't need to worry about keeping lots of sampling points for the old object if we've moved off of it. 

@vkakerbeck what do you think? It almost makes me wonder whether we would be better off having the compositional dataset to evaluate these approaches on, rather than trying to shoehorn it into the current YCB object cycling setup. Maybe something we can also discuss on Monday.

nielsleadholm (commented):
Nice, thanks for putting this together! I like the idea of using the slope as a signal for sampling, i.e. even when the total evidence counts might be low. 

Just a couple comments, e.g. I think it would be worth also emphasizing the uniform sampling, because I think this will still be useful, otherwise we might not see a good slope for the new object in the first case? At least based on the second plot, it looks like it takes a while for the slope to appear.

I've also just written a high-level comment about the dataset we want to evaluate this work on, as this might be something we want to revisit.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+
+![Monty Reset Logic](0000_hypotheses_resampling/monty_reset_logic.png)
+
+If we simply disable this resetting signal for the Monty class (and by extension SMs and LMs) between episodes, there will not be enough evidence update in the first step of a new object to get out of a terminal state. Monty will still think it is seeing the old object after getting a single observation of the new object. See the plot below.
+
+![Single Step Terminal State](0000_hypotheses_resampling/terminal_state_reached.png)
+
+To overcome this, I manually `reset_episode_steps()` such that the `matching_steps` would still be under the `min_eval_steps` and allow Monty time to gather enough evidence on the new object. Additionally, I manually `switch_to_matching_step()` between episodes since the `_compute_possible_matches` function that accumulates evidence is only called during matching, not exploration. This results in the following plot.
```
vkakerbeck, line 28: A more long-term solution would be to update the `check_terminal_conditions` function. Basically, if we start running experiments where the agent can move around from object to object without the environment being reset, we want to redefine what a terminal condition is for an episode. I could imagine in the future we have a totally different terminal condition (like achieving a goal state in the world or just a simple time limit). Then we log mlh at every step and compare it to whatever is sensed at that step in the post-experiment analysis. We could use a similar logic as we currently use in the terminal condition check to switch between matching and exploratory policies. This is a larger rethinking of the experimental setup so worth it's own RFC when we get there (as part of the "Setup YCB testbed that removes 1 episode-per-object assumption" task).

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+
+![No Resampling](0000_hypotheses_resampling/no_resampling.png)
+
+This reveals the main problem. Monty is still unable to accumulate evidence on the existing hypotheses. The current implementation of Monty uses `_get_all_informed_possible_poses()` to initialize hypotheses after seeing a single pose of the object. This is a smart way to reduce the number of initial hypotheses based on the principal curvature but it assumes that the object doesn't change and that these hypotheses will always be valid. However, when we change the object we would need to update these initial hypotheses based on a new pose observation of the new object. A simple test of sampling additional hypotheses (with informed poses) on the second object pose shows that we are able to accumulate evidence on these new hypotheses. See figure below.
```
vkakerbeck, line 32: You could also mention that even when there is just one object, it introduces an overreliance on the first observation which can lead to problems if that observation is noisy.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+
+![Resampling](0000_hypotheses_resampling/resampling_banana_mug.png)
```
vkakerbeck, line 34: Not something that needs to be done for this RFC, but something that always throws me off with these plots is that the legend (order and colors) changes between plots. Could you somehow fix this in your plotting code?

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+
+Why:
+* **Faster**: we don't have to wait for high unbounded evidence to decay enough to realize that a new hypothesis is more likely. We also may not need to worry about initializing new hypotheses with mean evidence, giving them fighting chance against other old hypotheses. Average slope is more fair in this sense.
+* **Accurate resampling**:  If we sample new hypotheses close to the hypotheses with high total accumulated evidence (e.g., particle filter), we could be sampling from the incorrect hypotheses (if we had just switched objects). If we sample close to the hypotheses with high evidence slope, we may converge faster.
+* **Practical**: The total evidence can still be unbounded, it doesn't matter because we only consider the slope. This metric does not care about how much evidence we've accumulated already. It other words, a hypothesis with a high evidence, can be removed if it hasn't accumulated evidence in a while, while a consistently growing hypothesis is less likely to be removed even if it was just added.
```
vkakerbeck, line 48: Are we still using the total evidence to determine the MLH? And then just using the slope for determining which hypotheses to resample? I think this would make sense because using the slope for MLH as well would come with the same problems as the bounded evidence approach (not having a long enough memory to explore a larger part of the object). I don't have a good intuition of whether this problem would also come up for this metric when it is used to determine which hypotheses to resample but I agree that it seems more reasonable than the total evidence.

vkakerbeck: Oh I should have kept reading before writing this :D

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Assumptions and constraints:
+* Number of hypotheses should not scale up with steps, if anything they should decrease. For now, any sampled hypothesis must replace an old "unlikely" hypothesis.
+* Sampling of likely and unlikely hypotheses is based on evidence change instead of absolute evidence.
+* Terminal state calculation is still based on accumulated evidence and `x_percent_threshold`.
+* The resampling procedure only occurs if principal curvature is defined.
```
vkakerbeck, line 54: This could be a pretty limiting constraint. Consider an object like ball where PC is not well defined anywhere on the object. Why not still use the point normal to inform new hypotheses? If you are worries about adding too many new hypotheses we could set the number of sampled hypotheses in `get_more_directions_in_plane` lower.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+
+## The Resampling Procedure
+
+1) For every object, at every new step, get initial hypotheses based on the observed pose.
+2) If principal curvature is not defined, skip
+3) Uniformly sample M percent from these new hypotheses (e.g., M = 20% = 1000 hyp)
+4) Sample N percent from the existing hypotheses distribution of highest evidence slope (e.g., N = 10% = 500 hyp). These sampled hypotheses should be "close" to the most likely hypotheses.
```
vkakerbeck, line 61: Could you give some more details on this step? Would you take the to 10% hypotheses and then slightly vary their location and orientation?

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+5) Replace unlikely M+N percent of existing hypotheses (e.g., 30% = 1500 hyp) with these newly sampled hypotheses. The unlikely hypotheses are chosen based on evidence change.
```
vkakerbeck, line 62: How is the evidence count for these new hypotheses initialized?
On a technical level, we have this large matrix of hypotheses, do you basically in-place replace the low evidence one's with the new ones and thereby don't change the shape of the hypothesis space?

rfcs/0000_hypotheses_resampling.md
vkakerbeck: Good points. I think this may relate to the mechanism for switching between matching and exploration policies we discussed briefly:
- If we have recognized an object and are now exploring it, we don't need to keep testing hypotheses, we just make predictions using our model of this object and check if these predictions are correct. 
- We would exit this mode if 1) our prediction is suddenly incorrect (the artificial setup where you suddenly replace the object, or you were wrong about the object) or 2) the sensor moves to a location where the model wouldn't predict the object to exist anymore (moving off object) which is the usual, more natural case. 
- The second case would give us a model-based signal for switching back into matching mode and potentially reinitializing the entire hypothesis space.
- In my mind we are currently dealing with the case where we have actually not recognized the object yet so we can't make model-based predictions. I mean you could argue that we can make model-based predictions using the mlh (which we do when testing that hypothesis) but it is not as much of a reliable signal that it could be used for a hard reset.

In terms of the test bed we use, I think that we can keep using the current setup for the hypothesis resampling work since this mechanism doesn't rely on physical movement from one object onto another. It should actually be useful for recognizing a single object in general (removing reliance on first observation and allowing us to test less hypotheses at any point in time).

However, it does bring up a good point on whether this item is really the most impactful one for us to tackle for the unsupervised learning milestone. We could evaluate whether it makes more sense to start with the prediction error/off-object prediction signal I outlined. In that case we would definitely need a multi object environment setup since this relies on a physical separation between the two objects.

Lmk what you think. Happy to talk about this more in a meeting too.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Alternatives Considered:
+* Detect change with evidence decay and hard reset?
+	* Here we would focus on detecting the change that triggers full reset of Monty states. 
+	* Detecting the object change will likely be based on evidence decay and can be problematic. If an object was added and quickly changed before accumulating enough evidence, it will be difficult to detect change based on evidence decay. This will still result in incorrect initial hypotheses. Resampling without a hard reset is more general.
+
+# Future possibilities
+* This only works in inference assuming that we have good object models learned in isolation. It likely breaks learning from scratch experiments.
```
vkakerbeck, line 70: Why would you think it would break learning from scratch? Do you mean because we would never have all hypotheses go below 0 evidence and can therefor never classify no_match? I would be curious about the details of how you plan to initialize the evidence values for the new hypotheses. I think we should try to make this mechanism work in the learning setting as well. Maybe by initializing the average evidence, which could be negative? I'd have to think about this a bit more.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+* I manually force Monty back into the matching phase and reset it's counters between episodes. 
+	* Maybe we should brainstorm about what a terminal state should be in a multi-object environment.
+	* Maybe also brainstorm about ways to more seamlessly switch between matching and exploration? I don't like forcing Monty to always be in matching mode, by doing this it feels that we are making the gap between learning and inference wider.
```
vkakerbeck, line 73: Definitely two good topics to talk about more. I wrote my thoughts on both in earlier comments. lmk what you think.

rfcs/0000_hypotheses_resampling.md
vkakerbeck, line 69: I think another future work item would be to test if we can use this mechanism to make Monty more efficient. Basically we don't have to initialize 1000s of hypotheses on the first step. We can initialize a small subset and then iteratively refine these hypotheses. This was one of my initial excitements about this idea.

vkakerbeck (commented):
Nice outline and great figures! Overall really like it. I just asked about clarifications in a few places. On a major note, we should talk more about the longer thread at the future possibilities section (is this really the most impactful next step). This shouldn't keep this RFC from being merged but may determine whether it's actually the next thing to work on.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
+- Start Date: 2025-02-27
+- RFC PR: (leave this empty, it will be filled in after RFC is merged)
+
+# Summary
+
+Resample hypotheses at every step in a manner inspired by particle-filters. This is the first step for Monty to interact with multiple objects and recognize compositional objects.
+
+# High-Level Motivation
+
+In an unsupervised learning experiment setup, Monty may be presented with multiple objects in a single episode. Ideally, we would like to move away from the traditional data loading setup of machine learning where there is a strict definition of an epoch, episode and step. As Monty starts to interact with the real world, the definition of epoch and episode will start to fade away and we'll be left with simple time discretization (i.e., step). The current definitions are:
```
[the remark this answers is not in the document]
ramyamounir: Yes, I agree with you that "unsupervised experiment" seems more fitting here. I like how you phrased this paragraph and added it as-is.

rfcs/0000_hypotheses_resampling.md
ramyamounir: Yes, good catch. Fixed.

rfcs/0000_hypotheses_resampling.md
ramyamounir: I've modified the summary section. Let me know if it's not clear enough.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,75 @@
...
+* Epoch: Used by the experiment class to denote one full pass through all the objects
+* Episode: Denotes a change in object
```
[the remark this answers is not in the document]
ramyamounir: I see. Thanks for explaining this.

rfcs/0000_hypotheses_resampling.md
ramyamounir: Yes, it's been throwing me off too :sweat_smile:. I'll fix it for future plots.

rfcs/0000_hypotheses_resampling.md
ramyamounir: I'm just worried that we would be replacing more accurate hypotheses with less accurate ones, **especially if the object hasn't changed**. For example, going from an area of defined PC to undefined PC. The hypotheses generated without a principal curvature are many and very coarse (9 per axis). It may not be too much of a concern because we only replace the unlikely hypotheses, I'm just worried that resampling every step will eventually run out of unlikely hypotheses and start replacing some good hypotheses too. What do you think?

You raise a good point on the ball object. Maybe add fewer "semi-informed" hypotheses when PC is undefined?

rfcs/0000_hypotheses_resampling.md
ramyamounir: I think I am more concerned about environments with multiple objects during the learning phase. If we assume objects are learned in isolation, then these pretrained graphs will work fine during inference. But in the learning from scratch, don't we use the same environment for learning and inference? If we start a multiobject episode, wouldn't we have to use it during learning too? Then objects won't be learned in isolation and the graphs we use for matching might end up being graphs of multiple objects. I could be missing something here..

rfcs/0000_hypotheses_resampling.md
ramyamounir: Yes, this makes sense. The terminal condition will need to be redefined based on the experiment type.

rfcs/0000_hypotheses_resampling.md
ramyamounir: Yeah good point! I find it interesting that you said start with less hypotheses and add more, because that's the opposite of what they explained [here](https://youtu.be/NrzmH_yerBU?si=zbfz5N44zZtzu0wz&t=766) on particle filters. They started with many particles and reduced the number as the model started narrowing down the location (i.e., simpler distribution can be represented with less particles). I can say good arguments for both ways actually.

I know you mentioned this as part of future work, but I think it makes sense to accommodate for it here. I've updated the RFC to allow for changing count of hypotheses. So if we want to start with a small number and increase the hypotheses by 10% every step, we can define `hypotheses_count_ratio=1.1`. We can also control where these hypotheses come from using the `hypotheses_old_to_new_ratio` and `hypotheses_informed_to_reenforce_ratio`.

ramyamounir: Upon reading your comment again, I see now that you said "iteratively refine" not increase :sweat_smile: Yeah makes sense!

nielsleadholm: Yeah I think my concern with optimizing for this current problem setup is that we may end up with solutions that aren't actually that necessary / useful. E.g. clearly the hypothesis resampling is useful, but the exact details of it might be lead on a tangent by trying to deal with this adversarial (and artificial) problem setup.

- One option is just use this as you say, but we should be cognizant of this caveat and not worry about getting super high accuracy in this setting, and just consider it as a test-bed for resampling.
- The other is the option of, after this RFC, to refocus on the actual dataset we want to deal with. A lot of unsupervised **_learning_** features aren't actually important if we are learning a dataset where compositional objects can be learned in isolation, in which case we can potentially refocus the items under the first milestone.

ramyamounir: I agree with @nielsleadholm that this setup is not very naturalistic and might lead to some unnecessary solutions because we're optimizing for a much harder experiment. I like the option where we consider this a test-bed for resampling and not worry about super high accuracy. In practice, Monty should be able to always accumulate evidence even when we switch objects in place, but we don't need to spend too much time optimizing the speed of inference between objects.

I also like @vkakerbeck  distinction between model-based and model-free evidence accumulation. If we have not recognized a model, we can't really use a model to make good predictions. This is where hypotheses resampling is most useful - to seamlessly adapt to changes while still trying to recognize the object. But if we have recognized the model, we can make predictions and figure out when a "Hard" reset might be needed. **This would require that we continuously test the recognized model at every exploratory step.** My only concern here is how do we know that a prediction error due to an unseen part of the object (e.g., previously unseen handle on a mug) should be treated as part of the object (i.e., add it to the cylinder graph) vs. trigger a prediction error reset for recognizing a new object that happens to be spatially close? Is this problem specific to learning-from-scratch?

I am a bit biased towards keeping this setup until we implement a decent hypotheses resampling LM because multiple objects would require us to think about (hierarchical) action policies to move between objects. 
* If it's a compositional object, such as wheels on a car, we would use a hypotheses-based hierarchical action policy. For example, if I know the higher level LM is on the car and the lower level LM is on the front wheel, now the higher level LM can instruct the lower level LM to move to the back wheel and reset its hypotheses (and perhaps even bias the new hypotheses with extra evidence for the back wheel object as Niels mentioned)
* If we are not on a compositional object, we need to think about model-free action policies to move from one object to another. Maybe we can hard-code that in for now as a hack (i.e., after x steps move the sensor to object 2 at a known location)?

@nielsleadholm when you say compositional dataset, are you talking about the dinner set scenes?

nielsleadholm: Great, Viviane and I were discussing this earlier this morning and yes what we are thinking is:
- Clarify that the current work to enable a first version of compositional objects is about unsupervised *inference*, not learning. For now, we assume that we are going to learn sub-objects in isolation (e.g. mug and logo, see below), and so it's more about equipping Monty to handle inference where it may move from one object onto another. 
- Continue the current work you are doing. Adding the ability to resample hypotheses will be important for several aspects of compositional objects. This first version would actually be best tested by seeing whether we can get more robust inference with the same number of initial points, or equally achieve the same accuracy by starting with fewer initial points. We can still use this artificial object-swapping task as an adversarial condition to magnify the benefits of resampling, but we don't need to optimize parameters to achieve high performance on it. 
- Later, create an RFC proposing how we can use more model-based resetting based on moving off of an object. I agree this will be an interesting question to explore in terms of untangling prediction errors that should inform learning vs. hypothesis resetting. E.g. maybe the default is that the hypothesis space is reset, but with some evidence accumulating that the model might be incomplete. Given enough movements with a consistent prediction error, this might trigger a more learning-style phase. But there are clearly complications to work through. 
- The compositional dataset would initially be a [simple one consisting of different objects with different logos on their surfaces](https://thousandbrainsproject.readme.io/docs/make-dataset-to-test-compositional-objects). That way we can learn the sub-components in isolation. This ties in with simplifying the first part of research to focus on unsupervised *inference*, and in the future we can return to unsupervised *learning* to enable more complex compositional datasets.

nielsleadholm (commented):
Thanks for the great updates! I like the proposed changes to the code, as well as the parameters you've introduced. Just added a few comments.

rfcs/0000_hypotheses_resampling.md
ramyamounir: Yeah makes sense. For what it's worth, I won't be directly controlling the number of different rotations to be sampled based on PC defined/undefined. Instead, I will be uniformly sampling a fixed number of informed hypotheses based on the [parameters introduced](https://github.com/ramyamounir/tbp.monty/blob/bd0665c84781030340692692038d3c1032679187/rfcs/0000_hypotheses_resampling.md?plain=1#L84) in this RFC. For example, if I need to get 100 informed hypotheses, I will uniformly sample them from 2000 hypotheses (with PC defined) or 8000 hypotheses (with PC undefined).

My concern here was mainly for scenarios where the sensor moves on the side of the coffee mug for a while (i.e., PC defined) and then moves on the bottom side (i.e., PC undefined). Sampling from a bigger pool of inaccurate hypotheses to replace more accurate hypotheses that were sampled when PC was well defined may cause some problems. That being said, it may turn out to be empirically insignificant.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -45,21 +52,104 @@ We currently use the total evidence score to decide which hypotheses are more pr
...
 * The resampling procedure only occurs if principal curvature is defined.
 
 ## The Resampling Procedure
 
-1) For every object, at every new step, get initial hypotheses based on the observed pose.
-2) If principal curvature is not defined, skip
-3) Uniformly sample M percent from these new hypotheses (e.g., M = 20% = 1000 hyp)
-4) Sample N percent from the existing hypotheses distribution of highest evidence slope (e.g., N = 10% = 500 hyp). These sampled hypotheses should be "close" to the most likely hypotheses.
-5) Replace unlikely M+N percent of existing hypotheses (e.g., 30% = 1500 hyp) with these newly sampled hypotheses. The unlikely hypotheses are chosen based on evidence change.
+1) If principal curvature is not defined, skip
```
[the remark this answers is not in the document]
ramyamounir: Sounds good, I've removed this from the resampling procedure and the constraints.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -45,21 +52,104 @@ We currently use the total evidence score to decide which hypotheses are more pr
...
+This should take care of dynamically increasing or decreasing the hypotheses count.
+For example, if we want to start by coarse sampling of hypotheses and then increase the number of hypotheses (and vice versa), we should be able to do that here.
+
+I'm introducing three new parameters. The naming of these parameters is preliminary and subject to change.
+
+| Parameter | Description | Range |
+| ----------|-------------|-------|
+| **hypotheses_count_ratio** | A multiplier for the needed number of hypotheses at this new step | [0, inf) |
+| **hypotheses_old_to_new_ratio** | How many old to new hypotheses to be added. `0` means all old, `1` means all new | [0, 1] |
+| **hypotheses_informed_to_reinforce_ratio** | How many informed (sampled based on newly observed pose) to reinforced hypotheses (sampled close to existing likely hypotheses) to be added. `0` means all informed, `1` means all reinforced | [0, 1] |
```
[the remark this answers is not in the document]
ramyamounir: yeah, good idea. I've added them to the summary. Also I came up with "reinforced" and I'm not sure if it's the best term, but we can go with it for now if no objections.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -45,21 +52,104 @@ We currently use the total evidence score to decide which hypotheses are more pr
...
+# 0 means all sampled from old, 1 means all sampled from new
+hypotheses_old_to_new_ratio = 0.5
+
+
+# 0 means all sampled from informed hypotheses, 1 means all sampled from reinforced hypotheses
+hypotheses_informed_to_reinforce_ratio = 0.5
+
+
+def calculate_new_hypotheses_counts(
+    curr_hypotheses_count, new_informed_hypotheses_count
```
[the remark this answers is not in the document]
ramyamounir: These are the total number of informed hypotheses sampled based on the current pose observation. Assuming there are `x` points in the object graph, these can be `2x` if the PC is defined or `8x` if PC is undefined. When calculating the needed number of informed hypotheses (i.e., `needed_new_informed_hyp`), we should not request more than the available `new_informed_hypotheses_count`. If we request more, the difference will come from reinforced hypotheses which are technically infinite. 

*Note that we do not need to actually get the informed hypotheses at this step to know how many we have, we can just use the number of points in the graph and whether the pose is defined.*

I'll add some description.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -45,21 +52,104 @@ We currently use the total evidence score to decide which hypotheses are more pr
...
+3) Calculate the needed hypotheses counts to be sampled based on the defined parameters as shown [here](#The-Resampling-Count-Calculation).
+4) Sample `needed_old_sampled_hyp` from the existing hypotheses based on highest evidence slope. We will keep these old hypotheses and remove the rest.
+5) Uniformly sample `needed_new_informed_hyp` from the new informed hypotheses. We will add these new hypotheses.
+6) Sample `needed_new_reinforced_hyp` from the existing hypotheses distribution of highest evidence slope. These sampled hypotheses should be "close" to the most likely hypotheses.
+
+
+## High Level Code Changes
+
+The needed modification will only change the `EvidenceLM` class. More specifically, we would be either be modifying the `_update_evidence` function
+directly or defining another function that calls `_update_evidence` as one of its steps. A rough proposal of the needed changes is shown below.
```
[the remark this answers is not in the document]
ramyamounir: Yes, makes sense. I may end up rewriting and factoring out parts of `_update_evidence` in the process. I was thinking of just rewriting `_update_evidence` to support resampling and breaking it up into manageable smaller functions.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -45,21 +52,104 @@ We currently use the total evidence score to decide which hypotheses are more pr
...
+
```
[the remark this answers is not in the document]
ramyamounir: Good point. The main purpose of the proposed figure is to show roughly where the changes will occur within the `matching_step` but I imagine that the exact number of functions will change as I start implementing. I may even do all the changes in `update_evidence` as discussed above. But at least this gives you an idea of where I plan to start.

rfcs/0000_hypotheses_resampling.md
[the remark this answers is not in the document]
nielsleadholm: Nice, yeah that would be great. It's a pretty big method at the moment.

nielsleadholm (approved):
Nice, thanks for making those changes! Just a couple more very minor comments but otherwise it looks ready to merge from my perspective.

rfcs/0000_hypotheses_resampling.md
vkakerbeck: There is nothing that forces us to use the same environment setup during training as during evaluation. We could first show objects in isolation while not providing labels (making it a learning from scratch setup) and then show them in multi object setups. I'm not saying this would work great, even in the current unsupervised learning setup with isolated objects we often merge multiple objects into the same graph. Just saying that not providing labels doesn't keep us from showing objects in isolation initially.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -45,21 +54,125 @@ We currently use the total evidence score to decide which hypotheses are more pr
...
+## High Level Code Changes
+
+The needed modification will only change the `EvidenceLM` class. More specifically, we would modify either the `_update_evidence` function
+directly or define another function that calls `_update_evidence` as one of its steps. A rough proposal of the needed changes is shown below.
+
+![code change](0000_hypotheses_resampling/high_level_code_change.png)
+
+*Note that the two types (reinforced and informed) of resampled hypotheses are treated differently. 
+Unlike reinforced hypotheses, the current positions of the informed hypotheses do not need to be rotated and displaced at resampling.
+Informed hypotheses are added after the reinforced (and old) hypotheses are updated/displaced.*
```
vkakerbeck, line 81: You still plan on added feature evidence to the new informed hypotheses, right? (like in `_calculate_feature_evidence_for_all_nodes`)

rfcs/0000_hypotheses_resampling.md
```diff
@@ -45,21 +54,125 @@ We currently use the total evidence score to decide which hypotheses are more pr
...
+
+This is a proposed approach to calculating the new set of hypotheses at every step. 
+This should take care of dynamically increasing or decreasing the hypotheses count.
+For example, if we want to start by coarse sampling of hypotheses and then increase the number of hypotheses (and vice versa), we should be able to do that here.
+
+I'm introducing three new parameters. The naming of these parameters is preliminary and subject to change.
+
+| Parameter | Description | Range |
+| ----------|-------------|-------|
+| **hypotheses_count_ratio** | A multiplier for the needed number of hypotheses at this new step | [0, inf) |
```
vkakerbeck, line 94: Does this mean if it is 1, the number of hypotheses remains the same?

vkakerbeck: Ah I see this below now :) Could be worth adding to the description here too since you have that in the description of the other params as well and it's useful info

rfcs/0000_hypotheses_resampling.md
```diff
@@ -45,21 +54,125 @@ We currently use the total evidence score to decide which hypotheses are more pr
...
+):
+    """
+    This is a proposed function for calculating the needed count of each type of
+    hypotheses ("informed", "reinforced", "old").
+
+    Inputs:
+    - `curr_hypotheses_count`: The current number of hypotheses
+    - `new_informed_hypotheses_count`: The total number of informed hypotheses
+        sampled based on the current pose observation. Assuming there are x points
+        in the object graph, these can be 2x if the PC is defined or 8x if PC is undefined.
```
vkakerbeck, line 126: Related to my earlier comment, I think it would be more efficient if we sampled the informed hypotheses after we calculated the counts so we don't need to calculate possible poses for all points in the graph. You could take the `needed_new_informed_hyp` and sample that number/2 points from the graph (could even be sampled in an informed way by first picking the ones where the features actually match the sensed features. 

This would remove the need for this parameter but would introduce the risk that `needed_new_informed_hyp` is larger than the number of points in the graph * 2. Is this actually something that would realistically happen? In what case would this happen? If `hypotheses_count_ratio` is >1? Do we want to support this case? And if so, could we just duplicate hypotheses with some noise added to them to fill up the number of requested hypotheses?

rfcs/0000_hypotheses_resampling.md
```diff
@@ -45,21 +54,125 @@ We currently use the total evidence score to decide which hypotheses are more pr
...
 
 1) For every object, at every new step, get initial hypotheses based on the observed pose.
-2) If principal curvature is not defined, skip
-3) Uniformly sample M percent from these new hypotheses (e.g., M = 20% = 1000 hyp)
-4) Sample N percent from the existing hypotheses distribution of highest evidence slope (e.g., N = 10% = 500 hyp). These sampled hypotheses should be "close" to the most likely hypotheses.
-5) Replace unlikely M+N percent of existing hypotheses (e.g., 30% = 1500 hyp) with these newly sampled hypotheses. The unlikely hypotheses are chosen based on evidence change.
+2) Calculate the needed hypotheses counts to be sampled based on the defined parameters as shown [here](#The-Resampling-Count-Calculation).
+3) Sample `needed_old_sampled_hyp` from the existing hypotheses based on highest evidence slope. We will keep these old hypotheses and remove the rest.
+4) Uniformly sample `needed_new_informed_hyp` from the new informed hypotheses. We will add these new hypotheses.
+5) Sample `needed_new_reinforced_hyp` from the existing hypotheses distribution of highest evidence slope. These sampled hypotheses should be "close" to the most likely hypotheses.
```
vkakerbeck, line 69: I'm not sure what you are planning exactly for how to initialize the evidence values but I think it would be worth considering using the average of the most likely hypotheses (the ones that the new hypotheses are sampled from) for this here instead of the average evidence for all hypotheses. We can use the average evidence of all hypotheses for step 4 (+evidence based on the sensed feature match)

rfcs/0000_hypotheses_resampling.md
```diff
@@ -3,21 +3,28 @@
 
 # Summary
 
-Resample hypotheses at every step in a manner inspired by particle-filters. This is the first step for Monty to interact with multiple objects and recognize compositional objects.
+Resample hypotheses at every step in a manner inspired by particle-filters. This is the first step for Monty to interact with multiple objects and recognize compositional objects. The newly sampled hypotheses will come from:
+1) **Informed Hypotheses:** a subset (uniformly sampled) of new hypotheses initialized based on the current step observation (i.e., sensed pose).
+2) **Reinforced Hypotheses:** a set of newly sampled hypotheses from the distribution of the most rapidly rising hypotheses.
```
vkakerbeck, line 8: Is "reinforced hypotheses" a known term from particle filters? To me it sounds maybe a bit ambiguous as to what they actually are. If there is a common terminology for this already we should use this. I also find "informed hypotheses" a bit confusing in this context now (since technically all 3 types are informed, just by different things). Initially I used the term to contrast the initialization to uniformly initialized hypotheses. I get how it would be useful to keep using the same terminology so I don't have a good solution. I just noticed reading through the code that I was getting quite confused at time and the variable names don't seem so intuitive. I don't know you @nielsleadholm you have any thoughts or if we should just stick with it and add good documentation.

vkakerbeck (approved):
Really nice, thanks for the updates! I just left one more major comment around the execution order of step 1 and 2. But happy to approve this already so you can get started on implementing this.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -45,21 +54,125 @@ We currently use the total evidence score to decide which hypotheses are more pr
...
 
 ## Assumptions and constraints:
-* Number of hypotheses should not scale up with steps, if anything they should decrease. For now, any sampled hypothesis must replace an old "unlikely" hypothesis.
 * Sampling of likely and unlikely hypotheses is based on evidence change instead of absolute evidence.
 * Terminal state calculation is still based on accumulated evidence and `x_percent_threshold`.
-* The resampling procedure only occurs if principal curvature is defined.
 
 ## The Resampling Procedure
 
 1) For every object, at every new step, get initial hypotheses based on the observed pose.
```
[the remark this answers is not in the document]
ramyamounir: Yes, that makes sense. This is what I planned out in the high-level code change diagram, but I forgot to update this here in the resampling procedure.

rfcs/0000_hypotheses_resampling.md
ramyamounir: Yes. I would need to compare with sensed features and add evidence if needed.

ramyamounir: At this point, I'm probably better off rewriting `_update_evidence` with support for resampling.

rfcs/0000_hypotheses_resampling.md
ramyamounir: Yeah, I'll add that.

rfcs/0000_hypotheses_resampling.md
ramyamounir: Yes, that's a good point. I don't plan on sampling the informed hypotheses first, but I only need the number of expected informed hypotheses at this step. We can quickly figure that out from the number of graph points and whether PC is defined. Then, I can know if the `needed_new_informed_hyp` is more than the actual informed hypotheses and handle it. I will only sample what I need from the graph later.

Yes, we run this risk if `hypotheses_count_ratio` is > 1. Another case is if we started off with undefined PC at step 1 (graph points * 8) and moved to defined PC at step 2 (graph points * 2). Even if `hypotheses_count_ratio` is < 1, we could still request too many informed hypotheses to maintain the high number of hypotheses we started with. A bit of an edge case, but could happen.

rfcs/0000_hypotheses_resampling.md
ramyamounir: Yes, great idea!

rfcs/0000_hypotheses_resampling.md
ramyamounir: It is not a known term, I just needed a place holder and this is the best I could come up with :sweat_smile: 

My intuition is we are reinforcing existing hypotheses that are most likely correct by adding ones that are "close" to them. Open to suggestions though..

rfcs/0000_hypotheses_resampling.md
ramyamounir: I see. I had the wrong assumptions about the learning from scratch setup then.

scottcanoe (approved):
Wow, great write-up and an interesting read. It looks like there are quite a few ideas in here to try out, and I'm looking forward to seeing how it pans out.

I don't have much detailed feedback since my knowledge of `EvidenceLM`s details is incomplete. I'll keep digesting it but, but I'm approving so as not to hold up its merge.

ramyamounir:
Great, thanks @scottcanoe! And thanks @nielsleadholm and @vkakerbeck for the thorough and quick review.

ramyamounir:
@vkakerbeck @nielsleadholm @tristanls @codeallthethingz @scottcanoe @hlee9212 

I'd like to motion for final comment period. All maintainer please state your final comments/vote.

**I will take a thumbs up on this comment as a vote for "merge", and thumbs down as a vote for "No merge"**. However feel free to also write comments to explain or provide additional feedback if needed.

codeallthethingz (commented):
LGTM

codeallthethingz (approved):

rfcs/0000_hypotheses_resampling.md
vkakerbeck: We could call it resampled hypotheses, similar to here: https://stonesoup.readthedocs.io/en/v1.5/auto_tutorials/sampling/ResamplingTutorial.html Or offspring hypotheses (borrowing from evolutionary algorithm terminology) as it is a resampling of the "fittest" hypotheses. But yeah not sure those are much better.

rfcs/0000_hypotheses_resampling.md
tristanls: Both of these links would be useful to include in the RFC.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,187 @@
...
+This is a proposed approach to calculating the new set of hypotheses at every step. 
+This should take care of dynamically increasing or decreasing the hypotheses count.
+For example, if we want to start by coarse sampling of hypotheses and then increase the number of hypotheses (and vice versa), we should be able to do that here.
+
+I'm introducing three new parameters. The naming of these parameters is preliminary and subject to change.
+
+| Parameter | Description | Range |
+| ----------|-------------|-------|
+| **hypotheses_count_ratio** | A multiplier for the needed number of hypotheses at this new step. `1` means maintain the same number of hypotheses. | [0, inf) |
+| **hypotheses_old_to_new_ratio** | How many old to new hypotheses to be added. `0` means all old, `1` means all new | [0, 1] |
```
tristanls, line 94: **issue**: I find the use of `_ratio` in `hypothesis_old_to_new_ratio` (and this table) confusing. I would expect a ratio to not be bound to [0, 1]. When I think ratio of old to new I think of phrases like 2 old to 1 new 2:1, which would be 2 numerically. Maybe this could be called `hypothesis_old_to_new_range|mix|factor`?

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,187 @@
...
+
+This is a proposed approach to calculating the new set of hypotheses at every step. 
+This should take care of dynamically increasing or decreasing the hypotheses count.
+For example, if we want to start by coarse sampling of hypotheses and then increase the number of hypotheses (and vice versa), we should be able to do that here.
+
+I'm introducing three new parameters. The naming of these parameters is preliminary and subject to change.
+
+| Parameter | Description | Range |
+| ----------|-------------|-------|
+| **hypotheses_count_ratio** | A multiplier for the needed number of hypotheses at this new step. `1` means maintain the same number of hypotheses. | [0, inf) |
```
tristanls, line 93: **suggestion**: Rename to `hypothesis_count_multiplier` as per the description.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,187 @@
...
+        needed_hypotheses_count * (1 - hypotheses_old_to_new_ratio),
+        needed_hypotheses_count * hypotheses_old_to_new_ratio,
+    )
+    # needed old hypotheses should not exceed the existing hypotheses
+    # if trying to sample more hypotheses, set the available count as ceiling
+    if needed_old_sampled_hyp > curr_hypotheses_count:
+        needed_old_sampled_hyp = curr_hypotheses_count
+        needed_new_sampled_hyp = needed_hypotheses_count - curr_hypotheses_count
+
+    # calculate how many informed and re-enforced hypotheses needed
```
tristanls, line 150: **typo**: Should be `reinforced`.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,187 @@
...
+
+1) Calculate the needed hypotheses counts to be sampled based on the defined parameters as shown [here](#The-Resampling-Count-Calculation).
+2) Select `needed_old_sampled_hyp` from the existing hypotheses based on highest evidence slope. We will keep these old hypotheses and remove the rest.
+3) Uniformly sample `needed_new_informed_hyp` from the new informed hypotheses based on the observed pose. We will add these new hypotheses.
+4) Sample `needed_new_reinforced_hyp` from the existing hypotheses distribution of highest evidence slope. These sampled hypotheses should be "close" to the most likely hypotheses.
+
+
+## High Level Code Changes
+
+The needed modification will only change the `EvidenceLM` class. More specifically, we would modify either the `_update_evidence` function
```
tristanls, line 73: **typo**: [There is no `EvidenceLM` class](https://github.com/search?q=repo%3Athousandbrainsproject%2Ftbp.monty+%22class+EvidenceLM%22&type=code). I think you mean `EvidenceGraphLM`?

tristanls (commented):
I added some comments. Voting to merge.

rfcs/0000_hypotheses_resampling.md
ramyamounir: Yeah both are great! I am a bit biased towards offspring because resampled sounds a bit more generic, we've used the term resampling in many other places. I'll go with offspring for now but we can change it later in the actual PR.

rfcs/0000_hypotheses_resampling.md
ramyamounir: I see your point here. Yes, the actual parameter is not a ratio, but instead it allows us to set the ratio (e.g., 0.5 would give us 1:1). I like `factor`. We can also call it `weight`, as in the weight of old to new in the hypotheses mix.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,187 @@
...
+* **Practical**: The total evidence can still be unbounded, it doesn't matter because we only consider the slope. This metric does not care about how much evidence we've accumulated already. In other words, a hypothesis with a high evidence, can be removed if it hasn't accumulated evidence in a while, while a consistently growing hypothesis is less likely to be removed even if it was just added.
+
+## Assumptions and constraints:
+* Sampling of likely and unlikely hypotheses is based on evidence change instead of absolute evidence.
+* Terminal state calculation is still based on accumulated evidence and `x_percent_threshold`.
+
+## The Resampling Procedure
+
+1) Calculate the needed hypotheses counts to be sampled based on the defined parameters as shown [here](#The-Resampling-Count-Calculation).
+2) Select `needed_old_sampled_hyp` from the existing hypotheses based on highest evidence slope. We will keep these old hypotheses and remove the rest.
```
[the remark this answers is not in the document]
ramyamounir: Great, I've added this to the RFC. I've also changed `old_sampled` to `old_maintained` since it's not really stochastically sampled as Niels has pointed out in another discussion.

rfcs/0000_hypotheses_resampling.md
```diff
@@ -0,0 +1,187 @@
...
+
+Why:
+* **Faster**: we don't have to wait for high unbounded evidence to decay enough to realize that a new hypothesis is more likely. We also may not need to worry about initializing new hypotheses with mean evidence, giving them fighting chance against other old hypotheses. Average slope is more fair in this sense.
+* **Accurate resampling**:  If we sample new hypotheses close to the hypotheses with high total accumulated evidence (e.g., particle filter), we could be sampling from the incorrect hypotheses (if we had just switched objects). If we sample close to the hypotheses with high evidence slope, we may converge faster.
+* **Practical**: The total evidence can still be unbounded, it doesn't matter because we only consider the slope. This metric does not care about how much evidence we've accumulated already. In other words, a hypothesis with a high evidence, can be removed if it hasn't accumulated evidence in a while, while a consistently growing hypothesis is less likely to be removed even if it was just added.
```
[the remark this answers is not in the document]
ramyamounir: Yes, good point. Considering the average evidence slope over the last S steps only will allow us to forget about objects that disappear in the favor of other objects with rapidly rising evidence. It doesn't matter if the old object has quickly accumulated evidence in the past (i.e., before S steps). But that's another parameter to tune..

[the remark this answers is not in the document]
vkakerbeck: Yeah, many thoughts around this! It seems like cotton candy disappearing in water may actually be a learned object behavior (and if you haven't learned it, like this raccoon, you are quite surprised). Just making the evidence horizon shorter will have similar effects as we saw when using bounded evidence (past_weight+present_weight=1) where our current policies are not efficient enough to explore a sufficient area of the object to make a confident classification. Of course, we could mitigate this more with more efficient policies, but it seems like forgetting about stuff is more than just a time horizon parameter. I originally thought of it this way as well (which is why I suggested testing the bounded evidence first as a solution for switching hypotheses when moving from one object to another) but it seems like there should be more rapid model-free and model-based mechanisms at play here. Such as using predictions and prediction errors to reset hypotheses (like described here https://github.com/thousandbrainsproject/tbp.monty/pull/196#discussion_r1975422218).

PR #201, opened 2025-02-28
build: remove explicit mask that maybe causes 401

tristanls:
Upload errors with `('Authorization token is no longer valid', 401)`. Maybe it is the explicit `::add-mask::` GitHub directive messing things up. Removing it to test.

I'll monitor and rotate the token immediately if we leak it this way.


codeallthethingz (approved):
🎲

tristanls:
Thanks, the token works locally, so I assume it is the crafting of the CLI command in the action that's the issue.

tristanls:
No leak. Token is not displayed 👍

PR #202, opened 2025-02-28
build: set target platform before capturing (platform-dependent) path

tristanls:
Because I'm capturing the output path before setting target platform in the configuration, the output does not correspond to the actual target platform and nothing is found when trying to upload. In `osx` case, platform is `osx-arm64` but builds for `osx-64` and things don't line up.

This pull request ensures configuration is present before we read the output path.

This should get `osx-64` upload to work.


codeallthethingz (approved):
LGTM

PR #203, opened 2025-02-28
build: tag version commit with version tag

tristanls:
This pull request introduces a `tag_version_monty_publish` job to the Monty Publish workflow.

The job is only triggered when Monty workflow on the `main` branch succeeds, which is after a commit is merged to `main`.
```yml
  tag_version_monty_publish:
    name: tag-version-monty-publish
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
    ...
```

The version is tagged only when that commit contains a version change in `src/tbp/monty/__init__.py`.
```yml
      ...
      - name: Tag version
        if: ${{ steps.version_updated.outputs.version_updated == 'true' }}
        working-directory: tbp.monty
        run: |
          git tag v${{ steps.version_updated.outputs.version }}
          git push origin v${{ steps.version_updated.outputs.version }}
```

The Version updated action now additionally returns `version` from `src/tbp/monty/__init__.py` (in addition to currently returning the `version_updated` boolean). This `version` is used in the tag name. For example, if the version is `0.0.7` then the tag will be `v0.0.7`

There are fewer `if:` conditionals in this job because it only triggers on the conclusion of Monty workflow and will not run on the manual trigger.

### DIFF
--- a/.github/actions/version_updated/check_version_updated.sh
+++ b/.github/actions/version_updated/check_version_updated.sh
@@ -5,6 +5,8 @@ do
   echo $changed_file
   if [[ $changed_file == src/tbp/monty/__init__.py ]]; then
     echo "version_updated=true" >> $GITHUB_OUTPUT
+    version=$(grep '__version__\s*=\s*' src/tbp/monty/__init__.py | awk -F'"' '{print $2}')
+    echo "version=$version" >> $GITHUB_OUTPUT
     exit 0
   fi
 done < changed_files.txt
--- a/.github/actions/version_updated/check_version_updated.sh
+++ b/.github/actions/version_updated/check_version_updated.sh
@@ -5,6 +5,9 @@ do
   echo $changed_file
   if [[ $changed_file == src/tbp/monty/__init__.py ]]; then
     echo "version_updated=true" >> $GITHUB_OUTPUT
+    # grep finds the version line. It will only have two quotes. Then awk uses
+    # the quotes as separators and prints the middle value, which corresponds to
+    # the version inside the quotes.
     version=$(grep '__version__\s*=\s*' src/tbp/monty/__init__.py | awk -F'"' '{print $2}')
     echo "version=$version" >> $GITHUB_OUTPUT
     exit 0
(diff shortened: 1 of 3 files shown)


.github/actions/version_updated/check_version_updated.sh
```diff
@@ -5,6 +5,11 @@ do
   echo $changed_file
   if [[ $changed_file == src/tbp/monty/__init__.py ]]; then
     echo "version_updated=true" >> $GITHUB_OUTPUT
+    # grep finds the version line. It will only have two quotes. Then awk uses
+    # the quotes as separators and prints the middle value, which corresponds to
+    # the version inside the quotes.
+    version=$(grep '__version__\s*=\s*' src/tbp/monty/__init__.py | awk -F'"' '{print $2}')
```
codeallthethingz, line 11: you could use the tools/print_version command here instead.

codeallthethingz (approved):
LGTM.  One comment around the use of the print_version tool.

.github/actions/version_updated/check_version_updated.sh
tristanls: Yes, I considered it. I believe that in order to use the `print_version` tool, I need to create the conda environment, install the tool, and invoke it. Is that correct?

codeallthethingz: Yep.

tristanls: > Yes. Though you don't have install anything.

What do you mean?

When I run the tool without installing via `python -m tools.print_version.cli -h` I get an error:
```
(tbp.monty) me@not_invented_here tbp.monty % python -m tools.print_version.cli -h 
  File "/opt/anaconda3/envs/tbp.monty/lib/python3.8/runpy.py", line 194, in _run_module_as_main
  File "/opt/anaconda3/envs/tbp.monty/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/me/tbp/tbp.monty/tools/print_version/cli.py", line 15, in <module>
    import semver
ModuleNotFoundError: No module named 'semver'
```

tristanls: I'm going to merge this as is for now and we can revisit in the future.

PR #204, opened 2025-03-04
build: source dependencies from conda

tristanls:
In our current setup, we use two package managers to satisfy Monty dependencies: `conda` and `pip`. This was fine when we were not distributing Monty. However, to publish and distribute Monty, we need to pick a single source for our dependencies. Today, this needs to be `conda` as not all dependencies are available on `pip` (e.g., `habitat_sim`).

By using `conda` to source all dependencies, when we publish `thousandbrainsproject::tbp.monty`, we can specify dependencies that should be installed when someone installs `thousandbrainsproject::tbp.monty`. If we do not do this, there is no standard install method for Monty on either `conda` or `pip`. (Without a standard install method, we would be asking users to execute a custom sequence of steps; not impossible, but not as accessible as `conda env create`).

With the changes in this pull request, the following conda `environment.yml` is expected to work out of the box (once this version is published) and install _all_ dependencies needed for Monty to run:
```yml
name: monty_lab
channels:
  - aihabitat
  - pytorch
  - pyg
  - defaults
  - conda-forge
  - thousandbrainsproject

dependencies:
  - thousandbrainsproject::tbp.monty
```
> [!NOTE]
> In `environment.yml` above, we need to specify all the channels that `tbp.monty` depends on because `conda` dependency file does not support specifying channels (see: https://github.com/conda/conda-build/issues/532). For example, even though `tbp.monty` depends on `aihabitat::habitat-sim`, in the `conda` dependency file this will be truncated to `habitat-sim`. Without specifying the necessary channels, installation will fail as `habitat-sim` is not in the `defaults`. When we specify `aihabitat` in the `channels:` property, `conda` install will include `aihabitat::habitat-sim` in the search for `habitat-sim`.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -29,6 +29,7 @@ dependencies:
   - conda-forge::cmake>=3.14.0
   - conda-forge::importlib_resources
   - conda-forge::matplotlib
+  - conda-forge::mkl==2024.0.0 # see https://github.com/pytorch/pytorch/issues/123097
   - conda-forge::numpy
   - conda-forge::pandas
   - conda-forge::pillow
--- a/environment.yml
+++ b/environment.yml
@@ -18,20 +18,31 @@ channels:
   - aihabitat
   - pytorch
   - pyg
-  - defaults
   - conda-forge
+  - defaults
 
 dependencies:
   - python=3.8
-  - cmake>=3.14.0
-  - pyg::pyg
-  - wget
 
   - aihabitat::habitat-sim=0.2.2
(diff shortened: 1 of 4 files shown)


tristanls:
FYI: @ramyamounir, @nielsleadholm, @hlee9212 , @scottcanoe

codeallthethingz (approved):
LGTM

vkakerbeck (approved):
Looks good :)

PR #209, opened 2025-03-11
chore: sanitize snippets

codeallthethingz:
This PR runs nh3.clean on any snippets as part of the markdown process when syncing the documentation.

It also adds an info statement to the sync process that prints out the URL to the documentation.


tristanls (approved):
Thank you for catching and updating this.

PR #210, opened 2025-03-13
minor modifications to evidence updating

ramyamounir:
While working on resampling hypotheses, I did some simplification to the `_update_evidence` and `_add_hypotheses_to_hpspace` functions as well as introduced a `ChannelMapper` class to replace the existing `channel_hypotheses_mapping` logic. I thought it would be a good idea to isolate these modifications from the actual resampling code and submit this as a separate PR (as opposed to a huge PR with big changes tangled up). Here is a brief summary of the changes:

### Channel Mapper
The current implementation stacks the hypotheses from all input channels in the same array to perform efficient vector operations on them. To identify which hypotheses correspond to which input channel, `channel_hypothesis_mapping` dictionary stores the ranges of indices for each channel and the total number of hypotheses over all channels. Here is an example:

```python
EvidenceGraphLM.channel_hypothesis_mapping = {'mug': {'patch_1':[0, 499], 'patch_2': [500, 789], 'num_hypotheses':790}
```
While this format doesn't necessarily pose a problem for resampling, it would not be easy to change the size of hypotheses for each input channel. For example, to change `patch_1` from [0,499] to [0,399], we would also have to update all subsequent channel ranges and update the total `num_hypotheses`. We want to allow for dynamic resizing of hypothesis space during resampling.

The proposed solution is to store only the size of the input channel hypotheses in a List or OrderedDict (any data structure that maintains the order will work fine). The ranges can be efficiently computed with cumulative sum calculation and the `num_hypotheses` is just the sum of all channel sizes. This functionality has been defined in the class `ChannelMapper()` in the `utils/evidence_matching.py` file. This change has already removed the need for much of the code that keeps track of `num_hypotheses`, such as [here](https://github.com/thousandbrainsproject/tbp.monty/blob/6d0f70db6440df99b46195d529217a7799ae9c93/src/tbp/monty/frameworks/models/evidence_matching.py#L880) and [here](https://github.com/thousandbrainsproject/tbp.monty/blob/6d0f70db6440df99b46195d529217a7799ae9c93/src/tbp/monty/frameworks/models/evidence_matching.py#L983).

### Initializing Hypothesis Space (Not implemented in this PR)
The `_update_evidence` draws a distinction between these two cases:
* We have no displacement
* We have displacement but no existing hypotheses

In both cases, we initialize a hypotheses space in roughly the same way. We `_get_initial_hypothesis_space` then add these hypotheses to the initialized space. I have rewritten this part of the `_update_evidence` function to apply the same logic for both cases. I also modified the `_add_hypotheses_to_hpspace` function to handle the case where there is already a hypothesis space initialized. In this case, we want to update the evidence with the mean of existing hypotheses.

### Explicit Evidence Reset (Not implemented in this PR)
The signal for resetting the evidence scores is not straightforward to find. The resetting depends on whether we have displacement or not. If we don't have displacement, we overwrite the existing evidence with new evidence scores, as shown [here](https://github.com/thousandbrainsproject/tbp.monty/blob/6d0f70db6440df99b46195d529217a7799ae9c93/src/tbp/monty/frameworks/models/evidence_matching.py#L990). I think this relies on the buffer to be reset. I suspect these indirect signals for evidence reset may cause some strange behavior in the unsupervised inference testbed. A direct `evidence={}` in the `EvidenceGraphLM.reset` function (called from pre_episode) will be easier to debug.

I explicitly reset evidence in the `EvidenceGraphLM.reset` function and removed the dependency on the `displacments is None` reset signal in `_update_evidence`. I think this is a more direct signal of reset. Since the evidence is now cleared out in the reset function, we can just check if `graph_id` is in evidence.keys() to decide on initializing hypotheses space vs. add to an existing space. This has been done in `_add_hypotheses_to_hpspace` now as proposed in this PR.

---

## Testing on Benchmarks
I ran benchmark tests on 10 objects (surface and distant agent), as well as 77 objects (surface agent), before and after the changes. I've seen no change in performance. The exact benchmark experiments I ran are:
* `randrot_noise_10distinctobj_surf_agent`
* `randrot_noise_10distinctobj_dist_agent`
* `randrot_noise_77obj_surf_agent`


scottcanoe (commented):
Nice @ramyamounir, I think this is a nice approach, and it helps quite a bit with the readability of `EvidenceLM._update_evidence`.

My generic feedback is that it'd be really nice to include type annotation for `ChannelMapper`, and if you've got the time, anywhere in `EvidenceLM` while the parameters are fresh in your mind (though type annotation for `EvidenceLM` could be done in a separate PR). I always struggle a bit when reading `EvidenceLM` code since it can be hard to keep track of what the data structures are, so any extra annotation/documentation would be much appreciated.

And just for my understanding, it seems like an `input_channel` refers a source (either an SM or LM), but `EvidenceLM._update_evidence` seems to get a list of potential input channels from a dictionary of features. Are these features like `rgba`? Or principal curvature vectors? If so, how does that square with the idea that an input channel corresponds to different SMs or LMs?

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -857,33 +858,44 @@ def _add_hypotheses_to_hpspace(
...
-        new_num_hypotheses = old_num_hypotheses + len(new_loc_hypotheses)
-        self.channel_hypothesis_mapping[graph_id][input_channel] = [
-            old_num_hypotheses,
-            new_num_hypotheses,
-        ]
-        self.channel_hypothesis_mapping[graph_id]["num_hypotheses"] = new_num_hypotheses
+
+        if graph_id in self.evidence.keys():
+            current_mean_evidence = np.mean(self.evidence[graph_id])
+            new_evidence = new_evidence + current_mean_evidence
```
tristanls, line 872: **suggestion**: `+=`?

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -857,33 +858,44 @@ def _add_hypotheses_to_hpspace(
...
+                    new_rot_hypotheses,
+                ]
+            )
+            self.evidence[graph_id] = np.hstack([self.evidence[graph_id], new_evidence])
+
+            self.channel_hypothesis_mapping[graph_id].increase_channel_size(
+                input_channel, len(new_evidence)
+            )
+
+        else:
```
tristanls, line 891: **issue**: This `else` is surprising to me because the state of `self` is such that `graph_id` is not present in `self.evidence`, nor `self.possible_locations`, nor `self.possible_poses`. However, `graph_id` _is_ present in `self.channel_hypothesis_mapping`. It feels to me like the initialization is split between two methods, here and in `_update_evidence`.

**question**: Would it work to move `ChannelMapper()` initialization here too?
```python
        if graph_id not in self.channel_hypothesis_mapping:
            self.channel_hypothesis_mapping[graph_id] = ChannelMapper()
```

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -857,33 +858,44 @@ def _add_hypotheses_to_hpspace(
         new_rot_hypotheses,
         new_evidence,
     ):
-        """Add new hypotheses to hypothesis space."""
+        """Add new hypotheses to hypothesis space.
+
+        If the hypothesis space doesn't exist, it will be initialized here.
```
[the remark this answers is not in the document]
ramyamounir: I agree, it would be nice to move towards type annotations everywhere. I've added type annotations and google-style docstrings to `ChannelMapper`, and I'll add those to `EvidenceGraphLM` in the next PR(s) as I dive deeper into the LM code.

src/tbp/monty/frameworks/utils/evidence_matching_utils.py
```diff
@@ -0,0 +1,127 @@
...
+        if channel_name not in self.channel_sizes:
+            raise ValueError(f"Channel '{channel_name}' not found.")
+
+        start = 0
+        for name, size in self.channel_sizes.items():
+            if name == channel_name:
+                return (start, start + size)
+            start += size
+
+    def increase_channel_size(self, channel_name, value):
```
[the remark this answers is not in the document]
ramyamounir: Yes, It can shrink channel size if `value` is <0. I've changed the name to `resize_channel_by` as suggested by Tristan. The resizing will be useful for resampling hypotheses.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -941,161 +953,108 @@ def _update_evidence(self, features, displacements, graph_id):
...
-            num_hypotheses = 0
-            for input_channel in input_channels_to_use:
+
+        if graph_id not in self.channel_hypothesis_mapping:
+            self.channel_hypothesis_mapping[graph_id] = ChannelMapper()
+
+        # get all usable input channels
+        input_channels_to_use = [
+            ic
+            for ic in list(features.keys())
```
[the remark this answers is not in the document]
ramyamounir: Good point, fixed.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -941,161 +953,108 @@ def _update_evidence(self, features, displacements, graph_id):
...
+            if ic in self.get_input_channels_in_graph(graph_id)
```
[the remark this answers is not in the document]
ramyamounir: Yes, good point. It gets called every iteration but it calls `GraphMemory.get_input_channels_in_graph` which just returns the keys of a dictionary (`GraphMemory.models_in_memory[graph_id].keys()`). There is not much compute going on, so it seems better to just call it every iteration.

src/tbp/monty/frameworks/utils/evidence_matching_utils.py
```diff
@@ -0,0 +1,127 @@
...
+    This allows us to dynamically change the range of one channel without
+    having to recompute the range of all subsequent channels. We also do not
+    need to store "num_hypotheses" as it can be computed with the `total_size`
+    property.
+    """
+
+    def __init__(self, channel_sizes=None):
+        """Initialize the ChannelMapper with an ordered dictionary of channel sizes.
+
+        :param channel_sizes: Dict of {channel_name: size}, maintaining order.
```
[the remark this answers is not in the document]
ramyamounir: Fixed.

src/tbp/monty/frameworks/utils/evidence_matching_utils.py
```diff
@@ -0,0 +1,127 @@
...
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+
+from collections import OrderedDict
+
+
+class ChannelMapper:
+    """This class marks the range of hypotheses that correspond to each input channel.
+
+    Instead of storing the actual range as a List or Tuple of indices,
```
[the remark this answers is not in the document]
ramyamounir: Nice, yeah I fixed it.

src/tbp/monty/frameworks/utils/evidence_matching_utils.py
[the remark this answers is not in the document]
ramyamounir: Fixed, added and added.

src/tbp/monty/frameworks/utils/evidence_matching_utils.py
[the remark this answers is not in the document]
ramyamounir: Good point. I renamed it to `resize_channel_by` and renamed `set_channel_size` to `resize_channel`. Thanks.

src/tbp/monty/frameworks/utils/evidence_matching_utils.py
```diff
@@ -0,0 +1,127 @@
...
+    def add_channel(self, channel_name, size, position=None):
+        """Adds a new channel at a specified position (default is at the end).
+
+        Raises:
+            ValueError: If the channel already exists.
+        """
+        if channel_name in self.channel_sizes:
+            raise ValueError(f"Channel '{channel_name}' already exists.")
+
+        if position is None or position >= len(self.channel_sizes):
```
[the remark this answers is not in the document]
ramyamounir: Yeah, good point. Fixed.

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: I see your point. If I move the initialization of `ChannelMapper` to the `_add_hypotheses_to_hpspace` function, it will give an error on new episodes because of this check below:

```python
for input_channel in input_channels_to_use:
            # If the channel doesn't exist, initialize it and add it.
            if input_channel not in self.channel_hypothesis_mapping[graph_id].channels:
```
I check if the input_channel is in the ChannelMapper, so it has to be initialized beforehand. Note that if it's not a new episode, we will already have a ChannelMapper, but we may not have the required input_channel. Does that answer your question?

ramyamounir:
Thanks for the reviews, I've addressed most of the comments.

@scottcanoe the matching step receives the observations in CMP-compliant format, but the function `_select_features_to_use` is used on these observations to extract the "features_to_use" and arranges them with the input_channel as the key. It just extracts the required features based on tolerances defined in configs. So, in a hierarchy, we will want to define different features to be extracted because hsv will not be a higher-level feature. Not sure if I got everything here right, or if @nielsleadholm or @vkakerbeck would like to add or correct something.

@tristanls I think I got all of your comments, except for the ones suggesting to remove unused function in `ChannelMapper`. I've gotten into the habit of defining useful functions and properties for classes so that other users can use those helper functions without trying to understand the logic of the whole class. Is code bloat your concern here?

src/tbp/monty/frameworks/utils/evidence_matching_utils.py
```diff
@@ -0,0 +1,127 @@
...
+        """Returns the existing channel names."""
+        return list(self.channel_sizes.keys())
+
+    @property
+    def dict(self):
+        """Returns the ordered dictionary."""
+        return self.channel_sizes
+
+    @property
+    def total_size(self):
```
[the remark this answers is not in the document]
ramyamounir: Yes, with `ChannelMapper` we won't need to track `num_hypotheses`. I just left the `total_size` property there because it's useful to quickly figure out how many hypotheses over all input channels exist. I've compared this to the number of hypotheses as a check in an integration test. I also expect to use some of these functions in hypotheses resampling.

nielsleadholm (commented):
Looks great, I think this is a nice and clean approach to this. Just added a few comments on documentation/naming.

One other question - you alluded in the PR description to the changes possibly affecting unsupervised learning, did you have a chance to check whether these benchmark experiments give the same result?

tristanls:
> I've gotten into the habit of defining useful functions and properties for classes so that other users can use those helper functions without trying to understand the logic of the whole class. Is code bloat your concern here?

@ramyamounir, adding unused functions to code is problematic for several reasons. As you mentioned, it creates code bloat. It increases the complexity of the code; there are already multiple unused functions in Monty that make understanding the code difficult. Unused code must be maintained so it adds to the maintenance overhead. Changing the implementation requires updating unused code. The presence of unused code confuses analysis in general. You spent time writing, and we now had a few rounds of conversation about code that never runs. This is coding and review time we could have spent on other things. From experience, assumptions of future use often, not always, turn out wrong. In addition to writing the code, there's also the overhead of writing and reviewing tests for code that never runs.

jeremyshoemaker:
> > I've gotten into the habit of defining useful functions and properties for classes so that other users can use those helper functions without trying to understand the logic of the whole class. Is code bloat your concern here?
> 
> @ramyamounir, adding unused functions to code is problematic for several reasons. As you mentioned, it creates code bloat. It increases the complexity of the code; there are already multiple unused functions in Monty that make understanding the code difficult. Unused code must be maintained so it adds to the maintenance overhead. Changing the implementation requires updating unused code. The presence of unused code confuses analysis in general. You spent time writing, and we now had a few rounds of conversation about code that never runs. This is coding and review time we could have spent on other things. From experience, assumptions of future use often, not always, turn out wrong. In addition to writing the code, there's also the overhead of writing and reviewing tests for code that never runs.

I just want to second everything Tristan said. I have an example from this morning of trying to understand what a method was doing that seemed to not be doing anything, only to realize after half an hour that it was overridden in every subclass, and thus dead code. Sometimes tooling can quickly help with identifying dead code and prevent these kinds of problems, but with a language as dynamic as Python, tooling can only do so much.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -941,161 +952,109 @@ def _update_evidence(self, features, displacements, graph_id):
...
+            self.channel_hypothesis_mapping[graph_id] = ChannelMapper()
+
+        # get all usable input channels
+        input_channels_to_use = [
+            ic
+            for ic in features.keys()
+            if ic in self.get_input_channels_in_graph(graph_id)
+        ]
+
+        if displacements is None and len(input_channels_to_use) == 0:
```
vkakerbeck, line 966: These were two separate if statements before. I don't understand how this should work if they are combined. If the displacement is None, we still want to use the observed features to initialize the hypothesis space and determine the initial evidence for it. I think right now you check whether the input_channel is in the hypothesis mapping instead, which may be why the tests still work but I am not sure those are always equivalent. Did you make sure this logic works in all cases? Maybe you can document this a bit more. To me the previous logic was very clearly distinguishing between the first observation before we have move (not the same as resetting!) and observations after a movement. Now this logic seems to be implemented as more of a side effect of whether a channel is stored in the hypothesis space. Maybe I am missing something.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -941,161 +952,109 @@ def _update_evidence(self, features, displacements, graph_id):
...
-                num_hypotheses += len(initial_possible_channel_locations)
-            self.channel_hypothesis_mapping[graph_id]["num_hypotheses"] = num_hypotheses
-            self.possible_locations[graph_id] = np.concatenate(
-                initial_possible_locations, axis=0
-            )
-            self.possible_poses[graph_id] = np.concatenate(
-                initial_possible_rotations, axis=0
-            )
-            self.evidence[graph_id] = (
-                np.concatenate(initial_evidence, axis=0) * self.present_weight
```
vkakerbeck, line 991: Where did this part go? This is not resetting the evidence, it is using the sensed features to determine the initial evidence. I don't see where this is done in your code now.

vkakerbeck (commented):
Nice idea! I like how it cleans some of this channel mapping up. Also nice that you added unit tests already. 
I didn't do a super detailed review (waiting for changes re the other comments to be incorporated) but just left 2 high-level comments for now. They are on an aspect of the old implementation that I don't see here anymore now. I am probably just missing something if the benchmark results are still exactly the same but want to make sure.

Also, since there is only one hierarchy unit test and this is probably the condition that is most affected, maybe you could run a before and after comparison of an experiment with `two_stacked_lms_config` (like your `two_lm_distant_heterarchy_sdr` experiment) and make sure the results are still exactly the same.

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: The way I think about this is that the `_update_evidence` function had three scenarios:
* `displacements is None`: This happens when the buffer is reset. It corresponds to the `len(self.buffer)<=1` as shown [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e62e1bca981f7ef3a4e2832e11f80056ce189981/src/tbp/monty/frameworks/models/graph_matching.py#L652). When this happens, we reinitialize the hypothesis space for every input channel and concatenate these hypotheses. The new hypothesis space [replaces](https://github.com/thousandbrainsproject/tbp.monty/blob/e62e1bca981f7ef3a4e2832e11f80056ce189981/src/tbp/monty/frameworks/models/evidence_matching.py#L990) the existing one.
* `displacements is not None` and `input_channel not in self.channel_hypothesis_mapping[graph_id].keys()`: This happens when we have a displacement but the incoming channel does not have a hypothesis space associated with it. In this case, we initialize a hypothesis space only for this `input_channel` and add it to the existing hypotheses of this `graph_id`. The evidence here is initialized with the mean of other existing hypotheses evidences from different channels.
* `displacements is not None` and `input_channel in self.channel_hypothesis_mapping[graph_id].keys()`: This is the default case of displacing the hypotheses and updating existing evidence for this channel on this graph_id.

The main thing I observed here is that the second scenario is just a special case of the first scenario. When `displacements is None`, we reinitialize the hypotheses for all input_channels. When `displacements is not None` and `input_channel` doesn't exist, we only initialize the `input_channel` hypotheses.

So I could just delete existing hypotheses for all channels if the `displacement is None`, and then loop over all the `input_channels`. If it can't find the input channel, it will just initialize it and add it. This should cover the first two scenarios in the same loop. Deleting existing hypotheses happens at the beginning of the episode by resetting `channel_hypotheses_mapping` and `evidence`.

Now that I think through all the cases again, there is one case where the results might differ. It's an easy fix, I'll update the PR for it and discuss it more. Otherwise, the logic above makes sense?

Note that the two separate if statements you highlighted above are there just to capture this [ValueError](https://github.com/thousandbrainsproject/tbp.monty/blob/e62e1bca981f7ef3a4e2832e11f80056ce189981/src/tbp/monty/frameworks/models/evidence_matching.py#L956). I wanted to keep all the errors you initially coded. Nothing is under this statement except for a `raise ValueError`

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: This has been moved to [here](https://github.com/thousandbrainsproject/tbp.monty/blob/a5c1fe7b825f05dbd968cc22b55d12fb6e2b365c/src/tbp/monty/frameworks/models/evidence_matching.py#L893). The `add_hypotheses_to_hpspace` now adds any new input_channel hypotheses to a hypotheses space, this also applies to when the `displacement is None`. As mentioned [here](https://github.com/thousandbrainsproject/tbp.monty/pull/210#discussion_r2003384408), I realized there is one case where this might give different results, I'll update the PR to handle it.

src/tbp/monty/frameworks/models/evidence_matching.py
vkakerbeck: Okay, thanks for the detailed explanation! That makes sense :) I just wanted to make sure it is still giving the same results in all the scenarios with the now changed if statements. I hadn't seen the line that you linked in the response to my other comment so now that I see that, I think it should be good.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -0,0 +1,168 @@
...
+class ChannelMapper:
+    """Marks the range of hypotheses that correspond to each input channel.
+
+    The `EvidenceGraphLM` implementation stacks the hypotheses from all input channels
+    in the same array to perform efficient vector operations on them. Therefore, we
+    need to keep track of which indices in the stacked array correspond to which input
+    channel. This class stores only the sizes of the input channels in an ordered data
+    structure (OrderedDict), and computes the range of indices for each channel. Storing
+    the sizes of channels in an ordered dictionary allows us to insert or remove
+    channels, as well as dynamically resize them.
```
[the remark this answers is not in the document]
ramyamounir: Yeah, I'll keep an eye on this. There is of course not much overhead to modifying the mapping, but there might be some overhead to the resampling processes and resizing the Numpy arrays where the hypotheses are stored.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -0,0 +1,168 @@
...
+    def channels(self) -> List[str]:
+        """Returns the existing channel names.
+
+        Returns:
+            List[str]: List of channel names.
+        """
+        return list(self.channel_sizes.keys())
+
+    @property
+    def dict(self) -> OrderedDictType[str, int]:
```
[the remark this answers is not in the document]
ramyamounir: I'll remove it anyway since it's one of the unused properties.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        if channel_name not in self.channel_sizes:
+            raise ValueError(f"Channel '{channel_name}' not found.")
+        if self.channel_sizes[channel_name] + value <= 0:
+            raise ValueError(
+                f"Channel '{channel_name}' size cannot be negative or zero."
+            )
+        self.channel_sizes[channel_name] += value
+
+    def resize_channel(self, channel_name: str, value: int) -> None:
```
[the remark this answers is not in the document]
ramyamounir: Yes, good point!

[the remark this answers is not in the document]
ramyamounir: This will also be removed because it's unused..

src/tbp/monty/frameworks/models/evidence_matching.py
tristanls: This isn't easy to wrap my head around. I'm going to wait for the next round of reviews to try again to understand this setup.

ramyamounir:
Thanks everyone for the reviews. I've made some changes, namely:
* I removed the unused properties and functions from `ChannelMapper` and the unit tests, as suggested by @tristanls and @jeremyshoemaker. I've only kept the `total_size` property as i'm currently using it in an EvidenceLM integration test.
* @nielsleadholm I've addressed your comments on documentation and variable names. It should be much clearer now.
* @vkakerbeck I realized there is a scenario (i.e.,  `len(input_channels_to_use)>1` and `displacement is None`) where we might get different evidence scores after initializing all input_channels one by one. The `_add_hypotheses_to_hpspace` adds the mean of existing evidence from other channels by default if those channels exist. I made sure this shouldn't happen by passing a parameter to control if we want to add evidence mean.
* @vkakerbeck I couldn't find `stacked_lms` experiment configs for testing. However, the unit tests have `test_two_lm_heterarchy_experiment` which runs fine.
* In `_add_hypotheses_to_hpspace` I've changed the main if statement to check if there exists other channels, instead of checking if graph_id is in evidence.keys(). I think this is equivalent but much easier to read and understand.
* I ran all 3 benchmarks again to validate performance doesn't change.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -857,18 +858,23 @@ def _add_hypotheses_to_hpspace(
...
+        will be added to the initial evidence. This is done to give the new input
+        channel a fighting chance.
         """
         # Add current mean evidence to give the new hypotheses a fighting
         # chance. TODO H: Test mean vs. median here.
+        new_evidence = np.array(new_evidence * self.present_weight)
 
-        if graph_id in self.evidence.keys():
-            new_evidence += np.mean(self.evidence[graph_id])
+        if len(self.channel_hypothesis_mapping[graph_id].channels):
```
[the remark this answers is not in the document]
ramyamounir: `graph_id` will always be in the `channel_hypotheses_mapping` when we reach the `_add_hypotheses_to_hpspace` function because it's added here in this [line](https://github.com/thousandbrainsproject/tbp.monty/blob/246aee3e3c0fbdebf93c5e9995d290ed2e7d6741/src/tbp/monty/frameworks/models/evidence_matching.py#L963) of `_update_evidence`

`channel_hypotheses_mapping` is responsible for all `input_channel` related checks now.  In my mind the logic is as follows:
* In `_update_evidence`, we check if a `ChannelMapper` is initialized for this `graph_id`, if not we initialize it [here](https://github.com/thousandbrainsproject/tbp.monty/blob/246aee3e3c0fbdebf93c5e9995d290ed2e7d6741/src/tbp/monty/frameworks/models/evidence_matching.py#L963)
* In `_update_evidence`, we also check if the current input_channel is one of the channels in the mapping. This check is done [here](https://github.com/thousandbrainsproject/tbp.monty/blob/246aee3e3c0fbdebf93c5e9995d290ed2e7d6741/src/tbp/monty/frameworks/models/evidence_matching.py#L981) and would determine if we need to initialize hypotheses for this `input_channel`
* When we reach `_add_hypotheses_to_hpspace`, we already know there exists a `ChannelMapper` (i.e., `graph_id is in channel_hypotheses_mapping.keys()`) and that the `input_channel` is not in the mapping channels. We can now check if there are other channels so that we can use their evidence as mean. This check is done [here](https://github.com/thousandbrainsproject/tbp.monty/blob/246aee3e3c0fbdebf93c5e9995d290ed2e7d6741/src/tbp/monty/frameworks/models/evidence_matching.py#L875)

Hopefully this clarifies things.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -890,7 +896,7 @@ def _add_hypotheses_to_hpspace(
         else:
             self.possible_locations[graph_id] = np.array(new_loc_hypotheses)
             self.possible_poses[graph_id] = np.array(new_rot_hypotheses)
-            self.evidence[graph_id] = np.array(new_evidence * self.present_weight)
+            self.evidence[graph_id] = np.array(new_evidence)
```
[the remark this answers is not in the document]
ramyamounir: I just thought it should apply to both conditions. Instead of applying it twice ([here](https://github.com/thousandbrainsproject/tbp.monty/blob/246aee3e3c0fbdebf93c5e9995d290ed2e7d6741/src/tbp/monty/frameworks/models/evidence_matching.py#L890) and [here](https://github.com/thousandbrainsproject/tbp.monty/blob/246aee3e3c0fbdebf93c5e9995d290ed2e7d6741/src/tbp/monty/frameworks/models/evidence_matching.py#L899)).

Also, it should be applied before adding the evidence mean from other channels. It just seemed like the right place for it but if it's confusing, we can move back under the `if` and `else` conditions.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -857,18 +858,23 @@ def _add_hypotheses_to_hpspace(
...
-        and initialize its evidence with the mean of existing hypotheses' evidence.
+        If the hypothesis space already exists, we will append new hypotheses to
+        this existing space. Otherwise, we will initialize the space. Note that
+        if `add_evidence_mean` is True, the mean of existing hypotheses evidence
+        will be added to the initial evidence. This is done to give the new input
+        channel a fighting chance.
         """
         # Add current mean evidence to give the new hypotheses a fighting
         # chance. TODO H: Test mean vs. median here.
+        new_evidence = np.array(new_evidence * self.present_weight)
```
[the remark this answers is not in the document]
ramyamounir: Yes, I missed that. I'll move it.

ramyamounir:
Thanks for the review and suggested experiments @vkakerbeck! These are all very reasonable suggestions considering how tricky these changes are. I want to point out that the next PR on hypotheses resampling will change the proposed code of this PR quite a bit. Specifically, the `_update_evidence` will change again and will be significantly refactored to accommodate the changes below.

Update Evidence

So, if you're anxious about merging this one because of big changes to updating evidence, maybe we could draft it for now and work towards the next PR directly. The important change in this PR here is the `ChannelMapper` because this will stay for the next PR and you seem to be okay with that change. Alternatively, I can restore some parts of the code that relate to update evidence and only keep the changes related to `ChannelMapper` in this PR. What do you think?

nielsleadholm:
@vkakerbeck just to say that Ramy and I discussed this and I think it makes sense to simplify this PR to just the ChannelMapper, since there will be more significant changes to the update_evidence logic anyways. @ramyamounir maybe if you simplify to those changes and then tag us, and we can have a final last look.

vkakerbeck:
Okay, that sounds good :) I like the addition of the ChannelMapper (and also don't have a problem with the other changes, just want to make sure it preserves the same behaviors).

ramyamounir:
@nielsleadholm @vkakerbeck I've updated the PR so that only changes related to the `ChannelMapper` are incorporated. As discussed, other simplifications to the `_update_evidence` will be in follow-up PRs.

nielsleadholm (approved):
Great! Did you have a chance to do a quick sanity check with one of the benchmark experiments to ensure performance is the same now that you've reverted most of these changes, but still have the channel mapper? Otherwise happy to merge.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -0,0 +1,128 @@
...
+        if channel_name not in self.channel_sizes:
+            raise ValueError(f"Channel '{channel_name}' not found.")
+
+        start = 0
+        for name, size in self.channel_sizes.items():
+            if name == channel_name:
+                return (start, start + size)
+            start += size
+
+    def resize_channel_by(self, channel_name: str, value: int) -> None:
```
vkakerbeck, line 77: When I read this name it wasn't clear to me if this function set's the channel size to `value` or adds more entries to the existing hypotheses. Maybe you could use a less ambiguous name for the function (like `increase_channel_by`) or for the `value` parameter (like `num_entries_to_add`)

vkakerbeck (approved):
Looks great, thanks for adding this channel mapper :)

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: This function was originally named `increase_channel_size`, but Scott and Tristan [pointed out](https://github.com/thousandbrainsproject/tbp.monty/pull/210#discussion_r1994342810) that the function is also used for decreasing the size as well, so it seemed more fitting to rename it to `resize`. Tristan also suggested `resize_channel_by` to highlight that value here would refer to incrementing or decrementing value. I had another [function](https://github.com/thousandbrainsproject/tbp.monty/pull/210#discussion_r1998932415) named `resize_channel_to` but ended up removing it because it's currently unused.

ramyamounir:
> Did you have a chance to do a quick sanity check.

@nielsleadholm, Yes. I re-ran all three benchmark experiments listed in the PR summary, all looks good.

@tristanls, please take another look and let me know if you have additional feedback. This PR now only introduces the `ChannelMapper`. Thanks.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -340,10 +339,11 @@ def __init__(
...
         self.possible_poses = {}
-        # Stores start and end indices of hypotheses in the above arrays for each graph
-        # corresponding to each input channel. This is used to make sure the right
-        # displacement is applied to the right hypotheses. Channel hypotheses are stored
-        # contiguously so we can just specify ranges here.
+        # Initializes a `ChannelMapper` object for each graph_id. The mapper stores the
+        # length of each input_channel in the hypotheses space defined in the above
+        # arrays. This is used to make sure the right displacement is applied to the
+        # right hypotheses. Channel hypotheses are stored contiguously so we can just
+        # specify the lengths of each channel in an ordered data structure.
```
tristanls, line 346: **issue**: This description is confusing because it describes a `ChannelMapper` and not `self.channel_hypothesis_mapping` dictionary.

**suggestion**: Either remove or update to something that describes this dictionary [as suggested by @nielsleadholm](https://github.com/thousandbrainsproject/tbp.monty/pull/210/files#r1998698279), e.g., `# Dictionary from graph_id to ChannelMapper`.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -877,13 +878,9 @@ def _add_hypotheses_to_hpspace(
...
-        old_num_hypotheses = self.channel_hypothesis_mapping[graph_id]["num_hypotheses"]
-        new_num_hypotheses = old_num_hypotheses + len(new_loc_hypotheses)
-        self.channel_hypothesis_mapping[graph_id][input_channel] = [
-            old_num_hypotheses,
-            new_num_hypotheses,
-        ]
-        self.channel_hypothesis_mapping[graph_id]["num_hypotheses"] = new_num_hypotheses
+        self.channel_hypothesis_mapping[graph_id].add_channel(
+            input_channel, len(new_loc_hypotheses)
+        )
```
tristanls, line 883: **suggestion (style,  non-blocking)**: There are a lot of chained property accesses here. Maybe something like:
```python
channel_mapper = self.channel_hypothesis_mapping[graph_id]
channel_mapper.add_channel(input_channel, len(new_loc_hypothesis))
```

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -962,8 +959,7 @@ def _update_evidence(self, features, displacements, graph_id):
             initial_possible_locations = []
             initial_possible_rotations = []
             initial_evidence = []
-            self.channel_hypothesis_mapping[graph_id] = dict()
-            num_hypotheses = 0
+            self.channel_hypothesis_mapping[graph_id] = ChannelMapper()
```
tristanls, line 962: **suggestion (style, non-blocking)**: To avoid chained property accesses in the loop below on line 974, maybe:
```python
channel_mapper = ChannelMapper()
for ...
    ...
    channel_mapper.add_channel(input_channel, len(channel_evidence))

self.channel_hypothesis_mapping[graph_id] = channel_mapper
```

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1010,7 +1003,7 @@ def _update_evidence(self, features, displacements, graph_id):
                 # hypotheses for them.
                 if (
                     input_channel
-                    not in self.channel_hypothesis_mapping[graph_id].keys()
+                    not in self.channel_hypothesis_mapping[graph_id].channels
```
tristanls, line 1006: **suggestion (style, non-blocking)**: Could pull out channel mapper as a variable before
```python
channel_mapper = self.channel_hypothesis_mapping[graph_id]
```
So then this would be:
```python
if (input_channel not in channel_mapper.channels)
```

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1036,7 +1029,7 @@ def _update_evidence(self, features, displacements, graph_id):
                     # Get the IDs in hypothesis space for this channel
                     channel_start, channel_end = self.channel_hypothesis_mapping[
                         graph_id
-                    ][input_channel]
+                    ].channel_range(input_channel)
```
tristanls, line 1032: **suggestion (style, non-blocking)**: With `channel_mapper` in a variable, this could become:
```python
channel_start, channel_end = channel_mapper.channel_range(input_channel)
```

**note**: The above suggestion makes me think that `range(...)` might be a better name instead of `channel_range(...)`.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -0,0 +1,128 @@
...
+    the sizes of channels in an ordered dictionary allows us to insert or remove
+    channels, as well as dynamically resize them.
+
+    """
+
+    def __init__(self, channel_sizes: Optional[Dict[str, int]] = None) -> None:
+        """Initializes the ChannelMapper with an ordered dictionary of channel sizes.
+
+        Args:
+            channel_sizes (Optional[Dict[str, int]]): Dictionary of {channel_name: size}
```
tristanls, line 32: **nitpick**: Missing period at the end of the sentence.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -0,0 +1,128 @@
...
+            if name == channel_name:
+                return (start, start + size)
+            start += size
+
+    def resize_channel_by(self, channel_name: str, value: int) -> None:
+        """Resizes the channel by a specific amount.
+
+        Args:
+            channel_name (str): The name of the channel.
+            value (int): The value used to modify the channel size.
```
tristanls, line 82: **suggestion (non-blocking)**: Highlight that it can be negative.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -0,0 +1,128 @@
...
+            raise ValueError(f"Channel '{channel_name}' not found.")
+
+        start = 0
+        for name, size in self.channel_sizes.items():
+            if name == channel_name:
+                return (start, start + size)
+            start += size
+
+    def resize_channel_by(self, channel_name: str, value: int) -> None:
+        """Resizes the channel by a specific amount.
```
tristanls, line 78: **suggestion (non-blocking)**: Maybe `Increases or decreases the channel by a specific amount.` to highlight the decrease use case?

tests/unit/channel_mapper_test.py
```diff
@@ -0,0 +1,67 @@
...
+        self.assertEqual(self.mapper.channel_range("B"), (5, 15))
+        self.assertEqual(self.mapper.channel_range("C"), (15, 30))
+        with self.assertRaises(ValueError):
+            self.mapper.channel_range("D")
+
+    def test_resize_channel_by(self):
+        """Test increasing and decreasing channel sizes."""
+        self.mapper.resize_channel_by("B", 5)
+        self.assertEqual(self.mapper.channel_range("B"), (5, 20))
+        self.assertEqual(self.mapper.total_size, 35)
```
tristanls, line 37: **suggestion (non-blocking)**: Add a passing case for resize by a negative value.

tests/unit/channel_mapper_test.py
```diff
@@ -0,0 +1,67 @@
...
+        self.assertEqual(self.mapper.channel_range("D"), (30, 38))
+
+        with self.assertRaises(ValueError):
+            self.mapper.add_channel("A", 3)
+
+    def test_add_channel_at_position(self):
+        """Test inserting a channel at a specific position."""
+        self.mapper.add_channel("X", 7, position=1)
+        self.assertEqual(self.mapper.channels, ["A", "X", "B", "C"])
+        self.assertEqual(self.mapper.channel_range("X"), (5, 12))
```
tristanls, line 55: **suggestion (non-blocking)**: Check ranges of the "moved" channels: "B" and "C".

tristanls (commented):
Thank you for the updates @ramyamounir.

I have one **issue** regarding a comment. Aside from that, other things are **non-blocking**.

src/tbp/monty/frameworks/models/evidence_matching.py
jeremyshoemaker: `range` might get confused for the builtin function of the same name. This _is_ a method though, so it might not be as big of a problem.

ramyamounir:
Thanks for the Review @tristanls! I've pushed updates to fix the issue and applied your non-blocking suggestions.

tristanls (approved):
Thank you for the updates. Looks good!

tristanls:
@ramyamounir, if I understand correctly, I think this is now a backward-compatible/non-breaking code refactor (`refactor:`). I don't think there are any breaking changes here, right?

ramyamounir:
@tristanls Yes, it is should be backward compatible/non-breaking. All the changes are to `channel_hypothesis_mapping` and I do not see it being used or referenced outside of `EvidenceGraphLM` too.

PR #213, opened 2025-03-14
chore: fix for bug where glossary has old description

codeallthethingz:
This PR fixes a bug where the description was not updating from the frontmatter in the document pages.

This was two fold
1. Because in readme.io it is called except, not description.
1. The description was not being read in from the document's frontmatter.

A term in the glossary page was also corrected to move to the right location and become a heading.


tristanls (approved):
🥈

PR #215, opened 2025-03-18
refactor: BufferEncoder allows registering handlers for different types

tristanls:


tristanls:
Never mind this pull request. There's a more nuanced use case where we don't want `GoalState` to be caught by an encoder for `State` via `isinstance`. That is, we need to find the most specific encoder available.

PR #216, opened 2025-03-19
refactor: `BufferEncoder` registration for types and encoders

scottcanoe:
**Summary**

This PR modifies `BufferEncoder` in two main ways:
 - Encoders and types aren't hard-coded in `BufferEncoder`. Types and their associated encoders are registered with the `BufferEncoder` class.
 - The lookup mechanism for encoders respects type hierarchy. This means that if you register an encoder for a subclass, then the encoder registered for its parent class will not be returned. This can happen with `isinstance` checks, the current method for encoder lookup.

This PR is in preparation for an update that allows `GoalState` objects to be serialized, a necessary feature for the Demonstrating Monty's Capabilities paper. For context, `GoalState` is a subclass of `State`, and they'll need different encoders to handle repackaging of their different attributes. This is the reason for the new lookup method.

Unit tests have also been added.


scottcanoe:
Tagging @vkakerbeck, @nielsleadholm, @hlee9212, and @ramyamounir for visibility since this came up in our research meeting.

tristanls:
Adding @jeremyshoemaker to also take a look.

tests/unit/buffer_encoder_test.py
```diff
@@ -0,0 +1,206 @@
...
+
+        # Test bools
+        for val in array.astype(bool):
+            val_string = str(val).lower()
+            self.assertEqual(val_string, json.dumps(val, cls=BufferEncoder))
+
+    def test_scipy_rotation(self):
+        rot = Rotation.from_euler("xyz", [30, 45, 60], degrees=True)
+        rot_array = rot.as_euler("xyz", degrees=True)
+        rot_string = json.dumps(rot_array.tolist())
```
scottcanoe, line 178: **Feedback Request**: I'm not sure its appropriate to test the encoders themselves here since their behavior is not under the purview of the `BufferEncoder` whose job is just to find encoders. These will also fail if the encoders themselves ever change. Would it be more principled to make tests of encoders their own `TestCase`s?

jeremyshoemaker (commented):
This all looks reasonable to me.

One thing I discussed with @tristanls the other day on this topic was how on a codebase I've worked on in the past, we defined a decorator to help with the calls to something like `register` and we wrapped the serialization using `json.dumps` in a helper function that made it so we didn't have to worry about remembering to pass in the encoder.

We weren't using a class based approach at all though for what `BufferEncoder` is doing here, so it's somewhat different, but the advantage of the decorator idea is you can do something like this, where in this example the decorator is called `encodes`:

```python
@encodes(ClassName)
def class_name_encoder(obj):
   # do the transform of `obj` into a json.dumps compatible structure
   ...
```
   
And that `encodes` decorator handles the call to `BufferEncoder.register`(it's been a couple years since I wrote a decorator, or I'd write an example now to show how it would be defined).

That works for the case where the encoder is a function, but we could do something similar with a class decorator that you could apply to JSONEncoder subclasses we create, and have it handle registering the class as it's being defined.

These are all just ways to try to make the process easier to use, and less error prone. Not something that we need to do in this PR.

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -582,33 +583,89 @@ def _fill_old_values_with_nans(self, existing_vals, new_val_len):
 class BufferEncoder(json.JSONEncoder):
     """Encoder to turn the buffer into a JSON compliant format."""
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-        self.action_encoder = ActionJSONEncoder(**kwargs)
+    encoders: Dict[type, Union[Callable, json.JSONEncoder]] = {}
```
tristanls, line 586: **todo (if-true)**: I think this should be `_encoders`. I believe it is intended to be a private class attribute.

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -582,33 +583,89 @@ def _fill_old_values_with_nans(self, existing_vals, new_val_len):
...
+        encoders, then we will always return the encoder registered for `Child`.
+        This is in contrast to lookup via `isinstance` checks in which case an
+        the `Parent` encoder may be returned for a `Child` instance, depending on the
+        order in which their encoders were registered.
+
+        Args:
+            obj: The object for in need of an encoder.
+
+        Returns:
+            An encoder for `obj` if one exists, otherwise `None`.
```
tristanls, line 634: **praise**: Thanks for the attention to detail on these Google Style docstrings, formatting, full sentences, etc. It's nice to read.

tests/unit/buffer_encoder_test.py
tristanls, line 1: **issue (non-blocking)**: Could you put this in `tests/unit/frameworks/models/buffer_encoder_test.py`? You can take a look at the action tests for reference (you'll need `__init__.py`s in the folder hierarchy).

tristanls (approved):
Thank you for this improvement @scottcanoe.

I added one **todo (if-true)** and have many thoughts on the tests. However, I feel it'll be easier for me to approve this PR and then do a follow on PR with test updates and have you review that instead of trying to do it via comments here.

I am approving the PR pending the decision on the **todo (if-true)**.

tests/unit/buffer_encoder_test.py
scottcanoe: Done.

src/tbp/monty/frameworks/models/buffer.py
scottcanoe: Yep, fixed.

scottcanoe:
Thanks for the review, y'all. 

@tristanls: I've made your suggested edits. I'm looking forward to seeing/hearing more about tests and developing better practices.

@jeremyshoemaker: The decorator approach is pretty nice. I think if we were to do that, I'd probably make it a class method of `BufferEncoder` (`BufferEncoder.encodes`).  It's light years from being a priority, but it'd be nice if we had other formats available for logging someday (i.e., HDF5) since these JSON files are ~3-4x larger than they need to be and don't deserialize nicely. In which case, we'd also have `HDF5Encoder.encodes` or something like that.

PR #217, opened 2025-03-20
refactor: Avoid possible locations double initialization.

ramyamounir:
This is a 1 line PR to remove unnecessary initialization of possible locations in `EvidenceGraphLM`.

We should not need to set the `self.possible_locations` using `self.graph_memory.get_initial_hypotheses` in the `EvidenceGraphLM.reset` [function](https://github.com/thousandbrainsproject/tbp.monty/blob/e62e1bca981f7ef3a4e2832e11f80056ce189981/src/tbp/monty/frameworks/models/evidence_matching.py#L366). Whatever we set here at reset will be overwritten later in `_update_evidence` initialization of hypothesis space [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e62e1bca981f7ef3a4e2832e11f80056ce189981/src/tbp/monty/frameworks/models/evidence_matching.py#L987).

It's not a big thing, but it causes `self.possible_locations` and `self.possible_poses` to be out of sync (i.e., locations are initialized and poses are empty) for some part of the code.

*Unit tests and a couple of benchmarks still ran fine.*

### DIFF
--- a/src/tbp/monty/frameworks/models/evidence_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching.py
@@ -1841,24 +1841,6 @@ class EvidenceGraphMemory(GraphMemory):
     # ------------------- Main Algorithm -----------------------
 
     # ------------------ Getters & Setters ---------------------
-    def get_initial_hypotheses(self):
-        possible_matches = self.get_memory_ids()
-        possible_locations = {}
-        for graph_id in possible_matches:
-            # Initialize vertical shape of np array so we can easily stack it. Will be
-            # removed after concatenating loop.
-            all_locations_for_graph = np.zeros(3)
-            for input_channel in self.get_input_channels_in_graph(graph_id):
-                # All locations stored in graph
-                all_locations_for_graph = np.vstack(
-                    [
-                        all_locations_for_graph,
-                        self.get_locations_in_graph(graph_id, input_channel),
-                    ]
-                )
-            possible_locations[graph_id] = all_locations_for_graph[1:]
-        return possible_matches, possible_locations
-
     def get_rotation_features_at_all_nodes(self, graph_id, input_channel):
         """Get rotation features from all N nodes. shape=(N, 3, 3).
 
--- a/src/tbp/monty/frameworks/models/evidence_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching.py
@@ -212,7 +212,7 @@ class EvidenceGraphLM(GraphLM):
             2) Within one object, possible poses are considered possible if their
                 evidence is larger than the most likely pose of this object - x percent
                 of this poses evidence.
-            # TODO: should we use a separate theshold for within and between objects?
+            # TODO: should we use a separate threshold for within and between objects?
             If this value is larger, the model is usually more robust to noise and
             reaches a better performance but also requires a lot more steps to reach a
             terminal condition, especially if there are many similar object in the data
@@ -222,7 +222,7 @@ class EvidenceGraphLM(GraphLM):
         pose_similarity_threshold: difference between two poses to be considered
             unique when checking for the terminal condition (in radians).
         required_symmetry_evidence: number of steps with unchanged possible poses
-            to classify an object as symetric and go into terminal condition.
+            to classify an object as symmetric and go into terminal condition.
 
     Model Attributes:
         graph_delta_thresholds: Thresholds used to compare nodes in the graphs being
@@ -251,7 +251,7 @@ class EvidenceGraphLM(GraphLM):
         use_multithreading: Whether to calculate evidence updates for different
             objects in parallel using multithreading. This can be done since the
             updates to different objects are completely independent of each other. In
-            general it is recommended to use this but it can be usefull to turn it off
+            general it is recommended to use this but it can be useful to turn it off
             for debugging purposes.
     """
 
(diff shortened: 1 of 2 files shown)


ramyamounir:
> @ramyamounir, this looks like a non-breaking refactor for semantic versioning purposes since all the tests pass, yes?

Yes, should be so.

tristanls:
@ramyamounir OK, thank you. I added a https://www.conventionalcommits.org to the pull request name (which is proposed as the default commit message) so that it is easier to remember at a glance what sort of change it is (in the context of semantic versioning).

vkakerbeck (commented):
Thanks for catching this! Can you also delete the `get_initial_hypotheses` function (https://github.com/thousandbrainsproject/tbp.monty/blob/eb5ddeb6d68efaf01d38f026f22a3e64f150cd72/src/tbp/monty/frameworks/models/evidence_matching.py#L1849) so we don't have dead code floating around?

tristanls:
**question**: Since @vkakerbeck pointed out removing `get_initial_hypotheses`, there is a pattern of calling `get_initial_hypotheses` in `graph_matching.py` and `feature_location_matching.py` as well. Should those patterns be evaluated based on this PR changing this pattern?

ramyamounir:
> **question**: Since @vkakerbeck pointed out removing `get_initial_hypotheses`, there is a pattern of calling `get_initial_hypotheses` inside `reset()` calls in `graph_matching.py` and `feature_location_matching.py` as well. Should those patterns be evaluated based on this PR changing this pattern?

Yes, good point. I've seen it in `displacement_matching.py` too, but they have different return variables sometimes. I'm not sure if the same simplification would apply to other graph matching classes than EvidenceGraphLM.

tristanls:
@ramyamounir @vkakerbeck, regarding `get_initial_hypotheses`, it is a method defined in the parent `GraphMemory` class. If it is removed, then the implementation from `GraphMemory` will be used. Is that appropriate `EvidenceGraphMemory`?

ramyamounir:
> @ramyamounir @vkakerbeck, regarding `get_initial_hypotheses`, it is a method defined in the parent `GraphMemory` class. If it is removed in `EvidenceGraphMemory`, then [the implementation from `GraphMemory`](https://github.com/thousandbrainsproject/tbp.monty/blob/eb5ddeb6d68efaf01d38f026f22a3e64f150cd72/src/tbp/monty/frameworks/models/graph_matching.py#L1265) will be used. Is that appropriate `EvidenceGraphMemory`?

I think it should be alright since the `get_initial_hypotheses` is never called from `EvidenceGraphMemory` or `EvidenceGraphLM` now. Seems like we don't need it for the evidence graph matching classes.

vkakerbeck:
Yes, I think it would be too involved removing the function in all the LM classes. The older LMs still rely on it. An alternative could be to move the code that is [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e62e1bca981f7ef3a4e2832e11f80056ce189981/src/tbp/monty/frameworks/models/evidence_matching.py#L962-L992) into the `get_initial_hypotheses` function and then call that function there. I think that could actually be a nice way to do it (but not sure how that will play with the other restructurings of the `update_evidence` function you are already working on Ramy).

nielsleadholm:
What about keeping `get_initial_hypotheses` in the EvidenceLM, but it just returns `possible_matches`? That way line 362 also doesn't reach into `self.graph_memory`'s methods?

ramyamounir:
> An alternative could be to move the code that is [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e62e1bca981f7ef3a4e2832e11f80056ce189981/src/tbp/monty/frameworks/models/evidence_matching.py#L962-L992) into the `get_initial_hypotheses` function and then call that function there.

I agree, we can refactor this part into it's own hypotheses initialization function. I still think that initialization due to `reset()` and initialization due to a missing input channel (without reset) can be combined together into a general "extend hypothesis space" functionality that loops over missing input channels. I will propose this in a future PR, but [PR#210](https://github.com/thousandbrainsproject/tbp.monty/pull/210) is a good reference. I'm leaning towards leaving `_update_evidence` as is for now until a PR for hypotheses resampling is submitted, I believe the function will change quite a bit.


> What about keeping `get_initial_hypotheses` in the EvidenceLM, but it just returns `possible_matches`?

Yea, we could do that if we want to maintain the pattern of calling `get_initial_hypotheses` in all LMs. To maintain the correct pattern maybe we would keep it in `EvidenceGraphMemory` (instead of `EvidenceGraphLM`) as it currently is? Also, I feel that the function name would not be correct now because it will only be returning `possible_matches`. If we decide to

nielsleadholm:
Yeah in terms of using `get_initial_hypotheses`, I think from my mind there is nothing wrong with the scope of initial hypotheses only being the objects that the LM could be on, rather than the locations or poses as well. At the moment, every `get_initial_hypotheses` is different (returning anywhere from 2-4 arguments), so it wouldn't be breaking with precedent that these are bespoke to the hypothesis-space of that type of LM. Curious what you think @vkakerbeck ?

ramyamounir:
Great, thanks!

nielsleadholm (approved):
Looks great, thanks for making this change!

tristanls (approved):

vkakerbeck (approved):
Great, thanks for the update!

PR #218, opened 2025-03-20
chore: update BufferEncoder tests

tristanls:
@scottcanoe, @jeremyshoemaker, here are my thoughts and proposal for `BufferEncoder` test updates.

The following diagram helps to explain the framing that guided my changes.

Screenshot 2025-03-20 at 16 49 26

In general, given an object under test, we want to always A) test messages received from others, B) never test messages sent to self, and C) sometimes test messages sent to others. 

There are two flavors of "messages:" ones that don't cause side effects (_queries_) and ones that do (_commands_).

A) Messages received from others means method calls. As in `BufferEncoder.register(...)`. We want to test the return value of that method to be what is expected. This is anticlimactic as `BufferEncoder.register(...)` returns `None`, so we don't bother. However, `BufferEncoder.register(...)` causes a side-effect, so we need to test the side-effect via `json.dumps(...)` and observe that the side-effect (encoding correctly) is as desired.

B) Messages sent to self are manipulations of the internal state and private message calls. I removed tests inspecting `BufferEncoder._encoders` because that is a private state. We don't test this so that we can change implementation without having to update tests all over the place. Testing B) is usually responsible for test brittleness when any change to the code requires updating tests. The `setUp` method was manipulating the object-under-test's private state, so this highlighted a need for the `unregister(...)` method to operate our object under test appropriately. That's why I added `unregister(...)` and tests for its side effects. This type of design update is what is meant by the phrase "Test Driven Design".

C) Messages sent to others are the objects under test calling methods on other objects. `BufferEncoder` does not do this, so there are no tests of this type here. _If_ there were outgoing messages, I mentioned that we test them sometimes. We don't need to test outgoing _queries_, but we do want to test the side effects of outgoing _commands_ (e.g. writing a file, saving a database record).

A note on why I checked for equality using `json.loads(json.dumps(...` instead of constructing a string to test with is due to the nuance of JSON encoding, in that multiple distinct JSON strings can be equivalent when parsed, e.g., `{"data": 1}` would be the same as `{"data"   :   1}`. This is an extreme example, but just highlighting the point. As a result, I don't want the test to get tripped up on this, and instead rely on the `json.loads` to parse the `json.dumps` result into a unique Python representation.

The "fakes" come from the Mocks, Stubs, and Fakes terminology (there are a zillion articles on this, I always have to look up which one is which).

I separated the fakes into their own place so that the test code resembles the actual object under test use. For example, `BufferEncoder.register(FakeClass, FakeJSONEncoder)` is closer to what code will look like in usage than `BufferEncoder.register(self.dummy_class, self.dummy_encoder)`.


tests/unit/frameworks/models/buffer_encoder_test.py
```diff
@@ -24,182 +23,158 @@
...
-            return dict(data=obj.data)
-
-        self.dummy_class = Dummy
-        self.dummy_subclass_1 = DummySubclass1
-        self.dummy_subclass_2 = DummySubclass2
-        self.dummy_encoder = dummy_encoder
-        self.dummy_subclass_1_encoder = dummy_subclass_1_encoder
-        self.dummy_encoder_class = DummyEncoderClass
-
+        """Ensure BufferEncoder does not contain any encoders created during tests."""
```
jeremyshoemaker, line 45: Wouldn't it be easier to just reach into the BufferEncoder class and set the registry to an empty list? That way if another test class is added to these tests later, we won't accidentally forget to add it here.

tests/unit/frameworks/models/buffer_encoder_test.py
```diff
@@ -24,182 +23,158 @@
...
-
-    def test_register(self):
-        # Test registering a a function.
-        BufferEncoder.register(self.dummy_class, self.dummy_encoder)
-        self.assertEqual(BufferEncoder._encoders[self.dummy_class], self.dummy_encoder)
+    def test_register_function_encoder(self):
+        BufferEncoder.register(FakeClass, fake_encoder)
+        fake = FakeClass(data=0)
+        self.assertEqual(
+            json.loads(json.dumps(fake, cls=BufferEncoder)),
```
jeremyshoemaker, line 60: Somewhat replying to your comments in the description, I was thinking about this a little and wondering if it might make more sense to write these tests to inspect the return value from `BufferEncoder.default` rather than round-tripping through the `json` library. The `json` library should be expected to work (it's tested elsewhere outside our codebase) and so doing this every time we want to check that the output is what we want seems like extra work to me. Maybe testing this round trip behaviour once would make more sense and then use the return values from `default` everywhere else? 

I'm not sure how much I think this is a better approach, so I want to open it up for discussion.

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -608,6 +608,15 @@ def register(
         else:
             raise ValueError(f"Invalid encoder: {encoder}")
 
+    @classmethod
+    def unregister(cls, obj_type: type) -> None:
```
jeremyshoemaker, line 612: Does this exist only for the tests, or do we expect to use this elsewhere? If it's just for the tests, then maybe my suggestion below for how to reset the BufferEncoder class in the tests is a better approach.

jeremyshoemaker (approved):
Looks fine to me overall. I have a few comments, none of them blockers.

scottcanoe (approved):
**praise:** ‏This is an excellent refactor with a fantastic write-up (it's going directly into my notes). The fakes system and a clear understanding of what the tests are responsible for is so much cleaner and more principled. You also included a mixed-inheritance case that I hadn't considered.


**note:** ‏It occurred to me while writing the original tests that being able to register lambdas does mean that we no longer have any reference to the registered encoder, and we end up being on the hook for recreating their full behavior during testing (brittle if those encoders ever change). I can see the value in only registering functions or classes that we can test independently, as you did with actions.

jeremyshoemaker:
> It occurred to me while writing the original tests that being able to register lambdas does mean that we no longer have any reference to the registered encoder, and we end up being on the hook for recreating their full behavior

I agree, which is why I was suggesting the decorator approach. It makes it easier to use stand alone functions for encoders instead of lambdas. Nothing about the current `register` method prevents using functions, but the decorator makes it much easier, and to me at least more Pythonic. But as I said before, an improvement for later. 😄

tests/unit/frameworks/models/buffer_encoder_test.py
jeremyshoemaker: That's a good point. This is fine for now. We can always revisit it if it becomes a maintenance problem.

Also just noticed, this shouldn't be a Python string because in this location it's not a docstring, it should be a regular comment.
```suggestion
        # Ensure BufferEncoder does not contain any encoders created during tests.
```

tests/unit/frameworks/models/buffer_encoder_test.py
tristanls: Yes, I've been thinking about this for a while myself.

My use of `json` library here is not testing the library. I do expect `json` to work correctly, and so I am using it to construct the assertion about the object under test `BufferEncoder`. My argument here would be analogous to one where my use of a `for` loop does not test `for` loops but is using code expected to work to make assertions. Less trivially, if I had a function that produced a prime number, I would create a helper `check_if_prime`, an assertion tool, not the code being tested.

I agree that testing something like `BufferEncoder.default` would be more unit-like. However, those unit tests would have a downside in that we would still not know (the code didn't execute) whether the output of `BufferEncoder.default` serializes to JSON, which is the side-effect we want to occur for every invocation of `BufferEncoder.default`.

I considered moving these tests to `tests/serialization/frameworks/models/buffer_encoder_test.py` instead. My thinking there would be that we probably want to cross the serialization boundary for every serialization code to make sure it works. But that always feels a bit more integration-like and less unit-like.

src/tbp/monty/frameworks/models/buffer.py
tristanls: Yes, this emerged from testing needs.

tests/unit/frameworks/models/buffer_encoder_test.py
tristanls: I incorporated @scottcanoe's suggestion here by segmenting encoder classes into `fakes/encoder_classes.py` and unregistering all of them in https://github.com/thousandbrainsproject/tbp.monty/pull/218/commits/76ec2feedc14baf6f71e96a7b58dc2300b85f5df.

jeremyshoemaker: I like it!

src/tbp/monty/frameworks/models/buffer.py
jeremyshoemaker: I'm a little suspicious of adding functionality just for the sake of tests, but it's probably fine here.

tristanls:
Thank you for the review and comments. Since we're discussing things in a non-blocking manner, I will merge the current code, but we can continue the discussion in the comment after the merge as long as we need to inform any follow-up PRs.

src/tbp/monty/frameworks/models/buffer.py
jeremyshoemaker: I think the approach of having it shared on the class makes it more flexible, and to me that's worth more than a little discomfort around an extra method for the tests.

PR #220, opened 2025-03-21
refactor!: MontyExperiment context manager part 2

jeremyshoemaker:
This is the second phase of [the work I proposed on Discourse][forum] to make MontyExperiment into a context manager.

This moves the config setup into a constructor for the class, leaving the part that initializes the model and datasets in the `setup_experiment` method. Then we call the simplified setup method in the context manager `__enter__` method.

The rest of this is updating the call sites to use the new context manager style more completely.

[forum]: https://thousandbrains.discourse.group/t/investigation-into-linux-test-failures-and-possible-improvements/468

### DIFF
--- a/src/tbp/monty/frameworks/experiments/monty_experiment.py
+++ b/src/tbp/monty/frameworks/experiments/monty_experiment.py
@@ -51,19 +51,25 @@ class MontyExperiment:
     the outermost loops for training and evaluating (including run epoch and episode)
     """
 
+    def __init__(self, config):
+        """Initialize the experiment based on the provide configuration.
+
+        Args:
+            config: config specifying variables of the experiment.
+        """
+        # Copy the config and store it so we can modify it freely
+        config = copy.deepcopy(config)
+        config = config_to_dict(config)
+        self.config = config
+
+        self.unpack_experiment_args(config["experiment_args"])
+
     def setup_experiment(self, config):
         """Set up the basic elements of a Monty experiment and initialize counters.
 
         Args:
             config: config specifying variables of the experiment.
         """
-        # Save a copy of the config used to specify the experiment before modifying
-        config = copy.deepcopy(config)
-        # Convert any dataclass back to dict for backward compatibility
-        config = config_to_dict(config)
-        self.config = config
-
-        self.unpack_experiment_args(config["experiment_args"])
         self.model = self.init_model(
             monty_config=config["monty_config"],
             model_path=self.model_path,
@@ -530,9 +536,6 @@ class MontyExperiment:
             self.run_epoch()
         self.logger_handler.post_train(self.logger_args)
 
-        if not self.do_eval:
-            self.close()
-
     def evaluate(self):
         """Run n_eval_epochs."""
         # TODO: check that number of eval epochs is at least as many as length
@@ -542,7 +545,6 @@ class MontyExperiment:
         for _ in range(self.n_eval_epochs):
             self.run_epoch()
         self.logger_handler.post_eval(self.logger_args)
-        self.close()
 
     def state_dict(self):
         """Return state_dict with total steps."""
@@ -612,7 +614,7 @@ class MontyExperiment:
         Returns:
             MontyExperiment self to allow assignment in a with statement.
         """
-        # TODO: Move some of the initialization code from `setup_experiment` into this.
+        self.setup_experiment(self.config)
         return self
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
@@ -623,7 +625,5 @@ class MontyExperiment:
         Returns:
             bool to indicate whether to supress any exceptions that were raised.
         """
-        # TODO: We call self.close inside `train` and `evaluate`.
-        #   Those should probably be removed.
         self.close()
         return False  # don't silence exceptions inside the with block
--- a/src/tbp/monty/frameworks/experiments/monty_experiment.py
+++ b/src/tbp/monty/frameworks/experiments/monty_experiment.py
@@ -52,7 +52,7 @@ class MontyExperiment:
     """
 
     def __init__(self, config):
-        """Initialize the experiment based on the provide configuration.
+        """Initialize the experiment based on the provided configuration.
 
         Args:
             config: config specifying variables of the experiment.
@@ -99,7 +99,7 @@ class MontyExperiment:
         """Initialize the Monty model.
 
         Args:
-            monty_config: confguration for the Monty class.
+            monty_config: configuration for the Monty class.
             model_path: Optional model checkpoint. Can be full file name or just the
                 directory containing the "model.pt" file saved from a previous run.
 
@@ -221,7 +221,7 @@ class MontyExperiment:
         Raises:
             TypeError: If `dataset_class` is not a subclass of `EnvironmentDataset`
         """
-        # Require dataset_class to be EnvironmentDataset now, generalzie later
+        # Require dataset_class to be EnvironmentDataset now, generalize later
         if not issubclass(dataset_class, EnvironmentDataset):
             raise TypeError("dataset class must be EnvironmentDataset (for now)")
 
--- a/tests/unit/base_config_test.py
+++ b/tests/unit/base_config_test.py
@@ -79,17 +79,14 @@ class BaseConfigTest(unittest.TestCase):
         """
         pprint("...parsing experiment...")
         base_config = copy.deepcopy(self.base_config)
-        self.exp = MontyExperiment()
-        with self.exp:
-            self.exp.setup_experiment(base_config)
+        with MontyExperiment(base_config) as self.exp:
+            pass
 
     # @unittest.skip("debugging")
     def test_can_run_episode(self):
         pprint("...parsing experiment...")
         base_config = copy.deepcopy(self.base_config)
-        self.exp = MontyExperiment()
-        with self.exp:
-            self.exp.setup_experiment(base_config)
+        with MontyExperiment(base_config) as self.exp:
             pprint("...training...")
             self.exp.model.set_experiment_mode("train")
             self.exp.dataloader = self.exp.train_dataloader
@@ -99,9 +96,7 @@ class BaseConfigTest(unittest.TestCase):
     def test_can_run_train_epoch(self):
         pprint("...parsing experiment...")
         base_config = copy.deepcopy(self.base_config)
-        self.exp = MontyExperiment()
-        with self.exp:
-            self.exp.setup_experiment(base_config)
+        with MontyExperiment(base_config) as self.exp:
             self.exp.model.set_experiment_mode("train")
             self.exp.run_epoch()
 
@@ -109,9 +104,7 @@ class BaseConfigTest(unittest.TestCase):
     def test_can_run_eval_epoch(self):
         pprint("...parsing experiment...")
         base_config = copy.deepcopy(self.base_config)
-        self.exp = MontyExperiment()
-        with self.exp:
-            self.exp.setup_experiment(base_config)
+        with MontyExperiment(base_config) as self.exp:
             self.exp.model.set_experiment_mode("eval")
             self.exp.run_epoch()
 
@@ -120,10 +113,7 @@ class BaseConfigTest(unittest.TestCase):
         """Make sure this test uses very small n_actions_per_epoch for speed."""
         pprint("...parsing experiment...")
         base_config = copy.deepcopy(self.base_config)
-        self.exp = MontyExperiment()
-        with self.exp:
-            self.exp.setup_experiment(base_config)
-
+        with MontyExperiment(base_config) as self.exp:
             monty_module_sids = set(
                 [s.sensor_module_id for s in self.exp.model.sensor_modules]
             )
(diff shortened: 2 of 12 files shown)


src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -51,19 +51,25 @@ class MontyExperiment:
     the outermost loops for training and evaluating (including run epoch and episode)
     """
 
-    def setup_experiment(self, config):
-        """Set up the basic elements of a Monty experiment and initialize counters.
+    def __init__(self, config):
+        """Initialize the experiment based on the provide configuration.
```
tristanls, line 55: **nitpick (typo)**: "provided"

tests/unit/base_config_test.py
```diff
@@ -79,17 +79,14 @@ def test_can_set_up(self):
         """
         pprint("...parsing experiment...")
         base_config = copy.deepcopy(self.base_config)
-        self.exp = MontyExperiment()
-        with self.exp:
-            self.exp.setup_experiment(base_config)
+        with MontyExperiment(base_config) as self.exp:
```
tristanls, line 82: **suggestion (non-blocking)**: I don't see a need to attach `exp` to `self` in these tests. `setUp` and `tearDown` do not interact with `self.exp`. I think here and in all other tests (in this file and all the others) this can be:
```python
with MontyExperiment(base_config) as exp:
```

tests/unit/evidence_lm_test.py
```diff
@@ -1702,9 +1683,7 @@ def test_5lm_evidence_experiment(self):
         """Test 5 evidence LMs voting with two evaluation settings."""
         pprint("...parsing experiment...")
         config = copy.deepcopy(self.evidence_5lm_config)
-        self.exp = MontyObjectRecognitionExperiment()
-        with self.exp:
-            self.exp.setup_experiment(config)
+        with MontyObjectRecognitionExperiment(config) as self.exp:
```
tristanls, line 1686: **issue**: Comment on line 1718/1697 below is outdated.
```
 # not running self.exp.evaluate here to avoid exp.close being called.
```

tests/unit/graph_learning_test.py
```diff
@@ -547,16 +547,13 @@ def test_can_set_up(self):
         """
         pprint("...parsing experiment...")
         base_config = copy.deepcopy(self.base_config)
-        self.exp = MontyObjectRecognitionExperiment()
-        with self.exp:
-            self.exp.setup_experiment(base_config)
+        with MontyObjectRecognitionExperiment(base_config) as self.exp:
+            pass
```
tristanls, line 551: **issue**: I think the test should be renamed to something like `test_can_initialize` and the docstring updated accordingly.

tests/unit/graph_learning_test.py
```diff
@@ -689,9 +674,7 @@ def test_fixed_actions_disp(self):
         """
         pprint("...parsing experiment...")
         config = copy.deepcopy(self.fixed_actions_disp)
-        self.exp = MontyObjectRecognitionExperiment()
-        with self.exp:
-            self.exp.setup_experiment(config)
+        with MontyObjectRecognitionExperiment(config) as self.exp:
             # self.exp.model.set_experiment_mode("eval")
```
tristanls, line 678: **issue (non-blocking)**: Remove commented out code.

tests/unit/graph_learning_test.py
```diff
@@ -713,9 +696,7 @@ def test_fixed_actions_ppf(self):
         """Like test_fixed_actions_disp but using point pair features for matching."""
         pprint("...parsing experiment...")
         config = copy.deepcopy(self.fixed_actions_ppf)
-        self.exp = MontyObjectRecognitionExperiment()
-        with self.exp:
-            self.exp.setup_experiment(config)
+        with MontyObjectRecognitionExperiment(config) as self.exp:
             # self.exp.model.set_experiment_mode("eval")
```
tristanls, line 700: **issue (non-blocking)**: Remove commented out code.

tristanls (commented):
Thank you for the follow-up. This is a breaking change since how someone would invoke an experiment has changed. It is perfectly fine, but let's be explicit about it.

**issue**: This is a breaking change in that an experiment now must be initialized with a `config` as an initialization parameter. Please update the commit and PR prefix to `refactor!:` to indicate the breaking change.

**issue**: `ProfileExperimentMixin` calls `super().setup_experiment(config)` inside its `setup_experiment(self, config)` method and invokes `super().close()` directly. Does that code still work?

tests/unit/base_config_test.py
jeremyshoemaker: I didn't think so either, but I was trying to make the minimal changes necessary. I'll go ahead and change all of them to not stash it on the test instance.

jeremyshoemaker:
> Thank you for the follow-up. This is a breaking change since how someone would invoke an experiment has changed. It is perfectly fine, but let's be explicit about it.
>
> **issue**: This is a breaking change in that an experiment now must be initialized with a `config` as an initialization parameter. Please update the commit and PR prefix to `refactor!:` to indicate the breaking change.

Will do. 

> **issue**: `ProfileExperimentMixin` calls `super().setup_experiment(config)` inside its `setup_experiment(self, config)` method and invokes `super().close()` directly. Does that code still work?

Good catch, I'm not sure how I missed this call site. Looking at what the code does, I'm pretty sure it will still work, except that we won't be profiling all the config setting work that I moved to the MontyExperiment constructor. It shouldn't be much from a performance perspective, and I think the profiling is really trying to capture how long it takes to setup the model and load the datasets.

All that said, we don't appear to have any test coverage for this mixin, so it might make sense for me to add some to test that this change doesn't break anything and to make sure it's covered in the future.

src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -51,19 +51,25 @@ class MontyExperiment:
...
 
         Args:
             config: config specifying variables of the experiment.
         """
-        # Save a copy of the config used to specify the experiment before modifying
+        # Copy the config and store it so we can modify it freely
         config = copy.deepcopy(config)
-        # Convert any dataclass back to dict for backward compatibility
```
vkakerbeck, line 62: Just curious why you removed this comment. Is it not accurate?

vkakerbeck (approved):
Looks good, thanks for the update!

src/tbp/monty/frameworks/experiments/monty_experiment.py
jeremyshoemaker: I thought it was redundant because it's basically summarizing the docstring of `config_to_dict`. 

The other comment above got rewritten because I originally copied the code up from `setup_experiment` without the comment and wrote a new one. I didn't realize I ended up just rewriting the original comment until I saw the diff. I think my version _might_ be explaining the "why" of the copying better than the original, but I could just as easily keep the original comment. 🤷

nielsleadholm (approved):
Just had a quick look through but overall looks good from my end, thanks for making this change.

jeremyshoemaker:
For posterity, I added unit tests for the **ProfileExperimentMixin** in #222. I'll pull those changes in here and make sure everything still works as expected.

src/tbp/monty/frameworks/experiments/monty_experiment.py
vkakerbeck: okay yeah makes sense. I'm happy with how it is. I was just curious if something changed in the functionality that I missed.

jeremyshoemaker:
Rebased onto `main` to pull in the work I did in #222, and addressed PR feedback. Most of it is just changing `self.exp` to `exp` in the tests since we don't need to be saving it on the test case objects.

tests/unit/graph_building_test.py
```diff
@@ -224,23 +224,22 @@ def check_graph_formatting(self, graph, features_to_check):
...
-            self.exp.setup_experiment(self.supervised_pre_training_in_habitat)
-            self.exp.model.set_experiment_mode("train")
+        config = self.supervised_pre_training_in_habitat
+        with MontySupervisedObjectPretrainingExperiment(config) as exp:
+            exp.model.set_experiment_mode("train")
 
             pprint("...training...")
-            self.exp.train()
+            exp.train()
+        return exp
```
tristanls, line 233: **note**: TIL that `with` does not create a scope, huh.

tests/unit/graph_learning_test.py
```diff
@@ -689,98 +674,86 @@ def test_fixed_actions_disp(self):
         """
         pprint("...parsing experiment...")
         config = copy.deepcopy(self.fixed_actions_disp)
-        self.exp = MontyObjectRecognitionExperiment()
-        with self.exp:
-            self.exp.setup_experiment(config)
-            # self.exp.model.set_experiment_mode("eval")
+        with MontyObjectRecognitionExperiment(config) as exp:
+            # exp.model.set_experiment_mode("eval")
```
tristanls, line 678: **issue (non-blocking)**: Remove commented out code.

tests/unit/graph_learning_test.py
```diff
@@ -689,98 +674,86 @@ def test_fixed_actions_disp(self):
...
     def test_fixed_actions_ppf(self):
         """Like test_fixed_actions_disp but using point pair features for matching."""
         pprint("...parsing experiment...")
         config = copy.deepcopy(self.fixed_actions_ppf)
-        self.exp = MontyObjectRecognitionExperiment()
-        with self.exp:
-            self.exp.setup_experiment(config)
-            # self.exp.model.set_experiment_mode("eval")
+        with MontyObjectRecognitionExperiment(config) as exp:
+            # exp.model.set_experiment_mode("eval")
```
tristanls, line 700: **issue (non-blocking)**: Remove commented out code.

tests/unit/graph_learning_test.py
```diff
@@ -689,98 +674,86 @@ def test_fixed_actions_disp(self):
...
         eval_cfg_1 = copy.deepcopy(config)
         eval_cfg_1["experiment_args"].model_name_or_path = os.path.join(
-            self.exp.output_dir,
+            exp.output_dir,
             "2",  # latest checkpoint
         )
-        self.eval_exp_1 = MontyObjectRecognitionExperiment()
-        with self.eval_exp_1:
-            self.eval_exp_1.setup_experiment(eval_cfg_1)
+        with MontyObjectRecognitionExperiment(eval_cfg_1) as self.eval_exp_1:
```
tristanls, line 755: **question**: Is there a need for `eval_exp_1` to be attached to `self` here?

It looks like you didn't remove `self.` from anything with the `self.eval_exp_` pattern here and below.

tests/unit/graph_learning_test.py
```diff
@@ -1131,30 +1078,28 @@ def test_time_out(self):
...
         config = copy.deepcopy(self.feature_pred_tests_time_out)
-        self.exp = MontyObjectRecognitionExperiment()
-        with self.exp:
-            self.exp.setup_experiment(config)
-            self.exp.model.set_experiment_mode("train")
+        with MontyObjectRecognitionExperiment(config) as exp:
+            exp.model.set_experiment_mode("train")
             pprint("...training...")
-            # self.exp.train()
+            # exp.train()
```
tristanls, line 1084: **issue**: Remove commented-out code.

tests/unit/run_parallel_test.py
```diff
@@ -241,10 +240,7 @@ def test_parallel_runs_n_epochs_lt(self):
 
         # In serial like normal
         pprint("...Setting up serial experiment...")
-        self.eval_exp = MontyObjectRecognitionExperiment()
-        with self.eval_exp:
-            self.eval_exp.setup_experiment(self.eval_config)
-
+        with MontyObjectRecognitionExperiment(self.eval_config) as self.eval_exp:
```
tristanls, line 243: **question**: Is there a need to keep `self.eval_exp` here?

tests/unit/run_parallel_test.py
```diff
@@ -292,10 +288,7 @@ def test_parallel_runs_n_epochs_lt(self):
 
         # In serial like normal
         pprint("...Setting up serial experiment...")
-        self.eval_exp_lt = MontyObjectRecognitionExperiment()
-        with self.eval_exp_lt:
-            self.eval_exp_lt.setup_experiment(self.eval_config_lt)
-
+        with MontyObjectRecognitionExperiment(self.eval_config_lt) as self.eval_exp_lt:
```
tristanls, line 291: **question**: Is there a need to keep `self.eval_exp_lt` here?

src/tbp/monty/frameworks/experiments/monty_experiment.py
tristanls: **nitpick (typo)**: "provided" (I wouldn't hold up a merge because of this, but since I left other comments, just highlighting this typo again).

tristanls (commented):
Thank you for the improvements. Readability is definitely improved as a result.

I think there are still a few omissions to address.

src/tbp/monty/frameworks/experiments/monty_experiment.py
jeremyshoemaker: That's weird, I thought I changed that.

tristanls (approved):
Great, thank you for the remaining updates!

tristanls:
NOTE: I force-pushed as the result of updating the branch via a rebase.

Issue #214, opened 2025-03-16
RFC 9 Hypotheses Resampling Implementation


ramyamounir:
This issue tracks the implementation of [RFC 9 Hypotheses Resampling](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0009_hypotheses_resampling.md).

PR #221, opened 2025-03-21
feat: YCB testbed for removing reset between episodes

ramyamounir:
This PR falls under the "Setup YCB testbed that removes 1 episode-per-object assumption during inference" objective.

A new Monty class that overwrites the `pre_episode` function is proposed. Using this class disables resetting the LMs and their buffers between episodes. A detailed write-up on the problems associated with disabling the reset signal is provided in the [RFC for Hypotheses Resampling](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0009_hypotheses_resampling.md). This PR (and follow-up PRs) will become a benchmark for hypotheses resampling.

New configs are provided as examples for how to use this new class. The configs are set up with the goal of fast prototyping and testing (i.e., less objects, less rotations, no noise, etc.)

Follow-up PRs will introduce logging, plotting and accuracy metrics.

A couple of basic unit tests are added with the goal of testing the buffer reset between episodes.


benchmarks/configs/load.py
```diff
@@ -76,12 +77,13 @@ def select_config(experiment: str) -> dict:
...
+    unsupervised_inference_experiment_names = [
+        field.name for field in fields(UnsupervisedInferenceExperiments)
+    ]
     my_experiment_names = [field.name for field in fields(MyExperiments)]
 
     if experiment in monty_world_experiment_names:
-        from benchmarks.configs.monty_world_experiments import (
-            CONFIGS as MONTY_WORLD,
-        )
+        from benchmarks.configs.monty_world_experiments import CONFIGS as MONTY_WORLD
```
vkakerbeck, line 86: What's the reason these formatting changes are added here?

benchmarks/configs/unsupervised_inference_experiments.py
```diff
@@ -0,0 +1,199 @@
...
+    SurfaceViewFinderMountHabitatDatasetArgs,
+)
+
+"""
+The configs provided in this file are used for unsupervised inference experiments.
+
+These experiments are simpler versions of the benchmark experiments designed for fast
+prototyping and testing. More specifically, they are configured for less number of
+rotations, less number of objects, and no noise. These configs call
+`MontyForUnsupervisedEvidenceGraphMatching`, which removes explicit reset logic.
```
vkakerbeck, line 50: Since someone else reading this may not be aware of what this means, maybe you could add a bit more about what the reset usually does and what the implications are if it is removed.

benchmarks/configs/unsupervised_inference_experiments.py
```diff
@@ -0,0 +1,199 @@
...
+    dataset_class=ED.EnvironmentDataset,
+    dataset_args=PatchViewFinderMountHabitatDatasetArgs(),
+    train_dataloader_class=ED.InformedEnvironmentDataLoader,
+    train_dataloader_args=get_env_dataloader_per_object_by_idx(start=0, stop=10),
+    eval_dataloader_class=ED.InformedEnvironmentDataLoader,
+    eval_dataloader_args=EnvironmentDataloaderPerObjectArgs(
+        object_names=objects_list,
+        object_init_sampler=PredefinedObjectInitializer(rotations=test_rotations),
+    ),
+)
```
vkakerbeck, line 192: I know that these experiments are currently just for quick experimentation (just testing one rotation and 2 objects) but I think if this becomes a benchmark experiment, the way you define those is a bit error prone. What I would do is import a ycb benchmark config, deepcopy it, and only update the `monty_class`. That way it is immediately clear what variable is varied and you can easily compare results between reset and no reset. The way you do it here I would be super paranoid that some other setting are accidentally different as well.

src/tbp/monty/frameworks/models/evidence_unsupervised_inference_matching.py
```diff
@@ -0,0 +1,50 @@
...
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+
+from tbp.monty.frameworks.models.evidence_matching import MontyForEvidenceGraphMatching
+
+
+class MontyForUnsupervisedEvidenceGraphMatching(MontyForEvidenceGraphMatching):
+    """This class removes explicit reset logic in unsupervised inference experiments."""
```
vkakerbeck, line 14: Similarly here, I would expand a bit on this docstring. Why does removing the reset make the experiment unsupervised?

src/tbp/monty/frameworks/models/evidence_unsupervised_inference_matching.py
```diff
@@ -0,0 +1,50 @@
...
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+
+from tbp.monty.frameworks.models.evidence_matching import MontyForEvidenceGraphMatching
+
+
+class MontyForUnsupervisedEvidenceGraphMatching(MontyForEvidenceGraphMatching):
```
vkakerbeck, line 13: The name of the class could be easily misunderstood as being an unsupervised learning experiment. How about something like `MontyWithoutReset` or `MontyWithoutEpisodeBorders`?

src/tbp/monty/frameworks/models/evidence_unsupervised_inference_matching.py
```diff
@@ -0,0 +1,50 @@
...
+    """This class removes explicit reset logic in unsupervised inference experiments."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # This keeps track of whether the first `pre_episode` is called.
+        # The experiment will not run unless the first `pre_episode` function
+        # is called, which suggests that some class initialization code is
+        # inside `pre_episode`. Initialization and resetting code should be separate.
+        # TODO: Remove initialization logic from `pre_episode`
```
vkakerbeck, line 23: I'm not sure I would phrase it this way. `pre_episode` can contain any code that should be executed before an episode. For all our experiments so far, this was both initializing and resetting. For this Monty class we just want to initialize at the beginning of the experiment and do no initialization or resetting in `pre_episode`. It seems a bit messy to put this code into the `pre_episode` call and then do the if statement. Why not put initialization into `pre_epoch` or `pre_eval` or even just into Monty's `init` function for this custom class? Then overwrite `pre_episode` to only update the terminal state and target for logging.

tests/unit/unsupervised_inference_evidence_lm_test.py
```diff
@@ -0,0 +1,245 @@
...
+    MontyForUnsupervisedEvidenceGraphMatching,
+)
+from tbp.monty.simulators.habitat.configs import (
+    EnvInitArgsPatchViewMount,
+    PatchViewFinderMountHabitatDatasetArgs,
+)
+from tests.unit.resources.unit_test_utils import BaseGraphTestCases
+
+
+class UnsInfEvidenceLMTest(BaseGraphTestCases.BaseGraphTest):
```
vkakerbeck, line 46: I would write this out. It's not obvious what the abbreviation means and there is no harm in having a longer class name

vkakerbeck (commented):
Thanks for putting together this first version of the new benchmark experiments. This will be useful as we iterate on solutions for this problem. 

I am a bit divided on whether to put this in the benchmarks folder already. On the one hand, this looks more like an early testbed that is more akin to a config in `monty_lab`. You also didn't add results of the two new experiments into the benchmarks table (making a new section for this in the docs). On the other hand I understand that you don't want these experiments to run super long yet and want to be able to quickly iterate. My proposed solution:
Update the configs here to inherit from one of the ycb 10 objects experiments. Run them and add a table to the benchmark results with the current performance (likely very bad). When you experiment locally, you simply overwrite the number of epochs and objects locally. When you make a new PR (presumably to improve performance on these new benchmarks) you rerun the experiment with 10 objects and all rotations and update the benchmarks results table. This way we can already keep track of the impact that changes have on this benchmark. 
What do you think?

ramyamounir:
Thanks for the review @vkakerbeck, I agree with you that this doesn't look like benchmark configs. I didn't really think of those configs as benchmarks but rather simpler experiments to quickly iterate over changes in hypotheses resampling. Perhaps `monty_labs` is a better fit for these configs. I want to point out that this PR is one of 3 PRs in the pipeline, as outlined in the YCB Testbed epic stories:
1) **PR removing resetting of LM state when new object appears**: Addressed by this PR.
2) **PR adding logging and plotting code for theoretical-limit measure**: Follow-up to log and plot things like stepwise pose error of mlh and target object, theoretical limit over time, etc.
3) **PR adding accuracy measure for episodes without matching condition**: Here I plan to do proper benchmark experiments and report accuracy metrics. The accuracy calculations here would be more than simple matching, but perhaps based on whether the mlh is correct after fixed amount of steps for each episode, how much evidence has accumulated on the theoretical limit hypotheses, or the quality of the theoretical limit compared to how many hypotheses exist (i.e., we want better theoretical limit at low number of hypotheses). These metrics are meant for benchmarking progress on [RFC 9](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0009_hypotheses_resampling.md). Additionally, I would be reporting the regular matching accuracy which will be very bad at the moment (perhaps 0). 

I can update this PR by inheriting and overwriting the benchmark experiments like you said and just report the current matching accuracy, then create a table for it in the docs reporting this expected low accuracy. I was just waiting on this until the third PR to report more meaningful accuracy measures.

benchmarks/configs/unsupervised_inference_experiments.py
ramyamounir: Interesting, I see your point. I always felt nervous about inheriting configs and overwriting them, especially if these configs already go through multiple layers of inheritance and have been changed multiple times. I always feel worried that at one of these levels something changed and I'm not aware of it. My concern is that with more inheritance layers, others would find understanding the configs very difficult and that it's tedious to track back to the base config. But it makes sense that for the new benchmark configs we would want to only define the changes the way you explain it so that we are sure all the other details are similar to the current benchmarks.

I know I brought this up last summer, but I still suggest we follow a similar config structure as done here in [MMDet](https://github.com/OpenGVLab/InternImage/tree/master/detection/configs). We would have a single folder of base configs and then additional folders (with nice readme.md files) for other types of configs ("benchmarks", "montymeetsworld", "omniglot", etc.) that inherit and modify base configs. Each experiment is defined in it's own file and experiments don't inherit from each other, they inherit only from base configs. This may not be a research priority at the moment, but I think it's a big obstacle for others trying to use Monty for the first time; configs are the entry point. Curious about your thoughts @tristanls and @jeremyshoemaker.

benchmarks/configs/unsupervised_inference_experiments.py
jeremyshoemaker: You could also replace "less number of" with "fewer" and I think it would read better.

benchmarks/configs/unsupervised_inference_experiments.py
```diff
@@ -0,0 +1,199 @@
...
+These experiments are simpler versions of the benchmark experiments designed for fast
+prototyping and testing. More specifically, they are configured for less number of
+rotations, less number of objects, and no noise. These configs call
+`MontyForUnsupervisedEvidenceGraphMatching`, which removes explicit reset logic.
+"""
+
+# === Configs parameters section === #
+
+# Changes the number of rotations per object
+num_rotations = 1
```
jeremyshoemaker, line 56: These look like constants to me, are they ever going to be changed by the code below? If not, then stylistically they should be all caps to call attention to that fact.

src/tbp/monty/frameworks/models/evidence_unsupervised_inference_matching.py
jeremyshoemaker: > I think these should be defined in the `init` function and reset in `pre_episode`.

I agree, and it's something I've noticed all over the codebase especially since PyCharm by default warns about all instance variables that aren't first defined in an `__init__` method somewhere in the inheritance tree. I've even seen a few places where an object of one class reaches into an object of another class and adds instance variables that didn't exist before (or at least PyCharm thinks so, it's usually pretty accurate). It's a recipe for trouble, but I don't think it's high on our priorities as far as I know.

benchmarks/configs/unsupervised_inference_experiments.py
nielsleadholm: Yeah for what it's worth, the benchmark experiments almost never change. So while I agree with your concerns under more fluid instances, I think in general it is safe to inherit from the benchmark experiments. When they do change (e.g. update when Scott removed the semantic sensor), it is something that is usually discussed with the whole team. This is something we can maybe emphasize to ensure everyone is aware of it.

vkakerbeck: I am worried about exactly the opposite. If something does change in the other benchmark experiments and I am not inheriting from them, then the two experiments are not comparable anymore (i.e. I am not isolating the factor that I want to test for anymore). To me it is the desired behavior that if something in the other config changes, this one changes as well.

src/tbp/monty/frameworks/models/evidence_unsupervised_inference_matching.py
vkakerbeck: Thanks for the explanation Ramy. Like we discussed in our meeting on Monday, I think in order for this not to become it's whole own project you can just update the comment to include this context for when we look into tackling this TODO some time in the future.

ramyamounir:
A quick update on this PR's status. After setting up the full benchmark experiments (random rotations of 10 objects with noise), I realized we would have a memory problem running all these episodes back to back without buffer reset (as @nielsleadholm predicted). A manual reset of the buffer between episodes would not work out of the box because of how `EvidenceGraphLM` depends on the [buffer size](https://github.com/thousandbrainsproject/tbp.monty/blob/b70a23e7a20943ebefd189577dd36d43114f111e/src/tbp/monty/frameworks/models/graph_matching.py#L652) to reset the hypothesis space and [calculate displacement](https://github.com/thousandbrainsproject/tbp.monty/blob/b70a23e7a20943ebefd189577dd36d43114f111e/src/tbp/monty/frameworks/models/graph_matching.py#L1059) with the last location. The SM also keeps its [own buffer](https://github.com/thousandbrainsproject/tbp.monty/blob/b70a23e7a20943ebefd189577dd36d43114f111e/src/tbp/monty/frameworks/models/sensor_modules.py#L108) (e.g., raw_observations) that gets reset in `pre_episode`. I will be discussing a couple of different solutions with Niels and updating this PR accordingly.

vkakerbeck:
One quick thing to double-check could be that you set `save_raw_obs` to `False` (in the `sensor_module_args`). I would assume that this is using up most of the memory.

benchmarks/configs/load.py
ramyamounir: I'm not really sure, Ruff did it automatically when I ran it on the file, line width should still be under 88 to pass the style test. If this is a problem I could look into it, but I think writing the import on one line looks better if it fits in under 88 characters.

nielsleadholm (commented):
Looks great, thanks for implementing this, including the nice documentation and unit test. Just added a few minor suggestions.

benchmarks/configs/load.py
vkakerbeck: Yeah the new format is not an issue itself, it's more just that it introduces a code change into this PR that is unrelated to the PRs purpose (and makes reviewing harder). I'm not sure which editor you are using but for example in VScode there is an option to setup the autoformatter so that it only formats lines that you changes (given that you have version control set up). No need to reverse this change here but this  might be a useful thing to set up in general.

benchmarks/configs/unsupervised_inference_experiments.py
```diff
@@ -0,0 +1,242 @@
...
+each episode via a reset signal. This includes resetting evidence scores, hypothesis
+space, and internal counters. In this unsupervised inference setup, that reset signal
+is removed — allowing us to simulate real-world dynamics where object boundaries are
+not clearly marked.
+
+Here are some key characteristics of the available configs:
+    - **Evaluation-only**: No learning or graph updates occur during these runs.
+        Pre-trained object models are loaded from model_path_10distinctobj before
+        the experiment begins.
+    - **Matching phase only**: The defined Monty class runs in matching mode, without
```
vkakerbeck, line 61: What's the difference between this and the evaluation-only condition? Does it still update graphs, but just with observations seen during matching? Why do we test this?

benchmarks/configs/unsupervised_inference_experiments.py
```diff
@@ -0,0 +1,242 @@
...
+unsupervised_inference_distinctobj_surf_agent = copy.deepcopy(
+    randrot_noise_10distinctobj_surf_agent
+)
+unsupervised_inference_distinctobj_surf_agent["logging_config"].wandb_handlers = []
+
+# distant agent benchmarks configs
+unsupervised_inference_distinctobj_dist_agent = copy.deepcopy(
+    randrot_noise_10distinctobj_dist_agent
+)
+unsupervised_inference_distinctobj_dist_agent["logging_config"].wandb_handlers = []
```
vkakerbeck, line 84: Why are you removing the wandb handlers?

docs/overview/benchmark-experiments.md
```diff
@@ -94,6 +94,20 @@ An object is classified as detected correctly if the detected object ID is in th
...
+
+More specifically, these experiments are run purely in evaluation mode (i.e., pre-trained object graphs are loaded before the experiment begins) with no training or graph updates taking place. Monty stays in the matching phase, continuously updating its internal hypotheses based on sensory observations. For each object, the model performs a fixed number of matching steps before the object is swapped. At the end of each segment, we evaluate whether Monty’s most likely hypothesis correctly identifies the current object. All experiments are performed on 10 distinct objects from the YCB dataset and 10 random rotations for each object. Random noise is added to sensory observations.
+
+## Results
+
+!table[../../benchmarks/results/ycb_unsupervised_inference.csv]
+
+*Note: These benchmark experiments track the progress on [RFC 9: Hypotheses resampling](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0009_hypotheses_resampling.md)*
```
vkakerbeck, line 109: Can you add more of a callout box for this? you could use
> [!NOTE]
> title
>
> text

or maybe even better [!WARNING] instead of [!NOTE]
Then just make it explicit that we don't expect good performance on these benchmarks yet.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -0,0 +1,149 @@
...
+                not_moved = False
+            else:
+                displacement = np.zeros(3)
+            o.set_displacement(displacement)
+            self.last_location[o.sender_id] = o.location
+        return obs, not_moved
+
+    def matching_step(self, observations):
+        """Update the possible matches given an observation."""
+        buffer_data, not_moved = self._add_displacements(observations)
```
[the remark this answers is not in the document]
vkakerbeck: I like the general idea. Just one note is that I would make the naming a bit clearer. When I read get_if_moved I would think it tells me if the agent moved since the last time the step function was called whereas adding something like _since_reset or _since_episode_start would make it more specific

vkakerbeck (commented):
Nice, thanks for all the changes! I just left a few minor questions

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -0,0 +1,149 @@
...
+            - not_moved (bool): True if none of the observations had a prior location
+              (i.e., no movement detected), False otherwise.
+        """
+        not_moved = True
+        for o in obs:
+            if o.sender_id in self.last_location.keys():
+                displacement = o.location - self.last_location[o.sender_id]
+                not_moved = False
+            else:
+                displacement = np.zeros(3)
```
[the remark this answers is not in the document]
ramyamounir: Yes that's correct. In `EvidenceGraphLM`, the displacement is set to zeros at the beginning of every new episode. In `NoResetEvidenceGraphLM` it is set to zeros only the the beginning of the first episode (or if the agent hasn't moved ofc). Adding a zeros displacement is probably just to avoid a logging error (and keep all buffers lengths in sync) when we save the states to the buffer, I don't think we use it in `_update_evidence` if `not_moved=True`. The buffer [expects a displacement variable](https://github.com/thousandbrainsproject/tbp.monty/blob/f67f439cccd5cc19bba05e329e39f648f1e4855c/src/tbp/monty/frameworks/models/buffer.py#L132).

I will add a clarifying comment in the function docstring.

benchmarks/configs/unsupervised_inference_experiments.py
ramyamounir: Good point, I shouldn't have. This was for my testing and I forgot these statements there.

benchmarks/configs/unsupervised_inference_experiments.py
ramyamounir: I think it's a subtle difference. In the matching phase only I'm referring to preventing the switch to exploration phase (regardless of whether the exploration phase is updating the graphs). I'm not really describing a new test, but I'm just describing the configs. We don't do any pre-training or learning (only evaluate), and when we do evaluate, we don't switch to exploration (only matching). I could be using wrong terminology here.. ?

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -0,0 +1,149 @@
...
+            lm.primary_target_rotation_quat = primary_target["quat_rotation"]
+
+        # reset LMs and SMs buffers to save memory
+        self._reset_modules_buffers()
+
+    def _reset_modules_buffers(self):
+        """Resets buffers for LMs and SMs."""
+        for lm in self.learning_modules:
+            lm.buffer.reset()
+            lm.gsg.wait_factor = 1
```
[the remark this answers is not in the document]
ramyamounir: Yeah, nice idea. This will remove the supervisory signal.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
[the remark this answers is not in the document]
ramyamounir: Yeah, good idea! The current logic of `not_moved` in `GraphLM` is a little confusing to me. The `len(buffer)>1` is not really checking if the agent has moved, it instead checks if this is the first step in the episode. There is a scenario where we are in a middle of the episode (i.e., len(buffer)>1), but the agent has not moved (i.e., none of the LMs have moved). In this case, we don't want to process zero displacements in `_update_evidence` because that's too many computations for nothing. The problem is `_update_evidence` will reinitialize the hypothesis space based on this `not_moved` variable.

My suggestion is:
1. Change the `not_moved` logic in `GraphLM` to loop over observations. If the displacements of all channels are (0,0,0), then set `not_moved` to `True`, otherwise set it to `False`. This logic should work in `GraphLM`, `EvidenceGraphLM`, and `NoResetGraphLM`
2. The evidence dictionary should be reset in `GraphLM.reset()`. 
3. `_update_evidence` should not depend on `not_moved` to figure out if it needs to initialize a hypothesis space. It would check the evidence dictionary for that. 
4. The `not_moved` will skip evidence update to save computations (assuming a hypothesis space exists) if the agent decided not to move in the middle of the episode. I don't know if an agent not moving is an option, but I think the code should handle this edge case anyway and not displace the `possible_locations` with (0,0,0) displacements.

The alternative is to rename `not_moved` to something else (e.g., initialize_space), but I think it's still useful to handle the case where an agent doesn't move efficiently.

*The double negative in `not_moved=False` is a little annoying :sweat_smile: , I'm wondering if it's better to at least change it to `agent_moved` or similar*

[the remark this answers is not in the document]
jeremyshoemaker: I generally try to avoid negation in the name of boolean variables because it gets confusing very quickly. I'd support changing it, but then we have to be sure to change everywhere it's used. If that's a lot of places it might be best to do in an isolated pull request.

tests/unit/no_reset_evidence_lm_test.py
```diff
@@ -0,0 +1,227 @@
...
+                    break
+            return step
+
+        pprint("...parsing experiment...")
+        train_config = copy.deepcopy(self.evidence_config)
+        self.train_exp = MontyObjectRecognitionExperiment(train_config)
+
+        with self.train_exp:
+            pprint("...training...")
+            self.train_exp.model.set_experiment_mode("train")
```
tristanls, line 172: **issue**: This line is extraneous. The second line of `.train()` invoked on the line below sets the experiment mode to `"train"`.

```python
    def train(self):
        """Run n_train_epochs."""
        self.logger_handler.pre_train(self.logger_args)
        self.model.set_experiment_mode("train")
        for _ in range(self.n_train_epochs):
            self.run_epoch()
        self.logger_handler.post_train(self.logger_args)
```

tests/unit/no_reset_evidence_lm_test.py
```diff
@@ -0,0 +1,227 @@
...
+            self.eval_exp.post_episode(episode_2_steps)
+            self.assertGreater(
+                len(self.eval_exp.model.learning_modules[0].evidence),
+                0,
+                "evidence dict should contain evidence values",
+            )
+            self.eval_exp.post_epoch()
+
+    def tearDown(self):
+        """Code that gets executed after every test."""
```
tristanls, line 222: **issue**: This docstring is extraneous. Either remove or describe what is happening in the setup.

tests/unit/no_reset_evidence_lm_test.py
```diff
@@ -0,0 +1,227 @@
...
+from tbp.monty.simulators.habitat.configs import (
+    EnvInitArgsPatchViewMount,
+    PatchViewFinderMountHabitatDatasetArgs,
+)
+from tests.unit.resources.unit_test_utils import BaseGraphTestCases
+
+
+class NoResetEvidenceLMTest(BaseGraphTestCases.BaseGraphTest):
+    def setUp(self):
+        """Code that gets executed before every test."""
```
tristanls, line 50: **issue**: This docstring is extraneous. Either remove or describe what is happening in the setup.

tests/unit/no_reset_evidence_lm_test.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+
+        def run_episode(exp):
+            for step, observation in enumerate(exp.dataloader):
+                exp.pre_step(step, observation)
+                exp.model.step(observation)
+                exp.post_step(step, observation)
+                if exp.model.is_done or step >= exp.max_steps:
+                    break
+            return step
```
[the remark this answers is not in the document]
tristanls: **issue**: It seems, you are using all of `MontyObjectRecognitionExperiment`, but want to override `run_episode_steps()`. 

**suggestion**: Create `class TestExperiment(MontyObjectRecognitionExperiment)` and override `run_episode_steps()` with the code here.

This would also allow you to simplify:
```python
            self.eval_exp.pre_episode()
            episode_1_steps = run_episode(self.eval_exp)
            self.eval_exp.post_episode(episode_1_steps)
```
into
```python
            eval_exp.run_episode()
```
Although, you'd still be doing second episode step-by-step if `pre_episode()` is needed before the assertion:
```python
            # second episode
            self.eval_exp.pre_episode()
            self.assert_dicts_equal(
                post_episode1_evidence,
                self.eval_exp.model.learning_modules[0].evidence,
                "evidence dict should not change between episodes",
            )
            episode_2_steps = run_episode(self.eval_exp)
            self.eval_exp.post_episode(episode_2_steps)
```
into
```python
            # second episode
            eval_exp.pre_episode()
            self.assert_dicts_equal(
                post_episode1_evidence,
                eval_exp.model.learning_modules[0].evidence,
                "evidence dict should not change between episodes",
            )
            episode_2_steps = eval_exp.run_episode_steps()
            eval_exp.post_episode(episode_2_steps)
```

tristanls (commented):
Howdy, I saw something in a test that drew my attention and ended up reviewing the test code (I only reviewed `tests/unit/no_reset_evidence_lm_test.py`).

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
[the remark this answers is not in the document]
vkakerbeck: I agree it is a confusing how it currently is. not_moved doesn't actually mean that the agent hasn't moved yet in the edge case where we are not on step 0. It means that the LM has not received input yet. So you can imagine the 5LM setup with feature change SM. An LM might not be getting inputs since it is not on the object while other LMs are getting inputs and calling the step function. Also, the LM might have gotten one input (len(buffer==1) but hasn't received a second input because the features haven't changed significantly. The second time the LM will get an input is when len(buffer)>1 and at this point we can calculate a displacement between the first and second location. This is when we set `not_moved=False`.

We could call it something like `has_sensed_displacement` or `first_movement_sensed` to avoid the negation in the variable name.

regarding 3. I am not sure I understand what you mean. update_evidence initializes the evidence space before not_moved is updated. not_moved is used to check whether we can calculate a displacement. If we have >1 observation in the buffer then we calculate a displacement and pass it to the update_evidence function. The update_evidence function initializes hypotheses on the first step (len(buffer==1) and not_moved=True) because no displacement was provided. Then on following steps it uses the displacement between the previous and current step to update the evidence. It seems like this distinction between having a displacement and not having one is important. How would you remove that?

regarding 4. I think this should be answered above but in short, not moved means that there is only one observation in the buffer and therefor no way to calculate a displacement. It is not used to skip steps. We have plenty of steps where an LM doesn't receive input from the sensor module and the logic for avoiding unnecessary steps in that case is here: https://github.com/thousandbrainsproject/tbp.monty/blob/f67f439cccd5cc19bba05e329e39f648f1e4855c/src/tbp/monty/frameworks/models/graph_matching.py#L234

Does that make sense?

benchmarks/configs/unsupervised_inference_experiments.py
vkakerbeck: Where is this concretely in the configs? Generally, there is no use case for having an exploratory phase during evaluation, so we never do that. I don't think we need to make a distinction here and can just say that we are doing evaluation only.

tests/unit/no_reset_evidence_lm_test.py
```diff
@@ -0,0 +1,227 @@
...
+
+        pprint("...parsing experiment...")
+        train_config = copy.deepcopy(self.evidence_config)
+        self.train_exp = MontyObjectRecognitionExperiment(train_config)
+
+        with self.train_exp:
+            pprint("...training...")
```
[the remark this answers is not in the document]
ramyamounir: Good question, I'm not sure. I've copied over the structure of running an experiment in unit tests from other unit tests and they all have these [pprint statements](https://github.com/thousandbrainsproject/tbp.monty/blob/f67f439cccd5cc19bba05e329e39f648f1e4855c/tests/unit/evidence_lm_test.py#L809). I assume this is for debugging but I'm not sure how this works within pytest.

Personally, when debugging a unit test I usually just call it from a python script so that my breakpoints and logging would work, but maybe there is a better way..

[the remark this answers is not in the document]
tristanls: Looking into the historical repository, this looks like a pattern from long ago when the architecture of everything was being figured out. It appears to have been copied over since then. @vkakerbeck, it looks like you were involved in adding these. Have these `pprint`s been used for anything recently?

benchmarks/configs/unsupervised_inference_experiments.py
ramyamounir: My bad then, I thought that during evaluation we can still go to exploration phase but that the buffer doesn't update the graphs. It makes more sense to go directly to the next episode when reaching a terminal state in evaluation mode. I'll remove this for the configs description.

tests/unit/no_reset_evidence_lm_test.py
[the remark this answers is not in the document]
vkakerbeck: no, I don't think so. This started with the very first unit tests and then was just copied over many times. When I debugged unit tests I often called `python tests/unit/test_file.py TestClass.test_name` (which I think doesn't work anymore since some recent changes?) and then these print were sometimes useful. But by now we have the log level which usually prints all the necessary info from the logging.info("...") calls all over the code. So I would say we can remove these print statements.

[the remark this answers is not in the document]
tristanls: @vkakerbeck , regarding calling individual tests, I'm not familiar with invoking them via `python`, but here's what I do with `pytest`:
```
pytest tests/unit/base_config_test.py -k test_can_run_eval_epoch
```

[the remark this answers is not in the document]
tristanls: And this call to show the logs (since pytest hides logs by default):
```
pytest tests/unit/base_config_test.py -rP -k test_can_run_eval_epoch
```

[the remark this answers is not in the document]
vkakerbeck: oh that's good to know! I didn't know about that log-showing option

tests/unit/no_reset_evidence_lm_test.py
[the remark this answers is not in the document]
ramyamounir: Nice suggestions, thanks! I've looked into this and figured that I didn't need to rewrite `run_episode_steps`, the one that exists in `MontyObjectRecognitionExperiment` would work fine. So I deleted the `run_episode` function I had there and just called `exp.run_episode_steps`.

The function that needs to be overwritten in `MontyObjectRecognitionExperiment` is `run_episode` so that I can add the assert statements around `pre_episode`. Do you still think it's a good idea to overwrite the `run_episode` so that I can add assert statements? also I assert for different things depending on which episode we're at, so I may be a bit more involved than simply overwriting this function.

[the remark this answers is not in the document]
tristanls: Your integration test would look like:
```python
# first episode
            self.assertEqual(
                len(self.eval_exp.model.learning_modules[0].evidence),
                0,
                "evidence dict should be empty before the first episode",
            )
            self.eval_exp.run_episode()
            post_episode1_evidence = copy.deepcopy(
                self.eval_exp.model.learning_modules[0].evidence
            )
            self.assertGreater(
                len(post_episode1_evidence),
                0,
                "evidence dict should now contain evidence values of the first episode",
            )

            # second episode
            self.eval_exp.run_episode()
            self.assertGreater(
                len(self.eval_exp.model.learning_modules[0].evidence),
                0,
                "evidence dict should contain evidence values",
            )
            self.eval_exp.post_epoch()
```

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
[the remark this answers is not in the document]
ramyamounir: Thanks for the detailed explanation. On number 3, I was suggesting that we don't depend on the buffer size and `not_moved` to decide whether we should initialize a hypothesis space. 

The current logic of initializing a hypothesis space is:
* `EvidenceGraphLM.reset()` resets the buffer
* The buffer lengths sets `not_moved`
* `not_moved` initializes the hypothesis space

Why don't we just reset the hypothesis space between episodes (e.g., `self.evidence={}`), then when we `_update_evidence`, we check if we have an existing hypothesis space in `self.evidence`. If not, we create one. We can initialize the hyp space based on a condition like `if graph_id not in self.evidence`. We wouldn't need the logic of `len(buffer)>1` and `not_moved` here, right?

The `MontyForNoResetEvidenceGraphMatching` does not call the LM reset in `pre_episode`, so it will not delete the hyp space and `_update_evidence` will not reinitialize it. Which should work out fine for unsupervised inference.

benchmarks/configs/unsupervised_inference_experiments.py
```diff
@@ -0,0 +1,242 @@
...
+    "experiment_args"
+].max_eval_steps = EVAL_STEPS
+
+# === End Benchmark Configs === #
+
+
+# === Rapid Prototyping Configs === #
+
+# This enables or disables rapid prototyping configs
+APPLY_RAPID_CONFIGS = False
```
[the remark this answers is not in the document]
ramyamounir: I prefer to merge them because I think they're useful for anyone working on RFC 9, not just me. Is it a good idea to create a sub-issue [here](https://github.com/thousandbrainsproject/tbp.monty/issues/214) as a reminder to remove associated rapid prototype configs? If you think it's still an issue, I can push it to a different branch on my fork.

tests/unit/no_reset_evidence_lm_test.py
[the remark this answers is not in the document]
ramyamounir: I could split it into two tests, one testing that that evidence values are not changed by `pre_episode` and the other checking the size of evidence values across episodes. It's just the same exact experiment and I found it unnecessary to run the evaluation twice. Note that even in the unit test of `pre_episode` I still need to run 2 episodes because `pre_episodes` for the first episodes behaves differently than for the second episode in this class. I left some comments [here](https://github.com/thousandbrainsproject/tbp.monty/blob/1605f88cfa2d7ffe74ef1e4759f205f90ef5ee91/src/tbp/monty/frameworks/models/no_reset_evidence_matching.py#L42).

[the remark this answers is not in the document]
tristanls: > I could split it into two tests, one testing that evidence values are not changed by pre_episode and the other checking the size of evidence values across episodes

This wasn't the goal of my suggestion. I thought `pre_episode` had consistent behavior and could be tested in isolation. I did not notice that it does completely different things on the first run vs follow-on runs.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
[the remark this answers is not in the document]
vkakerbeck: Maybe I am missing something but I think the issue is that we look at `len(buffer)>1`, not `len(buffer)>0`. So this check looks at whether we have received a **second** observation yet. The hypothesis space is already initialized on the first observation. `not_moved` is used to check if we can calculate a displacement (requires two observations) and use that to **update** the evidence.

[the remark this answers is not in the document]
ramyamounir: Yes, this makes sense; on the first observation we initialize, on the second observation we update. I detect the first observation by relying on `EvidenceGraphLM.reset()` at the beginning of each episode instead of relying on the buffer.

I did these changes in a separate branch on my fork to explain what I mean,

1) Add an explicit reset of `self.evidence` to `EvidenceGraphLM.reset()` [here](https://github.com/ramyamounir/tbp.monty/blob/5d8cdbb329af45218db3386ede8418712811acae/src/tbp/monty/frameworks/models/evidence_matching.py#L370)
2) Rely on the state of `self.evidence` to reinitialize the hypothesis space when needed, as shown [here](https://github.com/ramyamounir/tbp.monty/blob/5d8cdbb329af45218db3386ede8418712811acae/src/tbp/monty/frameworks/models/evidence_matching.py#L944)
3) Remove the need to decide on `not_moved` in `matching_step`, as shown [here](https://github.com/ramyamounir/tbp.monty/blob/5d8cdbb329af45218db3386ede8418712811acae/src/tbp/monty/frameworks/models/graph_matching.py#L649)
4) Remove `not_moved` from the query [here](https://github.com/ramyamounir/tbp.monty/blob/5d8cdbb329af45218db3386ede8418712811acae/src/tbp/monty/frameworks/models/graph_matching.py#L987) and always return the current displacements (even if zeros).

This worked fine for `EvidenceGraphLM`, but it may not be backward-compatible for other LM types (e.g., `FeatureGraphLM`) since we're changing main logic in `GraphLM`.

*I will not make any of these changes in this PR, but I just wanted to explain what I meant and check if it makes sense or if I'm missing something. We can discuss this later in more details in a technical meeting.*

benchmarks/configs/unsupervised_inference_experiments.py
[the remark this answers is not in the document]
tristanls: @nielsleadholm, I'm not sure I understand what part of the RFC process is in play here. 

**issue**: It seems arbitrary to me that this flag should change what configurations are. 

**question**: Why aren't there two extra configs with separate names?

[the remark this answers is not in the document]
tristanls: **issue**: It also seems to me these extra configs are `monty_lab`-like configurations intended for internal use and not to be released with Monty. Shouldn't these extra configs go into `monty_lab` or a new test repo if they need to be shared?

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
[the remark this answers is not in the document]
ramyamounir: I have isolated this refactoring change in a separate PR for clarity; it affects the code in different places and I wanted to make sure I test this isolated change separately. Once [PR#231](https://github.com/thousandbrainsproject/tbp.monty/pull/231) is merged, I will update this PR to incorporate Niels suggestion.

benchmarks/configs/unsupervised_inference_experiments.py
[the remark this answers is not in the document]
ramyamounir: I will remove the rapid prototype configs from this PR and use them locally. I was mostly thinking that it's not just for internal TBP researchers, but they would be useful for others from the community if they decide to work on RFC 9 in parallel. But no need to do that if it causes (temporary) confusion.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -94,6 +92,7 @@ class NoResetEvidenceGraphLM(EvidenceGraphLM):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.last_location = {}
+        self.gsg.wait_growth_multiplier = 1
```
nielsleadholm, line 95: **suggestion (non-blocking):** Maybe just add a brief comment for why this is hardcoded when no resetting - i.e. doesn't make sense to exponentially grow when objects are swapped without any supervised resetting.

nielsleadholm (approved):
Looks really nice and clean, thanks for responding to all those changes!

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -0,0 +1,137 @@
...
+        super().__init__(*args, **kwargs)
+        self.last_location = {}
+        self.gsg.wait_growth_multiplier = 1
+
+    def reset(self):
+        super().reset()
+        self.evidence = {}
+        self.last_location = {}
+
+    def _add_displacements(self, obs):
```
vkakerbeck, line 102: Why can't you use the existing displacement calculation code in the buffer class (`get_current_displacement`)? I would be very cautious about duplicating this functionality here for 3 reasons:
1. now we have two places where we keep track of recent observations: the buffer and the LM itself
2. If we ever change the way displacements should be calculated (or how we deal with input channels) we have to change it in multiple places
3. related to 2, we already have different ways displacements are calculated. In the DisplacementGraphLM we have the option to work with point pair features instead of x-y kind of displacements. There may be other cases like that in the future. It would be easiest to just overwrite the buffers `get_current_displacement` function with whatever custom displacement calculations we want to do. But if you have this functionality in another place now, that will be very error-prone and easily overlooked.

I don't see an issue with just checking the keys in the buffer and calling the buffers function here. Then when we call reset, we reset the buffer's last location and not a new attribute of the LM. Maybe I am missing something.

vkakerbeck (commented):
Thanks for the thorough review und updates! It looks very nice now. However I left one more comment on why the displacement calculation is now duplicated in another place.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
ramyamounir: I think I see your point, but I don't fully understand the suggested solution.

My understanding of the displacement logic is that: 
* `GraphLM._add_displacement` is responsible for [calculating the current displacement](https://github.com/thousandbrainsproject/tbp.monty/blob/40ec4225b4f697480e2f07cc8e86e57ef28edb1d/src/tbp/monty/frameworks/models/graph_matching.py#L1061) for all channels, then [modifies the observation](https://github.com/thousandbrainsproject/tbp.monty/blob/40ec4225b4f697480e2f07cc8e86e57ef28edb1d/src/tbp/monty/frameworks/models/graph_matching.py#L1066) to include this calculated displacement.
* These calculated displacements get [added](https://github.com/thousandbrainsproject/tbp.monty/blob/40ec4225b4f697480e2f07cc8e86e57ef28edb1d/src/tbp/monty/frameworks/models/graph_matching.py#L647) to the buffer, as shown [here](https://github.com/thousandbrainsproject/tbp.monty/blob/40ec4225b4f697480e2f07cc8e86e57ef28edb1d/src/tbp/monty/frameworks/models/buffer.py#L134)
* We then **only** retrieve these calculated displacement back from the buffer in `GraphLM._compute_possible_matches`, as shown [here](https://github.com/thousandbrainsproject/tbp.monty/blob/40ec4225b4f697480e2f07cc8e86e57ef28edb1d/src/tbp/monty/frameworks/models/graph_matching.py#L1003).

My point is we do not calculate any displacements in `Buffer.get_current_displacement()`, this is only used for retrieving the displacement originally calculated in `GraphLM._add_dispalcement`. We want to change the logic by which we calculate the displacement, not retrieve it. If I try to use the `Buffer.get_current_displacement` after a `Buffer.reset`, it will not find any stored displacements, and I cannot calculate it there because the reset has also deleted the location buffer.

The main problem is that we want to fully reset the buffer between episodes (even in unsupervised inference), due to memory restrictions, but we want to keep the last location (for displacement calculations) despite the reset which deletes everything in the location buffer. One possible solution is to define another buffer class with a `last_location` attribute that does not get reset in the `buffer.reset()` signal, but I think it's confusing and we will still have to overwrite the `GraphLM._add_displacement` function to use this new attribute `buffer.last_location` instead of `buffer.get_current_location`.

I could have completely misunderstood your solution.

vkakerbeck: Oh good point. just ignore my comment :D

vkakerbeck (approved):

PR #222, opened 2025-03-25
chore: add tests for ProfileExperimentMixin

jeremyshoemaker:
While working on https://github.com/thousandbrainsproject/tbp.monty/pull/220, we realized that this mixin might be affected by the changes there. Upon investigation, I noticed that it was lacking any tests to make sure at a bare minimum that it works, so I'm adding some here.

I also added an `__init_subclass__` method to it to check and ensure the required invariants are true when it is used.

### DIFF
--- a/src/tbp/monty/frameworks/experiments/profile.py
+++ b/src/tbp/monty/frameworks/experiments/profile.py
@@ -14,6 +14,8 @@ import os
 import pandas as pd
 import wandb
 
+from tbp.monty.frameworks.experiments import MontyExperiment
+
 
 def make_stats_df(stats):
     """Convert cProfile.Profile() stats to dataframe.
@@ -44,6 +46,27 @@ class ProfileExperimentMixin:
 
         NOTE: make sure this class is leftmost in mixin order.
     """
+
+    def __init_subclass__(cls, **kwargs):
+        """Ensure that the mixin is used in the correct way.
+
+        We want to ensure that the mixin is always the leftmost class listed in
+        the base classes when used so that the methods defined here override the ones
+        defined in MontyExperiment or its subclasses. We also want to ensure that
+        any subclasses are actually extending MontyExperiment. This ensures that by
+        raising an error if it is not.
+
+        Raises:
+            TypeError: when the mixin isn't the leftmost base class of the subclass
+            being initialized or the base classes don't include a subclass of
+            MontyExperiment.
+        """
+        super().__init_subclass__(**kwargs)
+        if cls.__bases__[0] is not ProfileExperimentMixin:
+            raise TypeError("ProfileExperimentMixin must be leftmost base class.")
+        if not any([issubclass(b, MontyExperiment) for b in cls.__bases__]):
+            raise TypeError("ProfileExperimentMixin must be mixed in with a subclass "
+                            "of MontyExperiment.")
 
     def make_profile_dir(self):
         self.profile_dir = os.path.join(self.output_dir, "profile")
--- a/tests/unit/profile_experiment_mixin_test.py
+++ b/tests/unit/profile_experiment_mixin_test.py
@@ -1,0 +1,148 @@
+# Copyright 2025 Thousand Brains Project
+# Copyright 2022-2024 Numenta Inc.
+#
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+import copy
+import shutil
+import tempfile
+import unittest
+from pathlib import Path
+from pprint import pprint
+from typing import Set
+
+import pytest
+
+from tbp.monty.frameworks.config_utils.config_args import (
+    LoggingConfig,
+    SingleCameraMontyConfig,
+)
+from tbp.monty.frameworks.config_utils.make_dataset_configs import (
+    DebugExperimentArgs,
+    EnvironmentDataLoaderPerObjectEvalArgs,
+    EnvironmentDataLoaderPerObjectTrainArgs,
+    NotYCBEvalObjectList,
+    NotYCBTrainObjectList,
+)
+from tbp.monty.frameworks.environments import embodied_data as ed
+from tbp.monty.frameworks.experiments import MontyExperiment, ProfileExperimentMixin
+from tbp.monty.simulators.habitat.configs import (
+    EnvInitArgsSinglePTZ,
+    SinglePTZHabitatDatasetArgs,
+)
+
+
+class ProfiledExperiment(ProfileExperimentMixin, MontyExperiment):
+    pass
+
+
+class ProfileExperimentMixinTest(unittest.TestCase):
+    def setUp(self):
+        """Code that gets executed before every test."""
+        self.output_dir = tempfile.mkdtemp()
+
+        base = dict(
+            experiment_class=ProfiledExperiment,
+            experiment_args=DebugExperimentArgs(),
+            logging_config=LoggingConfig(
+                output_dir=self.output_dir, python_log_level="DEBUG"
+            ),
+            monty_config=SingleCameraMontyConfig(),
+            dataset_class=ed.EnvironmentDataset,
+            dataset_args=SinglePTZHabitatDatasetArgs(
+                env_init_args=EnvInitArgsSinglePTZ(data_path=None).__dict__
+            ),
+            train_dataloader_class=ed.EnvironmentDataLoaderPerObject,
+            train_dataloader_args=EnvironmentDataLoaderPerObjectTrainArgs(
+                object_names=NotYCBTrainObjectList().objects,
+            ),
+            eval_dataloader_class=ed.EnvironmentDataLoaderPerObject,
+            eval_dataloader_args=EnvironmentDataLoaderPerObjectEvalArgs(
+                object_names=NotYCBEvalObjectList().objects,
+            ),
+        )
+
+        self.base_config = base
+
+    def tearDown(self):
+        """Code that gets executed after every test."""
+        shutil.rmtree(self.output_dir)
+
+    def get_profile_files(self) -> Set[str]:
+        """Helper to get the files in the profile directory in a set.
+
+        Returns:
+            set of filenames in the profile directory.
+        """
+        path = Path(self.output_dir, "profile")
+        filenames = [f.name for f in path.iterdir() if f.is_file()]
+        # returning a set so the order doesn't matter
+        return set(filenames)
+
+    @staticmethod
+    def test_correct_subclassing_does_not_error() -> None:
+        # this shouldn't raise an exception at evaluation time because
+        # the mixin is the leftmost base class listed
+        class GoodSubclass(ProfileExperimentMixin, MontyExperiment):
+            pass
+
+    @staticmethod
+    def test_incorrect_subclassing_raises_error() -> None:
+        with pytest.raises(TypeError):
+            class BadSubclass(MontyExperiment, ProfileExperimentMixin):
+                pass
+
+    @staticmethod
+    def test_missing_experiment_base_raises_error() -> None:
+        with pytest.raises(TypeError):
+            class BadSubclass(ProfileExperimentMixin):
+                pass
+
+    def test_can_run_episode(self) -> None:
+        pprint("...parsing experiment...")
+        base_config = copy.deepcopy(self.base_config)
+        with ProfiledExperiment() as exp:
+            exp.setup_experiment(base_config)
+            pprint("...training...")
+            exp.model.set_experiment_mode("train")
+            exp.dataloader = exp.train_dataloader
+            exp.run_episode()
+
+        self.assertSetEqual(self.get_profile_files(), {
+            "profile-setup_experiment.csv",
+            "profile-train_epoch_0_episode_0.csv",
+        })
+
+    def test_can_run_train_epoch(self) -> None:
+        pprint("...parsing experiment...")
+        base_config = copy.deepcopy(self.base_config)
+        with ProfiledExperiment() as exp:
+            exp.setup_experiment(base_config)
+            exp.model.set_experiment_mode("train")
+            exp.run_epoch()
+
+        self.assertSetEqual(self.get_profile_files(), {
+            "profile-setup_experiment.csv",
+            "profile-train_epoch_0_episode_0.csv",
+            "profile-train_epoch_0_episode_1.csv",
+            "profile-train_epoch_0_episode_2.csv",
+        })
+
+    def test_can_run_eval_epoch(self) -> None:
+        pprint("...parsing experiment...")
+        base_config = copy.deepcopy(self.base_config)
+        with ProfiledExperiment() as exp:
+            exp.setup_experiment(base_config)
+            exp.model.set_experiment_mode("eval")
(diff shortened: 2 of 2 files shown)


tests/unit/profile_experiment_mixin_test.py
jeremyshoemaker, line 1: A lot of the code here is copied from `base_config_test.py`. I'm not sure how much of this is repeated in other places, but it might make sense to have a base test case that other tests can inherit from and modify instead of copy-pasting.

I'm also not sure that I'm effectively exercising this code. I did work on it on top of my branch from #220, so it works there as well as here, but I might be missing something that is broken.

src/tbp/monty/frameworks/experiments/profile.py
```diff
@@ -45,6 +47,27 @@ class ProfiledExp(ProfileExperimentMixin, SomeExp):
...
+            TypeError: when the mixin isn't the leftmost base class of the subclass
+            being initialized or the base classes don't include a subclass of
+            MontyExperiment.
+        """
+        super().__init_subclass__(**kwargs)
+        if cls.__bases__[0] is not ProfileExperimentMixin:
+            raise TypeError("ProfileExperimentMixin must be leftmost base class.")
+        if not any([issubclass(b, MontyExperiment) for b in cls.__bases__]):
+            raise TypeError("ProfileExperimentMixin must be mixed in with a subclass "
+                            "of MontyExperiment.")
```
tristanls, line 69: **praise**: TIL this.

tests/unit/profile_experiment_mixin_test.py
```diff
@@ -0,0 +1,148 @@
+# Copyright 2025 Thousand Brains Project
+# Copyright 2022-2024 Numenta Inc.
```
tristanls, line 2: **issue**: Follow the [style guide for copyright headers](https://thousandbrainsproject.readme.io/docs/style-guide#source-code-copyright-and-license-header).

tests/unit/profile_experiment_mixin_test.py
```diff
@@ -0,0 +1,148 @@
...
+)
+from tbp.monty.frameworks.environments import embodied_data as ed
+from tbp.monty.frameworks.experiments import MontyExperiment, ProfileExperimentMixin
+from tbp.monty.simulators.habitat.configs import (
+    EnvInitArgsSinglePTZ,
+    SinglePTZHabitatDatasetArgs,
+)
+
+
+class ProfiledExperiment(ProfileExperimentMixin, MontyExperiment):
```
tristanls, line 39: **issue**: Let's call this a `FakeExperiment`, this is a fake in the mocks, stubs, fake context.

tests/unit/profile_experiment_mixin_test.py
```diff
@@ -0,0 +1,148 @@
...
+        """
+        path = Path(self.output_dir, "profile")
+        filenames = [f.name for f in path.iterdir() if f.is_file()]
+        # returning a set so the order doesn't matter
+        return set(filenames)
+
+    @staticmethod
+    def test_correct_subclassing_does_not_error() -> None:
+        # this shouldn't raise an exception at evaluation time because
+        # the mixin is the leftmost base class listed
```
tristanls, line 89: **nitpick**: This comment is redundant with the test name.

**suggestion**: Remove.

tests/unit/profile_experiment_mixin_test.py
```diff
@@ -0,0 +1,148 @@
...
+    def test_missing_experiment_base_raises_error() -> None:
+        with pytest.raises(TypeError):
+            class BadSubclass(ProfileExperimentMixin):
+                pass
+
+    def test_can_run_episode(self) -> None:
+        pprint("...parsing experiment...")
+        base_config = copy.deepcopy(self.base_config)
+        with ProfiledExperiment() as exp:
+            exp.setup_experiment(base_config)
```
tristanls, line 109: **question**: I was a bit thrown off by this. I assume this will change to below in a follow-on PR?
```python
with ProfiledExperiment(base_config) as exp:
```

tests/unit/profile_experiment_mixin_test.py
```diff
@@ -0,0 +1,148 @@
...
+)
+
+
+class ProfiledExperiment(ProfileExperimentMixin, MontyExperiment):
+    pass
+
+
+class ProfileExperimentMixinTest(unittest.TestCase):
+    def setUp(self):
+        """Code that gets executed before every test."""
```
tristanls, line 45: **issue**: Extraneous docstring.

**suggestion**: Remove, or describe what is happening below.

tests/unit/profile_experiment_mixin_test.py
```diff
@@ -0,0 +1,148 @@
...
+            eval_dataloader_class=ed.EnvironmentDataLoaderPerObject,
+            eval_dataloader_args=EnvironmentDataLoaderPerObjectEvalArgs(
+                object_names=NotYCBEvalObjectList().objects,
+            ),
+        )
+
+        self.base_config = base
+
+    def tearDown(self):
+        """Code that gets executed after every test."""
```
tristanls, line 72: **issue**: Extraneous docstring.

**suggestion**: Remove, or describe the line below.

tests/unit/profile_experiment_mixin_test.py
tristanls: Copying `base_config_test.py` makes sense to me for now.

tests/unit/profile_experiment_mixin_test.py
```diff
@@ -0,0 +1,148 @@
...
+    SingleCameraMontyConfig,
+)
+from tbp.monty.frameworks.config_utils.make_dataset_configs import (
+    DebugExperimentArgs,
+    EnvironmentDataLoaderPerObjectEvalArgs,
+    EnvironmentDataLoaderPerObjectTrainArgs,
+    NotYCBEvalObjectList,
+    NotYCBTrainObjectList,
+)
+from tbp.monty.frameworks.environments import embodied_data as ed
```
tristanls, line 31: **issue**: Let's import the classes we use instead of treating `embodied_data as ed` as `numpy as np`. I know we did this in `base_config_test.py`, but I'm unsure why. @vkakerbeck, do you have historical context for this usage of `embodied_data`?

tests/unit/profile_experiment_mixin_test.py
```diff
@@ -0,0 +1,148 @@
...
+            class BadSubclass(MontyExperiment, ProfileExperimentMixin):
+                pass
+
+    @staticmethod
+    def test_missing_experiment_base_raises_error() -> None:
+        with pytest.raises(TypeError):
+            class BadSubclass(ProfileExperimentMixin):
+                pass
+
+    def test_can_run_episode(self) -> None:
```
tristanls, line 105: **issue**: I think the test name should be along the lines of `test_run_episode_is_profiled`. Same for remaining profile tests.

tests/unit/profile_experiment_mixin_test.py
```diff
@@ -0,0 +1,148 @@
...
+            exp.setup_experiment(base_config)
+            pprint("...training...")
+            exp.model.set_experiment_mode("train")
+            exp.dataloader = exp.train_dataloader
+            exp.run_episode()
+
+        self.assertSetEqual(self.get_profile_files(), {
+            "profile-setup_experiment.csv",
+            "profile-train_epoch_0_episode_0.csv",
+        })
```
tristanls, line 118: **issue**: To your point above about exercising the code, you should check that there's something in these CSVs that corresponds to the expected cProfile being captured in them.

tristanls (commented):
Thank you for putting this together and improving our code coverage.

I think adding tests is a `chore:` and not a `fix:`.

tests/unit/profile_experiment_mixin_test.py
tristanls: **suggestion**: I think for the sake of the profiler, we can create no-op `setup_experiment`, `run_episode`, `train`, `evaluate`, and `close` methods.

To test the profiler we just want to make sure these are invoked and that profiler writes out profile data. This way, we don't have to copy over the base config setup or anything of the sort.

tests/unit/profile_experiment_mixin_test.py
tristanls: Nevermind my previous comment, I think we can do without copying `base_config_test.py`. See https://github.com/thousandbrainsproject/tbp.monty/pull/222/files#r2012936994

tests/unit/profile_experiment_mixin_test.py
jeremyshoemaker: Well, this touches on what I was thinking of when I said I'm not sure I'm covering everything in these tests that needs to be tested.

We want to test two things:
1. The profiling methods are called correctly when the experiments are run
2. The experiment methods are called correctly / the experiment is run correctly when the profiling mixin is being used.

I'm actually more concerned about testing the second one than the first one, but I don't know specifics of what I should check for to confirm that the experiment is being run correctly.

tests/unit/profile_experiment_mixin_test.py
jeremyshoemaker: Yeah, I had some similar confusion, because I originally wrote these changes on top of my other branch, and when I switched to a new branch based off of `main`, I got errors about the constructor not taking an argument, because that change is part of the changes in #220.

I'll have to rebase that branch after I merge this one and update these call sites.

tests/unit/profile_experiment_mixin_test.py
jeremyshoemaker: Ah, I was just copying from another file. I didn't realize the format was different. I guess that makes sense, since this isn't Numenta code.
- [ ] update copyright notice

tests/unit/profile_experiment_mixin_test.py
jeremyshoemaker: Fair point, I'll remove it. I think my original thinking was that the test seems to do nothing, and I wanted to be explicit about what we're testing here, but anyone working on this _should_ be able to tell what it's testing based off the name.
- [ ] remove superfluous comment

tests/unit/profile_experiment_mixin_test.py
jeremyshoemaker: I thought of that, but I'm not sure there will be anything we can consistently assert on, since it's profiling CSV data, that will be slightly different on every run.

We could do some basic checks, like confirm the file isn't empty. Maybe check and see that it has the expected number of columns and rows, but beyond that I'm not sure.

I'll take a look at what is in them and see if there are any reasonable tests we can add.

src/tbp/monty/frameworks/experiments/profile.py
```diff
@@ -45,6 +47,27 @@ class ProfiledExp(ProfileExperimentMixin, SomeExp):
...
+
+        Raises:
+            TypeError: when the mixin isn't the leftmost base class of the subclass
+            being initialized or the base classes don't include a subclass of
+            MontyExperiment.
+        """
+        super().__init_subclass__(**kwargs)
+        if cls.__bases__[0] is not ProfileExperimentMixin:
+            raise TypeError("ProfileExperimentMixin must be leftmost base class.")
+        if not any([issubclass(b, MontyExperiment) for b in cls.__bases__]):
```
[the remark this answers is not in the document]
jeremyshoemaker: I'm adding a test to verify this works as expected.

tests/unit/profile_experiment_mixin_test.py
jeremyshoemaker: Thinking about this some more, maybe we need two sets of tests. One set using mocks like you suggest to verify the first thing, and a set using an actual experiment to test the second thing. That way they're isolated and testing a single thing instead of two things at once.

jeremyshoemaker: Well, as I started down that path I ran into some trouble because the mixin is expecting some of its method to be called by methods in MontyExperiment. For example, it overrides `run_episode` and outputs files with the episode and epoch number in the filenames, but that method is called by `run_epoch` which the mixin doesn't override, and so that method never gets called if we mock those methods on MontyExperiment.

The mixin is, understandably, intimately tied to how MontyExperiement works, so I don't think trying to test it with a mocked version makes as much sense.

tests/unit/profile_experiment_mixin_test.py
tristanls: **suggestion**: Maybe the test name could be made even more explicit `test_leftmost_subclassing_does_not_error`

tests/unit/profile_experiment_mixin_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @staticmethod
+    def test_correct_subclassing_does_not_error() -> None:
+        # this shouldn't raise an exception at evaluation time because
+        # the mixin is the leftmost base class listed
+        class GoodSubclass(ProfileExperimentMixin, MontyExperiment):
+            pass
+
+    @staticmethod
+    def test_incorrect_subclassing_raises_error() -> None:
```
tristanls, line 94: **suggestion**: With `test_leftmost_subclassing_does_not_error` test name above, this could be made more explicit: `test_other_than_leftmost_subclassing_raises_error`.

tests/unit/profile_experiment_mixin_test.py
tristanls: What does the file look like? I can't imagine a CSV file changing in its entirety on every invocation. At the very least, the column names should be the same for every profile of the same method.

tristanls: I think testing that the headers are `,func,ncalls,ccalls,tottime,cumtime,callers` and that there is a non-zero number of rows would suffice for the CSV.

Can we do a passthrough mock to verify that the super method gets called? https://stackoverflow.com/a/25608441

jeremyshoemaker: > I think testing that the headers are ,func,ncalls,ccalls,tottime,cumtime,callers and that there is a non-zero number of rows would suffice for the CSV.

I updated the test to do that in the time between my last comment and when you posted this one :grin:

> Can we do a passthrough mock to verify that the super method gets called?

Unless I'm missing something, I don't think that approach using `wraps` would do it. That wouldn't let us distinguish calls to methods that came from MontyExperiment vs ones that are overridden in the mixin. 

I'm not sure trying to hook into the `super()` calls would help either, though admittedly, this is reaching the edges of my esoteric Python knowledge.

I think we _could_ use that technique to verify that the correct methods in the experiment are being called when we expect them to be, but then that would tie the test to the internals of MontyExperiment more than I'd like leading to the same churn I mentioned above in reference to the profile output.

🤔 

Thinking about this some more and looking at how the mixin works, I don't think we need to actually check that MontyExperiment is doing what we expect, because we can kinda infer that from the output profile files. 

I think I'd probably be fine with leaving this at just testing the profile files exist and have some reasonable contents. Thoughts?

tristanls: Well, stepping back a bit. The test existing is an improvement from before. So I'm up for merging it as is.

jeremyshoemaker: That's my feeling as well.

tristanls (approved):
Thank you for the updates.

vkakerbeck (approved):
This is great! Thanks for adding those tests and the thorough review @tristanls and @jeremyshoemaker

PR #227, opened 2025-04-04
feat: unsupervised inference logging and plotting functions

ramyamounir:
This PR contributes to the YCB testbed by introducing additional logging and plotting functions. More specifically:
* `NoResetEvidenceGraphLM` now includes new functions to calculate the theoretical limit and pose errors.
* The LM function `_add_detailed_stats` is overwritten in `NoResetEvidenceGraphLM` to log the calculated theoretical limit and pose errors at every step of every episode. This information is only logged if using a detailed logger.
* Two plotting functions are provided under `src/tbp/monty/frameworks/plots/no_reset_evidence.py` to visualize the maximum evidence scores for every object, pose error and theoretical error. The plots should look like the examples below.

evidence
pose_error

**Note: It may be useful to comment out this logic [here](https://github.com/thousandbrainsproject/tbp.monty/blob/8312bd67fc82acaefae1d55894dd4da1c1d0ebd6/src/tbp/monty/frameworks/experiments/monty_experiment.py#L376) to avoid storing the raw observations. This logic overwrites the configs and should be addressed in a different PR**


src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -128,6 +133,71 @@ def _add_displacements(self, obs):
...
+        """Add detailed statistics to the logging dictionary.
+
+        This includes metrics like the max evidence score per object, the theoretical
+        limit of Monty i.e., pose error of Monty's best hypothesis on the target object)
+        , and the pose error of the MLH hypothesis on the target object.
+
+        Args:
+            stats (Dict[str, Any]): The existing statistics dictionary to augment.
+            get_rotations (bool): Flag indicating if rotation stats are needed
+                                  (currently unused in implementation).
```
tristanls, line 148: **issue**: `_add_detailed_stats` should not have been overridden with an extra parameter in `EvidenceGraphLM`. Let's fix that here and get rid of this unused param.

**suggestion**: `EvidenceGraphLM` overrides `collect_stats_to_save` _and_ it overrides `_add_detailed_stats` with extra parameter:
```python
    def collect_stats_to_save(self):
        """Get all stats that this LM should store in the buffer for logging.

        Returns:
            The stats dictionary.
        """
        stats = {
            "possible_matches": self.get_possible_matches(),
            "current_mlh": self.get_current_mlh(),
        }
        if self.has_detailed_logger:
            # Save possible poses once since they don't change during episode
            get_rotations = False
            if "possible_rotations" not in self.buffer.stats.keys():
                get_rotations = True

            stats = self._add_detailed_stats(stats, get_rotations)
        return stats

    def _add_detailed_stats(self, stats, get_rotations):
        stats["possible_locations"] = self.possible_locations
        if get_rotations:
            stats["possible_rotations"] = self.get_possible_poses(as_euler=False)
        stats["evidences"] = self.evidence
        stats["symmetry_evidence"] = self.symmetry_evidence
        return stats
```
There is no need for this. You can fix the overrides by inlining `EvidenceGraphLM._add_detailed_stats(...)` inside its `EvidenceGraphLM.collect_stats_to_save` override, like so:
```python
    def collect_stats_to_save(self):
        """Get all stats that this LM should store in the buffer for logging.

        Returns:
            The stats dictionary.
        """
        stats = {
            "possible_matches": self.get_possible_matches(),
            "current_mlh": self.get_current_mlh(),
        }
        if self.has_detailed_logger:
            # Save possible poses once since they don't change during episode
            get_rotations = False
            if "possible_rotations" not in self.buffer.stats.keys():
                get_rotations = True

            stats["possible_locations"] = self.possible_locations
            if get_rotations:
                stats["possible_rotations"] = self.get_possible_poses(as_euler=False)
            stats["evidences"] = self.evidence
            stats["symmetry_evidence"] = self.symmetry_evidence
            
        return stats
```
Then, delete `EvidenceGraphLM._add_detailed_stats(...)`.

Once you do that, you can override `collect_stats_to_save` directly with what you want without relying on the `_add_detailed_stats` mechanism, which seems inadequate for `EvidenceGraphLM` and its subclasses.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -128,6 +133,71 @@ def _add_displacements(self, obs):
...
+        This considers all possible hypotheses rotations on the target object
+        and compares them to the target's rotation. The theoretical limit conveys the
+        best achievable performance if Monty selects the best hypothesis as its most
+        likely hypothesis (MLH).
+
+        Returns:
+            float: The minimum achievable rotation error (in radians).
+        """
+        poses = self.possible_poses[self.primary_target].copy()
+        hyp_rotations = Rotation.from_matrix(poses).inv().as_quat().tolist()
```
tristanls, line 172: **issue**: `.as_quat()` is extraneous, it is undone in the loop below on line 177 via `Rotation.from_quat(rot)`.

**suggestion**:
```python
hyp_rotations = Rotation.from_matrix(poses).inv()
...
for rot in hyp_rotations:
    ...
```

tristanls (commented):
**issue**: Currently, `src/tbp/monty/frameworks/plots/no_reset_evidence.py` is unused dead code. How are these intended to be used?

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
tristanls: I'm gonna make a separate PR for this real quick...

tristanls: Created https://github.com/thousandbrainsproject/tbp.monty/pull/238 to address this.

tristanls: OK, I found a simpler update in https://github.com/thousandbrainsproject/tbp.monty/pull/239. Once #239 is merged, you'll be able to remove the unused param.

ramyamounir: Perfect, thanks!

nielsleadholm:
Thanks for putting this together Ramy! I haven't had a chance to go through the whole PR in detail yet, but just a high-level comment:

Re. `_target_object_theoretical_limit` and `_target_object_pose_error`, I wonder whether these would be better off somewhere else? In particular, when thinking about the responsibility of the `MontyForNoResetEvidenceGraphMatching` class, it should do graph matching, object recognition etc, but it feels like calculating specific values which are of interest to the experimenter, but not for internal use by the model's intelligence, falls outside of this scope. Two suggestions:

1) A (maybe complicated) approach would be to move these calculations to the buffer, as that is more of an experimenter-facing object that already does a lot of calculations for external metrics.
2) A simpler approach could just be to make them utility functions that are imported and called in `_add_detailed_stats`, i.e. don't make them methods of a class at all? 

I know that's a subtle difference, and maybe I'm missing something, but I think it would just help prevent the `GraphMatching` classes from growing in complexity, where their responsibility becomes increasingly blurred.

Curious what others think about this including @tristanls

tristanls:
> Re. _target_object_theoretical_limit and _target_object_pose_error, I wonder whether these would be better off somewhere else? In particular, when thinking about the responsibility of the MontyForNoResetEvidenceGraphMatching class, it should do graph matching, object recognition etc, but it feels like calculating specific values which are of interest to the experimenter, but not for internal use by the model's intelligence, falls outside of this scope. Two suggestions:
>
>    1. A (maybe complicated) approach would be to move these calculations to the buffer, as that is more of an experimenter-facing object that already does a lot of calculations for external metrics.
>    2. A simpler approach could just be to make them utility functions that are imported and called in _add_detailed_stats, i.e. don't make them methods of a class at all?
> 
> I know that's a subtle difference, and maybe I'm missing something, but I think it would just help prevent the GraphMatching classes from growing in complexity, where their responsibility becomes increasingly blurred.
>
> Curious what others think about this including @tristanls

@nielsleadholm Yes, they would be better somewhere else. I would rather we stop growing the LM inheritance hierarchy in general and not add `NoResetEvidenceGraphLM` as another subclass. Still, I didn't want to get in the way of research as we haven't prepped an alternative using composition instead of inheritance. Inheritance hierarchies should be very shallow (e.g., one level), and our existing ones seem pretty deep. Our inheritance use is based on functionality reuse, and that's usually not what inheritance is good for. But, back to the topic, anything reducing the customization of deep subclasses is beneficial at this point.

nielsleadholm:
Thanks for the input @tristanls ! Good point about the depth of our classes. At some point it wouldn't be a bad idea to consolidate the various versions of Displacement --> Feature --> Evidence etc. LMs into a single, actually used LM (keeping any older features like storing displacements where necessary). This would be much easier for new users to understand and use. I think this is also something we've talked about before but just haven't gotten round to. Tagging @vkakerbeck just as a reminder for something to have in the back of our minds. 

Re. this PR, in that case Ramy I think I would go with the utils function route, it's the simplest to implement and also avoids making the buffer class more complex.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -128,6 +133,71 @@ def _add_displacements(self, obs):
             self.last_location[o.sender_id] = o.location
         return obs
 
+    def _add_detailed_stats(
+        self, stats: Dict[str, Any], get_rotations: bool
+    ) -> Dict[str, Any]:
+        """Add detailed statistics to the logging dictionary.
+
+        This includes metrics like the max evidence score per object, the theoretical
+        limit of Monty i.e., pose error of Monty's best hypothesis on the target object)
```
nielsleadholm, line 142:
```suggestion
        limit of Monty (i.e., pose error of Monty's best potential hypothesis on the target object)
```

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -128,6 +133,71 @@ def _add_displacements(self, obs):
...
+        """
+        stats["max_evidence"] = {k: max(v) for k, v in self.evidence.items()}
+        stats["target_object_theoretical_limit"] = (
+            self._target_object_theoretical_limit()
+        )
+        stats["target_object_pose_error"] = self._target_object_pose_error()
+        return stats
+
+    def _target_object_theoretical_limit(self) -> float:
+        """Compute the theoretical minimum rotation error on the target object.
```
nielsleadholm, line 161: **suggestion:** See separate high-level comments on the PR about moving these to utility functions.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -128,6 +133,71 @@ def _add_displacements(self, obs):
...
+
+        This considers all possible hypotheses rotations on the target object
+        and compares them to the target's rotation. The theoretical limit conveys the
+        best achievable performance if Monty selects the best hypothesis as its most
+        likely hypothesis (MLH).
```
nielsleadholm, line 166: **suggestion:** It might be worth noting there is a caveat to this - it doesn't account for the location of the hypothesis. A particular rotation might be perfect in terms of geodesic distance to the ground-truth rotation, but if it's on a totally different part of the object, it's not actually a very good hypothesis. Accounting for location would be complex/hard to do, since the ground truth is not readily accessible. However just highlighting this caveat might help people grok unexpected results if using this metric in the future.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -128,6 +133,71 @@ def _add_displacements(self, obs):
...
+
+        Returns:
+            float: The minimum achievable rotation error (in radians).
+        """
+        poses = self.possible_poses[self.primary_target].copy()
+        hyp_rotations = Rotation.from_matrix(poses).inv().as_quat().tolist()
+        target_rotation = Rotation.from_quat(self.primary_target_rotation_quat)
```
nielsleadholm, line 173: **suggestion:** Minor point, but maybe do the inverse here, so you don't need to do it many times within the for-loop.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -128,6 +133,71 @@ def _add_displacements(self, obs):
...
+
+        min_error = np.pi
+        for rot in hyp_rotations:
+            rot = Rotation.from_quat(rot)
+            error = (rot * target_rotation.inv()).magnitude()
+            min_error = min(min_error, error)
+
+        return min_error
+
+    def _target_object_pose_error(self) -> float:
```
nielsleadholm, line 183: **suggestion:** I think in my head, comparing to the ground-truth/target is given by it being an "error", but this function name doesn't specify that it is the MLH whose error is being calculated. Might be worth renaming to something that highlights this more, like `_mlh_object_pose_error`? 

The earlier `_target_object_theoretical_limit` also doesn't specify it is a rotation error, so maybe that could be `_theoretical_limit_object_pose_error` to follow the same pattern?

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -128,6 +133,71 @@ def _add_displacements(self, obs):
...
+        """Compute the actual rotation error between predicted and target pose.
+
+        This compares the most likely hypothesis pose (based on evidence) on the target
+        object with the ground truth rotation of the target object.
+
+        Returns:
+            float: The rotation error (in radians).
+        """
+        target_rot = Rotation.from_quat(self.primary_target_rotation_quat)
+        obj_mlh_id = np.argmax(self.evidence[self.primary_target])
```
nielsleadholm, line 193: I think these two lines would be cleaner if they used the `get_current_mlh()` getter and just accessed the rotation, which should be stored in the returned dictionary.

nielsleadholm (commented):
Really nice, thanks for putting this together. Just some general comments, main one being the earlier discussion about moving the calculation methods into functions.

ramyamounir:
> make them utility functions that are imported and called in _add_detailed_stats, i.e. don't make them methods of a class at all?

@nielsleadholm Yes I agree. this would be a better place to keep these functions. Perhaps `utils/logging_utils.py`? there are somewhat similar functionality there, e.g., `get_pose_error`.

@tristanls Thanks for your review. The plotting functions are meant to be used as post-processing functions, similar to e.g., `plot_evidence_transitions` and many others in `plot_utils.py`. It will not run as part of the benchmark experiments. To use it, we can do something like:
```python
plot_objects_evidence_over_time("path/to/exp/logs")
#OR
plot_pose_error_over_time("path/to/exp/logs")
```
What is the best way to clarify this? We can add a section in the [Logging and Analysis](https://thousandbrainsproject.readme.io/docs/logging-and-analysis) docs, or perhaps just a comment at the top of the file?

nielsleadholm:
Great, yeah I think `logging_utils.py` would be a good location for the functions.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
ramyamounir: My thinking here was that it's important to highlight that this MLH is on the target object, not an absolute MLH across all objects. Monty's best hypothesis can be on the wrong object (especially after swapping the objects without reset), these functions get the MLH only on the ground truth object.

How about `_mlh_target_object_pose_error` and `_theoretical_limit_target_object_pose_error`?

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
ramyamounir: I think the `get_current_mlh()` will get the absolute mlh over all objects, but here I'm trying to get the mlh only for `self.primary_target`.

nielsleadholm: Ah fair point, then can you use `get_mlh_for_object`, passing in the `self.primary_target`?

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
nielsleadholm: Makes sense and sounds good!

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
ramyamounir: Nice, I wasn't aware of this function.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
ramyamounir: Yeah good point. I couldn't find a way to iterate over a Scipy `Rotation` object. But, I found a simpler way that avoids the for loop. I will apply the same simplification to the `get_pose_error()` in [logging_utils.py](https://github.com/thousandbrainsproject/tbp.monty/blob/4cc6a799b03fadd20ac8d9cfc8f2761345b19cfa/src/tbp/monty/frameworks/utils/logging_utils.py#L391) in a separate PR.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
ramyamounir: After simplifying the `_theoretical_limit_target_object_pose_error` and `_mlh_target_object_pose_error` functions, they ended up being 3 lines each. Only the last line from each function can be moved to utils. The first two lines prepare the predicted and target rotations, which would occur in the LM class. I'm not sure if it's worth moving these functions over to utils now, could you take another look at the simplified functions and lmk what you think?

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -128,6 +133,58 @@ def _add_displacements(self, obs):
...
+
+        This compares the most likely hypothesis pose (based on evidence) on the target
+        object with the ground truth rotation of the target object.
+
+        Returns:
+            float: The rotation error (in radians).
+        """
+        obj_rotation = self.get_mlh_for_object(self.primary_target)["rotation"].inv()
+        target_rotation = Rotation.from_quat(self.primary_target_rotation_quat)
+        error = (obj_rotation * target_rotation.inv()).magnitude()
```
vkakerbeck, line 185: Similarly here, both detected and target rotation are inverted. 
Would it be worth writing a helper function for calculating pose error (or using the existing `get_pose_error` function) and then calling that here? That way we make sure we always do the same calculations when retrieving the pose error. Something like 
```
def get_pose_error(target, predicted):
   error = (predicted * target.inv()).magnitude()
   return error
```

src/tbp/monty/frameworks/plots/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2025 Thousand Brains Project
```
vkakerbeck, line 1: Is there a particular reason you need a new folder for this instead of using the utils folder where we have the plot_utils.py file with all the other plotting functions? I get that this may be cleaner if we add a lot more plotting functionality. In that case, we should move the plot_utils.py file here as well (in a separate PR as this would also require documentation updates)

tests/unit/no_reset_evidence_lm_test.py
```diff
@@ -87,7 +87,7 @@ def setUp(self):
             # NOTE: could make unit tests faster by setting monty_log_level="BASIC" for
             # some of them.
             logging_config=LoggingConfig(
-                output_dir=self.output_dir, python_log_level="DEBUG"
+                output_dir=self.output_dir, python_log_level="DEBUG", monty_handlers=[]
```
vkakerbeck, line 90: Just out of curiosity, why do you need to remove the monty_handler here? Does it now take up too much memory? And is the logging now not tested anymore?

vkakerbeck (approved):
Nice! Just left a few clarifying questions and comments

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
nielsleadholm: @ramyamounir just checking if you saw this comment? Happy to discuss if you disagree.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
nielsleadholm: Hmm yeah that's a good question. I don't have a strong opinion, the way I see it we have two options:
1. Move the two lines of retrieval for each calculation into `_add_detailed_stats`, which then calls some kind of `get_pose_error` function in utils (if this can be compartmentalized, related to Viviane's comments). This has the advantage of not increasing the complexity of the class by adding new methods. A comment above each could provide context for their purpose.
2. Leave it as it is, except for the potential reuse of a `get_pose_error` at the end of each. This has the advantage of compartmentalizing the code more, which increases readability in some ways, but at the cost of the above responsibility/complexity issue. If we go for this approach, I would just include a `# ------------------ Logging & Saving ----------------------` above these methods as we've done in e.g. `EvidenceGraphLM` to make their responsibilities clear.

I'm fine with going for (2), but as this is a new type of LM with a reasonable amount of complexity, it would be worth adopting those organizing `------` sections as used in the other LMs. Note that they each have a Public and Private version.

On that note, I just noticed that ideally `_agent_moved_since_reset` in the previous PR should have been placed under 

```
    # ======================= Private ==========================

    # ------------------- Main Algorithm -----------------------
```

not the Main Algorithm public methods. Could also be worth amending that.

ramyamounir: Thanks, both options sound good to me. I had a chat with @tristanls yesterday and he suggested adding these logging functions as a Mixin implemented in a different file. I thought this was also a good idea, any thoughts? It would be named something like `NoResetEvidenceGraphLMLoggingMixin`.

The shared logic of calculating pose error from given rotations is just this one line `error=(r1*r2.inv()).magnitude()`. I could make this into a utils function and call it. It could be a good place to add some description about how this error is calculated and that `r1` could be a single rotation or a set of rotations. With type annotations and docstrings, it will be clear that these are Scipy Rotation objects as well.

Note: I will open a separate PR to simplify `get_pose_error` in `utils/logging_utils.py`. We should be able to remove the for loop there similar to how I did it in this PR.

ramyamounir: I will fix `_agent_moved_since_reset`'s location in this PR.

nielsleadholm: Sounds good, including using a mixin. With that approach, should it be called something like `TheoreticalLimitLMLoggingMixin` or `UnsupervisedInferenceLMLoggingMixin`? It could be used with a normal evidence LM I think, where it would highlight the poor theoretical limits. That would also fit better with it being a mixin (i.e. it's something intended for more than one class to use).

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
ramyamounir: Yeah, I saw it, just needed a bit more time to think through this one :sweat_smile: 

The way I've been thinking about it is that we store these rotation hypotheses at every location on the object. Predicting the correct rotation also means predicting the location where we sampled this rotation. This rotation can rotate the object until the sensor position matches the location on the object where we initialized this hypothesis. The sampled rotation depends on the location where we sampled it from.

Since we're not accounting for symmetry here, Monty will have to find the same exact rotation that forces the location to be correct (i.e., similar to the location where we initially sampled the hypothesis). Each sampled rotation describes how much we need to rotate the object to go back to the original orientation, and because these sampled rotations are at different locations, they also implicitly store the location. So to predict the correct rotation, you must have known where you are on the object.

I may be missing something big here.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
ramyamounir: Yes, good point. `TheoreticalLimitLMLoggingMixin` sounds good!

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
nielsleadholm: Ok yeah I realize the way I worded it was a bit confusing/overly strong. I didn't mean to suggest this measure doesn't account for location at all. However, it's possible to have a hypothesis that gets a good theoretical limit value (because it's rotation is similar), even though it isn't a good hypothesis, because the location might be off. 

Imagine the cracker box (large object). The rotation hypotheses initialized for two points on the same front surface of the box (one near the top, and one near the bottom) can be the same, because they can both have a global rotation of the object that is consistent with their local perception. For these hypotheses, the global rotation will not change unless the points are resampled. Let's assume that the sensor was actually near the top of the box. From the point of view of the theoretical limit, these hypotheses are both equally good, however only the one initialized near the top of the object will enable good predictions. 

I think it's worth pointing out, because it means that having a good theoretical limit is not *sufficient* for having a good hypothesis. Furthermore, because we don't account for symmetry, a good theoretical limit is also not *necessary* for a good hypothesis. Despite both of these points, it is still a really useful measure, it's just important that we don't get confused by results related to it. Let me know if I'm still not making sense.

ramyamounir: I see what you mean, thanks for the explanation!

Does this suggest that we may need to consider predicting the global translation of the hypotheses in addition to the global rotations? The theoretical limit here would be a function of comparing both, translation and rotation. Maybe we could discuss this later in one of the meetings. I'll add your comment to the theoretical limit calculation as suggested.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
ramyamounir: As mentioned in a previous discussion, if we end up factoring out the calculation of pose error, or using `get_pose_error`, I prefer to leave the double `.inv()` because they will be split into two functions and it will not be easy to understand why we canceled them out.

Note: I plan on simplifying the existing `get_pose_error()` to remove the unnecessary for-loop and use this simplified function to calculate pose error in the theoretical limit functions. I will do that in a separate PR after because modifying `get_pose_error` may require other changes in `add_evidence_lm_episode_stats` and `get_graph_lm_episode_stats`.

Tagging @nielsleadholm here for this note here as well.

src/tbp/monty/frameworks/plots/__init__.py
ramyamounir: No reason other than avoiding to contribute to the evergrowing `plot_utils.py`. With [PR#244](https://github.com/thousandbrainsproject/tbp.monty/pull/244), I think it is proposed that we keep each plot in a separate file. I may end up just deleting this `frameworks/plots` directory, if we agree on how it's done there.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
nielsleadholm: No worries, sounds good! 

And yeah I think it would be a good discussion topic. I don't think we currently have a way of measuring the ground-truth location of the object, and how it compares to an estimated location, but I think it would be possible to establish. That would be useful for metrics like this, but also tie in with hierarchy and discussions we've had (and never settled) about what kind of "object location" information we should pass to the next level of the hierarchy.

nielsleadholm: For what it's worth, storing rotations that are identified by Monty as symmetric has [been on our to-do-list for a while](https://thousandbrainsproject.readme.io/docs/improve-handling-of-symmetry). It could be a relatively simple thing to add which would enable accounting for it in the theoretical limit, as well as some interesting efficiency boosts when re-sampling (no need to sample multiple hypotheses if they are all from a symmetric set - one will do). Something to think about!

tests/unit/no_reset_evidence_lm_test.py
ramyamounir: The `monty_handlers` were giving errors here because it was looking for the graph_id key `self.primary_target` of "capsule3DSolid" in the logging functions I added in this PR (e.g., `_mlh_target_object_pose_error`). The problem is it wouldn't find this key because I was using unsupervised training to learn ["capsule3DSolid", "cubeSolid"] as ["new_object0", "new_object1"]. I've fixed it properly now by using `MontySupervisedObjectPretrainingExperiment` for pretraining and `MontyForNoResetEvidenceGraphMatching` for evaluation.

ramyamounir:
Thanks @nielsleadholm and @vkakerbeck for the detailed reviews. I've pushed a few commits to address your concerns.

Notes:
* I will open a follow-up PR for refactoring the pose error calculations in `logging_utils.py` and calling this function from the Mixin.
* I will wait to merge this PR until after [PR#244](https://github.com/thousandbrainsproject/tbp.monty/pull/244) has been merged. I may end up deleting the plots directory in favor of the plotting tool folder structure proposed by @tristanls.

tristanls:
@ramyamounir , you don't have to wait for #244. I can add deleting your plot fikes to that PR.

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -0,0 +1,73 @@
...
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+
+from typing import Any, Dict
+
+from scipy.spatial.transform import Rotation
+
+
+class TheoreticalLimitLMLoggingMixin:
```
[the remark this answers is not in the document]
ramyamounir: Sure, I added some description.

nielsleadholm (approved):

Issue #117, opened 2024-12-17
test_5lm_feature_experiment fails


Teuling:
### Describe the bug

tests/unit/graph_learning_test.py:1714: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/tbp/monty/frameworks/experiments/monty_experiment.py:522: in train
    self.run_epoch()
src/tbp/monty/frameworks/experiments/monty_experiment.py:486: in run_epoch
    self.run_episode()
src/tbp/monty/frameworks/experiments/object_recognition_experiments.py:41: in run_episode
    last_step = self.run_episode_steps()
src/tbp/monty/frameworks/experiments/object_recognition_experiments.py:106: in run_episode_steps
    self.model.step(observation)
src/tbp/monty/frameworks/models/monty_base.py:141: in step
    self._matching_step(observation)
src/tbp/monty/frameworks/models/abstract_monty_classes.py:25: in _matching_step
    self._vote()
src/tbp/monty/frameworks/models/graph_matching.py:428: in _vote
    self.send_vote_to_lm(self.learning_modules[i], i, combined_votes)
src/tbp/monty/frameworks/models/graph_matching.py:82: in send_vote_to_lm
    lm.receive_votes(combined_votes[lm_id])
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <tbp.monty.frameworks.models.feature_location_matching.FeatureGraphLM object at 0x7ab2462f7730>
vote_data = {'neg_object_id_votes': {'new_object0': 0}, 'pos_location_votes': {'new_object0': array([[ 1.01622774e-02,  1.49800395...tion.Rotation object at 0x7ab23f4cbe40>, <scipy.spatial.transform._rotation.Rotation object at 0x7ab23f4ed030>, ...]]}}

    def receive_votes(self, vote_data):
        """Use votes to remove objects and poses from possible matches.
    
        NOTE: Add object back into possible matches if majority of other modules
                think it is correct? Could help with dealing with noise but may
                also prevent LMs from narrowing down quickly. Since we are not
                working with this LM anymore, we probably wont add that.
    
        Args:
            vote_data: positive and negative votes on object IDs + positive
[...]

PR #228, opened 2025-04-04
chore: upgrade Python and HabitatSim

jeremyshoemaker:
This updates the version of HabitatSim so that we can update to a newer, supported version of Python.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -22,9 +22,9 @@ channels:
   - defaults
 
 dependencies:
-  - python=3.8
+  - python=3.9
 
-  - aihabitat::habitat-sim=0.2.2
+  - aihabitat::habitat-sim=0.2.4
   - aihabitat::withbullet
   - conda-forge::cmake>=3.14.0
   - conda-forge::importlib_resources
@@ -48,4 +48,3 @@ dependencies:
   - pip
   - pip:
       - -e .[dev]
-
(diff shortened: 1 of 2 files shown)


jeremyshoemaker:
Just upgrading `habitat-sim` to version 0.2.4 leads to a single test failure (at least on MacOS). Specifically [this assert fails](https://github.com/thousandbrainsproject/tbp.monty/blob/f67f439cccd5cc19bba05e329e39f648f1e4855c/tests/unit/evidence_lm_test.py#L1173-L1177) because the first value is negated in the new version for some reason.

```
>       self.assertListEqual(
            list(graph_lm.get_possible_poses()["new_object0"][-1]),
            [180.0, 0.0, 180.0],
            "Since have symmtry here 180, 0, 180 should also be a possible pose.",
        )
E       AssertionError: Lists differ: [-180.0, 0.0, 180.0] != [180.0, 0.0, 180.0]
E       
E       First differing element 0:
E       -180.0
E       180.0
E       
E       - [-180.0, 0.0, 180.0]
E       ?  -
E       
E       + [180.0, 0.0, 180.0] : Since have symmtry here 180, 0, 180 should also be a possible pose.
```

jeremyshoemaker:
After a lot of false starts, and some efforts by @hlee9212, we decided to set up an action runner that has a GPU since those are available now.

Doing this still leads to the one test failure I mentioned above, but it also **finally** reproduces the 5LM test failure first reported in #117 that I've been working on rewriting.

So, that's progress.

I think we're going to need the fix that before we can finish this work.

PR #231, opened 2025-04-07
refactor: first movement logic in GraphLM

ramyamounir:
This change refactors the logic of detecting the first movement since the reset signal from `GraphLM.matching_step` to `GraphLM._agent_moved_since_reset`. This will allow us to overwrite this logic in other PRs (e.g., [PR#221](https://github.com/thousandbrainsproject/tbp.monty/pull/221#discussion_r2021342279)) without modifying the `matching_step` function.

Additionally, `not_moved` has been renamed to `first_movement_detected` to avoid the slightly confusing double negatives in cases such as `not_moved = False`.

This change required other changes in `DisplacementGraphLM` and `EvidenceSDRIntegrationTest`.

I ran the benchmark experiments for `randrot_noise_10distinctobj_surf_agent` and `randrot_noise_10distinctobj_dist_agent` for a quick sanity check; they gave the same results as before the change.

### DIFF
--- a/src/tbp/monty/frameworks/models/graph_matching.py
+++ b/src/tbp/monty/frameworks/models/graph_matching.py
@@ -994,7 +994,7 @@ class GraphLM(LearningModule):
 
         Args:
             observations: Observations to use for computing possible matches.
-            first_movement_detected: Whether the agent has moved since the reset
+            first_movement_detected: Whether the agent has moved since the buffer reset
                 signal.
         """
         if first_movement_detected:
--- a/src/tbp/monty/frameworks/models/graph_matching.py
+++ b/src/tbp/monty/frameworks/models/graph_matching.py
@@ -21,10 +21,7 @@ from tbp.monty.frameworks.loggers.graph_matching_loggers import (
     DetailedGraphMatchingLogger,
     SelectiveEvidenceLogger,
 )
-from tbp.monty.frameworks.models.abstract_monty_classes import (
-    LearningModule,
-    LMMemory,
-)
+from tbp.monty.frameworks.models.abstract_monty_classes import LearningModule, LMMemory
 from tbp.monty.frameworks.models.buffer import FeatureAtLocationBuffer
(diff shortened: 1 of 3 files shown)


src/tbp/monty/frameworks/models/graph_matching.py
```diff
@@ -21,10 +21,7 @@
     DetailedGraphMatchingLogger,
     SelectiveEvidenceLogger,
 )
-from tbp.monty.frameworks.models.abstract_monty_classes import (
-    LearningModule,
-    LMMemory,
-)
+from tbp.monty.frameworks.models.abstract_monty_classes import LearningModule, LMMemory
```
vkakerbeck, line 24: It would be good to turn off your autoformatter for lines that you didn't change so that PRs only contain the relevant lines to review

src/tbp/monty/frameworks/models/graph_matching.py
```diff
@@ -988,22 +989,23 @@ def load_state_dict(self, state_dict):
...
 
     # ------------------- Main Algorithm -----------------------
-    def _compute_possible_matches(self, observations, not_moved=False):
+    def _compute_possible_matches(self, observations, first_movement_detected=True):
         """Use graph memory to get the current possible matches.
 
         Args:
             observations: Observations to use for computing possible matches.
-            not_moved: Whether the observations are not moved.
+            first_movement_detected: Whether the agent has moved since the reset
```
vkakerbeck, line 997:
```suggestion
            first_movement_detected: Whether the agent has moved since the buffer reset
```

vkakerbeck (approved):
Very nice, this does look a lot cleaner and easier to understand :)

nielsleadholm (approved):
Really nice, thanks for making this change even though it was a bit of a detour!

src/tbp/monty/frameworks/models/graph_matching.py
ramyamounir: Yes, I still haven't looked into this. I'll do it before the next PR.

PR #232, opened 2025-04-08
refactor!: remove unused MotorSystem.update method

tristanls:
While this is marked as a breaking change (removing a method from the API), there are no subclass implementations of this abstract method, meaning that if it were ever called, it would have thrown an error. Since we observe no such errors, there are no calls of this method in the code, and there's a `TODO` questioning needing the method, this method is now removed.


vkakerbeck (approved):
nice, that looks good to remove

PR #233, opened 2025-04-08
chore: clean up ruff's ignore list to remove rules that pass

hlee9212:
I wrote a short script to loop through the ignore list under [tool.ruff.lint] and remove those that pass. 

Core logic of script for reference:
```python
def extract_ignore_rules(pyproject_path: Path) -> List[str]:
    """Extract the ruff ignore rules from pyproject.toml."""

    with open(pyproject_path, "rb") as f:
        data = tomli.load(f)

    ignore_rules = data["tool"]["ruff"]["lint"]["ignore"]
    return ignore_rules


def check_rule(rule: str, src_dirs: List[str], tbp_monty_dir: Path) -> bool:
    """
    Check if a rule produces any errors when run with ruff check.
    Returns True if no errors are found (rule can be removed).
    """
    cmd = ["ruff", "check"] + src_dirs + ["--select", rule]

    try:
        # Run the command and capture output
        result = subprocess.run(
            cmd,
            cwd=tbp_monty_dir,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            text=True,
            check=False,
        )

        # If the exit code is 0, no issues were found
        return result.returncode == 0
    except Exception as e:
        print(f"Error checking rule {rule}: {e}", file=sys.stderr)
        return False
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -187,7 +187,6 @@ ignore = [
     # Original inherited flake8 ignores
     "C4", # C4XX: ignore all Flake8 comprehensions
     "D1", # D1XX: Missing Docstrings
-    "E203", # E203: whitespace before ':'
     "ERA001", # ERA001: Found commented out code
     "F541", # F541: f-string is missing placeholders
     "FIX001", # FIX001: Line contains FIXME, consider resolving the issue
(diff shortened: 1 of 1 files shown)


tristanls (approved):
Rules are not ignored, and style checks pass; looks good.

PR #235, opened 2025-04-08
refactor!: remove setting unused MotorSystem.step_type attribute

tristanls:
The `MotorSystem` `step_type` attribute is unused but is set as part of `MontyBase` methods (and incorrectly). This pull request removes setting the unused attribute.


vkakerbeck (approved):
Yes, that was definitely a typo/copy paste error that we don't set it to matching_step in the first instance. I think the idea was that depending on the step type, the motor system would select different actions or use a different policy. Currently this is not used so I am happy with removing it. When we add swapping of policies in the motor system we might want something similar back or have a different mechanism to switch between matching and exploration policies.

PR #237, opened 2025-04-09
chore: run tools workflows when pyproject.toml changes

tristanls:
Tools workflow does not run when `pyproject.toml` changes, so if linting or other tooling configuration changes (for example: https://github.com/thousandbrainsproject/tbp.monty/pull/233), it won't be noticed until something alters the tools.

This pull request ensures tools checks run when `pyproject.toml` changes.

The newly effective linter rules are addressed for checks to pass in `tools`.

### DIFF
--- a/tools/github_readme_sync/tests/hierarchy_test.py
+++ b/tools/github_readme_sync/tests/hierarchy_test.py
@@ -8,6 +8,7 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
+import contextlib
 import http.server
 import os
 import socketserver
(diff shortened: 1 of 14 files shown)


tools/github_readme_sync/tests/hierarchy_test.py
```diff
@@ -178,10 +179,8 @@ def run_server():
             f.write(f"[Fragment]({self.server_url}/valid#fragment)\n")
 
         with self.assertLogs(level="ERROR") as log:
-            try:
+            with contextlib.suppress(SystemExit):
```
codeallthethingz, line 182: oh this is cool!

codeallthethingz (approved):
LGTM

tools/github_readme_sync/tests/hierarchy_test.py
tristanls: Yeah, Ruff told me to do it https://docs.astral.sh/ruff/rules/suppressible-exception/

PR #243, opened 2025-04-10
chore: only strings should be added to sys.path

tristanls:
Follow on to #237, where I incorrectly add `Path` objects to `sys.path` instead of strings.

> A program is free to modify this list for its own purposes. **Only strings and bytes** should be added to [sys.path](https://docs.python.org/3.8/library/sys.html#sys.path); all other data types are ignored during import.
> -- https://docs.python.org/3.8/library/sys.html#sys.path


codeallthethingz (approved):
👍

PR #244, opened 2025-04-10
chore: initial implementation of plot tool

tristanls:
This is an initial implementation of a plot tool. It supports plotting figures proposed in #227.

```
(tbp.monty) tslominski@Overlords-MacBook-Pro tbp.monty % python -m tools.plot.cli -h                                     
usage: cli.py [-h] [--debug] {objects_evidence_over_time,pose_error_over_time} ...

Plot a figure

positional arguments:
  {objects_evidence_over_time,pose_error_over_time}
    objects_evidence_over_time
                        Plot evidence scores for each object over time.
    pose_error_over_time
                        Plot MLH pose error and theoretical limits over time.

optional arguments:
```

When we added new plots in https://github.com/thousandbrainsproject/tbp.monty/pull/227, I noticed how different that code is from anything in the framework. Yes, it depends on specific logging output in the framework, but other than that, it has different dependencies, and it is mostly `matplotlib` configuration and log file parsing code. Additionally, as a heuristic, whenever I see the string `utils`, it usually highlights to me a group of functionality that likely belongs elsewhere.

If we keep plot utilities in the tools directory, we'll still be able to keep them in sync with the rest of the framework. We can set it up so that the plot tool tests run with Monty tests in case we change something in logging, although there's only a `get_action_name` test.

As a tool, I think plot code will still be usable in a Jupyter Notebook via something like `from tools.plot.objects_evidence_over_time import objects_evidence_over_time` while reducing the dependencies needed for the framework.

https://github.com/user-attachments/assets/de28ec6f-8f2b-48ad-8d55-b35c0e68b134

### DIFF
--- a/tools/plot/README.md
+++ b/tools/plot/README.md
@@ -1,0 +1,38 @@
+# Plot a figure
+
+A tool that plots a figure from experiment log output.
+
+## Setup
+
+Ensure you setup Monty. See [Getting Started - 2. Set up Your Environment](https://thousandbrainsproject.readme.io/docs/getting-started#2-set-up-your-environment).
+
+No additional dependencies are required.
+
+## Usage
+
+```
+$ python -m tools.plot.cli -h
+
+usage: cli.py [-h]
+
+usage: cli.py [-h] [--debug] {objects_evidence_over_time,pose_error_over_time} ...
+
+Plot a figure
+
+positional arguments:
+  {objects_evidence_over_time,pose_error_over_time}
+    objects_evidence_over_time
+                        Plot evidence scores for each object over time.
+    pose_error_over_time
+                        Plot MLH pose error and theoretical limits over time.
+
+optional arguments:
+  -h, --help            show this help message and exit
+  --debug               Enable debug logging
+```
+
+## Tests
+
+```
+pytest
+```
--- a/tools/plot/README.md
+++ b/tools/plot/README.md
@@ -31,6 +31,20 @@ optional arguments:
   --debug               Enable debug logging
 ```
 
+### Examples
+
+The tool can be invoked in multiple ways for convenience:
+
+```
+tbp.monty$ python -m tools.plot.cli pose_error_over_time ~/your_experiment_log_directory
+
+tbp.monty$ python tools/plot/cli.py pose_error_over_time ~/your_experiment_log_directory
+
+tbp.monty/tools$ python plot/cli.py pose_error_over_time ~/your_experiment_log_directory
+
+tbp.monty/tools/plot$ python cli.py pose_error_over_time ~/your_experiment_log_directory
+```
+
 ## Tests
 
 ```
--- a/tools/plot/cli.py
+++ b/tools/plot/cli.py
@@ -12,10 +12,10 @@ import logging
 import sys
 from pathlib import Path
 
-from tools.plot import objects_evidence_over_time, pose_error_over_time
-
 monty_root = Path(__file__).resolve().parent.parent.parent
 sys.path.append(str(monty_root))
+
+from tools.plot import objects_evidence_over_time, pose_error_over_time  # noqa: E402
 
 
 def main():
--- a/tools/plot/cli.py
+++ b/tools/plot/cli.py
@@ -1,0 +1,58 @@
+# Copyright 2025 Thousand Brains Project
+#
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+
+import argparse
+import logging
+import sys
+from pathlib import Path
+
+from tools.plot import objects_evidence_over_time, pose_error_over_time
+
+monty_root = Path(__file__).resolve().parent.parent.parent
+sys.path.append(str(monty_root))
+
+
+def main():
+    parent_parser = argparse.ArgumentParser(add_help=False)
+    parent_parser.add_argument(
+        "--debug",
+        action="store_true",
+        help="Enable debug logging",
+        dest="debug",
+    )
+
+    parser = argparse.ArgumentParser(
+        description="Plot a figure", parents=[parent_parser]
+    )
+    subparsers = parser.add_subparsers(dest="command")
+
+    objects_evidence_over_time.add_subparser(subparsers, parent_parser)
+    pose_error_over_time.add_subparser(subparsers, parent_parser)
+
+    args = parser.parse_args()
+
+    log_level = logging.DEBUG if args.debug else logging.INFO
+    handler = logging.StreamHandler(sys.stderr)
+    logging.basicConfig(
+        level=log_level,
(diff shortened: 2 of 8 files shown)


tristanls:
FYI: @hlee9212 @scottcanoe

tools/plot/pose_error_over_time.py
```diff
@@ -0,0 +1,195 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+
+from tbp.monty.frameworks.utils.logging_utils import load_stats
+
+logger = logging.getLogger(__name__)
+
+
+def pose_error_over_time(exp_path: str) -> int:
```
vkakerbeck, line 26: Why don't these functions utilize the functions that Ramy added in `frameworks/plots/`? Are you proposing that we remove all the plotting code from Monty src and move it into the tools folder?

tristanls: Yes, that's what I'm proposing.

tristanls:
> (...) Just a clarifying question about what should be done with the plotting code in monty's framework code. Should we also move the code from plot_utils.py over here and have those plots executable from the tools dir? The current usage is described here: https://thousandbrainsproject.readme.io/docs/logging-and-analysis#analysis

@vkakerbeck, yes, I think we should move that code over to the tool. We can do it gradually or all at once.

When we added new plots in #227, I noticed how different that code is from anything in the framework. Yes, it depends on specific logging output in the framework, but other than that, it has different dependencies, and it is mostly `matplotlib` configuration and log file parsing code. Additionally, as a heuristic, whenever I see the string `utils`, it usually highlights to me a group of functionality that likely belongs elsewhere.

If we keep plot utilities in the tools directory, we'll still be able to keep them in sync with the rest of the framework. We can set it up so that the plot tool tests run with Monty tests in case we change something in logging, although there's only a `get_action_name` test.

As a tool, I think plot code will still be usable in a Jupyter Notebook via something like `from tools.plot.objects_evidence_over_time import objects_evidence_over_time` while reducing the dependencies needed for the framework.

ramyamounir (approved):
Nice! I like this very much. 

Would it make sense if, in the future, we consider separating the tool cli code into a separate package that gets installed as a dependency of tbp.monty. The folder would then just have plotting functions and none of the cli parsing stuff. 

It would be similar in structure to pytest. Then we'd do something like `tbp.plot [plot_name] [exp_path]`

tristanls:
> Would it make sense if, in the future, we consider separating the tool cli code into a separate package that gets installed as a dependency of tbp.monty. The folder would then have plotting functions and none of the cli parsing stuff.

@ramyamounir, maybe, but there are benefits of keeping plotting code as part of `tbp.monty`. We can test the tool at the same time we test the framework, and plotting being heavily tied to the logging format, makes me want to keep these together for now to see if we want to integrate testing the two together better.

Since there aren't many tests, maybe this won't have as big of an impact. I'm not against separation in the long term, but I'd like to see how we interact with the tool before separating it out completely.

tools/plot/README.md
```diff
@@ -0,0 +1,38 @@
+# Plot a figure
+
+A tool that plots a figure from experiment log output.
+
+## Setup
+
+Ensure you setup Monty. See [Getting Started - 2. Set up Your Environment](https://thousandbrainsproject.readme.io/docs/getting-started#2-set-up-your-environment).
+
+No additional dependencies are required.
```
nielsleadholm, line 9: **suggestion:** Just because someone using our framework might find it helpful, maybe include a concrete example of a command like:

```bash
python -m tools.plot.cli pose_error_over_time ~/your_data_director
```

tools/plot/cli.py
```diff
@@ -0,0 +1,58 @@
...
+        dest="debug",
+    )
+
+    parser = argparse.ArgumentParser(
+        description="Plot a figure", parents=[parent_parser]
+    )
+    subparsers = parser.add_subparsers(dest="command")
+
+    objects_evidence_over_time.add_subparser(subparsers, parent_parser)
+    pose_error_over_time.add_subparser(subparsers, parent_parser)
```
jeremyshoemaker, line 36: I'm confused. How does this work? Looking at [the `argparse` documentation](https://docs.python.org/3/library/argparse.html#sub-commands), I don't see anything about this feature.

It looks like you're calling `add_parser` on a function object, but where is that defined?

---

Never mind, I see where it's defined. That is very confusing because the module and the function in the module have the same name, and I didn't see that we were importing the module and not the function from the module.

I feel like this could be handled better to prevent confusion but I'm not sure the best way to do that. Maybe rename the functions to not match the module names?

tools/plot/cli.py
```diff
@@ -0,0 +1,58 @@
...
+
+import argparse
+import logging
+import sys
+from pathlib import Path
+
+from tools.plot import objects_evidence_over_time, pose_error_over_time
+
+monty_root = Path(__file__).resolve().parent.parent.parent
+sys.path.append(str(monty_root))
```
nielsleadholm, line 18: **question:** Do we actually need to use this `sys.path.append` approach here? The `.parent.parent.parent` line looks a little fragile.

jeremyshoemaker (approved):
There's some ambiguity in how the sub parsers are defined that I feel could be handled better, but I'm not gonna block this just for that. See my comment below for details.

nielsleadholm (approved):
Really nice, thanks for starting this change! Just a couple minor suggestions.

tools/plot/cli.py
jeremyshoemaker: I was wondering that myself, but I haven't written many CLI tools in Python to know what might be a better way to handle it.

nielsleadholm: Yeah I'm not sure either, just thought it might be something that our package installation might already / theoretically be able to handle?

nielsleadholm: Great thanks for the context.

tools/plot/README.md
tristanls: Good point, thank you. I updated the README with examples.

tools/plot/objects_evidence_over_time.py
```diff
@@ -0,0 +1,146 @@
...
+    particular target object is active.
+
+    Args:
+        exp_path (str): Path to the experiment directory containing the detailed stats
+            file.
+
+    Returns:
+        int: Exit code.
+    """
+    if not os.path.exists(exp_path):
```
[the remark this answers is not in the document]
tristanls: Thank you, great point. Updated.

tools/plot/pose_error_over_time.py
vkakerbeck: okay, makes sense. Now that Ramy's code is merged into the plots folder, should this PR delete that to avoid having the same code in two places?

tristanls: I have a draft ready to go after we merge this one: https://github.com/thousandbrainsproject/tbp.monty/pull/247. I didn't want to sneak in deletions after so many approvals already.

PR #245, opened 2025-04-11
fix: Fix Omniglot Environment

vkakerbeck:
As I was writing the Monty for custom applications tutorial, I noticed that the Omniglot environment was not working properly anymore. This is largely due to the fact that we don't have unit tests for it, so previous changes that broke it were unnoticed. These fixes make it work as intended again. Followup work would include actually testing it, however I think it is a bit more involved since we don't want to have to download the dataset for the unit tests (maybe there is an easy solution I am not aware of).

### DIFF
--- a/src/tbp/monty/frameworks/environments/two_d_data.py
+++ b/src/tbp/monty/frameworks/environments/two_d_data.py
@@ -118,7 +118,7 @@ class OmniglotEnvironment(EmbodiedEnvironment):
         Returns:
             observation (dict).
         """
-        amount = 1 if action.rotation_degrees < 1 else action.rotation_degrees
+        amount = max(action.rotation_degrees, 1)
         self.step_num += int(amount)
         query_loc = self.locations[self.step_num % self.max_steps]
         patch = self.get_image_patch(
--- a/src/tbp/monty/frameworks/environments/two_d_data.py
+++ b/src/tbp/monty/frameworks/environments/two_d_data.py
@@ -111,8 +111,9 @@ class OmniglotEnvironment(EmbodiedEnvironment):
         different points on the second pass.
 
         Args:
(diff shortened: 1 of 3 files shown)


src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -111,14 +111,14 @@ def step(self, _action, amount):
...
+            action: Not used at the moment since we just follow the draw path. However,
+            we do use the rotation_degrees to determine the amount of pixels to move at
+            each step.
 
         Returns:
             observation (dict).
         """
-        if amount < 1:
-            amount = 1
+        amount = max(action.rotation_degrees, 1)
```
tristanls, line 121: **issue**: Not every `Action` has `rotation_degrees`.

**suggestion**: 
```python
amount = 1
if hasattr(action, "rotation_degrees"):
    amount = max(action.rotation_degrees, 1)
```

tristanls:
@vkakerbeck, I think this PR qualifies as a `fix:`. Since Omniglot is part of the framework, this fixes a bug in the codebase.

vkakerbeck:
Ah okay yeah makes sense. I was looking for what would be most appropriate but didn't see fix in the recent PR titles. Can we maybe add the conventions into our style guide https://thousandbrainsproject.readme.io/docs/style-guide ?

tristanls:
> Ah okay yeah makes sense. I was looking for what would be most appropriate but didn't see fix in the recent PR titles. Can we maybe add the conventions into our style guide https://thousandbrainsproject.readme.io/docs/style-guide ?

@vkakerbeck, I created [RFC Conventional Commits](https://github.com/thousandbrainsproject/tbp.monty/pull/246) to official adopt if this is what we want to do.

tristanls (approved):

PR #246, opened 2025-04-11
RFC 10 Conventional Commits

tristanls:
Proposal to adopt conventional commits.

### DIFF
--- a/rfcs/0000_conventional_commits.md
+++ b/rfcs/0000_conventional_commits.md
@@ -17,6 +17,9 @@ Adopting conventional commits allows us to rapidly determine what version ([RFC 
 
 ## Commit Messages
 
+> [!NOTE]
+> The following commit message guidance applies only to commits made to the `main` branch when a Pull Request is merged. This guidance does not apply to commit messages elsewhere, for example, on your working branch while working on a Pull Request.
+
 The commit message to `main` branch should be structured as follows:
 
 ```
--- a/rfcs/0000_conventional_commits.md
+++ b/rfcs/0000_conventional_commits.md
@@ -29,15 +29,18 @@ The commit message to `main` branch should be structured as follows:
 
 ### Types
 
+> [!NOTE]
+> As of writing, the **Public API** referenced below consists of codebase in **src/tbp/monty**, **benchmarks**, and **tools/plot**. The definition of Public API may change in the future.
+
 `tbp.monty` code adopts the following `<type>`s:
 
-- `fix`: Fix to a bug in the **src/tbp/monty** codebase. This correlates with `PATCH` in [RFC 7 - Monty versioning](0007_monty_versioning.md).
-- `feat`: Introduction of a new feature to the **scr/tbp/monty** codebase. This correlates with `MINOR` in [RFC 7 - Monty versioning](0007_monty_versioning.md).
+- `fix`: Fix to a bug in the **Public API**. This correlates with `PATCH` in [RFC 7 - Monty versioning](0007_monty_versioning.md).
+- `feat`: Introduction of a new feature to the **Public API**. This correlates with `MINOR` in [RFC 7 - Monty versioning](0007_monty_versioning.md).
 - `build`: Change to the build system or external dependencies.
 - `ci`: Change to our GitHub Actions confguration files and scripts.
 - `docs`: Documentation only update.
 - `perf`: Performance improvement.
-- `refactor`: A **src/tbp/monty** code change that neither fixes a bug nor adds a feature.
+- `refactor`: A **Public API** code change that neither fixes a bug nor adds a feature.
 - `style`: Change that does not affect the meaning of the code (white-space, formatting, etc.).
 - `test`: Adding or correcting tests.
 - `chore`: The commit is a catch all for work outside of the types identified above. For example, the commit affects infrastructure, tooling, development, or other non-Monty framework code.
@@ -51,9 +54,9 @@ Even with the above guidance, sometimes there might be doubt or disagreement on 
 A breaking change is communicated by appending `!` after the type/scope. This correlates with `MAJOR` in [RFC 7 - Monty versioning](0007_monty_versioning.md).
 
 > [!NOTE]
-> `fix`, `feat`, and `refactor` types refer only to the **src/tbp/monty** codebase. Adding a new tool is not a feature. Fixing a tool is not a feature. Refactoring a tool is not a refactor of `src/tbp/monty`.
+> `fix`, `feat`, and `refactor` types refer only to the **Public API** of the codebase. As of writing, the Public API consists of **src/tbp/monty**, **benchmarks**, and **tools/plot**. Any work outside of the "public API", does not qualify for `fix`, `feat`, or `refactor` types.
 >
-> By restricting these types to **src/tbp/monty**, it enables us to rapidly distinguish when we need to increment the `tbp.monty` version for publishing. Only `fix`, `feat`, `refactor` commits will be relevant to determine whether a `MINOR` or `MAJOR` version increment is required.
+> By restricting these types to the Public API, it enables us to rapidly distinguish when we need to increment the `tbp.monty` version for publishing. Only `fix`, `feat`, `refactor` commits will be relevant to determine whether a `MINOR` or `MAJOR` version increment is required.
 >
 > By default, a version increment is `PATCH`. If there is a `feat` commit present, then the version increment is `MINOR`. If there is a breaking change commit present: `fix!`, `feat!`, `refactor!`, then the version increment is `MINOR` if and only if `MAJOR == 0`, and it is `MAJOR` otherwise.
 
--- a/rfcs/0000_conventional_commits.md
+++ b/rfcs/0000_conventional_commits.md
@@ -1,0 +1,86 @@
+- Start Date: 2025-04-11
+- RFC PR: (leave this empty, it will be filled in after RFC is merged)
+
+# Summary
+
+[Conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) allow for rapid human and automated interpretation of commits.
+
+# Motivation
+
+As of this writing, we largely adopted the habit of labeling our commits as per https://www.conventionalcommits.org/en/v1.0.0/. 22 of the last 23 commits were done formatted in conventional commit style and currently active PRs authored by the TBP team adopt conventional commits as the pull request name.
+
+# Conventional Commit Style
+
+Monty commits use the [conventional commits 1.0.0 standard](https://www.conventionalcommits.org/en/v1.0.0/).
+
+## Commit Messages
+
+The commit message to `main` branch should be structured as follows:
+
+```
+<type>[optional scope]: <description>
+
+[optional body]
+
+[optional footer(s)]
+```
+
+### Types
+
+Monty code adopts the following `<type>`s:
+
+- `fix`: Fix to a bug in the **src/tbp/monty** codebase. This correlates with `PATCH` in [RFC 7 - Monty versioning](0007_monty_versioning.md).
+- `feat`: Introduction of a new feature to the **scr/tbp/monty** codebase. This correlates with `MINOR` in [RFC 7 - Monty versioning](0007_monty_versioning.md).
+- `build`: Change to the build system or external dependencies.
+- `ci`: Change to our GitHub Actions confguration files and scripts.
+- `docs`: Documentation only update.
+- `perf`: Performance improvement.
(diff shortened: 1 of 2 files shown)


vkakerbeck (approved):
Nice, this will make things much easier to parse (as they already are with us unofficially using this). Once this RFC is merged, can you make sure to update the documentation accordingly?

rfcs/0000_conventional_commits.md
```diff
@@ -0,0 +1,86 @@
...
+# Conventional Commit Style
+
+Monty commits use the [conventional commits 1.0.0 standard](https://www.conventionalcommits.org/en/v1.0.0/).
+
+## Commit Messages
+
+The commit message to `main` branch should be structured as follows:
+
+```
+<type>[optional scope]: <description>
```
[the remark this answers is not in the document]
tristanls: I added the following examples, does that help clarify what these are?

### Commit message with scope

> [!NOTE]
> We do not use scopes.

```
feat(motor_sys): add MotorModule
```

### Commit message with scope and `!` to draw attention to breaking change

> [!NOTE]
> We do not use scopes.

```
feat(api)!: on the wire CMP
```

### Commit message with multi-paragraph body and multiple footers

> [!NOTE]
> We typically don't use these types of commits.

```
fix: prevent racing of requests

Introduce a request id and a reference to latest request. Dismiss
incoming responses other than from latest request.

Remove timeouts which were used to mitigate the racing issue but are
obsolete now.

Reviewed-by: Z
Refs: #123
```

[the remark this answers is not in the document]
vkakerbeck: great, thanks! This clarifies things

rfcs/0000_conventional_commits.md
```diff
@@ -0,0 +1,143 @@
...
+[optional body]
+
+[optional footer(s)]
+```
+
+### Types
+
+Monty code adopts the following `<type>`s:
+
+- `fix`: Fix to a bug in the **src/tbp/monty** codebase. This correlates with `PATCH` in [RFC 7 - Monty versioning](0007_monty_versioning.md).
```
nielsleadholm, line 32: **suggestion:** "Corresponds" might be a clearer word to use than correlates (also on the next line), assuming there is meant to be a 1:1 mapping. If `fix` is an _example_ of a `PATCH` change, then I would suggest saying something along those lines.

rfcs/0000_conventional_commits.md
```diff
@@ -0,0 +1,143 @@
...
+
+Monty does not use optional scopes.
+
+### Description
+
+This is the usual one-line message summarizing the commit.
+
+### Optional Body
+
+Monty, usually, does not use optional body.
```
nielsleadholm, line 71: **nitpick:** Is it possible to change "Monty" here to `tbp.monty`? (also on line 75) I feel like it overloads a term that usually refers to Monty as an intelligent system rather than a codebase.

rfcs/0000_conventional_commits.md
```diff
@@ -0,0 +1,143 @@
...
+
+Reviewed-by: Z
+Refs: #123
+```
+
+## Pull Requests
+
+By default, Pull Request titles are used as the suggested commit message when merging. **Maintainers** are responsible for ensuring that any merges to `main` contain a conventional commit message as defined in this RFC.
+
+It is helpful to rename Pull Requests to match the intended commit message to communicate the scope of the Pull Request and to reduce the chance of mistakes in commit message when merging. However, we will not enforce a Pull Request naming standard so as not to introduce an additional barrier to contributors.
```
nielsleadholm, line 143: **suggestion:** I think it's worth adding a note that this does not affect the commit messages people make in their own branches. I.e. you don't need to follow this convention every-time you make a random commit on your working branch, it is only when you move to merge a PR into main. I know this is implicitly covered by the description already here, but I think it is worth spelling this out to avoid confusion.

nielsleadholm (approved):
Looks great! Just a few minor suggestions.

rfcs/0000_conventional_commits.md
```diff
@@ -0,0 +1,143 @@
...
+### Types
+
+Monty code adopts the following `<type>`s:
+
+- `fix`: Fix to a bug in the **src/tbp/monty** codebase. This correlates with `PATCH` in [RFC 7 - Monty versioning](0007_monty_versioning.md).
+- `feat`: Introduction of a new feature to the **scr/tbp/monty** codebase. This correlates with `MINOR` in [RFC 7 - Monty versioning](0007_monty_versioning.md).
+- `build`: Change to the build system or external dependencies.
+- `ci`: Change to our GitHub Actions confguration files and scripts.
+- `docs`: Documentation only update.
+- `perf`: Performance improvement.
```
ramyamounir, line 37: Some of these types can be co-dependent. For example, a `feat` or a `fix` may lead to `perf` and it may be difficult to separate those into two PRs.  Is there a way to signal that this `fix` also introduces performance improvements (e.g., accuracy increase)?

It's probably a nitpick but I don't think that `perf` will ever get used because it will always be superseded by some other code change type that caused the performance improvement.

ramyamounir (approved):
Looks great, I have been trying to use this for PR titles, but I didn't know we should also use it for the merging commit message. Just left one non-blocking comment on the introduced types.

rfcs/0000_conventional_commits.md
```diff
@@ -0,0 +1,86 @@
...
+- `chore`: The commit is a catch all for work outside of the types identified above. For example, the commit affects infrastructure, tooling, development, or other non-Monty framework code.
+- `rfc`: RFC proposal.
+
+### Breaking Changes
+
+A breaking change is communicated by appending `!` after the type/scope. This correlates with `MAJOR` in [RFC 7 - Monty versioning](0007_monty_versioning.md).
+
+
+> [!NOTE]
+> `fix`, `feat`, and `refactor` types refer only to the **src/tbp/monty** codebase. Adding a new tool is not a feature. Fixing a tool is not a feature. Refactoring a tool is not a refactor of `src/tbp/monty`.
```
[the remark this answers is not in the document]
tristanls: So, to understand my reasoning behind this, we need to take a step back and examine `tbp.monty` from the perspective of an external user. 

External users do not care about our infrastructure for managing the repository, any tests, or the tooling that helps us deliver `tbp.monty` to them. As such, any "features," "fixes," or other things of interest to us internally are irrelevant to our users. Nobody cares whether we have a new feature in our readme sync file tool, whether we have a breaking change in that tool, or whether we introduce a new tool; they don't come to `tbp.monty` for that.

Similarly, we use semver as a mechanism to communicate to `tbp.monty` users when they can use a new version without rewriting anything (`fix/PATCH`), when they can start using a new feature (`feat/MINOR`), and when they will have to rewrite some code they wrote (`!/MAJOR`). We will not roll out a MAJOR change because we will start hosting our documentation somewhere other than readme, requiring the removal of an old tool and the creation of a new one.

So, if we look at things from the perspective of `tbp.monty` users, only `src/tbp/monty` is relevant for versioning decisions and for what is to be considered a feature, a fix, or a breaking change. If we were to use components, it would be within the scope of `src/tbp/monty`... like `(cmp)`, or `(lm)`, or `(motor_sys)`, or something like that. However, we frankly don't need that level of detail right now. I don't think our codebase and contributor community are large enough to start separating things at such a fine level of detail yet.

Lastly, we are introducing conventional commits to inform the PATCH/MINOR/MAJOR version increment for our versioning scheme. We are not introducing them to determine whether a commit introduces a new feature to the tool. If we start conflating what users care about with the infrastructure we maintain in our conventional commits, then we aren't helping the goal of easily deciding what changes our users are exposed to.

rfcs/0000_conventional_commits.md
tristanls: I hear you, but I'm copying and pasting the wording from the standard we use, where they use the word "correlates."  

> fix: a commit of the type fix patches a bug in your codebase (this correlates with [PATCH](http://semver.org/#summary) in Semantic Versioning).
> feat: a commit of the type feat introduces a new feature to the codebase (this correlates with [MINOR](http://semver.org/#summary) in Semantic Versioning).
> -- https://www.conventionalcommits.org/en/v1.0.0/

Also, it is not a `1:1` mapping; e.g., a `fix!:` is a `MAJOR` server change.

rfcs/0000_conventional_commits.md
tristanls: I think additional clarity here helps. I updated it to `tbp.monty` here and elsewhere.

rfcs/0000_conventional_commits.md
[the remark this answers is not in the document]
tristanls: I added to the `# Motivation` section, which now reads:

# Motivation

As of this writing, we largely adopted the habit of labeling our commits as per https://www.conventionalcommits.org/en/v1.0.0/. 22 of the last 23 commits were done formatted in conventional commit style and currently active PRs authored by the TBP team adopt conventional commits as the pull request name.

Adopting conventional commits allows us to rapidly determine what version ([RFC 7 - Monty versioning](0007_monty_versioning.md)) to assign to new changes.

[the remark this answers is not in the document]
vkakerbeck: yeah, I was aware that we can use these for versioning and that they are useful to indicate to external people what changed. That's why I suggested using the scope in addition to the commit type. That way, external people who are just interested in the frameworks code (and us when we think about semver updates) just need to look at `(monty)` commits. But people who are interested in code outside of the frameworks folder (like plotting or benchmarks) still get more detailed info on what kind of change happened there.
But I also understand that enforcing the addition of scope adds more complexity so I am also okay with keeping it this way for now.

rfcs/0000_conventional_commits.md
tristanls: I think it's okay if `perf` isn't used. It won't affect anything. It's on this list because the Angular convention uses it.

I updated the motivation, but let's remember the purpose of these commits: to rapidly determine what type of version update our users will experience: PATCH/MINOR/MAJOR. We only need to communicate enough information for this. If someone wants more detail, they can review the PR and read the code changes.

If it is a backward compatible `fix:` and a performance improvement `perf:`, then I'd go with `fix:` because it is more useful to tell users that something broken is now fixed, and performance isn't the important bit. However, if `perf:` is chosen, it doesn't matter because either choice will only ever update `PATCH`. So, if there were a PR that we're discussing, I'd advocate going with the author's first choice.

rfcs/0000_conventional_commits.md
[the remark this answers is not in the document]
tristanls: If we want to make benchmarks and plotting part of our maintained public "API", then we can extend `feat/fix/refactor` to them. I can see an argument for those being part of our public API.

Right now, plotting is in `src/tbp/monty`.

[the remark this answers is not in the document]
tristanls: I reframed the RFC to be in terms of a **Public API** initially defined as code in **src/tbp/monty**, **benchmarks**, and **tools/plot**. How does it read to you now? It's starting to get a bit complicated. I doubt we'll remember that Public API consists of those three groups. However, this allows us not to add any optional scopes, which will complicate things even more.

I don't think adding an optional scope of `(monty)` and using that for versioning will be sufficient because I have a hard time imagining someone working on `tools/plot` and remembering to label their new plot feature as `feat(monty):`. 

Also, now that **benchmarks** and **tools/plot** are part of the Public API as defined here, I'm questioning those being a part of the Public API. If I imagine myself as a user, I'd question that a new plot in the plot tool would be a "Monty feature." Similarly, I'd question that a new benchmark is a "Monty feature." Furthermore, why would we want to release a new MAJOR version of Monty if we break the plot tool or if we get rid of a benchmark?

codeallthethingz (approved):
LGTM

rfcs/0000_conventional_commits.md
tristanls: I added a note higher up in the RFC:

## Commit Messages

> [!NOTE]
> The following commit message guidance applies only to commits made to the `main` branch when a Pull Request is merged. This guidance does not apply to commit messages elsewhere, for example, on your working branch while working on a Pull Request.

scottcanoe (approved):
LGTM. I was initially unclear about exactly where these commit messages go, but yesterday's post-Niels edit cleared it up for me.

nielsleadholm (approved):
Thanks for those changes, looks great.

tristanls:

I interpret @codeallthethingz, @hlee9212, @vkakerbeck, @nielsleadholm, @ramyamounir, and @scottcanoe current PR acceptance as **approval**.

I will wait a few hours in case anyone has any last comments, but I intend to merge the RFC afterward.

PR #249, opened 2025-04-11
refactor: pose error calculation logic

ramyamounir:
This is a follow-up to [PR#227](https://github.com/thousandbrainsproject/tbp.monty/pull/227). This PR:
* Simplifies the logic of calculating the minimum pose error between predicted poses and a target pose (removes unnecessary for loop)
* fixes all places where this function is called, e.g., `add_evidence_lm_episode_stats` and `get_graph_lm_episode_stats`
* calls this simplified function from `TheoreticalLimitLMLoggingMixin`


src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -434,13 +449,13 @@ def print_overall_stats(stats):
         / len(stats)
         * 100
     )
-    print(f"Detected {np.round(acc,2)}% correctly")
+    print(f"Detected {np.round(acc, 2)}% correctly")
```
tristanls, line 452: **issue (non-blocking)**: Unrelated code change

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -434,13 +449,13 @@ def print_overall_stats(stats):
...
     rt = np.sum(stats["time"])
     rt_per_step = np.mean(stats["time"] / stats["monty_matching_steps"])
     print(
-        f"overall run time: {np.round(rt,2)} seconds ({np.round(rt/60,2)} minutes),"
-        f" {np.round(rt/len(stats),2)} seconds per episode, {np.round(rt_per_step,2)} "
-        "seconds per step."
+        f"overall run time: {np.round(rt, 2)} seconds ({np.round(rt / 60, 2)} minutes),"
+        f" {np.round(rt / len(stats), 2)} seconds per episode,"
+        f" {np.round(rt_per_step, 2)} seconds per step."
```
tristanls, line 458: **issue (non-blocking)**: Unrelated code change

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -468,10 +483,10 @@ def print_unsupervised_stats(stats, epoch_len):
         * 100
     )
     print(
-        f"Detected {np.round(first_epoch_acc,2)}% correctly as new object"
+        f"Detected {np.round(first_epoch_acc, 2)}% correctly as new object"
```
tristanls, line 486: **issue (non-blocking)**: Unrelated code change

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -468,10 +483,10 @@ def print_unsupervised_stats(stats, epoch_len):
...
         "in first epoch"
     )
-    print(f"Detected {np.round(later_acc,2)}% correctly after first epoch")
+    print(f"Detected {np.round(later_acc, 2)}% correctly after first epoch")
```
tristanls, line 489: **issue (non-blocking)**: Unrelated code change

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -480,8 +495,8 @@ def print_unsupervised_stats(stats, epoch_len):
             print("     " + string)
     rt = np.sum(stats["time"])
     print(
-        f"overall run time: {np.round(rt,2)} seconds ({np.round(rt/60,2)} minutes),"
-        f" {np.round(rt/len(stats),2)} seconds per episode."
+        f"overall run time: {np.round(rt, 2)} seconds ({np.round(rt / 60, 2)} minutes),"
+        f" {np.round(rt / len(stats), 2)} seconds per episode."
```
tristanls, line 499: **issue (non-blocking)**: Unrelated code change

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -532,7 +547,7 @@ def get_graph_lm_episode_stats(lm):
         dict with stats of one episode.
     """
     primary_performance = "patch_off_object"  # Performance on the primary target in
-    # the environmnet, typically the target object we begin the episode on
+    # the environment, typically the target object we begin the episode on
```
tristanls, line 550: **issue (non-blocking)**: Unrelated code change

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -663,7 +678,7 @@ def add_pose_lm_episode_stats(lm, stats):
     """Add possible poses of lm to episode stats.
 
     Args:
-        lm: LM istance from which to add the statistics.
+        lm: LM instance from which to add the statistics.
```
tristanls, line 681: **issue (non-blocking)**: Unrelated code change

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -841,7 +856,7 @@ def lm_stats_to_dataframe(stats, format_for_wandb=False):
         {0: {LM_0: stats, LM_1: stats...}, 1:...} --> dataframe
 
     Currently we are reporting once per episode, so the loop over episodes is only over
-    a singel key, value pair, but leaving it here because it is backward compatible.
+    a single key, value pair, but leaving it here because it is backward compatible.
```
tristanls, line 859: **issue (non-blocking)**: Unrelated code change

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -901,8 +916,7 @@ def maybe_rename_existing_directory(path, report_count):
     if (report_count == 0) and os.path.exists(path):
         new_path = path + "_old"
         logging.warning(
-            f"Output path {path} already exists. This path will be moved"
-            f"to {new_path}"
+            f"Output path {path} already exists. This path will be moved to {new_path}"
```
tristanls, line 919: **issue (non-blocking)**: Unrelated code change

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -952,7 +966,7 @@ def total_size(o):
 
 
     The recursive recipe universally cited on stack exchange and blogs for gauging the
-    size of python objets in memory.
+    size of python objects in memory.
```
tristanls, line 969: **issue (non-blocking)**: Unrelated code change

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -388,18 +388,31 @@ def get_time_stats(all_ds, all_conditions):
...
+        predicted_rotation (Rotation): Predicted rotation(s). Can be a single or batched
+            rotation.
+        target_rotation (Rotation): Target rotation. Must represent a single rotation.
+
+    Returns:
+        float: The minimum angular error in radians.
+    """
+    error = (predicted_rotation * target_rotation.inv()).magnitude()
+    if isinstance(error, np.ndarray):
+        error = error.min()
```
tristanls, line 414: **issue**: `isinstance` should be avoided. Its use hints at problems with design.

**suggestion**:
```python
error = np.min((predicted_rotation * target_rotation.inv()).magnitude())
```

ramyamounir:
@tristanls, yes, Viviane commented on this before. I don't use VScode (or similar), I am going to replicate this behavior with my editor. Planning to look into that over the weekend :)

Some of these edits (e.g., typos) are useful, I manually fix them as I spot a typo somewhere. Would you prefer I create a separate PR just for typos and ruff edits?

tristanls:
@ramyamounir, I think all of those extra changes are useful. However, they are not part of the refactoring of pose error calculation logic. For a code review, it is an additional cognitive load that slows down the integration of your changes because of some spacing or alignment updates. It is very confusing to review, especially as the count of unrelated changes goes up. A one-off style change is probably not a problem, but as we need to pick some criteria, we generally pick _no_ extraneous changes as the least contentious.

Typos and unrelated formatting should be compartmentalized to a separate `style:` PR.

ramyamounir:
Makes sense, I've drafted this PR until I revert the unrelated changes.

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -388,18 +388,29 @@ def get_time_stats(all_ds, all_conditions):
...
-        if error < min_error:
-            min_error = error
-    return min_error
+def compute_pose_error(
+    predicted_rotation: Rotation, target_rotation: Rotation
+) -> float:
+    """Computes the minimum angular pose error between predicted and target rotations.
+
+    Both inputs must be instances of `scipy.spatial.transform.Rotation`. The
+    `predicted_rotation` may contain a single rotation or a batch of rotations,
```
vkakerbeck, line 397:
```suggestion
    `predicted_rotation` may contain a single rotation or a list of rotations,
```

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -388,18 +388,29 @@ def get_time_stats(all_ds, all_conditions):
...
+    while `target_rotation` must contain exactly one rotation.
```
vkakerbeck, line 398:
```suggestion
    while `target_rotation` must be exactly one rotation.
```

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -388,18 +388,29 @@ def get_time_stats(all_ds, all_conditions):
...
+
+    The pose error is defined as the geodesic distance on SO(3) — the angle of the
+    relative rotation between predicted and target. If `predicted_rotation` contains
+    multiple rotations, this function returns the minimum error among them.
+
+    Args:
+        predicted_rotation (Rotation): Predicted rotation(s). Can be a single or batched
```
vkakerbeck, line 405:
```suggestion
        predicted_rotation (Rotation): Predicted rotation(s). Can be a single or list of
```

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -420,7 +431,9 @@ def get_overall_pose_error(stats, lm_id="LM_0"):
         detected = stats[episode][lm_id]["detected_rotation_quat"]
         if detected is not None:  # only checking accuracy on detected objects
             target = stats[episode][lm_id]["target"]["quat_rotation"]
-            err = get_pose_error(detected, target)
+            err = compute_pose_error(
+                Rotation.from_quat(detected), Rotation.from_quat(target)
```
vkakerbeck, line 435: previously the conversion of detected pose quaternions into Rotations happened inside a for loop over the list. Did you make sure the `from_quat` functions works as expected when you give it a list of quaternions instead of just one?

vkakerbeck: (I know it looks like it is just one rotation here but it can be a list if symmetry was detected https://github.com/thousandbrainsproject/tbp.monty/blob/380d689a2d23d18b29e01442373213d5122beb23/src/tbp/monty/frameworks/utils/logging_utils.py#L571 )

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -388,18 +388,29 @@ def get_time_stats(all_ds, all_conditions):
     return time_stats
 
 
-def get_pose_error(detected_pose, target_pose):
-    if type(detected_pose) != list:
-        detected_pose = [detected_pose]
-    target_r = Rotation.from_quat(target_pose)
-    min_error = np.pi
```
vkakerbeck, line 395: It's been a really long time since I wrote this, so I just want to double-check: Is it really sufficient to just use np.min? Does the magnitude function always return values < pi?

tests/unit/frameworks/utils/compute_pose_error_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def test_known_small_angle_difference(self):
+        """Test pose error matches the known angle between simple rotations."""
+        rot1 = Rotation.from_euler("z", 0, degrees=True)
+        rot2 = Rotation.from_euler("z", 10, degrees=True)
+        error = compute_pose_error(rot1, rot2)
+        self.assertAlmostEqual(error, np.deg2rad(10), places=6)
+
+    def test_min_error_from_batch(self):
+        """Test function returns the minimum error from a batch of predicted rotations.
```
vkakerbeck, line 35:
```suggestion
        """Test function returns the minimum error from a list of predicted rotations.
```

vkakerbeck (approved):
Nice, thanks for restructuring this! My comments are mostly suggested edits to reword batch to list to be more clear about the actual type. Otherwise just a few small questions to double check it is all working as intended.

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -388,18 +388,29 @@ def get_time_stats(all_ds, all_conditions):
...
+) -> float:
+    """Computes the minimum angular pose error between predicted and target rotations.
+
+    Both inputs must be instances of `scipy.spatial.transform.Rotation`. The
+    `predicted_rotation` may contain a single rotation or a batch of rotations,
+    while `target_rotation` must contain exactly one rotation.
+
+    The pose error is defined as the geodesic distance on SO(3) — the angle of the
+    relative rotation between predicted and target. If `predicted_rotation` contains
+    multiple rotations, this function returns the minimum error among them.
```
nielsleadholm, line 402: **suggestion:** Maybe just add a short comment that the `.inv()` taken in this method is how geodesic distance between two rotations is calculated, not a side-effect of whether the target rotation is stored in its normal form, or as its inverse. The function therefore assumes that the orientations are already in the same coordinate system before comparing.

tests/unit/frameworks/utils/compute_pose_error_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+        The target rotation is 12°, so the closest in the batch [0°, 10°, 20°] is 10°,
```
nielsleadholm, line 37:
```suggestion
        The target rotation is 12°, so the closest in the list [0°, 10°, 20°] is 10°,
```

tests/unit/frameworks/utils/compute_pose_error_test.py
```diff
@@ -0,0 +1,62 @@
...
+        with a 2° error.
+        """
+        rot_batch = Rotation.from_euler("z", [0, 10, 20], degrees=True)
+        target = Rotation.from_euler("z", 12, degrees=True)
+        error = compute_pose_error(rot_batch, target)
+        self.assertAlmostEqual(error, np.deg2rad(2), places=6)
+
+    def test_180_degree_rotation(self):
+        """Test that the pose error for a 180-degree rotation is pi radians."""
+        rot1 = Rotation.from_quat([0, 0, 1, 0])  # 180° rotation around z-axis
```
nielsleadholm, line 47: **suggestion:** Maybe to @vkakerbeck separate comment, it could be worth adding a test with a rotation >180, and show that the error is <180?

nielsleadholm (commented):
Really nice and clean, thanks for putting this together! Other than a couple small comments, did you have a chance to run some of the 77-object benchmark experiments to make sure results are the same?

src/tbp/monty/frameworks/utils/logging_utils.py
jeremyshoemaker: In my experience, spelling corrections in comments generally won't happen unless they happen in passing like this, so I'm generally not opposed to them, even if they seem unrelated to the PR they happen in.

ramyamounir: Yeah, good point, what do you think @tristanls, should we limit unrelated changes to typos in docstrings and comments? I don't think I will ever open a PR to fix some typos.

nielsleadholm: I feel like at least for this PR it might be worth an exception? Personally it doesn't add much cognitive overhead when reviewing and a few of these are thrown in, I think the main detractor I can think of is it might make the git-blame history a bit confusing for a given line, but then again it's rare that we go down to that granularity.

src/tbp/monty/frameworks/utils/logging_utils.py
ramyamounir: Yes, the `Rotation` object can accept a list of quaternions, euler rotations, or rotation matrices. Then you can perform vectorized operations on these rotations. This is what allowed me to get rid of the for loop in `get_pose_error`. I have added a unit test (`test_rotation_object_from_quaternion_list`) to show this.

src/tbp/monty/frameworks/utils/logging_utils.py
ramyamounir: Yes, I added a unit test for this: `test_pose_error_never_exceeds_pi`.

tests/unit/frameworks/utils/compute_pose_error_test.py
ramyamounir: Yes, I added a unit test for this: `test_pose_error_never_exceeds_pi`

ramyamounir:
@nielsleadholm and @vkakerbeck, thanks for your reviews! I have added unit tests to address your concerns and have fixed the other issues you pointed out. I also checked the performance of the benchmark `randrot_noise_77obj_surf_agent` and found it to be the same.

src/tbp/monty/frameworks/utils/logging_utils.py
tristanls: > I don't think I will ever open a PR to fix some typos.

@ramyamounir, what's the argument being presented? Typos are not important to warrant a PR. But typos are important enough to detract from a PR containing needed code changes? I think a [priority inversion](https://en.wikipedia.org/wiki/Priority_inversion) is happening here.

**note**: Leaving typos around allows for typo-only PRs as first-commits for new contributors. These PRs can be very straightforward and serve to familiarize new contributors with the PR process without the overhead of changing code, etc.

src/tbp/monty/frameworks/utils/logging_utils.py
vkakerbeck: great, thank you!

src/tbp/monty/frameworks/utils/logging_utils.py
vkakerbeck: Thanks!

vkakerbeck (approved):
Great, thanks for double checking and the additional tests!

src/tbp/monty/frameworks/utils/logging_utils.py
ramyamounir: @nielsleadholm Sure, sounds reasonable. I'm fine with either approach.

> what's the argument being presented? Typos are not important to warrant a PR. But typos are important enough to detract from a PR containing needed code changes?

@tristanls The argument is I think the importance of fixing typos is understated while the cognitive load of reviewing a typo is overstated. FWIW, I would not open a PR for a typo not because it's not important enough to fix them but because I don't think typo fixes should undergo the same review process. If I can assign style type PRs only to myself and merge them without asking for reviews from other maintainers, I will do that right away for every file I edit. In most cases, typo fixes do not require review of others, but they are important.

jeremyshoemaker: > That way functions and scripts that aren't being modified by the PR won't suddenly get pulled in

That's the rubric I would use to draw the boundary, but there have been several times when I'm working on a file and I notice typos in unrelated parts of the file that I'd prefer to just correct, "while I've got the hood open".

Are we saying that we shouldn't do that and keep typo changes focused on _just_ the functions and methods being changed in the PR?

tristanls: @ramyamounir, I am not overstating _my experience of reviewing code_. It is what it is.

I don't understand the hesitation about opening PRs with some typos. We have plenty of trivial PRs that get rapidly approved and integrated.

https://github.com/thousandbrainsproject/tbp.monty/commit/4e2fd88e62260a8d859f0f68a2e76c6529d1cfcc
https://github.com/thousandbrainsproject/tbp.monty/commit/c00d89c8b87fd580e7f8504a27793e6b6f058ce4
https://github.com/thousandbrainsproject/tbp.monty/commit/a0bf79160e56385339d50a416bbdc567eff6580c
https://github.com/thousandbrainsproject/tbp.monty/commit/8312bd67fc82acaefae1d55894dd4da1c1d0ebd6
https://github.com/thousandbrainsproject/tbp.monty/commit/05dd6e61a82bfbea25f5b0a8a60b7786bac116f7
https://github.com/thousandbrainsproject/tbp.monty/commit/104277d44a02a3507df9e04921a15aef7f86ed1b
https://github.com/thousandbrainsproject/tbp.monty/commit/1cf8a07ca5cca165b052bbcbfda62e7daa009b3c
etc..

ramyamounir: I'm only trying to explain why, in my mind, it is not a case of priority inversion. I could be wrong, or my experience could be different.

Fair point, maybe I have a higher than usual boundary for opening a PR because I'm new to this process.

tristanls: What I mean by my personal experience is me, after a couple hours of code reviewing prior PRs line by line, trying to keep a new PR code change in my head, experiencing running out of working memory because of a typo or formatting change throwing a "is this spelled right" task into the mix. I'm not talking about importance. I already mentioned that these are worthwhile changes. The specific problem is changing unrelated things together as they enduce context switches. It's not as if I'm incapable of eventually getting through it. But if it's easier for reviewers then I think that speeds review and integrating the code.

ramyamounir: Fair enough, I see your point. Btw, I've set my editor to format only the diff hunks so we shouldn't see any more of these unrelated changes in future PRs.

PR #250, opened 2025-04-11
refactor: enforce TheoreticalLimitLMLoggingMixin compatibility requirement

tristanls:
This pull request enforces `TheoreticalLimitLMLoggingMixin` compatibility via code instead of convention.

### DIFF
--- a/src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
+++ b/src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
@@ -11,6 +11,11 @@ from typing import Any, Dict
 
 from scipy.spatial.transform import Rotation
 
+from tbp.monty.frameworks.models.evidence_matching import EvidenceGraphLM
+from tbp.monty.frameworks.models.no_reset_evidence_matching import (
+    NoResetEvidenceGraphLM,
+)
+
 
 class TheoreticalLimitLMLoggingMixin:
     """Mixin that adds theoretical limit and pose error logging for learning modules.
@@ -28,6 +33,24 @@ class TheoreticalLimitLMLoggingMixin:
         - EvidenceGraphLM
         - NoResetEvidenceGraphLM
     """
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        """Ensure the mixin is used only with compatible learning modules.
+
+        Raises:
+            ValueError: If the mixin is used with a non-compatible learning module.
+        """
+        super().__init_subclass__(**kwargs)
+        if not any(
+            [
+                issubclass(b, (EvidenceGraphLM, NoResetEvidenceGraphLM))
+                for b in cls.__bases__
+            ]
+        ):
+            raise ValueError(
+                "TheoreticalLimitLMLoggingMixin is compatible with EvidenceGraphLM or "
+                f"NoResetEvidenceGraphLM, got {cls}"
+            )
 
     def _add_detailed_stats(self, stats: Dict[str, Any]) -> Dict[str, Any]:
         """Add detailed statistics to the logging dictionary.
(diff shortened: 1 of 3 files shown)


src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -26,9 +28,21 @@ class TheoreticalLimitLMLoggingMixin:
...
-        - NoResetEvidenceGraphLM
     """
 
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        """Ensure the mixin is used only with compatible learning modules.
+
+        Raises:
+            TypeError: If the mixin is used with a non-compatible learning module.
+        """
+        super().__init_subclass__(**kwargs)
```
ramyamounir, line 39: Can you elaborate on the function of this line?

tests/unit/frameworks/models/mixins/no_reset_evidence_test.py
```diff
@@ -0,0 +1,32 @@
...
+from tbp.monty.frameworks.models.abstract_monty_classes import LearningModule
+from tbp.monty.frameworks.models.evidence_matching import EvidenceGraphLM
+from tbp.monty.frameworks.models.mixins.no_reset_evidence import (
+    TheoreticalLimitLMLoggingMixin,
+)
+
+
+class InheritanceTheoreticalLMLoggingMixinTest(TestCase):
+    @staticmethod
+    def test_mixin_used_with_compatible_learning_module_does_not_erorr() -> None:
```
ramyamounir, line 23:
```suggestion
    def test_mixin_used_with_compatible_learning_module_does_not_error() -> None:
```

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -26,9 +28,21 @@ class TheoreticalLimitLMLoggingMixin:
 
     Compatible with:
         - EvidenceGraphLM
-        - NoResetEvidenceGraphLM
```
ramyamounir, line 29: Is it enough to just say compatible with `EvidenceGraphLM`? Should we maybe add "and its subclasses"? I assume you removed `NoResetEvidenceGraphLM` because it would become a chore to keep it updated with any new class we come up with, correct?

ramyamounir (commented):
Thanks! This looks really nice.

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
ramyamounir: Yeah, sorry.. My understanding of `__init_subclass__` is that it gets called **only** if its class gets subclassed (e.g., `Model(mixin, OtherModel)`), and `cls` here would be the class that used the mixin (e.g., Model).

What role does calling the `__init_subclass__` of the superclass play here? I understand that this is the typical inheritance logic as you mentioned, but I'm trying to figure out if it has an actual role in our Mixin use case. Would this be useful e.g., in cases where we have a nested Mixin structure? This completely for my understanding, and definitely non-blocking.

**Note**: I checked out your branch and commented this line, unit tests still pass and experiments run fine.

tristanls: I see. The answer is still the same: proper use of inheritance. Just because today, the superclass is not doing anything (hence, commenting out this line does nothing) does not mean that the superclass code won't change and start doing something essential in the future.

Today, `object.__init_subclass__` "[does nothing, but raises an error if it is called with any arguments.](https://docs.python.org/3.8/reference/datamodel.html#object.__init_subclass__)"

So, it does function as a check that we consume all `kwargs`.

tests/unit/frameworks/models/mixins/no_reset_evidence_test.py
tristanls: Thanks! Updated.

ramyamounir (approved):

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -26,9 +28,21 @@ class TheoreticalLimitLMLoggingMixin:
...
     """
 
+    def __init_subclass__(cls, **kwargs: Any) -> None:
```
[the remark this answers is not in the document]
tristanls: I think ensuring it is a leftmost mixin is too restrictive.

To me, this is the beginning of eliminating the deeply nested inheritance hierarchy and a shift toward composing behavior via mixin/behaves-like-a relationships. With that goal, we will likely compose multiple orthogonal mixins, and it doesn't seem meaningful to restrict the ordering of those.

PR #251, opened 2025-04-13
Refactor | Typing for Models

artyom-dehtiar:
Disclaimer:
Proposed data types was extracted in a **runtime** from actual data which was passed into functions and returned from them. Obvious downside of this approach is that the data types are not exhaustive, it's just what I've seen in a **debugger** while running experiments (I used few), but it's better than nothing
_They are meant to be extended and specified further_

Anticipating some questions:

Q: Why add type when there's default value from which type could be inferred?
A: Because now you could safely remove default value without losing type info

Q: Why use NDArray instead of np.ndarray?
A: With small modifications NDArray could be provided with Shape info, and that could be very useful

Q: Why use Dict[str, Any] instead of simple dict
A: In most cases, structure of content in those dicts are the same (per function) and later this structure could be defined in dataclasses or in pydantic models (they're richer than defaultdict) , and then it will look like Dict[str, SpecificStructure] which is useful

Q: Why TYPE_CHECKING?
A: This prevents issues with circular imports

Q: Why remove '-> None' typing
A: None is default return for any function in python, types are for everything but explicit None, eg:
```
def foo():
    return None

def foo():
    return 

def foo(): ...
```
each function will equally return None when called, other cases with None is covered with Optional (when we expect value or None)

Some functions are not covered, because they weren't triggered in used experiments

### DIFF
--- a/src/tbp/monty/frameworks/models/buffer.py
+++ b/src/tbp/monty/frameworks/models/buffer.py
@@ -592,7 +592,7 @@ class BufferEncoder(json.JSONEncoder):
         cls,
         obj_type: type,
         encoder: Union[Callable, Type[json.JSONEncoder]],
-    ) -> None:
+    ):
         """Register an encoder.
 
         Args:
@@ -611,7 +611,7 @@ class BufferEncoder(json.JSONEncoder):
             raise ValueError(f"Invalid encoder: {encoder}")
 
     @classmethod
-    def unregister(cls, obj_type: type) -> None:
+    def unregister(cls, obj_type: type):
         """Unregister an encoder.
 
         Args:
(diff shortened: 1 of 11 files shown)


artyom-dehtiar:
https://github.com/thousandbrainsproject/cla/issues/27#event-17239170271

tristanls (commented):
**praise**: Hi @artyom-dehtiar, thank you for getting started on these. We appreciate it.

**todo**: Please review the style checks. Running `ruff check src tests benchmarks` locally should help reproduce and anticipate any other styling issues.

I skimmed through the PR (I will look more closely once all the checks pass), but I noticed one thing I will reject.

**issue**: As we treat everything in `src/tbp/monty` and `benchmarks` as part of our public API, we explicitly mark `-> None` for return types. Please don't remove these.

PR #252, opened 2025-04-14
fix: monty meets world streaming code

vkakerbeck:
As I was testing the monty meets world streaming experiment I noticed a few minor issues. Those are fixed here.

### DIFF
--- a/src/tbp/monty/frameworks/environment_utils/server.py
+++ b/src/tbp/monty/frameworks/environment_utils/server.py
@@ -60,6 +60,9 @@ class MontyRequestHandler(http.server.SimpleHTTPRequestHandler):
 
 if __name__ == "__main__":
     ip_address = "0.0.0.0"  # Set this to your WiFi's IP address
+    assert ip_address != "0.0.0.0", (
+        "Please set the IP address to your WiFi's IP address."
+    )
     port = 8080
     server = http.server.HTTPServer((ip_address, port), MontyRequestHandler)
     print(f"Waiting for data at {ip_address}:{port}...")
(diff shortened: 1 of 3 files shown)


src/tbp/monty/frameworks/environment_utils/server.py
```diff
@@ -55,5 +59,8 @@ def do_PUT(self):  # noqa: N802
 
 
 if __name__ == "__main__":
-    server = http.server.HTTPServer(("0.0.0.0", 8080), MontyRequestHandler)
+    ip_address = "0.0.0.0"  # Set this to your WiFi's IP address
```
nielsleadholm, line 62: **suggestion:** Would it maybe be better to get this from an environment variable? Then if it's not set, you can raise an informative error advising this?

nielsleadholm (approved):
Looks good! Just a small suggestion.

src/tbp/monty/frameworks/environment_utils/server.py
vkakerbeck: I would not extract it automatically because people may have different wifis and I think it would be less error prone if the user sets this himself and therefor also check if both laptop and iPad are connected to the right WiFi network. However, I added an assert reminding the user to set it. Does that sound good?

nielsleadholm: What I mean by automatic - this would be a made up environment variable we define, e.g. `WIFI_IP_FOR_MONTY`

By default, this would not be defined. The user would have to manually run something like 
```bash
export WIFI_IP_FOR_MONTY=your_wifi_IP_address_here

# Confirm set correctly
env | grep WIFI_IP_FOR_MONTY
```

Until they have set it, it would give an error message advising them they need to do the above. But that way it's not a hard-coded line in a script. Does that make sense?

vkakerbeck: ah yes, makes sense!

nielsleadholm (approved):

PR #253, opened 2025-04-15
refactor: replace type checks with isinstance (E721)

hlee9212:
Note on fixing [Ruff rule E721](https://docs.astral.sh/ruff/rules/type-comparison/):

Replaced direct type comparisons like `type(var) == SomeClass` with `isinstance(var, SomeClass)`, as recommended.

Just for posterity:
`isinstance(var, SomeClass)` returns `True` even if `var` is an instance of a subclass of `SomeClass`. This is usually the desired behavior when checking types.

However, if strict type equality is required—i.e., to ensure var is exactly an instance of a given class and not a subclass—use `type(var) is SomeClass`. For example:

```python
type(var) is SubClass and type(var) is not SuperClass)
```
That kind of check should be rare and very intentional, since it breaks the typical polymorphic behavior expected in Python.

### DIFF
--- a/src/tbp/monty/frameworks/models/states.py
+++ b/src/tbp/monty/frameworks/models/states.py
@@ -200,10 +200,10 @@ class State:
             self.confidence >= 0 and self.confidence <= 1
         ), f"Confidence must be in [0,1] but is {self.confidence}"
         assert isinstance(self.use_state, bool), (
-            "use_state must be a boolean but is " f"{type(self.use_state)}"
+            f"use_state must be a boolean but is {type(self.use_state)}"
         )
         assert isinstance(self.sender_id, str), (
-            "sender_id must be string but is " f"{type(self.sender_id)}"
+            f"sender_id must be string but is {type(self.sender_id)}"
         )
         assert (
             self.sender_type in self.allowable_sender_types
@@ -316,10 +316,10 @@ class GoalState(State):
             self.confidence >= 0 and self.confidence <= 1
         ), f"Confidence must be in [0,1] but is {self.confidence}"
         assert isinstance(self.use_state, bool), (
-            "use_state must be a boolean but is " f"{type(self.use_state)}"
+            f"use_state must be a boolean but is {type(self.use_state)}"
         )
         assert isinstance(self.sender_id, str), (
-            "sender_id must be string but is " f"{type(self.sender_id)}"
+            f"sender_id must be string but is {type(self.sender_id)}"
         )
         # Note *only* GSGs should create GoalState objects
         assert (


jeremyshoemaker (approved):
This looks like a good improvement to me. I think the first time I ran into one of these `type` checks I was a little confused since it's usually done with `isinstance`.

tristanls (approved):
Looks good, thank you.

tristanls (changes requested):
**issue**: I'm withdrawing approval. A function was added with latest push.

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -388,6 +388,18 @@ def get_time_stats(all_ds, all_conditions):
...
+        detected_pose = [detected_pose]
+    target_r = Rotation.from_quat(target_pose)
+    min_error = np.pi
+    for det_r in detected_pose:
+        detected_r = Rotation.from_quat(det_r)
+        difference = detected_r * target_r.inv()
+        error = difference.magnitude()
+        if error < min_error:
+            min_error = error
+    return min_error
```
tristanls, line 402: **issue**: This doesn't look like it belongs in this PR.

src/tbp/monty/frameworks/models/states.py
```diff
@@ -189,22 +189,22 @@ def _check_all_attributes(self):
...
-        ), f"use_state must be a boolean but is {type(self.use_state)}"
-        assert (
-            type(self.sender_id) == str
-        ), f"sender_id must be string but is {type(self.sender_id)}"
+        assert isinstance(self.use_state, bool), (
+            "use_state must be a boolean but is " f"{type(self.use_state)}"
+        )
+        assert isinstance(self.sender_id, str), (
+            "sender_id must be string but is " f"{type(self.sender_id)}"
+        )
```
tristanls, line 207: **nitpick**: Is there any reason this can't be a single `f` string? This formatting looks unusual.

src/tbp/monty/frameworks/models/states.py
```diff
@@ -315,12 +315,12 @@ def _check_all_attributes(self):
...
```
tristanls, line 323: **nitpick**: Is there any reason this can't be a single `f` string? This formatting looks unusual.

src/tbp/monty/frameworks/models/states.py
tristanls: Correct. I only commented on the lines that appeared to fit on a single line but were broken into two side-by-side strings.

tristanls (approved):

PR #255, opened 2025-04-15
refactor: reduce dataloader and motor system policy state interactions

tristanls:
This pull request reduces the number of interactions between DataLoader and MotorSystem policies regarding policy `state`. This slightly reduces the complexity in a few places for follow-on refactoring.

There are two types of changes:

First, capture the state from the dataset in a local `state` variable, interact with it, and then set `motor_system._policy.state` once. This ends up looking as follows:

```python
self._observation, state = self.dataset.reset()
# do stuff with state
self.motor_system._policy.state = state
```

Second, don't set observations or state when they will be immediately overridden. This ends up looking as follows:

```python
_, _ = self.dataset[set_agent_pose]
self._observation, self.motor_system._policy.state = self.dataset[
    set_sensor_rotation
]
```

### DIFF
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -379,16 +379,16 @@ class EnvironmentDataLoaderPerObject(EnvironmentDataLoader):
 
     def reset_agent(self):
         logging.debug("resetting agent------")
-        self._observation, self.motor_system._policy.state = self.dataset.reset()
+        self._observation, state = self.dataset.reset()
         self._counter = 0
 
         # Make sure to also reset action variables when resetting agent during
         # pre-episode
         self._action = None
         self._amount = None
-        self.motor_system._policy.state[self.motor_system._policy.agent_id][
-            "motor_only_step"
-        ] = False
+        state[self.motor_system._policy.agent_id]["motor_only_step"] = False
+
+        self.motor_system._policy.state = state
 
         return self._observation
 
@@ -457,13 +457,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                     self._observation, view_sensor_id="view_finder"
                 )
 
-                self.motor_system._policy.state[self.motor_system._policy.agent_id][
-                    "motor_only_step"
-                ] = True
-
-            self._observation, self.motor_system._policy.state = self.dataset[
-                self._action
-            ]
+            self._observation, state = self.dataset[self._action]
 
             # Check whether sensory information is just for feeding back to motor policy
             # TODO refactor so that the motor policy itself is making this update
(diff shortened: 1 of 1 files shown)


src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -457,13 +457,7 @@ def __next__(self):
                     self._observation, view_sensor_id="view_finder"
                 )
 
-                self.motor_system._policy.state[self.motor_system._policy.agent_id][
-                    "motor_only_step"
-                ] = True
```
tristanls, line 462: **note**: Setting the state here is overridden by the time we get to the old line 464 (or new line 473).

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -701,16 +693,11 @@ def execute_jump_attempt(self):
             self.motor_system._policy.derive_habitat_goal_state()
         )
 
-        # Update observations and motor system-state based on new pose
```
tristanls, line 704: **note**: No update happens since everything is overridden on old line 723 (new line 711) below. Comment removed.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -701,16 +693,11 @@ def execute_jump_attempt(self):
...
         set_agent_pose = SetAgentPose(
             agent_id=self.motor_system._policy.agent_id,
             location=target_loc,
             rotation_quat=target_np_quat,
         )
-        self._observation, self.motor_system._policy.state = self.dataset[
-            set_agent_pose
-        ]
-
```
tristanls, line 713: **note**: Moved further down without capturing output to demonstrate that we are taking two actions in the environment and only care about the observation and state from the last one.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -803,43 +791,36 @@ def handle_failed_jump(self, pre_jump_state, first_sensor):
             location=pre_jump_state["position"],
             rotation_quat=pre_jump_state["rotation"],
         )
-        self._observation, self.motor_system._policy.state = self.dataset[
-            set_agent_pose
-        ]
-
```
tristanls, line 809: **note**: Moved further down (new line 800) without capturing output to demonstrate that we are taking two actions in the environment and only care about the observation and state from the last one.

src/tbp/monty/frameworks/environments/embodied_data.py
nielsleadholm: Good catch, might be worth updating the comment on line 701 to something like:

```python
        # Update observations and motor system-state based on new pose, accounting
        # for resetting both the agent, as well as the poses of its coupled sensors; this
        # is necessary for the distant agent, which pivots the camera around

```

nielsleadholm (approved):
Nice change! Definitely helps reduce the complexity.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -379,16 +379,16 @@ def add_distractor_objects(
...
         self._counter = 0
 
         # Make sure to also reset action variables when resetting agent during
         # pre-episode
         self._action = None
         self._amount = None
-        self.motor_system._policy.state[self.motor_system._policy.agent_id][
-            "motor_only_step"
-        ] = False
+        state[self.motor_system._policy.agent_id]["motor_only_step"] = False
```
jeremyshoemaker, line 389: Since we access it quite a few times here, it might make sense to assign `self.motor_system._policy.agent_id` to a local variable, e.g. `agent_id` to make all these indexes into `state` shorter and easier to read. The lines are pretty noisy otherwise.

I guess it's actually spread across several methods here, so it might be a narrower change than I'm thinking. 🤷

tristanls: Yes, to this and many other things I didn't address here 🙂. I'm deferring holding on to `agent_id` and reusing it as future work. This PR aims to reduce the number of `_policy.state` interactions.

PR #256, opened 2025-04-16
refactor: update np.infty to np.inf (NPY201)

hlee9212:
Updating to use `np.inf`. Not critical as our current numpy version is less than 2.0, but will be removed once we upgrade. Here is the [NumPy 2.0 migration guide](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html).


nielsleadholm (approved):
Thanks for catching this Hojae!

tristanls (approved):

PR #257, opened 2025-04-16
refactor: address PLW2901

hlee9212:
Documentation for [PLW2901](https://docs.astral.sh/ruff/rules/redefined-loop-name/). Prioritized because this can cause subtle bugs.

### DIFF
--- a/src/tbp/monty/frameworks/utils/dataclass_utils.py
+++ b/src/tbp/monty/frameworks/utils/dataclass_utils.py
@@ -68,11 +68,7 @@ def from_dataclass_dict(datadict):
     # Check for nested dataclass
     kwargs = {}
     for k, v in datadict.items():
-        if isinstance(v, dict):
-            nested_value = from_dataclass_dict(v)
-        else:
-            nested_value = v
-        kwargs[k] = nested_value
+        kwargs[k] = from_dataclass_dict(v) if isinstance(v, dict) else v
 
     if _DATACLASS_TYPE not in kwargs:
         # Not a dataclass dict
--- a/tools/github_readme_sync/readme.py
+++ b/tools/github_readme_sync/readme.py
@@ -219,10 +219,10 @@ class ReadMe:
 
                     # Process headers and build alignment lookup
                     alignments = {}
-                    for i, raw_header in enumerate(headers):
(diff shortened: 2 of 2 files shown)


src/tbp/monty/frameworks/utils/spatial_arithmetics.py
```diff
@@ -40,11 +40,11 @@ def rot_mats_to_quats(rot_mats, invert=False):
         Quaternions
     """
     quats = []
-    for rot_mat in rot_mats:
-        rot_mat = Rotation.from_matrix(rot_mat)
+    for mat in rot_mats:
```
vkakerbeck, line 43: Can we give this a more expressive name instead of a less expressive name? Something like `rotation_matrix`?

src/tbp/monty/frameworks/utils/dataclass_utils.py
```diff
@@ -69,8 +69,10 @@ def from_dataclass_dict(datadict):
     kwargs = {}
     for k, v in datadict.items():
         if isinstance(v, dict):
-            v = from_dataclass_dict(v)
-        kwargs[k] = v
+            nested_value = from_dataclass_dict(v)
+        else:
+            nested_value = v
+        kwargs[k] = nested_value
```
tristanls, line 75: **suggestion (non-blocking)**: How about...
```python
kwargs[k] = from_dataclass_dict(v) if isinstance(v, dict) else v
```
...which you effectively used elsewhere for this.

tools/github_readme_sync/readme.py
```diff
@@ -219,10 +219,10 @@ def replace_match(match):
 
                     # Process headers and build alignment lookup
                     alignments = {}
-                    for i, header in enumerate(headers):
+                    for i, raw_header in enumerate(headers):
                         title_attr = ""
                         align_style = ""
-                        parts = [p.strip() for p in header.split("|")]
+                        parts = [p.strip() for p in raw_header.split("|")]
```
tristanls, line 225: **nitpick**: Maybe `unparsed_header`? Not sure what "raw" would be in this context.

tristanls (approved):
Looks good. Thanks for reaching into `tools` as well. I left some non-blocking comments.

PR #261, opened 2025-04-17
docs: additional formatting, structuring and notes

vkakerbeck:
Adding a couple of minor notes to the documentation. Also added a bit of formatting (bolding some text) and restructured the page on sensor modules, transforms and the CMP.


codeallthethingz (changes requested):
Few typos, one suggested rewording.

PR #262, opened 2025-04-17
feat: MotorSystemState type definition

tristanls:
This pull request defines the type of the (currently) convention-driven motor system state definition.


tristanls:
It feels a bit like I am introducing dead code, though. Perhaps this should wait for actual use 🤔.

PR #234, opened 2025-04-08
refactor!: motor policies are an attribute of MotorSystem

tristanls:
This pull request removes motor policies from the `MotorSystem` inheritance hierarchy. 

This is a breaking change (`refactor!:`) due to configuration changes. The functionality, once configuration is updated, works the same.

A new inheritance hierarchy is introduced with `MotorPolicy` replacing `MotorSystem` at the top. The rest of the policy hierarchy remains intact.

Going forward, motor policies are an attribute of `MotorSystem` instances, specifically: `motor_system._policy`.

`MotorSystem` configuration changes as follows:
- `motor_system_class` is always now an instance of `MotorSystem`
- `motor_system_args` is now a dictionary `{"policy_class":..., "policy_args":...}` where `policy_args` contain what used to be in `motor_system_args`, i.e., the policy configuration.

For example:
```python
@dataclass
class MotorSystemConfig:
    motor_system_class: MotorSystem = MotorSystem
    motor_system_args: Union[Dict, Dataclass] = field(
        default_factory=lambda: dict(
            policy_class=BasePolicy,
            policy_args=make_base_policy_config(
                action_space_type="distant_agent",
                action_sampler_class=UniformlyDistributedSampler,
            ),
        )
    )
```

`MotorSystem` initialization changes as follows, new portions highlighted with comments:
```python
        # Create motor system
        motor_system_config = monty_config.pop("motor_system_config")
        motor_system_class = motor_system_config["motor_system_class"]
        motor_system_args = motor_system_config["motor_system_args"]
        assert issubclass(motor_system_class, MotorSystem)  # expecting MotorSystem class
        policy_class = motor_system_args["policy_class"]  # extract "policy_class" from motor_system_args
        policy_args = motor_system_args["policy_args"]  # extract "policy_args" from motor_system_args
        assert issubclass(policy_class, MotorPolicy)  # policy now inherits from MotorPolicy
        policy = policy_class(rng=self.rng, **policy_args)  # policy class gets rng during initialization
        motor_system = motor_system_class(policy=policy)  # motor system only gets policy on initialization
```

All direct `motor_system` accesses are now `motor_system._policy` accesses, highlighting that we are reaching into intended-to-be-private internals of motor system implementation. Follow on pull requests will address this.

PR #263, opened 2025-04-17
refactor!: state is stored in motor system instead of policies

tristanls:
In a previous refactor (#234), I separated the `MotorSystem` from `MotorPolicy` and set things up so that a policy was a `motor_system._policy` attribute. However, the motor system state remained attached to the motor policy.

Since we want to be able to swap motor policies, we want to maintain the motor system state outside of the motor policy.

This pull request consolidates motor system state to be a `MotorSystem._state` property. Additionally, since motor policies no longer track motor system state, all motor policy code is passed the motor state explicitly via a `state: Optional[MotorSystemState]` argument.

One consequence of this is in dataloaders (`embodied_data.py`). We see two directly invoked policy methods being passed the motor system state explicitly, for example:
```python
                self._action = self.motor_system._policy.touch_object(
                    self._observation, view_sensor_id="view_finder"
                    self._observation,
                    view_sensor_id="view_finder",
                    state=self.motor_system._state,
                )
```
```python
                actions = self.motor_system._policy.orient_to_object(
                    self._observation,
                    sensor_id,
                    target_semantic_id=self.primary_target["semantic_id"],
                    multiple_objects_present=multiple_objects_present,
                    state=self.motor_system._state,
                )
```
Otherwise, most explicit state passing is confined to `motor_policies.py` internally.

Two outliers in `motor_policies.py` indicate a need for future refactoring: `is_motor_only_step()` and `convert_motor_state()`. 

`is_motor_only_step()` seems to no longer belong on a policy, as it is solely concerned with the motor system, but it is attached to `agent_id`, which is still tied to policy.

Similarly, `convert_motor_state`, only converts motor system state. ~~This particular one is left in place in case of weird things with the state assignment that InformedPolicy does. A future goal is to move it to the motor system after further analysis.~~ This was moved to `MotorSystemState.convert_motor_state` (unaltered for now).

I added a dash of unit tests for `BasePolicy` since the methods are easier to test with `state` being explicitly passed in. I deferred writing unit tests for other policies as they look complicated. I figure we'll get around to those eventually.

FYI: @scottcanoe @hlee9212 @ramyamounir @jeremyshoemaker

### DIFF
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -27,7 +27,10 @@ from tbp.monty.frameworks.models.motor_policies import (
     SurfacePolicy,
 )
 from tbp.monty.frameworks.models.motor_system import MotorSystem
-from tbp.monty.frameworks.models.motor_system_state import MotorSystemState
+from tbp.monty.frameworks.models.motor_system_state import (
+    MotorSystemState,
+    ProprioceptiveState,
+)
 
 from .embodied_environment import EmbodiedEnvironment
 
@@ -86,7 +89,7 @@ class EnvironmentDataset(Dataset):
 
         if self.transform is not None:
             observation = self.apply_transform(self.transform, observation, state)
-        return observation, MotorSystemState(state) if state else None
+        return observation, ProprioceptiveState(state) if state else None
 
     def close(self):
         self.env.close()
@@ -104,7 +107,7 @@ class EnvironmentDataset(Dataset):
         state = self.env.get_state()
         if self.transform is not None:
             observation = self.apply_transform(self.transform, observation, state)
-        return observation, MotorSystemState(state) if state else None
+        return observation, ProprioceptiveState(state) if state else None
 
     def __len__(self):
         return math.inf
@@ -139,14 +142,16 @@ class EnvironmentDataLoader:
         self.dataset = dataset
         self.motor_system = motor_system
         self.rng = rng
-        self._observation, self.motor_system._state = self.dataset.reset()
+        self._observation, state = self.dataset.reset()
+        self.motor_system._state = MotorSystemState(state)
         self._action = None
         self._amount = None
         self._counter = 0
 
     def __iter__(self):
         # Reset the environment before iterating
-        self._observation, self.motor_system._state = self.dataset.reset()
+        self._observation, state = self.dataset.reset()
+        self.motor_system._state = MotorSystemState(state)
         self._action = None
         self._amount = None
         self._counter = 0
@@ -160,7 +165,8 @@ class EnvironmentDataLoader:
         else:
             action = self.motor_system()
             self._action = action
-            self._observation, self.motor_system._state = self.dataset[action]
+            self._observation, state = self.dataset[action]
+            self.motor_system._state = MotorSystemState(state)
             self._counter += 1
             return self._observation
 
@@ -381,6 +387,7 @@ class EnvironmentDataLoaderPerObject(EnvironmentDataLoader):
     def reset_agent(self):
         logging.debug("resetting agent------")
         self._observation, state = self.dataset.reset()
+        state = MotorSystemState(state)
         self._counter = 0
 
         # Make sure to also reset action variables when resetting agent during
@@ -461,6 +468,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                 )
 
             self._observation, state = self.dataset[self._action]
+            state = MotorSystemState(state)
 
             # Check whether sensory information is just for feeding back to motor policy
             # TODO refactor so that the motor policy itself is making this update
@@ -576,7 +584,8 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                     state=self.motor_system._state,
                 )
                 for action in actions:
-                    self._observation, self.motor_system._state = self.dataset[action]
+                    self._observation, state = self.dataset[action]
+                    self.motor_system._state = MotorSystemState(state)
 
         if allow_translation:
             # Move closer to the object, if not already close enough
@@ -589,7 +598,8 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
             # Continue moving to a close distance to the object
             while not close_enough:
                 logging.debug("moving closer!")
-                self._observation, self.motor_system._state = self.dataset[action]
+                self._observation, state = self.dataset[action]
+                self.motor_system._state = MotorSystemState(state)
                 action, close_enough = self.motor_system._policy.move_close_enough(
                     self._observation,
                     sensor_id,
@@ -613,7 +623,8 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                 state=self.motor_system._state,
             )
             for action in actions:
-                self._observation, self.motor_system._state = self.dataset[action]
+                self._observation, state = self.dataset[action]
+                self.motor_system._state = MotorSystemState(state)
             on_target_object = self.motor_system._policy.is_on_target_object(
                 self._observation,
                 sensor_id,
@@ -706,7 +717,8 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
             rotation_quat=quaternion.one,
         )
         _, _ = self.dataset[set_agent_pose]
-        self._observation, self.motor_system._state = self.dataset[set_sensor_rotation]
+        self._observation, state = self.dataset[set_sensor_rotation]
+        self.motor_system._state = MotorSystemState(state)
 
         # Check depth-at-center to see if the object is in front of us
         # As for methods such as touch_object, we use the view-finder
@@ -817,7 +829,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                 == pre_jump_state["sensors"][current_sensor]["rotation"]
             ), "Failed to return sensor to orientation"
 
-        self.motor_system._state = state
+        self.motor_system._state = MotorSystemState(state)
 
         # TODO explore reverting to an attempt with touch_object here,
         # only moving back to our starting location if this is unsuccessful
@@ -859,7 +871,8 @@ class OmniglotDataLoader(EnvironmentDataLoaderPerObject):
             )
         self.dataset = dataset
         self.motor_system = motor_system
-        self._observation, self.motor_system._state = self.dataset.reset()
+        self._observation, state = self.dataset.reset()
+        self.motor_system._state = MotorSystemState(state)
         self._action = None
         self._amount = None
         self._counter = 0
@@ -948,7 +961,8 @@ class SaccadeOnImageDataLoader(EnvironmentDataLoaderPerObject):
             )
         self.dataset = dataset
         self.motor_system = motor_system
-        self._observation, self.motor_system._state = self.dataset.reset()
+        self._observation, state = self.dataset.reset()
+        self.motor_system._state = MotorSystemState(state)
         self._action = None
         self._amount = None
         self._counter = 0
@@ -1041,7 +1055,8 @@ class SaccadeOnImageFromStreamDataLoader(SaccadeOnImageDataLoader):
         # TODO: call super init instead of duplication code & generally clean up more
         self.dataset = dataset
         self.motor_system = motor_system
-        self._observation, self.motor_system._state = self.dataset.reset()
+        self._observation, state = self.dataset.reset()
+        self.motor_system._state = MotorSystemState(state)
         self._action = None
         self._amount = None
         self._counter = 0
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -454,7 +454,9 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                 and self._action is None
             ):
                 self._action = self.motor_system._policy.touch_object(
-                    self._observation, view_sensor_id="view_finder"
+                    self._observation,
+                    view_sensor_id="view_finder",
+                    state=self.motor_system._policy.state,
                 )
 
             self._observation, state = self.dataset[self._action]
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -142,16 +142,20 @@ class EnvironmentDataLoader:
         self.dataset = dataset
         self.motor_system = motor_system
         self.rng = rng
-        self._observation, state = self.dataset.reset()
-        self.motor_system._state = MotorSystemState(state) if state else None
+        self._observation, proprioceptive_state = self.dataset.reset()
+        self.motor_system._state = (
+            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
+        )
         self._action = None
         self._amount = None
         self._counter = 0
 
     def __iter__(self):
         # Reset the environment before iterating
-        self._observation, state = self.dataset.reset()
-        self.motor_system._state = MotorSystemState(state) if state else None
+        self._observation, proprioceptive_state = self.dataset.reset()
+        self.motor_system._state = (
+            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
+        )
         self._action = None
         self._amount = None
         self._counter = 0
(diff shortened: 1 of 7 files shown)


src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -138,14 +138,14 @@ def __init__(self, dataset: EnvironmentDataset, motor_system: MotorSystem, rng):
         self.dataset = dataset
         self.motor_system = motor_system
         self.rng = rng
-        self._observation, self.motor_system._policy.state = self.dataset.reset()
+        self._observation, self.motor_system._state = self.dataset.reset()
         self._action = None
         self._amount = None
         self._counter = 0
```
jeremyshoemaker, line 144: This is probably outside the scope of this PR, but what is `_counter` counting? It really needs a better name.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -221,20 +248,34 @@ def post_episode(self):
...
-            Agent state.
+            (AgentState): Agent state.
         """
-        return self.state[self.agent_id]
+        return state[self.agent_id]
 
-    @property
-    def is_motor_only_step(self):
-        agent_state = self.get_agent_state()
+    def is_motor_only_step(self, state: MotorSystemState):
```
jeremyshoemaker, line 265: I'm bothered that this is no longer a property, and it makes me think maybe this functionality should be moved up into whatever is calling it. It will have access to the `state` and the `policy` and can call `policy.get_agent_state(state)` and do the rest of the checks in this method. 

`get_agent_state` also feels misnamed because it's not a simple getter. These two things seem to be tangled together more than they should be. The state isn't owned by the policy, so it shouldn't be gotten (getted?) from the policy but instead whatever owns the state.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -498,14 +546,16 @@ def fixme_undo_last_action(self):
...
+    ) -> None:
         self.action = action
         self.timestep += 1
         self.episode_step += 1
-        state_copy = self.convert_motor_state()
+        state_copy = self.convert_motor_state(state)
         self.action_sequence.append([action, state_copy])
 
-    def convert_motor_state(self):
+    def convert_motor_state(self, state: MotorSystemState):
```
jeremyshoemaker, line 558: I'd have to open this in PyCharm to be sure, but I think it would tell me that this could be a static method, since `self` is no longer referenced.

Similar to my comments above, I think this method belongs on the MotorSystemState and not the policy.

src/tbp/monty/frameworks/models/motor_system_state.py
jeremyshoemaker, line 1: Why not have these be `dataclass`es? That way an IDE would be able to complete the field names better and you'd be less likely to get a KeyError at runtime. Maybe the first part already works with the type hinting, but it still seems better to me to do `state.position` instead of `state["position"]`.

jeremyshoemaker (commented):
When I got to the bottom and realized that `MotorSystemState` was a type alias for a dictionary, a lot of my comments become more complicated. It feels like it should probably be a proper class with some of the functionality that is currently in the policy moved there. It could always extend the dictionary class to keep that functionality if we really want that.

src/tbp/monty/frameworks/models/motor_policies.py
tristanls: The problem isn't moving it to whatever has access to `state`; this property should not be stored on the (motor system) `state` at all. This change surfaces that (existing) issue by explicitly passing in `state`. The entire `motor_only_step` in motor system state is going to go away as per the docstring TODO:
> TODO: This information is currently stored in motor system state, but
        should be stored in the policy state instead as it is tracking policy
        state, not motor system state. This will remove MotorSystemState param.

Once it doesn't take any parameters, we can make it a `@property` again.

Yes, `get_agent_state` is misnamed because it has `get_` in it 😛. However, it will also eventually disappear, as it only slices out the agent state from the global state. Instead, as you say, policy should only receive the state for whatever agent is relevant and not know that there is a global state with all agent states.

src/tbp/monty/frameworks/environments/embodied_data.py
tristanls: It is a hack to distinguish the first from the rest. Some examples:
```python
    def __next__(self):
        if self._counter == 0:
            # Return first observation after 'reset' before any action is applied
            self._counter += 1
            return self._observation
        else:
            action = self.motor_system()
            self._action = action
            self._observation, self.motor_system._state = self.dataset[action]
            self._counter += 1
            return self._observation
```
```python
    def __next__(self):
        if self._counter == 0:
            return self.first_step()
        ...
```

jeremyshoemaker (approved):

tristanls:
Thanks @jeremyshoemaker!, could you take another look at my latest commit: https://github.com/thousandbrainsproject/tbp.monty/pull/263/commits/47414adebbefad06d5a8a58c8f763ce3f38e7691

jeremyshoemaker (approved):

src/tbp/monty/frameworks/models/motor_system_state.py
tristanls: Historical reasons. There was nothing before this, so I only wanted to document the dictionary convention and assumed I wouldn't instantiate these anywhere. However, since you pointed it out, I now instantiate `MotorSystemState(state)` in https://github.com/thousandbrainsproject/tbp.monty/pull/263/commits/47414adebbefad06d5a8a58c8f763ce3f38e7691 . And since these now get instantiated, then yeah, dataclass seems more appropriate. Will take another look and see if I can make it work.

tristanls: @jeremyshoemaker, although, any objections to doing this in a separate PR? This change will touch a lot more code and I'd rather separate it out for review.

jeremyshoemaker: Yeah, sure. I kinda thought it might touch a lot of things and be better in a separate PR anyway. Probably should have mentioned that in my comment.

src/tbp/monty/frameworks/models/motor_policies.py
tristanls: It occurred to me that `get_agent_state` will disappear during the refactor, where policies will no longer get `MotorSystemState` but only `AgentState`.

jeremyshoemaker: That's the difficulty with all this being piecemeal, not that I'd want to do it all at once. It's hard to notice that looking at just this PR.

tristanls: I hesitate to touch `get_agent_state` for now because I am still a bit ambiguous about the correspondence between agent, policy, and the motor system (as they currently are). In the code today, it seems there is a 1:1:1 agent:policy:motor-system correspondence right now. However, I'm unclear (until we talk more with the team) what the future correspondence should be. I know we want policy:motor-system correspondence to be n:1. However, is agent:policy correspondence 1:1 or n:1? If it is 1:1, then no need for `get_agent_state`, but if it's n:1, then this method might remain.

src/tbp/monty/frameworks/models/motor_system_state.py
```diff
@@ -0,0 +1,102 @@
...
+    sensors: Dict[str, SensorState]
+    """The proprioceptive state of the agent's sensors."""
+    position: Any  # TODO: Stop using magnum.Vector3 and decide on Monty standard
+    """The agent's position relative to some global reference frame."""
+    rotation: Any  # TODO: Stop using quaternion.quaternion and decide on Monty standard
+    """The agent's rotation relative to some global reference frame."""
+
+
+class MotorSystemState(Dict[str, AgentState]):
+    """The proprioceptive state of the motor system."""
```
nielsleadholm, line 35: **suggestion:** I think it would be clearer to omit `proprioceptive` in the case of the `MotorSystemState`. 

While at the moment it may only be a dictionary of `AgentState`s, intuitively it would be more than this, because the motor system will also orchestrate the policies - i.e. its state is also about coordination and interfacing with the LMs. For example, part of its state could be whether it should only be performing model-based actions (blocking all model-free policies), analogous to the cortex overriding reflexive actions.

If we want to specifically keep the information to just proprioception type information, then I think that should be in the class name.

nielsleadholm (approved):
Great stuff, just a couple high level comments for discussion.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -27,6 +27,7 @@
     SurfacePolicy,
 )
 from tbp.monty.frameworks.models.motor_system import MotorSystem
+from tbp.monty.frameworks.models.motor_system_state import MotorSystemState
```
vkakerbeck, line 30: Why is this called `MotorSystemState`? Isn't it the state of the agent that is passed to the motor system since some policies rely on the agent state to decide the next action?

src/tbp/monty/frameworks/models/monty_base.py
```diff
@@ -437,11 +437,11 @@ def get_agent_state(self):
         Returns:
             State of the agent.
         """
-        return self.motor_system._policy.get_agent_state()
+        return self.motor_system._policy.get_agent_state(self.motor_system._state)
```
vkakerbeck, line 440: Why do we need to call `get_agent_state` on the policy by passing it the state? Wouldn't it make much more sense to do `return self.motor_system._state` or `return self.motor_system.get_agent_state()`? If it is just because of the `agent_id` property of the policy, couldn't that be extracted in a `get_agent_state` function of the motor system? like `return self._state[self.policy.agent_id]`

src/tbp/monty/frameworks/models/motor_system_state.py
vkakerbeck: Interesting, I hadn't thought about including LM command in the motor system's state. I agree that if we just store information about the agent's state here (which we currently do) we should reflect this in the class name (calling it `AgentState` for example). This was confusing me a bit when reading this PR. If we plan to include other info I am okay with calling it `MotorSystemState` but I am still not super sure what kind of other information this will be and if a general state variable is the best place to store it. We could also keep calling it `AgentState` for now and then change it once the need arises.

tests/unit/frameworks/models/motor_policies_test.py
```diff
@@ -0,0 +1,100 @@
...
+    def setUp(self):
+        self.rng = np.random.RandomState(42)
+        self.agent_id = f"agent_id_{self.rng.randint(0, 999_999_999)}"
+        self.default_sensor_state: SensorState = {
+            "position": (0.0, 0.0, 0.0),
+            "rotation": (1.0, 0.0, 0.0, 0.0),
+        }
+        self.agent_sensors: Dict[str, SensorState] = {
+            f"sensor_id_{self.rng.randint(0, 999_999_999)}": self.default_sensor_state,
+        }
```
vkakerbeck, line 35: Do the old ways of defining agent state still work (just nested dictionaries)? Do I need to update anything in the tutorials (open PR)? I think it's fine because the conversion happens after the state dict is returned from the dataloader, right?

vkakerbeck (commented):
Nice update! Just a few small comments, mostly around what MotorSystemState actually is and how it is different from AgentState

src/tbp/monty/frameworks/models/motor_system_state.py
nielsleadholm: Yeah agree we should keep it simple if we can. A couple variables that might already be relevant to include here are `driving_goal_state` and `use_goal_state_driven_actions`. These are currently part of `motor_system._policy`, but intuitively, the motor system (motor module) should receive the goal-state, and decide what model-free policies to use in order to satisfy it.

vkakerbeck: Good points. I am not opposed to the idea of having a `MotorSystemState` that keeps track of all these things. Thinking about it a bit more, the main thing that bothers me is that in `self._observation, self.motor_system._state = self.dataset[action]` the agent state suddenly becomes the `MotorSystemState`. I would at least want to preserve the notion that this is the state of the agent, which does not equate to the state of the motor system. Maybe something like `motor_system._state[agent_state] = AgentState`.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -221,20 +248,34 @@ def post_episode(self):
...
+        return state[self.agent_id]
 
-    @property
-    def is_motor_only_step(self):
-        agent_state = self.get_agent_state()
+    def is_motor_only_step(self, state: MotorSystemState):
+        """Check if the current step is a motor-only step.
+
+        TODO: This information is currently stored in motor system state, but
+        should be stored in the policy state instead as it is tracking policy
```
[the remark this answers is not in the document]
tristanls: @nielsleadholm, I don't have a preferred way of tracking this. It tracks itself as far as I'm concerned since it bugs me every time I see it in code 🙂. 

Regarding purpose (1), skipping passing the first sensory input will be obsolete with the introduction of positioning procedures. The first sensory input to Monty will always be the correct input. I already have a GetGoodView in a draft PR, which sets up exactly that.

src/tbp/monty/frameworks/models/motor_system_state.py
tristanls: > Thinking about it a bit more, the main thing that bothers me is that in self._observation, self.motor_system._state = self.dataset[action] the agent state suddenly becomes the MotorSystemState. I would at least want to preserve the notion that this is the state of the agent, which does not equate to the state of the motor system. Maybe something like motor_system._state[agent_state] = AgentState.

I don't want to solve this here. The Dataloder should not be setting the state of the motor system.

Eventually, I intend to have the Dataloader return `(observation, state)` to the `MontyExperiment`, which then invokes `Monty.step(observation, state)`.

```mermaid
sequenceDiagram
  participant ME as MontyExperiment
  participant M as Monty
  participant DL as Dataloader
  participant D as Dataset

  
  loop
  ME ->>+ M: step(observation, state)
  activate ME
  M -->>- ME: action
  ME ->>+ DL: step(action)
  DL ->>+ D: __getitem__(action)
  D -->>- DL: (observation, state)
  DL -->>- ME: (observation, state)
  deactivate ME
  end
```

I see the distinction y'all are making between the proprioceptive and other motor system states. I'll distinguish between them, showing that the proprioceptive state comes from the dataset, while the motor system tracks that and additional state.

Also, I need different names for a single agent's state vs the dictionary of agent states. So, AgentState is the name I chose for a single agent's state. I propose ProprioceptiveState as the name for a collection of AgentStates.

vkakerbeck: That sounds perfect

src/tbp/monty/frameworks/models/monty_base.py
vkakerbeck: okay that makes a lot of sense. Feel free to keep track of this however works best for you. If you add a TODO it would make it more obvious to me that you are already planning on changing this and I won't feel compelled commenting on it. But if you already have another system of tracking this I don't want to create any overhead for you.

tests/unit/frameworks/models/motor_policies_test.py
vkakerbeck: I was thinking of the `get_state` function in the DataLoader (like here https://github.com/thousandbrainsproject/tbp.monty/blob/6bf52739e17fa68cf047556618b9507408016725/src/tbp/monty/frameworks/environments/two_d_data.py#L149 ) But looking at it now I see that your PR won't change anything about that, it just converts once it gets to the motor system.

tristanls:
Per the discussion, I made an update to distinguish `ProprioceptiveState` from `MotorSystemState`.

This results in the general pattern of the Dataloader returning `(observation, ProprioceptiveState)` tuple.

With this distinction in place, we now convert `ProprioceptiveState` to the more general `MotorSystemState` when assigning to the motor system. I think this also does a good job of highlighting that these are two distinct states. The general pattern now is:
```python
self._observation, state = self.dataset.reset() # or self.dataset[action]
self.motor_system._state = MotorSystemState(state) if state else None
```

Asking for a review again @jeremyshoemaker @nielsleadholm as this seems worth an additional look.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -138,14 +142,16 @@ def __init__(self, dataset: EnvironmentDataset, motor_system: MotorSystem, rng):
         self.dataset = dataset
         self.motor_system = motor_system
         self.rng = rng
-        self._observation, self.motor_system._policy.state = self.dataset.reset()
+        self._observation, state = self.dataset.reset()
```
nielsleadholm, line 145: **suggestion:** Just looking at this again, I wonder whether it's worth assigning to `proprioceptive_state` e.g.:

```python
self._observation, proprioceptive_state = self.dataset.reset()
```

given `state` is quite generic and may cause confusion as to what it actually contains. Based on my understanding of the long term plan (Dataloader returns `(observation, state)`), I think that would still be a helpful clarification.

nielsleadholm (approved):
Thanks for that change, looks good!

src/tbp/monty/frameworks/environments/embodied_data.py
tristanls: I did as you suggested. In a few places I used `proprio_state` for better readability when running out of space.

PR #264, opened 2025-04-17
refactor!: remove unused policy is_exploring property

tristanls:
The _policy_ `is_exploring` property was set to `False` in `pre_episode` and never changed.

This pull request removes the property since it is unused.

I removed the check `if self.is_exploring is False:` in `InformedPolicy.post_action()` implementation. In the usual case, `pre_episode()` would set `self.is_exploring = True`. Technically, it is possible for `self.is_exploring` to be `None` where `InformedPolicy.use_goal_state_driven_actions` is `True`. In this case, `JumpToGoalStateMixin.pre_episode()` is called, which _does not set_ `self.is_exploring = False`. It does not strike me that deciding whether to append a copy of motor state to an action sequence depends on `InformedPolicy.use_goal_state_driven_actions`. However, if that is the case, I can replace `if self.is_exploring is False:` with `if self.use_goal_state_driven_actions:`. Although, that still might not be an exact replacement as it is unclear whether `InformedPolicy.use_goal_state_drive_actions` changes throughout execution.


ramyamounir (approved):

scottcanoe (approved):
LGTM. Thanks for cleaning this stuff up -- this is going to save me some time/confusion down the road.

PR #266, opened 2025-04-18
refactor: modernize syntax (UP031)

hlee9212:
[UP031](https://docs.astral.sh/ruff/rules/printf-string-formatting/)


vkakerbeck (approved):
Nice!

PR #267, opened 2025-04-18
refactor: update deprecated syntax (UP005)

hlee9212:
Not critical yet but `assertEquals` will be deprecated from python 3.12 ([UP005](https://docs.astral.sh/ruff/rules/deprecated-unittest-alias/)).


vkakerbeck (approved):
LGTM

PR #268, opened 2025-04-18
fix: temporary fix for featLM voting logic

vkakerbeck:
This is really more of a hacky patch than a fix. The featureLM voting logic looked at the number of incoming location votes to determine how many of them need to agree. However, a signed sending LM might have multiple hypotheses. For this to be correct, the receiving LM needs to use the number of LMs that it receives input from to determine k. However, I don't think it currently has access to this information so I hardcoded it to 4 since all our voting experiments use 5LMs. I would spend more time on an actual solution if this was the evidenceLM but since we don't really use the featureLM anymore I didn't want to go too far down this path.

@jeremyshoemaker this should be the last remaining fix necessary for the 5LM unit test to work (in addition to the changes on my branch here: https://github.com/vkakerbeck/tbp.monty/tree/my_5lm_test_branch )

### DIFF
--- a/src/tbp/monty/frameworks/models/feature_location_matching.py
+++ b/src/tbp/monty/frameworks/models/feature_location_matching.py
@@ -160,12 +160,12 @@ class FeatureGraphLM(GraphLM):
                 # Check that object is still in matches after ID update
                 if possible_obj in self.possible_matches:
                     # TODO: better way to dynamically adapt k
-                    if vote_data["pos_location_votes"][possible_obj].shape[0] < 5:
+                    if vote_data["pos_location_votes"][possible_obj].shape[0] < 4:
                         k = vote_data["pos_location_votes"][possible_obj].shape[0]
                         print(f"only received {k} votes")
                     else:
-                        k = 5
-
+                        # TODO: k should not be > num_lms
+                        k = 4
                     vote_location_tree = KDTree(
                         vote_data["pos_location_votes"][possible_obj],
                         leaf_size=2,


src/tbp/monty/frameworks/models/feature_location_matching.py
```diff
@@ -160,12 +160,12 @@ def receive_votes(self, vote_data):
...
                 if possible_obj in self.possible_matches:
                     # TODO: better way to dynamically adapt k
-                    if vote_data["pos_location_votes"][possible_obj].shape[0] < 5:
+                    if vote_data["pos_location_votes"][possible_obj].shape[0] < 4:
                         k = vote_data["pos_location_votes"][possible_obj].shape[0]
                         print(f"only received {k} votes")
                     else:
-                        k = 5
-
+                        # TODO: k should not be > num_lms
```
tristanls, line 167: **question (non-blocking)**: Should this comment say `should not be > (num_lms - 1)`? If I understand correctly, we assume a 5 LM experiment, so any LM will receive a maximum of 4 (i.e. 5 - 1) votes.

vkakerbeck: yes, good catch!

src/tbp/monty/frameworks/models/feature_location_matching.py
```diff
@@ -160,12 +160,12 @@ def receive_votes(self, vote_data):
                 # Check that object is still in matches after ID update
                 if possible_obj in self.possible_matches:
                     # TODO: better way to dynamically adapt k
-                    if vote_data["pos_location_votes"][possible_obj].shape[0] < 5:
+                    if vote_data["pos_location_votes"][possible_obj].shape[0] < 4:
```
jeremyshoemaker, line 163: Normally I would say that this "[magic constant](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants)" 4 should be replaced with an actual constant, especially since it's used again below and without it, it's not obvious that the two 4s are related. 

But since this is a temporary fix, it's probably fine.

jeremyshoemaker (approved):
Looks good to me

src/tbp/monty/frameworks/models/feature_location_matching.py
vkakerbeck: that's a good point. I updated it

nielsleadholm (commented):
In principal looks good, but just a high level comment/question - as you say, we don't use the featureLM anymore. Would it be better to actually remove the implementation of voting for the Feature LM, as well as the associated unit tests, rather than introduce a hard-coded hack? Are we gaining anything by keeping them, or is it just making the code more complex? We could basically just have voting as an EvidenceLM only implementation.

vkakerbeck:
Hm we could consider this. I think one argument against it is that many unit tests still use this LM that would have to be rewritten to use evidenceLM so its not as easy as just deleting some code. Also just to point out, this PR does not introduce the hardcoded hack, I just replaced the hard-coded 5 with a hardcoded 4. Then I rewrote it a bit to make it clearer that this is hardcoded.

nielsleadholm (approved):
Ok yeah, maybe it's a wider conversation about collapsing all of the LMs into a single evidence-like LM, and getting rid of pretty much everything outside of that. But yeah that would be a major change.

PR #269, opened 2025-04-18
test: refactor 5LM test to use fake observations

### LINKED ISSUE #117: test_5lm_feature_experiment fails

jeremyshoemaker:
This test currently fails on Linux machines (including in GH actions when run on a GPU instance) for unknown reasons, so we're trying to change the test to use faked observations instead of observations from HabitatSim.

Fixes #117

### DIFF
--- a/tests/unit/graph_learning_test.py
+++ b/tests/unit/graph_learning_test.py
@@ -16,7 +16,7 @@ import unittest
 from dataclasses import dataclass, field
 from pathlib import Path
 from pprint import pprint
-from typing import Dict, Union
+from typing import Any, Dict, List, Union
 
 import numpy as np
 import pandas as pd
@@ -46,6 +46,7 @@ from tbp.monty.frameworks.experiments import MontyObjectRecognitionExperiment
 from tbp.monty.frameworks.loggers.wandb_handlers import DetailedWandbMarkedObsHandler
 from tbp.monty.frameworks.models.displacement_matching import DisplacementGraphLM
 from tbp.monty.frameworks.models.feature_location_matching import FeatureGraphLM
+from tbp.monty.frameworks.models.graph_matching import GraphLM
 from tbp.monty.frameworks.models.motor_system import MotorSystem
 from tbp.monty.frameworks.utils.follow_up_configs import (
     create_eval_config_multiple_episodes,
@@ -99,6 +100,36 @@ class MotorSystemConfigOffObject:
             ),
         )
     )
+
+
+@dataclass
+class ObjectEpisode:
+    """An episode of observations for a single named object."""
+    name: str
+    observations: List[Any]
+
+    def __len__(self):
+        return len(self.observations)
+
+
+@dataclass
+class GraphModule:
+    """Class to bundle up a GraphLM and its object observations."""
+    learning_module: GraphLM
+    episodes: List[ObjectEpisode]
+
+    @property
+    def mode(self) -> str:
+        """Helper property to make setting and reading the LM mode easier."""
+        return self.learning_module.mode
+
+    @mode.setter
+    def mode(self, mode: str):
+        self.learning_module.mode = mode
+
+    def pre_episode(self, primary_target):
+        """Delegates pre_episode calls to the LM."""
+        self.learning_module.pre_episode(primary_target)
 
 
 class GraphLearningTest(BaseGraphTestCases.BaseGraphTest):
@@ -1317,6 +1348,29 @@ class GraphLearningTest(BaseGraphTestCases.BaseGraphTest):
         eval_stats = pd.read_csv(os.path.join(exp.output_dir, "eval_stats.csv"))
         self.check_eval_results(eval_stats)
 
+    def gm_learn_object(self, graph_lm, obj_name, observations, offset=None):
+        if offset is None:
+            offset = np.zeros(3)
+
+        graph_lm.mode = "train"
+        graph_lm.pre_episode(self.placeholder_target)
+
+        offset_obs = []
+        for observation in observations:
+            obs_to_learn = copy.deepcopy(observation)
+            obs_to_learn.location += offset
+            offset_obs.append(obs_to_learn)
+            graph_lm.exploratory_step([obs_to_learn])
+
+        graph_lm.detected_object = obj_name
+        graph_lm.detected_rotation_r = None
+        graph_lm.buffer.stats["detected_location_rel_body"] = (
+            graph_lm.buffer.get_current_location(input_channel="first")
+        )
+
+        graph_lm.post_episode()
+        return offset_obs
+
     def get_gm_with_fake_object(self):
         graph_lm = FeatureGraphLM(
             max_match_distance=0.005,
@@ -1327,21 +1381,10 @@ class GraphLearningTest(BaseGraphTestCases.BaseGraphTest):
                 }
             },
         )
-        graph_lm.mode = "train"
-        for observation in self.fake_obs_learn:
-            graph_lm.exploratory_step([observation])
-        graph_lm.detected_object = "new_object0"
-        graph_lm.detected_rotation_r = None
-        graph_lm.buffer.stats["detected_location_rel_body"] = (
-            graph_lm.buffer.get_current_location(input_channel="first")
-        )
 
-        self.assertEqual(
-            len(graph_lm.buffer.get_all_locations_on_object(input_channel="first")),
-            4,
-            "Should have stored exactly 4 locations in the buffer.",
-        )
-        graph_lm.post_episode()
+        self.gm_learn_object(graph_lm, obj_name="new_object0",
+                             observations=self.fake_obs_learn)
+
         self.assertEqual(
             len(graph_lm.get_all_known_object_ids()),
             1,
@@ -1353,6 +1396,29 @@ class GraphLearningTest(BaseGraphTestCases.BaseGraphTest):
             "Learned object ID should be new_object0.",
         )
         return graph_lm
+
+    def get_5lm_gm_with_fake_objects(self, objects) -> List[GraphModule]:
+        graph_lms = []
+        for lm in range(5):
+            graph_lm = FeatureGraphLM(
+                max_match_distance=0.005,
+                tolerances={
+                    "patch": {
+                        "hsv": [0.1, 1, 1],
+                        "principal_curvatures_log": [1, 1],
+                    }
+                },
+            )
+            object_obs = []
+            for i, obj in enumerate(objects):
+                obj_name = f"new_object{i}"
+                offset_obs = self.gm_learn_object(graph_lm, obj_name=obj_name,
+                                                  observations=obj,
+                                                  offset=self.lm_offsets[lm])
+                object_obs.append(ObjectEpisode(obj_name, offset_obs))
+            graph_lms.append(GraphModule(graph_lm, object_obs))
+
+        return graph_lms
 
     def test_same_sequence_recognition(self):
         """Test that the object is recognized with same action sequence."""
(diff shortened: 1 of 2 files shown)


tests/unit/graph_learning_test.py
jeremyshoemaker, line 1: I'm not certain I like the names of the helper classes I'm adding here, so I'm open to suggestions for better alternatives. They _do_ make the test easier to read to my eyes though.

jeremyshoemaker:
This PR depends on #268, so I'll need to rebase onto `main` once that one is merge in.

jeremyshoemaker:
All the tests pass on my Linux machine. The force push was me rebasing my copy of the branch onto main so I could test locally. Since that's the version I tested, I'd rather have that be the version that's merged in, even if it's the same as the one with the merge from main.

tests/unit/graph_learning_test.py
```diff
@@ -1661,18 +1727,36 @@ def test_5lm_feature_experiment(self):
...
 
             pprint("...evaluating...")
-            exp.evaluate()
+            monty = exp.model
+            monty.set_experiment_mode("eval")
+            monty.learning_modules = [gm.learning_module for gm in graph_modules]
+
+            for gm in graph_modules:
+                gm.mode = "eval"
+                gm.pre_episode(self.placeholder_target)
```
tristanls, line 1740: **issue**: Line 1747 (`monty.pre_episode(self.placeholder_target)`), will call pre_episode on each learning module. Do you think this is necessary?

tests/unit/graph_learning_test.py
```diff
@@ -1366,6 +1409,29 @@ def get_gm_with_fake_object(self):
...
+                        "principal_curvatures_log": [1, 1],
+                    }
+                },
+            )
+            object_obs = []
+            for i, obj in enumerate(objects):
+                obj_name = f"new_object{i}"
+                offset_obs = self.gm_learn_object(graph_lm, obj_name=obj_name,
+                                                  observations=obj,
+                                                  offset=self.lm_offsets[lm])
```
tristanls, line 1429: **nitpick**: The formatting seems unusual. In the Monty codebase, we typically don't align parameters on open parens.

tests/unit/graph_learning_test.py
```diff
@@ -1339,21 +1393,10 @@ def get_gm_with_fake_object(self):
...
-        )
 
-        self.assertEqual(
-            len(graph_lm.buffer.get_all_locations_on_object(input_channel="first")),
-            4,
-            "Should have stored exactly 4 locations in the buffer.",
-        )
-        graph_lm.post_episode()
+        self.gm_learn_object(graph_lm, obj_name="new_object0",
+                             observations=self.fake_obs_learn)
```
tristanls, line 1398: **nitpick**: The formatting seems unusual. In the Monty codebase, we typically don't align parameters on open parens.

tests/unit/graph_learning_test.py
```diff
@@ -98,6 +99,36 @@ class MotorSystemConfigOffObject:
...
+    """An episode of observations for a single named object."""
+    name: str
+    observations: List[Any]
+
+    def __len__(self):
+        return len(self.observations)
+
+
+@dataclass
+class GraphModule:
```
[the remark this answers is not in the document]
jeremyshoemaker: That's better than anything I came up with.

tests/unit/graph_learning_test.py
```diff
@@ -98,6 +99,36 @@ class MotorSystemConfigOffObject:
     )
 
 
+@dataclass
+class ObjectEpisode:
```
vkakerbeck, line 103: Maybe we could name this `EpisodeObservations` or `OneObjectEpisodeObservations` to be a bit more explicit that this contains the observations?

tests/unit/graph_learning_test.py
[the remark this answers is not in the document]
vkakerbeck: Do we need a dataclass for this? It seems a bit of an odd concept to bundle observations with an LM. If we keep it I would agree with changing the name of it. I'm not sure TrainedGraphLM is entirely accurate since it won't always be trained, right? Maybe `LMAndItsObservations` or `ObservationLMPackage`. I can't think of a great name since the concept of combining them into one instance is strange to me.

tests/unit/graph_learning_test.py
```diff
@@ -1339,21 +1393,10 @@ def get_gm_with_fake_object(self):
...
-            graph_lm.buffer.get_current_location(input_channel="first")
-        )
 
-        self.assertEqual(
-            len(graph_lm.buffer.get_all_locations_on_object(input_channel="first")),
-            4,
-            "Should have stored exactly 4 locations in the buffer.",
-        )
-        graph_lm.post_episode()
+        self.gm_learn_object(graph_lm, obj_name="new_object0",
```
vkakerbeck, line 1397: Doesn't this function return `offset_obs`?

tests/unit/graph_learning_test.py
```diff
@@ -1661,18 +1727,36 @@ def test_5lm_feature_experiment(self):
...
-
-            train_stats = pd.read_csv(
-                os.path.join(exp.output_dir, "train_stats.csv")
-            )
-            # The following check is brittle and depends on sensor arrangement. Leaving
-            # the rest of the test intact to detect run failures, but disabling checking
-            # of particular results.
-            # self.check_multilm_train_results(train_stats, num_lms=5, min_done=3)
+            objects = [self.fake_obs_learn, self.fake_obs_house_3d]
+            graph_modules = self.get_5lm_gm_with_fake_objects(objects)
```
vkakerbeck, line 1731: I would just call this `learning_modules` instead of `graph_modules` to avoid confusion

tests/unit/graph_learning_test.py
```diff
@@ -1661,18 +1727,36 @@ def test_5lm_feature_experiment(self):
...
+            monty.set_experiment_mode("eval")
+            monty.learning_modules = [gm.learning_module for gm in graph_modules]
+
+            for gm in graph_modules:
+                gm.mode = "eval"
+                gm.pre_episode(self.placeholder_target)
+
+            exp.pre_epoch()
+            # Use the first graph module's observations for the episodes
+            # since it has no offsets like the others do.
```
vkakerbeck, line 1744: Why does this matter? I thought we are just using this to know how many episodes there are and how many steps in each episode but the observations themselves are extracted for each lm separately on line 1752

tests/unit/graph_learning_test.py
```diff
@@ -1661,18 +1727,36 @@ def test_5lm_feature_experiment(self):
...
+            for episode in graph_modules[0].episodes:
+                exp.pre_episode()
+                monty.pre_episode(self.placeholder_target)
+                for observation in episode.observations:
```
vkakerbeck, line 1748: does this have to have the same name as when we extract observations on line 1752? To me this would be more intuitive to be `for step in range(episode.observations)` (kind of how I did it on my branch on line 1728 and 1732 https://github.com/thousandbrainsproject/tbp.monty/compare/main...vkakerbeck:tbp.monty:my_5lm_test_branch)

vkakerbeck: I am probably missing something about what type `observation` has here and how you extract the observations at each step on the next line. But I am confused where the `step` comes in in general. Like where do you specify which step in the experiment we are in to not keep extracting the first observation each time.

vkakerbeck (commented):
Great to hear that this solves the issue now! I just left a few clarifying comments. I am a bit confused now with the new dataclass logic and just want to make sure we extract the correct observations with the correct offset at each step.

tests/unit/graph_learning_test.py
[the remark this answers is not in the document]
jeremyshoemaker: In the rough code we made while pairing, we had it return a dictionary that contained the two values that are now in this dataclass. We were already bundling them together in that version, this just gives it more structure. I think your `ObservationLMPackage` might be a good name for it.

I'll look at it again and see if there isn't a better way to structure it, though.

tests/unit/graph_learning_test.py
jeremyshoemaker: That's how we did it on the last Python project I worked on, so it's habit. I'll update these call sites.

tests/unit/graph_learning_test.py
jeremyshoemaker: Your right, when I was refactoring this code, I think I misread what we were doing. This isn't accurate at all.

tests/unit/graph_learning_test.py
jeremyshoemaker: Yep, your right, I messed this up during the refactor of our original rough code to the state it's in now. Let me clean it up and see if it makes more sense then.

jeremyshoemaker:
Well, now that I've cleaned up the test code to address @vkakerbeck's feedback, it's failing the test again in the same place.

Walking through it in the debugger, it doesn't appear to be exactly the same as before because some of the calls to `post_episode` work fine. It seems like one of the LMs is getting into a bad state instead of all of them like before.

jeremyshoemaker:
> Well, now that I've cleaned up the test code to address @vkakerbeck's feedback, it's failing the test again in the same place.
> 
> Walking through it in the debugger, it doesn't appear to be exactly the same as before because some of the calls to `post_episode` work fine. It seems like one of the LMs is getting into a bad state instead of all of them like before.

Turns out I was making an invalid assumption. The episodes have a differing number of observations, 4 and 5, respectively, and we were only giving the models the first four observations for the second object, so they couldn't reach a consensus.

This makes me wonder, shouldn't that be a possible state for the LMs to get into? If they can't decide, I don't feel like the code should blow up the way it does.

tests/unit/graph_learning_test.py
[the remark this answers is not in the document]
jeremyshoemaker: Thinking about it some more, I think `TrainedGraphLM` makes the most sense. These are _only_ created in `get_5lm_gm_with_fake_objects`; we're specifically training them on the fake objects there. That should remain true unless someone comes along and reuses this for a different test.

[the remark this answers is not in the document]
vkakerbeck: Okay, maybe we could just expand the docstring in that case to make it clear what exactly this is and how it is supposed to be used. My fear is that someone will search the codebase for `GraphLM` and come across a `TrainedGraphLM` class. They may then assume the wrong things about this class just based on the name.

vkakerbeck:
This just happens because in this unit test we manually loop over the observations and we stop looping over them after 4 or 5 observations. In a normal Monty experiment, Monty can keep moving in the environment until it reaches a terminal state. So normally we use a dataloader to get the observations and the dataloader would be written to keep returning observations when we do `for step, observation in enumerate(self.dataloader):`.

tests/unit/graph_learning_test.py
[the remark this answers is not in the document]
jeremyshoemaker: I'll expand the docstring a bit to make it clear what it's for. I'd be a little concerned if someone found this and tried to use it somewhere in Monty, given that it's defined in a test file, but better to be more explicit about it.

tristanls (approved):

tests/unit/resources/unit_test_utils.py
```diff
@@ -138,6 +146,17 @@ def setUp(self):
...
             self.fake_obs_house_trans = [State(**obs_dic) for obs_dic in config_list]
 
+            fo_house_5 = copy.deepcopy(fo_house)
+            fo_house_5["location"] = np.array([0.5, 1.5, 1.0])
+            self.fake_obs_house_3d = [
+                State(**fo_house),
+                State(**fo_house_1),
+                State(**fo_house_2),
+                State(**fo_house_3),
+                # replacing fo_house_4 with fo_house_5
```
vkakerbeck, line 156:
```suggestion
                # replacing fo_house_4 with fo_house_5 to make it 3D and it's pose unambiguous
```

vkakerbeck (approved):
Great, thanks for the updates!

PR #270, opened 2025-04-18
fix: MontyBase._step_learning_modules now works

tristanls:
We probably never noticed this because `MontyForGraphMatching(MontyBase)` overrides this method with its own implementation. Within that override, you'll notice the correct usage of `getattr`:
```python
                lm_step_method = getattr(self.learning_modules[i], self.step_type)
                assert callable(lm_step_method), f"{lm_step_method} must be callable"
                lm_step_method(sensory_inputs)
```


ramyamounir (approved):
Nice catch!

PR #271, opened 2025-04-18
refactor!: remove unusued Monty._post_step method

tristanls:
This method is unused. While it mentions updating counters, we update counters in `_set_step_type_and_check_if_done`.

### DIFF
--- a/src/tbp/monty/frameworks/models/abstract_monty_classes.py
+++ b/src/tbp/monty/frameworks/models/abstract_monty_classes.py
@@ -27,7 +27,6 @@ class Monty(metaclass=abc.ABCMeta):
         self._pass_goal_states()
         self._pass_infos_to_motor_system()
         self._set_step_type_and_check_if_done()
-        self._post_step()
 
     def _exploratory_step(self, observation):
         """Step format for adding data to an existing model.
@@ -38,7 +37,6 @@ class Monty(metaclass=abc.ABCMeta):
         self._step_learning_modules()
         self._pass_infos_to_motor_system()
         self._set_step_type_and_check_if_done()
-        self._post_step()
 
     @abc.abstractmethod
     def step(self, observation):
@@ -88,11 +86,6 @@ class Monty(metaclass=abc.ABCMeta):
 
         Update what self.is_done returns to the experiment.
         """
-        pass
-
-    @abc.abstractmethod
-    def _post_step(self):
-        """Hook for doing things like updating counters."""
         pass
 
     ###
--- a/src/tbp/monty/frameworks/models/monty_base.py
+++ b/src/tbp/monty/frameworks/models/monty_base.py
@@ -321,9 +321,6 @@ class MontyBase(Monty):
                         f"finished evaluating after {self.matching_steps} steps"
                     )
 
-    def _post_step(self):
-        pass
-
     ###
     # Methods (other than step) that interact with the experiment
     ###


ramyamounir (approved):
Nice!

vkakerbeck:
I'm happy to remove this since it is unused. Just wondering if it would be a more general solution to keep the `post_step()` call and instead move `_set_step_type_and_check_if_done` inside that function.

scottcanoe (approved):
LGTM.

**suggestion [non-blocking]**: Update `_set_step_type_and_check_if_done`'s docstring to say that it updates counters (like `_post_step`'s docstring says it does).

**note**: I'm also on board with @vkakerbeck's suggestion about moving the contents of `_set_step_type_and_check_if_done` into `_post_step` since I do like the `pre_step` ... `post_step` pattern.

tristanls:
>note: I'm also on board with @vkakerbeck's suggestion about moving the contents of _set_step_type_and_check_if_done into _post_step since I do like the pre_step ... post_step pattern.

@scottcanoe , there is no `pre_step` ... `post_step` pattern here. Note that `pre_step` is absent.

vkakerbeck:
I think generally we have this `pre_` and `post_` pattern throughout the code, although we don't always make use of it or enforce it. For example we have `pre_epoch`, `pre_episode`, `pre_step`, `post_step`, `post_episode`, `post_epoch` functions for different classes (loggers, Monty, lms,...) that are called at particular points in the code. It's not just MontyExperiment that calls those kind of functions. I generally like this pattern as it makes it clear at what point which things should be happening. To me it is cleaner to have a general `post_step` function that each Monty instance calls (even though some may not execute any code in them) instead of adding specific `_set_step_type_and_check_if_done` function that does a bunch of things (indicated by the `and` in the name and even incrementing counters that is not indicated in the name).

scottcanoe:
> > note: I'm also on board with @vkakerbeck's suggestion about moving the contents of _set_step_type_and_check_if_done into _post_step since I do like the pre_step ... post_step pattern.
> 
> @scottcanoe , there is no `pre_step` ... `post_step` pattern here. Note that `_pre_step` is absent.
> 
> Additionally, the way we have things set up, it is the `MontyExperiment` that calls pre and post step methods (kind of) here:
> 
> ```python
>     def run_episode(self):
>         """Run one episode until model.is_done."""
>         self.pre_episode()
>         for step, observation in enumerate(self.dataloader):
>             self.pre_step(step, observation)
>             self.model.step(observation)
>             self.post_step(step, observation)
>             if self.model.is_done or step >= self.max_steps:
>                 break
>         self.post_episode(step)
> ```
> 
> ... I say "kind of" because it is the `MontyExperiment.pre/post_step`, not the model. Although neither `MontySupervisedObjectPretrainingExperiment` nor `MontyObjectRecognitionExperiment` call any pre/post step methods.
> 
> In general, our code doesn't have a `pre_step`/`post_step` convention. Only `MontyExperiment` calls those methods, and they just do some logging.

I retract my comment. It seemed from afar that the `pre_step` ... `post_step` was universal, but I hadn't checked. I'm going to abstain on the issue of renaming or refactoring of `_set_matching_step_and_check_if_done`.

mf-maymay (approved):

tristanls-tbp:
It's been almost a year. We got some clarity on the framework since then. One aspect is that we want to remove `pre/post` methods from the `Monty` framework.

In light of that and the context we gained in the intervening time, are there still any objections to removing this unused method?

nielsleadholm (approved):
I don't have any issue with this. 

[Edit] managed to hit approve in the wrong tab/PR, hence why my initial comment may not have made sense. Trying to get through my PR backlog too quickly...

vkakerbeck:
I'm fine with this change too. My comment last year was just a suggestion.

PR #272, opened 2025-04-18
refactor!: replace ValueError with TypeError (TRY004)

hlee9212:
TIL Python/ruff prefers TypeError over ValueError when you do `if isinstance(...)`. 😄


tristanls:
**note**: I updated PR name to be `refactor!`, since raising a different error is a public API change.

PR #273, opened 2025-04-18
refactor!: replace generic Exception with specific Errors (TRY002)

hlee9212:
Minor update and mostly it would help / enforce using specific Errors in future code contributions.

### DIFF
--- a/src/tbp/monty/frameworks/environment_utils/transforms.py
+++ b/src/tbp/monty/frameworks/environment_utils/transforms.py
@@ -90,7 +90,7 @@ class AddNoiseToRawDepthImage:
             observation, same as input, with added gaussian noise to depth values.
 
         Raises:
-            NoDepthSensorPresentError: if no depth sensor is present.
+            NoDepthSensorPresent: if no depth sensor is present.
         """
         # loop over sensor modules
         for sm in observation[self.agent_id].keys():
@@ -102,7 +102,7 @@ class AddNoiseToRawDepthImage:
                 )
                 observation[self.agent_id][sm]["depth"] += noise
             else:
-                raise NoDepthSensorPresentError(
+                raise NoDepthSensorPresent(
                     "NO DEPTH SENSOR PRESENT. Don't use this transform"
                 )
         return observation
@@ -143,7 +143,7 @@ class GaussianSmoothing:
             observation, same as input, with smoothed depth values.
 
         Raises:
-            NoDepthSensorPresentError: if no depth sensor is present.
+            NoDepthSensorPresent: if no depth sensor is present.
         """
         # loop over sensor modules
         for sm in observation[self.agent_id].keys():
@@ -155,7 +155,7 @@ class GaussianSmoothing:
                 )
                 observation[self.agent_id][sm]["depth"] = filtered_img
             else:
-                raise NoDepthSensorPresentError(
+                raise NoDepthSensorPresent(
                     "NO DEPTH SENSOR PRESENT. Don't use this transform"
                 )
         return observation
--- a/src/tbp/monty/frameworks/models/evidence_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching.py
@@ -1794,7 +1794,7 @@ class EvidenceGraphLM(GraphLM):
             The evidence update threshold.
 
         Raises:
-            InvalidEvidenceUpdateThresholdError: If `self.evidence_update_threshold` is
+            InvalidEvidenceUpdateThreshold: If `self.evidence_update_threshold` is
                 not in the allowed values
         """
         if type(self.evidence_update_threshold) in [int, float]:
@@ -1821,7 +1821,7 @@ class EvidenceGraphLM(GraphLM):
         elif self.evidence_update_threshold == "all":
             return np.min(self.evidence[graph_id])
         else:
-            raise InvalidEvidenceUpdateThresholdError(
+            raise InvalidEvidenceUpdateThreshold(
                 "evidence_update_threshold not in "
                 "[int, float, '[int]%', 'mean', 'median', 'all', 'x_percent_threshold']"
             )
--- a/tools/github_readme_sync/readme.py
+++ b/tools/github_readme_sync/readme.py
@@ -83,7 +83,7 @@ class ReadMe:
         response = get(f"{PREFIX}/docs/{slug}", {"x-readme-version": self.version})
 
         if not response:
-            raise DocumentNotFoundError(f"Document {slug} not found")
+            raise DocumentNotFound(f"Document {slug} not found")
 
         front_matter = OrderedDict()
         front_matter["title"] = response.get("title")


src/tbp/monty/frameworks/environment_utils/transforms.py
```diff
@@ -102,7 +102,7 @@ def __call__(self, observation, state=None):
                 )
                 observation[self.agent_id][sm]["depth"] += noise
             else:
-                raise Exception("NO DEPTH SENSOR PRESENT. Don't use this transform")
+                raise RuntimeError("NO DEPTH SENSOR PRESENT. Don't use this transform")
```
tristanls, line 105: **issue (non-blocking)**: This should probably become a `raise NoDepthSensorPresent`, where `class NoDepthSensorPresent(RuntimeError)`.

src/tbp/monty/frameworks/environment_utils/transforms.py
```diff
@@ -153,7 +153,7 @@ def __call__(self, observation, state=None):
                 )
                 observation[self.agent_id][sm]["depth"] = filtered_img
             else:
-                raise Exception("NO DEPTH SENSOR PRESENT. Don't use this transform")
+                raise RuntimeError("NO DEPTH SENSOR PRESENT. Don't use this transform")
```
tristanls, line 156: **issue (non-blocking)**: This should probably become a `raise NoDepthSensorPresent`, where `class NoDepthSensorPresent(RuntimeError)`.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1776,7 +1776,7 @@ def _get_evidence_update_threshold(self, graph_id):
         elif self.evidence_update_threshold == "all":
             return np.min(self.evidence[graph_id])
         else:
-            raise Exception(
+            raise ValueError(
```
tristanls, line 1779: **issue (non-blocking)**: This should probably become a `raise InvalidEvidenceUpdateThreshold`, where `class InvalidEvidenceUpdateThreshold(ValueError)`.

tools/github_readme_sync/readme.py
```diff
@@ -83,7 +83,7 @@ def get_doc_by_slug(self, slug: str) -> str:
         response = get(f"{PREFIX}/docs/{slug}", {"x-readme-version": self.version})
 
         if not response:
-            raise Exception(f"Failed to fetch document: {response}")
+            raise ValueError(f"Failed to fetch document: {response}")
```
tristanls, line 86: **issue**: This is not a `ValueError`. We failed to fetch the document. 
```python
    if response.status_code == 404:
        return None
    if response.status_code < 200 or response.status_code >= 300:
        logging.error(f"Failed to get {url} {response.text}")
        return None
```

tristanls (commented):
It looks good for the most part. Most issues are non-blocking, where I'd prefer we have more specific errors. One issue is blocking in that `ValueError` is not appropriate.

tristanls:
I updated this PR to be a breaking change (`refactor!:`) as raising more specific errors is a public API breaking change.

jeremyshoemaker (commented):
I agree with @tristanls comments here. The spirit of that lint is to make specific exceptions so that if we want to handle them we have the option to be specific in what we catch.

To expand a bit on his suggestions, the definition for the exception classes is as simple as:
```python
class NoDepthSensorPresent(RuntimeError):
    pass
```

src/tbp/monty/exceptions.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2025 Thousand Brains Project
+#
```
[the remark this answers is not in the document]
jeremyshoemaker: I've seen it done both ways.
- Exceptions defined where they are mainly used
- Exceptions defined in a centralized place

I think for starters, centralizing it makes sense. It makes it easier to see what all errors Monty can potentially raise. This also seems like a reasonable place for them to live to me.

I believe we have a check for whether the copyright is present or not, so I think you have to include it.

jeremyshoemaker (approved):

src/tbp/monty/exceptions.py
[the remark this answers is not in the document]
jeremyshoemaker: Oh yeah, I asked @tristanls about that the other day, and he said we shouldn't be including it, since we're not Numenta.

src/tbp/monty/exceptions.py
tristanls, line 1: **issue**: Please put these in the context of where they occur.

src/tbp/monty/exceptions.py
```diff
@@ -0,0 +1,19 @@
...
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+
+
+class NoDepthSensorPresentError(RuntimeError):
```
tristanls, line 11: **issue**: Move to `src/tbp/monty/frameworks/environment_utils/transforms.py`

src/tbp/monty/exceptions.py
```diff
@@ -0,0 +1,19 @@
...
+    """Raised when a depth sensor is expected but not found."""
+
+    pass
+
+class InvalidEvidenceUpdateThresholdError(ValueError):
```
tristanls, line 16: **issue**: Move to `src/tbp/monty/frameworks/models/evidence_matching.py`

tools/github_readme_sync/readme.py
```diff
@@ -82,8 +82,11 @@ def get_category_docs(self, category: Any) -> List[Any]:
     def get_doc_by_slug(self, slug: str) -> str:
         response = get(f"{PREFIX}/docs/{slug}", {"x-readme-version": self.version})
 
-        if not response:
-            raise Exception(f"Failed to fetch document: {response}")
+        if response.status_code == 404:
+            return None
+        if response.status_code < 200 or response.status_code >= 300:
+            logging.error(f"Failed to get {response.url} {response.text}")
+            return None
```
tristanls, line 89: **issue**: As far as I understand, not finding a document by slug in this case is an exceptional circumstance that should not occur. Returning `None` is not OK here.

**suggestion**: Create `DocumentNotFound(RuntimeError)`.

tristanls (changes requested):

src/tbp/monty/exceptions.py
jeremyshoemaker: I argued the other way in [my comment here](https://github.com/thousandbrainsproject/tbp.monty/pull/273#discussion_r2060694703) but I'm fine with either approach.

src/tbp/monty/exceptions.py
tristanls: **issue**: Also, please rename to `NoDepthSensorPresent`.

src/tbp/monty/exceptions.py
tristanls: **issue**: Also, please rename to `InvalidEvidenceUpdateThreshold`.

src/tbp/monty/exceptions.py
jeremyshoemaker: That's a good point. I've seen it both ways, and I'm just used to going with whatever already exists, but I tend to also prefer not to put what is essentially the type into the name of a class or variable.

tristanls: @hlee9212, please add the following ignore rule for error naming: `"N818", # Exception name should be named with an Error suffix`. Ruff is wrong in this case 🙂.

hlee9212: :) Thanks @tristanls. Updated.

tristanls (changes requested):
**issue**: Aside from the class declaration, all the docstrings and use of the errors still have the `*Error` suffix.

tristanls (approved):
🙌

PR #274, opened 2025-04-21
refactor: enable most comprehension lints

jeremyshoemaker:
This change enables all of the comprehension lints but configures ruff to ignore `dict()` calls that have keyword arguments, because our configs use `dict()` all the time instead of dictionary literals.

### DIFF
--- a/src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
+++ b/src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
@@ -112,7 +112,7 @@ class EnvInitArgsMontyWorldMultiObjectScenes:
 @dataclass
 class OmniglotDatasetArgs:
     env_init_func: Callable = field(default=OmniglotEnvironment)
-    env_init_args: Dict = field(default_factory=lambda: dict())
+    env_init_args: Dict = field(default_factory=lambda: {})
     transform: Union[Callable, list, None] = None
 
     def __post_init__(self):
@@ -143,7 +143,7 @@ class WorldImageDatasetArgs:
 @dataclass
 class WorldImageFromStreamDatasetArgs:
     env_init_func: Callable = field(default=SaccadeOnImageFromStreamEnvironment)
-    env_init_args: Dict = field(default_factory=lambda: dict())
+    env_init_args: Dict = field(default_factory=lambda: {})
     transform: Union[Callable, list, None] = None
 
     def __post_init__(self):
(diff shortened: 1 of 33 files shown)


src/tbp/monty/frameworks/experiments/profile.py
```diff
@@ -64,7 +64,7 @@ def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
         if cls.__bases__[0] is not ProfileExperimentMixin:
             raise TypeError("ProfileExperimentMixin must be leftmost base class.")
-        if not any([issubclass(b, MontyExperiment) for b in cls.__bases__]):
+        if not any(issubclass(b, MontyExperiment) for b in cls.__bases__):
```
jeremyshoemaker, line 67: I totally forgot about this when I wrote this code. The list brackets aren't needed because the `x for x in items` bit is actually making a [generator](https://docs.python.org/3.8/reference/expressions.html#generator-expressions). In a single argument function that expects an iterable, that just works. So we don't have to create a list.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -112,7 +112,7 @@ class EnvInitArgsMontyWorldMultiObjectScenes:
 @dataclass
 class OmniglotDatasetArgs:
     env_init_func: Callable = field(default=OmniglotEnvironment)
-    env_init_args: Dict = field(default_factory=lambda: dict())
+    env_init_args: Dict = field(default_factory=lambda: {})
```
jeremyshoemaker, line 115: Note to self: the change here really should be to remove the `lambda` and use the collection function directly.

jeremyshoemaker: Actually, I take it back, we can't do that, because the factory has to create a **new** dictionary each time. I'm pretty sure it's the same gotcha that happens if you put a list or dictionary literal as the default argument to a function or method. It gets shared between all calls of the function as a single object.

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -60,7 +60,7 @@ def report_episode(self, data, output_dir, episode, mode="train", **kwargs):
         for handler in self.wandb_handlers:
             handler.report_episode(data, output_dir, episode, mode=mode, **kwargs)
 
-        wandb.log(dict())  # TODO: What is this for?
+        wandb.log({})  # TODO: What is this for?
```
tristanls, line 63: **note**: 🤣 "What is this for?"

tristanls (approved):
Let there be `{}`.

PR #275, opened 2025-04-21
refactor!: GetGoodView positioning procedure

tristanls:
fyi @scottcanoe @hlee9212 @ramyamounir 

> [!NOTE]
> When reviewing, I highly recommend the "hide whitespace" view option for looking through the test changes
> Screenshot 2025-04-23 at 14 45 40

This pull request extracts the `GetGoodView` positioning procedure (as a policy) from the `InformedEnvironmentDataLoader`. 

Note that this PR _only_ extracts the positioning procedure as a policy. It does not separate positioning procedures from policies. This is future work. This requires that `GetGoodView` take in unused arguments on initialization and invoke its `positioning_call` method. This awkwardness will be removed when positioning procedures are made distinct from policies.

This code from updated `InformedEnvironemntDataLoader.get_good_view` demonstrates how dataloaders will use positioning procedures in general:
```python
_configured_policy = self.motor_system._policy

self.motor_system._policy = GetGoodView(...)
result = self.motor_system._policy.positioning_call(
    self._observation, self.motor_system._state
)
while not result.terminated and not result.truncated:
    for action in result.actions:
        self._observation, proprio_state = self.dataset[action]
        self.motor_system._state = (
            MotorSystemState(proprio_state) if proprio_state else None
        )

    result = self.motor_system._policy.positioning_call(
        self._observation, self.motor_system._state
    )

self.motor_system._policy = _configured_policy
```
While the above code works within the current procedures-are-policies constraint, in the future, this will look like:
```python
procedure = GetGoodView(...)
result = procedure(observation, proprioceptive_state)
while not result.terminated and not result.truncated:
  for action in result.actions:
    observation, proprioceptive_state = self.dataset[action]
  result = procedure(observation, proprioceptive_state)
```
The goal is that at the end of any number of positioning procedures, the data loader will return `(observation, proprioceptive_state)`, which will be Monty's initial state.

It is worth highlighting that `positioning_call` takes in different arguments than `dynamic|predefined_call` policy methods. This further reinforces the idea that positioning procedures differ from motor policies.

As another highlight, the essence of `GetGoodView` is now captured in `GetGoodView.positioning_call`:
```python
    def positioning_call(
        self,
        observation: Mapping,
        state: Optional[MotorSystemState] = None,
    ) -> PositioningProcedureResult:
        if self._multiple_objects_present:
            on_target_object = self.is_on_target_object(observation)
            if not on_target_object:
                return PositioningProcedureResult(
                    actions=self.orient_to_object(observation, state)
                )

        if self._allow_translation:
            action = self.move_close_enough(observation)
            if action is not None:
                logging.debug("Moving closer to object.")
                return PositioningProcedureResult(actions=[action])

        on_target_object = self.is_on_target_object(observation)
        if (
            not on_target_object
            and self._num_orientation_attempts < self._max_orientation_attempts
        ):
            self._num_orientation_attempts += 1
            return PositioningProcedureResult(
                actions=self.orient_to_object(observation, state)
            )

        if on_target_object:
            return PositioningProcedureResult(success=True, terminated=True)
        else:
            return PositioningProcedureResult(truncated=True)
```
Notice that we `return` in multiple places. `GetGoodView` effectively functions as a state machine that we call repeatedly until done. This is essentially what data loaders are currently doing, but instead of `return`, the data loaders apply the outputs to the environment, for example:
```python
                actions = self.motor_system._policy.orient_to_object(
                    self._observation,
                    sensor_id,
                    target_semantic_id=self.primary_target["semantic_id"],
                    multiple_objects_present=multiple_objects_present,
                    state=self.motor_system._state,
                )
                for action in actions:
                    self._observation, proprio_state = self.dataset[action]
                    self.motor_system._state = (
                        MotorSystemState(proprio_state) if proprio_state else None
                    )
```
At a high level, both code paths do the same thing, but the action actuation is _outside_ the positioning procedure instead of within it.

## Proving that it works

A feature flag `use_get_good_view_positioning_procedure` is added to the `InformedEnvironmentDataLoader` so that we can run the old and new code paths and verify that there is no difference.

Tests are updated so that they execute both code paths.

> [!NOTE]
> While we have both code paths, tests take twice as long, since we run most of them twice.

We have two paths forward:
1. Merge this PR as is, followed by another PR to remove the feature flag and the old get_good_view code.
2. Remove feature flag and get_good_view code as part of this PR, all at once.

I chose path 1 by default as it allows us to run some benchmarks as long as we want by flipping the feature flag back and forth and ensuring everything works as intended. Path 2 goes all-in without verifying that all benchmarks work as intended.

## Existing bug?

Notice that in `GetGoodView.positioning_call`, we do not limit the number of orientation attempts when we have multiple objects or when we allow translation. This is copied as the current behavior. However, that could go into an infinite loop in a pathological case.

### DIFF
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -650,6 +650,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
 
         Returns:
             Whether the sensor is on the object.
+
         """
         self.get_good_view("view_finder")
         for patch_id in ("patch", "patch_0"):
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -590,8 +590,8 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
 
             self.motor_system._policy = GetGoodView(
                 agent_id=self.motor_system._policy.agent_id,
-                desired_object_distance=0.03,
-                good_view_percentage=0.5,
+                desired_object_distance=_configured_policy.desired_object_distance,
+                good_view_percentage=_configured_policy.good_view_percentage,
                 multiple_objects_present=self.num_distractors > 0,
                 sensor_id=sensor_id,
                 target_semantic_id=self.primary_target["semantic_id"],
@@ -600,7 +600,14 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                 # TODO: Remaining arguments are unused but required by BasePolicy.
                 #       These will be removed when PositioningProcedure is split from
                 #       BasePolicy
-                rng=self.rng,
+                #
+                # Note that if we use rng=self.rng below, then the following test will
+                # fail:
+                #   tests/unit/evidence_lm_test.py::EvidenceLMTest::test_two_lm_heterarchy_experiment  # noqa: E501
+                # The test result seems to be coupled to the random seed and the
+                # specific sequence of rng calls (rng is called once on GetGoodView
+                # initialization).
+                rng=np.random.RandomState(),
                 action_sampler_args=dict(actions=[LookUp]),
                 action_sampler_class=UniformlyDistributedSampler,
                 switch_frequency=0.0,
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -142,20 +142,14 @@ class EnvironmentDataLoader:
         self.dataset = dataset
         self.motor_system = motor_system
         self.rng = rng
-        self._observation, proprioceptive_state = self.dataset.reset()
-        self.motor_system._state = (
-            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
-        )
+        self._observation, self.motor_system._state = self.dataset.reset()
         self._action = None
         self._amount = None
         self._counter = 0
 
     def __iter__(self):
         # Reset the environment before iterating
-        self._observation, proprioceptive_state = self.dataset.reset()
-        self.motor_system._state = (
-            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
-        )
+        self._observation, self.motor_system._state = self.dataset.reset()
         self._action = None
         self._amount = None
         self._counter = 0
@@ -169,10 +163,7 @@ class EnvironmentDataLoader:
         else:
             action = self.motor_system()
             self._action = action
-            self._observation, proprioceptive_state = self.dataset[action]
-            self.motor_system._state = (
-                MotorSystemState(proprioceptive_state) if proprioceptive_state else None
-            )
+            self._observation, self.motor_system._state = self.dataset[action]
             self._counter += 1
             return self._observation
 
@@ -472,7 +463,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                 self._action = self.motor_system._policy.touch_object(
                     self._observation,
                     view_sensor_id="view_finder",
-                    state=self.motor_system._policy.state,
+                    state=self.motor_system._state,
                 )
 
             self._observation, proprioceptive_state = self.dataset[self._action]
@@ -596,10 +587,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                     state=self.motor_system._state,
                 )
                 for action in actions:
-                    self._observation, proprio_state = self.dataset[action]
-                    self.motor_system._state = (
-                        MotorSystemState(proprio_state) if proprio_state else None
-                    )
+                    self._observation, self.motor_system._state = self.dataset[action]
 
         if allow_translation:
             # Move closer to the object, if not already close enough
@@ -612,10 +600,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
             # Continue moving to a close distance to the object
             while not close_enough:
                 logging.debug("moving closer!")
-                self._observation, proprio_state = self.dataset[action]
-                self.motor_system._state = (
-                    MotorSystemState(proprio_state) if proprio_state else None
-                )
+                self._observation, self.motor_system._state = self.dataset[action]
                 action, close_enough = self.motor_system._policy.move_close_enough(
                     self._observation,
                     sensor_id,
@@ -639,10 +624,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                 state=self.motor_system._state,
             )
             for action in actions:
-                self._observation, proprio_state = self.dataset[action]
-                self.motor_system._state = (
-                    MotorSystemState(proprio_state) if proprio_state else None
-                )
+                self._observation, self.motor_system._state = self.dataset[action]
             on_target_object = self.motor_system._policy.is_on_target_object(
                 self._observation,
                 sensor_id,
@@ -668,7 +650,6 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
 
         Returns:
             Whether the sensor is on the object.
-
         """
         self.get_good_view("view_finder")
         for patch_id in ("patch", "patch_0"):
@@ -735,10 +716,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
             rotation_quat=quaternion.one,
         )
         _, _ = self.dataset[set_agent_pose]
-        self._observation, proprioceptive_state = self.dataset[set_sensor_rotation]
-        self.motor_system._state = (
-            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
-        )
+        self._observation, self.motor_system._state = self.dataset[set_sensor_rotation]
 
         # Check depth-at-center to see if the object is in front of us
         # As for methods such as touch_object, we use the view-finder
@@ -891,10 +869,7 @@ class OmniglotDataLoader(EnvironmentDataLoaderPerObject):
             )
         self.dataset = dataset
         self.motor_system = motor_system
-        self._observation, proprioceptive_state = self.dataset.reset()
-        self.motor_system._state = (
-            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
-        )
+        self._observation, self.motor_system._state = self.dataset.reset()
         self._action = None
         self._amount = None
         self._counter = 0
@@ -983,10 +958,7 @@ class SaccadeOnImageDataLoader(EnvironmentDataLoaderPerObject):
             )
         self.dataset = dataset
         self.motor_system = motor_system
-        self._observation, proprioceptive_state = self.dataset.reset()
-        self.motor_system._state = (
-            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
-        )
+        self._observation, self.motor_system._state = self.dataset.reset()
         self._action = None
         self._amount = None
         self._counter = 0
@@ -1079,10 +1051,7 @@ class SaccadeOnImageFromStreamDataLoader(SaccadeOnImageDataLoader):
         # TODO: call super init instead of duplication code & generally clean up more
         self.dataset = dataset
         self.motor_system = motor_system
-        self._observation, proprioceptive_state = self.dataset.reset()
-        self.motor_system._state = (
-            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
-        )
+        self._observation, self.motor_system._state = self.dataset.reset()
         self._action = None
         self._amount = None
         self._counter = 0
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -48,8 +48,6 @@ __all__ = [
     "SaccadeOnImageFromStreamDataLoader",
 ]
 
-USE_GET_GOOD_VIEW_POSITIONING_PROCEDURE = True
-
 
 class EnvironmentDataset(Dataset):
     """Wraps an embodied environment with a :class:`torch.utils.data.Dataset`.
@@ -439,8 +437,18 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
     iv) Supports hypothesis-testing "jump" policy
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(
+        self, use_get_good_view_positioning_procedure: bool = False, *args, **kwargs
+    ):
         super(InformedEnvironmentDataLoader, self).__init__(*args, **kwargs)
+        self._use_get_good_view_positioning_procedure = (
+            use_get_good_view_positioning_procedure
+        )
+        """Feature flag to use the GetGoodView positioning procedure.
+
+        This is a temporary feature flag to allow for testing the GetGoodView
+        positioning procedure.
+        """
 
     def __iter__(self):
         # Overwrite original because we don't want to reset agent at this stage
@@ -585,7 +593,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
         TODO M : move most of this to the motor systems, shouldn't be in embodied_data
             class
         """
-        if USE_GET_GOOD_VIEW_POSITIONING_PROCEDURE:
+        if self._use_get_good_view_positioning_procedure:
             _configured_policy = self.motor_system._policy
 
             self.motor_system._policy = GetGoodView(
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -472,7 +472,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                 self._action = self.motor_system._policy.touch_object(
                     self._observation,
                     view_sensor_id="view_finder",
-                    state=self.motor_system._state,
+                    state=self.motor_system._policy.state,
                 )
 
             self._observation, proprioceptive_state = self.dataset[self._action]
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -143,7 +143,7 @@ class EnvironmentDataLoader:
         self.motor_system = motor_system
         self.rng = rng
         self._observation, state = self.dataset.reset()
-        self.motor_system._state = MotorSystemState(state)
+        self.motor_system._state = MotorSystemState(state) if state else None
         self._action = None
         self._amount = None
         self._counter = 0
@@ -151,7 +151,7 @@ class EnvironmentDataLoader:
     def __iter__(self):
         # Reset the environment before iterating
         self._observation, state = self.dataset.reset()
-        self.motor_system._state = MotorSystemState(state)
+        self.motor_system._state = MotorSystemState(state) if state else None
         self._action = None
         self._amount = None
         self._counter = 0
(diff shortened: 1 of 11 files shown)


tests/unit/graph_building_test.py
```diff
@@ -74,14 +78,14 @@ def setUp(self):
...
-            train_dataloader_args=EnvironmentDataLoaderPerObjectTrainArgs(
+            train_dataloader_args=InformedEnvironmentDataLoaderTrainArgs(
                 object_names=["capsule3DSolid", "cubeSolid"],
                 object_init_sampler=PredefinedObjectInitializer(
                     rotations=self.habitat_learned_rotations
                 ),
             ),
             eval_dataloader_class=ED.InformedEnvironmentDataLoader,
-            eval_dataloader_args=EnvironmentDataLoaderPerObjectTrainArgs(
+            eval_dataloader_args=InformedEnvironmentDataLoaderEvalArgs(
```
tristanls, line 88: **note**: I think it was a copy/paste bug when this was previously `EnvironmentDataLoaderPerObjectTrainArgs`.

tests/unit/graph_building_test.py
```diff
@@ -111,14 +115,14 @@ def setUp(self):
...
```
tristanls, line 125: **note**: I think it was a copy/paste bug when this was previously `EnvironmentDataLoaderPerObjectTrainArgs`.

scottcanoe (commented):
**praise**: This looks great, and its exciting to see the new motor policy/data loader design come into being in a really concrete way. Looking forward to working with it.

**thought**: I wonder if we'll ever need to pass processed observations to `GetGoodView` or an initialization policy more generally. For example, a semantic map used by `GetGoodView` is currently computed by `DepthTo3DLocations`, meaning the transform is doing sensor processing that probably ought to be done by a sensor module in the future. Nothing we couldn't figure out of course, but its food for **thought**.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -576,6 +591,51 @@ def get_good_view(
...
+                target_semantic_id=self.primary_target["semantic_id"],
+                allow_translation=allow_translation,
+                max_orientation_attempts=max_orientation_attempts,
+                # TODO: Remaining arguments are unused but required by BasePolicy.
+                #       These will be removed when PositioningProcedure is split from
+                #       BasePolicy
+                #
+                # Note that if we use rng=self.rng below, then the following test will
+                # fail:
+                #   tests/unit/evidence_lm_test.py::EvidenceLMTest::test_two_lm_heterarchy_experiment  # noqa: E501
```
jeremyshoemaker, line 612: We could just leave out the file directory parts `tests/unit` or even the whole file path and just reference the class and test method, and then we wouldn't need the `noqa` comment.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -576,6 +591,51 @@ def get_good_view(
...
+                # initialization).
+                rng=np.random.RandomState(),
+                action_sampler_args=dict(actions=[LookUp]),
+                action_sampler_class=UniformlyDistributedSampler,
+                switch_frequency=0.0,
+            )
+            result = self.motor_system._policy.positioning_call(
+                self._observation, self.motor_system._state
+            )
+            while not result.terminated and not result.truncated:
```
jeremyshoemaker, line 624: We could apply De Morgan's law here. I feel like this reads better, but it could just be me.
```suggestion
            while not (result.terminated or result.truncated):
```

tests/unit/evidence_lm_test.py
```diff
@@ -866,270 +878,294 @@ def test_patch_off_object_logging(self):
...
-            # min_steps is reached and the sensor moves off the object). In the second
-            # episode the sensor moves off the sphere on episode steps 6+
+        for c in [
+            config,
+            create_config_with_get_good_view_positioning_procedure(config),
+        ]:
+            with MontyObjectRecognitionExperiment(c) as exp:
+                pprint("...training...")
+                # First episode will be used to learn object (no_match is triggered before  # noqa: E501
+                # min_steps is reached and the sensor moves off the object). In the second  # noqa: E501
```
jeremyshoemaker, line 916: We could reflow the comment to avoid the length limit.

tests/unit/evidence_lm_test.py
```diff
@@ -866,270 +878,294 @@ def test_patch_off_object_logging(self):
...
+                    len(exp.model.learning_modules[0].get_all_known_object_ids()),
+                    1,
+                    "No new objects should be added to memory after time out.",
+                )
+                self.assertLessEqual(
+                    exp.model.learning_modules[0]
+                    .get_graph("new_object0", input_channel="first")
+                    .x.shape[1],
+                    32,  # max_train_steps + exploratory_steps
+                    "No new points should be added to an existing graph after time out.",  # noqa: E501
```
jeremyshoemaker, line 1039: We could break the string into two parts. I can't remember if you'd need to wrap it in parentheses to do it or not.

This _might_ do it.
```suggestion
                    ("No new points should be added "
                     "to an existing graph after time out."),
```

tests/unit/evidence_lm_test.py
jeremyshoemaker, line 1: I'm not sure how easy this would be to do, but instead of having to change each test to run both versions, we could try subclassing `EvidenceLMTest` and changing just the setup so that the tests themselves don't have to change.

Pytest will see the inherited test methods and run them a second time in the subclass without having to repeat them.

tests/unit/feature_flags.py
```diff
@@ -0,0 +1,57 @@
...
+from typing import Dict
+
+from tbp.monty.frameworks.config_utils.config_args import Dataclass
+from tbp.monty.frameworks.config_utils.make_dataset_configs import (
+    InformedEnvironmentDataLoaderEvalArgs,
+    InformedEnvironmentDataLoaderTrainArgs,
+)
+
+
+def to_dict(maybe_dataclass: Dataclass | Dict) -> dict:
```
jeremyshoemaker, line 23: Do we not have this method somewhere else already? I feel like I've seen it before. Unless you're moving it here from elsewhere and I haven't gotten to that part yet.

jeremyshoemaker (commented):
I think it might be worth looking into the suggest I made below around subclassing and altering the test case setup instead of changing all the tests. 

I've done something similar before on a previous project, but we had a more robust feature flagging system and you could just decorate the test case class to flip the flag for all the code ran inside it.

tristanls:
> **thought**: I wonder if we'll ever need to pass processed observations to `GetGoodView` or an initialization policy more generally. For example, a semantic map used by `GetGoodView` is currently computed by `DepthTo3DLocations`, meaning the transform is doing sensor processing that probably ought to be done by a sensor module in the future. Nothing we couldn't figure out of course, but its food for **thought**.

@scottcanoe, I think we'll want to avoid anything like that. Instead, I would propose that a `PositioningProcedure` is an oracle that can do _anything_. So, there is no reason it can't import a sensor module function and use that as part of whatever it does. On first thought, I would extract common code that computes a bunch of stuff to a library, then use that library in a `SensorModule` and a `PositioningProcedure` separately. Our default should always be: _Thou shalt not access Monty internals from outside Monty_. 

Having said that, I'm thinking only of `PositioningProcedure`s, which in my mental model are the experimenter placing an experimental subject in a specific configuration. However, a _different_ `GetGoodView` is also used after `execute_jump_attempt`. This will probably be a policy and not a positioning procedure. We'll know more when I extract `execute_jump_attempt` into a policy.

tristanls:
I have no idea what happened to all the checks 😕 

Screenshot 2025-04-23 at 17 57 47

tests/unit/evidence_lm_test.py
tristanls: My reasoning here and all the E501 ignores was that I indented everything for the feature flag, and I intend to remove the indents after the feature flag is removed, so I didn't want to futz around with line formatting.

tests/unit/evidence_lm_test.py
tristanls: I really don't want to do this for this PR. I can reach out, and we can explore that more for another PR. But for this PR, there was a lot of throwaway code to write, and I'd rather not rewrite the throwaway code since I already wrote it.

src/tbp/monty/frameworks/environments/embodied_data.py
tristanls: I'm split on this. Right now, there is one for and one against. If someone else prefers this, I'll switch.

tests/unit/feature_flags.py
tristanls: I haven't seen it anywhere. There's a bunch of this in configs, which might make `Dataclass | Dict` look familiar?
```python
    motor_system_args: Union[Dict, Dataclass] = field(
        default_factory=lambda: dict(
            policy_class=InformedPolicy,
            policy_args=make_informed_policy_config(
                action_space_type="distant_agent_no_translation",
                action_sampler_class=ConstantSampler,
                rotation_degrees=5.0,
                file_name=Path(__file__).parent / "resources/fixed_test_actions.jsonl",
            ),
        )
    )
```

tests/unit/evidence_lm_test.py
tristanls: Do you feel my excuse is good enough? 😬

vkakerbeck (approved):
Very nice! Great to see this take shape :)
I just left a few suggestions, none of them blocking. Did you already run a check with this new positioning procedure on some of the benchmark experiments? I think this would be worth doing just to confirm we don't get minor deviations that don't show up in the unit tests.
I'm happy with either option 1 or 2. I assume the code won't stay in this state for long so running those unit tests twice for a few days sounds fine.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -353,6 +356,386 @@ def derive_habitat_goal_state(self):
...
+            idx_loc_to_look_at[0], idx_loc_to_look_at[1], :3
+        ]
+        camera_location = self.get_agent_state(state)["sensors"][
+            f"{self._sensor_id}.depth"
+        ]["position"]
+        agent_location = self.get_agent_state(state)["position"]
+        # Get the location of the object relative to sensor.
+        relative_location = location_to_look_at - (camera_location + agent_location)
+
+        return relative_location
```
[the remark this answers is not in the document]
nielsleadholm: I'm wondering if it's that terrible to call it a relative location? It is a location with respect to the sensor's reference frame. Like any vector that is not the origin, that can be formulated as a displacement, but it's not that it represents a displacement that the agent is going to perform (i.e., the "eye" location is not going to undergo a displacement in 3D space like this). Instead, `compute_look_amounts` takes this location and determines the re-orienting displacement that is necessary. Feeding in a location (in some coordinate system) to `compute_look_amounts` is kind of nice because it's similar to our aim of passing a goal-state (a location) and having a the policy figure out how to orient towards it. 

Tagging @scottcanoe in case you want to add anything given you've spend a fair amount of time digging around these.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -353,6 +356,386 @@ def derive_habitat_goal_state(self):
...
+
+    actions: List[Action] = field(default_factory=list)
+    """Actions to take."""
+    success: bool = False
+    """Whether the procedure succeeded in its positioning goal."""
+    terminated: bool = False
+    """Whether the procedure reached a terminal state with success or failure."""
+    truncated: bool = False
+    """Whether the procedure was truncated due to a limit on the number of attempts or
+    other criteria."""
```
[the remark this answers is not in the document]
nielsleadholm: I agree, I found the positioning of the comments a bit confusing.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -353,6 +356,386 @@ def derive_habitat_goal_state(self):
...
+                    logging.debug("Moving forward")
+                    return MoveForward(agent_id=self.agent_id, distance=0.01)
+            else:
+                logging.debug("Close enough.")
+                return None
+        else:
+            logging.debug("Enough percent visible.")
+            return None
+
+    def orient_to_object(
```
nielsleadholm, line 665: **suggestion (non-blocking):** This can be a future change, but I wonder whether re-ordering these methods would make it easier to understand/read this class. E.g. it might make sense to start with `positioning_call`, then `orient_to_object` (the high-level methods), before the reader digs into the details of the methods they call like `find_location_to_look_at` and `compute_look_amounts`.

tests/unit/feature_flags.py
nielsleadholm: Maybe you were thinking of `config_to_dict` in `src/tbp/monty/frameworks/utils/dataclass_utils.py`? But yeah it's a bit different, otherwise I don't recall anything.

nielsleadholm (approved):
Awesome to see this coming together.

src/tbp/monty/frameworks/models/motor_policies.py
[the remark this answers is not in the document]
tristanls: So, these are docstrings, not comments. They document the dataclass properties. See the `ExampleClass`, specifically `attr5`, here https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html .

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -353,6 +356,386 @@ def derive_habitat_goal_state(self):
...
+        return [
+            LookDown(agent_id=self.agent_id, rotation_degrees=down_amount),
+            TurnLeft(agent_id=self.agent_id, rotation_degrees=left_amount),
+        ]
+
+    def positioning_call(
+        self,
+        observation: Mapping,
+        state: Optional[MotorSystemState] = None,
+    ) -> PositioningProcedureResult:
```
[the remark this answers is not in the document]
tristanls: Oh, I forgot an @override annotation to indicate the docstring is in the parent class. I'll add it. (Hovering over it in IDE brings up parent docstring automatically)

src/tbp/monty/frameworks/models/motor_policies.py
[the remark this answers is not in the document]
vkakerbeck: Yeah I think calling it `relative_location` here is probably enough of a hint. I guess it mostly bothered me in the docstring (like "Takes in a semantic 3D observation and returns an x,y,z location.") where it sounds like this returns a location but then when you get to the end of the function you see that it is actually a displacement. 
I think it is different than goal states since goal states will be locations in a common reference frame (i.e. relative to the body), not relative displacements like here.
But again, it's a minor difference that someone could infer from reading the function, I was just wondering if it could be made more clear so when you call this function you don't expect the x,y,z coordinates to be in the common reference frame.

src/tbp/monty/frameworks/models/motor_policies.py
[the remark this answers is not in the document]
vkakerbeck: okay, I see. I was more referring to a short explanation of the logic inside this specific function (since it is rather complex). Not just the input and output description.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -353,6 +356,386 @@ def derive_habitat_goal_state(self):
...
+            down_amount: Amount to look down (degrees).
+            left_amount: Amount to look left (degrees).
+        """
+        # Get the sensor's rotation relative to the world.
+        agent_state = self.get_agent_state(state)
+        # - The agent's rotation relative to the world.
+        agent_rotation = agent_state["rotation"]
+        # - The sensor's rotation relative to the agent.
+        sensor_rotation = agent_state["sensors"][f"{self._sensor_id}.depth"]["rotation"]
+        # - The sensor's rotation relative to the world.
```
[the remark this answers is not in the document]
tristanls: I don't know. This is copy pasted from the data loader.

tests/unit/evidence_lm_test.py
jeremyshoemaker: I'd have to test it out in a local copy, but I'm pretty sure you could just do something like this:

```python
class EvidenceLMTestWithGetGoodView(EvidenceLMTest):
    def setUp(self):
        super.setUp()
        self.evidence_config = create_config_with_get_good_view_positioning_procedure(self.evidence_config)
        # repeat for all the other configs
```

Then Pytest should find those tests and run them with the altered configs. Unless there's more being changed than just the configs. And then you don't have to make all the rest of the changes in this file (and the other tests could be handled similarly).

src/tbp/monty/frameworks/models/motor_policies.py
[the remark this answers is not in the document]
tristanls: Ah, hmm. I think the explanation you're thinking of is in the docstring for the `GetGoodView(PositioningProcedure)` class. I think that's the right place for it. When selecting a positioning procedure or a policy, we will look at `GetGoodView`, `TouchObject`, or `ExecuteJumpAttempt` to decide what to use. We'll be looking at what _class_ to use, not what _method_ to use. This `positioning_call` method, which will eventually end up as `__call__`, is just a generic "execute the purpose of this class" type of thing, which is reflected in its docstring: "Return a list of actions to position the agent in the scene."

tristanls:
A few `run_parallel.py` benchmark runs on my laptop (not seeing any meaningful differences):

| GetGoodView | Experiment | Correct (%) | Used MLH (%) | Num Match Steps | Rotation Error | WandB |
|---|---|---|---|---|---|---|
| old | base_config_10distinctobj_dist | 99.29 | 0 | 31 | 0.300 | https://wandb.ai/thousand-brains-project/Monty/runs/g7zosqpn |
| new | base_config_10distinctobj_dist | 99.29 | 0 | 31 | 0.299 | https://wandb.ai/thousand-brains-project/Monty/runs/bqiqfpbt |
| old | randrot_noise_10distinctobj_dist_agent | 100 | 0 | 73 | 0.382 | https://wandb.ai/thousand-brains-project/Monty/runs/atqtg2kk |
| new | randrot_noise_10distinctobj_dist_agent | 100 | 0 | 73 | 0.3561 | https://wandb.ai/thousand-brains-project/Monty/runs/s0ok9ijs |
| old | randrot_noise_10distinctobj_5lms_dist_agent | 100 | 0 | 37 | 0.804 | https://wandb.ai/thousand-brains-project/Monty/runs/h6wx9ei5 |
| new | randrot_noise_10distinctobj_5lms_dist_agent | 100 | 0 | 37 | 0.808 | https://wandb.ai/thousand-brains-project/Monty/runs/l7gankah

tristanls:
As an FYI @vkakerbeck , @nielsleadholm , I used @jeremyshoemaker 's approach to feature flag testing. It is much better and easier to apply. You can glance at the overall file changes and see how a test file is now changed in a very constrained way, instead of inside every test.

tests/unit/evidence_lm_test.py
tristanls: Thanks, @jeremyshoemaker. This approach is much better. I implemented what you suggested, and the tests look cleaner for it.

src/tbp/monty/frameworks/models/motor_policies.py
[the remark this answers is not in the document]
vkakerbeck: Ah okay yes, that should cover everything needed from a user perspective. I guess I was thinking mostly from the perspective of someone who plans to update the code who needs to look at this specific method and a docstring might provide useful context for that. But I guess in that case you would want to read and understand the entire function anyways.

vkakerbeck:
I don't think "no meaningful difference" is enough if we want to make this the default. We should make sure that the benchmark results are EXACTLY replicated. There should be no difference at all in the results. If fact, if the new positioning procedure was changing something in its logic I would expect it to only have a minor difference on the benchmark results (because it would probably just affect some edge case episodes and then just slightly change the starting position of the agent).
I guess one alternative cause for the difference could be a different order in which random seed are initialized (I think you mentioned this in one of your comments) but we need to make 100% sure that this is the cause. Otherwise, we might be missing some weird new behaviors.

PR #276, opened 2025-04-23
style: enable ruff format checking

jeremyshoemaker:
We were under the impression that `ruff check` checked for formatting, but that is under a separate subcommand `ruff format`. This change formats the files that have drifted from the formatting we want, and updates the GH checks to make sure we keep it that way going forward.

After this is merged, we'll add the merge commit hash to the `.git-blame-ignore-revs` file to keep the commit history clear in GH and other tools that support that feature.

### DIFF
--- a/src/tbp/monty/frameworks/models/evidence_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching.py
@@ -1078,8 +1078,8 @@ class EvidenceGraphLM(GraphLM):
         assert not np.isnan(np.max(self.evidence[graph_id])), "evidence contains NaN."
         logging.debug(
             f"evidence update for {graph_id} took "
-            f"{np.round(end_time - start_time,2)} seconds."
-            f" New max evidence: {np.round(np.max(self.evidence[graph_id]),3)}"
+            f"{np.round(end_time - start_time, 2)} seconds."
+            f" New max evidence: {np.round(np.max(self.evidence[graph_id]), 3)}"
         )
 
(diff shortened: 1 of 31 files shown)


src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1763,9 +1763,9 @@ def _get_evidence_update_threshold(self, graph_id):
         ) and self.evidence_update_threshold.endswith("%"):
             percentage_str = self.evidence_update_threshold.strip("%")
             percentage = float(percentage_str)
-            assert (
-                percentage >= 0 and percentage <= 100
-            ), "Percentage must be between 0 and 100"
+            assert percentage >= 0 and percentage <= 100, (
```
jeremyshoemaker, line 1766: TIL that this can be rewritten as `0 <= percentage <= 100` and it gives the result you'd expect, with short-circuiting if the first comparison is False. I think I knew that already and forgot about it.

Changing it is outside the scope of this PR.

See https://docs.python.org/3/reference/expressions.html#comparisons

tristanls (approved):
Thanks! I'm honestly surprised this is all there was.

tristanls:
**note**: I think this could be a `style:` commit
> `style:` Change that does not affect the meaning of the code (white-space, formatting, etc.).
> -- https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0010_conventional_commits.md#types

PR #278, opened 2025-04-25
refactor: address yoda condition (SIM300)

hlee9212:
Yoda condition, fixed it is.


jeremyshoemaker (approved):
I'm surprised we had instances of those. I've very rarely seen it in Python.

tristanls (approved):

PR #279, opened 2025-04-25
test: use GPU runner for monty tests

jeremyshoemaker:
Newer versions of HabitatSim seem to require an actual GPU based on the errors we've seen trying to run without it, so we're switching the tests to run on a GPU instance.

### DIFF
--- a/.github/workflows/monty.yml
+++ b/.github/workflows/monty.yml
@@ -305,7 +305,7 @@ jobs:
     name: test-monty
     runs-on:
       group: tbp.monty
-      labels: tbp-linux-x64-ubuntu2204-8core
+      labels: tbp-linux-x64-ubuntu-nvidia-4core-t4gpu
     timeout-minutes: 120
     needs:
       - check_dependencies_monty # Don't run if dependency check fails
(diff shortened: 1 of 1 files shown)


.github/workflows/monty.yml
```diff
@@ -321,33 +321,27 @@ jobs:
...
-        #
-        # NOTE: LD_PRELOAD is needed to provide what is discussed in https://stackoverflow.com/questions/71010343/cannot-load-swrast-and-iris-drivers-in-fedora-35/72200748#72200748
         if: ${{ needs.should_run_monty.outputs.should_run_monty == 'true' }}
         working-directory: tbp.monty
         run: |
           export PATH="$HOME/miniconda/bin:$PATH"
           source activate tbp.monty
           set -e
           mkdir -p test_results/pytest
-          export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libstdc++.so.6
```
jeremyshoemaker, line 349: [I tested this](https://github.com/thousandbrainsproject/tbp.monty/actions/runs/14670677100/job/41177141925) to see if we still need it, and it doesn't seem to be necessary, so I'm removing it.

.github/workflows/monty.yml
```diff
@@ -321,33 +321,27 @@ jobs:
...
-          xvfb-run pytest -n 8 --cov --cov-report html --cov-report term --junitxml=test_results/pytest/results.xml --verbose
+          xvfb-run pytest --cov --cov-report html --cov-report term --junitxml=test_results/pytest/results.xml --verbose
```
jeremyshoemaker, line 344: Rather than change `-n 8` -> `-n 4`, I'm removing it because by default Pytest uses the number of cores available on the host as the number of runners to spin up.

This way, if we change the host again, we don't have to remember to change this value.

codingkoi (dismissed):
Testing outside approval

.github/workflows/monty.yml
```diff
@@ -321,33 +321,27 @@ jobs:
           path: |
             ~/miniconda
           key: ${{ needs.install_monty.outputs.conda_env_cache_key_sha }}
+      - name: Install dependencies
+        # Installing before checkout because GitLFS is missing from the image
+        if: ${{ needs.should_run_monty.outputs.should_run_monty == 'true' }}
+        run: |
+          sudo apt update -y
+          sudo apt install -y --no-install-recommends xvfb libegl1-mesa-dev git-lfs
```
tristanls, line 329: **question (non-blocking)**: Are `xvfb` and `libegl1-mesa-dev` still needed on a GPU instance?

tristanls (approved):
This is awesome to see \o/.

One question on need for `xvfb`, but non-blocking.

.github/workflows/monty.yml
jeremyshoemaker: Looks like they're still both needed. Without both, a few of the tests fail while looking for a GL driver, and with just the Mesa driver, the workers crash probably because they can't access an X server but without extra logging I can't be sure, I'm pretty sure though.

tristanls (approved):
Thanks for double checking everything.

PR #281, opened 2025-04-29
chore: check for pending review requests

jeremyshoemaker:
Whenever someone requests a re-review, we aren't checking for that with the check-assignee-approval check, so this updates the check to do that.

I also realize a lot of the information we're using here is available in the context, so we don't need to make a separate API call to fetch the pull request object.

The new behaviour is:
- check for empty assignees
- check for non-maintainer assignees
- check for pending assignee reviews
- check latest reviews were approvals

### DIFF
--- a/.github/workflows/check_assignee_approval.yml
+++ b/.github/workflows/check_assignee_approval.yml
@@ -1,6 +1,8 @@
 name: Check Assignees Approval
 
 on:
+  pull_request:
+    types: [opened, assigned, review_requested, review_request_removed]
   pull_request_review:
     types: [submitted]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/check_assignee_approval.yml
```diff
@@ -28,25 +30,19 @@ jobs:
...
             
-            console.log(`Checking approvals for PR #${pull_number}`);
+            console.log(`Checking approvals for PR #${pr.number}`);
             
-            const { data: pr } = await github.rest.pulls.get({
-              owner,
-              repo,
-              pull_number,
-            });
+            // 1. Check for no maintainers in assignees
```
tristanls, line 38: **nitpick**: Misleading comment. The code checks for an empty assignees list. A list of `[not_a_maintainer]` would pass the check, but match the comment criteria.

tristanls (dismissed):
Thanks! Looks great.

.github/workflows/check_assignee_approval.yml
jeremyshoemaker: Oh, good catch, that should just say "check for no assignees".

tristanls:
**issue**: This is problematic:

Screenshot 2025-04-29 at 13 38 42

Notice there are two checks, as GitHub appends the trigger name to the check.

This makes it _impossible_ for us to require this check, defeating the point.

https://github.com/orgs/community/discussions/141018

PR #283, opened 2025-04-29
fix: GetGoodView positioning procedure matches previous code

tristanls:
This pull request corrects the GetGoodView positioning procedure to operate exactly like the data loader's current `get_good_view()` code.

## Benchmarks

Here are a few `run_parallel.py` benchmarks from my laptop with old and new code paths. They are for comparison with https://github.com/thousandbrainsproject/tbp.monty/pull/275#issuecomment-2828993390, which had some discrepancies. I now observe no discrepancies.

| GetGoodView | Experiment | Correct (%) | Used MLH (%) | Num Match Steps | Rotation Error | WandB |
|---|---|---|---|---|---|---|
| old | base_config_10distinctobj_dist_agent | 99.28571428571428 | 0 | 35.49285714285714 | 0.30008273381294964 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/l3isif50/overview)
| new | base_config_10distinctobj_dist_agent | 99.28571428571428 | 0 | 35.49285714285714 | 0.30008273381294964 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/thd7ok2b/overview)
| old | randrot_noise_10distinctobj_dist_agent | 100 | 3 | 53.04 | 0.3820000000000001 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/wodxhhmm/overview)
| new | randrot_noise_10distinctobj_dist_agent | 100 | 3 | 53.04 | 0.3820000000000001 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/ypvy54zo/overview)
| old | randrot_noise_10distinctobj_5lms_dist_agent | 100 | 0 | 54.73 | 0.80406485 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/g5cim14p/overview)
| new | randrot_noise_10distinctobj_5lms_dist_agent | 100 | 0 | 54.73 | 0.80406485 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/8iyeat6o/overview)
| old | base_77obj_dist_agent | 93.5064935064935 | 11.255411255411255 | 99.57142857142856 | 0.29177499999999995 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/t105xwl2/overview)
| new | base_77obj_dist_agent | 93.5064935064935 | 11.255411255411255 | 99.57142857142856 | 0.29177499999999995 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/yw56w04a/overview)
| old | base_77obj_surf_agent | 98.7012987012987 | 6.926406926406926 | 53.75757575757576 | 0.17195614035087717 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/iow7d4f7/overview)
| new | base_77obj_surf_agent | 98.7012987012987 | 6.926406926406926 | 53.75757575757576 | 0.17195614035087717 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/v29pgvvc/overview)

### DIFF
--- a/src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
+++ b/src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
@@ -312,6 +312,11 @@ class EnvironmentDataloaderPerObjectArgs:
 
 
 @dataclass
+class InformedEnvironmentDataLoaderPerObjectArgs(EnvironmentDataloaderPerObjectArgs):
+    use_get_good_view_positioning_procedure: bool = False
+
+
+@dataclass
 class EnvironmentDataLoaderPerObjectTrainArgs(EnvironmentDataloaderPerObjectArgs):
     object_names: List = field(default_factory=lambda: DefaultTrainObjectList().objects)
     object_init_sampler: Callable = field(default_factory=DefaultObjectInitializer)
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -592,12 +592,10 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
             class
         """
         if self._use_get_good_view_positioning_procedure:
-            _configured_policy = self.motor_system._policy
-
-            self.motor_system._policy = GetGoodView(
+            positioning_procedure = GetGoodView(
                 agent_id=self.motor_system._policy.agent_id,
-                desired_object_distance=_configured_policy.desired_object_distance,
-                good_view_percentage=_configured_policy.good_view_percentage,
+                desired_object_distance=self.motor_system._policy.desired_object_distance,
+                good_view_percentage=self.motor_system._policy.good_view_percentage,
                 multiple_objects_present=self.num_distractors > 0,
                 sensor_id=sensor_id,
                 target_semantic_id=self.primary_target["semantic_id"],
@@ -618,7 +616,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                 action_sampler_class=UniformlyDistributedSampler,
                 switch_frequency=0.0,
             )
-            result = self.motor_system._policy.positioning_call(
+            result = positioning_procedure.positioning_call(
                 self._observation, self.motor_system._state
             )
             while not result.terminated and not result.truncated:
@@ -628,11 +626,9 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                         MotorSystemState(proprio_state) if proprio_state else None
                     )
 
-                result = self.motor_system._policy.positioning_call(
+                result = positioning_procedure.positioning_call(
                     self._observation, self.motor_system._state
                 )
-
-            self.motor_system._policy = _configured_policy
 
             return result.success
 
--- a/src/tbp/monty/frameworks/models/motor_policies.py
+++ b/src/tbp/monty/frameworks/models/motor_policies.py
@@ -716,6 +716,8 @@ class GetGoodView(PositioningProcedure):
             self._multiple_objects_present
             and not self._executed_multiple_objects_orientation
         ):
+            # Current implementation only orients to the object once in the case
+            # where there are multiple objects.
             self._executed_multiple_objects_orientation = True
             on_target_object = self.is_on_target_object(observation)
             if not on_target_object:
@@ -729,6 +731,8 @@ class GetGoodView(PositioningProcedure):
                 logging.debug("Moving closer to object.")
                 return PositioningProcedureResult(actions=[action])
 
+        # If we get here, the current implementation will not attempt to move
+        # closer to the object again.
         self._allow_translation = False
 
         on_target_object = self.is_on_target_object(observation)
--- a/src/tbp/monty/frameworks/models/motor_policies.py
+++ b/src/tbp/monty/frameworks/models/motor_policies.py
@@ -724,6 +724,8 @@ class GetGoodView(PositioningProcedure):
                 logging.debug("Moving closer to object.")
                 return PositioningProcedureResult(actions=[action])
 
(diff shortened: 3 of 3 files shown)


src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -592,12 +592,10 @@ def get_good_view(
             class
         """
         if self._use_get_good_view_positioning_procedure:
-            _configured_policy = self.motor_system._policy
-
-            self.motor_system._policy = GetGoodView(
+            positioning_procedure = GetGoodView(
```
tristanls, line 595: **note**: We no longer need to swap anything to execute a positioning procedure, which is encouraging. There is no functional difference here. I learned we can do this while debugging the discrepancies.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -311,6 +311,11 @@ class EnvironmentDataloaderPerObjectArgs:
     object_init_sampler: Callable
 
 
+@dataclass
+class InformedEnvironmentDataLoaderPerObjectArgs(EnvironmentDataloaderPerObjectArgs):
```
tristanls, line 315: **note**: Needed to run `base_config_10distinctobj_dist_agent` benchmark.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -724,6 +724,8 @@ def positioning_call(
                 logging.debug("Moving closer to object.")
                 return PositioningProcedureResult(actions=[action])
 
+        self._allow_translation = False
```
tristanls, line 727: **note**: state machines... _sigh_

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -711,7 +712,11 @@ def positioning_call(
         observation: Mapping,
         state: Optional[MotorSystemState] = None,
     ) -> PositioningProcedureResult:
-        if self._multiple_objects_present:
+        if (
+            self._multiple_objects_present
+            and not self._executed_multiple_objects_orientation
+        ):
+            self._executed_multiple_objects_orientation = True
```
tristanls, line 719: **note**: Ensure we only execute multiple objects orientation at most once.

jeremyshoemaker (approved):

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
vkakerbeck: Why was this nescessary? Wouldn't it be False by default?

src/tbp/monty/frameworks/models/motor_policies.py
vkakerbeck: suggestion: could you add this as a comment in the code instead of on the PR review? (same with the comment below for the other flag)

vkakerbeck (approved):
Nice! Thanks for doing this deep dive and figuring out the issue :)

nielsleadholm (approved):
Nicely done, these can be such a pain to debug!

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
tristanls: Yes, it would be False by default. I needed this to set it to True for testing.

src/tbp/monty/frameworks/models/motor_policies.py
tristanls: I worry this will make it sound as if this is the _final_/_right_/_correct_ thing to do for the positioning procedure. These flags ensure that the code behaves exactly as the existing `get_good_view`. However, nothing is dictating that this ought always to be the case. For example, we could reorient multiple times. Or, we could move closer after reorienting. Those seem like different places in the solution space.

I added some comments that try to explain the code without entrenching the existing algorithm.

jeremyshoemaker: I don't want to start bike-shedding the comments, but I feel like this one and the other one don't quite explain _why_ we're setting these flags the way we are. Maybe it's a lack of understanding of what this flag does on my part that isn't helping.

I'd feel better if the comment was more along the lines of "We're setting this flag to true because ...". As it currently is, it's not clear if the current implementation is describing what this flag does, or if we're setting this flag to alter the current implementation.

src/tbp/monty/frameworks/models/motor_policies.py
jeremyshoemaker: Reading this one again, it sounds like "the current implementation" is what these flags are doing. In that case, it seems like this could just say something like "Set this to false so we don't attempt to move closer to the object again".

src/tbp/monty/frameworks/models/motor_policies.py
tristanls: This is why I left it without comments. The reason is "so that the code does what it does", as we ensure the code execution is identical to code that will be deleted soon.

Do you feel the following would be helpful:
```python
# This flag ensures that this state machine state is not revisited.
```

src/tbp/monty/frameworks/models/motor_policies.py
tristanls: That's not at all what these are for. These are set this way so that the algorithm executes the correct state machine that matches what the `get_good_view` algorithm does.

src/tbp/monty/frameworks/models/motor_policies.py
tristanls: I'm going to proceed without any comments.

I think a better solution for this is to use a state machine library that makes it evident that this is a state machine being executed.

tristanls: Ok, I added the explicit state machine comments.

jeremyshoemaker (approved):

PR #288, opened 2025-05-01
fix: non-informed Initial possible poses loop order

ramyamounir:
## Problem
While looking into the `_get_initial_hypothesis_space` of `EvidenceGraphLM`, I noticed that the order by which we initialize non-informed hypotheses differs from informed hypotheses. My understanding is that we store the evidence for all locations and then we tile this array for additional rotations. When using uniform (or given) initial poses, we do it the other way around which does not match how we store the initial evidence.

*Note that this looked like a bug to me so I'm labeling it as such, but I could be wrong and there might be something I'm missing here. Either way, I hope to get some feedback here because I'm working on this function for hypotheses resampling.*

Drawing 2025-05-01 08 03 00 excalidraw

## Solution
I just changed the order of the loop so that it would append the node_id first and then stack the rotations. This would match how we store the initial evidence scores.

## Impact
I only ran the benchmark experiment `randrot_noise_10distinctobj_dist_agent` with `uniform` initial poses. The overall accuracy increased from 91 to 96. We rarely use non-informed initial poses so I assume this fix may not have any impact on our workflow or the DMC paper (I hope). Also this is only for the initial evidence set, Monty will overcome this initial setback in subsequent evidence updates anyway.

### DIFF
--- a/src/tbp/monty/frameworks/models/evidence_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching.py
@@ -802,8 +802,8 @@ class EvidenceGraphLM(GraphLM):
                 graph_id, input_channel
             )
             # Initialize fixed possible poses (without using pose features)
-            for node_id in range(len(all_channel_locations)):
-                for rotation in self.initial_possible_poses:
+            for rotation in self.initial_possible_poses:
+                for node_id in range(len(all_channel_locations)):
                     initial_possible_channel_locations.append(
                         all_channel_locations[node_id]
                     )


tristanls:
I'm removing myself from review. I don't understand any of this code, and I hope others are familiar with it so they can review it meaningfully.

nielsleadholm (commented):
Nice job catching this! Based on my understanding and the empirical result, it does look like this could be a bug. That said, I would have maybe expected it to be more catastrophic if it was totally messing with the hypotheses.

**Re. notation describing the issue**
That is really helpful with the included notation diagram, although just to make sure we are all on the same page, I think it could be useful to be more specific about the rotations that are stored (maybe with the updated annotation from the paper 😅).

For Informed Initial Hypotheses
- We begin by determining all the possible values of $\prescript{B}{S}{\mathbf{R}}$. There will normally be either 2 (pose fully defined, but could be mirrored) or 8 (pose not fully defined, so we sample basically around a circle) of such rotations defined, depending on the nature of the first sensation.
- We then iterate through these values for $\prescript{B}{S}{\mathbf{R}}$. For each one, we determine the rotation of the object ($\prescript{B}{M}{\mathbf{R}}$) that would explain the local surface rotation stored in memory at that location of the model ($\prescript{M}{S}{\mathbf{R}_i}$). The result of this computation defines an object rotation hypothesis ($\prescript{B}{M}{\mathbf{R}}_k$, which becomes part of the evidence structure, although we technically store the inverse.
- So to be extra clear, your annotation is correct that $r_1$ is repeated for all of the model locations $i... N$ in a stack, as long as the "1" indexing is referring to the possible values of $\prescript{B}{S}{\mathbf{R}}$. However in practice, all of these rotations will be different values - the stored rotations aren't actually identical which the current notation might imply.

**Re. mismatch to evidence indexing**
- The one thing I'm unclear on is the indexing you point to at the bottom of the image/diagram, where the mismatch actually takes place. Are you able to point to the relevant part of the code where these evidence values are initialized with this ordering?

ramyamounir:
@nielsleadholm thanks for adding this context in more accurate notations, really helpful!

> I would have maybe expected it to be more catastrophic if it was totally messing with the hypotheses.

This is only applied at the hypothesis space initialization step. We currently only do that only once, at the beginning of the episode. During evidence update, the LM can recover from this by incrementing the correct hypothesis (quicker than incorrect ones) even if it started with low evidence. The problem is maybe that the initial low evidence of a correct hypothesis may be less than the `current_evidence_update_threshold`, which will prevent it from accumulating evidence and winning. This is somewhat arbitrary, thus the decrease in overall accuracy.

>  the stored rotations aren't actually identical which the current notation might imply.

Yes, I agree. For the informed hypotheses the stored rotations are not identical. But with uniform or given rotations, they are identical as shown [here](https://github.com/thousandbrainsproject/tbp.monty/blob/f939a837a2e63e8a8699c4e3a1889885c22c26e7/src/tbp/monty/frameworks/models/evidence_matching.py#L806)

> Are you able to point to the relevant part of the code where these evidence values are initialized with this ordering?

Yes, we first start by [calculating](https://github.com/thousandbrainsproject/tbp.monty/blob/f939a837a2e63e8a8699c4e3a1889885c22c26e7/src/tbp/monty/frameworks/models/evidence_matching.py#L822) the `node_feature_evidence`. This is calculated for every node in the graph, i.e., $[e_{l1}, e_{l2}, \dots]$. Then these are [stacked](https://github.com/thousandbrainsproject/tbp.monty/blob/f939a837a2e63e8a8699c4e3a1889885c22c26e7/src/tbp/monty/frameworks/models/evidence_matching.py#L830) for as many number of rotations per node we have.  Let me know if this doesn't answer your question.

> If these all start near 0, is this in some sense arbitrary as long as it has the same cardinality / array size?

Yes, I think this is the correct intuition. They all start with small evidence values (i.e., $\leq 1$), but as I mentioned above, the threshold we apply on updating evidence may prevent some from getting incremented/updated. Empirically, this has a big effect on overall accuracy and this threshold is my current explanation. If this doesn't make sense, we can further investigate why this bug seems to affect the performance more than expected.

PR #289, opened 2025-05-01
feat: initial implementation of hypotheses resampling mixin

ramyamounir:
This code mainly proposes the `ResamplingHypothesesEvidenceMixin` towards a full implementation of [RFC 9](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0009_hypotheses_resampling.md). The modifications are currently introduced as a Mixin to avoid confusion during the review process, but I imagine the modifications will later be directly applied to `EvidenceGraphLM`.

## Summary of additions
- Restructure of the `_update_evidence` function that enables for rebuilding the hypothesis space every step.
- Efficient sampling from informed hypotheses based on evidence values.
- Calculations of the required counts based on newly introduced parameters: `hypotheses_count_multiplier` and `hypotheses_existing_to_new_ratio`.

To avoid confusion, I have removed other WIP related to EvidenceSlopeTracker or offspring hypotheses resampling. This will come in a follow-up PR.

**I have not included unit tests yet because I want to gather feedback first. I have changed some of the logic in how `EvidenceGraphLM` does things, so I imagine you will have specific concerns and tests you want implemented before this PR can be merged. Therefore, this PR is still a draft**

## Preliminary Results
I have tested that I can reproduce the results from `randrot_10distinctobj_surf_agent` with the mixin applied and the sampling parameters are set to `hypotheses_count_multiplier=1.0` and `hypotheses_existing_to_new_ratio=0.0`.

Also, when allowing for the doubling of the hypothesis space size (`hypotheses_count_multiplier=2.0` and `hypotheses_existing_to_new_ratio=0.5`), we can see the theoretical limit as hypotheses based on less noisy observations are likely to be created at later steps (as opposed to relying on the first step only).
decreasing_th_limit

## Benchmark Comparisons

I branched off to a new branch "sample_informed_hypotheses_accuracy", where I only changed the default LM class in YCB experiments, as shown [here](https://github.com/ramyamounir/tbp.monty/blob/378535c99a37627a7e6ff5f0377bef5aaff705b7/benchmarks/configs/defaults.py#L74). It seems that we can accurately replicate the behavior of `EvidenceGraphLM` with `ResamplingEvidenceGraphLM` and the correct resampling parameters. The diff tables are shown below.

| Metric   |   main |   sample_informed_hypotheses_accuracy |   Δ (delta) |
|----------|--------|---------------------------------------|-------------|
| time     | 2.1854 |                                2.2167 |      0.0314 |

📌 Differences found in experiment: randrot_noise_10distinctobj_surf_agent
| Metric   |   main |   sample_informed_hypotheses_accuracy |   Δ (delta) |
|----------|--------|---------------------------------------|-------------|
| time     | 1.9791 |                                1.9868 |      0.0077 |

## Run time for efficient vs inefficient `_sample_informed`

I ran an experiment for around a 100 timesteps of one episode for the mug object. I set the count multiplier to 1.1 to force the `_sample_informed` function to sample the extra 10% of hypotheses from informed hypotheses. I repeated the experiment for `_sample_informed_inefficient`. I can see that the efficient version is better, and I expect it to be even better with larger objects (i.e., more points).

efficiency


src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+                self.evidence[graph_id][channel_end_ix:],
+            ]
+        )
+
+        self.channel_hypothesis_mapping[graph_id].resize_channel_by(
+            input_channel, len(new_evidence) - channel_size
+        )
+
+
+class ResamplingEvidenceGraphLM(ResamplingHypothesesEvidenceMixin, EvidenceGraphLM):
```
tristanls: **issue**: Put test classes in `test`, not `src`.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+                f"EvidenceGraphLM, got {cls.__bases__}"
+            )
+
+    def _num_hyps_per_node(self, features, input_channel):
+        if self.initial_possible_poses is None:
+            return 2 if features[input_channel]["pose_fully_defined"] else 8
+        else:
+            return len(self.initial_possible_poses)
+
+    def _update_evidence(
```
tristanls, line 91: **issue**: Don't override private implementations from other classes. This makes it something completely different from `EvidenceGraphLM`. Is there a reason this is not a different learning module?

Our goal for using mixins should be to  _compose_ functionality. This _overrides_ the functionality instead and ends up being as complicated as inheritance.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+        needed = current * self.hypotheses_count_multiplier
+
+        # calculate how many existing and new hypotheses needed
+        existing_maintained, new_informed = (
+            needed * (1 - self.hypotheses_existing_to_new_ratio),
+            needed * self.hypotheses_existing_to_new_ratio,
+        )
+        # needed existing hypotheses should not exceed the existing hypotheses
+        # if trying to maintain more hypotheses, set the available count as ceiling
+        # We also need to make new_informed divisible by the number of hypotheses per
```
nielsleadholm, line 207: **suggestion:** Is it possible to unpack this sentence, or reference a function that would help understand the significance of "sample by node evidence first"?

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+            graph_id (str): Identifier of the graph being queried.
+            informed_count (int): Number of fully informed hypotheses to sample.
+            input_channel: The channel for which hypotheses are sampled.
+
+        Returns:
+            Tuple[np.ndarray, np.ndarray, np.ndarray]: Tuple containing
+                selected locations, rotations, and evidence data.
+
+        """
+        # TODO: Remove this function after approving the more efficient version
```
nielsleadholm, line 356: **suggestion:** As discussed in the 1:1,  it would be worth having a unit test that checks the efficient method against this one.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+        # === Calculate selected rotations by top-k indices === #
+        if self.initial_possible_poses is None:
+            node_directions_filtered = (
+                self.graph_memory.get_rotation_features_at_all_nodes(
+                    graph_id, input_channel
+                )[top_indices]
+            )
+            sensed_directions = features[input_channel]["pose_vectors"]
+            if num_hyps_per_node == 2:
+                possible_s_d = [
```
nielsleadholm, line 304: **suggestion:** It would be nice to factor this out as a simple utility function. This is now the 3rd time it appears in the code (two times previous to your PR in `feature_location_matching` and `evidence_matching`.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+            graph_id (str): Identifier of the graph being queried.
+            existing_count (int): Number of existing hypotheses to sample.
+            input_channel: The channel for which hypotheses are sampled.
+
+        Returns:
+            Tuple[np.ndarray, np.ndarray, np.ndarray]: Tuple containing
+                selected locations, rotations, and evidence scores.
+
+        Notes:
+            Currently samples based on available count instead of evidence slope.
```
nielsleadholm, line 399: **suggestion:** Consider adding a TODO so that this comment gets updated in the future work.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+        """
+        # Return empty arrays for no hypotheses to sample
+        if existing_count == 0:
+            return np.zeros((0, 3)), np.zeros((0, 3, 3)), np.zeros(0)
+
+        # TODO implement sampling based on evidence slope.
```
nielsleadholm, line 405: **question:** It seems like it would make most sense to select these based on evidence count, if we are not using slope. Have you not implemented this just because you are planning on changing it soon anyways?

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+        self,
+        features: Dict,
+        locations: np.ndarray,
+        rotations: np.ndarray,
+        evidence: np.ndarray,
+        displacement: Dict,
+        graph_id: str,
+        input_channel: str,
+    ) -> Tuple[np.ndarray, np.ndarray]:
+        """Updates existing hypotheses by displacing them.
```
nielsleadholm, line 422: **suggestion:** It would be useful for some context for how this differs from the displacement code that exists within the baseline `_update_evidence`, e.g. just factored out, vs. a more efficient implementation. If more the former, then why not refactor the baseline class?

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+            )
+            min_update = np.clip(np.min(new_evidence), 0, np.inf)
+            evidence_to_add = np.ones_like(evidence) * min_update
+            evidence_to_add[hyp_ids_to_test] = new_evidence
+            evidence = (
+                evidence * self.past_weight + evidence_to_add * self.present_weight
+            )
+        return search_locations, evidence
+
+    def _calculate_evidence_for_new_locations(
```
nielsleadholm, line 459: **suggestion:** Again this seems to have a lot of overlap with the original / baseline version. Is it possible to clarify how it differs at a high level?

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+            return
+
+        channel_start_ix, channel_end_ix = self.channel_hypothesis_mapping[
+            graph_id
+        ].channel_range(input_channel)
+        channel_size = self.channel_hypothesis_mapping[graph_id].channel_size(
+            input_channel
+        )
+
+        self.possible_locations[graph_id] = np.concatenate(
```
nielsleadholm, line 581: **suggestion:** This looks like it could be factored out as a small utility function. It would take two arrays (e.g. `self.possible_locations` and `new_loc_hypotheses`) and the `channel_start_ix`/`channel_end_ix`. I think this would also help make it clearer what it is doing / add some implicit documentation.

nielsleadholm (commented):
Really nice! Overall looks like a clean approach, so mostly just some clarifying comments or suggestions.

The only high-level comment is that in the 1:1 you mentioned that this code should run faster. It would be great to see some speed comparisons, even if it is with a simple unit test or the like, just to sanity check the speed-ups are working as expected.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+
+    The resampling process is governed by two main parameters:
+      - `hypotheses_count_multiplier`: scales the total number of hypotheses every step.
+      - `hypotheses_existing_to_new_ratio`: controls the proportion of existing vs.
+          informed hypotheses during resampling.
+
+    Raises:
+        TypeError: If used in a class that is not a subclass of `EvidenceGraphLM`.
+
+    Compatible with:
```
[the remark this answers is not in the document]
ramyamounir: Sure, I will move the "Compatible with" block to above the "Raises" block. As for which classes go into the compatible block, I've discussed this briefly with Tristan in a previous PR [here](https://github.com/thousandbrainsproject/tbp.monty/pull/250#discussion_r2042047223). Let me know if it's still an issue so we can address it more.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+            self._replace_hypotheses_in_hpspace(
+                graph_id=graph_id,
+                input_channel=input_channel,
+                new_loc_hypotheses=channel_locations,
+                new_rot_hypotheses=channel_rotations,
+                new_evidence=channel_evidence,
+            )
+
+        end_time = time.time()
+        assert not np.isnan(np.max(self.evidence[graph_id])), "evidence contains NaN."
```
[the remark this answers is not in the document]
ramyamounir: This line and the logging beneath are carried over from the original `_update_evidence`. I figured @vkakerbeck may have encountered this enough times while debugging to write this assert statement, so I thought I'd also keep it here too. I could remove it if not longer needed..

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,613 @@
...
+            raise TypeError(
+                "ResamplingHypothesesEvidenceMixin must be mixed in with a subclass of "
+                f"EvidenceGraphLM, got {cls.__bases__}"
+            )
+
+    def _num_hyps_per_node(self, features, input_channel):
+        if self.initial_possible_poses is None:
+            return 2 if features[input_channel]["pose_fully_defined"] else 8
+        else:
+            return len(self.initial_possible_poses)
```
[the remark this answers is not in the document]
ramyamounir: The `initial_possible_poses` that is passed to the class arguments from the configs can contain a string (e.g., `uniform` or `informed`), but this goes through the `get_initial_possible_poses` function [here](https://github.com/thousandbrainsproject/tbp.monty/blob/a73d28f63b33963a2a7a6675871d669c6b54932d/src/tbp/monty/frameworks/models/evidence_matching.py#L330), which always resolves it to a set of rotations if a string was passed.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: Yes, good point. I could add a unit test that request a fraction of the available graph nodes (e.g., half the graph nodes) and see which one is more efficient. It is expected that as we request less informed hypotheses, we would see a bigger difference in efficiency between the two methods.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: Yes, there is a TODO [here](https://github.com/thousandbrainsproject/tbp.monty/blob/fdde7e5e44ff56eb43b2d743432e81b9fa88dd57/src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py#L405). I will just remove this redundant note from the docstring.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: Yes, I just didn't want to waste time on something that will change very soon.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: It looks like there is repetitive code that needs to be refactored, but it really is just one simple function. I feel refactoring this function only makes it harder to read. I have added a comment to explain what is going on in this concatenate operation. If you feel this is still confusing, let me know.

tests/unit/frameworks/models/mixins/resample_hypotheses_evidence_test.py
```diff
@@ -0,0 +1,32 @@
...
+from tbp.monty.frameworks.models.abstract_monty_classes import LearningModule
+from tbp.monty.frameworks.models.evidence_matching import EvidenceGraphLM
+from tbp.monty.frameworks.models.mixins.resampling_hypotheses_evidence import (
+    ResamplingHypothesesEvidenceMixin,
+)
+
+
+class InheritanceResampleHypothesesMixinTest(TestCase):
+    @staticmethod
+    def test_mixin_used_with_compatible_learning_module_does_not_error() -> None:
```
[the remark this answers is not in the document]
ramyamounir: At the moment, `NoResetEvidenceGraphLM` already uses this resampling mixin. Subclassing it here again as a unit test will raise an error.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: When writing this mixin and refactoring some of the logic in `_update_evidence`, I tried to make every refactored function self contained, i.e., takes some hypotheses, does something to them and return them. I tried to avoid relying on self.possible_locations, self.possible_poses or self.evidence. This way these functions can be easily reused, unit-tested, composed, etc.

So for example, when I displace hypotheses, I send the actual hypotheses to be displaced and I don't just pass indices or ranges and expect it to directly modify the full class hypotheses. This is the biggest difference in some of the functions with high overlap with baseline logic.

Also note that when I wrote this mixin, I intended for it to completely replace existing logic in `EvidenceGraphLM`, so eventually, we wouldn't have repeated code.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: Yes, valid point. See my answer [here](https://github.com/thousandbrainsproject/tbp.monty/pull/289#discussion_r2074348453) to a similar suggestion.

ramyamounir:
Thanks for your quick and thorough review @nielsleadholm! I have addressed your comments and kept a note of the tests you suggested, please let me know if you think of more tests.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
nielsleadholm: Ok thanks for the clarification. I think that's a really nice change, so if I understand correctly, these are refactors of existing code that should have the same functionality. I think that would be clearer by overwriting the original blocks of code in `EvidenceGraphLM`, which could be opened as a proceeding, smaller PR that just makes that change and verifies the same results on benchmark experiments.

That would also make it more transparent what potential changes are happening, e.g. it looks like some things like logging and the comments might differ.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -271,10 +288,16 @@ def _sample_informed(
             node_feature_evidence = self._calculate_feature_evidence_for_all_nodes(
                 features, input_channel, graph_id
             )
+            # Find the indices to nodes with highest evidence scores. The sorting
```
nielsleadholm, line 291:
```suggestion
            # Find the indices for the nodes with highest evidence scores. The sorting
```

src/tbp/monty/frameworks/utils/graph_matching_utils.py
```diff
@@ -462,3 +464,36 @@ def find_step_on_new_object(
...
+    sensed_directions: np.ndarray, num_hyps_per_node: int
+) -> np.ndarray:
+    """Returns the possible sensed directions for all nodes.
+
+    This function determines the possible sensed directions for a given set of sensed
+    direction. It relies on two different behaviors depending on the value of
+    num_hyps_per_node.
+
+    If num_hyps_per_node equals 2, then pose is well defined (i.e., PC1 != PC2).
+    Otherwise, this function is samples additional poses in the plane perpendicular
```
nielsleadholm, line 479:
```suggestion
    Otherwise, this function samples additional poses in the plane perpendicular
```

src/tbp/monty/frameworks/utils/graph_matching_utils.py
```diff
@@ -462,3 +464,36 @@ def find_step_on_new_object(
...
         return None
+
+
+def possible_sensed_directions(
+    sensed_directions: np.ndarray, num_hyps_per_node: int
+) -> np.ndarray:
+    """Returns the possible sensed directions for all nodes.
+
+    This function determines the possible sensed directions for a given set of sensed
+    direction. It relies on two different behaviors depending on the value of
```
nielsleadholm, line 475:
```suggestion
    directions. It relies on two different behaviors depending on the value of
```

src/tbp/monty/frameworks/utils/graph_matching_utils.py
```diff
@@ -462,3 +464,36 @@ def find_step_on_new_object(
...
+    num_hyps_per_node.
+
+    If num_hyps_per_node equals 2, then pose is well defined (i.e., PC1 != PC2).
```
nielsleadholm, line 478: **suggestion:** It would be worth mentioning however that this does not distinguish between the mirrored directions of PC1 and PC2, hence why we sample two.

nielsleadholm (commented):
Nice thanks for those responses and updates. The main points now are just about adding the unit tests, and I would suggest separating out the refactoring of evidence LM's update evidence. Otherwise I think it's ready to make a full PR.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: Yeah sounds good, this will include the restructuring of `_update_evidence` in `EvidenceGraphLM`, but none of the resampling stuff. I will do it with resampling in mind though so that we can just override a few functions in this PR.

ramyamounir:
This PR is now ready for review. No rush, this can be done after the Hackathon/brainstorming.

@vkakerbeck @nielsleadholm, I've added some comparison results to the description of this PR, mainly to make sure that we can get the same exact results of the branch `main` on a couple of YCB benchmark experiments, as well as to test if the new `_sample_informed` is actually as efficient as advertised.

@tristanls @jeremyshoemaker, I've written some unit tests for the `_sample_count` functionality but I'm not happy with how I did it. I wanted to split the tests into different functions (right now they are all under one function which counts as one test) but I didn't want to rerun the `setUp` function for every test because it takes time to train the model and get those pretrained graphs. Is there a workaround I'm not seeing?

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -367,6 +378,12 @@ def _sample_informed_inefficient(
             Tuple[np.ndarray, np.ndarray, np.ndarray]: Tuple containing
                 selected locations, rotations, and evidence data.
 
+        Note:
+            This function will not give the exact same results as the existing
+            `_sample_informed`. Here we directly sample the top `informed_count`
+            from based on the evidence scores. In `_sample_informed`, we sample
```
nielsleadholm:
```suggestion
            based on the evidence scores. In `_sample_informed`, we sample
```

tests/unit/frameworks/models/mixins/resample_hypotheses_evidence_test.py
```diff
@@ -30,3 +61,273 @@ def test_mixin_used_with_non_compatible_learning_module_raises_error() -> None:
...
+    def _count_multiplier(self, rlm):
+        """This tests that the count multiplier correctly scales the hypothesis space.
+
+        The count multiplier parameter is used to scale the hypothesis space between
+        steps. For example, a multiplier of 2, will request to double the number of
+        hypotheses.
+        """
+        graph_id = "capsule3DSolid"
+        graph_num_nodes = self._graph_node_count(rlm, graph_id)
+        before_count = graph_num_nodes * 2
```
nielsleadholm: I think including this at least once will make the test easier to read in the distant future. Alternatively, you could consider just for the unit tests defining a variable like `well_defined_num_poses` or the like, and replace all the hard-coded 2s in these tests.

```suggestion
        before_count = graph_num_nodes * 2  # Define before_count, where pose_defined = True
```

tests/unit/frameworks/models/mixins/resample_hypotheses_evidence_test.py
```diff
@@ -30,3 +61,273 @@ def test_mixin_used_with_non_compatible_learning_module_raises_error() -> None:
...
+        existing_count, informed_count = self.run_sample_count(
+            rlm=rlm,
+            count_multiplier=requested_count_multiplier,
+            existing_to_new_ratio=0.5,
+            pose_defined=True,
+            graph_id=graph_id,
+        )
+        self.assertEqual(expected_count, existing_count + informed_count)
+
+        # existing is 72 and informed is 8*36=288 (total is 360)
```
nielsleadholm:
```suggestion
        # Existing is 72 and informed is 8*36=288 (total is 360)
```

tests/unit/frameworks/models/mixins/resample_hypotheses_evidence_test.py
```diff
@@ -30,3 +61,273 @@ def test_mixin_used_with_non_compatible_learning_module_raises_error() -> None:
...
+
+    def _count_ratio(self, rlm):
+        """This tests that the resampling ratio of new hypotheses is correct.
+
+        The existing_to_new_ratio parameter is used to control the ratio of how many
+        existing vs. informed hypotheses to resample. This test ensures that the
+        `_sample_count` function follows the expected behavior of this ratio parameter.
+
+        Note that the `_sample_count` function will prioritize the multiplier count
+        parameter over this ratio parameter. In other words, if no enough existing
```
nielsleadholm:
```suggestion
        parameter over this ratio parameter. In other words, if not enough existing
```

tests/unit/frameworks/models/mixins/resample_hypotheses_evidence_test.py
```diff
@@ -30,3 +61,273 @@ def test_mixin_used_with_non_compatible_learning_module_raises_error() -> None:
...
+        count_multiplier = 2
+
+        for ratio in [0.0, 0.1, 0.5, 0.9, 1.0]:
+            requested_existing_count = (
+                available_existing_count * count_multiplier * (1.0 - ratio)
+            )
+            requested_informed_count = (
+                available_existing_count * count_multiplier * ratio
+            )
+            maximum_available_existing_count = available_existing_count
```
nielsleadholm, line 280: **nitpick:** Not something you need to change, but repeated variable assignments in for-loops are always worth looking out for to avoid!

tests/unit/frameworks/models/mixins/resample_hypotheses_evidence_test.py
```diff
@@ -30,3 +61,273 @@ def test_mixin_used_with_non_compatible_learning_module_raises_error() -> None:
...
+            maximum_available_informed_count = graph_num_nodes * 2
```
nielsleadholm: **suggestion (non-blocking):** It's unlikely but it feels like there is some symmetry in this test that could mask a potential issue, in particular `maximum_available_existing_count = maximum_available_informed_count`. As a result, some tests might pass simply because the conditions for the two groups of hypotheses are so similar. I think you could make it a bit more interesting and break this by assuming that the new points are based on an undefined PC (i.e. use the umbilical sampling for `maximum_available_informed_count`.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,454 @@
...
+        displacements: Optional[Dict],
+        graph_id: str,
+    ) -> None:
+        """Update evidence of hypotheses space with resampling.
+
+        Updates existing hypotheses space by:
+            1. Calculating sample count for existing and informed hypotheses
+            2. Sampling hypotheses for existing and informed hypotheses types
+            3. Displacing (and updating evidence of) existing hypotheses using
+                given displacements
```
vkakerbeck:
```suggestion
                given displacements and sensed features
```

tests/unit/frameworks/models/mixins/resample_hypotheses_evidence_test.py
```diff
@@ -30,3 +61,273 @@ def test_mixin_used_with_non_compatible_learning_module_raises_error() -> None:
...
+            )
+            expected_existing_count = min(
+                maximum_available_existing_count,
+                requested_existing_count,
+            )
+            self.assertEqual(existing_count, int(expected_existing_count))
+
+            missing_existing_hypotheses = (
+                requested_existing_count - expected_existing_count
+            )
```
nielsleadholm, line 302:
```suggestion
            )  # Will be zero, or otherwise the count that informed hypotheses need to fill in
```

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,454 @@
...
+                selected locations, rotations, and evidence data.
+
+        Note:
+            This function will not give the exact same results as the existing
+            `_sample_informed`. Here we directly sample the top `informed_count`
+            from based on the evidence scores. In `_sample_informed`, we sample
+            `informed_count/num_hyps_per_node` points then tile it for
+            `num_hyps_per_node`.
+        """
+        # TODO: Remove this function after approving the more efficient version
```
nielsleadholm: Thanks for including those speed test results in the PR. I think we can remove this function now before merging, unless you see any other reason to keep it.

tests/unit/frameworks/models/mixins/resample_hypotheses_evidence_test.py
ramyamounir: I created a simple function `_num_hyps_multiplier` for the unit tests to address this.

tests/unit/frameworks/models/mixins/resample_hypotheses_evidence_test.py
ramyamounir: Makes sense, I did some refactoring to rerun most of these tests with defined and undefined PCs. Please let me know if this addresses your concerns.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: Great, I'll plan to remove this function before merging unless the rest of reviewers bring up issues with the efficient version.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,454 @@
...
+        # Select the corresponding entries from the original arrays
+        selected_locations = initial_possible_channel_locations[top_indices]
+        selected_rotations = initial_possible_channel_rotations[top_indices]
+        selected_evidence = channel_evidence[top_indices]
+
+        return selected_locations, selected_rotations, selected_evidence
+
+    def _sample_existing(
+        self,
+        features: Dict,
```
tristanls: **issue**: `features` is unused.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,454 @@
...
+        ].channel_range(input_channel)
+        selected_locations = self.possible_locations[graph_id][
+            channel_start_ix:channel_end_ix
+        ][:existing_count]
+        selected_rotations = self.possible_poses[graph_id][
+            channel_start_ix:channel_end_ix
+        ][:existing_count]
+        selected_evidence = self.evidence[graph_id][channel_start_ix:channel_end_ix][
+            :existing_count
+        ]
```
tristanls: **issue**: This looks like it should be a `ChannelMapper` functionality or you should use `mapper.extract`.

```python
# ChannelMapper

def first(self, original: np.ndarray, channel: str, count: int) -> np.ndarray:
    arr = self.extract(original, channel)
    return arr[:count]
```

Then this code block becomes:
```python
selected_locations = mapper.first(self.possible_locations[graph_id], input_channel, existing_count)
selected_rotations = mapper.first(self.possible_poses[graph_id], input_channel, existing_count)
selected_evidence = mapper.first(self.evidence[graph_id], input_channel, existing_count)
```

Alternatively, use `mapper.extract`:

```python
selected_locations = mapper.extract(self.possible_locations[graph_id], input_channel)[:existing_count]
selected_rotations = mapper.extract(self.possible_poses[graph_id], input_channel)[:existing_count]
selected_evidence = mapper.extract(self.evidence[graph_id], input_channel)[:existing_count]
```

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,454 @@
...
+        ).shape[0]
+        num_hyps_per_node = self._num_hyps_per_node(features, input_channel)
+        full_informed_count = graph_num_points * num_hyps_per_node
+
+        # If hypothesis space does not exist, we initialize with informed hypotheses
+        if input_channel not in self.channel_hypothesis_mapping[graph_id].channels:
+            return 0, full_informed_count
+
+        # Calculate the total number of hypotheses needed
+        current = self.channel_hypothesis_mapping[graph_id].channel_size(input_channel)
```
tristanls: **suggestion**: Use mapper here for readability, as in:
```python
# If hypothesis space does not exist, we initialize with informed hypotheses
if input_channel not in mapper.channels:
            return 0, full_informed_count

# Calculate the total number of hypotheses needed
current = mapper.channel_size(input_channel)
```

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
tristanls: I didn't want to point out the problem without a solution, so #317 proposes refactoring `EvidenceGraphLM` to encapsulate evidence-updating functionality in an `EvidenceUpdater`. If we were to merge that code, instead of creating a mixin, you would make a `ResamplingEvidenceUpdater` (with the same functionality as this mixin) and configure `EvidenceGraphLM`.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,454 @@
...
+        """Samples the specified number of fully informed hypotheses.
+
+        This method selects hypotheses that are most likely to be informative based on
+        feature evidence. Specifically, it identifies the top-k nodes with the highest
+        evidence scores and samples hypotheses only from those nodes, making the process
+        more efficient than uniformly sampling from all graph nodes.
+
+        The sampling includes:
+          - Selecting the top-k node indices based on evidence scores, where k is
+             determined by the `informed_count` and the number of hypotheses per node.
```
vkakerbeck, line 269: This is a bit confusing. It sounds like you mean based on evidence scores that have been accumulated so far but as far as I understand the code it is actually based on the difference between the current sensed feature and the stored features in the graph (just at the current time step). I would not call those evidence scores since those are the ones we accumulate over time and those hypothesis locations will not correspond to locations in the graph.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,454 @@
...
+                node_feature_evidence[top_indices] * self.feature_evidence_increment
+            )
+        else:
+            num_nodes = self.graph_memory.get_num_nodes_in_graph(graph_id)
+            top_indices = np.arange(num_nodes)[
+                : int(informed_count // num_hyps_per_node)
+            ]
+            node_feature_evidence_filtered = np.zeros(len(top_indices))
+
+        selected_evidence = np.tile(node_feature_evidence_filtered, num_hyps_per_node)
```
vkakerbeck: Same with the code block here. I would make it more clear in the variable namings that this is different from the hypothesis evidence counts we are maintaining. Maybe just call it node_feature_match or best_match_nodes.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,454 @@
...
+        return selected_locations, selected_rotations, selected_evidence
+
+    def _sample_existing(
+        self,
+        features: Dict,
+        graph_id: str,
+        existing_count: int,
+        input_channel: str,
+    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+        """Samples the specified number of existing hypotheses.
```
vkakerbeck:
```suggestion
        """Samples the specified number of existing hypotheses to retain.
```

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
```diff
@@ -0,0 +1,454 @@
...
+
+        Returns:
+            Tuple[np.ndarray, np.ndarray, np.ndarray]: Tuple containing
+                selected locations, rotations, and evidence scores.
+        """
+        # Return empty arrays for no hypotheses to sample
+        if existing_count == 0:
+            return np.zeros((0, 3)), np.zeros((0, 3, 3)), np.zeros(0)
+
+        # TODO implement sampling based on evidence slope.
```
vkakerbeck, line 386: If you are going to run any experiments before implementing this I would at the very least sample the hypotheses randomly. Otherwise we just cut off entire chunks of the object out of the hypothesis space which will bias the results quite a bit.

vkakerbeck (approved):
Nice! Looking forward to further results from this :) Thanks for including the speed comparison between the two implementation versions.

One minor curiosity: When you looked at the benchmark changes in runtime, did you happen to have a look at one of the 77 object experiments? I wonder if the slow down is more pronounced there. And if so if it makes sense to add a parameter to only update the hypothesis space every n steps.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: Yes, you're right. This part of the code was written before we added the extract functionality in the other PR and I forgot to update it in this PR.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: Makes sense, how about we use `step_evidence` to distinguish between current step evidence and accumulated evidence scores over time. In my mind, these are still evidence scores that will be added to the cumulative evidence scores. I would then rename `selected_evidence` to `selected_step_evidence`.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: Nice, thanks. I agree that modifying the encapsulated functionality would be the preferred approach here. As discussed elsewhere, this would be implemented as a follow-up PR so as not to be a blocker to this PR.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: See comment [here](https://github.com/thousandbrainsproject/tbp.monty/pull/289#discussion_r2126807758)

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: This is more of a placeholder for sampling based on evidence slope. I am not planning to run any experiments that would require sampling a subset of the existing hypotheses space until I add the logic for evidence slope. @nielsleadholm had a similar concern [here](https://github.com/thousandbrainsproject/tbp.monty/pull/289#discussion_r2073665887).

ramyamounir:
Thank you all for the reviews!

@nielsleadholm 
> the handling of what happens when we request more hypotheses than can be satisfied based on informed etc. poses is quite complex.

Yes, I totally agree. The logic here is complex. We could simplify this a bit by changing the user parameters from `hypotheses_count_multiplier`, `hypotheses_existing_to_new_ratio`, `hypotheses_informed_to_offspring_ratio` to:
- `hypotheses_existing_multiplier`: Defines how many of the existing hypotheses to retain. Range is [0, 1], where zero means retain none of the existing hypotheses, and one retains all of them.
- `hypotheses_informed_multiplier`: Defines how many of the informed hypotheses to add to the hyp space. Range is [0, 1], where zero means do not add any informed hyps and one adds all available informed hypotheses (would vary depending on whether PC is defined or not).
- `hypotheses_offspring_multiplier`: Defines how many offspring hypotheses to add to the hyp space. Range is [0, ∞]. This value is multiplied by the number of points in the graph. So, a value of zero will not add any hyps, but a value of 2 will add twice the nodes in the graph as offspring hyps.

The problem here is that we would not directly control the total count multiplier, but it's much clearer. What do you think?

@vkakerbeck 
> When you looked at the benchmark changes in runtime, did you happen to have a look at one of the 77 object experiments? I wonder if the slow down is more pronounced there. And if so if it makes sense to add a parameter to only update the hypothesis space every n steps.

I should probably clarify that the benchmark comparison experiment at the top was only to show that we can reproduce the behavior of the default `EvidenceGraphLM` with the mixin by adjusting some parameters. I could essentially just use `hypotheses_count_multiplier = 1` and `hypotheses_existing_to_new_ratio = 0` and the resampling would be bypassed completely and we would get the same `EvidenceGraphLM` behavior. So no resampling is expected in this experiment, which is why only the time changed.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
vkakerbeck: Hm I would probably make it a bit more specific like `feature_match` or `feature_evidence`. This is not the whole evidence we would add at one step (doesn't include displacement/pose based evidence) and it is not even part of the evidence we will add at this step (the evidence for the current step is calculated based on the possible_locations, not based on the nodes in the graph). I think it would be good to make the distinction very clear here. We just look at how well the current observed features match features stored in the graph to initialize some new possible locations. This is very different from taking the current hypothesized locations, adding the displacement, doing kd tree search to find the nearest points in the graph, and then comparing features.

vkakerbeck:
Ah okay, makes sense re. the benchmark experiments. Is the time change just random variation (i.e. if hypotheses_existing_to_new_ratio==0 you don't call any of these new functions)? Just because I don't remember seeing such an explicit check but I may just be forgetting about it.

ramyamounir:
@vkakerbeck There isn't an explicit check, but this set of parameters will result in [`informed_count=0`](https://github.com/ramyamounir/tbp.monty/blob/fcb080cdbbadec21560bfa06e7214c4d3f5619b8/src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py#L130), which leads to the sampling bypass over [here](https://github.com/ramyamounir/tbp.monty/blob/fcb080cdbbadec21560bfa06e7214c4d3f5619b8/src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py#L293). I wanted to do the benchmark comparison to make sure that this works as expected and we don't get any mismatch other than random time variance. I would guess that having to go through `_sample_count` every function might be responsible for this insignificant increase in episode time, but we know it's not actually resampling or changing the logic of Monty.

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: I see. I have been thinking about this a little differently, each type of hypothesis (e.g., existing, informed, offspring) contribute its own local `step_evidence` scores to the cumulative evidence, regardless of how that evidence is calculated. If it gets [concatenated](https://github.com/ramyamounir/tbp.monty/blob/fcb080cdbbadec21560bfa06e7214c4d3f5619b8/src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py#L168) with other evidence scores then it qualifies as `evidence`. Some evidence scores are calculated with feature matching only because they don't have a displacement yet (i.e., informed and offspring), other scores are calculated using displacement and feature matching (i.e., existing). But these are all evidence scores that get added to the same [`self.evidence` array](https://github.com/ramyamounir/tbp.monty/blob/451ba5f576cfdf3eb1b14250d85125429b93d1e4/src/tbp/monty/frameworks/models/evidence_matching.py#L1095).

If I were to choose between `feature_match` and `feature_evidence`, I would probably go for the latter. I feel `feature_match` makes sense but it would be confusing when it gets magically concatenated with evidence scores.

nielsleadholm: (just quickly chiming in to say `feature_evidence` sounds intuitive to me for this value).

ramyamounir: Great, I've renamed evidence to feature_evidence for the informed hypotheses instances (e.g., `selected_feature_evidence`, `informed_hypotheses_feature_evidence`). If nothing else, I think this is ready to merge.

vkakerbeck: Oh okay, I was thinking of this as being used to determine the top_k locations to add to the hypothesis space. So basically looking at the match between currently sensed features and features stored in the graph to add new locations to the hypothesis space. Conceptually I was thinking of this being a separate process from how we initialize the evidence for the new hypotheses (as I thought this may be determined by average existing evidence or similar). If this is the same, the feature_evidence is fine by me. If we change this later (which I think you plan to?) I would make the difference more clear.

vkakerbeck:
Okay yeah I was just wondering if that unnecessary pass through _sample_count when we are not actually resampling might be more significant when there are more objects/hypotheses. But it probably isn't. Just another place we could optimize runtime when we need to (same reason I suggested not resampling hypotheses at every step). But yeah this can be followup work if we notice that runtimes on all object experiments get too long (still something I would test).

src/tbp/monty/frameworks/models/mixins/resampling_hypotheses_evidence.py
ramyamounir: Yeah, this makes sense. A better initialization of evidence scores for informed hypotheses will come in a follow-up PR. This PR just uses the selected feature_match scores as the initial evidence scores. I'll keep this distinction in mind for follow-up PRs.

PR #300, opened 2025-05-06
refactor: remove unused else case

tristanls:
Currently, there is no case where a model is not an instance of `MontyBase`. This pull request removes the else branch that will never be executed.


vkakerbeck (approved):
LGTM

PR #301, opened 2025-05-06
refactor: restructuring of _update_evidence

ramyamounir:
Based on some discussions with @nielsleadholm [here](https://github.com/thousandbrainsproject/tbp.monty/pull/289#discussion_r2074348453), we agreed to separate out the restructuring of `_update_evidence` in `EvidenceGraphLM` from the resampling logic.

This PR aims to refactor and simplify the `_update_evidence` function. The functionality remains the same but much of the logic has been refactored into other functions (e.g., `_displace_hypotheses`). Some of the refactoring aims to make the functions self-contained. For example instead of passing the range of indices in the hypothesis space and letting the function directly modify the hypothesis space (class variables), I pass the actual hypotheses and return the modified ones. This way these functions can be easily reused, unit-tested, composed, and are less prone to error.

I have tested running the benchmarks `randrot_noise_10distinctobj_dist_agent` and `randrot_noise_10distinctobj_surf_agent`. Both resulted in the exact same performance except for the average time per episode which has increased by 0.0296 seconds for the distant agent and 0.0314 for the surface agent.

### DIFF
--- a/src/tbp/monty/frameworks/models/evidence_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching.py
@@ -12,6 +12,7 @@ import copy
 import logging
 import threading
 import time
+from typing import Dict, Optional, Tuple
 
 import numpy as np
 from scipy.spatial import KDTree
@@ -898,188 +899,208 @@ class EvidenceGraphLM(GraphLM):
         self.possible_matches = self._threshold_possible_matches()
         self.current_mlh = self._calculate_most_likely_hypothesis()
 
-    def _update_evidence(self, features, displacements, graph_id):
-        """Update evidence for poses of graph_id.
+    def _update_evidence(
+        self,
+        features: Dict,
+        displacements: Optional[Dict],
+        graph_id: str,
+    ) -> None:
+        """Update evidence of hypotheses space based on sensor displacement.
 
-        replaces _update_matches_using_features
+        Updates existing hypotheses space or initializes a new hypothesis space
+        if one does not exist (i.e., at the beginning of the episode). This process
+        is repeated for each input channel in the graph.
 
-        - start with 0 evidence -> flat prior
-        - post-features & displacement: +/- evidence
-        - features: + evidence
-        Evidence is weighted by distance of hypothesis to point in model.
-        ----------not included in this function------
-        - votes: add average of nearby incoming evidence votes (could also be weighted
-          by distance)
-        ----------added later on perhaps?-----
-        - displacement recognition (stored edges in graph as ppf): + evidence for
-          locations
-
-        if first step (not moved yet):
-            - get initial hypothesis space using observed pose features
-            - initialize evidence for hypotheses using first observed features
-        else:
-            - update evidence for hypotheses using all observed features
-            and displacement
-
-        Raises:
-            ValueError: If no input channels are found to initializing hypotheses
+        Args:
+            features (dict): input features
+            displacements (dict or None): given displacements
+            graph_id (str): identifier of the graph being updated
         """
         start_time = time.time()
-        all_input_channels = list(features.keys())
-        input_channels_to_use = []
-        for input_channel in all_input_channels:
-            if input_channel in self.get_input_channels_in_graph(graph_id):
-                # NOTE: We might also want to check the confidence in the input channel
-                # features. This information is currently not available here. Once we
-                # pull the observation class into the LM we could add this (TODO S).
-                input_channels_to_use.append(input_channel)
-        # Before moving we initialize the hypothesis space:
-        if displacements is None:
-            if len(input_channels_to_use) == 0:
-                # QUESTION: Do we just want to continue until we get input?
-                raise ValueError(
-                    "No input channels found to initializing hypotheses. Make sure"
-                    " there is at least one channel that is also stored in the graph."
-                )
-            # This is the first observation before we moved -> check where in the
-            # graph the feature can be found and initialize poses & evidence
-            initial_possible_locations = []
-            initial_possible_rotations = []
-            initial_evidence = []
-            channel_mapper = ChannelMapper()
-            for input_channel in input_channels_to_use:
+
+        # Initialize a `ChannelMapper` to keep track of input channel range
+        # of hypotheses for a specific graph_id
+        if graph_id not in self.channel_hypothesis_mapping:
+            self.channel_hypothesis_mapping[graph_id] = ChannelMapper()
+
+        # Get all usable input channels
+        input_channels_to_use = [
+            ic
+            for ic in features.keys()
+            if ic in self.get_input_channels_in_graph(graph_id)
+        ]
+
+        for input_channel in input_channels_to_use:
+            # Determine if hypothesis space exists
+            initialize_hyp_space = bool(
+                input_channel not in self.channel_hypothesis_mapping[graph_id].channels
+            )
+
+            # Initialize a new hypothesis space using graph nodes
+            if initialize_hyp_space:
                 (
-                    initial_possible_channel_locations,
-                    initial_possible_channel_rotations,
-                    channel_evidence,
+                    locations,
+                    rotations,
+                    evidence,
                 ) = self._get_initial_hypothesis_space(
                     features, graph_id, input_channel
                 )
-                initial_possible_locations.append(initial_possible_channel_locations)
-                initial_possible_rotations.append(initial_possible_channel_rotations)
-                initial_evidence.append(channel_evidence)
-                channel_mapper.add_channel(input_channel, len(channel_evidence))
-            self.possible_locations[graph_id] = np.concatenate(
-                initial_possible_locations, axis=0
-            )
-            self.possible_poses[graph_id] = np.concatenate(
-                initial_possible_rotations, axis=0
-            )
-            self.evidence[graph_id] = (
-                np.concatenate(initial_evidence, axis=0) * self.present_weight
-            )
-            self.channel_hypothesis_mapping[graph_id] = channel_mapper
-            logging.debug(
-                f"\nhypothesis space for {graph_id}: {self.evidence[graph_id].shape[0]}"
-            )
-            assert (
-                self.evidence[graph_id].shape[0]
-                == self.possible_locations[graph_id].shape[0]
-            )
-        # ---------------------------------------------------------------------------
-        # Use displacement and new sensed features to update evidence for hypotheses.
-        else:
-            if len(input_channels_to_use) == 0:
-                logging.info(
-                    f"No input channels observed for {graph_id} that are stored in . "
-                    "the model. Not updating evidence."
-                )
+            # Retrieve existing hypothesis space for a specific input channel
+            else:
+                channel_start_ix, channel_end_ix = self.channel_hypothesis_mapping[
+                    graph_id
+                ].channel_range(input_channel)
+                locations = self.possible_locations[graph_id][
+                    channel_start_ix:channel_end_ix
+                ]
+                rotations = self.possible_poses[graph_id][
+                    channel_start_ix:channel_end_ix
+                ]
+                evidence = self.evidence[graph_id][channel_start_ix:channel_end_ix]
 
-            channel_mapper = self.channel_hypothesis_mapping[graph_id]
-            for input_channel in input_channels_to_use:
-                # If channel features are observed for the first time, initialize
-                # hypotheses for them.
-                if input_channel not in channel_mapper.channels:
-                    # TODO H: When initializing a hypothesis for a channel later on,
-                    # include most likely existing hypothesis from other channels?
-                    (
-                        initial_possible_channel_locations,
-                        initial_possible_channel_rotations,
-                        channel_evidence,
-                    ) = self._get_initial_hypothesis_space(
-                        features, graph_id, input_channel
+                # We only displace existing hypotheses since the newly sampled
+                # hypotheses should not be affected by the displacement from the last
+                # sensory input.
+                if not initialize_hyp_space:
+                    locations, evidence = self._displace_hypotheses(
+                        features,
+                        locations,
+                        rotations,
+                        evidence,
+                        displacements,
+                        graph_id,
+                        input_channel,
                     )
 
-                    self._add_hypotheses_to_hpspace(
-                        graph_id=graph_id,
-                        input_channel=input_channel,
-                        new_loc_hypotheses=initial_possible_channel_locations,
-                        new_rot_hypotheses=initial_possible_channel_rotations,
-                        new_evidence=channel_evidence,
-                    )
+            # === RE-BUILD HYPOTHESIS SPACE ===
+            self._replace_hypotheses_in_hpspace(
+                graph_id=graph_id,
+                input_channel=input_channel,
+                new_loc_hypotheses=locations,
+                new_rot_hypotheses=rotations,
+                new_evidence=evidence,
+            )
 
-                else:
-                    # Get the observed displacement for this channel
-                    displacement = displacements[input_channel]
-                    # Get the IDs range in hypothesis space for this channel
-                    channel_start, channel_end = channel_mapper.channel_range(
-                        input_channel
-                    )
-                    # Have to do this for all hypotheses so we don't loose the path
-                    # information
-                    rotated_displacements = self.possible_poses[graph_id][
-                        channel_start:channel_end
-                    ].dot(displacement)
-                    search_locations = (
-                        self.possible_locations[graph_id][channel_start:channel_end]
-                        + rotated_displacements
-                    )
-                    # Threshold hypotheses that we update by evidence for them
-                    current_evidence_update_threshold = (
-                        self._get_evidence_update_threshold(graph_id)
-                    )
-                    # Get indices of hypotheses with evidence > threshold
-                    hyp_ids_to_test = np.where(
-                        self.evidence[graph_id][channel_start:channel_end]
-                        >= current_evidence_update_threshold
-                    )[0]
-                    num_hypotheses_to_test = hyp_ids_to_test.shape[0]
-                    if num_hypotheses_to_test > 0:
-                        logging.info(
-                            f"Testing {num_hypotheses_to_test} out of "
-                            f"{self.evidence[graph_id].shape[0]} hypotheses for "
-                            f"{graph_id} "
-                            f"(evidence > {current_evidence_update_threshold})"
-                        )
-                        search_locations_to_test = search_locations[hyp_ids_to_test]
-                        # Get evidence update for all hypotheses with evidence > current
-                        # _evidence_update_threshold
-                        new_evidence = self._calculate_evidence_for_new_locations(
-                            graph_id,
-                            input_channel,
-                            search_locations_to_test,
-                            features,
-                            hyp_ids_to_test,
-                        )
-                        min_update = np.clip(np.min(new_evidence), 0, np.inf)
-                        # Alternatives (no update to other Hs or adding avg) left in
-                        # here in case we want to revert back to those.
-                        # avg_update = np.mean(new_evidence)
-                        # evidence_to_add = np.zeros_like(self.evidence[graph_id])
-                        evidence_to_add = (
-                            np.ones_like(
-                                self.evidence[graph_id][channel_start:channel_end]
-                            )
-                            * min_update
-                        )
-                        evidence_to_add[hyp_ids_to_test] = new_evidence
-                        # If past and present weight add up to 1, equivalent to
-                        # np.average and evidence will be bound to [-1, 2]. Otherwise it
-                        # keeps growing.
-                        self.evidence[graph_id][channel_start:channel_end] = (
-                            self.evidence[graph_id][channel_start:channel_end]
-                            * self.past_weight
-                            + evidence_to_add * self.present_weight
-                        )
-                    self.possible_locations[graph_id][channel_start:channel_end] = (
-                        search_locations
-                    )
         end_time = time.time()
         assert not np.isnan(np.max(self.evidence[graph_id])), "evidence contains NaN."
         logging.debug(
             f"evidence update for {graph_id} took "
             f"{np.round(end_time - start_time, 2)} seconds."
             f" New max evidence: {np.round(np.max(self.evidence[graph_id]), 3)}"
+        )
+
+    def _displace_hypotheses(
+        self,
+        features: Dict,
+        locations: np.ndarray,
+        rotations: np.ndarray,
+        evidence: np.ndarray,
+        displacement: Dict,
+        graph_id: str,
+        input_channel: str,
+    ) -> Tuple[np.ndarray, np.ndarray]:
+        """Updates existing hypotheses by displacing them.
+
+        Args:
+            features (dict): Input features
+            locations (np.ndarray): Hypothesized sensor locations for each hypothesis
+            rotations (np.ndarray): Hypothesized object rotations for each hypothesis
+            evidence (np.ndarray): Current evidence value for each hypothesis
+            displacement (dict): Sensor displacements for input channels
+            graph_id (str): The ID of the current graph
+            input_channel (str): The channel involved in hypotheses updating.
+
+        Returns:
+            Tuple[np.ndarray, np.ndarray]: Updated sensor locations and evidence values.
+        """
+        evidence_threshold = self._get_evidence_update_threshold(graph_id)
+
+        rotated_displacements = rotations.dot(displacement[input_channel])
+        search_locations = locations + rotated_displacements
+
+        hyp_ids_to_test = np.where(evidence >= evidence_threshold)[0]
+        num_hypotheses_to_test = hyp_ids_to_test.shape[0]
+        if num_hypotheses_to_test > 0:
+            new_evidence = self._calculate_evidence_for_new_locations(
+                graph_id,
+                input_channel,
+                search_locations[hyp_ids_to_test],
+                rotations[hyp_ids_to_test],
+                features,
+            )
+            min_update = np.clip(np.min(new_evidence), 0, np.inf)
+            evidence_to_add = np.ones_like(evidence) * min_update
+            evidence_to_add[hyp_ids_to_test] = new_evidence
+            evidence = (
+                evidence * self.past_weight + evidence_to_add * self.present_weight
+            )
+        return search_locations, evidence
+
+    def _replace_hypotheses_in_hpspace(
+        self,
+        graph_id: str,
+        input_channel: str,
+        new_loc_hypotheses: np.ndarray,
+        new_rot_hypotheses: np.ndarray,
+        new_evidence: np.ndarray,
+    ) -> None:
+        """Updates the hypothesis space for a given input channel in a graph.
+
+        Args:
+            graph_id (str): The ID of the current graph to update.
+            input_channel (str): Channel's name involved in updating the space
+            new_loc_hypotheses (np.ndarray): New sensor locations hypotheses
+            new_rot_hypotheses (np.ndarray): New object poses hypotheses
+            new_evidence (np.ndarray): New evidence values for the input channel
+        """
+        # Add a new channel to the mapping if the hypotheses space doesn't exist
+        if input_channel not in self.channel_hypothesis_mapping[graph_id].channels:
+            self.possible_locations[graph_id] = np.array(new_loc_hypotheses)
+            self.possible_poses[graph_id] = np.array(new_rot_hypotheses)
+            self.evidence[graph_id] = np.array(new_evidence)
+
+            self.channel_hypothesis_mapping[graph_id].add_channel(
+                input_channel, len(new_evidence)
+            )
+            return
+
+        channel_start_ix, channel_end_ix = self.channel_hypothesis_mapping[
+            graph_id
+        ].channel_range(input_channel)
+        channel_size = self.channel_hypothesis_mapping[graph_id].channel_size(
+            input_channel
+        )
+
+        # This concatenation operation for location, poses and evidence replaces
+        # the existing hypothesis space. The new hypothesis space in inserted
+        # in place of the existing hypothesis space by referencing the channel
+        # start and end indices.
+        self.possible_locations[graph_id] = np.concatenate(
+            [
+                self.possible_locations[graph_id][:channel_start_ix],
+                np.array(new_loc_hypotheses),
+                self.possible_locations[graph_id][channel_end_ix:],
+            ]
+        )
+
+        self.possible_poses[graph_id] = np.concatenate(
+            [
+                self.possible_poses[graph_id][:channel_start_ix],
+                np.array(new_rot_hypotheses),
+                self.possible_poses[graph_id][channel_end_ix:],
+            ]
+        )
+
+        self.evidence[graph_id] = np.concatenate(
+            [
+                self.evidence[graph_id][:channel_start_ix],
+                np.array(new_evidence),
+                self.evidence[graph_id][channel_end_ix:],
+            ]
+        )
+
+        self.channel_hypothesis_mapping[graph_id].resize_channel_by(
+            input_channel, len(new_evidence) - channel_size
         )
 
     def _update_evidence_with_vote(self, state_votes, graph_id):
@@ -1146,11 +1167,11 @@ class EvidenceGraphLM(GraphLM):
 
     def _calculate_evidence_for_new_locations(
         self,
-        graph_id,
-        input_channel,
-        search_locations,
-        features,
-        hyp_ids_to_test,
+        graph_id: str,
+        input_channel: str,
+        search_locations: np.ndarray,
+        rotations: np.ndarray,
+        features: Dict,
     ):
         """Use search locations, sensed features and graph model to calculate evidence.
 
@@ -1173,7 +1194,7 @@ class EvidenceGraphLM(GraphLM):
 
         pose_transformed_features = rotate_pose_dependent_features(
             features[input_channel],
-            self.possible_poses[graph_id][hyp_ids_to_test],
+            rotations,
         )
         # Get max_nneighbors nearest nodes to search locations.
         nearest_node_ids = self.get_graph(
--- a/src/tbp/monty/frameworks/models/evidence_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching.py
@@ -990,7 +990,7 @@ class EvidenceGraphLM(GraphLM):
                     )
                 )
 
-            self._replace_hypotheses_in_hpspace(
+            self._set_hypotheses_in_hpspace(
                 graph_id=graph_id,
                 input_channel=input_channel,
                 new_location_hypotheses=channel_possible_locations,
@@ -1083,7 +1083,7 @@ class EvidenceGraphLM(GraphLM):
             )
         return search_locations, channel_hypotheses_evidence
 
-    def _replace_hypotheses_in_hpspace(
+    def _set_hypotheses_in_hpspace(
         self,
         graph_id: str,
         input_channel: str,
--- a/src/tbp/monty/frameworks/models/evidence_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching.py
@@ -919,9 +919,6 @@ class EvidenceGraphLM(GraphLM):
             features (dict): input features
             displacements (dict or None): given displacements
             graph_id (str): identifier of the graph being updated
-
-        Raises:
-            ValueError: If no input channels are found to initializing hypotheses
         """
         start_time = time.time()
 
@@ -941,17 +938,11 @@ class EvidenceGraphLM(GraphLM):
         ]
 
         if len(input_channels_to_use) == 0:
-            if displacements is None:
-                # QUESTION: Do we just want to continue until we get input?
-                raise ValueError(
-                    "No input channels found to initializing hypotheses. Make sure"
-                    " there is at least one channel that is also stored in the graph."
-                )
-            else:
-                logging.info(
-                    f"No input channels observed for {graph_id} that are stored in . "
-                    "the model. Not updating evidence."
-                )
+            logging.info(
+                f"No input channels observed for {graph_id} that are stored in . "
+                "the model. Not updating evidence."
+            )
+            return
 
         for input_channel in input_channels_to_use:
             # Extract channel mapper
@@ -1002,8 +993,8 @@ class EvidenceGraphLM(GraphLM):
             self._replace_hypotheses_in_hpspace(
                 graph_id=graph_id,
                 input_channel=input_channel,
-                new_locations_hypotheses=channel_possible_locations,
-                new_poses_hypotheses=channel_possible_poses,
+                new_location_hypotheses=channel_possible_locations,
+                new_pose_hypotheses=channel_possible_poses,
                 new_evidence=channel_hypotheses_evidence,
             )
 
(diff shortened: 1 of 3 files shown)


src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -898,182 +899,88 @@ def _update_possible_matches(self, query):
...
-                        new_rot_hypotheses=initial_possible_channel_rotations,
-                        new_evidence=channel_evidence,
-                    )
+            # === RE-BUILD HYPOTHESIS SPACE ===
+            self._replace_hypotheses_in_hpspace(
+                graph_id=graph_id,
+                input_channel=input_channel,
+                new_loc_hypotheses=locations,
+                new_rot_hypotheses=rotations,
+                new_evidence=evidence,
```
vkakerbeck, line 981: why are these all called `new_`? shouldn't the output be new hypotheses, not the input?

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1082,6 +989,120 @@ def _update_evidence(self, features, displacements, graph_id):
...
+        self,
+        features: Dict,
+        locations: np.ndarray,
+        rotations: np.ndarray,
+        evidence: np.ndarray,
+        displacement: Dict,
+        graph_id: str,
+        input_channel: str,
+    ) -> Tuple[np.ndarray, np.ndarray]:
+        """Updates existing hypotheses by displacing them.
```
vkakerbeck, line 1002:
```suggestion
        """Updates evidences by comparing features after applying the sensed displacement.
```

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1082,6 +989,120 @@ def _update_evidence(self, features, displacements, graph_id):
...
+                rotations[hyp_ids_to_test],
+                features,
+            )
+            min_update = np.clip(np.min(new_evidence), 0, np.inf)
+            evidence_to_add = np.ones_like(evidence) * min_update
+            evidence_to_add[hyp_ids_to_test] = new_evidence
+            evidence = (
+                evidence * self.past_weight + evidence_to_add * self.present_weight
+            )
+        return search_locations, evidence
```
vkakerbeck, line 1037: Where did all the comments go? I think this function requires a bit more explanation, both in the comments and in the doc string.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1082,6 +989,120 @@ def _update_evidence(self, features, displacements, graph_id):
...
+        new_evidence: np.ndarray,
+    ) -> None:
+        """Updates the hypothesis space for a given input channel in a graph.
+
+        Args:
+            graph_id (str): The ID of the current graph to update.
+            input_channel (str): Channel's name involved in updating the space
+            new_loc_hypotheses (np.ndarray): New sensor locations hypotheses
+            new_rot_hypotheses (np.ndarray): New object poses hypotheses
+            new_evidence (np.ndarray): New evidence values for the input channel
```
vkakerbeck, line 1054: Similar to my comment earlier, it would be useful to define what "New" means here. I assume it means after the movement was already applied and evidence updated

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1082,6 +989,120 @@ def _update_evidence(self, features, displacements, graph_id):
...
+        """
+        # Add a new channel to the mapping if the hypotheses space doesn't exist
+        if input_channel not in self.channel_hypothesis_mapping[graph_id].channels:
```
vkakerbeck, line 1057: this seems like it should happen outside `_replace_hypotheses_in_hpspace`. Why are you not using the `_get_initial_hypothesis_space` function?

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1082,6 +989,120 @@ def _update_evidence(self, features, displacements, graph_id):
...
+        # the existing hypothesis space. The new hypothesis space in inserted
+        # in place of the existing hypothesis space by referencing the channel
+        # start and end indices.
+        self.possible_locations[graph_id] = np.concatenate(
+            [
+                self.possible_locations[graph_id][:channel_start_ix],
+                np.array(new_loc_hypotheses),
+                self.possible_locations[graph_id][channel_end_ix:],
+            ]
+        )
```
vkakerbeck, line 1084: I'm not sure I understand what is happening here. Why are you calling concatenate with the old possible_locations? Why can't we do what we used to do (`self.possible_locations[graph_id][channel_start:channel_end] = new_loc_hypotheses`)?

vkakerbeck (commented):
Nice, thanks for taking a look at cleaning this up a bit. I left some comments and questions. Mostly around changed terminology and variable names and comments, logs and error raises that seem to be missing in the new version.

nielsleadholm:
Just fyi I won't be able to review this given lack of time with travel, but I'm sure it is great and thanks for implementing it.

src/tbp/monty/frameworks/models/evidence_matching.py
tristanls: **note**: I think it may be more straightforward to understand what's happening here if the channel is resized before being overwritten.
```python
self.channel_hypothesis_mapping[graph_id].resize_channel_by(
    input_channel, len(new_evidence) - channel_size
)
```
Then, retrieve the indices:
```python
channel_start_ix, channel_end_ix = self.channel_hypothesis_mapping[
    graph_id
].channel_range(input_channel)
```
Then, can use the more straightforward previous assignment:
```python
self.possible_locations[graph_id][channel_start_ix:channel_end_ix] = new_loc_hypotheses
```

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1082,6 +989,120 @@ def _update_evidence(self, features, displacements, graph_id):
...
+            self.possible_locations[graph_id] = np.array(new_loc_hypotheses)
+            self.possible_poses[graph_id] = np.array(new_rot_hypotheses)
+            self.evidence[graph_id] = np.array(new_evidence)
+
+            self.channel_hypothesis_mapping[graph_id].add_channel(
+                input_channel, len(new_evidence)
+            )
+            return
+
+        channel_start_ix, channel_end_ix = self.channel_hypothesis_mapping[
```
tristanls, line 1067: **issue**: As I mentioned elsewhere, I think this replacement logic is better placed in the `ChannelMapper`.

**suggestion**: Create a `ChannelMapper` method that does this work for you. Something like:

```python
def update(self, mapping, channel, data):
    start, end = self.channel_range(channel)
    mapping[start:end] = data
```

Then, this whole code would turn into:
```python
mapper.resize_channel_by(
    input_channel, len(new_evidence) - channel_size
)
mapper.update(self.possible_locations[graph_id], input_channel, np.array(new_loc_hypothesis))
mapper.update(self.possible_poses[graph_id], input_channel, np.array(new_rot_hypothesis))
mapper.update(self.evidence[graph_id], input_channel, np.array(new_evidence))
```

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -54,6 +54,20 @@ def total_size(self) -> int:
         """
         return sum(self.channel_sizes.values())
 
+    def channel_size(self, channel_name: str) -> int:
```
tristanls, line 57: **issue**: Not needed if you implement my suggestion of `extract` and `update` methods.

tristanls (commented):
**issue**: This is a very verbose way to keep accessing the same `ChannelMapper`:
```python
self.channel_hypothesis_mapping[graph_id]
```
**suggestion**: Assign to a variable and reuse.
```python
```

src/tbp/monty/frameworks/models/evidence_matching.py
tristanls: **update**: I see now that what I suggested will not work since the data in `self.possible_locations`, `self.possible_rotations`, and `self.evidence` is not changing when `ChannelMapper` ranges change.

src/tbp/monty/frameworks/models/evidence_matching.py
tristanls: **update**: I see now that `self.possible_locations`, `self.possible_poses`, and `self.evidence` data is not updated when `ChannelMapper` is updated.

**suggestion**: Instead of the `update` function I suggested before, perhaps this version then:
```python
def update(self, mapping, channel, data):
    start, end = self.change_range(channel)
    return np.concatenate(mapping[:start], data, mapping[end:])
```

With the resize at the end as you have it:
```python
self.possible_locations[graph_id] = mapper.update(
    self.possible_locations[graph_id], input_channel, np.array(new_loc_hypothesis)
)
self.possible_poses[graph_id] = mapper.update(
    self.possible_poses[graph_id], input_channel, np.array(new_rot_hypothesis)
)
self.evidence[graph_id] = mapper.update(
    self.evidence[graph_id], input_channel, np.array(new_evidence)
)
mapper.resize_channel_to(input_channel, len(new_evidence))
```

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1173,7 +1194,7 @@ def _calculate_evidence_for_new_locations(
 
         pose_transformed_features = rotate_pose_dependent_features(
             features[input_channel],
-            self.possible_poses[graph_id][hyp_ids_to_test],
+            rotations,
```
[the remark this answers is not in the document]
ramyamounir: I'm honestly not quite sure what to use here. I've seen both being used interchangeably throughout the class. For example, in `_update_evidence` [here](https://github.com/thousandbrainsproject/tbp.monty/blob/2aefc52d2fe1230eae4c2cb92937feca1e017bbe/src/tbp/monty/frameworks/models/evidence_matching.py#L965), the output of the `_get_initial_hypothesis_space` are rotations but they are added to `possible_poses`. My thought process here was that a subset of `self.possible_poses` would be referred to as rotations.(?)

Should we pick one and use it for everything?

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -898,182 +899,88 @@ def _update_possible_matches(self, query):
...
+                rotations = self.possible_poses[graph_id][
+                    channel_start_ix:channel_end_ix
+                ]
+                evidence = self.evidence[graph_id][channel_start_ix:channel_end_ix]
+
+                # We only displace existing hypotheses since the newly sampled
+                # hypotheses should not be affected by the displacement from the last
+                # sensory input.
+                if not initialize_hyp_space:
+                    locations, evidence = self._displace_hypotheses(
```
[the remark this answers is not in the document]
ramyamounir: Is `_displace_hypotheses_and_compute_evidence` too long? I'm not very creative with variables and functions names :sweat_smile:

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: This function has a similar role to `_add_hypotheses_to_hpspace` [here](https://github.com/thousandbrainsproject/tbp.monty/blob/2aefc52d2fe1230eae4c2cb92937feca1e017bbe/src/tbp/monty/frameworks/models/evidence_matching.py#L1003), that's where I borrowed `new_`. It makes sense to me, these are the newly displaced or initialized hypotheses and we want to add them to the hypothesis space, or use them to replace existing ones. The function directly modifies the hypothesis space, therefore has no output.

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: I've added a note in the docstring.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -898,182 +899,88 @@ def _update_possible_matches(self, query):
...
+                ]
+                rotations = self.possible_poses[graph_id][
+                    channel_start_ix:channel_end_ix
+                ]
+                evidence = self.evidence[graph_id][channel_start_ix:channel_end_ix]
+
+                # We only displace existing hypotheses since the newly sampled
+                # hypotheses should not be affected by the displacement from the last
+                # sensory input.
+                if not initialize_hyp_space:
```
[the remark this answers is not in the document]
ramyamounir: Good catch, an obvious blunder..

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: I've added back the comments and some description in the docstring.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -898,182 +899,88 @@ def _update_possible_matches(self, query):
...
-        if first step (not moved yet):
-            - get initial hypothesis space using observed pose features
-            - initialize evidence for hypotheses using first observed features
-        else:
-            - update evidence for hypotheses using all observed features
-            and displacement
+    def _update_evidence(
+        self,
+        features: Dict,
+        displacements: Optional[Dict],
```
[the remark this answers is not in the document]
ramyamounir: Cool, thanks!

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -898,182 +899,88 @@ def _update_possible_matches(self, query):
...
+
+            # Initialize a new hypothesis space using graph nodes
+            if initialize_hyp_space:
                 (
-                    initial_possible_channel_locations,
-                    initial_possible_channel_rotations,
-                    channel_evidence,
+                    locations,
+                    rotations,
+                    evidence,
```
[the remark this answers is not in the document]
ramyamounir: Is `possible_locations`, `possible_rotations`, `evidence` better? Also would you rather have it be `possible_poses` instead of rotations? I'm worried they might get confused with `self.possible_locations`, `self.possible_poses`, `self.evidence`.

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: I'm using the `_get_initial_hypothesis_space` [here](https://github.com/thousandbrainsproject/tbp.monty/blob/254a2f433c7b637889c38d1179670aa96b254994/src/tbp/monty/frameworks/models/evidence_matching.py#L945) to get the `new_` space.

This logic of updating the hypothesis space is done at different places in the original code. There is the code [here](https://github.com/thousandbrainsproject/tbp.monty/blob/2aefc52d2fe1230eae4c2cb92937feca1e017bbe/src/tbp/monty/frameworks/models/evidence_matching.py#L958) for adding newly initialized hypotheses, then there is this [function](https://github.com/thousandbrainsproject/tbp.monty/blob/2aefc52d2fe1230eae4c2cb92937feca1e017bbe/src/tbp/monty/frameworks/models/evidence_matching.py#L1003) for adding newly initialized hypotheses (if displacement is not None). 

I'm trying to put all of this logic in one function for simplicity. The new function does the following:
- If the hypothesis space does not exist for any input channel, a new one is initialized
- If the hypothesis space only exists for other channels, a new channel is created with the mean evidence scores of the existing channels
- If the hypothesis space exists for the given input channel, the new space replaces the existing hypothesis space.

I've added this description to the docstring of the function.

ramyamounir:
Thanks for the detailed reviews @vkakerbeck and @tristanls! I think I've addressed most of your comments except for some pending ones.

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: This is great, much cleaner! I've added the update and extract functions, as well as some more unit tests for those. Thanks.

src/tbp/monty/frameworks/models/evidence_matching.py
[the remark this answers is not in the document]
vkakerbeck: How about `possible_channel_locations` and `possible_channel_poses` and `channel_hypotheses_evidence`? That way it expresses that these are just for a particular channel and that they are our hypotheses. I don't have a strong opinion on rotations vs. poses in general but I think it is better to be consistent across the code. If we call it poses in the class variables it would be best if the temporary variable names correspond. Another unofficial convention I used (when talking about this stuff) is that I call these "orientations" instead of "rotations" and then use the term "rotation" when we apply an RF transform.

src/tbp/monty/frameworks/models/evidence_matching.py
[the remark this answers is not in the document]
vkakerbeck: no I think this is fine and more expressive. It also points to an issue that these should probably be two separate functions (indicated by the "and" in the function name). But could be something for a future refactor.

src/tbp/monty/frameworks/models/evidence_matching.py
vkakerbeck: okay sounds good. As long as that is maybe explained a bit more in the function doc string

src/tbp/monty/frameworks/models/evidence_matching.py
vkakerbeck: okay, maybe you could just add a comment to it that explains this a bit. Maybe with same shape variables and how they correspond (especially in the case where the hypothesis space changed shape)

vkakerbeck:
Nice! All looks great so far and also nice suggestions @tristanls :) I replied to some of the open comments and will do a full re-review once those are incorporated.

src/tbp/monty/frameworks/models/evidence_matching.py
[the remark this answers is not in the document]
ramyamounir: Yeah, nice. I've renamed them to `channel_possible_locations`, `channel_possible_poses` and `channel_hypotheses_evidence` just so that they all start with the same prefix `channel_`. I'll stick to poses then instead of rotations.

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: I added some description in the ChannelMapper update function (where this logic happens now). Also, the added unit tests makes it clearer what is going on. Please let me know if more description is needed.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -898,182 +901,112 @@ def _update_possible_matches(self, query):
...
-        if first step (not moved yet):
-            - get initial hypothesis space using observed pose features
-            - initialize evidence for hypotheses using first observed features
-        else:
-            - update evidence for hypotheses using all observed features
-            and displacement
+    def _update_evidence(
+        self,
+        features: Dict,
+        displacements: Dict | None,
```
tristanls, line 907: **note**: With `from __future__ import annotations` in place, this can become `dict` instead of `typing.Dict`.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -898,182 +901,112 @@ def _update_possible_matches(self, query):
...
+
+        # Initialize a `ChannelMapper` to keep track of input channel range
+        # of hypotheses for a specific graph_id
+        if graph_id not in self.channel_hypothesis_mapping:
+            self.channel_hypothesis_mapping[graph_id] = ChannelMapper()
+
+        # Get all usable input channels
+        # NOTE: We might also want to check the confidence in the input channel
+        # features. This information is currently not available here. Once we
+        # pull the observation class into the LM we could add this (TODO S).
```
tristanls, line 936: **style (non-blocking)**: I know this is copy-and-pasted, but could you extract into a TODO that starts at the beginning of the line and remove the `S` as we don't assign TODOs. e.g.,

```python
        # NOTE: We might also want to check the confidence in the input channel
        # TODO: Once we pull the observation class into the LM we could add this.
```

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1082,6 +1015,152 @@ def _update_evidence(self, features, displacements, graph_id):
...
+                mapper.add_channel(input_channel, len(new_evidence))
+                return
+
+            # Add current mean evidence to give the new hypotheses a fighting
+            # chance. TODO H: Test mean vs. median here.
+            current_mean_evidence = np.mean(self.evidence[graph_id])
+            new_evidence = new_evidence + current_mean_evidence
+
+        # This mapper.update function replaces the hypothesis space with a new one
+        # at the specified input channel range.
```
tristanls, line 1146: **style**: This feels like it should be part of `mapper.update` docstring and should not be spelled out here.

**suggestion**: I think that a comment that should go here should explain why there are three update calls followed by a resize_channel_to call.

src/tbp/monty/frameworks/utils/evidence_matching.py
tristanls: **todo**: Remove unused code.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -120,6 +152,57 @@ def add_channel(
...
 
+    def extract(self, original: np.ndarray, channel: str) -> np.ndarray:
+        """Extracts the portion of the original array corresponding to a given channel.
+
+        Args:
+            original (np.ndarray): The full hypotheses array across all channels.
+            channel (str): The name of the channel to extract.
+
+        Returns:
+            np.ndarray: The extracted slice of the original array.
```
tristanls, line 163: **suggestion**: It may be helpful to highlight that this is a view, not a copy of the original array.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -120,6 +152,57 @@ def add_channel(
...
+        as such: (original[0:10], data, original[13:]). This will result in an array of
+        the shape (22, 3), i.e., we removed 3 rows and added new 5 rows.
+
+        Args:
+            original (np.ndarray): The original array.
+            channel (str): The name of the input channel.
+            data (np.ndarray): The new data to insert.
+
+        Returns:
+            np.ndarray: The resulting array after insertion.
```
tristanls, line 195: **suggestion**: It may be helpful to highlight that this is a new array and not a view.

```
>>> import numpy as np
>>> a = np.array([1,2])
>>> b = np.array([3,4])
>>> c = np.concatenate((a,b))
>>> c
array([1, 2, 3, 4])
>>> c[1] = 17
>>> c
array([ 1, 17,  3,  4])
>>> a
array([1, 2])
>>> b
array([3, 4])
```

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: I think "S" here possibly stands for sensory module. I've seen this in other places too (e.g., H for hierarchy, L for learning module). Should I start removing those as I see them?

tristanls: I did not know there was a convention. Looks like @vkakerbeck added these back in the day, 👋, what do these mean?

tristanls: From prehistory, a comment on this very... comment:
> -- Just curious, what does TODO S refer to? I was thinking "sensory" refactor but now I'm not so sure.
> -- Oh sorry, it refers to the State class and more generally anything CCP related. In this case it refers to my open TODO of maybe using the State class inside the LM

Then there's also
> -- Maybe add a TODO O ("optimize") to see if this for loop can be vectorized?

Still haven't found this documented anywhere.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1138,12 +1138,12 @@ def _replace_hypotheses_in_hpspace(
...
             # Add current mean evidence to give the new hypotheses a fighting
-            # chance. TODO H: Test mean vs. median here.
+            # chance.
+            # TODO H: Test mean vs. median here.
             current_mean_evidence = np.mean(self.evidence[graph_id])
             new_evidence = new_evidence + current_mean_evidence
 
-        # This mapper.update function replaces the hypothesis space with a new one
-        # at the specified input channel range.
+        # Update the hypothesis space with new locations, poses, and evidence scores
```
tristanls, line 1146: **suggestion**: Sorry, I should have made a better suggestion.

```
# The mapper update function calls below automatically resize the
# arrays they update. Afterward, we must update the channel indices 
# in the mapper via resize_channel_to to stay in sync with 
# the now resized arrays.
```

ramyamounir: Yeah makes sense, thanks.

src/tbp/monty/frameworks/models/evidence_matching.py
vkakerbeck: yes this is not documented anywhere. It was a convention Niels and I started when it was mostly the two of us working on the code. Basically so that we can quickly find all the TODOs related to a specific work package we are working on. I think there is:
M - motor system (niels)
H - heterarchy (me)
S - Pulling the state class into the LM to also be used there (future work I left when adding the State class)

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -898,182 +901,112 @@ def _update_possible_matches(self, query):
...
+        # features. This information is currently not available here.
+        # TODO S: Once we pull the observation class into the LM we could add this.
+        input_channels_to_use = [
+            ic
+            for ic in features.keys()
+            if ic in self.get_input_channels_in_graph(graph_id)
+        ]
+
+        if len(input_channels_to_use) == 0:
+            if displacements is None:
```
vkakerbeck, line 944: Now that I see this logic rewritten it looks kind of strange. I don't see why we would want to differentiate whether the displacement in None or not. I guess because when the displacement is None we got no usable input on the first step? I think it would be fine just to log the "not updating evidence" message in both cases and remove the value error. It should be able to still initialize the hypothesis space if it gets valid input later on, right?

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -898,182 +901,112 @@ def _update_possible_matches(self, query):
...
-                    (
-                        initial_possible_channel_locations,
-                        initial_possible_channel_rotations,
-                        channel_evidence,
-                    ) = self._get_initial_hypothesis_space(
-                        features, graph_id, input_channel
+                # We only displace existing hypotheses since the newly sampled
+                # hypotheses should not be affected by the displacement from the last
+                # sensory input.
+                channel_possible_locations, channel_hypotheses_evidence = (
```
vkakerbeck, line 990: as per your definition of `new_` later, shouldn't these two variables also have the `new_` prefix?

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1082,6 +1015,158 @@ def _update_evidence(self, features, displacements, graph_id):
...
+            - If the hypothesis space only exists for other channels, a new channel is
+                created with the mean evidence scores of the existing channels
+            - If the hypothesis space exists for the given input channel, the new space
+                replaces the existing hypothesis space
+
+        Args:
+            graph_id (str): The ID of the current graph to update.
+            input_channel (str): Channel's name involved in updating the space
+            new_locations_hypotheses (np.ndarray): New sensor locations hypotheses
+            new_poses_hypotheses (np.ndarray): New object poses hypotheses
```
vkakerbeck, line 1118: nit: could we call these `new_location_hypotheses` and `new_pose_hypotheses`? (removing the double plural)

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1082,6 +1015,158 @@ def _update_evidence(self, features, displacements, graph_id):
...
+            current_mean_evidence = np.mean(self.evidence[graph_id])
+            new_evidence = new_evidence + current_mean_evidence
+
+        # The mapper update function calls below automatically resize the
+        # arrays they update. Afterward, we must update the channel indices
+        # in the mapper via resize_channel_to to stay in sync with
+        # the now resized arrays. We do not resize before array updates
+        # because then, during the update, the indices would not correspond
+        # to the data in the arrays.
+        self.possible_locations[graph_id] = mapper.update(
```
vkakerbeck, line 1152: As far as I understand, the current version does not implement adding any new hypotheses into an existing hypothesis space (just adds a new space for a channel if it doesn't exist yet). I know you are planning to add this in your next PR. But would it make sense to omit all the resizing operations here since we can always just replace the values? And then when we actually sample new hypotheses we either add a check whether the size has changed before we resize or have a different version of this function for it? I'm just worried about the cost of all the array initialization and shaping functions when we don't actually need them.

src/tbp/monty/frameworks/models/evidence_matching.py
```diff
@@ -1082,6 +1015,158 @@ def _update_evidence(self, features, displacements, graph_id):
...
+        """Updates the hypothesis space for a given input channel in a graph.
+
+        This function updates the hypothesis space (for a specific graph and input
+        channel) with a new set of locations, rotations and evidence scores.
+            - If the hypothesis space does not exist for any input channel, a new one
+                is initialized
+            - If the hypothesis space only exists for other channels, a new channel is
+                created with the mean evidence scores of the existing channels
+            - If the hypothesis space exists for the given input channel, the new space
+                replaces the existing hypothesis space
```
vkakerbeck, line 1112: nit: two out of those three don't actually replace but initialize. Should those maybe be in a separate function? Or would it be more accurate to call this `_set_hypotheses_in_hpspace`?

vkakerbeck (approved):
Really nice updates and restructuring! All my remaining comments are minor questions/nits or things that could be done in a separate PR so I'm approving.

src/tbp/monty/frameworks/models/evidence_matching.py
tristanls: **nitpick**: As the self-designated `set/get` police, I believe there's usually a more specific name that does not include `set/get`, if we need a different one.

**note**: I think what might alleviate some of the tension about what is happening with hypotheses would be encapsulating them in something. The fact that we are passing around three arrays and doing (transactional) updates to them at each step points to the fact that they are all aspects of _one_ thing. We used to pass around `action, amount`, then combined that into `Action`. It seems we are missing a concept of `Hypothesis` or something adjacent, that would combine the three arrays and clarify the nature of these operations. Additionally, we also have a `HypothesisProcessor` or `...Engine` or something.. this is the mechanism where we stack all the hypotheses so that we can do efficient computations. 

**suggestion (non-blocking)**: (in a different PR), it would probably benefit us if we organized the things that go together. The code would end up something like:

```python
if initialize_hyp_space:
    hypothesis = self._get_initial_hypothesis_space(
        features, graph_id, input_channel
    )
else:
    hypothesis = self.hypothesis_processor.extract(graph_id, input_channel)
    # note that `hypothesis` would have `self.graph_id` and `self.input_channel` on it.
    hypothesis = self._displace_hypotheses_and_compute_evidence(
        features,
        hypothesis,
        displacements
    )
self.hypothesis_processor.update(hypothesis)
```

vkakerbeck: wow yeah that would be a great idea! I guess the main reason why we don't have a list of Hypothesis objects (each containing a location, orientation and evidence value) is because the we couldn't do our matrix operations on the entire hypothesis space anymore. But I still think it could be a good idea to at least bundle the three lists together in one HypothesisSpace object.

Just for my own learning: Is it bad practice to call function set_x and get_x? I always thought it was a nice thing to have these setter and getter function and have a common naming convention for them. Or is it more that functions with those names should do anything in addition to setting and getting?

tristanls: > the main reason why we don't have a list of Hypothesis objects (each containing a location, orientation and evidence value) is because the we couldn't do our matrix operations on the entire hypothesis space anymore

My suggestion would not change how we do the computations. The only difference would be how the code would read as we pass things around and read and write things to our matrix. The reads and writes would go through a HypothesisProcessor/Engine, but would still end up the same way in the same matrix. And what would get passed to and from HypothesisProcessor/Engine would be a Hypothesis, instead of three arrays. This would reduce the cognitive load when reading the code, allowing us to express the algorithm in 12 lines with no comments as I did above.

As for set/get, I remove set/get "noise" words as a naming trick to improve the clarity of method names. I found the code easier to read as a result. https://www.youtube.com/watch?v=OmtET5zNxuY (2m19s)

tristanls: As an example:
```python
if initialize_hyp_space:
    hypothesis = self._get_initial_hypothesis_space(
        features, graph_id, input_channel
    )
```
If that is changed to:
```python
if initialize_hyp_space:
    hypothesis = self._initialize_hypothesis(
        features, graph_id, input_channel
    )
```

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: Yeah makes sense. A log should be fine if we don't have input channels. You did leave a comment above the raised error:
```python
# QUESTION: Do we just want to continue until we get input?
raise ValueError(
      "No input channels found to initializing hypotheses. Make sure"
      " there is at least one channel that is also stored in the graph."
)
```
It makes sense to me that we just continue and not raise the ValueError here.

> It should be able to still initialize the hypothesis space if it gets valid input later on, right?

Yes, it will still initialize later on when it gets input_channels that are in the model.

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: Yes, I really like this. We should have some class for the hypothesis space that encapsulates the 3 arrays and some functionality of updating the space.

ramyamounir: as for avoiding set/get, how about we call this function `_update_hypothesis_space`? It still implies an existing hypothesis space that will be updated, which is not true when initializing, so idk..

This naming issue also reminds me of the name of `_update_evidence`, it says update, but at the beginning of the function we sometimes initialize evidence scores with `_get_initial_hypothesis_space`. I feel that the initialization at the beginning of the function is just handling an edge case but doesn't describe the main logic of the function.

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: I see your point. Do you prefer we implement both ways of updating the space here and having a check that decides which one to run, or overriding this method in the resampling mixin? I feel like the latter might avoid confusion as it ties the array shaping functions to the resampling logic.. What do you think?

src/tbp/monty/frameworks/models/evidence_matching.py
ramyamounir: technically yes, but if I do that I'd have to add `_replace_hypotheses_in_hpspace` to both the if condition cases, once with the prefix and once without.

```python
            # Initialize a new hypothesis space using graph nodes
            if initialize_hyp_space:
                # TODO H: When initializing a hypothesis for a channel later on (if
                # displacement is not None), include most likely existing hypothesis
                # from other channels?
                (
                    channel_possible_locations,
                    channel_possible_poses,
                    channel_hypotheses_evidence,
                ) = self._get_initial_hypothesis_space(
                    features, graph_id, input_channel
                )

                self._replace_hypotheses_in_hpspace(
                    graph_id=graph_id,
                    input_channel=input_channel,
                    new_location_hypotheses=channel_possible_locations,
                    new_pose_hypotheses=channel_possible_poses,
                    new_evidence=channel_hypotheses_evidence,
                )
            # Retrieve existing hypothesis space for a specific input channel
            else:
                channel_possible_locations = mapper.extract(
                    self.possible_locations[graph_id], input_channel
                )
                channel_possible_poses = mapper.extract(
                    self.possible_poses[graph_id], input_channel
                )
                channel_hypotheses_evidence = mapper.extract(
                    self.evidence[graph_id], input_channel
                )

                # sensory input.
                new_channel_possible_locations, new_channel_hypotheses_evidence = (
                    self._displace_hypotheses_and_compute_evidence(
                        features,
                        channel_possible_locations,
                        channel_possible_poses,
                        channel_hypotheses_evidence,
                        displacements,
                        graph_id,
                        input_channel,
                    )
                )

                self._replace_hypotheses_in_hpspace(
                    graph_id=graph_id,
                    input_channel=input_channel,
                    new_location_hypotheses=new_channel_possible_locations,
                    new_pose_hypotheses=channel_possible_poses,
                    new_evidence=new_channel_hypotheses_evidence,
                )
```

I feel it's okay leaving it the way it is because i've defined the `new_` prefix locally within the `_replace_hypotheses_in_hpspace` only in the context of replacing old hypotheses with new ones. We could also remove the prefix from everywhere..

src/tbp/monty/frameworks/models/evidence_matching.py
vkakerbeck: Yes I would also tend towards the second option. If it's easy to overwrite with the mixin you are working on that could be the cleanest way.

src/tbp/monty/frameworks/models/evidence_matching.py
vkakerbeck: no that is perfect

src/tbp/monty/frameworks/models/evidence_matching.py
vkakerbeck: I agree, especially since there is already this comment  this is good to remove.

src/tbp/monty/frameworks/models/evidence_matching.py
vkakerbeck: okay, that's fine too

src/tbp/monty/frameworks/models/evidence_matching.py
vkakerbeck: I agree that it would be great to have a class that bundles those three lists. I don't want to draw this conversation out too much here since this should be its own PR anyway and also don't want to do too much bikeshedding about naming stuff, so excuse me for adding some more of my humble opinions here :)

To me, having get and set in the name is useful because it tells me what to expect from the function. If it starts with get, I expect it to return whatever the remaining function name is. If it starts with set, I expect it to modify something that is named in the rest of the function name. I see the point of the video that "get" is not just retrieving but also changing a state but to me that is still compatible with how I use "get". If I just read `self._initialize_hypotheses` I expect it to change some internal state of the class, not necessarily to return that state. But maybe that's just me and how I've been using those conventions so far.
On a minor note, this function does not just initialize/get one hypothesis, it initializes many, which is why we should use plural (`hypotheses = ...`) or stay with `hypothesis_space`.

ramyamounir:
Thanks for your reviews and additional comments, @vkakerbeck  and @tristanls . One thing to add is that i've removed the now unused function `_add_hypotheses_to_hpspace` since this functionality was replaced with `_set_hypotheses_in_hpspace`. 

I ran benchmark comparisons for two experiments and the diffs between my PR and the main branch are shown below:

| Metric   |   main |   restructure_update_evidence |   Δ (delta) |
|----------|--------|-------------------------------|-------------|
| time     |  2.007 |                        2.0194 |      0.0124 |

📌 Differences found in experiment: randrot_noise_10distinctobj_surf_agent
| Metric   |   main |   restructure_update_evidence |   Δ (delta) |
|----------|--------|-------------------------------|-------------|
| time     | 1.8718 |                        1.8723 |      0.0005 |

If all looks good, I'll be merging this PR today.

PR #305, opened 2025-05-09
fix: SurfacePolicy touch_object and dynamic_call no longer interfere

tristanls:
This pull request removes the interference between SurfacePolicy's `dynamic_call` implementation and its embedded `touch_object` positioning procedure, which finds the object if the sensor loses it. Additionally, it resets the `touch_object` state between each object finding attempt. Previously, the `touch_object` procedure would resume from the random state it ended in from the previous attempt to find the object.

The `touch_object` state reset is done by having the data loader reset the `touch_object` positioning procedure on success. This is not great, but for now, `touch_object` might have succeeded and not know it. In the future, when it is refactored to the `TouchObject` positioning procedure, the reset will be contained within the positioning procedure itself.

The interference is removed by having `touch_object` no longer save its actions in `SurfacePolicy`'s `self.action`.

The test is updated with the example sequence of new expected behavior, demonstrating non-interference and resetting the `touch_object` state between each object finding attempt.

Lastly, return types of related methods are made specific instead of generic per best practice.

## Benchmark results
Multiple decimal places results here, the table updates are in the benchmark result changeset

| Code | Experiment | Correct (%) | Used MLH (%) | Num Match Steps | Rotation Error | WandB |
|---|---|---|---|---|---|---|
| new | base_config_10distinctobj_surf_agent | 100 | 0 | 27.642857142857142 | 0.20275 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/4d51r1go/overview)
| new | randrot_noise_10distinctobj_surf_agent | 100 | 1 | 28.46 | 0.3758580000000001 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/rhddvorw/overview)
| new | randrot_10distinctobj_surf_agent | 100 | 1 | 28.08 | 0.352194 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/bn6cwftb/overview)
| new | base_10simobj_surf_agent | 94.28571428571428 | 11.428571428571429 | 86.30714285714286 | 0.24581287878787875 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/elslasu1/overview)
| new | randrot_noise_10simobj_surf_agent | 93 | 36 | 181.35 | 0.5434129032258065 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/8dr5cmeu/overview)
[...]

PR #306, opened 2025-05-13
refactor!: extracted TouchObject positioning procedure from SurfacePolicy

tristanls:
This pull request extracts the `touch_object` positioning procedure from `SurfacePolicy` into the `TouchObject` positioning procedure. Now, when `SurfacePolicy.dynamic_call` cannot see the object, it raises `ObjectNotVisible`, and the data loader uses that to invoke the `TouchObject` positioning procedure.

Notably, this doesn't feel right as we use a _privileged_ (due to `"view_finder"` use) positioning procedure to keep a sensor in touch with the object. The future solution here will be to implement a similar touch object logic internally as part of the `SurfacePolicy`. This may be easier after refactoring the main Monty loop between the data loader, model, and the motor system to be more reinforcement-learning-like.

## Benchmark results
Multiple decimal places results here; the table updates are in the benchmark result changeset.

| Code | Experiment | Correct (%) | Used MLH (%) | Num Match Steps | Rotation Error | WandB |
|---|---|---|---|---|---|---|
| new | base_config_10distinctobj_surf_agent | 100 | 2.142857142857143 | 27.314285714285713 | 0.3148828571428572 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/wdnpj6fv/overview)
| recent (#305) | base_config_10distinctobj_surf_agent | 100 | 0 | 27.642857142857142 | 0.20275 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/4d51r1go/overview)
| new | randrot_noise_10distinctobj_surf_agent | 100 | 4 | 27.69 | 0.543061 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/refoixg5/overview)
| recent (#305) | randrot_noise_10distinctobj_surf_agent | 100 | 1 | 28.46 | 0.3758580000000001 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/rhddvorw/overview)
| new | randrot_10distinctobj_surf_agent | 100 | 2 | 27.11 | 0.33348799999999995 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/5jfg1hmz/overview)
| recent (#305) | randrot_10distinctobj_surf_agent | 100 | 1 | 28.08 | 0.352194 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/bn6cwftb/overview)
| new | base_10simobj_surf_agent | 95 | 10.714285714285714 | 54.628571428571426 | 0.2701473684210527 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/hvye109g/overview)
| recent (#305) | base_10simobj_surf_agent | 94.28571428571428 | 11.428571428571429 | 86.30714285714286 | 0.24581287878787875 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/elslasu1/overview)
| new | randrot_noise_10simobj_surf_agent | **gets stuck on step 65**
| recent (#305) | randrot_noise_10simobj_surf_agent | 93 | 36 | 181.35 | 0.5434129032258065 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/8dr5cmeu/overview)
| new | randomrot_rawnoise_10distinctobj_surf_agent | 68 | 72 | 15.35 | 1.5594602941176472 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/xwsio1sx/overview)
| recent (#305) | randomrot_rawnoise_10distinctobj_surf_agent | 69 | 74 | 15.19 | 1.945234782608695 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/rxy360ug/overview)
| new | base_77obj_surf_agent | 98.26839826839829 | 8.658008658008658 | 42.70995670995671 | 0.16050044052863435 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/h0w6zdvu/overview)
| recent (#305) | base_77obj_surf_agent | 99.13419913419914 | 7.35930735930736 | 57.064935064935064 | 0.2236707423580786 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/47nc07bu/overview)
| new | randrot_noise_77obj_surf_agent | _pending_
| recent (#305) | randrot_noise_77obj_surf_agent | 93.5064935064935 | 25.108225108225103 | 122.70995670995671 | 0.6740199074074075 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/t7te0nhh/overview)
| new | unsupervised_inference_distinctobj_surf_agent | _pending_
| recent (#305) | unsupervised_inference_distinctobj_surf_agent | 3.488372093023256 | 40.69767441860465 | 100 | 1.5507666666666668 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/4ffdsqfb/overview)

| Code | Experiment | Correct - 1st Epoch (%) | Correct - >1st Epoch (%) | Mean Objects per Graph | Mean Graphs per Object |
|---|---|---|---|---|---|
| new | surf_agent_unsupervised_10distinctobj | _pending_
| recent (#305) | surf_agent_unsupervised_10distinctobj | 80.00 | 87.78 | 1.22 | 1.1
| new | surf_agent_unsupervised_10distinctobj_noise | _pending_
| recent (#305) | surf_agent_unsupervised_10distinctobj_noise | 70.00 | 72.22 | 1.07 | 1.75
| new | surf_agent_unsupervised_10simobj | _pending_
| recent (#305) | surf_agent_unsupervised_10simobj | 40.00 | 80.00 | 2.33 | 1.4

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,7 +148,7 @@ show_contexts = true
 [tool.mypy]
 warn_unused_configs = true
 # TODO: This effectively disables mypy, don't ignore once types are added
-# ignore_errors = true
+ignore_errors = true
 # TODO: Remove global ignore and handle missing type stubs
 ignore_missing_imports = true
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,7 +148,7 @@ show_contexts = true
 [tool.mypy]
 warn_unused_configs = true
 # TODO: This effectively disables mypy, don't ignore once types are added
-ignore_errors = true
+# ignore_errors = true
 # TODO: Remove global ignore and handle missing type stubs
 ignore_missing_imports = true
 
@@ -227,6 +227,7 @@ ignore = [
     "ISC001", # ISC001: Implicitly concatenated string literals on one line
     "ISC003", # ISC003: Explicitly concatenated string should be implicitly concatenated
     "N804", # N804: First argument of a class method should be named `cls`
+    "N818", # N818: Exception name {exception} should be named with an Error suffix
     "NPY002", # NPY002: Replace legacy `np.random.normal` call with `np.random.Generator`
     "PD002", # PD002: `inplace=True` should be avoided; it has inconsistent behavior
     "PD901", # PD901: Avoid using the generic variable name `df` for DataFrames
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -502,8 +502,6 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
                 isinstance(self.motor_system._policy, SurfacePolicy)
                 and self._action.name != OrientVertical.action_name()
             ):
-                # We are not attempting to find the object, which means that we
-                # are executing the SurfacePolicy.dynamic_call action cycle.
                 # Out of the four actions in the
                 # MoveForward->OrientHorizontal->OrientVertical->MoveTangentially
                 # "subroutine" defined in SurfacePolicy.dynamic_call, we only
@@ -516,7 +514,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
 
             self.motor_system._state = motor_system_state
 
-            self._counter += 1  # TODO clean up incrementing of counter
+            self._counter += 1
 
             return self._observation
 
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -29,7 +29,10 @@ from tbp.monty.frameworks.actions.actions import (
 from tbp.monty.frameworks.models.motor_policies import (
     GetGoodView,
     InformedPolicy,
+    ObjectNotVisible,
+    PositioningProcedure,
     SurfacePolicy,
+    TouchObject,
 )
 from tbp.monty.frameworks.models.motor_system import MotorSystem
 from tbp.monty.frameworks.models.motor_system_state import (
@@ -459,59 +462,57 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
 
         # NOTE: terminal conditions are now handled in experiment.run_episode loop
         else:
-            self._action = self.motor_system()
-            attempting_to_find_object = False
-
-            # If entirely off object, use vision (i.e. view-finder)
-            # TODO refactor so that this check is done in the motor-policy, and we
-            # update the constraint separately/appropriately; i.e. the below
-            # code should be as general as possible
-            if (
-                isinstance(self.motor_system._policy, SurfacePolicy)
-                and self._action is None
-            ):
-                attempting_to_find_object = True
-                self._action = self.motor_system._policy.touch_object(
-                    self._observation,
-                    view_sensor_id="view_finder",
-                    state=self.motor_system._state,
-                )
-            else:
-                # TODO: Encapsulate this reset inside TouchObject positioning
-                #       procedure once it exists.
-                #       This is a hack to reset the current touch_object
-                #       positioning procedure state so that the next time
-                #       SurfacePolicy falls off the object, it will try to find
-                #       the object using its full repertoire of actions.
-                self.motor_system._policy.touch_search_amount = 0
+            try:
+                self._action = self.motor_system()
+            except ObjectNotVisible:
+                # TODO: Note that a PositioningProcedure is not a MotorPolicy, so
+                #       using it here to keep the agent in touch with the surface is
+                #       questionable. However, in order for the SurfacePolicy to
+                #       take over this responsibility, the main MontyExperiment loop
+                #       between Monty, MotorSystem, and DataLoaders needs to be updated
+                #       first. Once invoking the MotorSystem includes `observation` as
+                #       an argument, e.g. motor_system(observation), then the
+                #       SurfacePolicy will be capable of implementing logic similar to
+                #       the TouchObject positioning procedure instead of raising
+                #       ObjectNotVisible and relying on the dataloader to handle
+                #       this.
+                if isinstance(self.motor_system._policy, SurfacePolicy):
+                    self.touch_object(
+                        sensor_id="view_finder",
+                        desired_object_distance=self.motor_system._policy.desired_object_distance,
+                    )
+                    # TODO: Remove this short-circuit once SurfacePolicy is capable
+                    #       of receiving updated observation via
+                    #       `motor_system(observation)`. For now, we are relying
+                    #       on the experiment to check for motor_only_step and then
+                    #       to `pass_features_directly_to_motor_system(observation)`
+                    #       in order to pass the updated observation to the motor
+                    #       system policy.
+                    self.motor_system._state[self.motor_system._policy.agent_id][
+                        "motor_only_step"
+                    ] = True
+                    return self._observation
 
             self._observation, proprioceptive_state = self.dataset[self._action]
             motor_system_state = MotorSystemState(proprioceptive_state)
 
             # TODO: Refactor this so that all of this is contained within the
-            #       SurfacePolicy and/or positioning procedure.
-            if isinstance(self.motor_system._policy, SurfacePolicy):
-                # When we are attempting to find the object, we are always performing
-                # a motor-only step.
+            #       SurfacePolicy.
+            if (
+                isinstance(self.motor_system._policy, SurfacePolicy)
+                and self._action.name != OrientVertical.action_name()
+            ):
+                # We are not attempting to find the object, which means that we
+                # are executing the SurfacePolicy.dynamic_call action cycle.
+                # Out of the four actions in the
+                # MoveForward->OrientHorizontal->OrientVertical->MoveTangentially
+                # "subroutine" defined in SurfacePolicy.dynamic_call, we only
+                # want to send data to the learning module after taking the
+                # OrientVertical action. The other three actions in the cycle
+                # are motor-only to keep the surface agent on the object.
                 motor_system_state[self.motor_system._policy.agent_id][
                     "motor_only_step"
-                ] = attempting_to_find_object
-
-                if (
-                    not attempting_to_find_object
-                    and self._action.name != OrientVertical.action_name()
-                ):
-                    # We are not attempting to find the object, which means that we
-                    # are executing the SurfacePolicy.dynamic_call action cycle.
-                    # Out of the four actions in the
-                    # MoveForward->OrientHorizontal->OrientVertical->MoveTangentially
-                    # "subroutine" defined in SurfacePolicy.dynamic_call, we only
-                    # want to send data to the learning module after taking the
-                    # OrientVertical action. The other three actions in the cycle
-                    # are motor-only to keep the surface agent on the object.
-                    motor_system_state[self.motor_system._policy.agent_id][
-                        "motor_only_step"
-                    ] = True
+                ] = True
 
             self.motor_system._state = motor_system_state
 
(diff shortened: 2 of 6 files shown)


src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -459,63 +462,59 @@ def __next__(self):
...
-                    # OrientVertical action. The other three actions in the cycle
-                    # are motor-only to keep the surface agent on the object.
-                    motor_system_state[self.motor_system._policy.agent_id][
-                        "motor_only_step"
-                    ] = True
+                ] = True
 
             self.motor_system._state = motor_system_state
 
-            self._counter += 1  # TODO clean up incrementing of counter
```
tristanls, line 518: **note**: I don't think this TODO is valid anymore. We usually increment the counter in the data loader `__next__` method. I don't think there's anything here to "clean up".

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -1105,129 +1244,12 @@ def pre_episode(self):
...
-        )
-
     ###
     # Methods that define behavior of __call__
     ###
     def dynamic_call(
-        self, state: Optional[MotorSystemState] = None
-    ) -> OrientHorizontal | OrientVertical | MoveTangentially | MoveForward | None:
+        self, state: MotorSystemState | None = None
+    ) -> MoveForward | OrientHorizontal | OrientVertical | MoveTangentially | None:
```
vkakerbeck, line 1252: just out of curiosity, what's the reason that these are reordered? Do we follow a specific convention here?

vkakerbeck: Also, now that we raise ObjectNotVisible, is None still a possible output?

tests/unit/policy_test.py
```diff
@@ -894,109 +894,80 @@ def test_surface_policy_moves_back_to_object(self):
 
                 #  Step | Action           | Motor-only? | Obs processed? | Source
                 # ------|------------------|-------------|----------------|-------------
-                #  1    | MoveForward      | True        | False          | dynamic_call
+                #  0    | None             | True        | False          | first_step
+                #  1    | MoveForward      | True        | False          | touch_object
                 #  2    | OrientHorizontal | True        | False          | dynamic_call
-                #  3    | OrientVertical   | False       | True           | dynamic_call
+                #  3    | OrientVertical   | KeyError    | True           | dynamic_call
```
vkakerbeck, line 900: Why is this a key error now?

tests/unit/policy_test.py
```diff
@@ -894,109 +894,80 @@ def test_surface_policy_moves_back_to_object(self):
...
-                #  18   | OrientHorizontal | True        | False          | touch_object
-                #  19   | OrientHorizontal | True        | False          | touch_object
-                #  20   | OrientHorizontal | True        | False          | touch_object
-                #  21   | OrientHorizontal | True        | False          | touch_object
-                #  22   | OrientHorizontal | True        | False          | touch_object
-                #  23   | OrientHorizontal | True        | False          | touch_object
-                #  24   | OrientHorizontal | True        | False          | touch_object
-                #  25   | OrientVertical   | True        | False          | touch_object
+                # dataloder runs TouchObject positioning procedure in a tight loop only
+                # exposing a new observation at the end of the loop
```
vkakerbeck, line 912: does this mean, finding the object is not counted as Monty steps anymore now?

vkakerbeck (commented):
Tagging @nielsleadholm here to be aware of this change

From what I can tell it all looks good but I am not sure I understand why the benchmark performances change. It looks like they are getting worse after this change. Do you know why?

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -768,6 +790,157 @@ def sensor_rotation_relative_to_world(self, state: MotorSystemState) -> Any:
...
+        # The lateral movement will be down relative to the agent's facing position
+        # in the case of orient vertical, and left in the case of orient horizontal
+
+        # The below calculates the exsecant, which provides the necessary
+        # movement to bring the agent back to the same radius around the central
+        # point
+        move_forward_amount = distance_from_center * (
+            (1 - np.cos(np.radians(rotation_degrees)))
+            / np.cos(np.radians(rotation_degrees))
+        )
```
jeremyshoemaker, line 922: If I'm understanding the math right, we're moving a little bit along the tangent line at the point of intersection with the circle, and then moving inward to correct for the mismatch between the circle arc and the tangent line.

Is this more efficient than just rotating around the center of the circle? Seems like a complicated approach to do the same thing.

_I realize this code is just copied from below, but it's a good a time as any to ask these questions_

src/tbp/monty/frameworks/models/motor_policies.py
jeremyshoemaker: It looks like it's possible for `self.get_next_action` to return `None`, so that's probably why it's still here. That does seem like a rare case though, looking at its definition.

tests/unit/policy_test.py
```diff
@@ -894,109 +894,80 @@ def test_surface_policy_moves_back_to_object(self):
...
-                    or 29 <= loader_step <= 42
-                    or 46 <= loader_step <= 58
-                    or loader_step == 62
-                ):
+                # dataloder runs TouchObject positioning procedure in a tight loop only
+                # exposing a new observation at the end of the loop
+                #  28   | MoveForward      | True        | False          | touch_object
+
+                # Motor-only last step of touch_object positioning procedure
+                if loader_step in [1, 13, 18, 23, 28]:
```
jeremyshoemaker, line 943: These fixed numbers seem like they could be brittle. Where do they come from? How are they calculated? What steps would someone who needs to update this have to take to come up with new ones?

jeremyshoemaker (commented):
Seems reasonable but I don't have enough context to confidently approve.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -459,63 +462,59 @@ def __next__(self):
...
+                #       questionable. However, in order for the SurfacePolicy to
+                #       take over this responsibility, the main MontyExperiment loop
+                #       between Monty, MotorSystem, and DataLoaders needs to be updated
+                #       first. Once invoking the MotorSystem includes `observation` as
+                #       an argument, e.g. motor_system(observation), then the
+                #       SurfacePolicy will be capable of implementing logic similar to
+                #       the TouchObject positioning procedure instead of raising
+                #       ObjectNotVisible and relying on the dataloader to handle
+                #       this.
+                if isinstance(self.motor_system._policy, SurfacePolicy):
```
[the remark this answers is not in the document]
tristanls: Great points all around. I am taking advantage of the fact that only a `SurfacePolicy` can throw `ObjectNotVisible` here. But, now that I say that, that makes this instance check extraneous. I'll remove it.

The previous logic of SurfacePolicy relied on returning `None` as the sentinel value to communicate to the data loader to touch the object. Other policies never return `None` for an action. I removed this sentinel value and raised an exception instead. In the future, I expect `SurfacePolicy` to implement its own touch object routine, as I don't think it should ever ask the environment (data loader/dataset) for help.

Line 496 will eventually throw an error, but the reason is arcane. It'll try to access `None.agent_id` in `HabitatActuator.action_name`. Dataset access is defined by implementing the `__getitem__` method, so it wouldn't fail immediately there.

src/tbp/monty/frameworks/models/motor_policies.py
tristanls: The SurfacePolicy action cycle is MoveForward->OrientHorizontal->OrientVertical->MoveTangentially, so I did it to match these. I think alphabetical order may make the most sense, though.

@jeremyshoemaker, you're right that it can still return `None` 😕. I'm gonna remove that and see what happens, as I don't think it should be doing that anymore.

pyproject.toml
```diff
@@ -227,6 +227,7 @@ ignore = [
     "ISC001", # ISC001: Implicitly concatenated string literals on one line
     "ISC003", # ISC003: Explicitly concatenated string should be implicitly concatenated
     "N804", # N804: First argument of a class method should be named `cls`
+    "N818", # N818: Exception name {exception} should be named with an Error suffix
```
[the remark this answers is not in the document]
tristanls: 🙂. I adopted this habit after someone pointed out that nobody will ever be confused when reading code that says `raise DidntWork` instead of `raise DidntWorkError` or in a stack trace when it says `exception: DidntWork` instead of `exception: DidntWorkError`.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -710,7 +732,7 @@ def orient_to_object(
     def positioning_call(
         self,
         observation: Mapping,
-        state: Optional[MotorSystemState] = None,
+        state: MotorSystemState | None = None,
```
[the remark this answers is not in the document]
tristanls: I did it because Ruff told me to do it in another project I'm working on, so I started adopting this habit in our code base. Also, seven chars instead of 10 😬?

I guess it becomes "easier to read" in cases like: 
```python
Optional[MoveForward, MoveTangentially, OrientHorizontal, OrientVertical]
MoveForward | MoveTangentially | OrientHorizontal | OrientVertical | None
```

tests/unit/policy_test.py
tristanls: For others' benefit, we met and determined that finding the object is no longer counted as Monty's steps. I am still in the process of confirming that this is the cause of the observed benchmark differences.

tristanls:
I'm having issues demonstrating why the benchmarks are different. Additionally, `randrot_noise_10simobj_surf_agent` reproducibly gets stuck on step 65 using the new code. So, there's additional work here.

tests/unit/policy_test.py
vkakerbeck: I know this is a bit of a weird edge case but technically, shouldn't it also be True if `"motor_only_step" not in agent_state.keys()` and only be False if agent_state["motor_only_step"] is False? Basically if motor_only_step is not even in the keys then we for sure don't want to process observations in the LM.

tristanls: > if motor_only_step is not even in the keys then we for sure don't want to process observations in the LM

Could you explain more about this assertion? Why must the `motor_only_step` property be defined for the learning module to process anything? I thought we specifically set `motor_only_step` to bypass the learning module. Why would the learning module know this property exists?

tristanls:
I'm not sure how to move forward.

I'm writing down my current context in hopes that I can recreate it after the hackathon.

This PR has three algorithmic changes.

The first change is that the TouchObject positioning procedure now always ends with a MoveForward action taken in the scope of the positioning procedure (unprocessed by the model).

The second change is that the SurfacePolicy now begins with a MoveForward action. This is
because when the SurfacePolicy loses track of the object, the previous action was a MoveTangentially.
So, when SurfacePolicy resumes, the next action in the cycle, after TouchObject finds the object again, is a MoveForward.

The third change is that since the TouchObject positioning procedure occurs in a loop within the data loader, the steps used by the positioning procedure no longer increment the step counter inside Monty.

As a result of the algorithmic changes above, most of the relevant benchmarks are worse.

It is not apparent to me why additional `MoveForward` actions would cause benchmarks to worsen. When I hacked together a version of the code that removed the third change and ensured that all steps in the positioning procedure were still sent to the Monty model as motor-only steps (like before), the benchmarks still got worse.

Having identified the three algorithmic changes, I can attempt three separate PRs, each introducing just one of those changes.

tristanls:
I isolated one of the algorithmic changes in the #310 pull request (still draft as I'm writing this). Note that benchmarks get worse as a result, but at least we know the exact culprit and can discuss the single change there.

tristanls:
Closing this pull request in favor of #310, and follow-up pull requests.

PR #309, opened 2025-05-15
chore: add .rsyncignore to help with infra tooling

tristanls:
This pull request adds `.rsyncignore` to help with infrastructure tooling when using `rsync`.

### DIFF
--- a/.rsyncignore
+++ b/.rsyncignore
@@ -1,6 +1,9 @@
-.git
-.github
-.mypy_cache
-docs
-rfcs
-*.pyc
+- .git
+- .github
+- .mypy_cache
+- .ruff_cache
+- .vale
+- .vscode
+- docs
+- rfcs
+- *.pyc
--- a/.rsyncignore
+++ b/.rsyncignore
@@ -1,0 +1,6 @@
+.git
+.github
+.mypy_cache
+docs
+rfcs
+*.pyc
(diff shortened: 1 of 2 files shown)


.rsyncignore
jeremyshoemaker, line 1: The diff is making this confusing, but I'm pretty sure from reading the man page for `rsync` that this isn't the right format for these rules.

My reading of the man pages makes me think that these need to have a `-` at the start of each rule.

image

jeremyshoemaker (changes requested):
I don't think this file is currently using the right syntax.

.rsyncignore
tristanls: The file works in `monty_lab`: https://github.com/thousandbrainsproject/monty_lab/blob/main/.rsyncignore

jeremyshoemaker: I'm not sure how that can be the case, given what the documentation says. I initially looked for something about maybe it having a default TYPE of "exclude". 

But if you've tested this file and it works as expected, then it's fine I guess.

tristanls: Because we use `--exclude-from`, my assumption is that any pattern is interpreted as `-`. We might get a different behavior if we used `-f`.

tristanls: @jeremyshoemaker, I updated the format to be robust per the specification in the manual you pointed out.

jeremyshoemaker: Ah, that's probably what it is. Looking further up in the man pages it says:

> **--exclude-from**=_file_
> Load _patterns_ and _rules_ from _file_

So it accepts bare patterns, in addition to rules.

jeremyshoemaker (approved):

vkakerbeck (approved):
great, thanks for adding this!

PR #312, opened 2025-05-26
style: use elif instead of else then if (PLR5501)

hlee9212:
Found another one.

### DIFF
--- a/src/tbp/monty/frameworks/utils/plot_utils.py
+++ b/src/tbp/monty/frameworks/utils/plot_utils.py
@@ -2291,7 +2291,7 @@ def plot_evidence_transitions(
                     )
                 else:
                     detection = "true_negative"
-                        # Not plotted because the "default" result
+                    # Not plotted because the "default" result
 
                 # Log results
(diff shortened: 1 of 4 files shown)


src/tbp/monty/frameworks/utils/plot_utils.py
```diff
@@ -2276,6 +2273,15 @@ def plot_evidence_transitions(
                     else:
                         detection = "false_positive"
 
+                    ax.scatter(
+                        ii + 1,
+                        this_step_max,
+                        marker="x",
+                        s=85,
+                        color=detection_cmapping[detection],
+                    )
```
tristanls, line 2282: **note**: This entire if/else setup feels like it should be for setting `detection` and after `detection` is set, we should probably call `ax.scatter` when `detection != "true_negative"`. This way, we wouldn't repeat this code block inside the different `if/else` branches.

hlee9212: Agreed. I created Issue #315 to keep track / remind myself to open a separate PR for this.

PR #313, opened 2025-05-27
style: replace explicit loop-index with enumerate (SIM113)

hlee9212:


src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -118,9 +118,8 @@ def should_get_episode(start, stop, episodes, counter):
     detailed_json = {}
     stop = stop or np.inf
     with open(json_file, "r") as f:
-        line_counter = 0
-        for line in f:
-            if should_get_episode(start, stop, episodes, line_counter):
+        for line_counter, line in enumerate(f):
+            if should_get_episode(start, stop, episodes, counter=line_counter):
```
tristanls, line 122: **question (non-blocking)**: Any particular reason why `line_counter` is passed as a named argument?

tristanls (approved):

src/tbp/monty/frameworks/utils/logging_utils.py
hlee9212: No reason. Changing to positional.

PR #314, opened 2025-05-28
refactor!: remove extraneous reset by refactoring reset out of __iter__

tristanls:
This turned out to be more complicated than I expected.

This pull request removes an extraneous reset coded in `__iter__` and ensures all data loaders do the reset logic inside `pre_episode` instead.

Since the reset logic now always happens in `pre_episode`, two previous tests that were testing for logging `patch_off_object` now throw a `ValueError`, since the experiment is being initialized off the object. If those original tests intended to ensure that we know that an experiment started off object, the new tests maintain that intent. However, if those original tests intended to test `patch_off_object` logging functionality specifically, we'll need a different test. Based on the original docstrings, I believe the intent was the former.

The three Habitat tests and two embodied data tests were shortened. They implicitly relied on the reset in `__iter__` but didn't seem to do anything other than test that each observation is returned as intended.

## Benchmarks

_compared with d126aa160de8a1813b48aaf6b2788881a93bf644_

base_config_10distinctobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/tuintpm2/overview
base_config_10distinctobj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/v92grbhv/overview
randrot_noise_10distinctobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/lqgb599s/overview
randrot_noise_10distinctobj_dist_on_distm ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/1vswvcpe/overview
randrot_noise_10distinctobj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/829ix3w6/overview
randrot_10distinctobj_surf_agent,99.00,1.00,28,18.50,2,11 😐 (same as #317) - https://wandb.ai/thousand-brains-project/Monty/runs/giu1vcd4/overview
```diff
-randrot_10distinctobj_surf_agent,99.00,1.00,28,18.49,2,12
+randrot_10distinctobj_surf_agent,99.00,1.00,28,18.50,2,11
```
randrot_noise_10distinctobj_5lms_dist_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/2bek9su9/overview
base_10simobj_surf_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/qblu9uh4/overview
randrot_noise_10simobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/g6dqq4wd/overview
randrot_noise_10simobj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/7kvn6lha/overview
randomrot_rawnoise_10distinctobj_surf_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/k2hbjtda/overview
base_10multi_distinctobj_dist_agent,79.29,10.71,31,19.84,3,1 ❌ (degraded, same as in #317 although there's a typo there, percent correct should have been 79.29 from 79.28571428571428)  - https://wandb.ai/thousand-brains-project/Monty/runs/97fa80wg/overview
```diff
-base_10multi_distinctobj_dist_agent,79.41,9.56,32,19.47,3,1
+base_10multi_distinctobj_dist_agent,79.29,10.71,31,19.84,3,1
```
base_77obj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/2qhi4wj8/overview
base_77obj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/e3gwcyx4/overview
randrot_noise_77obj_dist_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/qgk2upco/overview
randrot_noise_77obj_surf_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/ngndnv2e/overview
randrot_noise_77obj_5lms_dist_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/zg82di9b/overview

### DIFF
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -264,6 +264,17 @@ class EnvironmentDataLoaderPerObject(EnvironmentDataLoader):
         self.epochs = 0
         self.primary_target = None
 
+    def __iter__(self):
+        """Do not reset the dataset when starting the iterator.
+
+        `self.pre_episode()` already resets the dataset before each episode, therefore,
+        `EnvironmentDataLoader.__iter__()` should not reset the dataset again.
+
+        Returns:
+            EnvironmentDataLoaderPerObject: The iterator.
+        """
+        return self
+
     def pre_episode(self):
         super().pre_episode()
         self.reset_agent()
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -155,13 +155,11 @@ class EnvironmentDataLoader:
         self._counter = 0
 
     def __iter__(self):
-        # Reset the environment before iterating
-        self._observation, proprioceptive_state = self.dataset.reset()
-        self.motor_system._state = (
-            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
-        )
-        self._action = None
-        self._counter = 0
+        """Implement the iterator protocol.
+
+        Returns:
+            EnvironmentDataLoader: The iterator.
+        """
         return self
 
     def __next__(self):
@@ -181,6 +179,14 @@ class EnvironmentDataLoader:
 
     def pre_episode(self):
         self.motor_system.pre_episode()
+
+        # Reset the dataset and the data loader state.
+        self._observation, proprioceptive_state = self.dataset.reset()
+        self.motor_system._state = (
+            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
+        )
+        self._action = None
+        self._counter = 0
 
     def post_episode(self):
         self.motor_system.post_episode()
@@ -264,20 +270,12 @@ class EnvironmentDataLoaderPerObject(EnvironmentDataLoader):
         self.epochs = 0
         self.primary_target = None
 
-    def __iter__(self):
-        """Do not reset the dataset when starting the iterator.
-
-        `self.pre_episode()` already resets the dataset before each episode, therefore,
-        `EnvironmentDataLoader.__iter__()` should not reset the dataset again.
-
-        Returns:
-            EnvironmentDataLoaderPerObject: The iterator.
-        """
-        return self
-
     def pre_episode(self):
         super().pre_episode()
-        self.reset_agent()
+
+        self.motor_system._state[self.motor_system._policy.agent_id][
+            "motor_only_step"
+        ] = False
 
     def post_episode(self):
         super().post_episode()
@@ -404,23 +402,6 @@ class EnvironmentDataLoaderPerObject(EnvironmentDataLoader):
                 primary_target_object=primary_target_obj,
             )
 
-    def reset_agent(self):
-        logging.debug("resetting agent------")
-        self._observation, proprioceptive_state = self.dataset.reset()
-        motor_system_state = MotorSystemState(proprioceptive_state)
-        self._counter = 0
-
-        # Make sure to also reset action variables when resetting agent during
-        # pre-episode
-        self._action = None
-        motor_system_state[self.motor_system._policy.agent_id]["motor_only_step"] = (
-            False
-        )
-
-        self.motor_system._state = motor_system_state
-
-        return self._observation
-
 
 class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
     """Dataloader that supports a policy which makes use of previous observation(s).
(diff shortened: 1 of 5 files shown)


src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -182,6 +180,14 @@ def __next__(self):
     def pre_episode(self):
         self.motor_system.pre_episode()
 
+        # Reset the dataset and the data loader state.
+        self._observation, proprioceptive_state = self.dataset.reset()
+        self.motor_system._state = (
+            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
+        )
+        self._action = None
+        self._counter = 0
```
vkakerbeck, line 189: Would it make sense to wrap this into a reset function and then just calling self.reset() here? That way it is easier to know how to modify pre_episode if we for some reason don't want to reset the dataloader every episode

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -435,15 +427,6 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
     iv) Supports hypothesis-testing "jump" policy
     """
 
-    def __iter__(self):
-        # Overwrite original because we don't want to reset agent at this stage
-        # (already done in pre-episode)
-
-        # TODO look into refactoring the parent __iter__ method so that we don't need
-        # to use this fix
```
vkakerbeck, line 443: Haha I totally forgot about this! Looks much better now

tests/unit/evidence_lm_test.py
```diff
@@ -856,33 +856,21 @@ def test_fixed_actions_evidence(self):
             "When using detailed logging we should store matches at every steps.",
         )
 
-    def test_patch_off_object_logging(self):
-        """Test that patch_off_object is logged when no object is present.
-
-        Test that if there is no object in the scene to begin with,
-        we log patch_off_object.
-        """
+    def test_pre_episode_raises_error_when_no_object_is_present(self):
```
vkakerbeck, line 877: It looks like this test was intended to test the logging of patch_off_object. What happens in this new version? Would we still log patch_off_object or just throw a ValueError?

vkakerbeck (commented):
Thanks for adding this fix!

Regarding the unit tests, they were aimed at testing that patch_off_object logging works as expected. It would be nice if this functionality is preserved because, while we generally want to avoid starting off an object, it can happen, and logging this correctly is useful for pinpointing issues.

Also, did you check whether this change affects any of the benchmarks? Does it change what the first episode in each episode will be since we don't step twice before sending observations to the LM?

tests/unit/evidence_lm_test.py
tristanls: Instead of logging patch_off_object, Monty will now throw ValueError.

vkakerbeck: I'm not sure this is the behavior we want. Previously, we had several experiments that had some episodes where we didn't get the patch on the object. While ideally we don't want this to happen in any episodes, not throwing an error but instead logging it allowed us to continue experimenting without having to spend a lot of time fixing the few off-object episodes. At that time, the ongoing research had a higher priority than having perfect initialization conditions for every episode, so it was good that we didn't have an error that forced us to take care of those episodes first. 
I can also see the argument for throwing the error, as it makes it more explicit when some episodes don't run as expected (instead of logging it to a file that may not be looked at by a user). 
@nielsleadholm what are your thoughts?

nielsleadholm: Yeah I think after the robustness changes @scottcanoe made, it would be very unusual to get patch off object now, and should be a fairly strong red flag, so I am supportive of it raising an error rather than just logging it.

vkakerbeck: Okay, sounds good :) I approved the PR now

jeremyshoemaker: Where are we raising the ValueError? I'm wondering if it might make sense to make a specific error class for this case. Depends on the context, though.

jeremyshoemaker (approved):
LGTM

tests/unit/evidence_lm_test.py
tristanls: I'll add a note to change it to `ObjectNotVisible`. I also need it in the upcoming SurfacePolicy/touch_object refactor as well: https://github.com/thousandbrainsproject/tbp.monty/pull/306/files#diff-db9da94ec9b5aff3a39254a57a114d20d9b458e0e06dc5fea9d4420b4d8bca97R855

src/tbp/monty/frameworks/environments/embodied_data.py
tristanls: I want to defer doing that until something other than reset happens in pre_episode. Then, we could refactor this code into a reset() method. For now, it is one less method lookup to follow to see what happens.

PR #316, opened 2025-06-03
RFC: Model-Free Saccades (Draft)

scottcanoe:


rfcs/model_free_saccades.md
```diff
@@ -0,0 +1,55 @@
+- Start Date: 2025-06-02
+- RFC PR: 
+
+# Summary
+This RFC proposes a mechanism for efficient model-free saccades to be used by the distant agent.
+
+# Motivation
+There are two primary motivations
+ 1. Efficiency: We would like to reduce the number of inference steps required by the distant agent.
+ 2. Multi-Object Scenes and Compositional Objects: We would like to constrain saccades to either keep sensors on an object or more sensors to a different objects.
```
vkakerbeck:
```suggestion
 2. Multi-Object Scenes and Compositional Objects: We would like to constrain saccades to either keep sensors on an object or move sensors to different objects.
```

rfcs/model_free_saccades.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Background
+The superior colliculus (SC) is a good mental model for what we need. Given visual input from a wide field-of-view, the SC generates one or more maps used to select the next location to attend to do.
```
vkakerbeck:
```suggestion
The superior colliculus (SC) is a good mental model for what we need. Given visual input from a wide field-of-view, the SC generates one or more maps used to select the next location to attend to.
```

rfcs/model_free_saccades.md
```diff
@@ -0,0 +1,55 @@
...
+ - persistent -- at least for a while, and
+ - live in world/body coordinates.
+
+More broadly, any map not imbued with world/body coordinates becomes stale after the next saccade. There is a workaround though -- maintain a memory of actions taken, and integrate those to populate a visitation map de novo after each observation. (I think this basically what the SC does, but only with the most recent action taken.)
+
+However, the question of whether we think maps should be persistent is a general consideration. As mentioned above, segmentation maps may be most useful if they reflect object classifications produced by LMs. On the other hand, building persistent maps in world/body coordinates in SC is moving into "where" pathway and/or hippocampus territory, and we might end up placing a lot of demands on the SC that are better off done elsewhere.
```
vkakerbeck: Good questions and concerns. Maybe for now we try and keep it as simple as possible and just add a bias in the SC not the reverse the previous action? I would be curious what we can find in the literature about this. 
It seems like the natural place to do this would be the LM, where we already have a buffer of recent observations in world coordinates. But then the question is how it would inform the SC about this. It could just inform what goal state or hotspot area the LM outputs.

rfcs/model_free_saccades.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Policy Switching
+The SC needs to know whether it should stay within an object or move to a new one. This makes it sound like the SC would be like a motor system with swappable motor policies, but that can't be quite right since the SC would perform lots of sensor processing.
```
vkakerbeck: Could this be done using the LM output? Like the LM outputs a broad goal state region to pay attention to within which the SC then selects the exact location. If the LM wants to look at another object, it would drastically change the location of that general region based on where it thinks another object would be. If the LM has no hypotheses yet, it does not define any region and the SC just looks at the most salient feature in the entire field of view.

rfcs/model_free_saccades.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Architecture
+How an SC would integrate into Monty's current architecture, including which components need to be added or augmented, will require a lot of feedback and discussion. Previously, the team put put together a version of this diagram:
```
vkakerbeck:
```suggestion
How an SC would integrate into Monty's current architecture, including which components need to be added or augmented, will require a lot of feedback and discussion. Previously, the team put together a version of this diagram:
```

rfcs/model_free_saccades.md
nielsleadholm: **suggestion:** Maybe add something like this - we had discussed this before but I was struggling to articulate it during our Monday meeting:

Efficiency also ties in with compositional objects - in order to support compositional objects, we are likely going to give [Monty a shorter memory horizon for its internal representations](https://thousandbrainsproject.readme.io/docs/implement-and-test-rapid-evidence-decay-as-form-of-unsupervised-memory-resetting). While this will enable the representations to change when we move onto a new child object (desired property), this comes at a cost - if Monty does not quickly move to useful features within an object, it may never recognize an object before it has "forgotten" any distinguishing features it has observed. A better saccade policy is important to addressing this.

rfcs/model_free_saccades.md
```diff
@@ -0,0 +1,55 @@
...
+
+Finally, the SC is thought to be THE place where the saccade command are ultimately issued. If Monty were built this way, it would probably mean passing the vision-receiving LMs' goal states to the SC where they would leave a huge impact on priority maps, perhaps only being overriden if something really drastic happens (like a moving object you need to evade).
+
+# Requirements
+### Segmentation
+Given the need to support saccades in multi-object scenes and compositional objects, its clear that we will need our superior colliculus to perform object segmentation, even if very approximate. One question is whether we want LMs to be able to inform segmentation maps. For example, if an LM has decided which object it is looking at, the region corresponding to the LM's observations could be "filled in" with an object ID. But this informed segmentation map would only be useful if the map persisted for some time. Another complication is what we do when there are multiple learning modules. We would have to know where the LM is "seeing" in the visual field in order to assign its output to a region in the visual field maintained by the SC.
```
nielsleadholm: "We would have to know where the LM is "seeing" in the visual field in order to assign its output to a region in the visual field maintained by the SC." --> if we follow our current thinking of goal-states, the LM would output a goal-state that is in shared/body-centric coordinates, so this itself might not be too much of an issue. The main difficulty might be converting from a 3D body-centric coordinate, to a more retinopic coordinate system, but see the higher level comment about maps and which ones we might use.

rfcs/model_free_saccades.md
nielsleadholm: I agree with keeping this simple for now. It might just be worth including in the write-up that the biological analogy for this policy switching might be more like the striatum. L5 projects to the striatum, and in that case the goal-states might specify more abstract policies and where to enact them (e.g. enact the sub-policy that will attend our sensors to whatever object is at the pose xyz, aka "stay on object at xyz", or its inverse of avoiding attending sensors to a particular region/object), rather than the more concrete, low-level goal-state of "move sensor A to this location", like in the superior colliculus.

rfcs/model_free_saccades.md
```diff
@@ -0,0 +1,55 @@
...
+  2. We need some way to switch the SC's "policy" between "stay-on-object" and "move-off-object" (or "move-to-a-specific-object").
+
+### Output
+What would an SC's output be? And who determines whether to enact the SC's suggestions or an LM's goal states? At first glance, it might seem like having the SC emit goal states would work, but the way goal states are handled by `InformedEnvironmentDataLoader` is much more than a saccade, including a full resetting of the agent's location and orientation (feel free to correct me on this).
+
+It might make more sense to emit actions that can be executed directly. In this case, something would need to receive both SC actions and LM goal states. I don't know what this entity would be.
+
+Note: there is one part of our code that already implements SC-like behavior -- `GetGoodView`. This code uses raw sensor data and coordinates obtained via `DepthTo3DLocations`. With this data, it decides where it ought to look, and computes the necessary "look" actions used directly by the motor system. This seems like a good template for an SC if we can work out how to do something similar within Monty's main execution loop.
```
nielsleadholm: I agree! Maybe a nice starting point would actually be refactoring `GetGoodView` to follow our finalized structure for how "Non-CMP Sensory Modules" and the motor system should interact, once we've decided on that.

nielsleadholm:
Another high level comment (mentioning it here so we don't forget), is that once we've agreed on the major details of the RFC, it could be worth presenting the proposal at a research meeting so that we can get feedback from Jeff.

rfcs/model_free_saccades.md
scottcanoe: I can't find the reference, but from what I recall, the SC's has at least a 1-step back "memory" that gets imprinted into a post-saccade salience map because the SC is the place where the previous command originated. I put "memory" in quotes because it seems more like the previous motor command isn't so much as stored in memory as it is just a briefly persistent signal that lasts long enough to comingle with the next salience map computation.

But whether the SC has a long or short term memory of recent actions, it'll still need some information about the action that was taken. In the first draft, I had assumed the SC would output actions (rather than a salience map) which would help -- but I had overlooked what would happen when those commands are ignored in favor of an LM's goal state.

So in short, it does seem like the NSM (new naming via @nielsleadholm) will need some knowledge about one or more recent actions. Receiving a "corollary discharge" from the motor system seems logical, but then I'm not sure how the coordinates work out. If we know that the action received from the motor system was a saccade, then we should be able to map those motions onto the salience map, but I haven't thought it through all the way.

scottcanoe:
@vkakerbeck and @nielsleadholm, this is great feedback. Thanks!

To summarize a few things:

It sounds like we be moving towards having the NSM (non-CMP-compliant sensor module) operating in world/body coordinates. A few thoughts: 

A salience map in world coordinates means some 3D representation of a salience map. The same would go for the attentional maps informed by LMs. I'm not sure the best way to go about partitioning space like this. We could either break the world into some kind of grid, but if we want to maintain a degree of resolution comparable to the view-finder's resolution, that's going to mean a huge amount of voxels. Alternatively, each feature in the salience map could be mapped to something like a scalar function at each point (f: R3 -> R), and if we want to know the "interestingness" at any given point, we sum over the functions evaluated at that point (like finding the electrical potential). The challenges then would be 
- transforming a salience map into a reasonable set of functions (which might not be hard for the NSM)
- finding the point(s) where the value is highest (probably harder outside the NSM, but probably easy within the NSM since it can first find the max in retinotopic coordinates and convert from there into world/body coordinates)

If we want LM input to the NSM, would it make sense to route LMs through the NSM? **I'm really hesitant to suggest this.** I bring it up though for a few reasons.

- Whatever solution we come up with for 3D salience maps in the NSM would have to be repeated in the motor system to, for example, build attentional maps having the same structure.
- How much control over choosing a location to explore should the motor system have? Of course, they have lots of freedom in this respect already with model-free policies. But I had a brief conversation about the motor system with @tristanls during the hackathon about what the motor system really is and what it's supposed to do, so I wanted to revisit that thought and try to clarify exactly what a motor system's responsibilities are. In some ways, it seems it would be cleanest if the motor system's only job was to figure out how to achieve a goal state rather than deciding which of many potential areas to explore that it should ultimately look at. Then again, this "decision" process is the "attention selector" in Viviane's picture. So that aspect is being somewhat separated out of the motor system with this model.
- If the NSM did receive LM input, theoretically it could be the one and only place operating with the 3D attention/salience construct, and it could just emit goal states. That might help constrain the kinds of data motor systems are responsible for receiving and maintaining and keep the new 3D salience/attentional construct encapsulated to within the NSM. We could still have an attentional selector within the motor system, but it would only make decisions based on received objects of the same type -- goal states.

Another note about goal states: Correct me if I'm wrong, but as of now, trying to achieve a goal state always uses `execute_jump_attempt` which also changes the agent's pose. It seems like there would be value in somehow restricting this behavior. Goal states sent by the NSM should be achievable only through saccades. Also, only-saccadable goal states would be able to work outside of a simulator environment without worrying about how to move an agent around an object (maybe a minor point).

The above point is arguable. As Viviane says, we don't want out goal states to be actuator specific. But do you think there's some value in somehow preventing the system from doing entire pose resets every time it attempts a goal state?

rfcs/model_free_saccades.md
vkakerbeck: Could we maybe solve it all with the goal state map of the LM? So basically have the subcortical sensor processing only be responsible for computing a saliency map based on sensed features. Then have the LM (which already has knowledge about how we previously moved) take the recently visited locations into account when outputting its's state. Or it could be part of the motor system to bias against reversing it's previous action (which we already have in some policies). I guess I don't see a strong argument right now of why this computation would have to happen in the subcortical sensor processing unit.

vkakerbeck:
How I am thinking of it right now I would split if up like this:

**Learning Module**: Output's goal states. Those goal states express a location in world coordinates that the sensor should move/attend to. We ideally want it to also express the size/shape of a region around that location instead of just a singular place to look at (distinguish between wanting to look into a general area, an object, or a specific location on an object).
**Subcortical Sensory Processing Unit**: Calculates saliency maps based on raw sensory input. The output could be a list of goal states (same as LM, with the option to specify size.shape of region around location) that indicate the peaks in the saliency map in the common coordinate frame (world/body).
**Action Selector**: This is kind of the grey circle within the motor system in my diagram but I agree it could make sense to talk about this as a separate unit. It get's all the goal states from all the LMs, and the subcortical processing units. It then looks at which location relative to the body/world wins out and sends this to the motor system to execute. Maybe a more complex version of this in our brain could be the basal ganglia? In that case it also takes into account the state of the system (hungry,...) but in our case we would just want to start with getting the location with the most support from all the goal states.
**Motor System** Takes the goal state (location and orientation in the world to move to) and moves the sensor there. This is where constraints of the actuator come in. If the actuator can't move freely (like an eye being fixed to a socket) then the motor system may just be able to achieve an approximation of the goal state. But I think this knowledge of the actuator and it's action space should remain in the motor system. I don't think we should implement actuator-specific subcortical processing like the SC.

What do you think?
This may also be a good topic to discuss on an excalidraw board :)

nielsleadholm:
A couple thoughts:

Agree with all your comments @vkakerbeck , including that we might take the Action Selector out of the Motor System.

Maybe the one I would want to discuss further is whether a goal-state can output a "shape" - to me this moves towards communicating whole models. I think using the current formulation of goal-states we could get close enough to describing a non-circular region that we want to attend to. This is at least what I was trying to describe in my comment 2 days ago under "Re. output goal states and a priority map".

Re. the representational format of the saliency map
- That's an interesting thought @scottcanoe about querying functions (if I'm following it correctly). If we did do something more like voxels in 3D space, I don't think there's any reason they couldn't be lower resolution, with resolution for a given volume of space dynamically determined based on the task.

Re. the Action/Goal-State Selector (AS)
- One possibility is that it receives goal-states from LMs, and non-goal state type representations from NSMs. I think the most key requirement is that both received inputs would be in shared (body/world) coordinates, so that the Action Selector doesn't need to transform the representation into a new space. The disadvantage of forcing the NSM to output goal states is it could limit its expressiveness/enforce an awkward implementation. But from a code and conceptual standpoint, it might be easier for everything to send goal-states to the AS. I'm currently on the fence about this.

Re. achieving goal-states (with model-free policies)
- 100% agree Scott that we want to modify this. Currently the jump-to-goal state was the quickest hack we could implement to enable achieving goal-states, but we would definitely want the ability to have more constrained ones (like only saccading). I think it would be the responsibility of the basal ganglia (which might also fit under the Action Selector) to decide what model-free sub-routines are allowed, like only performing saccades. A key part of this RFC (or if it's becoming too much, a follow-up one) should be thinking through how we can take the final goal-state received by the Motor System, and determine the necessary saccade (a model-free saccade policy). Future work can consider the sub-routine switching that might be orchestrated by the basal-ganglia/AS. For the purpose of this work, we can assume that the jump-to-goal-state would be disabled and never used by the Motor Policy when running saccade experiments.

nielsleadholm:
@vkakerbeck 

Scott and I just met and attached is an updated diagram from the [Excalidraw board ](https://app.excalidraw.com/s/gO8aypxgkZ/9FOwpMAufwG) with how we're picturing it at the moment. I think it's pretty consistent with what you are also suggesting but just thought it would be helpful to draw it out like you suggested.

overview_updated

second_example

scottcanoe:
@vkakerbeck and @nielsleadholm

Here's a summary and some thoughts about what I think we may be converging on.

### New Components

**NSM**
 - Input: The view-finder's RBG and depth data (and/or world coordinates associated with RGB pixels).
 - Output: a salience map (more on this below).
 
**ActionSelector**
 - Input: Goal States from LMs and a salience map from the `NSM`.
 - Output: A single goal state for the motor system?
 
### Refactors
 
**MotorSystem**: All goal states are currently handled with pose resets. I believe a refactor here is needed that enables the motor system to decide whether a saccade is sufficient to achieve a goal state or whether the current, pose-resetting behavior is needed. Maybe the `ActionSelector` would do this?

**DepthTo3DTransform** and/or **SensorModule**: Optional? Though we would like to eventually move `DepthTo3DTransform`'s behavior into the sensor module's processing pipeline rather than the dataset's, we don't need to wait on this refactor. We should be able to use the view-finder's data and coordinates as they are currently generated.

### Salience Maps

Salience maps are pretty much 2D things, but I think we're widening the definition to include some kind of 3D representation. My take is that this 3D representation will be a point cloud, and each point has an associated scalar that represents that point's salience. These points are naturally represented in the same way as goal states, so this is a natural choice.

While the NSM's output will be a 3D point cloud, my intuition is that almost all the work done by the NSM will happen in 2D retinotopic space in order to leverage all the existing techniques and theory about salience maps and segmentation. But depth is a pretty powerful segmentation cue, so if we do perform segmentation, we'll probably use that as well.

### More Thoughts about an `ActionSelector`

Having the `NSM` produce salience maps without any knowledge of the current task or desired behavior (i.e., staying on-object) means that salience maps will be behaviorally uninformed. If the `NSM` was task-aware, the region to target could be embedded in the salience map by, for example, returning only points that exist within some region, or by assigning them higher priority. Similarly, if the `NSM` was aware of previous actions, it could also embed these into salience maps by reducing the salience/priority values associated with previously visited points.

What I'm gathering is that these history- and task-aware behaviors will be implemented by the `ActionSelector`. In a sense, this makes the `ActionSelector` the superior colliculus in every way, minus the generation of low-level salience maps. I think this is very natural with respect to return inhibition (i.e., not revisiting areas) since the `ActionSelector` will "know" what has been previously visited, being that it is the place where previous actions were selected. As for task-aware behaviors, such as "stay on-object", I'm less sure. Since staying on-object (or moving off-object) feels closely related to segmentation, the `ActionSelector` will either need to perform segmentation, either through some novel algorithm that operates on point clouds or by temporarily projecting into retinotopic space and performing segmentation there.

vkakerbeck:
I was thinking of it slightly differently. Maybe I am just missing the reasons you discussed in your meeting to get the full picture of why you settled on this but here is how I was trying to describe it:
image

The big difference to what you are proposing is that all inputs to the action selector are goal states. There are goal states that are based on model-based information (from the LM) and goal states that are based on model-free information (from the sensor processing). I would think of both subcortical sensor processing units as sensor modules that output CMP compliant messages. However, the outputs from the one that processes the larger low-resolution receptive field goes to the action selector instead of to the learning module. Basically one SM outputs sensed features@locations while the other outputs desired locations to move to. 
Theoretically they could be the same class and just output both. I haven't thought this through enough to tell what would be better. If SMs can just all output features at locations to LMs and sensation based goal states then that simplifies things. The question is how useful it is for an LM to learn from data from those large low-res receptive fields and how useful it is to output goal states based on the smaller high-res patches. But I could think of reasons why this could be useful (i.e learning large low-res models, microsaccades, ...).

We still have a saliency map but it is local to the sensor module. It then picks the hotspots in the saliency map and outputs those as goal states. It doesn't really matter what space the saliency map is in as long as the SM can extract target locations in the common reference frame from it. The action selector then just picks the location with the most support from all the goal states it receives (could take into account confidence). Since all it's inputs are CMP compliant goal states (i.e. in a common reference frame) this is pretty straight forward. 

The motor system then gets one goal state and has to execute on it. I don't see why a refactor of the motor system is necessary for this. I though we would just implement different policies for the different actuators. If we have a ball and socket actuator, it's policy would execute a goal state differently than the policy of a more freely moving agent. Basically how we switch the `InformedPolicy` out for the `SurfacePolicy` when we want to do experiments with a surface instead of a distant agent. The main thing we would need to do is make sure each policy has a function that takes in a goal state and translated it into actions out of it's action space.

Let me know what you think and if this makes sense with what you were discussing.

scottcanoe:
@vkakerbeck I think we're basically on the same page, so I might not be clear in my language. LMs would emit goal states (like normal), and so would an `NSM` -- thought it would emit many. I had described the `NSM`'s output as a 3D salience map akin to a point cloud, but its essentially the same as a list of goal states (with salience scores, similar to confidence). This makes the `ActionSelector` only have to receive objects of one type, and it figures out which of the goal states to send to the motor system.

I do have a couple of thoughts though. As I see it, there are two main things we want out of this system. We want it to
 1. Select feature-rich areas to attend to, and
 2. Constrain its actions so that it will stay on and object (or move off an object).
 
Number 1 is the straightforward part, and it we should be able to accomplish this with some straightforward salience detection techniques.

Number 2 is harder. This is where I was thinking segmentation would come in. Since many of the salient areas in a wide FOV may be entirely off-object, I figured we could filter out the off-object goal-states with segmentation, but then we need to decide which segmented area we're going to consider "good". We could just say that whatever segmented area that is in the center of the FOV is the on-object area, but what if the task is to move to a different object? Or if the previous saccade ended up slightly off the object we were targeting and we want to move back onto it?
   The alternative, more complicated variant I mentioned as a possibility was let an LM's estimates determine which of the segmented regions we might want to consider "good". So model-free segmentation and model-free salience detection would define regions and especially feature-rich areas within regions, and the LM's model-based estimates could, in principle, be used to help figure out which of those segmented regions to ignore or focus on.

I have a feeling I'm not being super clear here, and it's a tricky to put this into words.

As for the motor state and refactors, I'll defer to you on this. I have limited experience with this part of the system. All I'm aware of is the `execute_jump_attempt` code, and I don't know how this is explicitly connected to actuators of a given type like a ball and socket).

vkakerbeck:
Ah okay thanks for the clafication, that does sound like we are pretty much on the same page. In this case I would not refer to it as NSM since it would be CMP compliant. I called it Action-focused SM but it could also just end up being a normal SM with a larger receptive field.

In terms of the segmentation we could start with the basic on/off object segmentation we have implemented which is based on the depth map. I recently posted some pictures and infos on this here: https://thousandbrains.discourse.group/t/sensor-fundamentals/634/3 We could use this map to make sure all of the SM goal states and up in the on-object area. When we want to move to a new object I would use the LM goal state for this. By the time we move to a new object the LM should have a good idea of which object it is sensing. It should therefor know where the object ends and which locations won't be on the object anymore. It could output those locations with high confidence to overwrite the SMs goal states. What do you think?

Good point on the `execute_jump_attempt` function. That should be moved out of the dataloader and become part of a policy for agents that can move freely. For example each policy could have an `move_to_location` or `execute_goal_state` function and the current `execute_jump_attempt` function could be called there if we have a surface agent. For a contained distant agent that can't move it's location we would have to write new code. So yeah, there will be some refactoring involved to move this function into a specific policy.

nielsleadholm:
Nice, yeah sounds like we're converging on a nice structure. I like your suggestion Viviane of explicitly defining the SM motor outputs as goal-states, and so they would actually still be CMP compliant. We might just want to update the definition of an SM to be able to also output a CMP goal-state, rather than just a CMP sensory signal. Some SMs might output one, some might output both. 

Re. the segmentation question, one thing to consider is that since they will be goal-states, they can have labels/object IDs associated with them. Naturally, the model-free SM won't have models of objects, but it could still use heuristics to say that a certain region (and therefore all the associated goal-states) is "Object A", and another region is "Object B". Since this information would be in the goal-states, the Action Selector is free to make use of it if e.g. the current goal is to stay on object, and the LMs still don't have a very good hypothesis for what they're seeing. This would hopefully address your concern Scott about needing to do segmentation in the Action Selector.

Re. how the motor system actualizes goal-states, I would just say that yeah at the moment the data-loader assumes that the only way to act on goal-states is the jumping policy, so that would need some changing, if it has not already been refactored by that point. But the basic principle of passing a goal-state to the motor system, and assuming it will try to enact it with some sub-routine, will stay the same.

nielsleadholm:
I updated the diagram to try to incorporate our standard colors/dashes for CMP, non-CMP, motor etc. to hopefully make things a bit clearer. 

Screenshot 2025-06-12 at 3 20 11 PM

vkakerbeck:
sounds great and nice update to the diagram. Just two thoughts:
1. We will have to experiment with this but I would guess that the number of goals states that the SM outputs doesn't have to be in the 1000s. In most cases I would assume it is sufficient to output 1-10 goal states. If the LM doesn't have a good hypothesis then we just move to the peak in the saliency map so no need to output locations with lower saliencies. If the LM does have a good hypothesis then the saliency in the sensor input won't matter too much so there is also no big need to output a lot of goal states. The only thing to consider is that the goal states of the SM should at least have some of them be on the object.
2. Related to segmentation I generally like that general idea to express "same-object-ness" in the CMP message. It might just get a bit complicated once the sensor starts moving and we have to keep track of objects to make sure they keep using the same segmentation mask ID. Not impossible but just something to consider. Maybe a topic to tackle when we actually notice a need for it.

scottcanoe:
@vkakerbeck and @nielsleadholm This draft is basically a rewrite. I'm pretty happy with it as a starting point for (more) discussion and feedback, so if you're good with it, I can open a new PR with it on Monday.

Just to note that the **Integration** is intentionally skeletal since I didn't really want to make a concrete plans or suggestions on those matters before talking it through with the team.

scottcanoe:
> 2\. Related to segmentation I generally like that general idea to express "same-object-ness" in the CMP message. It might just get a bit complicated once the sensor starts moving and we have to keep track of objects to make sure they keep using the same segmentation mask ID. Not impossible but just something to consider. Maybe a topic to tackle when we actually notice a need for it.

We can definitely give it a shot. I imagine we could put something together using basic heuristics that would work well most of the time, like when objects are easily segmentable and the agent doesn't move much. But ultimately, I think the day Monty starts observing the world in a more temporally continuous way will be the first day of its model-free life. Motion cues alone would probably unlock some pretty elegant solutions for segmentation and same-object-ness.

vkakerbeck:
Nice! I'm looking forward to reading the rewrite. Just let me know if I should do this here or on the new PR you plan to open.

Regarding the motion detection, I don't see a reason why we couldn't do this now already. You can update an SM as frequently as you like so it can be close to continuous (fully continuous would only work on different hardware as far as I can tell). You don't have to send an observation to an LM at every step (like we already do when we are off object or with the FeatureChangeSM) so you can run may frequent inputs through an SM, do motion processing and then send an observation to an LM every once in a while. Lmk if I am missing something.

nielsleadholm:
Great thanks Scott, yeah feel free to open it as a separate PR now. And yeah makes sense about keeping the code high level for now.

PR #317, opened 2025-06-03
refactor!: extract hypotheses update logic into HypothesesUpdaters

tristanls:
In #289, we added a `ResamplingHypothesesEvidenceMixin` to replace the logic of how evidence is updated in the `EvidenceGraphLM`. The technique used there is a Mixin, which overrides some functionality. Following our new project code organization guidance heuristics, this pull request extracts the current evidence updating logic from `EvidenceGraphLM` into a `DefaultHypothesesUpdater`. Then, following multiple refactors, it eventually refactors `ResamplingHypothesesEvidenceMixin` as a new `ResamplingHypothesesUpdater`, which is used to configure `EvidenceGraphLM`. This pull request is a lot, so I encourage you to hang in there.

## DefaultHypothesesUpdater extraction

The first step was to identify the general bundle of code that `ResamplingHypothesesEvidenceMixin` wanted to replace and extract it into its own component. This way, when we have a `ResamplingHypothesesUpdater` ready, we can plug it into `EvidenceGraphLM` instead of the default functionality. Hence, first we identify and extract the [DefaultHypothesesUpdater](https://github.com/thousandbrainsproject/tbp.monty/pull/317/files#diff-9cc2383adfac523f2f4d7410705bc394a811f1ce71ee947fb6485ef69819a6fd).

By default, `EvidenceGraphLM` instantiates an instance of `DefaultHypothesesUpdater` in its constructor:
```python
        hypotheses_updater_class: type[HypothesesUpdater] = DefaultHypothesesUpdater,
        hypotheses_updater_args: dict | None = None,
        (...)
        if hypotheses_updater_args is None:
            hypotheses_updater_args = {}
        # Every HypothesesUpdater gets at least the following arguments because they are
        # either constructed or edited in the constructor, or they are shared with the
        # learning module.
        hypotheses_updater_args.update(
            feature_evidence_increment=self.feature_evidence_increment,
            feature_weights=self.feature_weights,
            graph_memory=self.graph_memory,
            max_match_distance=self.max_match_distance,
            past_weight=self.past_weight,
            present_weight=self.present_weight,
            tolerances=self.tolerances,
        )
        self.hypotheses_updater = hypotheses_updater_class(**hypotheses_updater_args)
```

## Hypotheses bundling

The original functionality continuously passed around three arrays: locations, poses, evidence, and a channel identifier. This type of pattern is error-prone, so these were bundled into [`Hypotheses` and `ChannelHypotheses` data classes](https://github.com/thousandbrainsproject/tbp.monty/pull/317/files#diff-9783ba4c1804c32794b1519e41b037347c60494d412d6c45677bcb7673fbc4e6), thus reducing cognitive load when thinking about hypotheses updates. This refactor makes no implementation changes, only how arguments are passed and how the code reads are affected. (This is also one of the rare examples where an inheritance hierarchy use seems appropriate, as a `ChannelHypotheses` _is-a_ `Hypotheses`).

```python
@dataclass
class Hypotheses:
    """Set of hypotheses consisting of evidence, locations, and poses.
    The three arrays are expected to have the same shape. Each index corresponds to a
    hypothesis.
    """

    evidence: np.ndarray
    locations: np.ndarray
    poses: np.ndarray


@dataclass
class ChannelHypotheses(Hypotheses):
    """A set of hypotheses for a single input channel."""

    input_channel: str
```

## DefaultHypothesesDisplacer extraction

When implementing `ResamplingHypothesesUpdater`, it happens to reuse the hypotheses displacement functionality. But that was now embedded inside the `DefaultHypothesesUpdater`. So, to reuse the functionality, the hypotheses displacing functionality was extracted into the `DefaultHypothesesDisplacer` component. Both hypotheses updaters now use that displacer functionality by default:

```python
# DefaultHypothesesUpdater.__init__
        self.hypotheses_displacer = DefaultHypothesesDisplacer(
            feature_evidence_increment=self.feature_evidence_increment,
            feature_weights=self.feature_weights,
            graph_memory=self.graph_memory,
            max_match_distance=max_match_distance,
            max_nneighbors=max_nneighbors,
            past_weight=past_weight,
            present_weight=present_weight,
            tolerances=self.tolerances,
            use_features_for_matching=self.use_features_for_matching,
        )
```
```python
# ResamplingHypothesesUpdater.__init__
        self.hypotheses_displacer = DefaultHypothesesDisplacer(
            feature_evidence_increment=self.feature_evidence_increment,
            feature_weights=self.feature_weights,
            graph_memory=self.graph_memory,
            max_match_distance=max_match_distance,
            max_nneighbors=max_nneighbors,
            past_weight=past_weight,
            present_weight=present_weight,
            tolerances=self.tolerances,
            use_features_for_matching=self.use_features_for_matching,
        )
```

## EvidenceSDRMixin complications

The previous changes would have been sufficient, but the `EvidenceSDRMixin` replaces a different subset of functionality, so it required additional refactors to keep it working (although it seems to be unused dead code at this point, as it is not used in any benchmarks).

## FeatureEvidenceCalculator and FeaturesForMatchingSelector extraction

The `EvidenceSDRMixin` changes functionality internal to the hypotheses updaters. So, every functionality being replaced was turned into a component and extracted. The first one was the extraction of the `DefaultFeatureEvidenceCalculator`, the second was the extraction of `DefaultFeatureForMatchingSelector`.

Once the above were extracted, then, the `EvidenceSDRMixin` could implement its own custom versions: `SDRFeatureEvidenceCalculator` and `AllFeaturesForMatchingSelector`, and configure its hypotheses updater to use those instead of the defaults.

```python
# EvidenceSDRLMMixin.__init__
        # Configure the evidence updater with SDRFeatureEvidenceCalculator and
        # AllFeaturesForMatchingSelector by default.
        updater_args = kwargs.get("hypotheses_updater_args", {})
        if not hasattr(updater_args, "feature_evidence_calculator"):
            updater_args["feature_evidence_calculator"] = SDRFeatureEvidenceCalculator
        if not hasattr(updater_args, "features_for_matching_selector"):
            updater_args["features_for_matching_selector"] = (
                AllFeaturesForMatchingSelector
            )
        kwargs["hypotheses_updater_args"] = updater_args
```

By default, the hypotheses updaters use the default calculator and selector:
```python
class DefaultHypothesesUpdater:
    def __init__(
        self,
        ...,
        feature_evidence_calculator: Type[FeatureEvidenceCalculator] = (
            DefaultFeatureEvidenceCalculator
        ),
        ...,
        features_for_matching_selector: Type[FeaturesForMatchingSelector] = (
            DefaultFeaturesForMatchingSelector
        ),
        ...,
    ):
```
```python
class ResamplingHypothesesUpdater:
    ...
    def __init__(
        self,
        ...,
        feature_evidence_calculator: Type[FeatureEvidenceCalculator] = (
            DefaultFeatureEvidenceCalculator
        ),
        ...,
        features_for_matching_selector: Type[FeaturesForMatchingSelector] = (
            DefaultFeaturesForMatchingSelector
        ),
        ...,
    ):
```

One note is that the calculator and selector are pure functions, so they could be implemented as bare functions. However, I implemented them as static methods so that there are no `Callable`s in our configuration and the configuration remains serializable per our requirements.

## evidence_matching folder

With these separate functionalities being extracted, breaking functionality into individual files became necessary to avoid circular imports.

## Project Code Organization Guidance

Here is a partial list of heuristics I used to decide on how to conduct this refactoring:

- Protocols SHOULD be used to document usage and expectations
- Inheritance hierarchy SHALL be only one deep
- Inheritance hierarchy parent MUST be a concrete class
- Bare functions or class methods SHOULD be used to share a functionality implementation that does not access instance state
- Mixins SHOULD be used to share a functionality implementation that accesses state of the instance being mixed with
- Mixins SHALL NOT add state to the instance being mixed with
- Composition SHOULD be used to share a functionality implementation that needs its own state

---

> [!NOTE]
> `EvidenceSDRGraphLM` appears to be dead code, as it is not used in any benchmarks.

## Benchmarks

_compared with d126aa160de8a1813b48aaf6b2788881a93bf644_

base_config_10distinctobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/ljfpfbin/overview
base_config_10distinctobj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/5h0mb7ds/overview
randrot_noise_10distinctobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/a7ldsaj1/overview
randrot_noise_10distinctobj_dist_on_distm ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/aqf360nl/overview
randrot_noise_10distinctobj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/9bzora70/overview
randrot_10distinctobj_surf_agent,99.00,1.00,28,18.50,2,11 😐 (rotation error 0.01 degree diff) - https://wandb.ai/thousand-brains-project/Monty/runs/2150fgmv/overview
```diff
-randrot_10distinctobj_surf_agent,99.00,1.00,28,18.49,2,12
+randrot_10distinctobj_surf_agent,99.00,1.00,28,18.50,2,11
```
randrot_noise_10distinctobj_5lms_dist_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/mrfi07is/overview
base_10simobj_surf_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/bnlhjk51/overview
randrot_noise_10simobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/pxzzhjih/overview
randrot_noise_10simobj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/71jq2qer/overview
randomrot_rawnoise_10distinctobj_surf_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/6158haz1/overview
base_10multi_distinctobj_dist_agent,79.28,10.71,32,19.83,3,1 ❌ (degraded)  - https://wandb.ai/thousand-brains-project/Monty/runs/1dstf33i/overview
```diff
-base_10multi_distinctobj_dist_agent,79.41,9.56,32,19.47,3,1
+base_10multi_distinctobj_dist_agent,79.28,10.71,32,19.83,3,1
```
base_77obj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/flr0obwv/overview
base_77obj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/b2lax23s/overview
randrot_noise_77obj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/1d1o7fwk/overview
randrot_noise_77obj_surf_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/ngndnv2e/overview
randrot_noise_77obj_5lms_dist_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/n4xfp1se/overview

_previous refactor before resampling hypotheses mixin merge (compared with 16c8f1f297f5316b29fbd995c34847afa0742746)_

base_config_10distinctobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/zobmijs5/overview
base_config_10distinctobj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/9n09fksk/overview
randrot_noise_10distinctobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/u6jkd3nm/overview
randrot_noise_10distinctobj_dist_on_distm ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/zpw950q3/overview
randrot_noise_10distinctobj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/ioxc984f/overview
randrot_10distinctobj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/vernlwxw/overview
randrot_noise_10distinctobj_5lms_dist_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/hiojd86b/overview
base_10simobj_surf_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/7kks5eev/overview
randrot_noise_10simobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/8537h8sk/overview
randrot_noise_10simobj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/gxjwnoom/overview
randomrot_rawnoise_10distinctobj_surf_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/iy3hqcyd/overview
base_10multi_distinctobj_dist_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/jjz25zfx/overview

base_77obj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/h5hkts8l/overview
base_77obj_surf_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/ulq8n8b2/overview
randrot_noise_77obj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/jrgxk0qk/overview
randrot_noise_77obj_surf_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/sdv3lrj7/overview
randrot_noise_77obj_5lms_dist_agent ✅  - https://wandb.ai/thousand-brains-project/Monty/runs/cp9446t5/overview

> [!NOTE]
> There is a mistake in #318 for `base_10multi_distinctobj_dist_agent`. The diff reads:
> ```diff
> -base_10multi_distinctobj_dist_agent,79.29,10.71,31,20.05,43,1
> +base_10multi_distinctobj_dist_agent,79.41,9.56,32,19.47,3,1
> ```
> However, that looks like a typo, as clicking on the linked WandB run https://wandb.ai/thousand-brains-project/Monty/runs/jtqkdg22/overview shows:
>
> `base_10multi_distinctobj_dist_agent,79.29,10.71,31,19.84,3,1`
>
> This pull request corrects this mistake and verifies that this benchmark did not change since the run in #318.


docs/how-monty-works/evidence-based-learning-module/evidence-as-a-similarity-measure.md
```diff
@@ -75,4 +75,4 @@ Want to dive deeper into this topic? Have a look at this presentation:
   "href": "https://www.youtube.com/watch?v=qHdl8tlzY3k",
   "typeOfEmbed": "youtube"
 }
-[/block]
\ No newline at end of file
+[/block]
```
tristanls, line 78: **note**: newline at end of file

docs/how-to-use-monty/tutorials/unsupervised-continual-learning.md
```diff
@@ -227,4 +229,4 @@ After running this script, you should see a plot with three views of the bowl ob
...
 
-Thus far we have demonstrated how to build models that use a single sensor module connected to one learning module. In the [next tutorial](multiple-learning-modules.md), we will demonstrate how to configure multiple sensors and connect them to multiple learning modules to perform faster inference.
\ No newline at end of file
+Thus far we have demonstrated how to build models that use a single sensor module connected to one learning module. In the [next tutorial](multiple-learning-modules.md), we will demonstrate how to configure multiple sensors and connect them to multiple learning modules to perform faster inference.
```
tristanls, line 232: **note**: newline at end of file

tristanls:
FYI @nielsleadholm, @hlee9212, @scottcanoe, @jeremyshoemaker. This is an example of refactoring a bundle of functionality into an object component so that it can be swapped out. In this case, #289 wants to change how evidence is calculated. So, first, we extract the default functionality before implementing the alternative.

src/tbp/monty/frameworks/models/evidence_matching/evidence_updater.py
```diff
@@ -0,0 +1,782 @@
...
+    rotate_pose_dependent_features,
+)
+
+
+@dataclass
+class EvidenceUpdate:
+    evidence: np.ndarray
+    input_channel: str
+    locations: np.ndarray
+    poses: np.ndarray
```
vkakerbeck: Does this refer to possible_locations and possible_poses? Can we make these variable names more explicit? We have locations and poses at many places in the code (sensed location and pose at each step, locations and poses stored in the buffer, hypothesized locations and poses on the object,...) so I think just using locations and poses as variable names can be confusing.

src/tbp/monty/frameworks/models/evidence_matching/evidence_updater.py
```diff
@@ -0,0 +1,782 @@
...
+from tbp.monty.frameworks.utils.spatial_arithmetics import (
+    align_multiple_orthonormal_vectors,
+    get_angles_for_all_hypotheses,
+    get_more_directions_in_plane,
+    rotate_pose_dependent_features,
+)
+
+
+@dataclass
+class EvidenceUpdate:
```
vkakerbeck: Can we add a docstring to this class? What is it used for?

src/tbp/monty/frameworks/models/evidence_matching/evidence_updater.py
```diff
@@ -0,0 +1,782 @@
...
+
+    def update_evidence(
+        self,
+        evidence: np.ndarray,
+        features: dict,
+        displacements: dict | None,
+        graph_id: str,
+        locations: np.ndarray,
+        mapper: ChannelMapper,
+        poses: np.ndarray,
```
vkakerbeck: again here, the variable names are kind of confusing. Which of those refer to hypotheses and which refer to the current sensation? They should at least be grouped together but idealy this information should also be in the variable name

src/tbp/monty/frameworks/models/evidence_matching/evidence_updater.py
```diff
@@ -0,0 +1,782 @@
...
+        current_mlh: dict,
```
vkakerbeck: It looks like you are passing current_mlh here to calculate the evidence update threshold. This is not very straight forward to see though and someone may be confused why we need to know about the mlh to update evidences. Could we calculate the update threshold before calling `update_evidence` and then just pass the threshold to this function here?

vkakerbeck (commented):
Thanks for the nice example of how this could be refactored. I like the idea. I left a few comments but here are some more high-level notes:

- could we change the name `evidence_updater` to `hypothesis_updater`? This would be more accurate as it doesn't just update the amount of evidence for a hypothesis but also the hypothesized locations and orientations at every step. It also initialized the hypothesis space and can update the hypotheses in that space.
- This PR changes experiment configs so it will at the very least require changes to the tutorials code as well in ideally also to our monty_lab configs.
- I would be in favor of merging PR #289 first

src/tbp/monty/frameworks/models/evidence_matching/evidence_updater.py
tristanls: The point of (now renamed) `HypothesesUpdater` is only to bundle this data together so we don't have to pass around all four of these individually. The bundle has no context for what these are used for. I think what you are referring to is better addressed by naming the update variable. For example:
```python
possible_update = HypothesesUpdate(
    evidence=possible_evidence,
    input_channel="potato",
    locations=possible_locations,
    poses=possible_poses
)
```
If these are sensed, then it'll be `sensed_update = HypothesesUpdate(...)`, if they are hypothesized, they'll be `hypothesized_update = HypothesesUpdate(...)`.

In code usage, further below, is an example of this sort of thing:
```python
hypotheses_updates.append(
    HypothesesUpdate(
        evidence=channel_hypotheses_evidence,
        input_channel=input_channel,
        locations=channel_possible_locations,
        poses=channel_possible_poses,
    )
)
```
And when it gets applied in the learning module, it looks like this:
```python
        for update in hypotheses_updates:
            self._set_hypotheses_in_hpspace(
                graph_id=graph_id,
                input_channel=update.input_channel,
                new_location_hypotheses=update.locations,
                new_pose_hypotheses=update.poses,
                new_evidence=update.evidence,
            )
```

I think we only clutter and do not gain anything by having separate classes with context-specific variable names like:
```python
@dataclass
class PossibleHypothesesUpdate:
    evidence: np.ndarray
    input_channel: str
    possible_locations: np.ndarray
    possible_poses: np.ndarray

@dataclass
class SensedHypothesesUpdate:
    evidence: np.ndarray
    input_channel: str
    sensed_locations: np.ndarray
    sensed_poses: np.ndarray

@dataclass
class HypothesizedHypothesesUpdate:
    evidence: np.ndarray
    input_channel: str
    hypothesized_locations: np.ndarray
    hypothesized_poses: np.ndarray
```

vkakerbeck: Yes I agree, just renaming the class clarifies things enough. Just to clarify, I didn't mean that there are any other types of updates. We never update the sensed locations. We only update our hypotheses. So it wouldn't make sense to make separate classes for this. I was just referring to different places in the code where we talk about locations and that we want to be unambiguous in what kind of locations we are referring to.

tristanls: Yes, the name change you suggested is much better.

I also added this commit, showing that the context can change as the update is passed. The contents of HypothesesUpdate go from "possible" to "new".

https://github.com/thousandbrainsproject/tbp.monty/pull/317/commits/90e4a34bab1429390fd47b9a22d09806813f22fa

src/tbp/monty/frameworks/models/evidence_matching/evidence_updater.py
tristanls: This makes a lot of sense, thank you. I'll give that a try.

tristanls: I changed this to pass the `max_global_evidence` instead of `current_mlh`, but when I attempted to extract the threshold calculation back to the learning module I got really weird errors I don't quite understand. I'm setting that aside for now, but maybe I'll come back to it.

tristanls: I figured this out. Latest changes moved the threshold calculation out of the hypotheses updater.

tristanls:
Converting this pull request into a draft while I work through ResamplingHypothesesEvidenceMixin refactor.

tristanls:
@vkakerbeck, @ramyamounir, this pull request is once again ready for review. I updated the main description. Please take a look and let me know if you have any questions. We might need to do 1:1 sessions to help with the review.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
```diff
@@ -0,0 +1,385 @@
...
+            # here in case we want to revert back to those.
+            # avg_update = np.mean(new_evidence)
+            # evidence_to_add = np.zeros_like(channel_hypotheses_evidence)
+            evidence_to_add = np.ones_like(possible_hypotheses.evidence) * min_update
+            evidence_to_add[hyp_ids_to_test] = new_evidence
+
+            # If past and present weight add up to 1, equivalent to
+            # np.average and evidence will be bound to [-1, 2]. Otherwise it
+            # keeps growing.
+            updated_evidence = (
```
ramyamounir: One issue I see here is that `updated_evidence` is not defined outside of this `if num_hypotheses_to_test > 0` condition, so we cannot rely on it being available to initialize the returned `ChannelHypotheses`. 

A hacky suggestion is to define a default `updated_evidence = possible_hypotheses.evidence` before the condition.

tristanls: Good catch, thank you, I missed this. Updated.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -0,0 +1,390 @@
...
+
+class HypothesesUpdater(Protocol):
+    def update_hypotheses(
+        self,
+        hypotheses: Hypotheses,
+        features: dict,
+        displacements: dict | None,
+        graph_id: str,
+        mapper: ChannelMapper,
+        max_global_evidence: float,
```
ramyamounir: I noticed that `DefaultHypothesesUpdater` and `ResamplingHypothesesUpdater` do not follow the `HypothesesUpdater` protocol. Specifically, `update_hypotheses` accepts `max_global_evidence` but is passed `evidence_update_threshold` instead. Was this maybe more recently updated without updating the protocol?

tristanls: Thank you. This is an error on my part. The protocol should accept the evidence threshold as you stated.

hlee9212:
Just lurking and checking out the refactoring and wanted to say hallelujah! 😄

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
ramyamounir: Nice, thanks. While you're at it, maybe also fix the Displacer protocol. The signature for `displace_hypotheses_and_compute_evidence` in the protocol doesn't match the one in `DefaultHypothesesDisplacer`. `channel_displacement` and `channel_features` positions are swapped.

Probably doesn't matter, just a nitpick.

tristanls: Oof, thank you. I will look into enabling type checking sooner rather than later as well. Protocol use that I'm introducing here depends on type checking to catch these errors.

ramyamounir:
Really nice refactor, I learned a lot just going through the changes. Thanks!

One final comment is that perhaps the calculation of `evidence_update_threshold` should be part of the updater logic. Would moving `_get_evidence_update_threshold` to the updater make sense? Also in the near future, we may decide to remove the update threshold logic and it would be easier to do it in the `ResamplingHypothesesUpdater`.

tristanls:
Thanks @ramyamounir.

> One final comment is that perhaps the calculation of evidence_update_threshold should be part of the updater logic

The problem is that calculation logic requires `current_mlh`. See https://github.com/thousandbrainsproject/tbp.monty/pull/317#discussion_r2131960365 . Ended up moving out of the updater so that it did not receive `current_mlh` as a parameter.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
tristanls: Updated.

ramyamounir:
> The problem is that calculation logic requires current_mlh. See https://github.com/thousandbrainsproject/tbp.monty/pull/317#discussion_r2131960365 . Ended up moving out of the updater so that it did not receive current_mlh as a parameter.

I see, I lean more towards your initial attempt of passing `max_global_evidence`. I can give that a shot in a follow-up PR before I remove the threshold logic from resampling. This is a non-blocking issue.

ramyamounir (approved):

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
```diff
@@ -0,0 +1,388 @@
...
+    get_custom_distances,
+    get_relevant_curvature,
+)
+from tbp.monty.frameworks.utils.spatial_arithmetics import (
+    get_angles_for_all_hypotheses,
+    rotate_pose_dependent_features,
+)
+
+
+class HypothesesDisplacer(Protocol):
```
vkakerbeck, line 35: This name seems to narrow for what it is doing. When I read `HypothesisDisplacer` I would assume it just updates the hypothesized locations and rotations using the sensed movement. However, here we also update the evidence for all the hypotheses (which is actually the main part of the calculations here). So it is more of a `HypothesisUpdater`. But we already have a hypothesis updater. What is the difference here? their docstring description is very similar ("Updates evidence by comparing features after applying sensed displacement." and "Update hypotheses based on sensor displacement and sensed features.".)

tristanls: I can see how `HypothesesDisplacer` can be an unsatisfying name here. On the one hand, a good name helps a lot, on the other, attempting to name things so that they are clear out of context can be challenging, if not futile.

For example, while this is called a `HypothesesDisplacer`, its only method is `displace_hypotheses_and_compute_evidence`, which immediately highlights that it is computing evidence in addition to displacement. Additionally, when it is called in context:
```python
                # sensory input.
                channel_possible_hypotheses = (
                    self.hypotheses_displacer.displace_hypotheses_and_compute_evidence(
                       # ...
                    )
                )
```
also helps seeing that it updates evidence.

Lastly, `Hypotheses` class itself includes `evidence`, which implies that a hypothesis consists of location, pose, and evidence.

That said, there may well be a better name for it. Happy to rename it in the future if there's a better candidate. Renaming symbols is very straightforward.

I also agree that the docstrings are very similar. This is the same similarity exhibited by the docstrings in the code these replaced.

Screenshot 2025-06-10 at 16 22 57
Screenshot 2025-06-10 at 16 23 36

vkakerbeck: Sorry if this wasn't clear but my comment was hostly asking about what the difference is (not just asking you to come up with a better name). From the description I can't tell in what case I should use HypothesisDisplacer and when I should use HypothesisUpdater. Can you explain to me what the difference is and why we need both?

btw this naming for displacing hypothesis is something I already commented on in the original PR from @ramyamounir here https://github.com/thousandbrainsproject/tbp.monty/pull/301#discussion_r2077705819 I agree this doesn't have to be something to fix in the scope of this refactor.

nielsleadholm:
Just wanted to say thanks @tristanls for the huge effort putting this together, and the clear overview of the changes.

PR #320, opened 2025-06-10
chore: mypy configuration to recognize tbp.monty package

tristanls:
We still ignore errors for now, but if you want local `mypy` type checking and set `ignore_errors=false`, then this `mypy` configuration will now correctly load available Monty code for type checking.


codeallthethingz (approved):
LGTM

PR #322, opened 2025-06-10
chore: show mypy errors without failing check

tristanls:


tristanls:
FYI @scottcanoe, @hlee9212, @ramyamounir, @nielsleadholm, @vkakerbeck, if you have mypy configured in your IDE, you will see type errors from now on. You can ignore the existing errors; the GitHub checks will still pass. However, please don't add new type errors 🙂.

PR #310, opened 2025-05-15
refactor!: SurfacePolicy touch_object phase always ends with MoveForward action

tristanls:
The intended future refactor of `SurfacePolicy` is to separate the `TouchObject` positioning procedure. This was initially attempted in #306. However, after analysis, it turned out that #306 contained three algorithmic changes, and it was difficult to identify and accept the changes.

This pull request isolates and applies only one of the algorithmic changes: TouchObject positioning procedure now always ends with a MoveForward action.

This pull request ensures that the current embedded touch object positioning procedure (in the form of `touch_object` method) always runs until the viewfinder sees the object within range. Then, it executes the MoveForward action. The reason for implementing this change is that once the TouchObject positioning procedure is extracted in the future, it will always run automatically to completion and will always end with the MoveForward action. Before this pull request, `SurfacePolicy` could short-circuit `touch_object` by detecting that it is close enough to the object and begin the main `SurfacePolicy` (in `dynamic_call`) pre-emptively.

## Benchmark results
Multiple decimal places results here; the table updates are in the benchmark result changeset.

| Code | Experiment | Correct (%) | Used MLH (%) | Num Match Steps | Rotation Error | WandB |
|---|---|---|---|---|---|---|
| new | base_config_10distinctobj_surf_agent | 100 | 0 | 27.84285714285714 | 0.24448000000000003 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/86lgqo1w/overview)
| recent (#305) | base_config_10distinctobj_surf_agent | 100 | 0 | 27.642857142857142 | 0.20275 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/4d51r1go/overview)
| new | randrot_noise_10distinctobj_surf_agent | 99 | 0 | 28.81 | 0.3788070707070707 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/1q58zyn6/overview)
| recent (#305) | randrot_noise_10distinctobj_surf_agent | 100 | 1 | 28.46 | 0.3758580000000001 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/rhddvorw/overview)
| new | randrot_10distinctobj_surf_agent | 99 | 1 | 27.94 | 0.32283333333333336 | [run](https://wandb.ai/thousand-brains-project/Monty/runs/tfioufxd/overview)
[...]

PR #323, opened 2025-06-11
chore: version 0.6.0

tristanls:
Minor version release due to Major version 0 and a breaking changes since 0.5.0.

Screenshot 2025-06-11 at 09 43 22

## refactor!: replace generic Exception with specific Errors (TRY002) (#273)

If you were relying on the previously raised Errors, update to catching the newly raised Errors.

## refactor!: SurfacePolicy touch_object phase always ends with MoveForward action (#310)

This is a breaking change due to the change in behavior of the `SurfacePolicy`. If you are using the `SurfacePolicy`, you should evaluate whether the updated behavior meets your current needs.

## refactor!: extract hypotheses update logic into HypothesesUpdaters (#317)

Multiple API changes in how hypotheses are updated within the `EvidenceGraphLM`. The changes are drop-in without altering behavior; however, the code is refactored. If you changed or were working with `EvidenceGraphLM` internals, take note of the updated code organization.

## refactor!: remove extraneous reset by refactoring reset out of __iter__ (#314)

Previously, dataset and data loader were reset twice at the beginning of every episode, once in `__iter__` and once in `pre_episode`. Now, reset happens only once in `pre_episode`. This is a behavior change, and you should evaluate whether the updated behavior meets your current needs.

### DIFF
--- a/src/tbp/monty/__init__.py
+++ b/src/tbp/monty/__init__.py
@@ -8,4 +8,4 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
-__version__ = "0.5.0"
+__version__ = "0.6.0"


codeallthethingz (approved):
🎉

PR #325, opened 2025-06-11
refactor: extract `_get_evidence_update_threshold` into a bare function

ramyamounir:
[PR #317](https://github.com/thousandbrainsproject/tbp.monty/pull/317) has significantly refactored the functionality of hypotheses updating outside of `EvidenceGraphLM`. We can now use it as a component of the LM and swap it out with e.g., `ResamplingHypothesesUpdater`. 

This PR proposes to refactor the evidence_update_threshold functionality outside of `EvidenceGraphLM` as well. I'm proposing this for the following reasons:
- Conceptually, calculating the evidence_update_threshold is part of the hypotheses update logic. A new hypotheses updater should be able to redefine this heuristic.
- In follow-up resampling work, I plan to simplify this heuristic and remove the functionality from the updater logic. The default would be to update all hypotheses without using a threshold. It is part of the `ResamplingHypothesesUpdater`and refactoring this function into the scope of the `HypothesesUpdater` is the first step.

**I turned the class function `_get_evidence_update_threshold` into a bare function (i.e., not belonging to any class) to be used by multiple classes.**

> [!NOTE]
> I modified the `_get_evidence_update_threshold` to handle the edge case of empty evidence array at the beginning of the function.

---

### Benchmark Comparisons

| Metric   |   main |   refactor_evidence_threshold |   Δ (delta) |
|----------|--------|-------------------------------|-------------|
| time     | 2.1265 |                         2.116 |     -0.0105 |

📌 Differences found in experiment: unsupervised_inference_distinctobj_dist_agent
| Metric                        |     main |   refactor_evidence_threshold | Δ (delta)   |
|-------------------------------|----------|-------------------------------|-------------|
| time                          |   4.2224 |                        4.2791 | +0.0567     |

### DIFF
--- a/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
@@ -51,6 +51,7 @@ class HypothesesUpdater(Protocol):
         displacements: dict | None,
         graph_id: str,
         mapper: ChannelMapper,
+        max_global_evidence: float,
     ) -> list[ChannelHypotheses]:
         """Update hypotheses based on sensor displacement and sensed features.
 
@@ -61,6 +62,8 @@ class HypothesesUpdater(Protocol):
             graph_id (str): Identifier of the graph being updated
             mapper (ChannelMapper): Napper for the graph_id to extract data from
                 evidence, locations, and poses based on the input channel
+            max_global_evidence (float): Highest evidence of all hypotheses (i.e.,
+                current mlh evidence),
 
         Returns:
             list[ChannelHypotheses]: The list of channel hypotheses updates to be
@@ -209,6 +212,8 @@ class DefaultHypothesesUpdater:
             graph_id (str): Identifier of the graph being updated
             mapper (ChannelMapper): Mapper for the graph_id to extract data from
                 evidence, locations, and poses based on the input channel
+            max_global_evidence (float): Highest evidence of all hypotheses (i.e.,
+                current mlh evidence),
 
         Returns:
             list[ChannelHypotheses]: The list of hypotheses updates to be applied to
@@ -251,7 +256,7 @@ class DefaultHypothesesUpdater:
                     hypotheses, input_channel
                 )
 
-                # calculate the evidence_update_threshold
+                # Calculate the evidence_update_threshold
                 evidence_update_threshold = get_evidence_update_threshold(
                     self.evidence_update_threshold,
                     self.x_percent_threshold,
@@ -422,7 +427,7 @@ def all_usable_input_channels(
 
 
 def get_evidence_update_threshold(
-    evidence_update_threshold: int,
+    evidence_update_threshold: float | str,
     x_percent_threshold: float | str,
     max_global_evidence: float,
     evidence_all_channels: np.ndarray,
@@ -430,10 +435,13 @@ def get_evidence_update_threshold(
     """Determine how much evidence a hypothesis should have to be updated.
 
     Args:
-        evidence_update_threshold (float | str): update threshold type.
-        x_percent_threshold (int): x_percent threshold used in the calculation of
-            evidence threshold.
-        max_global_evidence (float): highest evidence of all hypotheses,
+        evidence_update_threshold (float | str): The heuristic for deciding which
+            hypotheses should be updated.
+        x_percent_threshold (float | str): The x_percent value to use for deciding
+            on the `evidence_update_threshold` when the `x_percent_threshold` is
+            used as a heuristic.
+        max_global_evidence (float): Highest evidence of all hypotheses (i.e.,
+            current mlh evidence),
         evidence_all_channels (np.ndarray): Evidence values for all hypotheses.
 
     Returns:
@@ -443,9 +451,6 @@ def get_evidence_update_threshold(
         InvalidEvidenceUpdateThreshold: If `self.evidence_update_threshold` is
             not in the allowed values
     """
-    # TODO: Better Args description
-    # max_global_evidence = np.max(evidence_all_channels)
-
     if type(evidence_update_threshold) in [int, float]:
         return evidence_update_threshold
     elif evidence_update_threshold == "mean":
--- a/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
@@ -192,6 +192,7 @@ class DefaultHypothesesUpdater:
         displacements: dict | None,
         graph_id: str,
         mapper: ChannelMapper,
+        max_global_evidence: float,
     ) -> list[ChannelHypotheses]:
         """Update hypotheses based on sensor displacement and sensed features.
 
@@ -254,6 +255,7 @@ class DefaultHypothesesUpdater:
                 evidence_update_threshold = get_evidence_update_threshold(
                     self.evidence_update_threshold,
                     self.x_percent_threshold,
+                    max_global_evidence,
                     hypotheses.evidence,
                 )
 
@@ -422,6 +424,7 @@ def all_usable_input_channels(
 def get_evidence_update_threshold(
     evidence_update_threshold: int,
     x_percent_threshold: float | str,
+    max_global_evidence: float,
     evidence_all_channels: np.ndarray,
 ):
     """Determine how much evidence a hypothesis should have to be updated.
@@ -430,6 +433,7 @@ def get_evidence_update_threshold(
         evidence_update_threshold (float | str): update threshold type.
         x_percent_threshold (int): x_percent threshold used in the calculation of
             evidence threshold.
+        max_global_evidence (float): highest evidence of all hypotheses,
         evidence_all_channels (np.ndarray): Evidence values for all hypotheses.
 
     Returns:
@@ -440,7 +444,7 @@ def get_evidence_update_threshold(
             not in the allowed values
     """
     # TODO: Better Args description
-    max_global_evidence = np.max(evidence_all_channels)
+    # max_global_evidence = np.max(evidence_all_channels)
 
     if type(evidence_update_threshold) in [int, float]:
         return evidence_update_threshold
--- a/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
@@ -51,7 +51,6 @@ class HypothesesUpdater(Protocol):
         displacements: dict | None,
         graph_id: str,
         mapper: ChannelMapper,
-        evidence_update_threshold: float,
     ) -> list[ChannelHypotheses]:
         """Update hypotheses based on sensor displacement and sensed features.
 
@@ -62,7 +61,6 @@ class HypothesesUpdater(Protocol):
             graph_id (str): Identifier of the graph being updated
             mapper (ChannelMapper): Napper for the graph_id to extract data from
                 evidence, locations, and poses based on the input channel
-            evidence_update_threshold (float): Evidence update threshold.
 
         Returns:
             list[ChannelHypotheses]: The list of channel hypotheses updates to be
@@ -91,6 +89,8 @@ class DefaultHypothesesUpdater:
         past_weight: float = 1,
         present_weight: float = 1,
         umbilical_num_poses: int = 8,
+        x_percent_threshold: int = 10,
+        evidence_update_threshold: float | str = "all",
     ):
         """Initializes the DefaultHypothesesUpdater.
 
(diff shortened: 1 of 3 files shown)


src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -387,3 +424,66 @@ def all_usable_input_channels(
     # features. This information is currently not available here.
     # TODO S: Once we pull the observation class into the LM we could add this.
     return [ic for ic in features.keys() if ic in all_input_channels]
+
+
+def get_evidence_update_threshold(
```
tristanls: **style**: `get` is a filler word. This function can be called `evidence_update_threshold`.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -387,3 +424,66 @@ def all_usable_input_channels(
...
+    evidence_update_threshold: float | str,
+    x_percent_threshold: float | str,
+    max_global_evidence: float,
+    evidence_all_channels: np.ndarray,
+):
```
tristanls: **issue**: What's the return type?

tristanls (changes requested):
It's a good idea to turn it into a bare function. However, I'm surprised that you moved the call location into hypotheses updaters and (as a result) had to add additional state to them.

**issue**: I expect the call location to be unchanged. This way, no state needs to be added to updaters.

**question**: What prompted moving the call location?

tristanls:
> Conceptually, calculating the evidence_update_threshold is part of the hypotheses update logic. A new hypotheses updater should be able to redefine this heuristic.

We can enable redefining this heuristic while keeping evidence_update_threshold out of the hypotheses updaters.

If this were to become an `EvidenceUpdateThresholdCalculator` (like `FeatureEvidenceCalculator`), then the `EvidenceGraphLM` could be configured with it (in the same way an updater is configured with FeatureEvidenceCalculator). Then, in the future, if a different updater needs a different threshold, we'd write a new `EvidenceUpdateThresholdCalculator` and configure `EvidenceGraphLM` with the new one instead.

```python
class EvidenceGraphLM(GraphLM):
  def __init__(
    self,
    ...,
    evidence_update_threshold_calculator: type[EvidenceUpdateThresholdCalculator] = (
      DefaultEvidenceUpdateThresholdCalculator
    ),
    ...
  ):
    self.evidence_update_threshold_calculator = evidence_update_threshold_calculator

  ...

  def _update_evidence(
    ...
  ) -> None:
    ...
    # Initialize a `ChannelMapper` to keep track of input channel range
    # of hypotheses for a specific graph_id
    if graph_id not in self.channel_hypothesis_mapping:
        self.channel_hypothesis_mapping[graph_id] = ChannelMapper()
        self.evidence[graph_id] = np.array([])
        self.possible_locations[graph_id] = np.array([])
        self.possible_poses[graph_id] = np.array([])
        evidence_update_threshold = 0
    else:
        evidence_update_threshold = self.evidence_update_threshold_calculator.calculate(
            evidence_update_threshold=self.evidence_update_threshold,
            max_global_evidence=self.current_mlh["evidence"],
            evidence_all_channels=self.evidence[graph_id],
            x_percent_threshold=self.x_percent_threshold,
        )
    ...
```

The future hypothesis updater could ignore the `evidence_update_threshold` parameter that the learning modules passed to it.

tristanls:
If you don't want to waste compute calculating the update threshold, you could implement a `NoOpEvidenceUpdateThresholdCalculator` that ignores the params and returns zero. It shouldn't matter if it goes with a `ResamplingHypothesesUpdater` that ignores the `evidence_update_threshold` param.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
ramyamounir: I think the name `evidence_update_threshold` is overused at the moment. We have:
* `evidence_update_threshold` (float | str) : defining the heuristic used for calculating the threshold (e.g., "all", "x_percent_threshold"). Usually defined in the configs
* `evidence_update_threshold` (float) : the actual threshold value returned from the `get_evidence_update_threshold` function
* [proposed] `evidence_update_threshold`: the function that calculates the `evidence_update_threshold` using the `evidence_update_threshold` :sweat_smile:

The filler word is somewhat reducing the confusion of: 
```python
evidence_update_threshold = evidence_update_threshold(self.evidence_update_threshold, ...)
```
I assume `calculate` instead of `get` would also be considered a filler word? Not sure how to resolve this but at least 2 of these 3 names should be different, maybe:
```python
evidence_update_threshold_value = evidence_update_threshold(self.evidence_update_threshold_heuristic, ...)
# OR
evidence_update_threshold_value = calculate_evidence_update_threshold(self.evidence_update_threshold, ...)
```

ramyamounir:
I see, yeah what you proposed would work too. The main reason for moving the call to the updater is that I think the actual calculation is part of the updater/displacer logic. Calculating the value in `EvidenceGraphLM` and passing it to the updater doesn't fit my mental model of the updater role. My plan was to completely bypass the calculation call of evidence_update_threshold in the `ResamplingHypothesesUpdater`.

Then in a follow-up PR I would define a `ResamplingHypothesesDisplacer` that ignores the `evidence_update_threshold` parameter (I think I have to keep the parameter in the function signature to follow the `HypothesesDisplacer` protocol). It would look something like this:

```python
    def displace_hypotheses_and_compute_evidence(
        self,
        channel_displacement: np.ndarray,
        channel_features: dict,
        evidence_update_threshold: float,
        graph_id: str,
        possible_hypotheses: ChannelHypotheses,
        total_hypotheses_count: int,
    ) -> ChannelHypotheses:
        # Have to do this for all hypotheses so we don't loose the path information
        rotated_displacements = possible_hypotheses.poses.dot(channel_displacement)
        search_locations = possible_hypotheses.locations + rotated_displacements


        # Get evidence update for all hypotheses
        evidence_to_add = self._calculate_evidence_for_new_locations(
            graph_id=graph_id,
            input_channel=possible_hypotheses.input_channel,
            search_locations=search_locations,
            channel_possible_poses=possible_hypotheses.poses,
            channel_features=channel_features,
        )
 
        # If past and present weight add up to 1, equivalent to
        # np.average and evidence will be bound to [-1, 2]. Otherwise it
        # keeps growing.
        evidence = (
            possible_hypotheses.evidence * self.past_weight
            + evidence_to_add * self.present_weight
        )

        return ChannelHypotheses(
            input_channel=possible_hypotheses.input_channel,
            evidence=evidence,
            locations=search_locations,
            poses=possible_hypotheses.poses,
        )
```
We would be saving some compute here because we wouldn't have to call `np.where` and indexing with `hyp_ids_to_test`.

To bypass the calculation of `evidence_update_threshold` when doing resampling, I see three options:
* As you proposed, we can create `EvidenceUpdateThresholdCalculator` and pass a `NoOpEvidenceUpdateThresholdCalculator`.
    * This looks great, but I'm not sure what the best way would be to ensure that a `ResamplingHypothesesUpdater` is always accompanied by a `NoOpEvidenceUpdateThresholdCalculator`. Would we have to explicitly set it in the `__init__` of `EvidenceGraphLM`?
* The function `get_evidence_update_threshold` is already written in a way that does different logic based on the heuristic, so I can pass it a new heuristic called e.g., "bypass" which would just skip any calculation and return `None`. This simulates a NoOp class.
    * This would still need explicit assigning of `self.evidence_update_threshold="bypass"` in `EvidenceGraphLM` since the call is still there.
* Move the call to the updater, where we can completely bypass it in `update_hypotheses` and pass None to the `ResamplingHypothesesDisplacer`.
    * This looks good to me but, as you pointed out, we will have to add more state to the updater.

I'm okay with any of these options, but if we decide to keep the call in `EvidenceGraphLM`, I would vote for the second option since the calculator is written in a way where the behavior can be modified given a heuristic parameter, so no need to make a class out of a simple util function.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
nielsleadholm: I like`calculate_evidence_update_threshold`. To me a getter is normally a method to access the value of a private / protected class property, whereas this method seems to be doing some more involved work. I would agree with you though Ramy that it's good to make it clear that this is an actual method rather than the variable or argument.

nielsleadholm (commented):
Overall looks good, just weighing in on the discussion - I can see how for at least a couple months, we would want to have backward support for an update threshold for the experiments you are running Ramy etc. However, in the longer term, we are likely to strip out `evidence_update_threshold` as a concept entirely. My general feeling therefore is that whatever solution we go for, we shouldn't spend too much time worrying about it. It could be worth adding a TODO for when we have updated all of our benchmark experiments to use your new resampling approach Ramy, that we remove `evidence_update_threshold` from the code.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
tristanls: `calculate` is a better word. I really only object to `get` and `set` fillers, as they mean everything and nothing.

The other thing I want to highlight about naming. Please consider the naming context during use and not during function creation. The names mean things in the context they are in. There are no universally ambiguous names. Attempting to create universally ambiguous names eventually leads to Java naming along the lines of `ProjectContractChargingPeriodProjectAccountReferenceVM`.

A way to think about it is to ask yourself is someone reading the code where the function is used will be confused or not.

```python
# clear in context, better
p = potato()
p = self.basket.potato()
p = self.potato(size="large")
self.basket.add(p)
# unnecessary disambiguation, worse
potato = get_potato()
potato = self.basket.get_potato()
potato = self.get_large_potato()
self.basket.set_potato(potato)
```

The other aspect about fillers like `get_` and `set_` is that, even if someone is confused about what `p = self.potato` means. They will learn it fairly quickly, on the order of seconds. Whereas if you commit to `self.get_potato()`, you will forever type `get_` everywhere and `get_` will be read everywhere. It is not about keystroke typing efficiency, but it is about the cognitive load of reading the code and parsing additional tokens.

So, if instead of the following
```python
                # Calculate the evidence_update_threshold
                evidence_update_threshold = get_evidence_update_threshold(
                    self.evidence_update_threshold,
                    self.x_percent_threshold,
                    max_global_evidence,
                    hypotheses.evidence,
                )

                # sensory input.
                channel_possible_hypotheses = (
                    self.hypotheses_displacer.displace_hypotheses_and_compute_evidence(
                        channel_displacement=displacements[input_channel],
                        channel_features=features[input_channel],
                        evidence_update_threshold=evidence_update_threshold,
                        graph_id=graph_id,
                        possible_hypotheses=channel_hypotheses,
                        total_hypotheses_count=hypotheses.evidence.shape[0],
                    )
                )
```
We have
```python
                update_threshold = evidence_update_threshold(
                    self.evidence_update_threshold,
                    self.x_percent_threshold,
                    max_global_evidence,
                    hypotheses.evidence,
                )

                # sensory input.
                channel_possible_hypotheses = (
                    self.hypotheses_displacer.displace_hypotheses_and_compute_evidence(
                        channel_displacement=displacements[input_channel],
                        channel_features=features[input_channel],
                        evidence_update_threshold=update_threshold,
                        graph_id=graph_id,
                        possible_hypotheses=channel_hypotheses,
                        total_hypotheses_count=hypotheses.evidence.shape[0],
                    )
                )
```
Do you think anyone will be confused what kind of threshold `update_threshold` is? It is only ever used as:
```python
update_threshold = evidence_update_threshold(
# ...
evidence_update_threshold=update_threshold
```

Also, the `# Calculate the evidence_update_threshold` comment kinda hints at what would be a better name, even with the `calculate` filler. What other interpretation could there be for the below code aside from `evidence_update_threshold` will give me back an update threshold, especially since it is a function? 
```python
update_threshold = evidence_update_threshold(
    self.evidence_update_threshold,
    self.x_percent_threshold,
    max_global_evidence,
    hypotheses.evidence,
)
```

tristanls: I forgot to add these examples, but I do want to include them to highlight that this isn't a Tristan thing but a good if not best practice. Is anyone confused about what these Numpy functions do?
```python
overall_stats["overall/percent_correct"] = np.mean(stats["episode/correct"]) * 100
# it isn't np.get_mean(...), it is not np.calculate_mean(...)

overall_stats["overall/run_time"] = np.sum(stats["episode/run_time"])
# it isn't np.get_sum(...) nor is it np.calculate_sum(...)

rotations = np.zeros([arr_len, 4])
# it isn't np.fill_with_zeros or similar
```

tristanls:
@ramyamounir, I like your option 2. Provide another `"bypass"` option. My pitch for the option name would be `"none"`, and have the function return `None`. This way, the code will error if someone tries to use it.

I also think you should rename `self.evidence_update_threshold` to `self.evidence_threshold_config`. ~~Based on the first if check, I assume this used to be the threshold, but evolved to be a configuration without a name change.~~ I think we're overdue for a name change. This will also make the code easier to read:

```python
update_threshold = evidence_update_threshold(
    self.evidence_threshold_config,
    self.x_percent_threshold,
    max_global_evidence,
    hypotheses.evidence,
)
```

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -62,7 +62,7 @@ def update_hypotheses(
             graph_id (str): Identifier of the graph being updated
             mapper (ChannelMapper): Napper for the graph_id to extract data from
                 evidence, locations, and poses based on the input channel
-            evidence_update_threshold (float): Evidence update threshold.
+            evidence_update_threshold (float): Evidence update_threshold
```
tristanls: **nitpick (non-blocking)**: I don't think the `_` belongs here. We don't have a category of `update_threshold`s where this is the evidence type of it. I think we have an evidence update threshold.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -248,3 +249,66 @@ def __repr__(self) -> str:
...
         ranges = {ch: self.channel_range(ch) for ch in self.channel_sizes}
         return f"ChannelMapper({ranges})"
+
+
+def evidence_update_threshold(
+    evidence_threshold_config: float | str,
+    x_percent_threshold: float | str,
+    max_global_evidence: float,
+    evidence_all_channels: np.ndarray,
+):
```
tristanls: **issue (non-blocking)**: What is the return type?

ramyamounir:
Great! Thanks for the feedback @tristanls and @nielsleadholm. I will be merging this one shortly.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -248,3 +249,66 @@ def __repr__(self) -> str:
...
+    elif evidence_threshold_config == "all":
+        return np.min(evidence_all_channels)
+    else:
+        raise InvalidEvidenceUpdateThreshold(
+            "evidence_threshold_config not in "
+            "[int, float, '[int]%', 'mean', 'median', 'all', 'x_percent_threshold']"
+        )
+
+
+class InvalidEvidenceUpdateThreshold(ValueError):
```
tristanls: **issue**: I just noticed that this error should be renamed to `InvalidEvidenceThresholdConfig`.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -302,13 +302,13 @@ def evidence_update_threshold(
...
     else:
-        raise InvalidEvidenceUpdateThreshold(
+        raise InvalidEvidenceThresholdConfig(
             "evidence_threshold_config not in "
             "[int, float, '[int]%', 'mean', 'median', 'all', 'x_percent_threshold']"
         )
 
 
-class InvalidEvidenceUpdateThreshold(ValueError):
+class InvalidEvidenceThresholdConfig(ValueError):
```
tristanls, line 311: **issue**: One more change in the docstring. I think Ruff should complain about it.

PR #326, opened 2025-06-12
refactor!: decouple actions from samplers and actuators

tristanls:


docs/how-to-use-monty/tutorials/implementing-actions.md
```diff
@@ -0,0 +1,161 @@
+---
+title: Implementing Actions
+---
+
+# What Are Actions?
+
+**Actions** are how Monty moves **Agents** in its **Environment**. Monty's **Motor System** outputs **Actions**, which are then actuated within the **Environment**. An **Agent** is what Monty calls its end effectors.
```
codeallthethingz: I'm not sure what an end effector is.

docs/how-to-use-monty/tutorials/implementing-actions.md
```diff
@@ -0,0 +1,161 @@
...
+
+With that in mind, before creating a new **Action**, consider what can be actuated in the **Environment**.
+
+# Action Lifecycle
+
+At a high level, **Actions** are created by the **Motor System** and actuated by the **Environment**.
+
+![](../../figures/how-to-use-monty/tutorials/action_lifecycle_high_level.png)
+
+Within the **Motor System**, **Policies** either explicitly choose a specific **Action** by creating it directly, or sample from a pool of actions available.
```
codeallthethingz: Is "sample" similar to clone?  Is that the reason you would do it?  Just to not have to create a new action?

codeallthethingz (approved):
👍

docs/how-to-use-monty/tutorials/implementing-actions.md
tristanls: 🤔 I need to improve the wording here.

The distinction I'm trying to make is between using some policy to decide an action, as in, "I'm too far away, I need to move forward" (create `MoveForward`), instead of randomly sampling an action to execute from a set of possibilities (sample `MoveForward` ... from the set {`MoveForward`,`MoveBack`}).

docs/how-to-use-monty/tutorials/implementing-actions.md
tristanls: 🤔 ... how about... 

An **Agent** is what Monty calls its end effectors, which carry out actions in the environment, physically or in simulation (like a robotic arm or a camera), or virtually (like a web browser navigating the Internet).

docs/how-to-use-monty/tutorials/implementing-actions.md
```diff
@@ -0,0 +1,161 @@
...
+
+![](../../figures/how-to-use-monty/tutorials/action_lifecycle_motor_system_detail.png)
+
+Within the **Environment**, the **Actions** are actuated either within a simulator, by a robot, or directly.
+
+![](../../figures/how-to-use-monty/tutorials/action_lifecycle_both_detail.png)
+
+Additionally, within an experimental framework, a **Positioning Procedure** can generate **Actions** before starting the experiment. This is analogous to an experimenter moving Monty into a starting position.
```
vkakerbeck:
```suggestion
Additionally, within an experimental framework, a **Positioning Procedure** can generate **Actions** before starting the experiment. This is analogous to an experimenter moving Monty into a starting position. The positioning procedure can use privileged information (such as labels, ground truth object models, or task instructions) and is independent of Monty and the models learned in its learning modules.
```

docs/how-to-use-monty/tutorials/implementing-actions.md
```diff
@@ -0,0 +1,161 @@
...
+
+With the above in place, the **Motor System** can now create the **Action** as needed.
+
+# Sampling Actions
+
+For the **Motor System** to be able to sample the **Action** with the help of a sampler, you will need to include a `sample` method and declare an action-specific sampler protocol:
+
+```python
+class JumpActionSampler(Protocol):
+    def sample_jump(self, agent_id: str) -> Jump: ...
```
vkakerbeck, line 70: It could help to make this more concrete by adding some code in this function. Like randomly sampling a `how_high` amount and returning the Jump action. Maybe the text could also explicitly talk about what samplers are used for (random selection of actions)

It could also be useful to explain why we need a separate Sampler class instead of just implementing the sampling code in the `sample` function of `Jump`. Like allowing for switching between sampling actions with a constant amount to sampling random or gaussian distributed amounts.

docs/how-to-use-monty/tutorials/implementing-actions.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+The actuator itself will be specific to the **Environment** and the simulator, robot, or otherwise. However, in general, it will implement the actuator protocol:
+
+```python
+class MyActuator:
+
+    def actuate_jump(self, action: Jump) -> None: ...
+        # custom code to make Jump happen
+```
```
vkakerbeck, line 136: I'm not sure this code block adds anything that isn't already in the previous code block? Why not add the comment into the previous code block or make it a bit more concrete with some example code in there?

docs/how-to-use-monty/tutorials/implementing-actions.md
```diff
@@ -0,0 +1,161 @@
...
+```python
+class SomeSimulator(MyActuator):
+
+    def step(self, action: SomeSimulatorAction) -> Observations
+        # ...
+        action.act(self)
+        # ...
+```
+
+For an example of an actuator, see `src/tbp/monty/simulators/habitat/simulator.py`
```
vkakerbeck, line 162: Nice tutorial! It helped me understand the new structure :) 
suggestion: One thing to consider could be making the example a bit more concrete. Like code that is concrete enough that someone could actually take it, know where to put it in Monty, and use it to add an action to one of our action spaces. For example, code that makes the camera in habitat move up by how_high amount and a sampler for that.

src/tbp/monty/frameworks/actions/actions.py
```diff
@@ -99,17 +128,20 @@ def __iter__(self):
                 continue
             yield key, value
 
-    @abstractmethod
-    def act(self, actuator: Actuator) -> None:
-        """Execute the action using the provided actuator."""
-        pass
+
+class LookDownActionSampler(Protocol):
+    def sample_look_down(self, agent_id: str) -> LookDown: ...
```
vkakerbeck, line 133: I'm sorry if this is a stupid question but where do we actually define what happens when sample_look_down is called?

src/tbp/monty/simulators/habitat/actuator.py
```diff
@@ -66,170 +65,119 @@ def action_name(self, action: Action) -> str:
         """
         return f"{action.agent_id}.{action.name}"
 
-    def actuate(self, action: Action, parameterizer: Type[HabitatParameterizer]):
```
vkakerbeck, line 69: What's the reason we can just remove the general `actuate` function here?

vkakerbeck (approved):
Nice changes & tutorial! This helped me a lot understanding the new structure & also the refactor looks great for adding more flexibility for future Mnty instantiations :) 
I didn't notice any issues with these changes and just left a few suggestions and questions. I hope you don't mind my naive questions on some of the refactor changes. It's nothing about your PR, I just want to make sure I understand everything that is happening.

docs/how-to-use-monty/tutorials/implementing-actions.md
```diff
@@ -0,0 +1,161 @@
...
+If at any point this action will be read from a file, then be sure to update the `ActionJSONDecoder` as well:
+
+```python
+class ActionJSONDecoder(JSONDecoder):
+
+    # ...
+
+    def object_hook(self, obj: dict[str, Any]) -> Any:
+        # ...
+        elif action = Jump.action_name():
```
jeremyshoemaker: Are you doing a comparison here, or trying to do an assignment with a check? The first should be `==` and the second, I believe, should be `:=`.

src/tbp/monty/frameworks/actions/actions.py
jeremyshoemaker: It looks like they're implemented in the Sampler concrete class(es). 

What seems to have happened here is that the various `sample_*` methods have been split out of the single ActionSampler ABC into direction-specific protocols. If I understand correctly, the idea is to make it so you only have to implement the ActionSampler protocol methods you care about, and not all of them, like you had to previously with the ABC.

src/tbp/monty/simulators/habitat/actuator.py
jeremyshoemaker: I _think_ this is just a rename from `actuate` to `to_habitat`. All of the `actuate_*` methods below were changed to call this method instead of `actuate`.

docs/how-to-use-monty/tutorials/implementing-actions.md
jeremyshoemaker: Looking at the actual code, I believe you meant this to be `==`.

docs/how-to-use-monty/tutorials/implementing-actions.md
tristanls: Protocol definitions do not contain the implementation body. The ellipsis `...` is syntactically correct Python and is how someone's implementation would look. For example, this is how our `LookDown` implementation looks now:

```python
class LookDownActionSampler(Protocol):
    def sample_look_down(self, agent_id: str) -> LookDown: ...


class LookDownActuator(Protocol):
    def actuate_look_down(self, action: LookDown) -> None: ...


class LookDown(Action):
    """Rotate the agent downwards by a specified number of degrees."""

    @staticmethod
    def sample(agent_id: str, sampler: LookDownActionSampler) -> LookDown:
        return sampler.sample_look_down(agent_id)

    def __init__(
        self, agent_id: str, rotation_degrees: float, constraint_degrees: float = 90.0
    ) -> None:
        super().__init__(agent_id=agent_id)
        self.constraint_degrees = constraint_degrees
        self.rotation_degrees = rotation_degrees

    def act(self, actuator: LookDownActuator) -> None:
        actuator.actuate_look_down(self)
```

> It could also be useful to explain why we need a separate Sampler class instead of just implementing the sampling code in the sample function of Jump. 

You can implement sampling code in the sample function of `Jump`. 

I added a paragraph to the beginning of the section to add more context based on your feedback.

docs/how-to-use-monty/tutorials/implementing-actions.md
tristanls: This wasn't meant to be a protocol definition. I updated it to

```python
class MyActuator:

    def actuate_jump(self, action: Jump) -> None:
        # custom code to make Jump happen
        pass
```

This being a general guide, implementing actuation could be anything. For an example implementation, readers are directed to the habitat actuator at the bottom of the section.

src/tbp/monty/frameworks/actions/actions.py
tristanls: Yes to what @jeremyshoemaker said. The `ConstantSampler` and `UniformlyDistributedSampler` are the ones that implement the `sample_look_down` method.

src/tbp/monty/simulators/habitat/actuator.py
tristanls: Yes to what @jeremyshoemaker said. This method was only called by `actuate_*` methods. However, it violated some design principles, so it got reworked into `to_habitat`, which is also only called by the `actuate_*` methods.

docs/how-to-use-monty/tutorials/implementing-actions.md
tristanls: Thank you, yes, this should be `==`.

docs/how-to-use-monty/tutorials/implementing-actions.md
vkakerbeck: Okay yeah I realized that later in the PR review and forgot to remove this comment. I haven't worked with protocols like this before but that makes sense :)

docs/how-to-use-monty/tutorials/implementing-actions.md
vkakerbeck: I think if the intent is for it to be a general guide the I would put it into the general "how to use Monty' section and not in the tutorials subsection. So far all the tutorials have concrete follow-along code that implements specific examples and illustrates how something is done by providing code that you can copy and run.

src/tbp/monty/frameworks/actions/actions.py
vkakerbeck: okay, thanks for the explanation!

docs/how-to-use-monty/tutorials/implementing-actions.md
tristanls: That's a good point, I'll move this over.

tristanls: https://github.com/thousandbrainsproject/tbp.monty/pull/334

PR #328, opened 2025-06-16
RFC: Efficient Saccades Driven by Model-Free and Model-Based Signals

scottcanoe:
**Note**: This RFC has been superseded by [RFC: Extended Goal State Functionality](https://github.com/thousandbrainsproject/tbp.monty/pull/376).

## Summary
This RFC proposes changes to Monty’s architecture aimed at improved support for compositional objects and multi-object environments. More specifically, this RFC addresses the intermediate goal of [implementing efficient saccades driven by model-free and model-based signals](https://thousandbrainsproject.readme.io/docs/implement-efficient-saccades-driven-by-model-free-and-model-based-signals).

## Notes
 - Early planning and discussions for this RFC can be viewed [here](https://github.com/thousandbrainsproject/tbp.monty/pull/316).


rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
+- Start Date: 2025-06-16
+- RFC PR: 
+
+# Summary
+This RFC proposes changes to Monty’s architecture designed to improve support for compositional objects and multi-object environments[^1]. More specifically, this document addresses the intermediate goal of [implementing efficient saccades driven by model-free and model-based signals](https://thousandbrainsproject.readme.io/docs/implement-efficient-saccades-driven-by-model-free-and-model-based-signals).
+
+In single-object, one-object-per-episode experiments, we make the following simplifying assumption -- all observations in an episode correspond to exactly one object (henceforth, "one-object criterion"). In this case, the following important conditions are are automatically guaranteed:
```
vkakerbeck:
```suggestion
In single-object, one-object-per-episode experiments, we make the following simplifying assumption -- all observations in an episode correspond to exactly one object (henceforth, "one-object criterion"). In this case, the following important conditions are automatically guaranteed:
```

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+  1. We propose a strategy for selecting saccade targets that are contained within a region estimated to satisfy the one-object criterion. This depends on first having reasonably good estimates of the locations and spatial extents of one-object regions. To accomplish this, we plan to apply model-free segmentation techniques to wide field-of-view imagery. Model-based signals may also be used to generate segmentation maps de novo and/or refine model-free maps, but this topic is currently marked as an [open question](#open-questions).
+  2. In addition, we propose a set of optimization strategies designed to improve efficiency. In this context, efficiency refers to the number of steps taken to converge on an object + pose estimate. While these optimization strategies should improve inference speed in any setting (single-object, multi-object, etc.), faster inference may serve an additional purpose in the compositional/multi-object context; by reducing the number of steps needed to recognize an object, we may also reduce the likelihood of collecting mislabeled, off-object observation[^3].
```
vkakerbeck: If there is an order of importance to these I would put this one first. I would start with the saliency map for more efficient convergence. Then in a second step I would see if we need to add an additional segmentation mechanism.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+
+To implement these strategies, we propose the addition of two new components.
+  - `SalienceMapSM`: a sensor module type that receives a wide field-of-view, performs model-free processing (e.g., segmentation salience estimation), and outputs a set of CMP-compliant goal states.
```
vkakerbeck:
```suggestion
  - `SalienceMapSM`: a sensor module type that receives a wide, low-resolution field-of-view, performs model-free processing (e.g., salience estimation or segmentation), and outputs a set of CMP-compliant goal states.
```

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+  - `GoalStateSelector`: an arbiter that receives goal states from all sources (LMs and `SalienceMapSM`)and outputs a single goal state for the motor system.
+
+# Architecture
+
+![Information Flow](efficient_saccades/flow.png)
```
vkakerbeck, line 25: this is not as important but in the goal state box from the new SM I would say "# of goal states *can be* large" or just leave that out. I am not sure we would actually want to output a lot of goalstates here. I think we will have to experiment if it makes more sense for the SM to do some preselection like only outputting the most salient locations. Similarly the learning modules could output multiple goal states.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+**Proposed Routing Pathways**. LM- and `SalienceMapSM`-derived goal states will be routed to the  `GoalStateSelector`, and the `GoalStateSelector`'s output will be routed to the motor system. All other routing pathways are unchanged.
+
+### `SalienceMapSM`
+
+  - **Type**: `DetailedLoggingSM`
+  - **Input**: Wide field-of-view imagery, depth data, and each pixel's associated 3D coordinates. No special handling of input data is required beyond standard processing steps performed for all other sensor modules.
+  - **Output**: A set of CMP-compliant `GoalState` objects.
```
vkakerbeck, line 32: I wonder if it would make sense to implement the goal state generation capability in a more SM independent way. I could imagine that in the future we want to be able to connect an LM to the SM with the large receptive field and/or that we want the standard SMs to be able to output goalstates (maybe for small adjusting movements like micro saccades). 
Maybe @tristanls has a good suggestion of how we could implement the feature extraction capability (giving a CMP sensory signal for an LM) and the target location extraction (giving a CMP goal state for the action selector) in an independent way so they can be mixed flexibly. But we can definitely start prototyping with a separate SM like you suggest for now and then think about this later.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+
+### `GoalStateSelector`
+
+  - **Type**: `object`
+  - **Input**: All `GoalState` objects produced by LMs and the `SalienceMapSM`).
+  - **Output**: `GoalState`
+  - **Consumers**: `MotorSystem`  
+
+With the addition of a `SalienceMapSM`, the number of goal states generated at each time step will jump from ~1 to ~1000 (roughly). Consequently, we need something to pool LM- and SM-derived goal states together and decide which should be acted upon by the motor system. These requirements place the `GoalStateSelector` in a uniquely powerful position to integrate model-free and model-based data. In addition, it can easily maintain a recent history of visited (or at least attempted) locations since it has the final say about which goal state the motor system should attempt.
```
vkakerbeck, line 48: Why would we have ~1000 goal states? Would there be that many interesting locations? If we have a large low-res receptive field, lets say 64x64 pixels, there are only ~4k possible locations it could want to go to. Would we really want to output a 4th of those? I would try to keep this number small to start with.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+In its simplest/dumbest form, a `GoalStateSelector` could choose a goal state like so:
+  - If an LM emitted a goal state, use it. In the case of multiple LM-derived goal states, pick the one with the greatest confidence value. If there are ties, break them at random.
+  - If no LM has emitted a goal state, use one of the `SalienceMapSM`'s most salient goal state.
+
+If we take advantage of the `GoalStateSelectors`'s unique input/output values, we could instead do any or all of the following:
+  - If an LM emitted a goal state, use it. In the case of multiple LM-derived goal states, pick the one with the greatest confidence value. If there are ties, select the one with the highest expected salience.
+  - If no LM has emitted a goal state, focus on the `SalienceMapSM`'s goal state.
+    - Initialize some scoring variable for each goal state. Let's call it `priority`.
+    - Weight `priority` values based on `region_id` (e.g., make it zero or `nan` if it's on the wrong region, or simply filter out the off-object ones).
+    - Weight `priority` values by salience.
```
vkakerbeck: Why couldn't this be done inside the SM? I would have this filtering take place there and have this weighting/subset already be contained in what it sends to the goalstateselector.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+
+In its simplest/dumbest form, a `GoalStateSelector` could choose a goal state like so:
+  - If an LM emitted a goal state, use it. In the case of multiple LM-derived goal states, pick the one with the greatest confidence value. If there are ties, break them at random.
+  - If no LM has emitted a goal state, use one of the `SalienceMapSM`'s most salient goal state.
+
+If we take advantage of the `GoalStateSelectors`'s unique input/output values, we could instead do any or all of the following:
+  - If an LM emitted a goal state, use it. In the case of multiple LM-derived goal states, pick the one with the greatest confidence value. If there are ties, select the one with the highest expected salience.
```
vkakerbeck, line 55: This sounds good for a first start. Later we could test some more variations. Like if there was an LM goal state, check if there are nearby salient locations from the SM. Then move there. This would compensate a bit for noisy LM hypotheses and make sure we actually saccade onto an interesting feature on the object.
We could also see if it make sense to have an option where the SM outputs a goal state with extremely high salience that can outweigh the LM's goal state. Like when our attention is suddenly drawn to some movement or a loud noise.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+  - If no LM has emitted a goal state, focus on the `SalienceMapSM`'s goal state.
+    - Initialize some scoring variable for each goal state. Let's call it `priority`.
```
vkakerbeck: We already have the `confidence` attribute in each CMP message. Why not use that?

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+    - Weight `priority` values based on `region_id` (e.g., make it zero or `nan` if it's on the wrong region, or simply filter out the off-object ones).
+    - Weight `priority` values by salience.
+    - Weight `priority` values by the magnitude of the displacement it would generate. (bigger displacements => faster hypothesis elimination)
+    - Weight `priority` values by a goal state's distance from previously visited areas.
+
+Some of these rules help satisfy the one-object criterion, and others support efficiency. There are a lot of things we could try here, and we won't know what works until we do. Other potential rules are left as an exercise for the reader.
```
vkakerbeck, line 63: I think before we start to far down the implementation we should think about some good experiments and metrics that we want to use to evaluate different options. Like I mentioned before, I would start with efficiency. There we could simply take the existing benchmark experiments and see if we can reduce #of steps and increase accuracy. For staying on object it is a bit more tricky. You could start with the multi object benchmark experiment (potentially adding one or two extra metrics) or we would have to make a new testbed for that.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Integration
+
+### Executing Goal States
+This RFC would move us down a very goal-state-heavy path. It's easy to imagine that a distant agent performing inference will produce achievable goal states at every step. This is a pretty big departure from Monty's existing behavior in which goal states are only occasionally generated and acted upon. The remainder of the time, actions are derived from model-free policies.
```
vkakerbeck, line 68: I think it will be important that not every policy will execute goal states at every step (and also not every SM or LM will output goalstates at every step). A policy like the surface agent could use its curvature following actions for the majority of steps and only consider a goal state if it has particularly high confidence or we haven't used one in a while. The distant agent could use goal states almost exclusively.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+
+Implementing this RFC is going to require at least some revisions to our goal-state execution framework. First, we
+only have one goal-state-attempter -- `InformedEnvironmentDataLoader.execute_jump_attempt` -- and has only one strategy. All attempts consist of an agent-pose and sensor-orientation setting actions. This RFC entails that most of goal states will have been produced by the `SalienceMapSM`, and are therefore achievable through saccades alone. At a mimimum, we should build out some system that's capable of deciding whether a saccade is sufficient or whether it should perform pose-and-orientation resets. Second, I believe there are existing plans (desires?) to move `execute_jump_attempt` out of `InformedEnvironmentDataLoader`. I'm assuming it's headed for a `MotorPolicy`, but I don't really know.
```
vkakerbeck: Those are good points. This will be a pretty involved but useful change.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+
+Also, I'd like to make one thing clear: The `MotorSystem`/`MotorPolicy` situation is probably my biggest Monty blindspot. I'm also aware that Tristan has been actively refactoring `Motor*` code, which means I'm at least as ignorant about the future as I am the present.
+
+### Modifying Monty Classes
+The `MontyBase` or one of its subclass will need a `goal_state_selector` attribute. Which subclass? TBD.
+  
+`SalienceMapSM` is sensor module, so it can be added to `Monty`, and it will be called like any other. But we will need to implement the passing of its up new routing pathways.
+
+**Routing**
+  - Add: `SalienceMapSM` ->  `GoalStateSelector`.
```
vkakerbeck, line 81: I think this should be more general. Any SM that emits a goal state should have it sent to the GoalStateSelector, not just this specific subclass.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+  - Add: LM/GSG -> `GoalStateSelector`.
+  - Add: `GoalStateSelector` -> `MotorSystem`.
+  - Remove: LM/GSG -> `MotorSystem`. See `MontyForEvidenceGraphMatching._pass_infos_to_motor_system`.
+  
+**Potential issue**
+  - All `MontyBase` subclasses appear to have the same set of attributes. I don't know how much code assumes that this is the case. If `MontyExperiment` classes assume a fixed set of attributes for initialization and logging, there could be a few pain points.
+
+# Open Questions
+ - Can/should we integrate model-based signals to inform segmentation or region selection?
```
vkakerbeck, line 104: I was thinking of them implicitly containing this information. If the LM hasn't confidently recognized an object yet or is learning more about it, it will output goal states that keep it on the object (using the model for its mlh). If it is done recognizing a model it can use it's knowledge of the object's shape to output a goalstate that take it off that object. No need to do any segmentation for this.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+ - Which segmentation methods, or combination thereof, might work well for multi-object environments (where objects do not share space) and compmositional objects (where objects *do* share space). For example, depth-based segmentation will be no help in separating a logo from the mug it is on.
```
vkakerbeck:
```suggestion
 - Which segmentation methods, or combination thereof, might work well for multi-object environments (where objects do not share space) and compositional objects (where objects *do* share space). For example, depth-based segmentation will be no help in separating a logo from the mug it is on.
```

vkakerbeck: I think for color we will probably be able to use a similar mechanism as for depth. Basically, detecting sudden shifts in the feature distribution as borders of objects.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+
+Finally, the SC is thought to be THE place where the saccade command are ultimately issued.
+
+The system proposed here doesn't have an exact 1:1 relationship with the superior colliculus, but they do rhyme. The `SalienceMapSM` resembles the visual-only superficial layer of the SC, while the `GoalStateSelector` is perhaps most similar to the SC's intermediate layer. The `GoalStateSelector` also resembles the SC in they both act as the input to the motor system.
+
+Note: there is one part of our code that already implements SC-like behavior -- `GetGoodView`. This code uses raw sensor data and coordinates obtained via `DepthTo3DLocations`. With this data, it decides where it ought to look, and computes the necessary "look" actions used directly by the motor system. Niels has proposed reworking `GetGoodView` as an intermediate step to guide the development of this new system. I haven't given it much thought yet, but given that `GetGoodView` employs both translating and orienting actions, it could be an interesting spot to explore how motor systems decide which actions to perform to achieve a goal state.
```
vkakerbeck, line 116: That's a good idea

rfcs/efficient_saccades.md
ramyamounir:
```suggestion
  2. In addition, we propose a set of optimization strategies designed to improve efficiency. In this context, efficiency refers to the number of steps taken to converge on an object + pose estimate. While these optimization strategies should improve inference speed in any setting (single-object, multi-object, etc.), faster inference may serve an additional purpose in the compositional/multi-object context; by reducing the number of steps needed to recognize an object, we may also reduce the likelihood of collecting mislabeled, off-object observations[^3].
```

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+  2. In addition, we propose a set of optimization strategies designed to improve efficiency. In this context, efficiency refers to the number of steps taken to converge on an object + pose estimate. While these optimization strategies should improve inference speed in any setting (single-object, multi-object, etc.), faster inference may serve an additional purpose in the compositional/multi-object context; by reducing the number of steps needed to recognize an object, we may also reduce the likelihood of collecting mislabeled, off-object observation[^3].
+
+To implement these strategies, we propose the addition of two new components.
+  - `SalienceMapSM`: a sensor module type that receives a wide field-of-view, performs model-free processing (e.g., segmentation salience estimation), and outputs a set of CMP-compliant goal states.
+  - `GoalStateSelector`: an arbiter that receives goal states from all sources (LMs and `SalienceMapSM`)and outputs a single goal state for the motor system.
```
ramyamounir:
```suggestion
  - `GoalStateSelector`: an arbiter that receives goal states from all sources (LMs and `SalienceMapSM`) and outputs a single goal state for the motor system.
```

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+
+# Architecture
+
+![Information Flow](efficient_saccades/flow.png)
+**Proposed Routing Pathways**. LM- and `SalienceMapSM`-derived goal states will be routed to the  `GoalStateSelector`, and the `GoalStateSelector`'s output will be routed to the motor system. All other routing pathways are unchanged.
```
ramyamounir, line 26: Can you add a brief description, in the caption or in a legend, for what different line colors (i.e., pink vs. blue) and types (e.g., dashed) mean in this context?

rfcs/efficient_saccades.md
ramyamounir: This is probably thinking too far ahead but my understanding is that we would ideally have one GSS for each agent. I can imagine a scenario where two or more groups of LMs send goal states to different selectors which control different agents. I only bring this up because it could be worth thinking about how to define which LMs send to which selector and which selector talks to which agent. I also think in the configs we would define a list of selectors not one, even if we only use one for now, just like we have a list of agents but only support one for now.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+  - Salience Detection: To support rapid inference by targeting an object's more informative areas.
+
+The results of these processes may be (a) used to filter out goal states and/or (b) embedded into `GoalState` objects. On the one hand, filtering is pretty unambiguous way to indicate whether a target location is worth considering at all. On the other, filtering out goal-states based on model-free criteria might prematurely eliminate candidates that would have been very promising had they been allowed to mix with model-based information. At this stage, I'm learning towards minimal output filtering since (a) we want to avoid repeatedly saccading between a handful of salient locations, and (b) the `GoalStateSelector`'s filtration should be "smarter" since it has access to model-based information[^4]. Some metadata will be added each goal state's dictionary of non-morphological features (such as `region_id` and `salience`) which the `GoalStateSelector` can use to filter and rank goal states[^5]. 
+
+### `GoalStateSelector`
+
+  - **Type**: `object`
+  - **Input**: All `GoalState` objects produced by LMs and the `SalienceMapSM`).
```
ramyamounir:
```suggestion
  - **Input**: All `GoalState` objects produced by LMs and the `SalienceMapSM`.
```

ramyamounir (approved):
Really nice and very exciting! I agree that implementing this RFC will be quite involved especially with all the refactoring that's going on..

rfcs/efficient_saccades.md
nielsleadholm: @vkakerbeck is that actually as important for compositional objects? In the previous draft PR discussion, I wrote:

"Efficiency also ties in with compositional objects - in order to support compositional objects, we are likely going to give [Monty a shorter memory horizon for its internal representations](https://thousandbrainsproject.readme.io/docs/implement-and-test-rapid-evidence-decay-as-form-of-unsupervised-memory-resetting). While this will enable the representations to change when we move onto a new child object (desired property), this comes at a cost - if Monty does not quickly move to useful features within an object, it may never recognize an object before it has "forgotten" any distinguishing features it has observed. A better saccade policy is important to addressing this."

However, with Ramy's most recent results, it does not seem like Monty will need to be limited to having a very short memory horizon. While it would obviously be nice to have a faster, better saccade policy, it's not clear to me that it is as important to compositional objects anymore. Instead, my feeling is that having a bias to stay within a region of similar features (e.g. color, depth, texture frequency) until we have reasonable object hypotheses might be more important to this specific problem.

For what it's worth, if we have the segmentation (through model-free and model-based methods) working pretty well to stay on object, then we can easily set the random-walk saccade sizes within that region to be much larger, resulting in more efficient recognition.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+  - Segmentation: To support action policies that require staying on-object (or moving off-object).
+  - Salience Detection: To support rapid inference by targeting an object's more informative areas.
+
+The results of these processes may be (a) used to filter out goal states and/or (b) embedded into `GoalState` objects. On the one hand, filtering is pretty unambiguous way to indicate whether a target location is worth considering at all. On the other, filtering out goal-states based on model-free criteria might prematurely eliminate candidates that would have been very promising had they been allowed to mix with model-based information. At this stage, I'm learning towards minimal output filtering since (a) we want to avoid repeatedly saccading between a handful of salient locations, and (b) the `GoalStateSelector`'s filtration should be "smarter" since it has access to model-based information[^4]. Some metadata will be added each goal state's dictionary of non-morphological features (such as `region_id` and `salience`) which the `GoalStateSelector` can use to filter and rank goal states[^5].
```
nielsleadholm:
```suggestion
The results of these processes may be (a) used to filter out goal states and/or (b) embedded into `GoalState` objects. On the one hand, filtering is pretty unambiguous way to indicate whether a target location is worth considering at all. On the other hand, filtering out goal-states based on model-free criteria might prematurely eliminate candidates that would have been very promising had they been allowed to mix with model-based information. At this stage, I'm learning towards minimal output filtering since (a) we want to avoid repeatedly saccading between a handful of salient locations, and (b) the `GoalStateSelector`'s filtration should be "smarter" since it has access to model-based information[^4]. Some metadata will be added to each goal state's dictionary of non-morphological features (such as `region_id` and `salience`) which the `GoalStateSelector` can use to filter and rank goal states[^5]. 
```

nielsleadholm: I would be tempted to re-use goal-state variables we already have, like `object_id` (which would just be a random number/string that could change each time the SM sends information, but which would be shared by goal-states in that time step), and `confidence` instead of `salience`. If necessary, we can update `confidence` to be called `confidence_priority_factor` or something like that in all the code, but I think we want to avoid proliferating different non-pose features unless absolutely necessary.

rfcs/efficient_saccades.md
nielsleadholm: I would agree with it not necessarily needing to be 1000s (or even dozens of goal-states) from the SM. However I also agree with Scott's current draft that we try to leave as much of the sub-selection to the action-selector as possible. The SM's job would be to measure salience, but not to filter what we pass on, as that tries to put too much policy into the SM, since under different conditions, we may choose to focus on salience in different ways.

rfcs/efficient_saccades.md
nielsleadholm: See separate comment under the "this is not as important but in the goal state box from the new SM I would say "# of goal states can be large" or just leave that out" comment.

rfcs/efficient_saccades.md
nielsleadholm:
```suggestion
only have one goal-state-attempter -- `InformedEnvironmentDataLoader.execute_jump_attempt` -- and it has only one strategy. All attempts consist of an agent-pose and sensor-orientation setting actions. This RFC entails that most of goal states will have been produced by the `SalienceMapSM`, and are therefore achievable through saccades alone. At a mimimum, we should build out some system that's capable of deciding whether a saccade is sufficient or whether it should perform pose-and-orientation resets. Second, I believe there are existing plans (desires?) to move `execute_jump_attempt` out of `InformedEnvironmentDataLoader`. I'm assuming it's headed for a `MotorPolicy`, but I don't really know.
```

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+
+Also, I'd like to make one thing clear: The `MotorSystem`/`MotorPolicy` situation is probably my biggest Monty blindspot. I'm also aware that Tristan has been actively refactoring `Motor*` code, which means I'm at least as ignorant about the future as I am the present.
+
+### Modifying Monty Classes
+The `MontyBase` or one of its subclass will need a `goal_state_selector` attribute. Which subclass? TBD.
+  
+`SalienceMapSM` is sensor module, so it can be added to `Monty`, and it will be called like any other. But we will need to implement the passing of its up new routing pathways.
```
nielsleadholm:
```suggestion
`SalienceMapSM` is a sensor module, so it can be added to `Monty`, and it will be called like any other. But we will need to implement the passing of its new routing pathways.
```

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+ - Salience Map: The visual-only superficial layer (SCs) of the superior colliculus is thought to generate maps based on features such as contrast, color opponency, and motion. Salience maps might help Monty move to areas that are more informative than a random-walk policy would. At a minimum, they should at least guarantee that the next viewed location is on some object (or very nearly so).
+ - Priority Map: The intermediate layer (SCi) is believed to generate a priority map that reflects an organism's goals. Unlike the visual-only layer, the intermediate layer receives top-down and multimodal input. In Monty, a priority map may reflect the general regions we would like to keep our saccades within.
+
+Finally, the SC is thought to be THE place where the saccade command are ultimately issued.
+
+The system proposed here doesn't have an exact 1:1 relationship with the superior colliculus, but they do rhyme. The `SalienceMapSM` resembles the visual-only superficial layer of the SC, while the `GoalStateSelector` is perhaps most similar to the SC's intermediate layer. The `GoalStateSelector` also resembles the SC in they both act as the input to the motor system.
```
nielsleadholm:
```suggestion
The system proposed here doesn't have an exact 1:1 relationship with the superior colliculus, but they do rhyme. The `SalienceMapSM` resembles the visual-only superficial layer of the SC, while the `GoalStateSelector` is perhaps most similar to the SC's intermediate layer. The `GoalStateSelector` also resembles the SC in that they both act as the input to the motor system.
```

rfcs/efficient_saccades.md
nielsleadholm: Yeah I would almost suggest that a third component that this RFC should implement, and perhaps the first one to work on, would be to implement an `OrientToGoalState` policy or the like in the `MotorSystem`, and a `LookForObject` in the `ActionSelector`, which can call `OrientToGoalState` .

Note this would be separate from the `GetGoodView` _positioning procedure_, which should only be used by experimenters to setup an episode. As such, any references to `GetGoodView` below is not talking about the positioning procedure, but rather how we currently use `GetGoodView` to look at an object again if we have lost visualization of it, after an episode has started.

This change would basically involve:

- Move all the sensory processing currently taking place in `GetGoodView` into a separate SM. This should output a goal-state (or states) corresponding to an object in the visual field.
- Refactor `GetGoodView` as a meta policy e.g. `LookForObject`, probably located within `ActionSelector`, that gets goal-states from the new SM (including whether they are on or off object), and passes a single goal-state to the motor system. This way `LookForObject` can also use model-based goal-states to look at an object, e.g. if we remember seeing something somewhere.
- Implement `OrientToGoalState` as a _motor-system_ policy that can figure out what rotation etc. is required to orient to a goal-state. This would be similar in flavor to `orient_to_object`, but would make use of a received goal-state, as well as the agent's current state.
 
It would be worth discussing as a group whether we want to have the possibility of policies in both `ActionSelector`, and within `MotorSystem`, as the above assumes. What I'm currently thinking is that the former could be policies that orchestrate what goal-state to act on, such as prioritizing getting a view of an object. The latter would be simpler policies that can only receive a single goal-state, and figure out how to carry it out.

Tagging @tristanls since this relates to a lot of what you have been working on.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+  - Remove: LM/GSG -> `MotorSystem`. See `MontyForEvidenceGraphMatching._pass_infos_to_motor_system`.
+  
+**Potential issue**
+  - All `MontyBase` subclasses appear to have the same set of attributes. I don't know how much code assumes that this is the case. If `MontyExperiment` classes assume a fixed set of attributes for initialization and logging, there could be a few pain points.
+
+# Open Questions
+ - Can/should we integrate model-based signals to inform segmentation or region selection?
+ - Which segmentation methods, or combination thereof, might work well for multi-object environments (where objects do not share space) and compmositional objects (where objects *do* share space). For example, depth-based segmentation will be no help in separating a logo from the mug it is on.
+ 
+# Appendix: Superior Colliculus
```
nielsleadholm, line 107: Love the neuroscience context - since adding `ActioSelector` is a pretty big change, I think it could also be nice to discuss how this will likely find future uses that relate to the role of the basal ganglia.

rfcs/efficient_saccades.md
nielsleadholm: Yeah I think we should definitely have some mention of combining the influence of the SM and LM goal states.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
+- Start Date: 2025-06-16
+- RFC PR: 
+
+# Summary
+This RFC proposes changes to Monty’s architecture designed to improve support for compositional objects and multi-object environments[^1]. More specifically, this document addresses the intermediate goal of [implementing efficient saccades driven by model-free and model-based signals](https://thousandbrainsproject.readme.io/docs/implement-efficient-saccades-driven-by-model-free-and-model-based-signals).
+
+In single-object, one-object-per-episode experiments, we make the following simplifying assumption -- all observations in an episode correspond to exactly one object (henceforth, "one-object criterion"). In this case, the following important conditions are are automatically guaranteed:
+  1. Objects models are stored in their own reference frames.
```
hlee9212: **typo:** Object models

rfcs/efficient_saccades.md
hlee9212: I have the same question of whether there was some back-of-the-envelop calculations for where ~1000 goal states came from.

rfcs/efficient_saccades.md
hlee9212: Just throwing out a thought - maybe a metric or a way to visualize staying on object is plot the state (patch off object, on object one, on object two) as a function of steps, similar to Ramy's evidence values over steps. Maybe even a single number, like % on object 1 or % on object 2 (or just % on target object if multiple objects present) could be a good metric, though it may not capture the whole story.

hlee9212:
Wow, thanks for explaining things very clearly!

Some additional resources for Saliency in OpenCV: https://pyimagesearch.com/2018/07/16/opencv-saliency-detection/ (has implementation to take advantage of three different types of saliency, i.e. static saliency, motion saliency, and "objectness").

Regarding segmentation: A non-DL method that is well known is called GrabCut. Basically Saliency map can help initialize GrabCut. GrabCut will output graphs (where nodes are pixels), and one can do connected component analysis to identify potentially different objects (if multiple graphs, or connected components, are present). 

Also, I think this paper is very relevant for both saliency and segmentation: https://ieeexplore.ieee.org/document/5206596 and one of the StaticSaliency methods that OpenCV implements. 

(Just looked at the paper and it was like a review of my Computer Vision class, haha.)

rfcs/efficient_saccades.md
jeremyshoemaker: Don't forget you can use the "suggestion" feature to give the author a change they can apply from the GitHub UI instead of having to go back to their editor and do it.
```suggestion
  1. Object models are stored in their own reference frames.
```

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+
+Finally, the SC is thought to be THE place where the saccade command are ultimately issued.
+
+The system proposed here doesn't have an exact 1:1 relationship with the superior colliculus, but they do rhyme. The `SalienceMapSM` resembles the visual-only superficial layer of the SC, while the `GoalStateSelector` is perhaps most similar to the SC's intermediate layer. The `GoalStateSelector` also resembles the SC in that they both act as the input to the motor system.
+
+Note: there is one part of our code that already implements SC-like behavior -- `GetGoodView`. This code uses raw sensor data and coordinates obtained via `DepthTo3DLocations`. With this data, it decides where it ought to look, and computes the necessary "look" actions used directly by the motor system. Niels has proposed reworking `GetGoodView` as an intermediate step to guide the development of this new system. I haven't given it much thought yet, but given that `GetGoodView` employs both translating and orienting actions, it could be an interesting spot to explore how motor systems decide which actions to perform to achieve a goal state.
+
+# Footnotes
```
jeremyshoemaker, line 118: This aren't getting rendered at the bottom of the page the way you might think they are.

In fact, you can inline the content of a footnote right after the paragraph where it is used (or anywhere else in the document) and they'll get pulled out and rendered at the bottom of the page.

I prefer that approach because it makes it easier to read the raw markdown when I don't have to scroll to the bottom to read the footnote's contents.

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+[^1]: For more discussion on how evidence drop-offs indicate changes to the sensed object's identity, see [RFC 9](https://github.com/thousandbrainsproject/tbp.monty/pull/196). Also see [here](https://thousandbrainsproject.readme.io/docs/use-off-object-observations) for a general discussion about how off-object observations may be used in the future.
```
jeremyshoemaker: This is more of a nitpick, but one I've tried to adhere to in my personal writing. A link with the text of "here" is somewhat less accessible (e.g. in screen readers) because it doesn't say anything about what the link is pointing to.

I might do something like:
```suggestion
[^1]: For more discussion on how evidence drop-offs indicate changes to the sensed object's identity, see [RFC 9](https://github.com/thousandbrainsproject/tbp.monty/pull/196). Also see ["Use Off-Object Observations"](https://thousandbrainsproject.readme.io/docs/use-off-object-observations) for a general discussion about how off-object observations may be used in the future. 
```

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+  1. We propose a strategy for selecting saccade targets that are contained within a region estimated to satisfy the one-object criterion. This depends on first having reasonably good estimates of the locations and spatial extents of one-object regions. To accomplish this, we plan to apply model-free segmentation techniques to wide field-of-view imagery. Model-based signals may also be used to generate segmentation maps de novo and/or refine model-free maps, but this topic is currently marked as an [open question](#open-questions).
+  2. In addition, we propose a set of optimization strategies designed to improve efficiency. In this context, efficiency refers to the number of steps taken to converge on an object + pose estimate. While these optimization strategies should improve inference speed in any setting (single-object, multi-object, etc.), faster inference may serve an additional purpose in the compositional/multi-object context; by reducing the number of steps needed to recognize an object, we may also reduce the likelihood of collecting mislabeled, off-object observations[^3].
```
jeremyshoemaker: This may just be me, but I think this reads better, so take it as a nitpick.
```suggestion
  2. In addition, we propose a set of optimization strategies designed to improve efficiency. In this context, efficiency refers to the number of steps taken to converge on an object and pose estimate. While these optimization strategies should improve inference speed in any setting (single-object, multi-object, etc.), faster inference may serve an additional purpose in the compositional/multi-object context; by reducing the number of steps needed to recognize an object, we may also reduce the likelihood of collecting mislabeled, off-object observations[^3].
```

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+`SalienceMapSM` is a sensor module, so it can be added to `Monty`, and it will be called like any other. But we will need to implement the passing of its new routing pathways.
+
+**Routing**
+  - Add: `SalienceMapSM` ->  `GoalStateSelector`.
+  - Add: LM/GSG -> `GoalStateSelector`.
+  - Add: `GoalStateSelector` -> `MotorSystem`.
+  - Remove: LM/GSG -> `MotorSystem`. See `MontyForEvidenceGraphMatching._pass_infos_to_motor_system`.
+  
+**Potential issue**
+  - All `MontyBase` subclasses appear to have the same set of attributes. I don't know how much code assumes that this is the case. If `MontyExperiment` classes assume a fixed set of attributes for initialization and logging, there could be a few pain points.
```
jeremyshoemaker, line 87: If they are a required part of the interface (or Protocol when we move to those) then I feel like the base class or protocol should define `@property` methods that subclasses implement/override. That decorator allows you to define getters and setters, and while they usually map to some underlying field, _because_ they're a part of the interface/protocol, defining them this way means that other classes can depend on them existing.

rfcs/efficient_saccades.md
vkakerbeck: Okay, we could also start with that if you think the efficient saccades are not required anymore. Although I did mostly read this RFC with that in mind. I think if we just care about staying on the object, there could be simpler solutions than segmentation. If we are not implementing larger saccades, that means we are still moving in a smooth random walk so it would be a much simpler solution to extend the off-object detection (leading to reversing the previous action) to other sudden feature changes besides depth. This would work without a large view-finder image and without calculating segmentation maps. Maybe I am missing something but to me that issue of needing segmentation maps to stay in a larger region only comes up after we introduce larger saccades to salient features.

I also think that while Ramy's results show pretty good performance on switching object hypotheses, the speed at which hypotheses are switched would probably be addressed more efficiently by implementing these saliency based saccades than by tweaking the hyperparameters of the hypothesis resampling.

I guess my main point is that the most intriguing part of the mechanism outlined in this RFC, to me, is the efficient movement to salient features and all the properties we would get from that (learning better models with fewer points, faster recognition, faster hypothesis switching, more robust recognition, potential for reanchoring hypotheses,...) But I understand if we want focus on the essentials for compositional objects and leave this to later work. In that case, I just wonder if a simpler mechanism for staying on the subobject until it is recognized is sufficient.

rfcs/efficient_saccades.md
vkakerbeck: I'm fine with starting with communicating many goal states for now to test this. But I am not sure it will be necessary. I don't see thresholding them as shifting policy-related responsibilities into the SM. I would rather see sending dense saliency maps to the motor system as shifting too much processing there.

rfcs/efficient_saccades.md
vkakerbeck: I like the idea of refactoring all our current motor related functions out of the dataloader and into the motor system with help of goal states from the SM.
I don't understand why we need policies in the Action Selector. How I am thinking of the action selector is a simple max function over all the goal states it receives. It may have some other basic heuristics like prioritizing LMs but it may not. I would be very hesitant to put any kind of processing or policy logic into the action selector. It seems like this will quickly move responsibilities that should be in the SM, LM or motor system there. 
What would be something that only the action selector can do? From what you describe it sounds like you want it to do further processing that should be happening in the SM. Like why would the SM send goal states that are off the object? It seems more like the GSG in the LM and SM could have different "policies" that change what goal states they emit. Maybe I am missing something?

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+  - **Type**: `object`
+  - **Input**: All `GoalState` objects produced by LMs and the `SalienceMapSM`.
+  - **Output**: `GoalState`
+  - **Consumers**: `MotorSystem`  
+
+With the addition of a `SalienceMapSM`, the number of goal states generated at each time step will jump from ~1 to ~1000 (roughly). Consequently, we need something to pool LM- and SM-derived goal states together and decide which should be acted upon by the motor system. These requirements place the `GoalStateSelector` in a uniquely powerful position to integrate model-free and model-based data. In addition, it can easily maintain a recent history of visited (or at least attempted) locations since it has the final say about which goal state the motor system should attempt.
+
+In its simplest/dumbest form, a `GoalStateSelector` could choose a goal state like so:
+  - If an LM emitted a goal state, use it. In the case of multiple LM-derived goal states, pick the one with the greatest confidence value. If there are ties, break them at random.
+  - If no LM has emitted a goal state, use one of the `SalienceMapSM`'s most salient goal state.
```
tristanls, line 52: **question**: How does `GoalStateSelector` know that the goal state came from `SalienceMapSM`? Should it?

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+
+If we take advantage of the `GoalStateSelectors`'s unique input/output values, we could instead do any or all of the following:
+  - If an LM emitted a goal state, use it. In the case of multiple LM-derived goal states, pick the one with the greatest confidence value. If there are ties, select the one with the highest expected salience.
+  - If no LM has emitted a goal state, focus on the `SalienceMapSM`'s goal state.
+    - Initialize some scoring variable for each goal state. Let's call it `priority`.
+    - Weight `priority` values based on `region_id` (e.g., make it zero or `nan` if it's on the wrong region, or simply filter out the off-object ones).
```
tristanls: **question**: How does `GoalStateSelector` know that a salience goal state is off-object?

rfcs/efficient_saccades.md
```diff
@@ -0,0 +1,111 @@
...
+    - Weight `priority` values by salience.
+    - Weight `priority` values by the magnitude of the displacement it would generate. (bigger displacements => faster hypothesis elimination)
+    - Weight `priority` values by a goal state's distance from previously visited areas.
```
tristanls: **question**: This seems to imply that the `GoalStateSelector` has memory. How does it know about previously visited areas?

tristanls (commented):
Sorry, running out of time today, but wanted to share at least some notes and questions.

**note**: The proposed architecture highlights the generally present synchronization problem. Specifically, in the context of this RFC, the path from `SalienceMapSM` to the `GoalStateSelector` will likely take less time than the path through `SensorModule` and `LearningModule` to the `GoalStateSelector`. Probably no need to address it yet, as I assume we will continue synchronizing on loop iterations for a while still.

**question**: What is the relationship between `GoalStateSelector` and an agent? one-to-one? one-to-many? many-to-one? I'm guessing the Superior Colliculus analogues would be one `GoalStateSelector` to two agents (two eyes).

**question**: How would a `GoalStateSelector` wired to two agents tell apart goal states generated for different agents (e.g., left-eye vs right-eye goal state)?

**note**: The relationship between `GoalStateSelector` and agents implies global (all Monty entities that wire into that `GoalStateSelector`) coupling to agents. This requires thinking through a managable way of configuring. Perhaps we should start thinking of addtional segmentation of Monty into areas/sectors depending on what they are wired to.

rfcs/efficient_saccades.md
nielsleadholm: Yeah I agree there are a lot of interesting benefits with saccading to salient features. I think it will be clearer when we rework the research objectives, which I think will clarify that both salient features and segmentation are worth pursuing, and then we can decide which one we focus on first.

rfcs/efficient_saccades.md
nielsleadholm: I think it would be helpful in the next research meeting for us to talk about where we want policies to exist in SM vs. Action Selector, because I think our philosophy on this is behind the main remaining open questions.

vkakerbeck:
That's a good point about synchronization. I think it will be interesting to see that once we move to a more asynchronous way of communicating information, model-free policies will be executed faster than model-based ones (similar to biology).

Regarding the goal state selector to agent mapping, I would think of it as a 1:1 mapping. With the two eyes, I actually conceptualize them more as one agent since the eyes always move together. You can't send one goal state to one eye and another to the other eye and execute both of them. If they could do that, then they would be two agents, and each of them would have a goal state selector. 
We will get into some complications here once agents have dependencies but are also not fully dependent (like the fingers on a hand can move relatively independently but depend on where the hand overall is moving. And the eyes can only do so much when the head is fixed. I think such hierarchical agent relationships will need to be represented eventually, but for now, Monty only supports one agent anyway, so maybe not as relevant to figure out for this RFC.

rfcs/efficient_saccades.md
nielsleadholm: All `State` objects, including `GoalState`, have a `sender_id` feature. In Monty, this is a unique ID indicating the source of the CMP signal. In biology, this would be hard-coded in the anatomical connectivity, i.e. if a signal came from a particular neuron's axon, then it is the same source as all previous signals from that axon.

rfcs/efficient_saccades.md
nielsleadholm: For any segmentation, we were discussing that we would probably use the `object_id`, non-morphological part of the CMP to encode a temporary mask for e.g. different objects. E.g. this region is "object_1", and this region is "object_2", but there is no persistence or meaning to these labels across saccades, so they are not a form of model-based representation. We probably wouldn't explicitly encode off-object regions, but if we did, I imagine we would use that here as well.

rfcs/efficient_saccades.md
nielsleadholm: We discussed having a very simple form of short-term memory, basically a trace, to reduce the chance of returning to certain areas after a short period. There is evidence that this exists in humans (https://pubmed.ncbi.nlm.nih.gov/25219515/). However, this would not be building models, and so would not be a form of model-based learning.

rfcs/efficient_saccades.md
tristanls: I think there is an important distinction between knowing that something came from the same sender and knowing the sender's identity. Is the `GoalStateSelector` deciding to "pick a goal state from senders", to "pick a goal state from the sender that consistently gives me goal states", or to "pick a goal state specifically from the SalienceMapSM sender"?

If I understand correctly, the setup here would specifically identify `SalienceMapSM` as a distinct type of sender from other LM senders. So, we would configure `GoalStateSelector` and explicitly give it `SalienceMapSM` sender_id to use when other sender_ids don't send anything.

rfcs/efficient_saccades.md
scottcanoe: This is a good point. I'll keep this in mind as I work with `Monty` classes on this project, and hopefully we can put together a protocol or similar and get it adopted.

rfcs/efficient_saccades.md
scottcanoe: Well that's a bummer. I had a hell of a time figuring out how to get footnotes working, and I thought I at least had it working for VSCode markdown preview. I'll have to follow up with your or google around to figure out how to inline the footnotes.

rfcs/efficient_saccades.md
scottcanoe: So I haven't weighed in this much lately because I find it difficult to picture ahead of time how efficient saccades and stay-on-object behaviors can work without each other in compositional or multi-object settings. My uncertainty is compounded by the fact that I haven't gotten a chance to work with Ramy's new hypothesis resampling stuff, so how that changes things is also a question mark. While I'd love to have this issue really well defined and decide which to prioritize right now, I tend to agree with @nielsleadholm last comment suggesting we defer until a little later in the process. I think the `GetGoodView` refactor will be a good opportunity to focus on salience, so that will likely be the first thing we tackle.

rfcs/efficient_saccades.md
scottcanoe: All good points. One of the reasons I've been leaning towards many goal states is that the `GoalStateSelector` is going to be downgrading previously visited goal states. If the SM only emits 5 goal states, they may all be visited after 5 steps. Maybe that's fine, and we just let a model-free motor-only policy take over after that. But my original thought was that by sending many goal states, we would ensure that there is usually some goal state available that hasn't been visited yet.

rfcs/efficient_saccades.md
scottcanoe: This is an interesting idea, and it sounds like it could be pretty powerful but also tricky in terms of routing as you've suggested. I've included a bit of discussion on this in the Objectives and Discussion section.

rfcs/efficient_saccades.md
scottcanoe: I've added a bit of discussion about goal-state-generating SMs, but as its been pointed out, figuring out how to decide the routing could by tricky. This RFC could go two ways -- it could outline goal-state-generation for SMs and GoalStateSelectors in a very generic way, or we could keep it more constrained to the particular motivating problem. I suppose I had written this up with the particular topic in mind, but it does seem like we're opening up a lot of discussion here about Monty's architecture more generally.

rfcs/efficient_saccades.md
scottcanoe: @hlee9212 and @vkakerbeck I'm just going to paste an answer in from a previous comment: 

"One of the reasons I've been leaning towards many goal states is that the GoalStateSelector is going to be downgrading previously visited goal states. If the SM only emits 5 goal states, they may all be visited after 5 steps. Maybe that's fine, and we just let a model-free motor-only policy take over after that. But my original thought was that by sending many goal states, we would ensure that there is usually some goal state available that hasn't been visited yet."

I'm certainly not married to any particular number or even order of magnitude. This is just one of those issues I had in the back of my mind. Also I forgot to mention in the above comment that I'm a bit worried that emitting few goal states might result in getting trapped in a cycle where we only visit the small handful of salient points.

rfcs/efficient_saccades.md
scottcanoe: @vkakerbeck I agree, I think we could make this as sophisticated as we like. And I can definitely see motion-based saliency in particular overriding an LM's goal state.

rfcs/efficient_saccades.md
scottcanoe: @vkakerbeck Since it looks like we're starting with a GetGoodView-like procedure, I think we'll be able to assess efficiency directly with existing benchmarks. Multi-object benchmarks would be the next logical benchmark for the second of our objectives (see the new Objectives section).

@hlee9212 Good ideas -- I think we may have talked about this before, but this is going to be really hard with the logo which is mostly empty space (and is therefore the parent object). Just putting that out there in case you have any ideas on how to handle that. Unfortunately, we don't have any way to specify that logos have a different semantic ID in that case since the parent object's texture has to contain the child object's image. I hope that's clear, it's hard to describe.

rfcs/efficient_saccades.md
scottcanoe: Yep, will do. I'll get a better version of this figure going soon.

rfcs/efficient_saccades.md
scottcanoe: Fair point. I've adjusted the document to reuse `object_id` and `confidence`. There's no reason to make things complicated right now, but who knows, maybe in the future there will be some reason to have separate categories for salience. For example, we may want tunable coefficients that weight salience vs confidence vs priority, etc.

rfcs/efficient_saccades.md
scottcanoe: We should maybe chat about the basal ganglia in a 1:1 or research meeting. I've thought about this comment a few times and clarity hasn't presented itself on this topic.

rfcs/efficient_saccades.md
scottcanoe: Commenting just to endorse @nielsleadholm's answer.

rfcs/efficient_saccades.md
scottcanoe: We could use `sender_id`, we could maybe have something like `sender_type`, but I think more likely we'll just embed that information in the `GoalState`'s `confidence` value. LMs will endow goal states with larger confidence values, and SMs will generate goal states with small ones. It'd be analogous to having proximal vs distal input. It's the simplest approach I can think of, but we can always make it more sophisticated if we need to.

rfcs/efficient_saccades.md
scottcanoe: I don't really have a great answer @tristanls. @nielsleadholm's comment is a good description of a model-free way, but we've also talked about using model-based signals from LMs to help. But how that would work is still pretty unclear to me.

rfcs/efficient_saccades.md
tristanls: For me they render how I expected

Screenshot 2025-07-10 at 18 28 50

rfcs/efficient_saccades.md
vkakerbeck: Yes I was thinking of this as a general change to Monty's architecture where we add the option for an SM to directly connect to the motor system and send goal states to it. If the SaliencyMapSM is a subclass of sensor module then I don't think we should give that capability only to the subclass. We would just have another routing matrix in the Monty class that specifies `sm_to_motor_system`. In the default setup, only the SaliencyMapSM would actually be connected to the motor system but it would be a general option for any SM

tristanls: I just want to drop in the reminder that we don't want to continue using inheritance for functionality reuse. So, no subclasses.

vkakerbeck: whether its a subclass or a type of SM following the SM Protocol, the way I am thinking about it, it would still be something specified in the `sensor_module_configs` dictionary and initialized in the same process as the other SMs instead of being treated as a new Monty component.

PR #339, opened 2025-06-20
refactor!: extract TouchObject positioning procedure

tristanls:
This pull request completes the extraction of the `TouchObject` positioning procedure from `SurfacePolicy`.

Now, when `SurfacePolicy.dynamic_call` cannot see the object, it raises `ObjectNotVisible` (instead of returning `None`), and the data loader will then invoke the `TouchObject` positioning procedure, placing Monty back on the object and returning the observation from the object.

```python
# DataLoader
def __next__(self):
    # ...
    try:
        self._action = self.motor_system()
    except ObjectNotVisible:
        self.touch_object(
            sensor_id="view_finder",
            desired_object_distance=self.motor_system._policy.desired_object_distance,
        )
        self.motor_system._state[self.motor_system._policy.agent_id][
            "motor_only_step"
        ] = True
        return self._observation
    else:
        # ...
```

Notably, this doesn't feel right as we use a _privileged_ (due to `"view_finder"` use) positioning procedure to keep a sensor in touch with the object. The future solution here will be to implement a similar touch object logic internally as part of the `SurfacePolicy`. This may be easier after refactoring the main Monty loop between the data loader, model, and the motor system to be more reinforcement-learning-like.

The algorithmic change is that all the positioning actions within the `TouchObject` positioning procedure are not passed to Monty for processing (except for the last observation when the procedure finds the object). This means Monty is not internally incrementing any steps while the `TouchObject` positioning procedure is running. The benchmarks below show that the surface benchmarks take longer (wall-clock time) to run, even though the average episode runtime is the same or shorter.

Many typing changes were made using `mypy` in strict mode. Mypy check for 121eb71150e97a5eee8397ff2d19e6078dfbce30 resulted in `Found 271 errors in 41 files (checked 159 source files)`. Mypy check for this pull request is `Found 267 errors in 40 files (checked 160 source files)`. The improvement is only slight because the `mypy` check does not run in strict mode yet, but a lot of typing was added.

I pre-emptively addressed some comments from the previous pull request inline.

## Benchmarks

Noticing NaNs for steps_to_individual_ts and individual_ts_rotation_error may indicate something (previous version on the left, current version on the right). The circles on the right chart are NaNs.
Screenshot 2025-06-21 at 07 54 58

Also, `randrot_noise_10simobj_surf_agent` benchmarks do not progress beyond step 65.

_comparison with 121eb71150e97a5eee8397ff2d19e6078dfbce30_

base_config_10distinctobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/ju0avl98/overview
base_config_10distinctobj_surf_agent,100.00,2.14,27,18.04,5,10 ❌ (worse) - https://wandb.ai/thousand-brains-project/Monty/runs/wmgbqut7/overview
```diff
-base_config_10distinctobj_surf_agent,100.00,0.00,28,14.01,2,11
+base_config_10distinctobj_surf_agent,100.00,2.14,27,18.04,5,10
```
randrot_noise_10distinctobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/skevdhmw/overview
randrot_noise_10distinctobj_dist_on_distm ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/4qoivhoc/overview
randrot_noise_10distinctobj_surf_agent,100.00,4.00,28,31.12,4,16 ❌ (worse) - https://wandb.ai/thousand-brains-project/Monty/runs/rcb1op26/overview
```diff
-randrot_noise_10distinctobj_surf_agent,100.00,1.00,29,20.42,3,21
+randrot_noise_10distinctobj_surf_agent,100.00,4.00,28,31.12,4,16
```
randrot_10distinctobj_surf_agent,100.00,2.00,27,19.11,4,10 ❌ (worse) - https://wandb.ai/thousand-brains-project/Monty/runs/nqugo4u3/overview
```diff
-randrot_10distinctobj_surf_agent,100.00,0.00,28,17.23,2,10
+randrot_10distinctobj_surf_agent,100.00,2.00,27,19.11,4,10
```
randrot_noise_10distinctobj_5lms_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/805xmwaz/overview
base_10simobj_surf_agent,95.00,10.71,55,15.48,11,21 😐 (different) - https://wandb.ai/thousand-brains-project/Monty/runs/p3cz9f2i/overview
```diff
-base_10simobj_surf_agent,93.57,8.57,70,13.98,6,27
+base_10simobj_surf_agent,95.00,10.71,55,15.48,11,21
```
randrot_noise_10simobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/ee3texb4/overview
randrot_noise_10simobj_surf_agent 💥 - does not proceed past step 65
  - ❓ (manually terminated) https://wandb.ai/thousand-brains-project/Monty/runs/hq85be39/overview
  - ❓ (manually terminated after 12 hours) https://wandb.ai/thousand-brains-project/Monty/runs/lfcv6lfm/overview

randomrot_rawnoise_10distinctobj_surf_agent,68.00,72.00,15,89.35,6,6 👍 (better) - https://wandb.ai/thousand-brains-project/Monty/runs/k63k9jwp/overview
```diff
-randomrot_rawnoise_10distinctobj_surf_agent,64.00,77.00,16,106.51,4,7
+randomrot_rawnoise_10distinctobj_surf_agent,68.00,72.00,15,89.35,6,6
```
base_10multi_distinctobj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/4qel8v62/overview

base_77obj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/z5tpd34y/overview
base_77obj_surf_agent,98.27,8.66,43,9.20,18,26 😐 (different) - https://wandb.ai/thousand-brains-project/Monty/runs/kg2hzeh9/overview
```diff
-base_77obj_surf_agent,98.70,6.49,56,12.94,12,34
+base_77obj_surf_agent,98.27,8.66,43,9.20,18,26
```
randrot_noise_77obj_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/vlga5xj9/overview; ❓ (manually terminated after 12 hours - https://wandb.ai/thousand-brains-project/Monty/runs/hr8cik2p/overview)

randrot_noise_77obj_surf_agent,93.51,28.57,107,39.36,38,86 ❌ (worse) - https://wandb.ai/thousand-brains-project/Monty/runs/jwgr6h12/overview
```diff
-randrot_noise_77obj_surf_agent,93.94,23.38,115,34.80,32,113
+randrot_noise_77obj_surf_agent,93.51,28.57,107,39.36,38,86
```
randrot_noise_77obj_5lms_dist_agent ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/3298dstn/overview


src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -496,11 +508,11 @@ def __next__(self):
                         "motor_only_step"
                     ] = True
 
-            self.motor_system._state = motor_system_state
+                self.motor_system._state = motor_system_state
 
-            self._counter += 1  # TODO clean up incrementing of counter
```
tristanls, line 503: **note**: I don't think this TODO is valid anymore. We usually increment the counter in the data loader `__next__` method. I don't think there's anything here to clean up.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -1113,175 +1246,39 @@ def pre_episode(self):
...
-        )
-
-    ###
-    # Methods that define behavior of __call__
-    ###
     def dynamic_call(
-        self, state: Optional[MotorSystemState] = None
-    ) -> OrientHorizontal | OrientVertical | MoveTangentially | MoveForward | None:
+        self, state: MotorSystemState | None = None
+    ) -> MoveTangentially | MoveForward | OrientHorizontal | OrientVertical:
```
tristanls, line 1255: **note**: Removed `None` as it can no longer be returned from this method since `self.get_next_action` no longer can return `None`. The order is alphabetical.

tests/unit/policy_test.py
```diff
@@ -901,116 +901,77 @@ def test_surface_policy_moves_back_to_object(self):
...
+                # the procedure completes, the DataLoader returns a motor-only step
+                # observation arrived at by the positioning procedure.
                 #  Step | Action           | Motor-only? | Obs processed? | Source
                 # ------|------------------|-------------|----------------|-------------
-                #  1    | MoveForward      | True        | False          | dynamic_call
+                #  0    | None             | True        | False          | first_step
+                #  1    | ObjectNotVisible | True        | False          | touch_object
                 #  2    | OrientHorizontal | True        | False          | dynamic_call
-                #  3    | OrientVertical   | False       | True           | dynamic_call
+                #  3    | OrientVertical   | KeyError    | True           | dynamic_call
```
tristanls, line 915: **note**: The below code will not set `"motor_only_step"` to `False` for `OrientVertical` action.

```python
            if (
                isinstance(self.motor_system._policy, SurfacePolicy)
                and self._action.name != OrientVertical.action_name()
            ):
                # Out of the four actions in the
                # MoveForward->OrientHorizontal->OrientVertical->MoveTangentially
                # "subroutine" defined in SurfacePolicy.dynamic_call, we only
                # want to send data to the learning module after taking the
                # OrientVertical action. The other three actions in the cycle
                # are motor-only to keep the surface agent on the object.
                motor_system_state[self.motor_system._policy.agent_id][
                    "motor_only_step"
                ] = True
```

When we look it up, we guard against it being a `KeyError`:
```python
    def is_motor_only_step(self, state: MotorSystemState):
        """Check if the current step is a motor-only step.

        TODO: This information is currently stored in motor system state, but
        should be stored in the policy state instead as it is tracking policy
        state, not motor system state. This will remove MotorSystemState param.

        Args:
            state (MotorSystemState): The current state of the motor system.

        Returns:
            bool: True if the current step is a motor-only step, False otherwise.
        """
        agent_state = self.get_agent_state(state)
        if "motor_only_step" in agent_state.keys() and agent_state["motor_only_step"]:
            return True
        else:
            return False
```

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -772,6 +785,157 @@ def sensor_rotation_relative_to_world(self, state: MotorSystemState) -> Any:
...
+        # The lateral movement will be down relative to the agent's facing position
+        # in the case of orient vertical, and left in the case of orient horizontal
+
+        # The below calculates the exsecant, which provides the necessary
+        # movement to bring the agent back to the same radius around the central
+        # point
+        move_forward_amount = distance_from_center * (
+            (1 - np.cos(np.radians(rotation_degrees)))
+            / np.cos(np.radians(rotation_degrees))
+        )
```
tristanls, line 919: @jeremyshoemaker's **question**: If I'm understanding the math right, we're moving a little bit along the tangent line at the point of intersection with the circle, and then moving inward to correct for the mismatch between the circle arc and the tangent line.

Is this more efficient than just rotating around the center of the circle? Seems like a complicated approach to do the same thing.

_I realize this code is just copied from below, but it's a good a time as any to ask these questions_

tests/unit/policy_test.py
```diff
@@ -901,116 +901,77 @@ def test_surface_policy_moves_back_to_object(self):
...
+                #  28   | ObjectNotVisible | True        | False          | touch_object
 
                 # Motor-only touch_object steps
-                if (
-                    13 <= loader_step <= 26
-                    or 31 <= loader_step <= 44
-                    or 49 <= loader_step <= 62
-                    or loader_step == 67
-                ):
+                if loader_step in [1, 13, 18, 23, 28]:
```
tristanls, line 947: **note**:  Yes, these numbers are very brittle. The step numbers are from the comment above.

To update the test, one would need to step through the code and observe what happens at each step and whether that is appropriate. Then, after documenting the trace in the comment above, update the assertions here with the corresponding numbers.

I haven't reviewed policy tests in depth, but at first glance, we want to test policy behavior and ensure that it executes deterministic behaviors given specific observations. An alternate approach could maybe be to initialize a policy, mock any side effects we care about, let policy execute for a specified count of steps, and then assert that sensor is at a specific endpoint and that side effects were as desired (i.e., the correct things got passed to the learning module).

src/tbp/monty/frameworks/environment_utils/transforms.py
```diff
@@ -314,7 +334,9 @@ def __init__(
         self.clip_value = clip_value
         self.depth_clip_sensors = depth_clip_sensors
 
-    def __call__(self, observations: dict, state: Optional[State] = None):
```
tristanls, line 317: **note**: `State` was the wrong type here.

src/tbp/monty/frameworks/models/motor_policies.py
nielsleadholm: Yes you're right. My guess is this approach was originally taken based on limited (perceived or actual) available actions in Monty, i.e. that the agent could only rotate about itself, rather than about an external point. I agree doing it via a single step would be cleaner.

In case some historical context is helpful, I just sent you a link on Slack to the original research meeting where this work was presented.

nielsleadholm:
Thanks for putting this together Tristan! Just a high level question/comment before I get a chance to review it in detail:

Do you have a sense for what is causing the i) ~doubling of inference time, ii) NaNs, and iii) non-terminating `randrot_noise_77obj_dist_agent`? A natural culprit would be that somehow the logic has changed so that `touch_object` is no longer working 100% of the time, and in particular it is sometimes failing to ever find the object.

What's strange is that from a quick look at the code, it seems like the core logic has just been copied over, so it's not clear to me why it would sometimes fail, and sometimes work. If I had to guess (again, without a thorough look through the code), it might be that some part of resolving touch-on-object, like the significant movement forward to be within `desired_object_distance` doesn't happen, or something like that. I can't imagine the random orienting around to find the object has changed, which makes me suspicious it is either the kick-off or the resolution of the touch-object process that isn't working as suspected. Anyways, just something to maybe think about.

tests/unit/policy_test.py
nielsleadholm: Yeah if I understand your suggestion, I agree it might be better for the tests to look for specific effects (e.g., can we get back on the object), rather than rigid requirements for each individual step.

tests/unit/policy_test.py
nielsleadholm: **question:** Sorry if I'm missing something, but is there a reason we don't update the logic such that we can return `motor_only_step=False`, instead of getting a `KeyError`? Beyond just making the code a bit easier to understand, the `motor_only_step`, and its logging, is useful for plotting how the agent's positions during an episode correspond to when information was passed to Monty.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -59,19 +59,22 @@
...
 
+    agent_id: str
+    """The agent ID corresponding to this motor policy."""
+
     def __init__(self) -> None:
         self.is_predefined = False
 
     @abc.abstractmethod
-    def dynamic_call(self, state: MotorSystemState | None = None) -> Action | None:
+    def dynamic_call(self, state: MotorSystemState | None = None) -> Action:
```
nielsleadholm, line 71: **question:** Is it necessary to make all the type changes in this PR? Is that generally how `mypy` will work going forward, i.e. running on any file (and in particular, the whole file) that is included in a PR?

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -772,6 +785,157 @@ def sensor_rotation_relative_to_world(self, state: MotorSystemState) -> Any:
         return agent_rotation * sensor_rotation
 
 
+class ObjectNotVisible(RuntimeError):
+    """Error raised when the object is not visible."""
+
+
+class TouchObject(PositioningProcedure):
+    """Positioning procedure to touch an object.
```
nielsleadholm, line 795: **suggestion (non-blocking):** Add a line along the lines of "Used by surface agents, i.e. agents with the `surface_agent` action space." I think this context is missing from the doc string now that it has been extracted.

tristanls:
@nielsleadholm, thanks for the suggestions. I wanted y'all to glance at it and see what you thought. I, too, am surprised by the scope of the benchmark differences. 

I am turning this into a Draft PR as I figure out what's happening.

nielsleadholm (commented):
Sounds good about investigating this further, just added a few smaller comments. For what it's worth I had a look through the logic of the new `class TouchObject` and unfortunately couldn't see any obvious culprit.

src/tbp/monty/frameworks/models/motor_policies.py
jeremyshoemaker: Looking at this again with some more context from the hackathon, a rotation followed by a lateral movement might be a better approach. I don't know if it results in less trig function calls though. I don't remember the derivation we were using for the drone to do orbits in this way. It might be simpler than this approach of moving along the tangent line and then moving again to correct for the displacement off the circle followed by a rotation to orient the camera back to the object. One way I could see the rotation followed by lateral movement approach could be problematic is if the intermediate camera state is being used, since it temporarily moves the camera off the target.

tristanls:
It turns out that running `randrot_noise_10simobj_surf_agent` benchmark using `run.py` instead of `run_parallel.py` [completes without issue (after 2h5m45s)](https://wandb.ai/thousand-brains-project/Monty/runs/kmmwnu1i/overview). 

randrot_noise_10simobj_surf_agent,93.00,40.00,146,26.39,23,14 with a diff of
```diff
-randrot_noise_10simobj_surf_agent,93.00,34.00,172,20.94,16,137
+randrot_noise_10simobj_surf_agent,93.00,40.00,146,26.39,23,14
```

This leads me to believe that benchmark getting stuck has something to do with parallelization, not algorithmic changes.

🤔, also, the stats reported 23 min overall run time, but WandB tracked 2h5m45s, so.. that's different.

src/tbp/monty/frameworks/models/motor_policies.py
tristanls: I'm not quite sure what you mean by `mypy` running.

We have thousands of type errors in the code, so it will be a while before `mypy` strict checks are turned on for PRs. In the meantime, before this PR, the `mypy` strict check was:
```
Found 3832 errors in 114 files (checked 161 source files)
```
This PR results in `mypy` strict check of:
```
Found 3765 errors in 113 files (checked 162 source files)
```

Regarding whether type additions and/or changes should be included in a PR like this, I use these types to help me write the code. They help me reason whether the code is correct, so these type updates are me debugging the code to get the code as correct as possible. I try to limit the scope as much as possible and not touch types that are too far from the active changes.

tristanls: Regarding the specific type change from `-> Action | None` to `-> Action`, I removed the only code that returned `None`. So, `dynamic_call` can no longer return `None`, hence the update here with some cascading effects throughout.

vkakerbeck:
I am not sure it would mean that. Running with run.py instead of run_parallel.py will have different random seed behavior, the agent will take different actions, code is executed in different orders, etc. so it may just randomly not get into the stuck state. I would assume it is similar to using run_parallel.py with a different random seed. If run_parallel worked fine before the changes here (which I assume it did since this is what we use for reporting the benchmark experiments) I wouldn't think it is caused by parallelization issues. I would guess more along the lines of what Niels suggested. That we get into an infinite loop because the agent doesn't find the object, and we don't have a proper termination check invoked. I will try to have a closer look at the code later.

tristanls:
@vkakerbeck, thank you, good points. I'll run the "stuck" benchmarks with a different seed to see what happens.

tristanls:
`randrot_noise_10simobj_surf_agent` with seed=1337 gets stuck on step 43: https://wandb.ai/thousand-brains-project/Monty/runs/8lw7d3mq/overview . I'll try a different seed.

tristanls:
`randrot_noise_10simobj_surf_agent` with seed=714157817 finished without issue: https://wandb.ai/thousand-brains-project/Monty/runs/4bbdno82/overview

vkakerbeck:
okay thanks for looking into this more. That sounds like in some edge cases the policy behavior has changed and it now get's into an infinite loop/stuck in those cases.

nielsleadholm:
Yeah it sounds like something has changed with the policy. I don't recall us having experienced this, even though it's fairly common for us to cause subtle changes to the seed with the different experiments we run and past PRs.

For debugging this kind of issue, it can be helpful to re-run a specific episode (using `make_detailed_follow_up_configs.py`, assuming it still works) with the view-finder visualization active. You might then directly see that the view-finder e.g. isn't moving at all or is only moving in a specific direction. I remember in the distant past (since fixed), there was an issue where the agent could spin into the distance, resulting in the object slowly fading into the ether, even as the agent spins round hoping to find it.

tristanls:
Identified a correlation for the NaNs, which is terminating due to maximum episode steps being reached:
```
INFO:tbp.monty.frameworks.experiments.object_recognition_experiments:run_episode_steps:98:Terminated due to maximum episode steps : 6000
```

tristanls:
So, here is the infinite loop problem, an example of what happens when a step gets "stuck". 

The video below shows a few invocation of TouchObject positioning procedure. When TouchObject succeeds, the title changes to `Object touched!`. As far as I can tell, the problem is using the `view_finder`, which sees something different from the sensor itself. The first few times are expected `touch_object` successes, but soon you'll see a loop where `touch_object` claims success based on `view_finder`, but we can see the sensor on the right has no object pixels in the center.

However, switching positioning procedure to use `patch` (the sensor) also doesn't work as it still gets stuck in ObjectNotVisible loop.

https://github.com/user-attachments/assets/77ce655a-a64a-49a8-a1d7-bde83ec10850

The view from the SurfacePolicy (much faster video, a touch object loop happens between each step):

https://github.com/user-attachments/assets/9cc39636-64e3-4b80-bedf-8916c1844793

@scottcanoe, you've dealt with something similar before, yes?

tristanls:
As before, I will break up this pull request into smaller pull requests to isolate the benchmark-altering change. #347 is one such pull request that should have no benchmark impact.

nielsleadholm:
Interesting, this feels like one of those "how did this ever work" instances. The issue that the viewfinder can find the object, but fail to provide the object for the SM, is indeed what @scottcanoe fixed a while back for `get_good_view`. I'm trying to understand why this *hasn't* been an issue for `touch_object`, given we also rely on the view-finder. From a quick look at the code, I wonder if the removal of the following lines (in particular `if self._action is None`), could be the culprit? I think this might previously have been ensuring that the search loop continues (or at least restarts) until the _sensor-module_ has visualized the object (view-finder is not sufficient), but I could be wrong.

```python
            # If entirely off object, use vision (i.e. view-finder)
            # TODO refactor so that this check is done in the motor-policy, and we
            # update the constraint separately/appropriately; i.e. the below
            # code should be as general as possible
            if (
                isinstance(self.motor_system._policy, SurfacePolicy)
                and self._action is None
            ):
```

tristanls:
@nielsleadholm, check for `self._action is None` was replaced by `ObjectNotVisible`. There is no change in logic due to that. It is also demonstrated in this draft PR https://github.com/thousandbrainsproject/tbp.monty/pull/349 (easier to just look at this commit: https://github.com/thousandbrainsproject/tbp.monty/pull/349/commits/b09599566476803e63c0cab3569386fcd182cc7f).

vkakerbeck:
Interesting find. Maybe worth pairing with @scottcanoe on this one?
If it helps, here are the two PRs that fixed a similar problem when setting up the experiment:
https://github.com/thousandbrainsproject/tbp.monty/pull/114
https://github.com/thousandbrainsproject/tbp.monty/pull/146

tristanls:
This pull request is superseded by #356.

PR #340, opened 2025-06-21
feat! : evidence slope tracker integration for resampling

ramyamounir:
This PR adds support for tracking evidence scores within the `ResamplingHypothesesUpdater`. This allows the updater to use evidence slope information to determine which existing hypotheses can be removed in favor of new informed hypotheses. There are 3 major changes in this PR:
- The `EvidenceSlopeTracker` (and corresponding `EvidenceSlopeTrackerTest`) class keeps track of the evidence scores over a window and returns which hypotheses can be removed, if any. See [here](https://github.com/thousandbrainsproject/tbp.monty/blob/21b8cc0a559cc45c457c6daa6753ccffabe4eb3a/src/tbp/monty/frameworks/utils/evidence_matching.py#L254) and [here](https://github.com/thousandbrainsproject/tbp.monty/blob/21b8cc0a559cc45c457c6daa6753ccffabe4eb3a/tests/unit/frameworks/utils/evidence_matching_test.py#L181).
- Modifications of `ResamplingHypothesesUpdater` to integrate the slope trackers. This happens in a few places within the `update_hypotheses` function:
    - A tracker is initialized for every `graph_id`. This tracker stores information for each channel separately. See [here](https://github.com/thousandbrainsproject/tbp.monty/blob/21b8cc0a559cc45c457c6daa6753ccffabe4eb3a/src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py#L208)
    - When calculating `_sample_count`, we use the tracker to get information about how many hypotheses we must keep due to their ages (hypotheses must be above a certain age before they can be removed from the hyp space). See [here](https://github.com/thousandbrainsproject/tbp.monty/blob/21b8cc0a559cc45c457c6daa6753ccffabe4eb3a/src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py#L362)
    - In `_sample_existing`, we use the tracker to retrieve the ids of the hypotheses to keep and the ones to remove. We also update the tracker data by removing the hypotheses. See [here](https://github.com/thousandbrainsproject/tbp.monty/blob/21b8cc0a559cc45c457c6daa6753ccffabe4eb3a/src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py#L404)
    - In `_sample_informed`, we add new hypotheses matching the number of sampled hypotheses. See [here](https://github.com/thousandbrainsproject/tbp.monty/blob/21b8cc0a559cc45c457c6daa6753ccffabe4eb3a/src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py#L546)
    - Finally, we update the tracker evidence scores. See [here](https://github.com/thousandbrainsproject/tbp.monty/blob/21b8cc0a559cc45c457c6daa6753ccffabe4eb3a/src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py#L272)
- Bypass of evidence_update_threshold. In [PR#325](https://github.com/thousandbrainsproject/tbp.monty/pull/325), we refactored the evidence update threshold logic into a bare function. In this PR, we are adding a bypass option when `None` is passed as the `evidence_threshold_config`. I also did a minor modification to the displacer logic to bypass some possibly inefficient operations (e.g., `np.where`). This can be simplified if we remove the update threshold functionality in the future. See [here](https://github.com/thousandbrainsproject/tbp.monty/blob/21b8cc0a559cc45c457c6daa6753ccffabe4eb3a/src/tbp/monty/frameworks/utils/evidence_matching.py#L448) and [here](https://github.com/thousandbrainsproject/tbp.monty/blob/21b8cc0a559cc45c457c6daa6753ccffabe4eb3a/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py#L145).

### Preliminary Results
I ran some quick rapid configs experiments with the following parameters:
- OBJECTS_LIST = ["strawberry", "banana", "mug", "potted_meat_can", "mustard_bottle"]
- EVAL_STEPS = 50
- evidence_threshold_config = None
- hypotheses_count_multiplier = 1.0
- hypotheses_existing_to_new_ratio = 0.1
- window_size = 3
- min_age = 5

The results are shown below and they seem okay for preliminary (non-benchmark) results. I noticed the evidence scores for "potted_meat_can" took more time to decay when the object changed to a "mustard_bottle", but it eventually decayed.

5_objects
5_objects_pose

### Benchmark Results
I ran the unsupervised inference benchmark to assess the performance of resampling. I noticed a significant improvement in performance on the "No Reset" experiments, but there were also increase in time which is expected at this stage (discussed in the next section)

First, I ran the main branch as a baseline, this experiment was set up with effectively no resampling (`hypotheses_existing_to_new_ratio=0.0`). It also had `evidence_threshold_config` set to 80%. But the EVAL_STEPS were set to 100, which is why it took more time than regular benchmark experiments. It is still a No Reset experiment.
| Experiment | Correct (%) | Avg Matching Steps | Run Time (mins) | Avg Episode Time (s) |
|---|---|---|---|---|
| unsupervised_inference_distinctobj_dist_agent | 10.0 | 100.0 | 7.08 | 4.25 |
| unsupervised_inference_distinctobj_surf_agent | 5.0 | 97.31 | 12.97 | 7.78 |

Then I ran the code from this PR. Here we update all hypotheses, and resample 10% new informed hypotheses every step (`hypotheses_existing_to_new_ratio=0.1`).
| Experiment | Correct (%) | Avg Matching Steps | Run Time (mins) | Avg Episode Time (s) |
|---|---|---|---|---|
| unsupervised_inference_distinctobj_dist_agent | 96.0 | 100.0 | 25.94 | 15.56 |
| unsupervised_inference_distinctobj_surf_agent | 94.0 | 98.82 | 19.0 | 11.4 |

Finally, I wanted to test what would happen if we use a smaller hypotheses space for all objects. I used one third of the hypotheses space size and I saw a speed up in the experiments without affecting performance.
| Experiment | Correct (%) | Avg Matching Steps | Run Time (mins) | Avg Episode Time (s) |
|---|---|---|---|---|
| unsupervised_inference_distinctobj_dist_agent | 97.0 | 100.0 | 14.27 | 8.56 |
| unsupervised_inference_distinctobj_surf_agent | 94.0 | 99.85 | 15.75 | 9.45 |

The reasons for resampling experiments taking more time are summarized below:
- The resampling logic itself, where we sample informed hypotheses at every step, takes time
- The No Reset benchmark experiments run for exactly 100 steps, so we cannot compare their time to regular benchmark experiments
- We have implemented resampling, but we still use the full hypotheses space size, we can now afford to operate on a smaller number of hypotheses. These hypotheses are always getting refined by continuous resampling and removal of the unlikely ones.
- In the resampling experiments we update "all" hypotheses. So experiments are expected to take longer if we don't optimize the number of hypotheses and allow for triggering terminal conditions.
- The No Reset experiments have to be run back-to-back and cannot be run in parallel with `run_parallel.py` like the default benchmarks.  

### Next Steps
In a follow-up PR, I want to run resampling to the regular benchmarks and perform a hyperparameter sweep. We should also consider adding a config parameter that defines the hyp space size as a factor of the number of object graph nodes.

### DIFF
--- a/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
@@ -39,7 +39,7 @@ class HypothesesDisplacer(Protocol):
         self,
         channel_displacement: np.ndarray,
         channel_features: dict,
-        evidence_update_threshold: float | None,
+        evidence_update_threshold: float,
         graph_id: str,
         possible_hypotheses: ChannelHypotheses,
         total_hypotheses_count: int,
@@ -132,7 +132,7 @@ class DefaultHypothesesDisplacer:
         self,
         channel_displacement: np.ndarray,
         channel_features: dict,
-        evidence_update_threshold: float | None,
+        evidence_update_threshold: float,
         graph_id: str,
         possible_hypotheses: ChannelHypotheses,
         total_hypotheses_count: int,
@@ -142,13 +142,9 @@ class DefaultHypothesesDisplacer:
         search_locations = possible_hypotheses.locations + rotated_displacements
 
         # Get indices of hypotheses with evidence > threshold
-        # TODO: Should simplify after the update threshold concept is removed
-        if evidence_update_threshold is None:
-            hyp_ids_to_test = np.arange(possible_hypotheses.evidence.shape[0])
-        else:
-            hyp_ids_to_test = np.where(
-                possible_hypotheses.evidence >= evidence_update_threshold
-            )[0]
+        hyp_ids_to_test = np.where(
+            possible_hypotheses.evidence >= evidence_update_threshold
+        )[0]
         num_hypotheses_to_test = hyp_ids_to_test.shape[0]
         if num_hypotheses_to_test > 0:
             logger.info(
--- a/src/tbp/monty/frameworks/utils/evidence_matching.py
+++ b/src/tbp/monty/frameworks/utils/evidence_matching.py
@@ -456,7 +456,7 @@ class EvidenceSlopeTracker:
 
 
 def evidence_update_threshold(
-    evidence_threshold_config: float | str | None,
+    evidence_threshold_config: float | str,
     x_percent_threshold: float | str,
     max_global_evidence: float,
     evidence_all_channels: np.ndarray,
@@ -475,6 +475,13 @@ def evidence_update_threshold(
 
     Returns:
         The evidence update threshold.
+
+    Note:
+        The logic of `evidence_threshold_config="all"` can be optimized by
+        bypassing the `np.min` function here and bypassing the indexing of
+        `np.where` function in the displacer. We want to update all the existing
+        hypotheses, therefore there is no need to find the specific indices for
+        them in the hypotheses space.
 
     Raises:
         InvalidEvidenceThresholdConfig: If `evidence_threshold_config` is
@@ -504,12 +511,12 @@ def evidence_update_threshold(
         x_percent_of_max = max_global_evidence / 100 * x_percent_threshold
         return max_global_evidence - x_percent_of_max
     elif evidence_threshold_config == "all":
-        return None
+        return np.min(evidence_all_channels)
     else:
         raise InvalidEvidenceThresholdConfig(
             "evidence_threshold_config not in "
             "[int, float, '[int]%', 'mean', "
-            "'median', 'all', 'x_percent_threshold', 'None']"
+            "'median', 'all', 'x_percent_threshold']"
         )
 
 
--- a/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
@@ -59,15 +59,14 @@ class EvidenceGraphLM(GraphLM):
             by this value before being added to the overall evidence of a hypothesis.
             This factor is only multiplied with the feature evidence (not the pose
             evidence as opposed to the present_weight).
-        evidence_threshold_config (float | str | None): How to decide which hypotheses
+        evidence_threshold_config: How to decide which hypotheses
             should be updated. When this parameter is either '[int]%' or
             'x_percent_threshold', then this parameter is applied to the evidence
             for the Most Likely Hypothesis (MLH) to determine a minimum evidence
             threshold in order for other hypotheses to be updated. Any hypotheses
             falling below the resulting evidence threshold do not get updated. The
             other options set a fixed threshold that does not take MLH evidence into
-            account. Using None will bypass threshold calculation and return None.
-            In [int, float, None, '[int]%', 'mean', 'median', 'all',
+            account. In [int, float, '[int]%', 'mean', 'median', 'all',
             'x_percent_threshold']. Defaults to 'all'.
         vote_evidence_threshold: Only send votes that have a scaled evidence above
             this threshold. Vote evidences are in the range of [-1, 1] so the threshold


src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
```diff
@@ -141,9 +141,13 @@ def displace_hypotheses_and_compute_evidence(
         search_locations = possible_hypotheses.locations + rotated_displacements
 
         # Get indices of hypotheses with evidence > threshold
-        hyp_ids_to_test = np.where(
-            possible_hypotheses.evidence >= evidence_update_threshold
-        )[0]
+        # TODO: Should simplify after the update threshold concept is removed
+        if evidence_update_threshold is None:
```
vkakerbeck: Why don't you just use the 'all' parameter? It seems a bit more intuitive to me (saying "I want to update all hypotheses") and it is a functionality we have already implemented. If you want to avoid the np.where lookup you could just change this if statement to `if evidence_update_threshold == "all"`. This would help make any other experiments that update all hypotheses and still use older configs more efficient too.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -165,6 +160,19 @@ def __init__(
             use_features_for_matching=self.use_features_for_matching,
         )
 
+        # Controls the shrinking or growth of hypothesis space size
+        # Cannot be less than 0
+        self.hypotheses_count_multiplier = max(0, hypotheses_count_multiplier)
```
vkakerbeck, line 173: Does this one mean if it is between 0 and 1 the hypothesis space will shrink and if it is >1 it will grow? What is a scenario in which we would want to grow the hypothesis space over time?

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -339,6 +359,12 @@ def _sample_count(
         if new_informed > full_informed_count:
             new_informed = full_informed_count
 
+        # Additional adjustment based on valid mask
+        must_keep = np.sum(~tracker.valid_indices_mask(input_channel))
```
vkakerbeck: I'm not sure what happens here based on the comment and function names. Maybe you could elaborate a bit? It looks like we set must_keep to the number of indices that are not valid? This seems counterintuitive. Why would we want to keep the invalid indices?

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+
+    Each input channel maintains its own hypotheses with independent evidence histories.
+    This tracker supports adding, updating, pruning, and analyzing hypotheses per
+    channel.
+
+    Attributes:
+        window_size (int): Number of past values to consider for slope calculation.
+        min_age (int): Minimum number of updates before a hypothesis can be considered
+            for removal.
+        data (Dict[str, np.ndarray]): Maps channel names to their hypothesis evidence
```
vkakerbeck: Why not give this a more describtive name like `evidence_buffer` or `evidence_history`?

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+        Args:
+            channel (str): Name of the input channel.
+
+        Returns:
+            int: Number of hypotheses currently tracked in the channel.
+        """
+        return self.data.get(channel, np.empty((0, self.window_size))).shape[0]
+
+    def valid_indices_mask(self, channel: str) -> np.ndarray:
+        """Returns a boolean mask for hypotheses valid for removal in a channel.
```
vkakerbeck: Ok this explains my earlier question. Maybe it would help to call this `removable_hypotheses` or to be more specific, `old_enough_hyptheses`?

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+
+        Args:
+            values (List[float] | np.ndarray): List or array of new evidence values.
+            channel (str): Name of the input channel.
+
+        Raises:
+            ValueError: If the channel doesn't exist or the number of values is
+                incorrect.
+        """
+        values = np.array(values, dtype=float)
```
vkakerbeck: Why are the evidence values not an np array already? I thought we are doing all the evidence update operations within numpy

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+        Args:
+            channel (str): Name of the input channel.
+
+        Returns:
+            np.ndarray: Array of average slopes, one per hypothesis.
+        """
+        diffs = np.diff(self.data[channel], axis=1)
+        valid_steps = np.count_nonzero(~np.isnan(diffs), axis=1)
+        valid_steps = np.where(valid_steps == 0, np.nan, valid_steps)
+        return np.nansum(diffs, axis=1) / valid_steps
```
vkakerbeck: I'm not sure I understand what is happening here. Ir at least it seems like it could be done more efficiently if I am not missing something. If I understand it right you
- Calculate the diffs along axis 1 of the evidence array. Some of those diffs might be nan because some values in the evidence array may be nan.
- You then count the number of elements in each row that is not nan.
- If all values in the row are nan you set that rows step count to nan, otherwise to the stepcount calculated in the previous row.
- you then get the sum of diffs (treating nans as zeros) and divide it by the number of elements in diffs that is not nan.

If this is what is happening I guess I have 2 questions:
1. Why do you need to use np.count_nonzero? Couldn't you just do `valid_steps = np.sum(~np.isnan(diffs), axis=1)`?
2. Why do you need to use np.where? Couldn't you just do `valid_steps[valid_steps == 0] = np.nan`? (you may have to cast to float first to be able to assign a nan value.

I am not an expert on this but it seems like those would be more efficient or at least more readable. It could also be worth documenting a bit more. Like that the returned values will be nan whenever valid_steps is nan.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+            )
+        total_ids = np.arange(total_size)
+        num_remove = total_size - num_keep
+
+        # Retrieve valid slopes and sort them
+        valid_mask = self.valid_indices_mask(channel)
+        slopes = self._calculate_slopes(channel)
+        valid_slopes = slopes[valid_mask]
+        valid_ids = total_ids[valid_mask]
+        sorted_indices = np.argsort(valid_slopes)
```
vkakerbeck: If we look more into optimizing this implementation, this one may be one to keep an eye on. It could be totally sufficient to just randomly subsample num_remove induces that are valid (adding an additional check for being valid requiring the slope to be <mean(slopes)) instead of having to sort the whole array.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+
+        # Calculate which ids to keep and which to remove
+        to_remove = valid_ids[sorted_indices[:num_remove]]
+        to_keep = np.setdiff1d(total_ids, to_remove)
```
vkakerbeck: This seems like it could be quite inefficient. Have you tried using a mask instead of figuring out the set diff? something like:
```
remove_mask = np.zeros(total_size, dtype=bool)
to_remove = valid_ids[sorted_indices[:num_remove]]
remove_mask[to_remove] = True

to_keep = total_ids[~remove_mask]
```

nielsleadholm:
Thanks for opening this Ramy, very exciting and looking forward to reviewing it in detail. Just thought it would be helpful if I added some context to your questions @vkakerbeck , obviously feel free to follow up with more detail though Ramy.

**Re. speed / computational efficiency.**
- Ramy has been exploring methods to actually reduce the number of hypotheses sampled, but I had suggested that in this particular PR, we not spend too much time addressing the runtime question, since that is not directly related to solving the issue of high accuracy on the unsupervised inference benchmark. I agree it's an important thing to resolve, but I don't feel the times here are awful (e.g. hours) - they can very much be run, and importantly this offers a first working solution to the unsupervised inference benchark. I would therefore suggest we resolve that in a follow-up PR, which should be very much possible with various approaches such as sampling fewer hypotheses, and the points you discuss. What do you think?
- When addressing this, I agree the recreation of Numpy arrays is likely an issue. As long as we have a maximum number of potential hypotheses, it should be possible to pre-initialize this and keep it fixed, which might resolve some of these problems. This would avoid memory allocation overhead, while still leveraging vectorized acceleration in Numpy.

**Re. re-sampling and slope-tracking vs. standard decay**
- Maybe I am missing something, but I think the difference very much emerges because we are dealing with the unsupervised inference benchmark, where objects can change, and yet in the original implementations, our hypotheses are rigidly fixed. This fixing of the initial hypothesis space is there whether we allow evidence values to decay below a sub-sampling threshold or not, and this is captured nicely by Ramy's theoretical limit metric. The importance of Ramy's implementation is therefore that we can identify hypotheses that can be eliminated, so that we can sample entirely new ones that are informed by current observations (including observations on new objects when these change).
- The slope-tracker is a good heuristic for identifying neurons for elimination. Biologically, I think you can still think of it as a threshold - if a neuron isn't receiving sufficient input within a short time horizon, then it might fall below a spiking threshold, and as a result become further suppressed by competitive inhibition - i.e. eliminated. Lateral inhibition and sparsity could then mean that another neuron (hypothesis) is able to become active since the first neuron has become suppressed. The key difference is that code-wise, this enables us to sample new hypotheses that were not relevant when seeing object $i$, but might be relevant now on object $i+1$ etc.

vkakerbeck:
Oh sorry I wasn't very clear with that. I didn't mean that all those ideas around speed improvements need to be addressed in this PR. I left a few inline comments which can be relatively quick to try but everything in the high level comment was meant as ideas to look into later on.

Regarding the resampling, I agree that adding of new hypotheses is definitely required. However, I was looking at the mechanism for identifying and deleting hypotheses as something different. Just off the top of my head, using the old mechanism for that I could imagine something like: We use the evidence threshold to mask which hypotheses are being updated (as we currently do). If hypotheses don't get evidence for a few steps, their evidence decays below the threshold (or with the current default settings, the threshold goes up, and hypotheses only decay if they receive negative evidence, which is similar). When we add new hypotheses, they can just replace any of the currently masked hypotheses. The new hypotheses are initialized above the threshold so that they can receive updates and have a chance to establish themselves or quickly be proven wrong. I am not sure if this way is better or worse, it just seems very similar in the effect it would have, and computationally easier. It wouldn't involve any separate slope tracking and would allow for a dynamic number of hypotheses tested at any step. It's nothing that needs to be addressed in this PR. I just want to think through this a bit more before we make it the default option and remove the simple thresholding.

nielsleadholm:
Ah ok nice, makes sense.

And yeah thanks for the clarification, in terms of how we eliminate hypotheses, I agree there are probably a few different ways to do it. On the surface, I think what you describe is pretty similar to the concept of the slope tracker, so yeah maybe something to consider as a simpler implementation. Just one other comment is that a general issue we ran into is that trying to initialize hypotheses with a high-evidence value, for example within x-percent of the top hypothesis, results in some unintuitive effects. For example, it seems that within a local neighborhood, it is often the case that at least one rotation hypothesis that is initialized will be consistent with the next step - it is only after integrating over many steps that we find evidence against a new hypothesis. This means that if we keep initializing new hypotheses with medium-to-high values, they can continuously "creep up" alongside the top hypothesis, even though they aren't actually good hypotheses. All they are is sufficiently good hypotheses for the _next_ sensation. In the follow-up step, another "sufficiently good" hypothesis is likely to be initialized, making it challenging to eliminate these. Maybe something we can chat about more next Monday and how we can work around this.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
tristanls: @vkakerbeck `evidence_update_threshold` is a pre-calculated `float`. You're thinking of what's now called the `evidence_threshold_config`. The `evidence_update_threshold` passed into this function is [calculated based on the `evidence_threshold_config` parameter (the one with `"all"` as an option)](https://github.com/thousandbrainsproject/tbp.monty/blob/64ba534e4c2dba306de2b6b81c18a1027ac48fb9/src/tbp/monty/frameworks/utils/evidence_matching.py#L254), but that config doesn't get passed into here. 

If you recall, you asked me to move the calculation of `evidence_update_threshold` out of the hypotheses displacer: https://github.com/thousandbrainsproject/tbp.monty/pull/317#discussion_r2131960365

benchmarks/configs/unsupervised_inference_experiments.py
```diff
@@ -85,6 +85,9 @@
 ] = False
 surf_monty_config.monty_class = MONTY_CLASS
 surf_monty_config.monty_args.min_eval_steps = EVAL_STEPS
+surf_monty_config.learning_module_configs["learning_module_0"]["learning_module_args"][
```
tristanls: **nitpick**: Seems like this would go better next to the other `surf_monty_config.learning_module_configs["learning_module_0"]` configuration above.

benchmarks/configs/unsupervised_inference_experiments.py
```diff
@@ -105,6 +108,9 @@
 ] = False
 dist_monty_config.monty_class = MONTY_CLASS
 dist_monty_config.monty_args.min_eval_steps = EVAL_STEPS
+dist_monty_config.learning_module_configs["learning_module_0"]["learning_module_args"][
```
tristanls: **nitpick**: Seems like this would go better next to the other `dist_monty_config.learning_module_configs["learning_module_0"]` configuration above.

benchmarks/results/ycb_unsupervised_inference.csv
```diff
@@ -1,3 +1,3 @@
 Experiment,Correct (%)|align right,Num Match Steps|align right,Run Time (mins)|align right,Episode Run Time (s)|align right
-unsupervised_inference_distinctobj_dist_agent,12.00,100,5,3.0
-unsupervised_inference_distinctobj_surf_agent,6.00,98,15,9
+unsupervised_inference_distinctobj_dist_agent,96.00,100,25.94,15.56
```
tristanls: **note**: I think our benchmark results convention has been to round Run Time (mins) to full minutes and Episode Run Time (s) to full seconds, as in: 26,16

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
```diff
@@ -131,7 +131,7 @@ def displace_hypotheses_and_compute_evidence(
         self,
         channel_displacement: np.ndarray,
         channel_features: dict,
-        evidence_update_threshold: float,
+        evidence_update_threshold: float | None,
```
tristanls: **issue**: This violates the `HypothesesDisplacer` protocol defined above.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
```diff
@@ -141,9 +141,13 @@ def displace_hypotheses_and_compute_evidence(
...
-        hyp_ids_to_test = np.where(
-            possible_hypotheses.evidence >= evidence_update_threshold
-        )[0]
+        # TODO: Should simplify after the update threshold concept is removed
+        if evidence_update_threshold is None:
+            hyp_ids_to_test = np.arange(possible_hypotheses.evidence.shape[0])
+        else:
+            hyp_ids_to_test = np.where(
+                possible_hypotheses.evidence >= evidence_update_threshold
+            )[0]
```
tristanls: **issue**: You are changing the default hypotheses displacer implementation. Why aren't you creating a new displacer instead?

From https://github.com/thousandbrainsproject/tbp.monty/pull/325#issuecomment-2970911032:
> Then in a follow-up PR I would define a `ResamplingHypothesesDisplacer` that ignores the `evidence_update_threshold`.

I expected a `ResamplingHypothesesDisplacer` implementation in this PR.

Checking `if evidence_update_threshold is None` does not ignore the parameter but uses `None` as a sentinel value to change the algorithm's behavior.

**suggestion**: I originally suggested returning `None` from `evidence_update_threshold(...)` so that it _wouldn't be used_. Now, since you intend to use it, I think returning `0` would be a better choice, as it would require no changes to the default hypotheses displacer. Can `possible_hypotheses.evidence` ever be negative?

src/tbp/monty/frameworks/utils/evidence_matching.py
tristanls: I echo @vkakerbeck's sentiment here, "valid for removal" is a confusing framing. This method and docstrings would be clearer without the word "valid" present. Also, the difference between the first line and the Returns docstring further confuses what "valid" means:
```
Returns a boolean mask for hypotheses valid for removal in a channel.
Returns:
    np.ndarray: Boolean array indicating valid hypotheses (age >= min_age).
```

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -165,6 +160,19 @@ def __init__(
...
+        self.hypotheses_count_multiplier = max(0, hypotheses_count_multiplier)
+
+        # Controls the ratio of existing to newly sampled hypotheses
+        # Bounded between 0 and 1
+        self.hypotheses_existing_to_new_ratio = max(
+            0, min(hypotheses_existing_to_new_ratio, 1)
+        )
+
+        # Dictionary of slope trackers, one for each graph_id
+        self.evidence_slope_trackers = {}
```
tristanls: **issue**: Missing type annotation for `self.evidence_slope_trackers`. Based on usage, I think this should be `dict[str, EvidenceSlopeTracker]`.

```suggestion
        self.evidence_slope_trackers: dict[str, EvidenceSlopeTracker] = {}
```

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -339,6 +359,12 @@ def _sample_count(
         if new_informed > full_informed_count:
             new_informed = full_informed_count
 
+        # Additional adjustment based on valid mask
+        must_keep = np.sum(~tracker.valid_indices_mask(input_channel))
+        if must_keep > existing_maintained:
+            existing_maintained = must_keep
+            new_informed = needed - existing_maintained
```
tristanls: **issue**: Return type from `np.sum()` is a `np.bool_`. The calculations are fine, but the type system is unhappy with all the coercion, as it makes the assignment `existing_maintained = must_keep` invalid. Coercion to `int` makes the type system happy and the intent of summing booleans explicit.
```suggestion
        must_keep = int(np.sum(~tracker.valid_indices_mask(input_channel)))
        if must_keep > existing_maintained:
            existing_maintained = must_keep
            new_informed = needed - existing_maintained
```

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -372,13 +401,20 @@ def _sample_existing(
                 evidence=np.zeros(0),
             )
 
+        keep_ids, remove_ids = tracker.calculate_keep_and_remove_ids(
+            num_keep=existing_count,
+            channel=input_channel,
+        )
+
+        # Update tracker by removing the remove_ids
+        tracker.remove_hyp(remove_ids, input_channel)
```
tristanls, line 419: **issue**: `tracker.remove_hyp(self, hyp_ids: List[int], channel: str) -> None: ...` expects a list of ids and not an ndarray.
```suggestion
        tracker.remove_hyp(remove_ids.tolist(), input_channel)
```

Alternatively, change `tracker.remove_hyp` signature to:
```python
def remove_hyp(self, hyp_ids: npt.NDArray[np.int_], channel: str) -> None: ...
```

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+
+        Args:
+            window_size (int, optional): Number of evidence points per hypothesis.
+            min_age (int, optional): Minimum number of updates before removal is
+                allowed.
+        """
+        self.window_size = window_size
+        self.min_age = min_age
+        self.data: Dict[str, np.ndarray] = {}
+        self.age: Dict[str, np.ndarray] = {}
```
tristanls: **suggestion**: Numpy has `numpy.typing` typically used as `import numpy.typing as npt` and that allows for more precision in specifying the data types in our arrays.

**suggestion**: The use of `typing.Dict` here is extraneous. With the `from __future__ import annotations` you can use `dict`.

```suggestion
        self.data: dict[str, npt.NDArray[np.float64] = {}
        self.age: dict[str, npt.NDArray[np.int_] = {}
```

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+
+        Args:
+            channel (str): Name of the input channel.
+
+        Returns:
+            int: Number of hypotheses currently tracked in the channel.
+        """
+        return self.data.get(channel, np.empty((0, self.window_size))).shape[0]
+
+    def valid_indices_mask(self, channel: str) -> np.ndarray:
```
tristanls: **suggestion**
```suggestion
    def valid_indices_mask(self, channel: str) -> npt.NDArray[np.bool_]:
```

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+        new_age = np.zeros(num_new_hyp, dtype=int)
+
+        if channel not in self.data:
+            self.data[channel] = new_data
+            self.age[channel] = new_age
+        else:
+            self.data[channel] = np.vstack((self.data[channel], new_data))
+            self.age[channel] = np.concatenate((self.age[channel], new_age))
+
+    def update(self, values: List[float] | np.ndarray, channel: str) -> None:
```
tristanls: **issue**: It looks like the only reason to accept `List[float]` is for testing.

**suggestion**: Use `np.array([1.0, 2.0])` in tests and tighten this type up to:
```suggestion
    def update(self, values: npt.NDArray[np.float64], channel: str) -> None:
```

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+        # Shift evidence data by one step
+        self.data[channel][:, :-1] = self.data[channel][:, 1:]
+
+        # Add new evidence data
+        self.data[channel][:, -1] = values
+
+        # Increment age
+        self.age[channel] += 1
+
+    def _calculate_slopes(self, channel: str) -> np.ndarray:
```
tristanls: **suggestion**:
```suggestion
    def _calculate_slopes(self, channel: str) -> npt.NDArray[np.float64]:
```

src/tbp/monty/frameworks/utils/evidence_matching.py
tristanls: **question**: Setting return type to `-> npt.NDArray[np.float64]:` resulted in a type error. I attempted to assign types, and this is what I came up with. Is this what is happening here? I'm unsure if `np.nansum` outcome is a single value or an array.

```python
        diffs = np.diff(self.data[channel], axis=1).astype(np.float64, copy=False)
        valid_steps = cast(
            npt.NDArray[np.int_], np.count_nonzero(~np.isnan(diffs), axis=1)
        )
        valid_steps = np.where(valid_steps == 0, np.nan, valid_steps).astype(
            np.float64, copy=False
        )
        nansum = cast(npt.NDArray[np.float64], np.nansum(diffs, axis=1))
        return nansum / valid_steps
```

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+            channel (str): Name of the input channel.
+        """
+        mask = np.ones(self.total_size(channel), dtype=bool)
+        mask[hyp_ids] = False
+        self.data[channel] = self.data[channel][mask]
+        self.age[channel] = self.age[channel][mask]
+
+    def calculate_keep_and_remove_ids(
+        self, num_keep: int, channel: str
+    ) -> Tuple[np.ndarray, np.ndarray]:
```
tristanls: **suggestion**:
```suggestion
    ) -> tuple[npt.NDArray[np.int_], npt.NDArray[np.int_]]:
```

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -276,7 +444,11 @@ def evidence_update_threshold(
         InvalidEvidenceThresholdConfig: If `evidence_threshold_config` is
             not in the allowed values
     """
-    # return 0 for the threshold if there are no evidence scores
+    # Return None to bypass update threshold calculation
+    if evidence_threshold_config is None:
+        return None
```
tristanls: **issue**: I mentioned in a different comment that I think `evidence_threshold_config=None` might be better off returning `0` instead of `None`. But as written, this violates the return type of `-> float`.

tristanls (commented):
Some type comments.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
vkakerbeck: Ok I should have probably placed my comment on a different line (I deliberately went back up to put it at the place where None is checked but I see now that this is not the best place to make the change). But I think my comment is still valid as the `evidence_threshold_config` now has a new None option which is functionally the same as the `all` option https://github.com/thousandbrainsproject/tbp.monty/blob/21b8cc0a559cc45c457c6daa6753ccffabe4eb3a/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L60 I would prefer to keep using the `all` option and rewrite the logic for it to do whatever is done for the None option now.

benchmarks/results/ycb_unsupervised_inference.csv
nielsleadholm: *26,16 right?

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
nielsleadholm: Yeah unless I'm missing something, I agree with your point Viviane that the logic around `all` should just be updated to reflect what `None` is currently doing. E.g. `Using None will bypass threshold calculation and return None` --> `Using "all" will bypass threshold calculation and return None`

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
nielsleadholm: Yeah this is something Ramy and I recently discussed --> between 0 and 1 they will eventually decay to 0, whereas 1 they will eventually grow to infinity. I think capturing your suggestion below Viviane would be good in a follow-up PR:

"When we are uncertain about what we are sensing, we want to keep many options open. But when we have a fairly certain hypothesis, we don't need to keep testing all the other hypotheses. If I am sure about what I am sensing I would just make predictions about that and only when my predictions don't match anymore, do I add more hypotheses back in again."

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -372,13 +401,20 @@ def _sample_existing(
                 evidence=np.zeros(0),
             )
 
+        keep_ids, remove_ids = tracker.calculate_keep_and_remove_ids(
+            num_keep=existing_count,
+            channel=input_channel,
+        )
+
+        # Update tracker by removing the remove_ids
```
nielsleadholm, line 418: **suggestion:** It seems strange that we pass `remove_ids` back to `ResamplingHypothesesUpdater`, only to give it straight back to `tracker`. Since `tracker` already has `input_channel` in the `calculate_keep_and_remove_ids` call, is there any reason you can't do `remove_hyp` as a private method within `tracker`? `calculate_keep_and_remove_ids` would then just be `calculate_keep_ids`

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -251,16 +251,184 @@ def __repr__(self) -> str:
...
+        values = np.array(values, dtype=float)
+        if channel not in self.data:
+            raise ValueError(f"Channel '{channel}' does not exist.")
+
+        if values.shape[0] != self.total_size(channel):
+            raise ValueError(
+                f"Expected {self.total_size(channel)} values, but got {len(values)}"
+            )
+
+        # Shift evidence data by one step
```
nielsleadholm: **suggestion (non-blocking):** Not for this PR, but I wonder whether one optimization might be to treat the array of tracked values as a ring-like structure. Rather than shifting the values every time they are updated, we could just  iterate an index which determines where in the ring we are. Then we would update one column, which based on the index, corresponds to the most recent values.

tests/unit/frameworks/utils/evidence_matching_test.py
```diff
@@ -169,11 +172,95 @@ def test_update_invalid_channel(self) -> None:
...
+    def test_add_hypotheses_initializes_channel(self) -> None:
+        """Test that hypotheses are correctly initialized in a new channel."""
+        self.tracker.add_hyp(2, self.channel)
+        self.assertEqual(self.tracker.total_size(self.channel), 2)
+        self.assertEqual(self.tracker.data[self.channel].shape, (2, 3))
+        self.assertTrue(np.all(np.isnan(self.tracker.data[self.channel])))
+        self.assertTrue(np.all(self.tracker.age[self.channel] == 0))
+
+    def test_update_correctly_shifts_and_sets_values(self) -> None:
+        """Test that update correctly shifts previous values and adds new ones."""
```
nielsleadholm, line 198: **suggestion:** It could be worth checking that if you add even more hypotheses (greater than `window_size`), that this works as expected.

nielsleadholm (commented):
Really nice, thanks Ramy for putting this together. Overall I like the documentation and unit tests, just a couple minor suggestions in addition to what has already been mentioned.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
ramyamounir: Yes, this makes sense. Under the function `evidence_update_threshold()`, if the `evidence_threshold_config` is set to "all", it will return `None` instead of `np.min(evidence_all_channels)`. The modified displacer can now handle this None value.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
ramyamounir: Yeah dynamically changing this value would be good, but it should always be pulled back to 1 since modifying this value will either grow to infinity or decay to 0 as Niels said. So any change here should be temporary (for 1 step, then returns back to 1). We discussed how it might be useful to be controlled by a higher level LM as well.

ramyamounir: > What is a scenario in which we would want to grow the hypothesis space over time?

I think the scenario where we are sure about the object and then we moved to another object (or the object is magically swapped) and now we are unsure. Maybe we would want to increase the space size now.

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: Yes, I could do that. I was just trying to simplify the naming since `data` within the context of `EvidenceSlopeTracker` is unambiguous. I'm referring to @tristanls 's comment about [naming within contexts](https://github.com/thousandbrainsproject/tbp.monty/pull/325#discussion_r2155632316). But I also agree that `data` is very generic, `buffer` might be better.

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: I agree, I'll change `valid` to `removable`, this will be clearer.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
ramyamounir: > Can possible_hypotheses.evidence ever be negative?

Yes, evidence scores can be negative.

I think @vkakerbeck brought up a good point about how we could just modify the logic of existing "all" to be more efficient by returning None in `evidence_update_threshold()` (bypassing `np.min`) and bypassing `np.where` in the displacer. I then realized that I was trying to make the "all" option more efficient which would fall under the `DefaultHypothesesDisplacer`, not really specific to the resampling logic.

I could also create a `ResamplingHypothesesDisplacer` that inherits from `DefaultHypothesesDisplacer` and modifies the `displace_hypotheses_and_compute_evidence`, but I don't think this is necessary now that we are modifying the default's "all" logic.

tristanls: > I could also create a ResamplingHypothesesDisplacer that inherits from DefaultHypothesesDisplacer and modifies the displace_hypotheses_and_compute_evidence

Remember that we don't want to use inheritance for functionality reuse.

ramyamounir: Yeah, I was thinking one level of hierarchy is okay in some scenarios? The alternative would be to copy the other 3 functions under `DefaultHypothesesDisplacer` into the new class, or try to extract the displacer functionality of `displace_hypotheses_and_compute_evidence()` into another class which seems like it would become difficult to understand with many levels of nested compositional components. I don't know, maybe I'm missing a clean way of doing it that doesn't require a hierarchy.

tristanls: @ramyamounir, a `HypothesesSelector` would do it:

```python
class HypothesesSelector(Protocol):
    def select(self, hypotheses, threshold) -> npt.NDArray[np.int_]: ...
```

```python
class DefaultHypothesesSelector:
    def select(self, hypotheses, threshold):
        return np.where(hypotheses.evidence >= threshold)[0]
```

```python
class AllHypothesesSelector:
    def select(self, hypotheses, threshold):
        return np.arange(hypotheses.evidence.shape[0])
```

Then, configure the `DefaultHypothesesDisplacer` as needed with `DefaultHypothesesSelector` being the default.

ramyamounir: Very cool! A simple solution :)

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: Interesting that it would give a type error, the return is indeed `npt.NDArray[np.float64]`.
pic-window-250626-1232-53

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: `0` would not work in this scenario because evidence scores can be less than zero, I kept `None` for now as the return when `evidence_threshold_config="all"`. Let me know if you want this handled in a different way.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
ramyamounir: Yeah, this would work. My bias would be to avoid removing hypotheses from within `calculate_keep_ids`. I wanted to have a simple function that calculates top-k slopes and does nothing else, so that it can be reused in other places if needed (testing, logging, plotting, etc.) If we were to call `remove_hyp` within the tracker itself, I would call the function `calculate_keep_ids_and_remove_unwanted` or something similar, but I prefer for these functions to have only one role if possible for usability and testing purpose.

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: Yes, very interesting idea.

src/tbp/monty/frameworks/utils/evidence_matching.py
tristanls: `mypy` does not execute the code to determine types, so it didn't have enough information to verify (by statically analyzing the code) that the return type is `npt.NDArray[np.float64]`. So, the type casting and explicit conversions in my code snippet seem correct, based on your confirmation.

One interesting aspect, from explicitly annotating the types, was that `valid_steps` starts as an `np.int_` array, but is converted to `np.float64` due to the use of `np.nan` (which appears to require `np.float64` dtype). It probably doesn't matter for performance, but that's one benefit I experienced with type-annotated numpy use: it becomes more visible what sorts of things are happening behind the interface.

src/tbp/monty/frameworks/utils/evidence_matching.py
tristanls: 🤔, what about returning negative Infinity, which is a float, and everything is greater than it:
```python
>>> from math import inf
>>> type(-inf)
<class 'float'>
>>> -23747312343 > -inf
True
```
@jeremyshoemaker, any thoughts on `-inf` instead of a `None` sentinel?

tristanls: I'd add that the other benefit of `-inf` threshold is that the current default displacer code wouldn't need to change and you wouldn't need to refactor `HypothesesSelector`:
```python
            hyp_ids_to_test = np.where(
                possible_hypotheses.evidence >= evidence_update_threshold
            )[0]
```

ramyamounir: Just to say that I was trying to bypass the computation of this `np.where` comparison operation, but it may not be that big of a deal. So, I'm okay with `-inf` or even returning back to `np.min()` of the evidence scores (which was the original implementation of "all").

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: Yeah, you're right. I think I had it in there to support initial testing with lists, but I've changed those unit tests to pass numpy arrays now based on Tristan's comment, so this statement is not needed.

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: Yes, those statements would be equivalent, this would work:
```python
        valid_steps = np.sum(~np.isnan(diffs), axis=1).astype(np.float64)
        valid_steps[valid_steps == 0] = np.nan
```
I don't really have a preference, but I agree that it could be more readable (especially the second line). I have changed the lines and I'll add a bit more description of what these lines are doing.

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: I checked and found out they are equivalent in run time, but what you suggested might be more readable. I've changed the `setdiff1d` to what you suggested.

ramyamounir:
Thank you all for the reviews! @vkakerbeck, I think Niels added great context and answers, I just want to add a couple more points.

> We are recreating np arrays for our hypotheses at every step now. How much overhead does this add? If we are not changing the hypothesis space size, can we solve this differently (like masking as we did before & inplace operations for the newly sampled hypotheses)?

Can you elaborate more on which arrays are recreated. We tackled this problem before [here](https://github.com/thousandbrainsproject/tbp.monty/blob/52d791cf4e5c9183accd55d406204b4a2563fefb/src/tbp/monty/frameworks/utils/evidence_matching.py#L235), but the addition of the slope tracker doesn't recreate these arrays. It only tells us which hypotheses to drop and we do that by [indexing](https://github.com/thousandbrainsproject/tbp.monty/blob/ce84000dfda949dcb4a294493bf7720931f0a1c0/src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py#L415). I could be missing something.

>  How does the slope tracker and deletion of hypotheses relate to evidence decay and thresholding? It seems like allowing evidence to decay if no new evidence is added (similar to past and present weights) is similar to tracking the slope of evidence updates. Removing hypotheses seems similar to masking them using a threshold.

I think the ideas are relatable. But deleting hypotheses that do not accumulate evidence is a much faster way than decaying, especially when moving between objects (No Reset setting). Because we are quickly deleting, we are then able to quickly sample more in their place. With decay and threshold, the previous object has to decay first before we can start accumulating evidence on new hypotheses, and assuming that we sample informed hypotheses every step, it will take a lot of time to decay before we can start accumulating on new hypotheses. That said, I'm sure there might be away to get decaying to work, I just think it will be much slower.

> Now that you are testing this on the other benchmarks, I am curious how it will interact with the terminal condition checks. I could imagine that we may run into some difficulties there.

Yeah, I'll work on getting it to run on one of the default benchmarks very soon. It will be interesting to see the runtime there.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
nielsleadholm: That makes sense, thanks for clarifying.

src/tbp/monty/frameworks/utils/evidence_matching.py
vkakerbeck: okay yeah I guess I like to tend more on the side of verbosity over risking confusion. I feel like even within the EvidenceSlopeTracker, data or buffer could be referring to an array of slopes, an array of differences, or an array of evidences.

src/tbp/monty/frameworks/utils/evidence_matching.py
jeremyshoemaker: > any thoughts on `-inf` instead of a `None` sentinel?

It seems reasonable to me in this context. Returning `None` requires an explicit check on the caller's part to avoid an erroneous comparison. The only thing that `None` would give you here that I think `-inf` doesn't is that it's falsey, but that depends on the context whether it's useful or not.

src/tbp/monty/frameworks/utils/evidence_matching.py
vkakerbeck: That's a good idea. Additionally we could also just actually only track the slopes here. There is no need to recalculate the diffs for all the values in the window at every step. We just need the diff between the newest set of values and the last one. Older diffs will remain the same.

vkakerbeck:
Very nice, thanks for the updates! It all looks good to me :)

Regarding the recreation of np.arrays I was referring to the np.vstack calls in `update_hypotheses` but I see now that that wasn't actually changed in this PR. It could still be something worth coming back to.
In general, maybe make sure to either add some comments in places that we highlighted here for investigating further or keep a list locally for you so the ideas for potential speedups don't get lost in the PR review.

Regarding the decaying vs. deleting, I agree that hypotheses get deleted faster here now compared to decay + thresholding. The decau mechanism would require the bounded evidence (for example achieved with past and present weight). If we had the evidence bound to a certain range then a couple of steps with inconsistent observations should quickly get us below a fixed threshold. But yeah, nothing to test right now.

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: Yeah makes sense, I've renamed `data` -> `evidence_buffer` and `age` -> `hyp_age`.

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: Since this is related to another discussion [here](https://github.com/thousandbrainsproject/tbp.monty/pull/340#discussion_r2168985636), I'm tagging @vkakerbeck and @nielsleadholm to reach a consensus.

The options are:
- Leave it as is (`None`), which bypasses `np.min` in `evidence_update_threshold()` and `np.where` in the displacer. This may not be the best solution for the reasons Tristan and Jeremy have mentioned.
- Change to `-inf`, which would only bypass the `np.min` but still do the comparison operation in the displacer.
- Change back to the original behavior of `np.min` and `np.where`. This is fine too, I don't think bypassing these functions saves a lot of computations unless the hyp space is too large.

ramyamounir:
Thanks for clarification. For what it's worth, I am not too worried about the runtime in these experiments, I added another reason in the description for why the default benchmark experiments would be faster than No Reset (`run_parallel.py` vs. `run.py`). I also started looking into running one of the default benchmarks with resampling and from preliminary testing I can already see around 50% reduction in runtime without sacrificing much performance. Although, this could just be the difference from running locally and on AWS; either way resampling doesn't seem to be adding much overhead overall. We can discuss this in more detail in the technical meeting.

vkakerbeck:
Hm it still looks like the relative runtime change in the tables you show is >2x. Even if it can be sped up with run_parallel for the normal benchmark that is still a relative change that will remain, right? Also, I wouldn't compare runtimes on the laptop to AWS runtimes. Usually the laptop is much faster.

Generally, while I don't think runtimes are going out of the range of practical (so no complete dealbreaker), I would still try to make it as efficient as possible since we run these experiments a lot and a 2x or even 1.5x factor quickly adds up to saving us a lot of time. But yeah something to talk about more later :)

src/tbp/monty/frameworks/utils/evidence_matching.py
vkakerbeck: @ramyamounir just sicne I saw you added a note for other optimization ideas, to remember this one too

src/tbp/monty/frameworks/utils/evidence_matching.py
nielsleadholm: Yeah I think if we haven't specifically demonstrated that this is a big slow-down, then it could just be cleaner to use the original behavior (option 3), assuming that doesn't require too many revert-style changes on your part.

No matter what option we go for, it is likely that we will make "all" the default in the long term, and `np.min`/`np.where` won't be relevant (can be entirely removed). The third option has the benefit of being the cleanest solution until that eventual refactor, which seems preferable if we it's not a significant difference in compute overhead.

ramyamounir: I agree, this is the cleanest solution for now. I'll revert back and leave a note there that it can be optimized later.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -250,12 +251,216 @@ def __repr__(self) -> str:
...
+        return to_keep, to_remove
+
+
 def evidence_update_threshold(
     evidence_threshold_config: float | str,
     x_percent_threshold: float | str,
     max_global_evidence: float,
     evidence_all_channels: np.ndarray,
-) -> float:
+) -> float | None:
```
tristanls, line 463: **todo**: @ramyamounir, please remove this typing change. With the revert, it is my understanding this method can no longer return `None`.

ramyamounir: Yes, I missed this. I opened a [PR](https://github.com/thousandbrainsproject/tbp.monty/pull/360).

PR #341, opened 2025-06-21
feat: DISABLE_WANDB envrionment variable

ramyamounir:
This adds support for DISABLE_WANDB environment variable which simply clears the `wandb_handlers` from the configs. To disable w&b, you can run the commands with inline environment variable, e.g., 

```bash
DISABLE_WANDB=1 python benchmarks/run.py -e base_config_10distinctobj_dist_agent
#OR
DISABLE_WANDB=1 python benchmarks/run_parallel.py -e base_config_10distinctobj_dist_agent
```

### DIFF
--- a/docs/how-to-use-monty/getting-started.md
+++ b/docs/how-to-use-monty/getting-started.md
@@ -215,7 +215,7 @@ If you want to run an experiment with parallel processing to make use of multipl
 python benchmarks/run_parallel.py -e my_experiment
 ```
 
-If you wish to disable W&B for benchmark experiments, you can pass the environment variable `DISABLE_WANDB=1` as a prefix to the run command.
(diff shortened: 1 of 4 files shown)


ramyamounir:
FYI, @thousandbrainsproject/monty-maintainers

tristanls (changes requested):
I think this pull request is unnecessary as WandB appears to already have `WANDB_MODE` environment variable:

> 
WANDB_MODE | If you set this to “offline” wandb will save your run metadata locally and not sync to the server. If you set this to `disabled` wandb will turn off completely.
-- | --

See https://docs.wandb.ai/guides/track/environment-variables/#optional-environment-variables

ramyamounir:
@tristanls, I have been using the command `wandb disabled` which returns `W&B disabled.`, but doesn't really work. The monty_handlers seem to turn it back on. However, I didn't know about the environment variable, setting `WANDB_MODE="disabled"` works great, thanks. I've closed this PR.

PR #342, opened 2025-06-23
refactor!: replace logging BasicConfig with standard tiered config

tristanls:
This pull request replaces the original global logging basic configuration with a standard tiered configuration.

This will allow us to create very detailed level, filter, handler, formatter, configuration per module level.

Note that `python_log_to_stdout` option was misnamed, as it attached a non-configured `logging.StreamHandler`, which logs to `sys.stderr` by default.

The updated Python logger usage is now to get the logger by module name at the top of the module:
```python
# ...
from tbp.monty.frameworks.models.object_model import GraphObjectModel

class MontyForGraphMatching(MontyBase):
    """General Monty model for recognizing object using graphs."""
# ...
```

Additionally, Python loggers are now initialized before the Monty model so that logging during initialization of the Monty model works.

The "logging" of metrics used for statistics and plotting is unaffected.

The default log format is updated to `"%(levelname)s:%(name)s:%(funcName)s:%(lineno)d:%(message)s"`

Example of new default python logging output:
```
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:235:Sending input from ['patch'] to learning_module_0
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 2730 out of 4440 hypotheses for mug (evidence > 2.8473207648464793)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 1206 out of 1342 hypotheses for dice (evidence > 2.8473207648464793)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 2558 out of 5888 hypotheses for bowl (evidence > 2.8473207648464793)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 2413 out of 4878 hypotheses for potted_meat_can (evidence > 2.8473207648464793)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 3692 out of 3850 hypotheses for spoon (evidence > 2.8473207648464793)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 1684 out of 4166 hypotheses for mustard_bottle (evidence > 2.8473207648464793)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 933 out of 1658 hypotheses for strawberry (evidence > 2.8473207648464793)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 640 out of 640 hypotheses for golf_ball (evidence > 2.8473207648464793)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 1970 out of 3814 hypotheses for c_lego_duplo (evidence > 2.8473207648464793)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 1482 out of 1490 hypotheses for banana (evidence > 2.8473207648464793)
INFO:tbp.monty.frameworks.models.evidence_matching.learning_module:_calculate_most_likely_hypothesis:1156:current most likely hypothesis: spoon with evidence 15.65
INFO:tbp.monty.frameworks.models.graph_matching:_get_union_of_possible_matches:287:Possible matches for LM 0: {'banana', 'dice', 'potted_meat_can', 'bowl', 'mug', 'spoon'}
INFO:tbp.monty.frameworks.models.monty_base:update_step_counters:488:--- Global Matching Step 11 ---
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:235:Sending input from ['patch'] to learning_module_0
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 926 out of 1658 hypotheses for strawberry (evidence > 3.129311527623278)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 2721 out of 4440 hypotheses for mug (evidence > 3.129311527623278)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 640 out of 640 hypotheses for golf_ball (evidence > 3.129311527623278)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 2395 out of 4878 hypotheses for potted_meat_can (evidence > 3.129311527623278)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 2545 out of 5888 hypotheses for bowl (evidence > 3.129311527623278)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 1681 out of 4166 hypotheses for mustard_bottle (evidence > 3.129311527623278)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 1197 out of 1342 hypotheses for dice (evidence > 3.129311527623278)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 3663 out of 3850 hypotheses for spoon (evidence > 3.129311527623278)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 1477 out of 1490 hypotheses for banana (evidence > 3.129311527623278)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:151:Testing 1915 out of 3814 hypotheses for c_lego_duplo (evidence > 3.129311527623278)
```

Lastly, types associated with the changes to the logger initialization are updated. This required an update to `config_to_dict` as it did more than it claimed and could return `Any`, which was unnecessarily broad. Instead, the recursive call to `config_to_dict` is now guarded by the `is_config_like` check.

### DIFF
--- a/src/tbp/monty/frameworks/run.py
+++ b/src/tbp/monty/frameworks/run.py
@@ -17,6 +17,8 @@ import time
 from tbp.monty.frameworks.config_utils.cmd_parser import create_cmd_parser
 from tbp.monty.frameworks.utils.dataclass_utils import config_to_dict
 
+logger = logging.getLogger(__name__)
+
 
 def merge_args(config, cmd_args=None):
     """Override experiment "config" parameters with command line args.
@@ -110,4 +112,4 @@ def main(all_configs, experiments=None):
         os.makedirs(exp_config["logging_config"]["output_dir"], exist_ok=True)
         start_time = time.time()
         run(exp_config)
-        logging.info(f"Done running {experiment} in {time.time() - start_time} seconds")
+        logger.info(f"Done running {experiment} in {time.time() - start_time} seconds")
--- a/src/tbp/monty/frameworks/utils/dataclass_utils.py
+++ b/src/tbp/monty/frameworks/utils/dataclass_utils.py
@@ -7,11 +7,16 @@
 # Use of this source code is governed by the MIT
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
+from __future__ import annotations
 
 import dataclasses
 import importlib
 from inspect import Parameter, signature
-from typing import Callable, Optional, Type
+from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Type
+from typing_extensions import TypeIs
+
+if TYPE_CHECKING:
+    from _typeshed import DataclassInstance
 
 __all__ = [
     "as_dataclass_dict",
@@ -150,7 +155,7 @@ def create_dataclass_args(
     return dataclasses.make_dataclass(dataclass_name, _fields, bases=bases, frozen=True)
 
 
-def config_to_dict(config):
+def config_to_dict(config: DataclassInstance | Dict[str, Any]) -> Dict[str, Any]:
     """Convert config composed of mixed dataclass and dict elements to pure dict.
 
     We want to convert configs composed of mixed dataclass and dict elements to
@@ -158,14 +163,33 @@ def config_to_dict(config):
 
     TODO: Remove once all other configs are converted to dict only
 
+    Args:
+        config(DataclassInstance | Dict[str, Any]): Config to convert to dict.
+
     Returns:
-        Pure dict version of config.
+        (Dict[str, Any]): Pure dict version of config.
     """
     if isinstance(config, dict):
-        return {k: config_to_dict(v) for k, v in config.items()}
+        return {
+            k: config_to_dict(v) if is_config_like(v) else v for k, v in config.items()
+        }
     if dataclasses.is_dataclass(config):
         return dataclasses.asdict(config)
-    return config
+
+
+def is_config_like(obj: Any) -> TypeIs[DataclassInstance | Dict[str, Any]]:
+    """Returns True if obj is a dataclass or dict, False otherwise.
+
+    Args:
+        obj(Any): Object to check.
+
+    Returns:
+        (TypeIs[DataclassInstance | Dict[str, Any]]): True if config is a dataclass or
+            dict, False otherwise.
+    """
+    return isinstance(obj, dict) or (
+        dataclasses.is_dataclass(obj) and not isinstance(obj, type)
+    )
 
 
 def get_subset_of_args(arguments, function):
--- a/src/tbp/monty/frameworks/config_utils/config_args.py
+++ b/src/tbp/monty/frameworks/config_utils/config_args.py
@@ -111,7 +111,7 @@ class LoggingConfig:
     wandb_handlers: list = field(default_factory=list)
     python_log_level: str = "INFO"
     python_log_to_file: bool = True
-    python_log_to_stdout: bool = True
+    python_log_to_stderr: bool = True
     output_dir: str = os.path.expanduser(
         os.path.join(monty_logs_dir, "projects/monty_runs/")
     )
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -49,6 +49,8 @@ __all__ = [
     "SaccadeOnImageFromStreamDataLoader",
 ]
 
+logger = logging.getLogger(__name__)
+
 
 class EnvironmentDataset(Dataset):
     """Wraps an embodied environment with a :class:`torch.utils.data.Dataset`.
@@ -321,7 +323,7 @@ class EnvironmentDataLoaderPerObject(EnvironmentDataLoader):
         Also add any potential distractor objects.
         """
         next_object = (self.current_object + 1) % self.n_objects
-        logging.info(
+        logger.info(
             f"\n\nGoing from {self.current_object} to {next_object} of {self.n_objects}"
         )
         self.change_object_by_idx(next_object)
@@ -366,7 +368,7 @@ class EnvironmentDataLoaderPerObject(EnvironmentDataLoader):
             "semantic_id": self.semantic_label_to_id[self.object_names[idx]],
             **self.object_params,
         }
-        logging.info(f"New primary target: {pformat(self.primary_target)}")
+        logger.info(f"New primary target: {pformat(self.primary_target)}")
 
     def add_distractor_objects(
         self, primary_target_obj, init_params, primary_target_name
@@ -632,7 +634,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
         Returns:
             The observation from the jump attempt.
         """
-        logging.debug(
+        logger.debug(
             "Attempting a 'jump' like movement to evaluate an object hypothesis"
         )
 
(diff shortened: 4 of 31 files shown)


src/tbp/monty/frameworks/utils/dataclass_utils.py
```diff
@@ -150,22 +155,41 @@ class SingleSensorAgentArgs:
...
+        (Dict[str, Any]): Pure dict version of config.
     """
     if isinstance(config, dict):
-        return {k: config_to_dict(v) for k, v in config.items()}
+        return {
+            k: config_to_dict(v) if is_config_like(v) else v for k, v in config.items()
+        }
     if dataclasses.is_dataclass(config):
         return dataclasses.asdict(config)
-    return config
```
tristanls, line 168: **note**: When taking typing into account, line 168 becomes unreachable and is removed.

jeremyshoemaker: This should be changed to an `else:` since there are only two possibilities given the typing.

src/tbp/monty/frameworks/utils/dataclass_utils.py
```diff
@@ -150,22 +156,41 @@ class SingleSensorAgentArgs:
...
+    """Returns True if obj is a dataclass or dict, False otherwise.
+
+    Args:
+        obj(Any): Object to check.
+
+    Returns:
+        (TypeIs[DataclassInstance | Dict[str, Any]]): True if config is a dataclass or
+            dict, False otherwise.
+    """
+    return isinstance(obj, dict) or (
```
jeremyshoemaker: Kinda wish I could do multi-line suggested changes, but I'm wondering if this might be better written like this:

```python
if isinstance(obj, type):
    return False
return isinstance(obj, dict) or dataclasses.is_dataclass(obj)
```

That said, I feel like this function _only_ needs to exist because our config dictionaries are `Dict[str, Any]` when we could work on making that type narrower. That's a different bit of work, though.

src/tbp/monty/frameworks/run.py
jeremyshoemaker, line 1: Since this is the entrypoint for Monty (or one of them anyway), should we add in a call to `logging.config.fileConfig()` to load the logging config from a file if it exists so that we can control the logging at runtime?

src/tbp/monty/frameworks/utils/dataclass_utils.py
tristanls: Updated.

src/tbp/monty/frameworks/utils/dataclass_utils.py
tristanls: **note**: You can do multi-line suggested changes

https://github.com/user-attachments/assets/4042f8af-5c66-4fb0-b22a-bc9b053677c4

---
Having said that, the `isinstance(obj, type)` and `dataclasses.is_dataclass(obj)` go together because they distinguish between `type[DataclassInstance]` and `DataclassInstance`, and we don't want to accept `type[DataclassInstance]` because `dataclasses.as_dict(...)` will throw a `TypeError`.

But overall, the `Dict[str, Any]` needs to change in the future. I figure `Any` will be our marker to return to this in the future to improve typing.

tristanls: 🤔 I could clarify the intent by making `is_dataclass_instance` check.

tristanls: Updated to be clearer.

src/tbp/monty/frameworks/run.py
tristanls: I was looking to do something like this. I feel it should be in a different PR because the configuration will be split. Currently, logging configuration is part of the one experiment configuration. If we are to load config from a file (I would probably do `dictConfig` and allow for overrides), then logging will be separate, leading to another discussion and design choices. Hence, it should be a different PR.

src/tbp/monty/frameworks/utils/dataclass_utils.py
tristanls: I have no idea why this comment thread detached from the `config_to_dict` function 🫠 .

tristanls:
@jeremyshoemaker, thanks for the review, I made a few updates per your suggestions, if you could take a look.

src/tbp/monty/frameworks/utils/dataclass_utils.py
jeremyshoemaker: > You can do multi-line suggested changes

I don't know how I managed to miss that, after trying to do exactly that. 🤷

jeremyshoemaker: Back on the topic at hand. I'm not sure that it's better to make a whole extra function just for this check. Just so we're on the same page, my thinking is like this. I'm sure you're aware of a lot of the details already, but it helps to be clear sometimes. 

The original code was:
```python
return isinstance(obj, dict) or (
    dataclasses.is_dataclass(obj) and not isisntance(obj, type)
)
```
After poking around and figuring out _why_ you were doing the check for a `type`, this block of code results in this mapping table:

| `obj` type | return value |
|------------|-------------|
| dict | True |
| dataclass (sortof) | True |
| type | False |
| anything else | False |

The _only_ reason we need to do the check for `type` is because given the following code...

```python
from dataclasses import dataclass

@dataclass
class Foo:
    bar: str

f = Foo()
```

...the result of `is_dataclass(Foo)` is True and `isinstance(Foo, type)` is also True.

What we're _actually_ trying to do here is limit the truthy values to instances of `dict` and instances of classes decorated with `@dataclass`. In either of those two cases, we don't want a `type`. The `type` check is implicit in checking whether it's a `dict`, so it's really only _needed_ in the check for the dataclass. That said, in either case, since we don't want a `type`, to me it's clearer to knock that test out of the way up front and then get more specific after that. Leading to the code I suggested:

```python
if isinstance(obj, type):
    return False
return isinstance(obj, dict) or dataclasses.is_dataclass(obj)
```

Yes, this does result in two instance checks instead of one, but it's probably not in a high-performance part of the code, so the optimization isn't worth the clarity tradeoff to me. :shrug:

The real problem here is that `is_dataclass` returns True for `type`s and not just instances. But that's part of the standard library. They really should provide the `is_dataclass_instance` function instead of just suggesting it in the documentation.

jeremyshoemaker (approved):

src/tbp/monty/frameworks/utils/dataclass_utils.py
tristanls: OK, I'll update per your recommendation.

PR #343, opened 2025-06-24
Abstraction Layer for KNN Backend 

collin-allen512:
## Overview
In this PR, I have replaced the direct use of the SciPy KDTree implementation for KNN search with a KNNIndex class which supports both CPU (KDTree SciPy) and GPU (FAISS IndexIVFFlat Index) implementations.

The implementation of the KNN backend is in src/tbp/monty/frameworks/utils/knn_search.py. This implementation includes the KNNIndex abstract class, the CPU implementation as the KDTreeIndex class, the GPU implementation as the FAISSIndex class, as well as a KNNSearchFactory class for creating the desired Index. 

## Profiling
In order to compare the impact of the CPU and GPU backends, I added a custom KNNProfiler in src/tbp/monty/frameworks/utils/knn_profiler.py. This tracks the build and search times throughout an experiment, saving the profiling results to csv files. 

This profiler is separate from the Monty logging code, and may not be appropriate for the full repository. I include it here for discussion, but we may decide it is more appropriate to remove or better integrate with the rest of the codebase. 

## Global KNN Backend Definition
In this initial implementation, I wanted to be able to adjust the KNN backend for the entire experiment definition, rather than for each learning module and object model independently. This way I could efficiently run the existing benchmarks with minimal configuration changes. However, as this approach does not seem to be used for any existing config parameters, this might not be the right implementation for the Monty codebase. 

In terms of implementation, I added a KNNBackendArgs config which is passed to each MontyConfig. Then in order to update each LM, the KNN backend settings are passed to each LM when it is initialized by the MontyExperiment.  It led to a couple of code changes that seemed like they might not be the right approach, so please provide feedback if there is a more elegant/more Monty appropriate approach to take.

## CPU implementation
The KDTreeIndex is designed to provide the same functionality as the current KNN approach and uses the same SciPy KDTree data structure. By setting the backend to CPU (which is the default), there should be no difference between experiments run before/after these changes. 

## GPU implementation
The GPU implementation, in the FAISSIndex class, uses the FAISS library's IVFFlatIndex to handle GPU acceleration. This algorithm divides the dataset into clusters, defined by the 'nlist' parameter, and is designed for fast parallel execution across GPU SIMD cores. For faster, but only approximate KNN search, the nprobes parameter can be set to a value lower than nlist. However, in order to stay consistent with the current exact KNN methodology I set nprobes equal to nlist to ensure that every cluster is visited. 

If multiple GPUs are available on the system they can be specified through the "gpu_id" parameter. The batch size for GPU inferences can also be adjusted through the "batch_size" parameter, which is set to a default of 65535 based on what my Nvidia 4090 Laptop GPU can support.

## Global FAISS GPU resources
I hit some issues before adjusting the code to only have one instance of the FAISS GPU resource, enforcing all instances of FAISSIndex to reference this one singleton. This is a bit clunky, and causes a Ruff error, so if there is a better approach then I am open to adjusting. 

## Analysis
Using the KNN Profiler to profile several of the benchmark experiments, there is a clear trend: the CPU backend is faster on average for the current typical Monty settings, but as the number of queries increases the GPU backend scales better and is more consistent. 

Below is a summary of the data for several of the benchmark experiments when using the CPU backend:
image

And here for the GPU backend:
image

Consistently the CPU runs have lower mean and total runtimes. However, the GPU experiments consistently have lower max values. Note that I turned multi-threading off for these experiments for more consistent profiling data. 

These plots below show the query time plotted against the size of the query for all queries in each experiment. Note that this does not include the number of nodes in the graph, which will also effect the runtime significantly. 

image

image

image

## Conclusion
The basic take away is that right now the current CPU-based approach is faster on average, and there isn't an advantage to switching to this GPU backend yet. However, this backend could still be of value for larger experiments down the line. Considering there does not seem to be value add for this code change right now, I understand if this PR is not ready to be added to the main codebase. I wanted to go ahead and submit this PR to share my findings, get feedback, and see if this is of value to the Monty community. Looking forward to hearing feedback!


tristanls:
Hi @collin-allen512, and thank you for this pull request! This is a great write-up and an excellent prototype. Thank you for sharing the code.

As you concluded, since the results do not yet convey an advantage, I think we should hold off on merging into Monty. 

I think it's worth keeping this pull request open as we continue to develop Monty and need to perform larger experiments where this technique will generate a speedup. The general approach of working with multiple backends is also a great example of what use cases Monty should support in the future. (If we were to merge this PR in the future, I would recommend/help with some refactoring to integrate the code into the framework.)

Additionally, I would like to keep this pull request open as an example of how to prototype functionality in Monty and evaluate it, so I very much appreciate you putting it together.

src/tbp/monty/frameworks/utils/knn_search.py
```diff
@@ -0,0 +1,402 @@
...
+        Returns:
+            Tuple of (distances, indices) or just indices if return_distance=False
+        """
+        # Ensure we're working with float32
+        query_points = np.ascontiguousarray(query_points.astype("float32"))
+
+        n_queries = query_points.shape[0]
+        start_time = time.time()
+
+        if n_queries > sekf.batch_size:
```
tristanls: **issue**: Looks like `self` is misspelled.

**question**: Did this code execute during your benchmarks? I assume it would throw an error if it did.

collin-allen512: This is what I get for trying to squeeze in a last minute commit ...

I saw I wasn't actually using the batch_size param, so did a small update but didn't rerun all of the tests. My bad :)

vkakerbeck:
Hi @collin-allen512, thanks for taking this deep dive and sharing the detailed results, evaluations, and code in this PR! This is really great to see :) Even though it currently doesn't outperform the kd-tree search, it is still good to have this result and the code available in this PR if it becomes more advantageous in the future.

PR #351, opened 2025-06-26
chore: demonstrate initializing TouchObject alters benchmarks

tristanls:
This pull request partially implements https://github.com/thousandbrainsproject/tbp.monty/pull/339. **It demonstrates that simply initializing `TouchObject` will cause benchmark deviations.** 🤔 

The goal of having this pull request separate is to isolate and merge changes that do not alter benchmarks, while working towards identifying the most minor possible change responsible for benchmark changes in https://github.com/thousandbrainsproject/tbp.monty/pull/339.

Since there are no algorithm changes, the culprit appears to be a change in `rng` sequence due to `BasePolicy.__init__` code, where a random action is initialized, which uses the `rng`:
```python
        # Ensure our first action only samples from those that can be random
        self.action: Action | None = self.get_random_action(
            self.action_sampler.sample(self.agent_id)
        )
```

## Benchmarks

_comparison with fb1e7d0d910454067627682b5796de00a2b21c4a_

base_config_10distinctobj_dist_agent,100.00,0.00,37,16.20,2,9 ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/0fqswz3r/overview
base_config_10distinctobj_surf_agent,100.00,0.71,27,17.02,2,11 ❌ (worse) - https://wandb.ai/thousand-brains-project/Monty/runs/39duz0x2/overview
```diff
-base_config_10distinctobj_surf_agent,100.00,0.00,28,14.01,2,11
+base_config_10distinctobj_surf_agent,100.00,0.71,27,17.02,2,11 
```
randrot_noise_10distinctobj_dist_agent,100.00,3.00,55,22.49,3,21 ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/3jw0robr/overview
randrot_noise_10distinctobj_dist_on_distm,99.00,1.00,38,12.15,2 ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/kio3z60p/overview
randrot_noise_10distinctobj_surf_agent,99.00,1.00,28,26.27,3,20 ❌ (worse) - https://wandb.ai/thousand-brains-project/Monty/runs/m3os8itx/overview
```diff
-randrot_noise_10distinctobj_surf_agent,100.00,1.00,29,20.42,3,21
+randrot_noise_10distinctobj_surf_agent,99.00,1.00,28,26.27,3,20
```
randrot_10distinctobj_surf_agent,100.00,1.00,27,18.99,2,11 ❌ (worse) - https://wandb.ai/thousand-brains-project/Monty/runs/feugazb7/overview
```diff
-randrot_10distinctobj_surf_agent,100.00,0.00,28,17.23,2,10
+randrot_10distinctobj_surf_agent,100.00,1.00,27,18.99,2,11
```
randrot_noise_10distinctobj_5lms_dist_agent,100.00,0.00,57,45.48,5,42 ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/svcsq7s0/overview
base_10simobj_surf_agent,95.71,7.86,68,16.67,5,25 😐 (different) - https://wandb.ai/thousand-brains-project/Monty/runs/8rzzeyco/overview
```diff
-base_10simobj_surf_agent,93.57,8.57,70,13.98,6,27
+base_10simobj_surf_agent,95.71,7.86,68,16.67,5,25
```
randrot_noise_10simobj_dist_agent,84.00,40.00,237,35.01,11,96 ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/bm6lqsl1/overview
randrot_noise_10simobj_surf_agent,91.00,38.00,181,22.39,18,147 ❌ (worse) - https://wandb.ai/thousand-brains-project/Monty/runs/grt39f4s/overview
```diff
-randrot_noise_10simobj_surf_agent,93.00,34.00,172,20.94,16,137
+randrot_noise_10simobj_surf_agent,91.00,38.00,181,22.39,18,147
```
randomrot_rawnoise_10distinctobj_surf_agent,68.00,74.00,15,89.75,3,6 👍 (better) - https://wandb.ai/thousand-brains-project/Monty/runs/q853aeog/overview
```diff
-randomrot_rawnoise_10distinctobj_surf_agent,64.00,77.00,16,106.51,4,7
+randomrot_rawnoise_10distinctobj_surf_agent,68.00,74.00,15,89.75,3,6
```
base_10multi_distinctobj_dist_agent,79.29,10.71,31,19.84,3,1 ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/ihry8ce4/overview

base_77obj_dist_agent,93.51,12.99,108,15.86,19,62 ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/s6669ojf/overview
base_77obj_surf_agent,98.27,4.76,52,6.47,12,36 😐 (different) - https://wandb.ai/thousand-brains-project/Monty/runs/7mw8tpof/overview
```diff
-base_77obj_surf_agent,98.70,6.49,56,12.94,12,34
+base_77obj_surf_agent,98.27,4.76,52,6.47,12,36
``` 
randrot_noise_77obj_dist_agent,89.61,22.51,152,37.47,34,123 ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/kyb6cr08/overview
randrot_noise_77obj_surf_agent,94.81,24.24,119,33.25,35,125 👍 (better) - https://wandb.ai/thousand-brains-project/Monty/runs/zfts3fbb/overview
```diff
-randrot_noise_77obj_surf_agent,93.94,23.38,115,34.80,32,113
+randrot_noise_77obj_surf_agent,94.81,24.24,119,33.25,35,125
```
randrot_noise_77obj_5lms_dist_agent,88.31,0.00,70,58.25,9,87 ✅ - https://wandb.ai/thousand-brains-project/Monty/runs/rp45zd7d/overview

### DIFF
--- a/src/tbp/monty/frameworks/models/motor_policies.py
+++ b/src/tbp/monty/frameworks/models/motor_policies.py
@@ -400,6 +400,25 @@ class PositioningProcedure(BasePolicy):
     indicates that the procedure has terminated or truncated.
     """
 
+    @staticmethod
+    def depth_at_center(agent_id: str, observation: Any, sensor_id: str) -> float:
+        """Determine the depth of the central pixel for the sensor.
+
+        Args:
+            agent_id (str): The ID of the agent to use.
+            observation (Observations): The observation to use.
+            sensor_id (str): The ID of the sensor to use.
+
+        Returns:
+            (float): The depth of the central pixel for the sensor.
+        """
+        # TODO: A lot of assumptions are made here about the shape of the observation.
+        #       This should be made robust.
+        observation_shape = observation[agent_id][sensor_id]["depth"].shape
+        return observation[agent_id][sensor_id]["depth"][
+            observation_shape[0] // 2, observation_shape[1] // 2
+        ]
+
     @abc.abstractmethod
     def positioning_call(
         self,
@@ -837,40 +856,6 @@ class InformedPolicy(BasePolicy, JumpToGoalStateMixin):
 
         return super().pre_episode()
 
-    def get_depth_at_center(self, raw_observation, view_sensor_id, initial_pose=True):
-        """Determine the depth of the central pixel.
-
-        Method primarily used by surface-agent, but also by distant agent after
-        performing a hypothesis-testing jump; determines the depth of the central pixel,
-        to inform whether the object is visible at all
-
-        initial_pose : Whether we are checking depth-at center as part of the first
-            start of an experiment; if using get_depth_at_center to check for
-            the observation after e.g. a hypothesis-testing jump, then we don't
-            want to throw an error if the object is nowhere to be seen; instead, we
-            simply move back
-
-        Returns:
-            Depth at the center of the view sensor.
-        """
-        observation_shape = raw_observation[self.agent_id][view_sensor_id][
-            "depth"
-        ].shape
-        depth_at_center = raw_observation[self.agent_id][view_sensor_id]["depth"][
-            observation_shape[0] // 2, observation_shape[1] // 2
-        ]
-        if initial_pose:
-            assert depth_at_center > 0, (
-                "Object must be initialized such that "
-                "agent can visualize it by moving forward"
-            )
-            # TODO investigate - I think this may have always been passing in the
-            # original surface-agent policy implementation because the surface
-            # sensor clips at 1.0, so even if the object isn't strictly visible (or
-            # we're inside the object))  there's a risk we have initialized without the
-            # object in view
-        return depth_at_center
-
     ###
     # Methods that define behavior of __call__
     ###
@@ -1147,7 +1132,11 @@ class SurfacePolicy(InformedPolicy):
             (MoveForward | OrientHorizontal | OrientVertical): Action to take.
         """
         # If the viewfinder sees the object within range, then move to it
-        depth_at_center = self.get_depth_at_center(raw_observation, view_sensor_id)
+        depth_at_center = PositioningProcedure.depth_at_center(
+            agent_id=self.agent_id,
+            observation=raw_observation,
+            sensor_id=view_sensor_id,
+        )
         if depth_at_center < 1.0:
             distance = (
                 depth_at_center
--- a/src/tbp/monty/frameworks/models/motor_policies.py
+++ b/src/tbp/monty/frameworks/models/motor_policies.py
@@ -1313,9 +1313,13 @@ class SurfacePolicy(InformedPolicy):
 
         # TODO: Remove this once TouchObject positioning procedure is implemented
         """
+        if self.attempting_to_find_object:
+            # When the TouchObject positioning procedure is separated, there
+            # will be no post_action calls when attempting to find the object.
+            return
+
         super().post_action(action, state)
-        if not self.attempting_to_find_object:
-            self.last_surface_policy_action = action
+        self.last_surface_policy_action = action
 
     def _orient_horizontal(self, state: MotorSystemState) -> OrientHorizontal:
         """Orient the agent horizontally.
--- a/src/tbp/monty/frameworks/models/motor_policies.py
+++ b/src/tbp/monty/frameworks/models/motor_policies.py
@@ -34,7 +34,10 @@ import quaternion as qt
 import scipy.ndimage
 from scipy.spatial.transform import Rotation as rot  # noqa: N813
 
-from tbp.monty.frameworks.actions.action_samplers import ActionSampler
+from tbp.monty.frameworks.actions.action_samplers import (
+    ActionSampler,
+    UniformlyDistributedSampler,
+)
 from tbp.monty.frameworks.actions.actions import (
     Action,
     ActionJSONDecoder,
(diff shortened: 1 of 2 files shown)


src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -1135,6 +1286,33 @@ def touch_object(
...
+        #     self.attempting_to_find_object = True
+        #     if isinstance(action, MoveForward):
+        #         # Reset TouchObject positioning procedure
+        #         self.touch_object_pos_proc = None
+        #         # Let dynamic_call proceed to take action
+        #         self.attempting_to_find_object = False
+        #     return action
+
+        # raise RuntimeError("touch_object should not have been called")
+
```
tristanls, line 1315: **note**: Please note that methods on `TouchObject` are _never called_.

tristanls:
@vkakerbeck, @nielsleadholm, take a look. I'm not looking for approval, but more of a discussion on how to move forward.

nielsleadholm:
Nice, thanks for isolating this! Just to double check my reading of the changes in the results, this shows that the benchmark results will change (as you say, due to the seed changing). On its own that's fine.

However, this code change does not cause several of the episode run times to approximately double, or to have a non-terminating experiment like `randrot_noise_10simobj_surf_agent` right?

Is it possible to further introduce the other changes until that undesirable behavior appears? Maybe it is just another seed change (and we will just see it when that "seed" is introduced), but given the inference time change is across most of the experiments, I'm still suspicious something else is changing. Even once there's essentially no algorithm code changes that will be introduced, then it could still be helpful to have a follow-up PR basically demonstrating that.

tristanls:
@nielsleadholm, if I comment back in the code in `SurfacePolicy.touch_object` right after `TouchObject` initialization, it ~~replicates~~ generates outcomes that are similar to the #339 benchmark differences (`RuntimeError` isn't raised, it's there to check for exit logic mistakes in this hack):

```python
        if self.touch_object_pos_proc is None:
            self.touch_object_pos_proc = TouchObject(
                agent_id=self.agent_id,
                sensor_id=view_sensor_id,
                desired_object_distance=self.desired_object_distance,
                rng=self.rng,
                # TODO: Remaining arguments are unused but required by BasePolicy.
                #       These will be removed when PositioningProcedure is split from
                #       BasePolicy.
                action_sampler_args=dict(actions=[LookUp]),
                action_sampler_class=UniformlyDistributedSampler,
                switch_frequency=0.0,
            )
        result = self.touch_object_pos_proc.positioning_call(raw_observation, state)
        if not result.terminated and not result.truncated:
            action = result.actions[0] # TouchObject returns one action at a time
            # Ensure dynamic_call gives us control by raising ObjectNotVisible
            self.attempting_to_find_object = True
            if isinstance(action, MoveForward): # TouchObject only returns MoveForward when it finishes
                # Reset TouchObject positioning procedure
                self.touch_object_pos_proc = None
                # Let dynamic_call proceed to take action
                self.attempting_to_find_object = False
            return action

        raise RuntimeError("touch_object should not have been called")
```
Oh, and fun fact 😄, tests pass for the above code without any test changes.

Benchmarks diff replications:
[base_config_10distinctobj_surf_agent,100.00,2.14,27,18.04,3,10](https://wandb.ai/thousand-brains-project/Monty/runs/gmwvk5zq/overview)
[randrot_noise_10distinctobj_surf_agent,99.00,4.00,27,29.18,3,18](https://wandb.ai/thousand-brains-project/Monty/runs/3kaqui9v/overview)
[randrot_10distinctobj_surf_agent,100.00,2.00,27,19.11,2,10](https://wandb.ai/thousand-brains-project/Monty/runs/hmz21fck/overview)
[base_10simobj_surf_agent,94.29,10.71,53,14.32,5,18](https://wandb.ai/thousand-brains-project/Monty/runs/icvghikv/overview)
randrot_noise_10simobj_surf_agent - TBD: https://wandb.ai/thousand-brains-project/Monty/runs/p3jds3ge/overview
[randomrot_rawnoise_10distinctobj_surf_agent,68.00,74.00,15,88.35,4,6](https://wandb.ai/thousand-brains-project/Monty/runs/c1bw2vxa/overview)

base_77obj_surf_agent - TBD: https://wandb.ai/thousand-brains-project/Monty/runs/mqw3qcpj/overview

Looks like [randrot_noise_77obj_dist_agent got stuck](https://wandb.ai/thousand-brains-project/Monty/runs/e0cybeg1/overview) (which might be a fluke, as I dont believe it would be affected by any changes).

vkakerbeck:
If I see it right, the benchmark changes here are minor and look like changes cause by random seed change.

Do you still see episodes getting stuck with only this change? This seems like it would be an issue that needs to be addressed. And it does not look like what you observe here where we just have slight variations in the results (explainable by random seed).

It makes sense that the tests are not affected by this. Tests only use very basic geometric objects (they don't have access to the YCB dataset) and are therefor much more simplistic and usually don't encounter the kind of edge cases we see in the benchmark experiments.

nielsleadholm:
Echoing @vkakerbeck , my overall feeling is that there is still something categorically different between the results introduced in this PR, and those from [PR 339](https://github.com/thousandbrainsproject/tbp.monty/pull/339).

I'd be up for discussing this in a call @tristanls today if you want? Will message you on Slack.

In terms of possible next steps:
- Since this PR (including the un-commented section) still does not seem to produce the troublesome behavior, it could be worth continuing to introduce the code changes necessary to replicate 339. That might help narrow down the cause.
- Another suggestion would be whether you can find any [infinite loop episodes](https://github.com/thousandbrainsproject/tbp.monty/pull/339#issuecomment-3006563701) like you show here, but in the `main` branch, i.e. before your PR. If so, then I'm more inclined to agree it's just a seed difference. 

Semi-independent of this issue, I thought it would be helpful at least for me, and maybe others, to just recap how the surface-agent uses `touch_object`, and how we might imagine it working in the longer term. The below identifies some potential issues, which might be contributing to the above problem, but again, I think the current PR results suggest that it is something new:
- For the surface agent's sensor-patch input (`SurfaceViewFinderMountHabitatDatasetArgs`), depth values are clipped in `DepthTo3DLocations` to 5cm (0.05m). 
- `touch_object` is used at the start of an episode, and if we _fall off_ an object during the standard surface-following policy. The latter can rarely happen on the edges of objects, esp. if they have sharp corners.
- `touch_object` is not directly used after an `execute_jump_attempt`. If the object is within 1m, the surface agent tries to move forward. If the object is not visible and we are using the surface agent, it just resets our agent and sensors to the positions they were in before the jump. However, I think jumps *indirectly* call `touch_object` if `0.05 < depth_at_center < 1.0`, as this will pass the first check, but then the sensor-module itself won't be able to see the object. This should trigger the `touch_object` procedure.

Take-aways, which I think are all consistent with your plans Tristan for the on-going refactor:
- We should use the view-finder (with its greater depth reading) at the start of the episode, for the positioning procedure. Similar to `GetGoodView`, it might make sense to have a second refinement stage that basically calls the `touch_object` below to make sure the SM itself starts on the object, in case moving forward resulted in a mismatch between the view-finder and the SM center-pixels.
- We should switch to using the SM (with its clipped 0.05m depth reading) to find the object when we fall off - this is akin to the finger blindly moving around until it touches the object. Because we will have just fallen off the corner, we shouldn't need the extreme depth readings of the `view_finder` to find it again. This should also reduce the chance that the view-finder finds the object, but the SM is not actually on it. If the surface agent falls off and fails to find the object with this change, then this is arguably a failure of the policy. We could look to update it, for example by increasing the "length" of the finger slightly, or having it pivot its own center, rather than about a point in front of it.
- For now, we can continue not explicitly invoking `touch_object` within `execute_jump_attempt`; however, it probably makes sense to change the line 

```python
if depth_at_center < 1.0:
    self.handle_successful_jump()
```
to use the depth reading available to that agent, rather than 1.0, once we move away from invoking view-finder based `touch_object` during an experiment.

tristanls:
Thank you for working through this one with me and identifying the infinite loop that caused the "stuck" problem. I appreciate it. This pull request is superseded by #356.

PR #352, opened 2025-06-26
style: no need for types in docstrings, they are automatic now

tristanls:
Thanks to @jeremyshoemaker pointing out this plugin, we no longer need to, nor will we ask for providing type information in the docstrings. The types for the API docs will now be automatically generated from the Python type information.

FYI @codeallthethingz, @vkakerbeck, @nielsleadholm, @hlee9212, @scottcanoe, @ramyamounir


src/tbp/monty/frameworks/utils/spatial_arithmetics.py
```diff
@@ -135,8 +135,8 @@ def get_angles_for_all_hypotheses(hyp_f, query_f):
     return shape = (num_hyp, num_nn)
 
     Args:
-        hyp_f (num_hyp, num_nn, 3): ?
-        query_f (num_hyp, 3): ?
+        hyp_f: ?
+        query_f: ?
```
jeremyshoemaker: This entire docstring is _almost_ useless. Not sure who wrote it, but I think we should try to fill in the blanks.

src/tbp/monty/simulators/simulator.py
```diff
@@ -49,17 +49,17 @@ def add_object(
...
-            semantic_id (Optional[str]): Optional override object semantic ID
-            enable_physics (bool): Whether to enable physics on the object
-            object_to_avoid (bool): If True, ensure the object is not colliding with
-              other objects
-            primary_target_bb (List | None): If not None, this is a list of the min and
+            name: Registered object name.
+            position: Initial absolute position of the object.
+            rotation: Initial orientation of the object.
+            scale: Initial object scale.
+            semantic_id: Optional override object semantic ID.
```
jeremyshoemaker: I know I just wrote this, and I might have just copied it from `HabitatSim`, but this just reads poorly. I think it should be something like this instead:
```suggestion
            semantic_id: Optional override for the object's semantic ID.
```

jeremyshoemaker (approved):
Looks good overall. I didn't check every instance, but I noticed there are a few where we're losing typing information, because the type in the docstring wasn't repeated in the type hints on the function definitions. We might want to preserve those. I could also see just taking them out and revisiting this when we start trying to type things in earnest.

tristanls:
Thanks @jeremyshoemaker, I made another pass and readded the type information to where I previously lost it. Additionally, did another pass over all `Returns:` docstrings and cleaned those up a bit.

PR #357, opened 2025-06-27
chore: add correct percentage figure to plot tool

ramyamounir:
This PR adds a simple script to the plot tool under the name `correct_percentage_per_episode`. It will plot a bar chart with the percentage of correct steps on the y-axis and the episode target on the x-axis. The idea was suggested by @nielsleadholm and it would be a nice metric to report as part of the unsupervised inference experiments.

```zsh
usage: cli.py correct_percentage_per_episode [-h] [--debug] [-lm LEARNING_MODULE] [-t {primary,stepwise}] experiment_log_dir

positional arguments:
  experiment_log_dir    The directory containing the experiment log with the detailed stats file.

optional arguments:
  -lm LEARNING_MODULE, --learning_module LEARNING_MODULE
                        The name of the learning module (default: "LM_0").
  -t {primary,stepwise}, --target_object {primary,stepwise}
                        Whether to use "primary_target_object" or "stepwise_target_object" for labels (default: "primary").
```

For example with the No Reset results below:
max_scores

The plot tool would output this:
Summary

### DIFF
--- a/tools/plot/correct_percentage_per_episode.py
+++ b/tools/plot/correct_percentage_per_episode.py
@@ -25,12 +25,16 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-def plot_correct_percentage_per_episode(exp_path: str, learning_module: str) -> int:
+def plot_correct_percentage_per_episode(
+    exp_path: str, learning_module: str, target_object_key: str
+) -> int:
     """Bar chart showing how many steps the correct object had the highest evidence.
 
     Args:
         exp_path: Path to the experiment directory containing the detailed stats file.
         learning_module: The learning module to use for extracting evidence data.
+        target_object_key: The key for extracting labels from stats file.
+            choices are "primary" or "stepwise".
 
     Returns:
         Exit code.
@@ -47,7 +51,7 @@ def plot_correct_percentage_per_episode(exp_path: str, learning_module: str) -> 
 
     for _, episode_data in enumerate(detailed_stats.values()):
         evidences_data = episode_data[learning_module]["max_evidence"]
-        target_obj = episode_data["target"]["primary_target_object"]
+        target_obj = episode_data["target"][target_object_key + "_target_object"]
 
         count = sum(1 for ts in evidences_data if max(ts, key=ts.get) == target_obj)
         percentage = (count / len(evidences_data)) * 100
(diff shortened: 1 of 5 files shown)


ramyamounir:
FYI, @vkakerbeck and @tristanls

tools/plot/correct_percentage_per_episode.py
```diff
@@ -0,0 +1,129 @@
...
+
+    fig.tight_layout()
+    plt.show()
+
+    return 0
+
+
+def add_subparser(
+    subparsers: argparse._SubParsersAction,
+    parent_parser: Optional[argparse.ArgumentParser] = None,
```
tristanls: **suggestion (non-blocking)**: Prefer `argparse.ArgumentParser | None` over `Optional[argparse.ArgumentParser]` (I think it'd require `from __future__ import annotations` at the top of the file.) 

The reason is that we'll eventually turn on Ruff's `UP045` rule: https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/

ramyamounir: Nice, fixed. Ruff complained about moving argparse into a type checking block, I applied the code action to make it happy but i'm not sure if this makes sense.

pic-window-250627-1819-44

tristanls: If a concrete class is only used for typing, then that's the usual recommendation. You can search codebase for TYPE_CHECKING and see bunch of examples. Basically, it's only imported when mypy runs and never during runtime.

tools/plot/correct_percentage_per_episode.py
```diff
@@ -0,0 +1,133 @@
...
+        return 1
+
+    # Load detailed stats
+    _, _, detailed_stats, _ = load_stats(exp_path, False, False, True, False)
+
+    correct_object_hits = []
+    episode_labels = []
+
+    for _, episode_data in enumerate(detailed_stats.values()):
+        evidences_data = episode_data["LM_0"]["max_evidence"]
```
nielsleadholm: **suggestion:** Consider making `LM_0` an optional argument for this function. This might be used relatively soon when we begin adding hierarchy for compositional objects.

tools/plot/correct_percentage_per_episode.py
```diff
@@ -0,0 +1,133 @@
...
+        target_obj = episode_data["target"]["primary_target_object"]
```
nielsleadholm: **suggestion:** Similarly, it might be worth trying to use `stepwise_target_object` (although you would need to index it), or again, making this an argument in the function. With compositional or multi-object scenes where we don't reset a `primary_target_object`, this is likely to be more reflective of what we're interested in.

nielsleadholm (commented):
Nice, thanks for adding this! Just a couple minor suggestions.

At a high level, you could consider having a summary statistic for the entire experiment, as in, what percentage of the entire experiment did we have the correct MLH. This could be useful for tables and tracking improvement gains.

I think this will be helpful when we try to improve performance (e.g. by resetting hypotheses triggered when moving out of reference frame, or adding voting), ensuring we consider both accuracy at some high-confidence threshold, as well this metric.

ramyamounir:
Thanks Niels, I've added the summary bar (see updated figure in the description) and the suggested optional arguments.

nielsleadholm (approved):
Look great, thanks Ramy!

PR #359, opened 2025-06-30
feat: add RFC for unsupervised object ID association in cross-modal learning

OgnjenX:
RFC proposal for this feature [https://thousandbrainsproject.readme.io/docs/generalize-voting-to-associative-connections](url)

### DIFF
--- a/rfcs/0000_unsupervised_object_id_association.md
+++ b/rfcs/0000_unsupervised_object_id_association.md
@@ -1,0 +1,100 @@
+# RFC 0011: Unsupervised Object ID Association for Cross-Modal Learning
+
+## Summary
+
+This RFC proposes implementing unsupervised object ID association mechanisms that enable learning modules to discover correspondences between their internal object representations without requiring predefined object labels. This addresses the current limitation where voting requires hardcoded object IDs shared across learning modules.
+
+## Motivation
+
+Currently, Monty's voting mechanism relies on all learning modules sharing the same object ID for any given object (e.g., "cup"), which acts as a supervised learning signal. In truly unsupervised scenarios, each learning module creates unique object IDs (e.g., `visual_object_1`, `touch_object_1`) for the same physical object, preventing effective cross-modal voting and consensus building.
+
+This limitation contradicts the goal of unsupervised learning and prevents Monty from working in real-world scenarios where object boundaries and identities are not predefined.
+
+## Detailed Design
+
+### Phase 1: Association Memory and Co-occurrence Learning
+
+1. **Association Memory Structure**
+   - Each LM maintains `association_memory`: `{my_object_id: {other_lm_id: {other_object_id: association_data}}}`
+   - `association_data` includes: co-occurrence count, spatial consistency scores, temporal context
+
+2. **Co-occurrence Detection**
+   - Track when multiple LMs have high evidence for their respective objects simultaneously
+   - Record spatial and temporal context of co-occurrences
+   - Build confidence scores for object ID associations
+
+### Phase 2: Modified Voting Protocol
+
+1. **Enhanced Vote Messages**
+   - Include sender LM ID, confidence scores, spatial context
+   - Send association metadata alongside traditional vote data
+
+2. **Probabilistic Vote Mapping**
+   - Map incoming votes to local object IDs using learned associations
+   - Weight votes by association confidence
+   - Handle uncertainty in associations gracefully
+
+### Phase 3: Consensus Building
+
+1. **Multi-Modal Hypothesis Clustering**
+   - Group hypotheses from different LMs based on spatial/temporal consistency
+   - Use clustering to identify likely same-object hypotheses
+
+2. **Dynamic Association Updates**
+   - Continuously refine associations as more evidence is gathered
+   - Implement feedback mechanisms for association validation
+
+## Implementation Plan
+
+### Step 1: Create Association Mixin
+- `UnsupervisedAssociationMixin` class
+- Association memory management
+- Co-occurrence tracking algorithms
+
+### Step 2: Extend Learning Module Classes
+- Modify `EvidenceGraphLM` to include association capabilities
+- Update voting methods (`send_out_vote`, `receive_votes`)
+- Add association learning logic
+
+### Step 3: Enhanced Voting Protocol
+- Modify vote data structures
+- Implement probabilistic vote mapping
+- Add association confidence calculations
+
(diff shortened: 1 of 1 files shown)


rfcs/0000_unsupervised_object_id_association.md
```diff
@@ -0,0 +1,103 @@
...
+
+2. **Co-occurrence Detection**
+   - Track when multiple LMs have high evidence for their respective objects simultaneously
+   - Record spatial and temporal context of co-occurrences
+   - Build confidence scores for object ID associations
+
+### Phase 2: Modified Voting Protocol
+
+1. **Enhanced Vote Messages**
+   - Include sender LM ID, confidence scores, spatial context
```
nielsleadholm: **question:** Can you clarify what you mean by spatial context here? We already share certain spatial details during voting, but the inclusion of this under "Enhanced Vote Messages" suggests that it is a different type of spatial information you are introducing.

rfcs/0000_unsupervised_object_id_association.md
```diff
@@ -0,0 +1,103 @@
...
+   - Send association metadata alongside traditional vote data
+
+2. **Probabilistic Vote Mapping**
+   - Map incoming votes to local object IDs using learned associations
+   - Weight votes by association confidence
+   - Handle uncertainty in associations gracefully
```
nielsleadholm: **question:** Can you please clarify what this means exactly? What kind of uncertainty are you anticipating?

rfcs/0000_unsupervised_object_id_association.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Phase 3: Consensus Building
+
+1. **Multi-Modal Hypothesis Clustering**
+   - Group hypotheses from different LMs based on spatial/temporal consistency
+   - Use clustering to identify likely same-object hypotheses
```
nielsleadholm: **question:** Can you please clarify what you mean by clustering here? This is not a term we've used before, so I'm wondering if the proposal to change how votes are processed by LMs?

rfcs/0000_unsupervised_object_id_association.md
```diff
@@ -0,0 +1,103 @@
...
+   - Record spatial and temporal context of co-occurrences
+   - Build confidence scores for object ID associations
+
+### Phase 2: Modified Voting Protocol
+
+1. **Enhanced Vote Messages**
+   - Include sender LM ID, confidence scores, spatial context
+   - Send association metadata alongside traditional vote data
+
+2. **Probabilistic Vote Mapping**
```
nielsleadholm: **question:** Can you please clarify what you mean by "probabilistic" in this setting? Monty does not currently have any probabilistic formulations in its algorithms.

nielsleadholm (commented):
Thanks again @OgnjenX for putting this together. The majority of my high-level comments are in PR https://github.com/thousandbrainsproject/tbp.monty/pull/358 , but I've added a few comments here for points that were only in the RFC.

rfcs/0000_unsupervised_object_id_association.md
OgnjenX: The “uncertainty” refers to noisy or conflicting association signals—e.g., multiple LMs proposing different object IDs with similar confidence, or spurious co-occurrences when objects briefly overlap. In the implementation (UnsupervisedAssociator in src/tbp/monty/frameworks/models/unsupervised_association.py), we address this by maintaining a decayed confidence history (AssociationData.update_confidence() / get_average_confidence()), computing spatial and temporal consistency scores, and combining all of these into a weighted strength via get_association_strength(). An association is only used once its strength clears min_association_threshold, and it keeps decaying if new evidence doesn’t arrive. Instead of locking onto the first co-occurrence, we continuously re-score the link and only forward votes when the accumulated evidence is sufficiently strong.

rfcs/0000_unsupervised_object_id_association.md
OgnjenX: The term "clustering" in the RFC was a bit of an overstatement—it refers to how we group temporally related associations in the 
_calculate_temporal_clustering()
 function. Instead of traditional clustering, we analyze how densely associations occur in time. For example, if multiple LMs report the same object ID in quick succession, we consider that a stronger signal than sporadic reports. The implementation in 
UnsupervisedAssociator
 uses a simple density metric (number of associations per time unit) to score this temporal grouping, which is then factored into the overall association strength. Would you prefer we update the RFC to use "temporal grouping" instead of "clustering" to avoid confusion?

OgnjenX:
> Thanks again @OgnjenX for putting this together. The majority of my high-level comments are in PR #358 , but I've added a few comments here for points that were only in the RFC.

I made changes in [this](https://github.com/thousandbrainsproject/tbp.monty/pull/359/commits/91558d1196542061c37e03edd5b51771e9b3b50e) commit that tries to improve the RFC for better clarity, related to the questions you had.

rfcs/0000_unsupervised_object_id_association.md
nielsleadholm: Thanks, that makes sense. Yes I think temporal consistency is already clear enough, in which case I don't think you need this line.

rfcs/0000_unsupervised_object_id_association.md
nielsleadholm: Thanks for clarifying.

rfcs/0000_unsupervised_object_id_association.md
```diff
@@ -96,7 +96,7 @@ This limitation contradicts the goal of unsupervised learning and prevents Monty
 1. **Hierarchical Associations**: Part-whole relationships across modalities
 2. **Temporal Sequence Learning**: Dynamic object and scene understanding
 3. **Language Grounding**: Associate learned words with grounded objects
-4. **Advanced Clustering**: More sophisticated hypothesis grouping algorithms
+4. **Richer Hypothesis Grouping**: Explore graph-based or probabilistic grouping atop the learned association strengths once the foundational pipeline is validated
```
nielsleadholm, line 99: **question:** Can you please provide a sentence or two what you mean by graph-based or probabilistic grouping here?

OgnjenX: The idea is that if object A and B appear at the same time, and B and C appear at the same time, maybe it makes sense to keep (small) possibility that A and C belong to the same 'scene' even though they never appeared together.

PR #361, opened 2025-07-03
feat: MuJoCo sim, basic add and remove

jeremyshoemaker:
This starts to implement the MuJoCo simulator implementing some basic methods from the Simulator protocol.

Specifically, adding primitive objects, and removing all objects from the scene, as well as the basic functionality for managing MuJoCo's state.


jeremyshoemaker:
I added the `unittest-parametrize` dependency so that I could write one of the unit tests with parametrized inputs rather than copy pasting the same test 5 times. I kinda expect to run into this sort of thing again, so I think adding it makes sense. Pytest has builtin functionality for this but it doesn't work with `unittest` style tests.

I'm not actually using any `unittest` specific features here -- all of the methods could be static and hence separate standalone functions -- so they could just be plain Pytest style tests, but I don't know if we want to keep our tests in a specific style or start to move towards Pytest in general.

vkakerbeck (commented):
I've never used Mujoco or worked very closely with our habitat simulator code but it all looks good to me. I'm just not ticking approve since I think Tristan will be the better person to review and approve.

jeremyshoemaker:
> I've never used Mujoco or worked very closely with our habitat simulator code but it all looks good to me. I'm just not ticking approve since I think Tristan will be the better person to review and approve.

That's fair. I wasn't sure if you'd have additional context, which was why I added you as a reviewer, but didn't assign it to you.

I don't think there's anything here that's straying too far from what the Habitat version is doing (aside from a smaller set of primitive objects available), but if there's anyone else on the team who might have more context, it might be good to get their feedback. I just don't know who that might be.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -0,0 +1,207 @@
...
+        # Note: We can't use the `model.ngeoms` for this since that will include parts
+        # of the agents, especially when we start to add more structure to them.
+        self._object_count = 0
+
+    def _recompile(self) -> None:
+        """Recompile the MuJoCo model while retaining any state data."""
+        self.model, self.data = self.spec.recompile(self.model, self.data)
+
+    def initialize_agent(self, agent_id, agent_state) -> None:
+        """Update agent runtime state."""
```
tristanls: **note**: I think this is extraneous, since the protocol has this docstring.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -0,0 +1,207 @@
...
+            rotation: Initial orientation of the object
+            scale: Initial object scale
+            semantic_id: Optional override object semantic ID
+            enable_physics: Whether to enable physics on the object
+            object_to_avoid: If True, ensure the object is not colliding with
+              other objects
+            primary_target_bb: If not None, this is a list of the min and
+              max corners of a bounding box for the primary object, used to prevent
+              obscuring the primary objet with the new object.
+        """
```
tristanls: **note**: I think this is extraneous since the Protocol docstring documents this method.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Arguments:
+            obj_name: Identifier for the object in the scene, must be unique.
+            shape_type: The primitive shape to add.
+            position: Initial position of the object
+            rotation: Initial orientation of the object
+            scale: Initial scale of the object
+
+        Raises:
+            ValueError: when the shape_type is unknown
```
tristanls: **suggestion**: `UnknownShapeType`.

vkakerbeck:
Most of the simulator code was written by Luiz at Numenta. I think @nielsleadholm added the `object_to_avoid` and `primary_target_bb` part but otherwise this simulator code has remained relatively untouched since the initial setup.

nielsleadholm:
Just to respond that yeah I wouldn't say I have a great understanding of the original implementation, other than as Viviane said, the functions I implemented. That said if you would like me to cast a more thorough look over this, or jump on a Huddle to talk about what I do understand, I'd be happy to help!

I think the basic requirements aren't going to be anything too controversial, being able to add and remove objects etc. I think the main thing that might be worth thinking about is the semantic sensor, which has been useful in the past for evaluations, and might continue to be so as we move towards more complex environments.

jeremyshoemaker:
> Just to respond that yeah I wouldn't say I have a great understanding of the original implementation, other than as Viviane said, the functions I implemented. That said if you would like me to cast a more thorough look over this, or jump on a Huddle to talk about what I do understand, I'd be happy to help!
> 
> I think the basic requirements aren't going to be anything too controversial, being able to add and remove objects etc. I think the main thing that might be worth thinking about is the semantic sensor, which has been useful in the past for evaluations, and might continue to be so as we move towards more complex environments.

Okay, that context helps. I might have some questions about the things Viviane mentioned, but I'll get back to you when I get there.

This initial PR isn't doing anything around sensors or agents yet, it's just the very basics of add an object (just primitives at the moment), reset the scene, and some tests to confirm that all that works as expected. I'm tackling the agents and sensors in separate PRs, to keep the changes bite sized and easy to understand.

On the topic of the semantic sensor, I'm pretty sure MuJoCo will let us do something similar with its segmentation rendering. It sounds like just a different name for the same thing, where each pixel corresponds to the ID of the object that the simulation knows it at that location.

nielsleadholm:
Ok perfect, sounds good.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def add_object(
+        self,
+        name: str,
+        position: VectorXYZ = (0.0, 0.0, 0.0),
+        rotation: QuaternionWXYZ = (1.0, 0.0, 0.0, 0.0),
+        scale: VectorXYZ = (1.0, 1.0, 1.0),
+        semantic_id: Optional[str] = None,
+        enable_physics=False,
+        object_to_avoid=False,
```
tristanls: **suggestion (non-blocking)**: Missing type annotation. They're needed despite default params.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -0,0 +1,207 @@
...
+        # Track how many objects we add to the environment.
+        # Note: We can't use the `model.ngeoms` for this since that will include parts
+        # of the agents, especially when we start to add more structure to them.
+        self._object_count = 0
+
+    def _recompile(self) -> None:
+        """Recompile the MuJoCo model while retaining any state data."""
+        self.model, self.data = self.spec.recompile(self.model, self.data)
+
+    def initialize_agent(self, agent_id, agent_state) -> None:
```
tristanls, line 50: **suggestion (non-blocking)**: Missing type annotation for function arguments.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -0,0 +1,207 @@
...
+    def add_object(
+        self,
+        name: str,
+        position: VectorXYZ = (0.0, 0.0, 0.0),
+        rotation: QuaternionWXYZ = (1.0, 0.0, 0.0, 0.0),
+        scale: VectorXYZ = (1.0, 1.0, 1.0),
+        semantic_id: Optional[str] = None,
+        enable_physics=False,
+        object_to_avoid=False,
+        primary_target_bb: Optional[List] = None,
```
tristanls, line 68: **suggestion (non-blocking)**: Missing type parameters for generic `List`.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -0,0 +1,207 @@
...
+        """Return agent instance."""
+        pass
+
+    def get_observations(self):
+        """Get sensor observations."""
+        pass
+
+    def get_states(self):
+        """Get agent and sensor states."""
+        pass
```
tristanls: **suggestion (non-blocking)**: Missing return type annotations.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -0,0 +1,207 @@
...
+    def apply_action(self, action: Action) -> Dict[str, Dict]:
+        """Execute the given action in the environment.
+
+        Args:
+            action (Action): the action to execute
+
+        Returns:
+            (Dict[str, Dict]): A dictionary with the observations grouped by agent_id
+        """
+        pass
```
tristanls: **suggestion (non-blocking)**: Should probably return `{}` to match the type.

tests/unit/simulators/mujoco/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+
+import pytest
+
+# Adding this at the test `mujoco` module level means we don't have to remember to
+# add it for every module beneath here.
+pytest.importorskip(
+    "mujoco",
+    reason="MuJoCo test suite skipped. MuJoCo optional dependency not installed.",
```
tristanls: **suggestion (non-blocking)**: The Habitat Sim skip has the reason "Habitat Sim optional dependency not installed." For consistency, I think this should be "MuJoCo optional dependency not installed." The being skipped part will be obvious from context.

tests/unit/simulators/mujoco/mujoco_simulator_test.py
```diff
@@ -0,0 +1,93 @@
...
+    @parametrize(
+        "shape",
+        [
+            param("box", id="box"),
+            param("capsule", id="capsule"),
+            param("cylinder", id="cylinder"),
+            param("ellipsoid", id="ellipsoid"),
+            param("sphere", id="sphere"),
+        ],
+    )
```
tristanls: **suggestion (non-blocking)**: I think we briefly discussed that this could become something like:

```python
SHAPES = ["box", "capsule", "cylinder", "ellipsoid", "sphere"]
SHAPE_PARAMS = [param(s, id=s) for s in SHAPES]
```

tests/unit/simulators/mujoco/mujoco_simulator_test.py
tristanls, line 1: **suggestion (non-blocking)**: I'm noticing that `sim.model.ngeom`, `sim.get_num_objects()`, `len(sim.spec.geoms)` must all be in sync. This prompts two suggestions:

1) A test assertion helper? 
```python
def assert_object_count(sim: MuJoCoSimulator, count: int) -> None:
    assert sim.model.ngeom == count
    assert sim.get_num_objects() == count
    assert len(sim.spec.geoms) == count
```

2) This seems like an invariant to uphold in `MuJoCoSimulator`, perhaps the sim should have an assert helper that is called after all operations impacting object count, such as:
```python
class MuJoCoSimulator(Simulator):
    # ...
    def _assert_object_count(self) -> None:
        assert sim.model.ngeom == sim._object_count == len(sim.spec.geoms)
```

tests/unit/simulators/mujoco/mujoco_simulator_test.py
```diff
@@ -0,0 +1,93 @@
...
+        assert sim.model.ngeom == 1
+        assert sim.get_num_objects() == 1
+        assert len(sim.spec.geoms) == 1
+
+        # 1. Check that the spec was updated
+        spec_xml = sim.spec.to_xml()
+        # Sphere is the default and so its type doesn't end up in the resulting XML
+        if shape != "sphere":
+            assert f'type="{shape}"' in spec_xml
+        assert f'name="{shape}_0"' in spec_xml
```
tristanls: **issue**: Searching for a string in serialized data can fail or succeed unpredictably. You should parse the XML and ensure everything (or what you're looking for) is where it's supposed to be in the XML structure.

tests/unit/simulators/mujoco/mujoco_simulator_test.py
```diff
@@ -0,0 +1,93 @@
...
+        # 2. Check that the model was updated
+        # This raises if it doesn't exist
+        sim.model.geom(f"{shape}_0")
+
+    def test_multiple_objects_have_different_ids(self) -> None:
+        """Test that multiple objects have different IDs.
+
+        To prevent name collisions in the MuJoCo spec, the names of objects are
+        suffixed with their "object number", an increasing index of the objects in the
+        scene. So, several objects should be numbered in the order they were added.
```
tristanls, line 57: **note**: I'm not entirely sure that this collision avoidance should be done inside `add_object` (instead of, for example, the caller ensuring it doesn't collide). However, I don't have enough simulator context at this time to challenge this approach.

tests/unit/simulators/mujoco/mujoco_simulator_test.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        shapes = ["box", "capsule", "cylinder"]
+        sim = MuJoCoSimulator()
+        for shape in shapes:
+            sim.add_object(shape)
+
+        assert sim.model.ngeom == len(shapes)
+        assert sim.get_num_objects() == len(shapes)
```
tristanls: **issue**: Are you missing `assert len(sim.spec.geoms) == len(shapes)` ?

tests/unit/simulators/mujoco/mujoco_simulator_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+        spec_xml = sim.spec.to_xml()
+        for i, shape in enumerate(shapes):
+            assert f'name="{shape}_{i}"' in spec_xml
```
tristanls: **issue**: As mentioned elsewhere, this should parse `spec_xml` instead of looking in a string.

tests/unit/simulators/mujoco/mujoco_simulator_test.py
```diff
@@ -0,0 +1,93 @@
...
+        assert sim.model.ngeom == 0
+        assert sim.get_num_objects() == 0
+        assert len(sim.spec.geoms) == 0
+
+    def test_primitive_object_positioning(self) -> None:
+        sim = MuJoCoSimulator()
+        sim.add_object("box", position=(1.0, 1.0, 2.0))
+
+        assert np.allclose(sim.model.geom("box_0").pos, np.array([1.0, 1.0, 2.0]))
+        assert f'pos="1 1 2"' in sim.spec.to_xml()
```
tristanls: **issue**: As mentioned elsewhere, this should parse `spec_xml` instead of looking in a string.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -0,0 +1,207 @@
...
+        world_body: MjsBody = self.spec.worldbody
+
+        # TODO: should we encapsulate primitive objects into bodies?
+
+        if shape_type == "sphere":
+            world_body.add_geom(
+                name=obj_name,
+                type=mjtGeom.mjGEOM_SPHERE,
+                # Use first scale component as radius
+                size=[scale[0] * 0.5, 0.0, 0.0],
```
tristanls: **question**: Why are all sizes being modified here and the other `if` branch arms instead of using the `scale` directly? This strikes me as an arbitrary, hardcoded convention for someone to reverse engineer before using `add_object`.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I'm going to leave this as-is for now. I'm planning to make `agent_id` a newtype of `str` when I add agents. As for `agent_state`, it seems like an internal HabitatSim concept, so it might not actually map to something in this simulator once I implement it, but either way, that type would most likely be defined when adding the agents.

tests/unit/simulators/mujoco/mujoco_simulator_test.py
jeremyshoemaker: The caller doesn't have any way to specify it, without requiring the use of `semantic_id`, which we currently don't do in the Habitat version (it's optional). 

The `name` field is basically the lookup name for the "template" of the object being added. In the current version, that's the name of the primitive object, but in the instance of having other custom object types registered, it would be the key for looking those up. So it can't be used as the ID without adding some uniqueness to it. 

We _could_ just use an internal ID scheme for the ID and exclude the name entirely, but I felt for readability of state, keeping the object type in the ID would be helpful.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I was going off the [documentation for adding primitives](https://mujoco.readthedocs.io/en/stable/XMLreference.html#body-geom), specifically the section on the `size` parameter. 

I'm not sure what will happen in the case of a sphere if you give it more than just the single parameter, I'll have to test it.

For some of the others, there's a scaling factor where in Habitat the size defines a bounding box, and in MuJoCo the size could be the half-length of a side of a cube, or the radius instead of the diameter of a circular object.

The current implementation allows, for the most part, using the same scale parameters as used in Habitat. That doesn't mean it's the best approach.

tests/unit/simulators/mujoco/mujoco_simulator_test.py
jeremyshoemaker: Parsing XML in a test seems a bit heavy when a simple substring search works fine. If this were testing anything more complicated than an attribute, I'd definitely want to use a proper parser, but I don't think the extra work is warranted here.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: Based on the Habitat version and the docstring I copied into Simulator, this _should_ be a list of bounding boxes, but I haven't defined a type for those yet.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: It looks like it works whether you pass in the extra values or not (because the shape of `size` still has to have three elements). That allows simplifying this code a bit.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -112,96 +96,58 @@ def _add_primitive_object(
...
         else:
-            raise ValueError(f"Unknown MuJoCo primitive: {shape_type}")
+            raise UnknownShapeType(f"Unknown MuJoCo primitive: {shape_type}")
+
+        world_body.add_geom(
+            name=obj_name,
+            type=geom_type,
+            # MuJoCo sizes are generally half-length or radii, so half of the
+            # sizes of the bounding box.
+            size=np.array(scale) * 0.5,
```
tristanls: **question**: What anomaly do you observe if `scale` is _not_ multiplied by `0.5`?

I'm still trying to understand why it is valid to multiply everything by `0.5`. A Monty experiment will pass in the `scale` param, which is sent to `HabitatSim`. `HabitatSim` will adjust things as needed internally. [For example](https://github.com/facebookresearch/habitat-sim/blob/6c26c4f9ae10ad7f534aa9e59a84dff00782741e/src/esp/physics/bullet/BulletPhysicsManager.cpp#L333)
```cpp
        case metadata::URDF::GEOM_BOX:
        visualMeshInfo.type = AssetType::Primitive;
        visualMeshInfo.filepath = "cubeSolid";
        scale = visual.m_geometry.m_boxSize * 0.5;
        break;
```
[where](https://github.com/facebookresearch/habitat-sim/blob/6c26c4f9ae10ad7f534aa9e59a84dff00782741e/src/esp/metadata/URDFParser.h#L96)
```cpp
  Magnum::Vector3 m_boxSize{1, 1, 1};
```

tests/unit/simulators/mujoco/mujoco_simulator_test.py
tristanls: As discussed in the description of #218, this is a type of test where we want to validate that `sim.add_object(shape)` caused a desired side effect. The desired side effect is not the presence of a string anywhere in the spec XML document, but, I assume, that the object was added appropriately to the spec, which would require parsing XML. In #218, I highlighted why I used the JSON parser, since multiple strings parse to the same JSON. Multiple strings can parse into the same XML, and this test hardcodes only one of those string representations.

**issue**: I now notice that the test's name doesn't match its function. For `test_multiple_objects_have_different_ids`, I would expect `shapes` to be `["box", "box", "box"]` to check if unique IDs are assigned.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: Isn't that the same thing I'm doing here? I'm making it so that a caller of `add_object` when passing in a `scale` of [1, 1, 1] will end up with a box that is 1 by 1 by 1 instead of 2 by 2 by 2.

Am I missing something here?

tests/unit/simulators/mujoco/mujoco_simulator_test.py
jeremyshoemaker: Both good ideas, I'll add them.

tests/unit/simulators/mujoco/mujoco_simulator_test.py
jeremyshoemaker: Good point on the `shapes`, I'll change that. I think I was thinking of something else when I wrote this.

I'll look into doing XML parsing, but my thinking here is that I'm assuming that `MjSpec` works correctly (that there are tests in MuJoCo that prove that) and that as a result the XML output can be relied upon to be stable (outside of version changes). What this test is testing is that when we call _our_ methods that add geoms to the spec, the spec is updated correctly. I feel like this test currently does that, but I can make it more rigid (not a bad thing) by parsing the XML and making sure it ends up where we expect it to.

src/tbp/monty/simulators/mujoco/simulator.py
tristanls: From what I can tell, the difference is that we don't do it in our code for HabitatSim. But here, we do it in our code for MuJoCo.

Regarding the observed anomaly, are you saying that you observed passing (1,1,1) to HabitatSim results in a box size of (1,1,1), but when you pass (1,1,1) to MuJoCo, you observe the box to be (2,2,2)?

tests/unit/simulators/mujoco/mujoco_simulator_test.py
jeremyshoemaker: Actually, looking at the code again, I think there _are_ ways in which the counts would differ. `_object_count` is keeping track of all the objects we add to the scene. That doesn't include any geoms that are part of an agent (since I think those shouldn't be counted as objects in the scene and I'm kinda following what was done in HabitatSim for this). I don't expect there to be any geoms in an agent at this stage, but in the future that might not be the case.

In these tests, those values should be the same because we aren't adding any agents to the scene. I could definitely make an assertion helper for the tests.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I haven't observed it. I'm going off how the MuJoCo documentation describes what the `size` parameter should be.

Looking at our [`HabitatSim`](https://github.com/thousandbrainsproject/tbp.monty/blob/7e57ee9dff069e44716e66fbc449c279b2f50927/src/tbp/monty/simulators/habitat/simulator.py#L253) code, I'm not exactly sure how the `scale` value is being propagated to habitat. It seems to be going through the template manager somehow. I've spent most of the day trying to figure out how I could write a test for habitat to determine what it does in these cases.

I went ahead and changed the MuJoCo simulator code to just pass the scale in directly and we can figure out if it's a problem later.

tristanls (approved):
Thank you for all the updates.

PR #362, opened 2025-07-06
Extract better features RFC

OgnjenX:
RFC for initial improvement related to the problem explained [here](https://thousandbrainsproject.readme.io/docs/extract-better-features).

### DIFF
--- a/rfcs/0000_enhanced-non-morphological-features.md
+++ b/rfcs/0000_enhanced-non-morphological-features.md
@@ -5,57 +5,193 @@
 
 ## Summary
 
-This RFC proposes enhancing non-morphological feature extraction in Monty's sensor modules by implementing multi-scale Local Binary Patterns (LBP) for texture analysis and enhanced depth processing for tactile-like features. The implementation uses a **pure composition approach** with a dedicated texture processor that can be integrated into any sensor module, improving object recognition and discrimination capabilities while maintaining rotation invariance as specified in the project documentation. This approach avoids multiple inheritance complexity and aligns with established architectural preferences in the codebase.
+This RFC proposes enhancing non-morphological feature extraction in Monty's sensor modules by implementing multi-scale
+Local Binary Patterns (LBP) for texture analysis and enhanced depth processing for tactile-like features. The
+implementation uses a **pure composition approach** with a dedicated texture processor that can be integrated into any
+sensor module, improving object recognition and discrimination capabilities while maintaining rotation invariance as
+specified in the project documentation. This approach avoids multiple inheritance complexity and aligns with established
+architectural preferences in the codebase.
 
 ## Motivation
 
-### Current Limitations
+Monty's current non-morphological feature extraction is limited to single-pixel sampling, extracting only RGBA/HSV
+values from the center pixel of a 64x64 patch (using just 1/4096 available pixels). This approach cannot distinguish
+between smooth and textured surfaces with identical colors, lacks rotation invariance, and provides only basic depth
+statistics without surface texture information.
 
-The current non-morphological feature extraction in sensor modules is limited to:
-- **Single-pixel sampling**: Only extracts RGBA/HSV from the center pixel (1/4096 pixels used)
-- **No texture information**: Cannot distinguish between smooth vs. textured surfaces with identical colors
-- **Basic depth statistics**: Only min/mean depth, missing surface texture information
-- **No rotation invariance**: Current features change with sensor orientation
-- **Limited cross-modal integration**: RGB and depth features are processed independently
+The [Extract Better Features](https://thousandbrainsproject.readme.io/docs/extract-better-features) documentation
+specifically calls for richer features using Local Binary Patterns for visual features and enhanced depth processing to
+approximate tactile texture, with an explicit requirement for **rotation invariant** features that detect textured
+patterns regardless of sensor orientation.
 
-### Problem Statement
+The main research question is how much improvement we can achieve with enhanced texture features, and how robust these
+features are to noise, scale variations, and different viewing conditions. Secondary questions include optimal
+integration of RGB and depth texture features, acceptable computational overhead, and correlation with existing
+morphological features.
 
-From the [Extract Better Features](https://thousandbrainsproject.readme.io/docs/extract-better-features) documentation:
-> "Currently non-morphological features are very simple, such as extracting the RGB or hue value at the center of the sensor patch. In the short term, we would like to extract richer features, such as using HTM's spatial-pooler or Local Binary Patterns for visual features, or processing depth information within a patch to approximate tactile texture."
-
-The documentation specifically requires **rotation invariant** features that can detect textured patterns regardless of sensor orientation.
-
-### Research Questions
-
-1. **Feature Complementarity**: Should LBP and HTM Spatial Pooler be used together for enhanced texture discrimination?
-2. **Cross-Modal Integration**: How can RGB and depth texture features be optimally combined?
-3. **Computational Trade-offs**: What is the acceptable computational overhead for improved recognition accuracy?
-4. **Feature Interaction**: How do new texture features correlate with existing morphological features?
 
 ## Detailed Design
 
 ### Architecture Overview
 
-The implementation follows a **pure composition approach** that can be integrated into any sensor module:
+To enable sensor modules to extract enhanced texture features, we propose a compositional approach where each sensor
+module owns feature-specific data processors that can be configured and utilized as needed. These processors serve as
+simple interfaces to standalone functions that don't require sensor module attributes or methods, allowing us to avoid
+complex inheritance patterns.
 
-1. **Texture Processing Utilities**: Standalone functions in `src/tbp/monty/frameworks/utils/texture_processing.py`
-2. **Texture Feature Processor**: Dedicated `TextureFeatureProcessor` class that handles all texture extraction logic
-3. **Composition Integration**: Sensor modules use composition ("has-a" relationship) to incorporate texture processing
-4. **Broad Applicability**: Not limited to Habitat-specific sensor modules, works with any RGBA+depth data
-5. **Clean Architecture**: Avoids multiple inheritance complexity and method resolution order issues entirely
-6. **Consistent with User Preferences**: Follows the established preference for composition over mixin patterns
+The implementation consists of standalone texture processing utilities, dedicated feature processors for LBP and
+depth-based features (kept separate for modularity), and clean integration into sensor modules through composition
+rather than inheritance. This approach works with any RGBA+depth data and maintains compatibility with existing sensor
+module patterns.
 
-### 1. Texture Processing Utilities
+### 1. LBP Texture Processing
 
-#### New File: `src/tbp/monty/frameworks/utils/texture_processing.py`
+The LBP processing module provides configurable multi-scale rotation-invariant texture analysis using Local Binary Patterns. The system is designed to be easily reconfigurable for different radius and point combinations - we propose starting with radii [1, 2] pixels and point counts [8, 16] as reasonable defaults based on literature, but these parameters should be systematically tested and optimized for different object types and scenarios.
+
+Key capabilities include fully configurable multi-scale analysis, robust input validation, normalized histogram generation, and graceful handling of edge cases. The module generates descriptive feature names like `lbp_texture_r1_p8` and `lbp_texture_r2_p16` that reflect the specific parameter combinations used, making it easy to experiment with different configurations.
+
+*Implementation details: See [Appendix A: LBP Processing Implementation](#appendix-a-lbp-processing-implementation)*
+
+### 2. Depth Texture Processing
+
+The depth texture processing module extracts tactile-like features from surface depth variations to approximate texture information that would be available through touch. It computes surface roughness (scaled standard deviation of depth), gradient magnitude (rate of depth change), and surface normal consistency (alignment with expected surface orientation).
+
+The implementation includes robust handling of edge cases such as empty patches, insufficient on-object points, and non-square patch geometries. It uses Gaussian smoothing before gradient computation to reduce noise and employs careful NaN handling to prevent runtime warnings. Features are generated with names like `depth_texture_roughness`, `depth_texture_gradient`, and `depth_texture_consistency`.
+
+*Implementation details: See [Appendix B: Depth Texture Processing Implementation](#appendix-b-depth-texture-processing-implementation)*
+
+#### Key Design Principles
+
+Both LBP and depth texture processing follow a modular design with standalone functions that can be easily tested and
+integrated into any sensor module with RGBA+depth data. LBP provides multi-scale analysis with rotation invariance
+through uniform patterns, while depth processing extracts tactile-like surface characteristics including roughness,
+gradient patterns, and surface normal consistency.
+
+### 3. Feature Processors
+
+We implement separate processors for LBP and depth-based features to enable independent usage and avoid coupling
+unrelated functionality.
+
+#### LBP Feature Processor
+
+The LBP processor provides configurable multiscale texture analysis with preset configurations (basic, enhanced,
+complete) to simplify adoption. It handles feature name generation, extraction coordination, and maintains compatibility
+with existing sensor module patterns.
+
+#### Depth Texture Feature Processor
+
+The depth processor extracts tactile-like features from surface variations, including roughness metrics, gradient
+analysis, and surface normal consistency checking. It provides robust handling of edge cases and invalid data points.
+
+### 4. Integration Approach
+
+Sensor modules integrate texture processing through composition, instantiating the appropriate processors (LBP and/or
+depth) based on their feature requirements. The processors provide a clean interface for feature extraction,
+configuration management, and compatibility with existing noise systems.
+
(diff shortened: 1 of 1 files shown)


rfcs/0000_enhanced-non-morphological-features.md
```diff
@@ -0,0 +1,1050 @@
...
+
+## Questions for Community
+
+### Technical Design Questions
+1. **Composition Architecture**: Is the TextureFeatureProcessor composition approach optimal?
+   - Should the processor be created lazily only when texture features are needed?
+   - How should we handle processor lifecycle (creation, updates, cleanup)?
+   - Does this approach align well with the existing NoiseMixin integration pattern?
+
+2. **Feature Naming Convention**: Are the proposed multi-scale feature names intuitive?
```
scottcanoe: **suggestion:** ‏I noticed that `TextureFeatureProcessor.extract_features` looks for strings that start with `depth_`. IMO, this casts too wide of a net. You're probably better off selecting prefix and naming convention that's less likely to clash with other features (whether they exist yet or not). Something more like what you did with `lbp_*` would be safer.

rfcs/0000_enhanced-non-morphological-features.md
```diff
@@ -0,0 +1,1050 @@
...
+- **Composition over Inheritance**: TextureFeatureProcessor uses composition to avoid multiple inheritance complexity
+- **Backward Compatibility**: All existing functionality is preserved, new features are opt-in
+- **Feature Classification**: Texture features are correctly classified as non-morphological
+- **Configuration Flexibility**: Multiple configuration presets for different use cases
+- **Performance Considerations**: Lazy initialization of texture processor only when needed
+
+## Questions for Community
+
+### Technical Design Questions
+1. **Composition Architecture**: Is the TextureFeatureProcessor composition approach optimal?
```
scottcanoe: **question/suggestion**: Is there a reason to combine LBP- and depth-based texture processing into one interface (i.e., `TextureFeatureProcessor`)? Though they are both related to textured features, I think there are good reasons to keep them separate.
  1. I'll expound on this elsewhere, but I'm thinking LBP might actually have the most immediate and clear utility with 2D datasets like MNIST or Omniglot. In that case, we'd like to leave depth-based texture processing out of the equation.
  2. Correct me if I'm wrong, but LBP- and depth-based methods don't seem to need or make use of each other. I'd suggest making separate processor interfaces to make it easier to mix and match different implementations as needed. I'm guessing others will want to weight in about separation of concerns as well.

I see you've got "Cross-modal RGB-depth texture integration" in Phase 4 of your plan, in which case you might want to have both methods inform each other in some way. Still, I think something like that is a ways off, and you can always experiment with some hybrid approach later.

rfcs/0000_enhanced-non-morphological-features.md
```diff
@@ -0,0 +1,1050 @@
...
+
+## Motivation
+
+### Current Limitations
+
+The current non-morphological feature extraction in sensor modules is limited to:
+- **Single-pixel sampling**: Only extracts RGBA/HSV from the center pixel (1/4096 pixels used)
+- **No texture information**: Cannot distinguish between smooth vs. textured surfaces with identical colors
+- **Basic depth statistics**: Only min/mean depth, missing surface texture information
+- **No rotation invariance**: Current features change with sensor orientation
```
scottcanoe: Which ones?

rfcs/0000_enhanced-non-morphological-features.md
```diff
@@ -0,0 +1,1050 @@
...
+- **Limited cross-modal integration**: RGB and depth features are processed independently
+
+### Problem Statement
+
+From the [Extract Better Features](https://thousandbrainsproject.readme.io/docs/extract-better-features) documentation:
+> "Currently non-morphological features are very simple, such as extracting the RGB or hue value at the center of the sensor patch. In the short term, we would like to extract richer features, such as using HTM's spatial-pooler or Local Binary Patterns for visual features, or processing depth information within a patch to approximate tactile texture."
+
+The documentation specifically requires **rotation invariant** features that can detect textured patterns regardless of sensor orientation.
+
+### Research Questions
```
scottcanoe: **suggestion:** Good questions, but we're still getting a handle on what the main proposal is at this point in the document. There's also another section called Research Direction Questions. I'd recommend consolidating the two sections, and probably holding off on any non-essential discussion points until much later. For example, HTM Spatial Pooling is discussed as a future possibility, but it's not core to the proposal here like LBP or the depth-based method you present.

At this point in the document, I think the main research questions are simply how much improvement we might see with the extracted features you're suggesting, and how robust these features are to noise, scale, etc.

rfcs/0000_enhanced-non-morphological-features.md
```diff
@@ -0,0 +1,1050 @@
...
+### Current Limitations
+
+The current non-morphological feature extraction in sensor modules is limited to:
+- **Single-pixel sampling**: Only extracts RGBA/HSV from the center pixel (1/4096 pixels used)
+- **No texture information**: Cannot distinguish between smooth vs. textured surfaces with identical colors
+- **Basic depth statistics**: Only min/mean depth, missing surface texture information
+- **No rotation invariance**: Current features change with sensor orientation
+- **Limited cross-modal integration**: RGB and depth features are processed independently
+
+### Problem Statement
```
scottcanoe: **nitpick/suggestion:** Problem Statement seems to logically precede Current Limitations while also being a bit redundant with the Current Limitations section. In my opinion, the Motivation section doesn't need subsections, and the simplification and removal of redundant content is going to make it easier to hold onto the thread. (Also see the next comment about the Research Questions subsection.)

rfcs/0000_enhanced-non-morphological-features.md
```diff
@@ -0,0 +1,1050 @@
...
+### Architecture Overview
+
+The implementation follows a **pure composition approach** that can be integrated into any sensor module:
+
+1. **Texture Processing Utilities**: Standalone functions in `src/tbp/monty/frameworks/utils/texture_processing.py`
+2. **Texture Feature Processor**: Dedicated `TextureFeatureProcessor` class that handles all texture extraction logic
+3. **Composition Integration**: Sensor modules use composition ("has-a" relationship) to incorporate texture processing
+4. **Broad Applicability**: Not limited to Habitat-specific sensor modules, works with any RGBA+depth data
+5. **Clean Architecture**: Avoids multiple inheritance complexity and method resolution order issues entirely
+6. **Consistent with User Preferences**: Follows the established preference for composition over mixin patterns
```
scottcanoe: **suggestion:** Bullet points 3, 5, and 6 are pretty much saying the same thing. Recommend consolidation.

rfcs/0000_enhanced-non-morphological-features.md
```diff
@@ -0,0 +1,1050 @@
...
+1. **Feature Complementarity**: Should LBP and HTM Spatial Pooler be used together for enhanced texture discrimination?
+2. **Cross-Modal Integration**: How can RGB and depth texture features be optimally combined?
+3. **Computational Trade-offs**: What is the acceptable computational overhead for improved recognition accuracy?
+4. **Feature Interaction**: How do new texture features correlate with existing morphological features?
+
+## Detailed Design
+
+### Architecture Overview
+
+The implementation follows a **pure composition approach** that can be integrated into any sensor module:
```
scottcanoe: **suggestion:** As a reader, it'd be really helpful if you'd provide more of a high-level description before getting lists or bullet points. For example, this section could start with something like

> To utilize the new features suggested in this RFC, we must first enable sensor modules to extract them. We propose a compositional approach wherein each sensor module owns a set of feature-specific data processors that can be configured and utilized as needed. These feature-specific data processors are intended to serve as a simple interface to standalone functions that do not require the attributes or methods of an owning sensor module. As such, we can avoid complex inheritance patterns incurred by a subclass- or mixin-based approach by preferring a compositional strategy.

I find something like the above easier to follow than a list since it walks you through the logic a bit.

scottcanoe (commented):
Hi @OgnjenX, thank you for this detailed RFC! I think there are some pretty promising ideas here, and I'm looking forward to seeing how this shapes up!

**Praise**
 - This RFC is solidly rooted in TBP's goals, and you've done a great job of interpreting the enhanced features objective as well some of our other guidelines (e.g., our preferences regarding deep learning).
  - Your design choices are principled (composition, easily unit-testable standalone functions, etc.) and well-suited to Monty.
  - You've clearly thought a lot about how to go about implementing these features, not just in terms of the code, but also in terms of intermediate goals and a roadmap.

**Suggestions**
I've left a handful of suggestions in the comments, but I wanted to explain a bit more here on a higher-level. The main challenge with this RFC is its length. It's pretty long and structurally complicated, and it'll be more inviting if you can streamline and reorganize it a bit. A few ideas:
  - Consolidate sections where possible. As an example, see my comments about "Motivation"'s subsections and consolidating the "Research Questions" and "Research Direction Questions" subsections.
  - There's a good amount of repetition. Composition is referenced 22 times (not including in code samples), and rotational invariance almost as much. I think it's good to stress these points earlier on, but I think we've got it after that.
  - The team can correct me if I'm wrong on this, but we're generally more focused on concepts than implementation details in the earlier stages of an RFC. Your code look great! But I wonder if there's a way to "hide" some of it so as not to get too distracted by it at this stage. For example, your points about feature dimensionality explosion are a lot more interesting than how you're thinking about configuration strategies. I also assume implementation strategies are subject to a lot of change after back-and-forth discussions.
  - There are a whole lot of lists. I tend to glaze over them. I'm not sure what to suggest here except maybe push lower priority content into an appendix or something.
    
**Discussion**
Here's a handful of questions I had while reading through. Curious to hear everyones thoughts.
  1. How sensitive is LBP to scale? A sensor patch will cover a much larger part of an object's surface if the agent is 20 cm from the object compared with 10 cm. In other words, if we've associated some LBP metrics with a given point when observed from when 20 cm away, what happens when we observe the same point at a distance of 10 cm during inference? It doesn't seem like this would be too bad to test.
  2. Similarly, how do we handle viewing an object's edge? I believe you touched on this, but I just wanted to echo your points about how we need to make some decisions about what to do when some of the image patch is off-object.
  3. In general, I'm very curious to see how LBP works in 3D. It seems really well-suited to 2D images. It may turn out that LBP makes a small amount of difference with 3D datasets but provides a huge boost in 2D. Perhaps even using a non-rotationally-invariant version in 2D would make sense, but I'd have to think about that more.
  4. What are your thoughts on the feasibility of either LBP or depth-based metrics for distant vs surface agents? My intuition is that the sensor patch would have to be surface-agent-close to the object for texture-level variations in depth to show up.

And thanks again for the contribution! Looking forward to keeping tabs on the discussion as it progresses.

rfcs/0000_enhanced-non-morphological-features.md
OgnjenX: I agree.

rfcs/0000_enhanced-non-morphological-features.md
OgnjenX: You are right. It is probably better to keep them as separate interfaces.

rfcs/0000_enhanced-non-morphological-features.md
OgnjenX: Good point, I will improve that in my next commit.

OgnjenX:
> Hi @OgnjenX, thank you for this detailed RFC! I think there are some pretty promising ideas here, and I'm looking forward to seeing how this shapes up!
> 
> **Praise**
> 
> * This RFC is solidly rooted in TBP's goals, and you've done a great job of interpreting the enhanced features objective as well some of our other guidelines (e.g., our preferences regarding deep learning).
> * Your design choices are principled (composition, easily unit-testable standalone functions, etc.) and well-suited to Monty.
> * You've clearly thought a lot about how to go about implementing these features, not just in terms of the code, but also in terms of intermediate goals and a roadmap.
> 
> **Suggestions** I've left a handful of suggestions in the comments, but I wanted to explain a bit more here on a higher-level. The main challenge with this RFC is its length. It's pretty long and structurally complicated, and it'll be more inviting if you can streamline and reorganize it a bit. A few ideas:
> 
> * Consolidate sections where possible. As an example, see my comments about "Motivation"'s subsections and consolidating the "Research Questions" and "Research Direction Questions" subsections.
> * There's a good amount of repetition. Composition is referenced 22 times (not including in code samples), and rotational invariance almost as much. I think it's good to stress these points earlier on, but I think we've got it after that.
> * The team can correct me if I'm wrong on this, but we're generally more focused on concepts than implementation details in the earlier stages of an RFC. Your code look great! But I wonder if there's a way to "hide" some of it so as not to get too distracted by it at this stage. For example, your points about feature dimensionality explosion are a lot more interesting than how you're thinking about configuration strategies. I also assume implementation strategies are subject to a lot of change after back-and-forth discussions.
> * There are a whole lot of lists. I tend to glaze over them. I'm not sure what to suggest here except maybe push lower priority content into an appendix or something.
> 
> **Discussion** Here's a handful of questions I had while reading through. Curious to hear everyones thoughts.
> 
> 1. How sensitive is LBP to scale? A sensor patch will cover a much larger part of an object's surface if the agent is 20 cm from the object compared with 10 cm. In other words, if we've associated some LBP metrics with a given point when observed from when 20 cm away, what happens when we observe the same point at a distance of 10 cm during inference? It doesn't seem like this would be too bad to test.
> 2. Similarly, how do we handle viewing an object's edge? I believe you touched on this, but I just wanted to echo your points about how we need to make some decisions about what to do when some of the image patch is off-object.
> 3. In general, I'm very curious to see how LBP works in 3D. It seems really well-suited to 2D images. It may turn out that LBP makes a small amount of difference with 3D datasets but provides a huge boost in 2D. Perhaps even using a non-rotationally-invariant version in 2D would make sense, but I'd have to think about that more.
> 4. What are your thoughts on the feasibility of either LBP or depth-based metrics for distant vs surface agents? My intuition is that the sensor patch would have to be surface-agent-close to the object for texture-level variations in depth to show up.
> 
> And thanks again for the contribution! Looking forward to keeping tabs on the discussion as it progresses.

Hi @scottcanoe, thank you for the detailed review.
I just added a commit that aims to incorporate your suggestions into the existing RFC.

I will also provide my view on the questions you raised:

1. LBP Scale Sensitivity (20cm vs 10cm viewing distance)

You're right that this is a key challenge. When the agent moves from 20cm to 10cm, the same sensor patch will capture very different texture granularity, from seeing multiple texture elements to seeing fine details within a single element.
Proposed approach:
Multi-scale mitigation: Using configurable radii (starting with [1,2] as educated guesses) to capture both fine and coarse patterns simultaneously
Expectation: Coarse patterns (radius 2) should be more stable across distance changes than fine patterns (radius 1)
Systematic testing: We've made radius selection fully configurable, specifically to test different combinations like [1,3], [1,2,3], [2,4], and find what works best
Proposed validation:
Controlled experiments with the same textured objects at 10cm, 15cm, and 20cm distances
Measure feature vector cosine similarity across distances (target >0.9 for stability)
If scale sensitivity is too high, we could add distance-adaptive radius selection
The good news is that this should be straightforward to test empirically, as you mentioned.


2. Edge Handling and Off-Object Regions

Completely agree, this needs careful handling. 
Proposed approach:
For LBP features:
Filter to on-object points only before LBP computation
Minimum coverage threshold (require ≥50% of the patch to be on-object)
Graceful degradation - return zero/default features for insufficient coverage
Feature names indicate when edge cases occur
For depth features:
Already implemented: filter obs_3d[:, 3] > 0 to on-object points only
Handle insufficient data points (< 2 valid points) by returning default values
Use robust statistics (ddof=0, nanmean) to prevent numpy warnings
Trade-off: We accept some information loss near edges to avoid crashes and invalid calculations. This seems like the right engineering choice.


3. LBP Performance in 3D vs 2D

I completely agree with your intuition here. LBP will likely show different effectiveness:
2D datasets (MNIST, Omniglot):
Should see clear benefits - LBP is proven for 2D texture analysis
Rotation invariance is particularly valuable for handwritten digits
This is where we expect the strongest initial results
3D datasets (YCB objects):
Effectiveness depends on texture richness and sensor resolution
Many YCB objects may not have rich enough texture for LBP to help significantly
Depth-based texture features specifically designed to complement LBP in 3D scenarios
Hybrid strategy:
LBP handles RGB color patterns (works in both 2D and 3D)
Depth texture features handle surface variations (3D-specific)
The combined approach should work better than either alone
Non-rotational LBP in 2D: Interesting idea! Could capture orientation-specific patterns that uniform LBP misses. Worth exploring if uniform LBP proves insufficient.


4. Distant vs Surface Agent Feasibility

Your intuition perfectly aligns with mine. Surface agents are the primary target:
Surface agents:
Proximity enables the detection of fine texture variations
Depth gradients are meaningful at surface-level distances
The full texture feature suite makes sense
Distant agents:
Fine-grained texture analysis probably won't help much
Might benefit from very coarse texture patterns for object-level discrimination
Should use simplified feature sets or skip texture entirely
Proposed approach:
Focus initial development and validation on surface agents
Make texture features completely configurable per agent type
Test the distant agent applicability as a secondary objective
Consider distance-adaptive feature selection (coarse features only for distant agents)
This reinforces our modular design - different agent types can configure different texture feature subsets based on their sensing capabilities.

PR #365, opened 2025-07-08
chore: Vedo interactive plot

ramyamounir:
This PR adds a vedo interactive plot with the goal of visualizing the target objects, MLH, sensor at every step of the `object_evidence_over_time` plot. The plot provides a way to manually step through the time steps to inspect sensor location and the accuracy of the MLH.

```bash
usage: cli.py interactive_objects_evidence_over_time [-h] [--debug] [--objects_mesh_dir OBJECTS_MESH_DIR] [-lm LEARNING_MODULE]
                                                     experiment_log_dir

positional arguments:
  experiment_log_dir    The directory containing the experiment log with the detailed stats file.

optional arguments:
  --objects_mesh_dir OBJECTS_MESH_DIR
                        The directory containing the mesh objects.
  -lm LEARNING_MODULE, --learning_module LEARNING_MODULE
                        The name of the learning module (default: "LM_0").
```

Here is a demo:

https://github.com/user-attachments/assets/48272108-bc82-4865-8d0e-1ea8d608faae

> [!NOTE]
> This visualization only works on unsupervised inference experiments because it requires logged data from [`TheoreticalLimitLMLoggingMixin`](https://github.com/thousandbrainsproject/tbp.monty/blob/08cbaae8a7c29dce14badd57ec1be664f62486e3/src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py?plain=1#L20). Additionally, the configs should specify detailed logging. I apply my rapid configs code on top of the unsupervised inference configs as follows:

```python

# === Rapid Prototyping Configs === #

# This enables or disables rapid prototyping configs
APPLY_RAPID_CONFIGS = True

# Changes the number of rotations per object
NUM_ROTATIONS = 1

# Changes the types of YCB objects evaluated
OBJECTS_LIST = ["strawberry", "banana", "mug", "potted_meat_can", "mustard_bottle"]

# Controls the number of evaluation steps for each object
EVAL_STEPS = 50

# Controls whether to output detailed logs
DETAILED_LOG = True

# === End Rapid Prototyping Configs === #

# define rotations
test_rotations = get_cube_face_and_corner_views_rotations()[:NUM_ROTATIONS]

# define monty loggers
monty_handlers = [
    BasicCSVStatsHandler,
]
if DETAILED_LOG:
    monty_handlers.append(DetailedJSONHandler)


# define data path for supervised graph models
model_path_10distinctobj = os.path.join(
    pretrained_dir,
    "surf_agent_1lm_10distinctobj/pretrained/",
)



# Apply prototyping configs
if APPLY_RAPID_CONFIGS:
    unsupervised_inference_distinctobj_surf_agent.update(
        dict(
            experiment_args=EvalExperimentArgs(
                model_name_or_path=model_path_10distinctobj,
                n_eval_epochs=NUM_ROTATIONS,
                max_eval_steps=EVAL_STEPS,
            ),
            logging_config=DetailedEvidenceLMLoggingConfig(
                monty_handlers=monty_handlers,
                wandb_handlers=[],
            ),
            eval_dataloader_args=EnvironmentDataloaderPerObjectArgs(
                object_names=OBJECTS_LIST,
                object_init_sampler=PredefinedObjectInitializer(
                    rotations=test_rotations
                ),
            ),
        )
    )

    unsupervised_inference_distinctobj_dist_agent.update(
        dict(
            experiment_args=EvalExperimentArgs(
                model_name_or_path=model_path_10distinctobj,
                n_eval_epochs=NUM_ROTATIONS,
                max_eval_steps=EVAL_STEPS,
            ),
            logging_config=DetailedEvidenceLMLoggingConfig(
                monty_handlers=monty_handlers,
                wandb_handlers=[],
            ),
            eval_dataloader_args=EnvironmentDataloaderPerObjectArgs(
                object_names=OBJECTS_LIST,
                object_init_sampler=PredefinedObjectInitializer(
                    rotations=test_rotations
                ),
            ),
        )
    )


```


hlee9212:
Hey @ramyamounir, I'm not officially assigned but I may take a look at the PR because I'm really curious how you implemented! I think I may be extending for visualization for resampling / re-anchoring, possibly. Also want to think about how to use the `wrapt` package that @scottcanoe mentioned, which may help us visualize in "real time".

ramyamounir:
Yeah that would be great, thanks @hlee9212!

hlee9212:
Hey @ramyamounir, just one high-level lint-related comment: I noticed you are using strings to represent paths. While these do not currently throw an error, we will likely address Ruff linting rules about Path (i.e. since Python 3.4, the `pathlib` library is preferred over `os.path` module), e.g. PTH103, 104, etc. 

Feel free to keep the code as is as I or someone else may bulk update across the codebase, but just FYI! 😄 (I'm a terrible offender of using `os.path.join` especially....but because I started using Python 3.2 before the `pathlib` library and I have kept my bad habit for....14 years. *ashamed* So this is also partly to help me break my bad habit.) 

```
# pyproject.toml
"PTH103", # PTH103: `os.makedirs()` should be replaced by `Path.mkdir(parents=True)`
"PTH104", # PTH104: `os.rename()` should be replaced by `Path.rename()`
"PTH107", # PTH107: `os.remove()` should be replaced by `Path.unlink()`
"PTH110", # PTH110: `os.path.exists()` should be replaced by `Path.exists()`
"PTH111", # PTH111: `os.path.expanduser()` should be replaced by `Path.expanduser()`
"PTH118", # PTH118: `os.path.join()` should be replaced by `Path` with `/` operator
"PTH119", # PTH119: `os.path.basename()` should be replaced by `Path.name`
"PTH120", # PTH120: `os.path.dirname()` should be replaced by `Path.parent`
"PTH123", # PTH123: `open()` should be replaced by `Path.open()`
```

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,769 @@
...
+        """
+
+        def find_glb_file(obj_name: str) -> str:
+            """Search for the .glb.orig file of a given YCB object in a directory.
+
+            Args:
+                obj_name: The object name to search for (e.g., "potted_meat_can").
+
+            Returns:
+                Full path to the .glb.orig file if found, else None.
```
scottcanoe:
```suggestion
                Full path to the .glb.orig file.
```

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,769 @@
...
+        """Reads a 3D object file in glb format and returns a Vedo Mesh object.
+
+        Args:
+            obj_name: Name of the object to load.
+
+        Returns:
+            vedo.Mesh object with UV texture and transformed orientation.
+        """
+
+        def find_glb_file(obj_name: str) -> str:
```
scottcanoe: **issue:**  The YCB object dataset's peculiar usage of the `.orig` suffix is assumed here, which is totally fine for now, but I noticed that there is a command-line argument for specifying mesh directories (`--objects_mesh_dir`). Presumably you were planning ahead so that this code will work with non-YCB datasets, but I wouldn't expect other datasets will have this `*.orig` pattern. For example, the compositional object dataset I just made is Habitat-compatible, and so the overall file and directory structure is very similar, but there are no `*.orig` files.

**thought:** In the near future, we should consider building standard interfaces to object datasets that would help make scripts like this more dataset agnostic. I recently made an interface to Habitat-compatible datasets that also performs file-finding, so a common interface would also cut down on duplicated effort.

tools/plot/interactive_objects_evidence_over_time.py
scottcanoe: &#8593; since a not-found file results in an exceptions -- not a `None` return.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,769 @@
...
+        self.sensor_object.pos(sensor_pos)
+
+        # Add or update the viewing ray line
+        if self.line_object is None:
+            line_obj = Line(sensor_pos, patch_loc, c="black", lw=2)
+            self.line_object = line_obj
+            plotter.at(self.renderer_ix).add(line_obj)
+        self.line_object.points = [sensor_pos, patch_loc]
+
+    def get_axes(self) -> dict[str, tuple[float, float]]:
```
scottcanoe: **suggestion**: Perhaps rename to `get_axes_ranges`, `get_axes_info`, or similar. As a first-time reader, I would have expected this to return some kind of Axes object.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,769 @@
...
+        # Initialize the MLH simulator
+        self.mlh_sim = MlhSimulator(data_extractor=self.data_extractor, renderer_ix=2)
+
+        # Initialize the Evidence Plotter
+        self.evidence_plotter = EvidencePlot(
+            data_extractor=self.data_extractor, renderer_ix=0
+        )
+
+        # Create a plotter with 3 renderers (2 on top, 1 on bottom)
+        self.plt = Plotter(shape="2/1", size=(1000, 1000), sharecam=False)
```
scottcanoe: **question/suggestion**: Is naming a `Plotter` object `plt` standard in `vedo` usage? If not, I suggest renaming this to `plotter` just because `plt` is so strongly associated with `matplotlib.pyplot`.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,769 @@
...
+    """An interactive 3-pane plot for sensor data, MLH hypotheses, and evidence scores.
+
+    Args:
+        exp_path: Path to the JSON directory containing detailed run statistics.
+        data_path: Path to the root directory of YCB object meshes.
+        learning_module: Which learning module to use for data extraction.
+        throttle_time: Minimum delay between slider callbacks (seconds).
+            Defaults to 0.2 seconds.
+
+    Attributes:
```
scottcanoe, line 560: **praise**: Seeing attributes documented like this is a nice surprise. Everything is well-documented and readable in general.

tools/plot/interactive_objects_evidence_over_time.py
scottcanoe: **comment**: Here's another weird quirk of the YCB dataset with habitat. All of the metadata files under `ycb/configs` tell habitat to use the `textured.glb` files. Yet I'm assuming that you were only able to read and render the `textured.glb.orig` files. I had that issue, but I never figured out why `textured.glb` files are unusable outside of habitat.

scottcanoe (commented):
Very nice, Ramy. The code is easy to understand, and the visualization itself looks great.

I'm in agreement with @hlee9212  about using `pathlib`. I only ever use strings to represent paths when downstream code doesn't support `pathlib.Path` objects. It'd be an easy refactor, and I promise you'll prefer `pathlib` over `str` (if you don't already).

I left a few small suggestions, mostly targeting readability for first-time readers.

tools/plot/interactive_objects_evidence_over_time.py
hlee9212: Can't agree more to this comment!

tools/plot/interactive_objects_evidence_over_time.py
ramyamounir: Yes, you're right. I originally had it return `None` and then changed it but forgot to update the docstring.

tools/plot/interactive_objects_evidence_over_time.py
ramyamounir: Yeah, I agree with everything you said here. A standard interface would be quite helpful! I found that `textured.glb` do not show texture for some reason. It could be that the texture data is stored differently and I couldn't extract it, but the texture wouldn't even show in blender when I tried for last year's hackathon. Monty reads those config files, I'm surprised that habitat can render the texture given that it's pointing to `textured.glb`.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,769 @@
...
+    Plotter,
+    Rectangle,
+    Slider2D,
+    Sphere,
+    Text2D,
+    Text3D,
+    settings,
+)
+
+from tbp.monty.frameworks.environments.ycb import DISTINCT_OBJECTS
```
nielsleadholm, line 32: **suggestion:** It would be best to avoid hard-coding the number of colors based on `DISTINCT_OBJECTS`.

tools/plot/interactive_objects_evidence_over_time.py
ramyamounir: I've renamed it to `get_axes_dict`, the `axes` argument of the `show` function can take a string or dict (for more control).

tools/plot/interactive_objects_evidence_over_time.py
ramyamounir: Good point!

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,769 @@
...
+
+        Args:
+            plotter: The vedo.Plotter instance with multiple renderers.
+            step: The timestep to visualize.
+        """
+        obj_name, target_pos, mlh_pos, mlh_rot = self.data_at_step(step)
+
+        # Create a transformed clone of the MLH mesh
+        mlh_obj = self.data_extractor.imported_objects[obj_name].clone(deep=True)
+        mlh_obj.rotate_x(mlh_rot[0])
```
nielsleadholm, line 378: **question:** Are we applying an inverse anywhere? I'm kind of surprised that this is working, as in the code, the MLH rotation is stored as the inverse of the ground-truth rotation.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,769 @@
...
+            pos=(0.9, 0.9),
+            states=["Align"],
+            size=30,
+            font="Calco",
+            bold=True,
+        )
+
+    def align_camera(self, cam1: Any, cam2: Any) -> None:
+        """Align the camera objects."""
+        cam1.SetPosition(cam2.GetPosition())
```
nielsleadholm: **nitpick:** It might be a bit clearer calling these `camA` and `camB` or the like, since `cam1` and `cam2` have more general meanings.

ramyamounir:
Thanks @hlee9212 and @scottcanoe. You guys are right about `pathlib`! `os.path` has been my comfort zone for so long, but this is much better.

tools/plot/interactive_objects_evidence_over_time.py
ramyamounir: Actually, I might just end up renaming it to `axes_dict` because we're trying to avoid `get_` and `set_` prefixes.

scottcanoe: I'm not a `get_` / `set_` hater, but I can go with the flow.

tristanls: _I felt a disturbance in the force..._

tools/plot/interactive_objects_evidence_over_time.py
ramyamounir: Yeah, ruff has a problem with the uppercase A and B in the variable names, so I named them `cam_{a,b}`

tools/plot/interactive_objects_evidence_over_time.py
ramyamounir: Yeah fair point, I looked into it and it seems that we are applying an inverse for logging in this [function](https://github.com/thousandbrainsproject/tbp.monty/blob/e701a15caa2febdb1531cf0b7075103c0537c07c/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L621-L642), which is called from `_add_detailed_stats` [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e701a15caa2febdb1531cf0b7075103c0537c07c/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L1173-L1184). That said, this inverse is applied to `possible_rotations` not the `current_mlh`, I ended up checking that the logged `possible_rotation` at the argmax of `evidences` is the same as the `current_mlh['rotation']`. So, the rotation of the `current_mlh` must have had an inverse somewhere, but I looked everywhere and can't find where we apply this `inv()` on the `current_mlh` before it gets logged. I think it's worth spending some time simplifying all the rotation operations, because it's becoming a bit difficult to manage.

tools/plot/interactive_objects_evidence_over_time.py
ramyamounir: I agree. The limit on the number of colors is more constrained by the color map `tab10`, it has 10 visually distinct colors. I didn't want to use a color map with more colors as some of these colors start to look similar. It just felt like the best option for unsupervised inference experiments which always have 10 objects only. `DISTINCT_OBJECTS` is also used in the same way in the matplotlib plot `objects_evidence_over_time` [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e701a15caa2febdb1531cf0b7075103c0537c07c/tools/plot/objects_evidence_over_time.py?plain=1#L51-L54), so I think it's probably best I keep it consistent and update both together when we try to use these visualization for 77 objects. What do you think?

ramyamounir:
Thanks for the review @nielsleadholm, these are really nice suggestions for improving the visualization. I especially like the toggle to show the path taken.

> What is the uv.lock file diff for? Is that necessary/intentional?

When I added `vedo` and `trimesh` to the `pyproject.toml` as dependencies, the github workflow `test_uv_monty` failed [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e701a15caa2febdb1531cf0b7075103c0537c07c/.github/workflows/monty.yml?plain=1#L389). `uv sync` runs on the `uv.lock` file and it seems to complain if the lock file is different from what we have in the `pyproject.toml`. @jeremyshoemaker might be the best person to ask about this because he was recently adding mujoco dependencies. Also @jeremyshoemaker, I only ran `uv lock`, was I supposed to run `uv lock --extra dev --extra simulator_mujoco`?

> Thanks for the documentation in the PR comment, is it possible to just remind me what the "rapid prototyping" configs are / what that term refers to?

Sure, the rapid prototyping configs just simplifies the configs to only a few objects and rotations instead of running the full benchmark. I had proposed to add this to the unsupervised inference configs but we ended up removing it because it didn't contribute to the benchmark tables. [Here](https://github.com/thousandbrainsproject/tbp.monty/pull/221#discussion_r2021255927) is the relevant discussion.

uv.lock
tristanls, line 1: **question**: Any idea why `uv.lock` changed?

pyproject.toml
tristanls, line 1: **issue**: I noticed that the `plot` tool README states that "No additional dependencies are required." However, it looks like to use the tool one needs to:

```
pip install -e '.[dev,analysis]'
```

Following the existing instructions results in:
```
(tbp.monty) me@nervous_energy tbp.monty % python -m tools.plot.cli -h
  File "/Users/me/miniconda3/envs/tbp.monty/lib/python3.8/runpy.py", line 194, in _run_module_as_main
  File "/Users/me/miniconda3/envs/tbp.monty/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/me/tbp/tbp.monty/tools/plot/cli.py", line 18, in <module>
    from tools.plot import (  # noqa: E402
  File "/Users/me/tbp/tbp.monty/tools/plot/interactive_objects_evidence_over_time.py", line 20, in <module>
    import trimesh
ModuleNotFoundError: No module named 'trimesh'
```

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+
+        # Make sure that the agent positions (sampled by processed steps) is the same
+        # length as the other logged data
+        assert (
+            len(self.target_positions)
+            == len(self.sensor_positions)
+            == len(self.patch_locations)
+        )
+
+    def __len__(self):
```
tristanls: **suggestion (non-blocking)**: `-> int`.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+                        return str(glb_orig_path)
+
+            raise FileNotFoundError(
+                f"Could not find .glb.orig file for '{obj_name}' in '{self.data_path}'"
+            )
+
+        # read the glb file
+        file_path = find_glb_file(obj_name)
+        with open(file_path, "rb") as f:
+            mesh = trimesh.load(f, file_type="glb")
```
tristanls: **question**: `trimesh.load` is deprecated. Is there a reason not to use its replacements?

**issue**: `trimesh.load() -> Geometry`, and `Geometry` has no `geometry` property used next in `mesh.geometry.values()`.

**suggestion**: I think you want `trimesh.load_scene()`, which will return a `trimesh.Scene` that has `geometry` property later called in `mesh.geometry.values()`.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+
+    def __call__(self, plotter: Plotter, step: int = 0) -> None:
+        """Render the target object, sensor, and viewing ray at a specific timestep.
+
+        Args:
+            plotter: The vedo.Plotter instance with multiple renderers.
+            step: The timestep to visualize.
+        """
+        # Extract relevant time-step data
+        obj_name, target_pos, target_rot, sensor_pos, sensor_rot, patch_loc = (
```
tristanls: **issue**: `sensor_rot` is unused. Any reason it is extracted in `data_at_step` in the first place?

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+            self.slider_callback,
+            xmin=0,
+            xmax=len(self.data_extractor) - 1,
+            value=0,
+            pos=[(0.2, 0.05), (0.8, 0.05)],
+            title="Step",
+        )
+        self.curr_slider_val = None
+        self.last_call_time = time.time()
+        self.slider_callback(self.slider, None)
```
tristanls: **issue**: `self.slide_callback` expects a string for the second argument, `None` is not valid.

**suggestion**: `self.slide_callback(self.slider, "")`

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+        self.plotter.render()
+
+    def mlh_callback(self, widget: Button, event: str) -> None:
+        """Align the MLH renderer's camera with the groundtruth renderer."""
+        cam_a = self.plotter.renderers[1].GetActiveCamera()
+        cam_b = self.plotter.renderers[2].GetActiveCamera()
+        self.align_camera(cam_b, cam_a)
+        self.plotter.render()
+
+    def slider_callback(self, widget: Slider2D, event: str) -> None:
```
tristanls: **issue**: `event` is unused.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+        learning_module: The learning module to use for extracting evidence data.
+
+    Returns:
+        Exit code.
+    """
+    if not Path(exp_path).exists():
+        logger.error(f"Experiment path not found: {exp_path}")
+        return 1
+
+    data_path = Path(data_path).expanduser()
```
tristanls: `data_path` is expected to be a `str` and not `Path`.

```suggestion
    data_path = str(Path(data_path).expanduser())
```

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+        """Reads a 3D object file in glb format and returns a Vedo Mesh object.
+
+        Args:
+            obj_name: Name of the object to load.
+
+        Returns:
+            vedo.Mesh object with UV texture and transformed orientation.
+        """
+
+        def find_glb_file(obj_name: str) -> str:
```
tristanls: **issue**: This function is defined anew every time `create_mesh` is called for the sole purpose of using `self.data_path`. 

**suggestion**: Make this a `DataExtractor._find_glb_file` method.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+        self.data_extractor = data_extractor
+        self.renderer_ix = renderer_ix
+
+        self.target_object = None
+        self.sensor_object = None
+        self.line_object = None
+        self.title_object = None
+        self.target_name = None
+
+    def data_at_step(self, step: int) -> tuple:
```
tristanls: **suggestion (non-blocking)**: Add type parameters to `tuple`, e.g., `tuple[str, list[float], list[float], list[float], list[float], list[float]]:` or even better, more informative along the lines of `tuple[str, list[VectorXYZ], list[QuaternionWXYZ], list[VectorXYZ], list[QuaternionWXYZ], list[VectorXYZ]]:`

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+            steps = np.arange(len(evidence_scores))
+            pts = np.c_[steps, evidence_scores, np.zeros_like(steps)]
+            line = Line(pts, c=colors.get(obj, (0.5, 0.5, 0.5)), lw=3)
+            self.lines.append(line)
+
+        # Add dynamic guide line. This will be modified with the plot slider.
+        self.guide_line = Line(p0=(0, 0, 0), p1=(0, 80, 0), lw=2, c="black")
+        self.added_plot_flag = False
+
+    def axes_dict(self) -> dict:
```
tristanls: **suggestion**: Missing type parameters `-> dict[str, Any]` or `Any` (since `show()` expects `Any | None`).

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+    def __init__(
+        self,
+        exp_path: str,
+        data_path: str,
+        learning_module: str,
+        throttle_time: float = 0.2,
+    ):
+        self.throttle_time = throttle_time
+
+        # Initialize the data extractor
```
tristanls: **nitpick**: Extraneous comment.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+        self.data_extractor = DataExtractor(exp_path, data_path, learning_module)
+
+        # Initialize the ground truth simulator
```
tristanls: **nitpick**: Extraneous comment.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+        self.gt_sim = GroundTruthSimulator(
+            data_extractor=self.data_extractor, renderer_ix=1
+        )
+
+        # Initialize the MLH simulator
```
tristanls: **nitpick**: Extraneous comment.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+        self.mlh_sim = MlhSimulator(data_extractor=self.data_extractor, renderer_ix=2)
+
+        # Initialize the Evidence Plotter
```
tristanls: **nitpick**: Extraneous comment.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+
+    def align_camera(self, cam_a: Any, cam_b: Any) -> None:
+        """Align the camera objects."""
+        cam_a.SetPosition(cam_b.GetPosition())
+        cam_a.SetFocalPoint(cam_b.GetFocalPoint())
+        cam_a.SetViewUp(cam_b.GetViewUp())
+        cam_a.SetClippingRange(cam_b.GetClippingRange())
+        cam_a.SetParallelScale(cam_b.GetParallelScale())
+
+    def simulator_callback(self, widget: Button, event: str) -> None:
```
tristanls: **issue**: `widget` and `event` are unused.

tools/plot/interactive_objects_evidence_over_time.py
```diff
@@ -0,0 +1,766 @@
...
+        """Align the ground truth renderer's camera with the MLH renderer."""
+        cam_a = self.plotter.renderers[1].GetActiveCamera()
+        cam_b = self.plotter.renderers[2].GetActiveCamera()
+        self.align_camera(cam_a, cam_b)
+        self.plotter.render()
+
+    def mlh_callback(self, widget: Button, event: str) -> None:
```
tristanls: **issue**: `widget` and `event` are unused.

uv.lock
tristanls: **note**: Oh, I see the comment where @nielsleadholm brought it up. I believe `uv lock` is correct, however, you should update your `uv` version to latest. I did the same thing as you and the thousands of line changes is due to using old `uv` version. After you update, the diff should be much much less (I think).

```
% uv -V
uv 0.7.15 (Homebrew 2025-06-25)
```

tristanls (commented):
Nice tool.

tools/plot/interactive_objects_evidence_over_time.py
nielsleadholm: Ok interesting, thanks for checking. @vkakerbeck by any chance do you remember where this might happen?

tools/plot/interactive_objects_evidence_over_time.py
nielsleadholm: Ok yeah makes sense, maybe just add a warning or the like if someone tries to pass in a larger dataset?

tools/plot/interactive_objects_evidence_over_time.py
vkakerbeck: Yes, the most likely hypothesis within an LM is always in the object-centric coordinates. We only convert it back to world coordinates when we do logging (here: https://github.com/thousandbrainsproject/tbp.monty/blob/c886e187c47aac6135a15b72052c71e34009a92b/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L529 ) setting the `lm_episode_stats` `detected_rotation_quat` field which is used for calculating the rotation error to that inverse https://github.com/thousandbrainsproject/tbp.monty/blob/c886e187c47aac6135a15b72052c71e34009a92b/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L548

uv.lock
ramyamounir: Yeah, that didn't help. I upgraded to the latest `0.7.20` and still get the same `uv.lock` file. I also tried deleting the file and regenerating it.

```bash
% uv -V
uv 0.7.20 (251420396 2025-07-09)
```

pyproject.toml
ramyamounir: Nice catch, I missed the README. 

Is `dev` really required to run the visualizations? I'm thinking I'll update the README to say install the `analysis` packages, unless you think `dev` is also needed.

tools/plot/interactive_objects_evidence_over_time.py
ramyamounir: Thanks, I ended up using `load_mesh` which returns `geometry`, so I no longer have to extract the geometry from mesh. I can just simplify to this:
```python
        file_path = find_glb_file(obj_name)
        with open(file_path, "rb") as f:
            geometry = trimesh.load_mesh(f, file_type="glb")

        # create mesh from vertices and faces
        obj = Mesh([geometry.vertices, geometry.faces])
```
but I think I'll call it mesh to avoid confusion.

tools/plot/interactive_objects_evidence_over_time.py
ramyamounir: before using the `patch_loc`, I was using `sensor_pos` and `sensor_rot` to cast a ray and detect it's intersection with the mesh to find the patch location. But now this is not needed, I've cleaned it up. Thanks!

ramyamounir:
@tristanls Thanks for the detailed review, I've addressed most of your questions, except for a couple issues of dependencies and uv lock.

@vkakerbeck Yes, this should be quite possible. I was thinking for the default YCB benchmark experiments, we can have 2 global slider controls, one switches episodes and another controls the steps within the episode. The line plot showing maximum evidence may not be as useful here, we could instead show additional plots that highlight the evidence scores as a colored point cloud for the first few most likely objects. The nice thing is that every renderer/subplot can have it's own controls (buttons and sliders) which we can use to show different things at any step. I have been thinking of a plot that shows the diff of hypothesis space between steps, as in for step t+1, it would show the added hypotheses in green and the deleted in red. This would be useful for debugging resampling.

vkakerbeck:
Cool! Sounds like there are lot sof exciting possible extensions. I think for a first start, even just the plot as it is would be useful for the one-object experiments, just to see how evidence scores between objects compare and to see the mlh vs. ground truth 3D visualization. One other thing I recently found useful was to plot the past path of observations transformed by the mlh (instead of just the current observations) onto the 3D object because it shows a bit clearer why the current object and pose are the mlh and how they fit the observations so far. Nothing to add to this PR though, just throwing it out there :)

nielsleadholm (approved):
Nice, thanks for those changes Ramy, looks great!

jeremyshoemaker:
> @jeremyshoemaker might be the best person to ask about this because he was recently adding mujoco dependencies. Also @jeremyshoemaker, I only ran `uv lock`, was I supposed to run `uv lock --extra dev --extra simulator_mujoco`?

Anytime dependencies are changed in `pyproject.toml` this is going to be needed. That said, the test isn't required yet, so you can safely ignore its failures for now and I'll clean it up as I run into it with my MuJoCo work.

At some point, I need to write up some documentation on how to use the `uv` environment. It's somewhat new to me so I need to figure out the right way to use it. I think what you did with `uv lock` was fine. The reason it changed almost the whole file is because we need to make sure people use recent versions of `uv`. We've seen it change the whole file when using mismatched versions (or ones that aren't recent enough).

uv.lock
jeremyshoemaker: So, it looks like it all changed because it's reversing the change @tristanls committed a while back with an older version. It's switching the `revision` back to 2 from 1.

jeremyshoemaker: What they should have done is made older versions of `uv` refuse to update lock files that have revisions that they don't recognize, rather than just recreating the file with the older revision format.

It was a good idea to have revisions in the file format, but then to not have the tool fail on ones it doesn't recognize is a missed opportunity.

pyproject.toml
tristanls: I haven't tried it without `dev`. You can reinstall things by doing:
```
conda env remove --name tbp.monty
# click y/Y/yes to remove all the things
```
Then follow the installation instructions again and trying `pip install -e '.[analysis]'` to see if the tool still works.

ramyamounir: It looks like `dev` is already [installed](https://github.com/thousandbrainsproject/tbp.monty/blob/c886e187c47aac6135a15b72052c71e34009a92b/environment.yml?plain=1#L49-L50) as part of the installation instructions with `environment.yml`. So the only thing extra that's needed is `analysis`. I've updated the README to say that.

PR #367, opened 2025-07-11
style: address os.rename (PTH104)

hlee9212:
CleanShot 2025-07-10 at 20 39 40@2x

I didn't fix `os.path.join` yet (other ruff rule for future PR), just trying to get the ball rolling replacing `os.path` module `pathlib`. Main purpose is to remove this from the ignore list in `pyproject.toml`.


tristanls:
Please create pull requests from your fork and not from `thousandbrainsproject/tbp.monty` branch.

PR #369, opened 2025-07-12
style: address os.rename (PTH104)

hlee9212:
From #367 but from my own fork. 

CleanShot 2025-07-11 at 18 33 48


jeremyshoemaker (approved):
Looks good to me.

PR #370, opened 2025-07-13
refactor: move test-only classes into test code

scottcanoe:
I noticed that `LearningModuleBase` is essentially a fake used only for testing, but it lives in `src/tbp/monty/frameworks/models/monty_base.py`:
```python
class LearningModuleBase(LearningModule):
    """Dummy placeholder class used only for tests."""

```
Its methods define some dummy testing behavior, so it could actually be bad if someone were to inherit from it (though none do). It is only used in two config classes, which are themselves only used for testing.

This PR migrates `LearningModuleBase` into `tests` as a fake. This is also done for the two config classes that use `LearningModuleBase`. Unit tests that used these test-only classes have been updated to use the fakes.

Note: Something similar may be done with `SensorModuleBase`, but that's a bit more complicated and may be a follow up PR.

### DIFF
--- a/src/tbp/monty/frameworks/config_utils/config_args.py
+++ b/src/tbp/monty/frameworks/config_utils/config_args.py
@@ -57,11 +57,6 @@ from tbp.monty.frameworks.models.evidence_matching.model import (
     MontyForEvidenceGraphMatching,
 )
 from tbp.monty.frameworks.models.graph_matching import MontyForGraphMatching
-from tbp.monty.frameworks.models.monty_base import (
-    LearningModuleBase,
-    MontyBase,
-    SensorModuleBase,
-)
 from tbp.monty.frameworks.models.motor_policies import (
     BasePolicy,
     InformedPolicy,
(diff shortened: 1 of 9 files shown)


jeremyshoemaker (approved):
Looks good to me.

tristanls (approved):
Thanks, looks good to me too.

**suggestion**: Like `get` & `set`, `utils` is another noise word that means "I don't have a good name yet." I think we can improve the situation. Do you think anything is lost or anyone will be confused if we rename things as follows?
```
tests/unit/frameworks/config_utils/* -> tests/unit/frameworks/config/*
```

**suggestion**: Another naming suggestion is not repeating the obvious (in usage) context in names. e.g.,:
```
tests/unit/framworks/config/fakes/config_args.py -> tests/unit/framework/config/fakes/args.py
```
This ends up used as:
```python
from tests.unit.frameworks.config.fakes.args import FakeSingleCameraMontyConfig
```

scottcanoe:
> Thanks, looks good to me too.
> 
> **suggestion**: Like `get` & `set`, `utils` is another noise word that means "I don't have a good name yet." I think we can improve the situation. Do you think anything is lost or anyone will be confused if we rename things as follows?
> 
> ```
> tests/unit/frameworks/config_utils/* -> tests/unit/frameworks/config/*
> ```
> 
> **suggestion**: Another naming suggestion is not repeating the obvious (in usage) context in names. e.g.,:
> 
> ```
> tests/unit/framworks/config/fakes/config_args.py -> tests/unit/framework/config/fakes/args.py
> ```
> 
> This ends up used as:
> 
> ```python
> from tests.unit.frameworks.config.fakes.args import FakeSingleCameraMontyConfig
> ```

@tristanls I like your suggestions, and I would like do so some renaming. But it seems like the emerging structure of `tests/unit` is meant to mirror the structure of `monty`, which I think is a helpful pattern. As such, it seems like it'd be a good idea to apply renaming to `config_args` in `monty` and `tests/unit` together, no? I'm going to leave it for the time being and merge.

But for the sake of discussion, it seems like there are a bunch of potential improvements to `config_args`, like splitting configs into different files based on object type, moving simulator-specific configs into appropriate folders, etc.

PR #374, opened 2025-07-15
refactor: improve buffer channel selection logic

OgnjenX:
- Add fallback to return first available channel when no 'patch' channel found
- This supports different sensor module naming schemes beyond the legacy 'patch' naming
- Addresses TODO comment about better channel selection that doesn't rely on naming
- Maintains backward compatibility with existing 'patch' channel preference

The change prevents ValueError exceptions when using sensor modules with different naming conventions while preserving the existing behavior for legacy setups.


OgnjenX:
> Thanks, @OgnjenX! This is an unfamiliar part of the code, so as I review it, I'm noticing a few nuances. I'm pulling in @nielsleadholm to help as he's more familiar with these.
> 
> @nielsleadholm, the method `get_first_sensory_input_channel` states that it should return input from a sensory module. Inspecting `State`, there is a `state_type` field that seems to be the data we are looking for: `"SM"`, `"LM"`, or `"GSG"` (as the TODO note highlights). However, there is a complication in that we have two different types of "SM", but the `state_type` field does not distinguish them. Instead, the two types we use are encoded in sensor module names via the "patch_" prefix or the "view_finder" sentinel value. It is my understanding that, in general, we don't want to pass "view_finder" to ~sensor~ learning modules as it is intended for use by the experimental machinery/instrumentation and not by Monty itself.
> 
> So, storing `sensory_type` (per the TODO note) seems like the appropriate solution to determine whether the input comes from a sensor module. However, are we missing a sensory type for the view finder?
> 
> If we were to make `sensory_type` available, then @OgnjenX could select the first channel where `sensory_type == "SM"`.
> 
> @OgnjenX, I think that returning `all_channels[0]` may return non-sensory input if the input comes from another learning module or a goal state generator.

Thanks for the clarification @tristanls. Since I honestly didn't know that, it helps me a lot.

OgnjenX:
Hi @tristanls,
I’ve implemented a solution based on my understanding of your suggestion. When you have a moment, I’d appreciate it if you could take a look.
Best regards.

nielsleadholm:
**praise:** Thank you for working on this @OgnjenX , great to address some of these TODOs.

Re. the particular solution, I agree with your assessment Tristan. With the new change we still need to check for "viewfinder". Intuitively, if we were going to add this kind of meta-data, then the `State` class would be a good place to have it, rather than repeating some of that here. 

On the plus side, my understanding is that the view-finder shouldn't actually output a `State` object, so a simpler solution might be possible. In particular, the viewfinder does not output CMP messages, and so there is no reason for it to have a State object.

When this code runs @OgnjenX , does line 613 (`self.channel_sensory_types[input_channel] = "view_finder"`) actually run? In other words, is there actually a `State` object who has the below values?
```
state.sender_type == "SM"
state.sender_id == "viewfinder"
```

OgnjenX:
> **praise:** Thank you for working on this @OgnjenX , great to address some of these TODOs.
> 
> Re. the particular solution, I agree with your assessment Tristan. With the new change we still need to check for "viewfinder". Intuitively, if we were going to add this kind of meta-data, then the `State` class would be a good place to have it, rather than repeating some of that here.
> 
> On the plus side, my understanding is that the view-finder shouldn't actually output a `State` object, so a simpler solution might be possible. In particular, the viewfinder does not output CMP messages, and so there is no reason for it to have a State object.
> 
> When this code runs @OgnjenX , does line 613 (`self.channel_sensory_types[input_channel] = "view_finder"`) actually run? In other words, is there actually a `State` object who has the below values?
> 
> ```
> state.sender_type == "SM"
> state.sender_id == "viewfinder"
> ```

Hi @nielsleadholm, thanks for the analysis, and sorry for my late response. I debugged code now, and it seems to me that you are right, this if is never executed:

```
                if "view_finder" in input_channel:
                    self.channel_sensory_types[input_channel] = "view_finder"
```

Can we then rely on something like I did in my last commit? Would it be better to extend `State` class with `sensory_type` instead? (per my understanding, that would require more code changes, but would probably result in more structured solution)

**question**: Is `get_first_sensory_input_channel` expected to return first sensory channel, or first sensory 'patch' channel? (in new implementation I made it to do as name suggests - return first sensory input channel)

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -456,23 +463,29 @@ def get_first_sensory_input_channel(self):
...
+        # This avoids relying on naming conventions like "patch"
+        for channel in all_channels:
+            if (
+                channel in self.channel_sender_types
+                and self.channel_sender_types[channel] == "SM"
+            ):
+                return channel
+
+        # If we reach here, no sensory channels were found but channels exist
+        # This means we have channels but none are sensory (e.g., only view_finder, LM)
```
nielsleadholm:
```suggestion
        # This means we have channels but none are SMs that output CMP-compliant State observations (e.g., only view_finder, LM)
```

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -456,23 +463,29 @@ def get_first_sensory_input_channel(self):
             The name of the first sensory (coming from SM) input channel in buffer.
 
         Raises:
-            ValueError: If no sensor channels are found in the buffer
+            ValueError: If no sensory channels are found in the buffer.
         """
         all_channels = list(self.locations.keys())
```
nielsleadholm: **suggestion:** I think it would be cleaner now to use `self.channel_sender_types` rather than `self.locations.keys()` as the iterable, now that it exists.

nielsleadholm (commented):
Thanks @OgnjenX for checking that and making the change.

It's not necessary to change the `State` class because SMs which do not pass CMP signals (like the view-finder), will never pass information to the LMs. LMs can only receive CMP inputs. The view-finder, on the other hand, is an SM that passes information directly to other parts of Monty (like the motor system). As such, it is sufficient to know that the sender-type is an `SM` to know that it is a "sensory" input. Let me know if I can clarify further.

Did you have a chance to run some of the benchmark experiments with this change, in order to make sure the results are numerically identical? Particularly some of the longer (77 object) experiments would be good to double check. Thanks!

vkakerbeck:
@nielsleadholm thanks for double-checking. Yes, this should produce the desired behavior without relying on the "patch" naming convention :)

@OgnjenX thanks for addressing this TODO and making the code more generally usable!

src/tbp/monty/frameworks/models/buffer.py
OgnjenX: I updated it in [this commit](https://github.com/thousandbrainsproject/tbp.monty/pull/374/commits/e9f6154737c9db45febe992476a9bb97f0c1d051). Thank you for suggesting, Niels.

OgnjenX:
> Thanks @OgnjenX for checking that and making the change. Overall I think this looks like a reasonable approach, and yes, that should be the intended approach. @vkakerbeck just double checking with you that you can't think of any reason that might be an issue?
> 
> It's not necessary to change the `State` class because SMs which do not pass CMP signals (like the view-finder), will never pass information to the LMs. LMs can only receive CMP inputs. The view-finder, on the other hand, is an SM that passes non-CMP information directly to other parts of Monty (like the motor system). As such, it is sufficient to know that the sender-type is an `SM` to know that it is a "sensory" input. Let me know if I can clarify further, and otherwise you might find some of our documentation on what Cortical Messaging Protocol (CMP) signals are helpful.
> 
> Did you have a chance to run some of the benchmark experiments with this change, in order to make sure the results are numerically identical? Particularly some of the longer (77 object) experiments would be good to double check. Thanks!

Thanks for the useful review @nielsleadholm.

I just compared `base_77obj_surf_agent` with these changes, to results without changes included in this PR, and results seem consistent to me.

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -59,6 +59,9 @@ def __init__(self):
 
         self.displacements = {}
 
+        # Store sender_type for each channel to avoid relying on naming conventions
```
tristanls: **issue**: Please remove this comment. The pull request discussion will better record why this property exists for the curious.

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -121,6 +124,10 @@ def append(self, list_of_data):
         any_obs_on_obj = False
         for state in list_of_data:
             input_channel = state.sender_id
+
+            # Store sender_type for this channel
```
tristanls: **issue**: Remove extraneous comment (the code speaks for itself).

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -456,23 +463,27 @@ def get_first_sensory_input_channel(self):
...
-                "get_first_sensory_input_channel assumes we have at least one"
-                f" sensor channel but channels are {all_channels}."
-            )
-        else:
+        all_channels = list(self.channel_sender_types.keys())
+        if len(all_channels) == 0:
             return None
 
+        # Find first channel with sender_type "SM"
+        # This avoids relying on naming conventions like "patch"
```
tristanls: **issue**: Please remove extraneous comments. The first comment describes what the code does (the code speaks for itself in this case). The second comment mentions a historical state made irrelevant from this moment forward due to your fix.

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -456,23 +463,27 @@ def get_first_sensory_input_channel(self):
...
+        for channel in all_channels:
+            if self.channel_sender_types[channel] == "SM":
+                return channel
+
+        # If we reach here, no sensory channels were found but channels exist
+        # This means we have channels but none are SMs that output CMP-compliant
+        # State observations (e.g., only view_finder, LM)
```
tristanls, line 476: **praise**: This is a good comment. It provides the context that is not necessarily obvious from the surrounding code, but helps understand the control flow.

OgnjenX:
> Great, thanks for making those changes @OgnjenX ! Looks really nice and clean.
> 
> Before we merge: re. the benchmark results, is it possible to include a CSV or summarize the results? Before and after this change they should be numerically identical. I know it's a bit of a pain, but unfortunately our unit tests do not cover all possible issues, and we are often surprised by how results change with a small PR.

[eval_stats.csv](https://github.com/user-attachments/files/21487583/eval_stats.csv)

**With Buffer Channel Selection Changes**:
```
    Total Episodes: 231
    Correct: 212 episodes
    Correct MLH: 17 episodes
    Confused: 2 episodes
    Overall Accuracy: 99.13% (229/231)
    MLH Usage: 7.36% (17/231)
```

**Baseline**:
```
    Accuracy: 98.70%
    MLH Usage: 6.49%
```

nielsleadholm:
Thanks @OgnjenX , those do indeed look comparable, although we would expect them to be numerically identical. The difference in your results from [those reported here](https://thousandbrainsproject.readme.io/docs/benchmark-experiments#longer-experiments-with-all-77-ycb-objects) are likely related to hardware differences, what happens if you run the experiment on the same hardware (i.e. whatever you have available), but using the main branch? Do you get the exact same result as on your PR/feature branch?

OgnjenX:
> Thanks @OgnjenX , those do indeed look comparable, although we would expect them to be numerically identical. The difference in your results from [those reported here](https://thousandbrainsproject.readme.io/docs/benchmark-experiments#longer-experiments-with-all-77-ycb-objects) are likely related to hardware differences, what happens if you run the experiment on the same hardware (i.e. whatever you have available), but using the main branch? Do you get the exact same result as on your PR/feature branch?

You’re right. I just ran the baseline on my machine (main branch), and the results appear numerically identical.

[baseline_eval_stats.csv](https://github.com/user-attachments/files/21488218/eval_stats.csv)

nielsleadholm:
Amazing, thanks for checking that. There is just one last comment that Tristan has asked to be removed, and otherwise this looks good to go!

src/tbp/monty/frameworks/models/buffer.py
OgnjenX: Okay, thanks. I removed extraneous comments in [this commit](https://github.com/thousandbrainsproject/tbp.monty/pull/374/commits/566315e99c5d78268ec0d63a7085bf1357551798)

OgnjenX:
> Amazing, thanks for checking that. There is just one last comment that Tristan has asked to be removed, and otherwise this looks good to go!

Great, thanks. I removed extraneous comments in [this commit](https://github.com/thousandbrainsproject/tbp.monty/pull/374/commits/566315e99c5d78268ec0d63a7085bf1357551798).

edit: Sorry, I just realized that I missed one of @tristanls comments. I updated it now.

nielsleadholm (approved):
Amazing, thanks for your thorough follow-up @OgnjenX , looks great!

tristanls:
Since @nielsleadholm mentioned benchmarks, I'll launch a benchmark set on our hardware and report the results here before merging.

tristanls:
FYI, benchmark runs failed to run. It might be infrastructure and unrelated to this pull request. There'll be a delay while I investigate.

nielsleadholm:
Interesting, thanks for double checking @tristanls

tristanls:
**note**: I'm renaming this pull request to `refactor:`. [We use `fix:` for bugs](https://github.com/thousandbrainsproject/tbp.monty/blob/876667db5adc7c0b6997afa50992e7d98d3e09ef/rfcs/0010_conventional_commits.md?plain=1#L37), however, I do not consider the previous behavior to be a bug. I view the current code as relaxing a constraint.

tristanls:
**note**: I confirmed that the benchmark issue is an infrastructure problem unrelated to this pull request.

tristanls:
**note**: Infrastructure is recovering. Once recovered, I'll launch the benchmarks and report the results as initially intended.

PR #376, opened 2025-07-15
RFC 12: Extended Goal State Functionality

scottcanoe:
## Summary
This RFC proposes changes to Monty’s architecture aimed at developing salience-based methods to improve saccade efficiency and to [support compositional and multi-object configurations](https://thousandbrainsproject.readme.io/docs/implement-efficient-saccades-driven-by-model-free-and-model-based-signals).

This RFC is an outgrowth of [a previous RFC](https://github.com/thousandbrainsproject/tbp.monty/pull/328). For reviewers familiar with that one, know that this one is short and targeted. For new/future reviewers, see that RFC for additional context, especially with respect to the improved saccade efficiency and supporting compositional and multi-object configurations.

For easier reading, see the [rendered markdown on my fork](https://github.com/scottcanoe/tbp.monty/blob/extended_goal_state_functionality/rfcs/extended_goal_state_functionality.md).


rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,59 @@
...
+
+All of the above changes are meant to mirror existing learning-module/GSG behavior.
+
+### `Monty`
+ - Will have a `gss` attribute to hold a `GoalStateSelector`.
+ - Existing machinery that collects goal states, performs selection, and updates the motor system will be replaced. More specifically, `Monty` will collect goal states from both learning modules and sensor modules, but it will defer goal state selection to its `gss` object. The output of the `gss` object will then be used to update the motor system.
+ - Optionally, `Monty` may be equipped with additional `lm_to_gss` and `sm_to_gss` connectivity matrices to support more complex routing pathways, but this is not an immediate need.
+
+### `MotorSystem`
+ - Will be initialized with any number of policies.
```
nielsleadholm:
```suggestion
 - Will be initialized with any number of policies. As before, these will all be model-free.
```

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,59 @@
...
+ - Will perform policy/action selection based on a given goal state. Note that a goal state may not always be set, in which case it is expected that an appropriate model-free policy will be selected to produce an action.
+
+### `GoalState`
+ - May be augmented with metadata indicating the target of the goal, such as the agent's position or a sensor's sensed location.
+
+Specific modifications to `GoalState` objects are not yet well understood. However, we expect that additional metadata may be required for the motor system to make correct judgments about which policy to select. For example, implementing behavior like `GetGoodView` will require some way to distinguish between goal states that specify an agent's translation and goal states specifying an orienting motion.
```
nielsleadholm: **suggestion:** Add something like the following:

However, adding new metadata in the form of non-morphological features will be limited as much as possible. For example, the concept of an "object ID" in a goal-state could be used to specify that the goal-state is for the agent itself, or for a sensed location. Just as a goal-state might specify "coffee machine (object ID) at location xyz to be ON (behavioral state)", here the goal state is "agent to be at location xyz" or "sensed observation to be at location xyz".

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,59 @@
...
+
+While the proposed modifications to sensor modules are meant to mirror the the goal-state handling implemented for learning modules, there is one annoying difference. As noted, the sensor module's `propose_goal_states` method is *pluralized*; learning modules have a `propose_goal_state` method which returns only one goal state, whereas the sensor module's `propose_goal_states` method produces any number of goal states. Ideally, SM and LM interfaces would be identifical in this respect. Since returning many goal states is a strict requirement for future work, aligning the two interfaces would entail modifying the learning module's interface to look like the sensor module's new pluralized interface. This detail may be too small to mention here, but I wanted to note it here to express my intention to build without introducing bothersome inconsistencies.
```
nielsleadholm: **thought:** It doesn't seem like it would be prohibitively complex in the code to just have these behaviors be different, such that the GSGs for SMs and LMs are slightly different in this method. But I agree it would be nice if we could have a name that captures both. I think just keeping `propose_goal_state`, and allowing this to be a list for SMs could be sufficient.

One other difference I anticipate - my current thinking is that SM-GSGs will be driven by innate (in biology, evolutionary) driving goals, which are not modified by any external force. It is up to the cortex/LMs _via the GSS_ to determine whether these kinds of model-free goal-states are acted on or ignored. This means that the `driving_goal_state` of an SM-GSG can only be modified by internal logic within the SM. This differs from LM-GSGs, which are driven by higher level goal-states, including human experimenters.

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,59 @@
...
+All of the above changes are meant to mirror existing learning-module/GSG behavior.
+
+### `Monty`
+ - Will have a `gss` attribute to hold a `GoalStateSelector`.
+ - Existing machinery that collects goal states, performs selection, and updates the motor system will be replaced. More specifically, `Monty` will collect goal states from both learning modules and sensor modules, but it will defer goal state selection to its `gss` object. The output of the `gss` object will then be used to update the motor system.
+ - Optionally, `Monty` may be equipped with additional `lm_to_gss` and `sm_to_gss` connectivity matrices to support more complex routing pathways, but this is not an immediate need.
+
+### `MotorSystem`
+ - Will be initialized with any number of policies.
+ - Will perform policy/action selection based on a given goal state. Note that a goal state may not always be set, in which case it is expected that an appropriate model-free policy will be selected to produce an action.
```
nielsleadholm:
```suggestion
 - Will perform policy/action selection based on a given goal state. Note that a goal state may not always be set, in which case it is expected that an appropriate model-free policy will be selected to produce an action. For example, when goal-states from `GetGoodView` are not used, the motor system can randomly saccade.
```

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,59 @@
...
+### `SensorModule`
+ - Will have an optional `gsg` attribute (i.e., it will be a `GoalStateGenerator` or `None`).
+ - Will call its `gsg`'s `step_gsg` method before exiting its own `step` method (if `gsg` is not `None`).
+ - Will have a `propose_goal_state` method that returns any number of goal states. Note that this differs from the learning module's `propose_goal_state` method which returns at most 1 goal state. Returing many goal states will enable planned work on salience-based targeting methods.
+
+All of the above changes are meant to mirror existing learning-module/GSG behavior.
+
+### `Monty`
+ - Will have a `gss` attribute to hold a `GoalStateSelector`.
+ - Existing machinery that collects goal states, performs selection, and updates the motor system will be replaced. More specifically, `Monty` will collect goal states from both learning modules and sensor modules, but it will defer goal state selection to its `gss` object. The output of the `gss` object will then be used to update the motor system.
```
nielsleadholm:
```suggestion
 - Existing machinery that collects goal states, performs selection, and updates the motor system will be replaced. More specifically, `Monty` will collect goal states from both learning modules and sensor modules, but it will defer goal state selection to its `gss` object. The output of the `gss` object will then be used to update the motor system. See the for-loop associated with `best_goal_confidence` in `_pass_infos_to_motor_system` for how this is currently implemented.
```

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,59 @@
...
+
+  1. Implement the functionality of `GetGoodView` for the distant agent within the proposed framework. Goal-state-generating SMs will provide the model-free signals used to orient the agent to on-object regions. These signals can be filtered out or overridden when learning modules emit goal states, and the filtering will be performed by a `GoalStateSelector`.
+  2. Integration of model-free segmentation and model-based policies to recognize an object before moving on to a new object. This objective is the primary motive that led to the formation of this specific RFC.
+  3. Improve efficiency by learning and using sparser models. It has been suggested that our learned models are unnecessarily dense owing to the fact that non-salient and salient parts of an object are represented equally. If we can ensure that only the most salient parts of an object are attended to during inference, then we can effectively ignore (or at least sample less frequently) the non-salient areas during training without degrading performance.
+
+Each of these tasks is significant, likely warranting its own RFC. Further discussion is outside the scope of this RFC.
+
+# Discussion
```
nielsleadholm: **suggestion:** Maybe add something like the following:

More complex agents in the future might have multiple Goal State Selectors, and these could each output a goal-state to different parts of the motor-system. This is non-trivial however, and not something necessary at the moment. The analogy to the body is that we can sometimes do multiple things at once (e.g. walking or driving while talking), but this is often challenging, especially if similar parts of the motor system are involved (e.g. brushing your teeth side-to-side while rubbing your stomach in a circular motion). For now, there will be one Goal-State Selector for the entire Monty system.

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,59 @@
...
+
+The changes proposed here may disrupt the existing goal-state-execution system. More specifically, the dataloader is currently responsible for checking whether a motor system has a valid goal state and attempting to satisfy the goal state by "jumping" to a new location. I strongly suspect that the best course of action is to move the dataloader's jump-execution behavior into the motor system where it can be triggered by a policy. While the details are not yet fully defined, this refactor does appear to be consistent with already planned changes to the dataloader.
```
nielsleadholm: **suggestion:** Another discussion point that could be worth documenting. It might also be worth adding sub-headers for these different discussion points.

We are still determining the best way to verify when a goal-state has been achieved. The most intuitive approach is for the system (SM or LM) generating the goal-state to receive a sensory input that confirms the goal state was achieved. It is not clear however if this sensory input will always be formulated so as to exactly match the output goal state, or if it is up to the SM/LM to learn/have logic that maps between output goal states and sensory predictions. For example, the output goal-state for filling most of the the visual field with an object (move agent to xyz) is quite different from the sensory prediction of its success (percentage of visual field filled by object passes a threshold). In addition, some LMs, SMs, or the motor system/motor modules might receive proprioceptive input that could confirm the satisfaction of certain goal-states (e.g. agent is location xyz).

Formulating goal-states as "displacements" more than "locations" might help with this, but this would be a more significant change.

Currently, the Hypothesis-Testing GSG knows that a successful goal-state means the sensory perception will change. For now, we will continue the approach that the GSG must be able to map from an output goal state to the sensory prediction that equates to success. This can be achieved either through learning or some innate knowledge, where the latter will be our primary approach for now.

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,59 @@
...
+### `GoalState`
+ - May be augmented with metadata indicating the target of the goal, such as the agent's position or a sensor's sensed location.
+
+Specific modifications to `GoalState` objects are not yet well understood. However, we expect that additional metadata may be required for the motor system to make correct judgments about which policy to select. For example, implementing behavior like `GetGoodView` will require some way to distinguish between goal states that specify an agent's translation and goal states specifying an orienting motion.
+
+# Long-Term Objectives
+
+Following discussions prior to this RFC, we have identified three objectives related to the functional and architectural changes proposed here.
+
+  1. Implement the functionality of `GetGoodView` for the distant agent within the proposed framework. Goal-state-generating SMs will provide the model-free signals used to orient the agent to on-object regions. These signals can be filtered out or overridden when learning modules emit goal states, and the filtering will be performed by a `GoalStateSelector`.
```
nielsleadholm:
```suggestion
  1. Implement the functionality of `GetGoodView` for the distant agent within the proposed framework. Goal-state-generating SMs will provide the model-free signals used to orient the agent to on-object regions. These signals can be filtered out or overridden when learning modules emit goal states, and the filtering will be performed by a `GoalStateSelector`. Note this `GetGoodView` will be an inherent part of Monty, and can be called, for example, after a `JumpToGoalState` causes Monty to lose the object. This is thus distinct from the version of `GetGoodView` being extracted as a `PositioningProcedure`. Positioning Procedures cannot be called by Monty, and happen outside of the inference or learning episode. 
```

nielsleadholm (commented):
This looks great! Really nice and clean. Just a few minor suggestions, but I think it's ready for review by the rest of the team.

nielsleadholm:
**nitpick:** The arrow in the diagram going from the motor system to the data-loader should be pink.

rfcs/extended_goal_state_functionality.md
scottcanoe: To be clear, are you suggesting we could encode the "subject" of a goal state with a non-morphological feature such as "object ID"?

If so, I'm starting to think we might need something a bit more privileged than that. In particular, breaking the assumption that a goal state means "sensed location" will require an additional mechanism that determines how a goal state should compared with the post-action state of the system. It's unclear to me whether the same 1-2 additional pieces of metadata will be enough to bind a goal state with its target, equality-testing method, etc.

rfcs/extended_goal_state_functionality.md
scottcanoe: We're on the same page. I started implementing this yesterday as went with `propose_goal_state` for everything.

Interesting thoughts on how driving goal states might differ in these two cases. But while we're on the topic, I was wondering yesterday about whether we'd ever want or allow multiple driving goal states. If so, the goal state selector could be a `GoalStateGenerator`.

rfcs/extended_goal_state_functionality.md
scottcanoe: Excellent suggestion. The verification step is tricky. Part of that probably involves notifying a goal state that it has been attempted so that not all goal states are checked during `GoalStateGenerator.step_gsg`. For what it's worth, it appears that this is already an issue for multi-LM experiments when several LMs produce a goal state simultaneously. If anything, it probably just inflates the number of goal states not achieved. But even with 16-LM experiments, I bet simultaneously emitted goal states have probably pretty infrequent thus far.

rfcs/extended_goal_state_functionality.md
nielsleadholm: Ok interesting, in my head the goal-state will drive the model-free policy in the motor system. These policies will know what kind of goal-states they can enact, like changing the location of the agent, changing the location that is sensed, or (eventually) exerting a force at a location. Thus they will only accept a goal-state that they can actually carry out, i.e. match to based on this "object" ID. They then will enact a sub-routine / policy that operates until that goal-state is satisfied. In this case I don't see an issue with the goal-state both determining the target, as well as equality-testing, but I may be oversimplifying it - might be helpful to discuss tomorrow. 

For what it's worth, I like how you have it worded in the RFC now.

rfcs/extended_goal_state_functionality.md
nielsleadholm: Yeah interesting question, I think that's something to think about, even if we don't need to implement the GSS that way for now.

rfcs/extended_goal_state_functionality.md
nielsleadholm: Yeah I agree this is something to think about for the longer term. Right now I'm thinking that if the mapping between goal-states and sensory observations is actually learned, then although there will be many cases where a goal-state does not get enacted, at the limit, the LM will notice a pattern between certain output goal states, and expected observations. Unless an LM has very high confidence in its output goal-state, it might then notice the discrepancy when it's goal-state isn't enacted, but generally won't "care". But yeah maybe there is an efferent feedback signal to help inform these expectations further.

rfcs/extended_goal_state_functionality.md
```diff
@@ -30,30 +30,43 @@ All of the above changes are meant to mirror existing learning-module/GSG behavi
...
+  1. The sensor module's GSG may have produced any number of goal states on the previous step. We need some way to indicate which goal state (if any) was selected and attempted. Note that multi-LM systems can also produce multiple goal states, so there is already a need to implement this behavior.
+  2. The observations delivered to the GSG's owing LM or SM may not necessarily contain the information needed to check whether a goal state was achieved. For example, an SM may generate a goal state targeting the agent's position. In this case a goal-state-generator may need access to proprioceptive data for computing goal-state-success. We can also imagine a goal might be to have a minimum percentage of the visual field occuppied an object. In this case, success can only be computed with raw sensor data which isn't available to the GSG (by default).
```
nielsleadholm:
```suggestion
  2. The observations delivered to the GSG's own LM or SM may not necessarily contain the information needed to check whether a goal state was achieved. For example, an SM may generate a goal state targeting the agent's position. In this case a goal-state-generator may need access to proprioceptive data for computing goal-state-success. We can also imagine a goal might be to have a minimum percentage of the visual field occuppied an object. In this case, success can only be computed with raw sensor data which isn't available to the GSG (by default).
```

nielsleadholm (commented):
Looks great, thanks for those changes! Motion to move this to non-draft status.

rfcs/extended_goal_state_functionality.md
scottcanoe: For policy selection, I wondered about giving policies a method like `can_handle(goal_state)`. It'd be a simple way for the motor system to find a policy without it needing to know anything about the particular policies it owns. It'd be a pretty self-contained way to do it, but I can imagine smarter or more dynamic ways to associate policies with goal states.

To determine whether a goal state has been achieved, which includes equality testing, it makes sense to me that we delegate this to the goal state itself as you suggested (as opposed to the GSG that created it). But who performs the check is maybe less of an issue than the overall problem of needing to supply the testing method with anything it needs.

Here's a funny example. I've only reasoned about this and not checked it, so let me know if you think this isn't how it works. Imagine a multi-SM, multi-LM system arranged like normal. So we have a "central" SM/LM and a bunch of "peripheral" SMs/LMs. Then...
- A peripheral LM emits a goal state to observe location "xyz".
- The motor system successfully accomplishes this goal state by putting that location in the **central** SMs field-of-view.
- The peripheral LM calls its GSG to see if the goal state has been accomplished. Since the peripheral SM is not observing this location, it will be registered as a failure.

This is a consequence of that fact that
1.  The policy always tries to observe location xyz with the central SM. (FWIW, I think this is a good thing.)
2. The goal state's success check is strongly coupled to its parent LM's data.

This is a related problem to success-checking a goal state that requires agent position. In both cases, the success-checker needs access to data it doesn't have. One idea I had was to attach extractor and a success-checker functions to a goal state. So it's look something like:
```python
# Method of `GoalState`
def achieved(self, all_of_the_data):
    expected = self.extract_prediction(self.output_goal_state)
    actual = self.extract_actual(all_of_the_data)
    return self.equivalent(expected, actual)
```
where `extract_predict`, `extract_actual`, and `equivalent` can be set dynamically or by the generator. The goal state could, in principle, be success-checked anywhere or by anyone, assuming they can supply it with the data required by `extract_actual`. I'm not sure if this is a good thing or a bad thing.

nielsleadholm: 1. Yeah I like the proposal for a `can_handle` method.

2. That's a great example to bring up, I agree it's tricky. The `achieved` method being bundled with the goal-state is interesting, although right now I'm struggling to think through how this would work in the brain. It feels like it is moving towards "non-locality" of information, with the potential long-term pitfalls that could cause. 

With the example of the peripheral LM, I'm not 100% sure, but a few thoughts:
- One way to maybe think about it is this peripheral LM basically knows / has learned that it is a peripheral, low-resolution LM.
- When it outputs the goal-state for sensing at a location, it is not expecting to be sensing at that location (in fact it already is); from its perspective, the goal-state is successful if it moves such that it *no longer* is sensing.
- There would probably be another level of goal-state checking (corresponding more to motor cortex or sub-cortical structures), that we are sensing where desired.
- This example is another demonstration that the success of the goal-state is not an exact match to the output goal-state, it needs to be learned in a way that will be specific for particular LMs. 

How might this be learned in an unsupervised way?
- Imagine we have a high-level goal-state of "reduce uncertainty/recognize what we are observing"
- The peripheral LM has really low resolution and, along with its neighbors, is really bad at this, so it doesn't receive any positive feedback by trying to be observing the object itself.
- However, when it output a goal-state to move sensors to its current location when an interesting feature occurred, then that was associated with the system (set of foveal LMs voting etc.) recognizing the object, and the high-level goal state was achieved.
- With enough reinforcement, peripheral LMs will realize their job is to move primary sensors on to interesting information, not to try to study them on their own. 
- This in turn will inform what input they are expecting as evidence that their goal-state was achieved (moved off location); it will be imperfect / only correlate with success, but that this is the cost of relying on locally available information.
- Other parts of the system (motor regions etc.) can use the local information available to them to also check for the success of the goal-state, and these checks might be more direct.

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,72 @@
...
+
+### `SensorModule`
+ - Will have an optional `gsg` attribute (i.e., it will be a `GoalStateGenerator` or `None`).
+ - Will call its `gsg`'s `step_gsg` method before exiting its own `step` method (if `gsg` is not `None`).
+ - Will have a `propose_goal_state` method that returns any number of goal states. Note that this differs from the learning module's `propose_goal_state` method which returns at most 1 goal state. Returing many goal states will enable planned work on salience-based targeting methods.
+
+All of the above changes are meant to mirror existing learning-module/GSG behavior.
+
+### `Monty`
+ - Will have a `gss` attribute to hold a `GoalStateSelector`.
```
vkakerbeck: minor note: It might be useful to spell these out to avoid confusion. I don't see a downside in calling these variables `goal_state_selector` and `goal_state_generator` for more clarity

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,72 @@
...
+ - Existing machinery that collects goal states, performs selection, and updates the motor system will be replaced. More specifically, `Monty` will collect goal states from both learning modules and sensor modules, but it will defer goal state selection to its `gss` object. The output of the `gss` object will then be used to update the motor system. See the for-loop associated with `best_goal_confidence` in `_pass_infos_to_motor_system` for how this is currently implemented.
+ - Optionally, `Monty` may be equipped with additional `lm_to_gss` and `sm_to_gss` connectivity matrices to support more complex routing pathways, but this is not an immediate need.
```
vkakerbeck: This would only be nescessary if we have multiple motor systems (like two agents that can move independently and achieve two goals at the same time). This is still further out in the future so I would not think about adding this right now.

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,72 @@
...
+
+### `MotorSystem`
+ - Will be initialized with any number of policies. As before, these will all be model-free.
+ - Will perform policy/action selection based on a given goal state. Note that a goal state may not always be set, in which case it is expected that an appropriate model-free policy will be selected to produce an action. For example, when goal-states from `GetGoodView` are not used, the motor system can randomly saccade.
+
+### `GoalState`
+ - May be augmented with metadata indicating the target of the goal, such as the agent's position or a sensor's sensed location.
```
vkakerbeck: Another potential extention of the goal state we talked about is that it may specify an area instead of a point in space. This could be interpreted as "observe somewhere in this rough area" (uncertainty) or "observe this area" (implicitly telling agent to move closer/further away from object)

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,72 @@
...
+  1. The sensor module's GSG may have produced any number of goal states on the previous step. We need some way to indicate which goal state (if any) was selected and attempted. Note that multi-LM systems can also produce multiple goal states, so there is already a need to implement this behavior.
+  2. The observations delivered to the GSG's owning/parent LM or SM may not necessarily contain the information needed to check whether a goal state was achieved. For example, an SM may generate a goal state targeting the agent's position. In this case a goal-state-generator may need access to proprioceptive data for computing goal-state-success. We can also imagine a goal might be to have a minimum percentage of the visual field occuppied an object. In this case, success can only be computed with raw sensor data which isn't available to the GSG (by default).
```
vkakerbeck: Isn't all this information available in the SM (this is what extracts the features like percent on obj. and combines them with proprioceptive information)?

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Discussion
+
+### Dataloader
+The changes proposed here may disrupt the existing goal-state-execution system. More specifically, the dataloader is currently responsible for checking whether a motor system has a valid goal state. If so, it attempts to satisfy the goal state by "jumping" to a new location. In the proposed system, goal states will no longer be coupled to a single execution method, and all conversions between goal states to actions will be performed within the motor system. Consequently, the dataloader's "jump-to-goal-state" code will need to live elsewhere, preferably somewhere it can be reused by different polices. Fortunately, removing any reliance on the dataloader to execute goal states is consistent with existing plans.
+
+### Computing Goal-State Success
+We are still determining the best way to verify when a goal state has been achieved. As it stands, each `GoalStateGenerator` performs this task by reaching into its parent LM, grabbing its input observations, and comparing them with the previously generated goal state. Setting aside parent-access, reimplementing this pattern for goal-state-generating sensor modules won't work for a couple of reasons:
```
vkakerbeck: What would this information be used for? Are you thinking of eventually having GSGs that learn by themselves what goalstates to output and use feedback about success to adjust this? It seems like that would be nice in the future, but for the current GSGs, we are planning on it is not necessary.

rfcs/extended_goal_state_functionality.md
ramyamounir: I think I made this [comment](https://github.com/thousandbrainsproject/tbp.monty/pull/328#discussion_r2153303697) on Scott's previous RFC because I noticed we already have parts of the code supporting multiple agents. Now after reading this RFC, I think it may also be useful for the class `Monty` to know which SMs are connected to the `gss` by looking at the `sm_to_gss` matrix. Not sure how Scott plans on implementing it but I was thinking it could be done in a loop over the SMs that are wired into the gss, just like how we collect votes or sensory observations.

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,72 @@
...
+
+Each of these tasks is significant, likely warranting its own RFC. Further discussion is outside the scope of this RFC.
+
+# Discussion
+
+### Dataloader
+The changes proposed here may disrupt the existing goal-state-execution system. More specifically, the dataloader is currently responsible for checking whether a motor system has a valid goal state. If so, it attempts to satisfy the goal state by "jumping" to a new location. In the proposed system, goal states will no longer be coupled to a single execution method, and all conversions between goal states to actions will be performed within the motor system. Consequently, the dataloader's "jump-to-goal-state" code will need to live elsewhere, preferably somewhere it can be reused by different polices. Fortunately, removing any reliance on the dataloader to execute goal states is consistent with existing plans.
+
+### Computing Goal-State Success
```
ramyamounir: Perhaps related to this, I was thinking whether the saliency SM will continue issuing goal states while the Motor System is still trying to achieve a previously selected Goal State, or will the SM stop generating goal states until the previous goal state is achieved? It could be interesting to think about the implications of a better goal state that has been generated while the previous one is still in progress. If there is a global ranking system for these goal states, then maybe the GSS can instruct the MS to change plans and use the better goal state even though the previous one has not yet been reached.

ramyamounir (approved):
Very nicely written! I like how you went into details about the anticipated complications, it's clearly well thought.

rfcs/extended_goal_state_functionality.md
scottcanoe: Great thoughts, it's a lot to take in. I have a feeling we'll be returning to this discussion a fair bit as things develop, so I'm going to start collecting these notes somewhere for future reference.

One part I zeroed in on was this: "The peripheral LM has really low resolution and, along with its neighbors, is really bad at this, so it doesn't receive any positive feedback by trying to be observing the object itself." This is interesting to think about in terms of how distinguishable object models ought to be for peripheral LMs, and how its goal state generators should be tuned. For example, a peripheral LM's "fork" and "spoon" object models might be indistinguishable, but they'd have different labels when supervisory signals are involved. In that case, a standard GSG probably wouldn't be useful, but I'd have to think about that more.

rfcs/extended_goal_state_functionality.md
scottcanoe: Agreed. I can delete this or leave it for posterity, but adding more complex routing pathways for a goal state selector would really be a footnote in a larger refactor involving multiple agents, agent <-> motor_system associations, etc.

rfcs/extended_goal_state_functionality.md
scottcanoe: Good point, I've added a bit of text to include this.

rfcs/extended_goal_state_functionality.md
scottcanoe: True. SM's don't (currently) hold on to a copy of its raw sensory data, but that's not a problem so long as the SM calls its GSG during `step`.

rfcs/extended_goal_state_functionality.md
scottcanoe: Sorry, not sure which bit of text you're referencing. But if you're talking about information like proprioceptive state, then the idea behind that was to be able to emit a goal state like "have the agent be at this location", in which case the sensor's data processed observation might not be sufficient to test whether that goal has been achieved. I was speculating about ideas further down the road when I wrote this, but I've mentally backtracked on this approach since this draft.

rfcs/extended_goal_state_functionality.md
scottcanoe: This is a million-dollar question, and I think the answer is "it depends". In general, I'm not really sure how multi-step actions and goals are going to be handled. At a minimum, goal states from the SM can always be ignored if the motor system is busy, but the problem here is that GSGs are the only place where goal states persist across steps. If we let GSGs continue to produce goal states while a previous goal state is in process, then the only reference to the goal state being attempted will be lost, and it'll therefore be untestable for success when its attempt is complete.

On the other hand, I'm not really sure how important it is to check an SM's goal states for success. Model-free goals might be more "fire and forget" than model-based ones.

scottcanoe: And as for SM-GSS connectivity, I think that'll happen naturally whenever we start adding more agents, and it'll be part of a wider conversation. To me, it makes sense that each agent would have its own motor system, in which case we might come up with a slightly different way to specify agent-wise connections between SMs, goal-state-selectors, motor systems, (though I'd think LM's are agent-independent).

rfcs/extended_goal_state_functionality.md
vkakerbeck: Yeah I was questioning what this information would be used for? Like how would it change the goal state generator's output if the goal state was achieved or not achieved? It seems like unless the GSG is actively learning and updating it's policy based on feedback (which we are not planning on for these first model-free policies) it would have no effect.

nielsleadholm: Yeah maybe just to add some clarification, the plan with the SM-GSG for GetGoodView was to have a bespoke, custom-made way of checking for goal-state success based on sensory observations (e.g. percentage of visual field filled). It would not be based on learning. Does that address your question @vkakerbeck ? If that's not clear in the text, maybe we can update part of it.

rfcs/extended_goal_state_functionality.md
hlee9212: I second this motion. 😄 (I know `lm_to_gss` will get long like ` lm_to_goal_state_selector`, but I think I still prefer being written out).

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,72 @@
...
+  2. Integration of model-free segmentation and model-based policies to recognize an object before moving on to a new object. This objective is the primary motive that led to the formation of this specific RFC.
+  3. Improve efficiency by learning and using sparser models. It has been suggested that our learned models are unnecessarily dense owing to the fact that non-salient and salient parts of an object are represented equally. If we can ensure that only the most salient parts of an object are attended to during inference, then we can effectively ignore (or at least sample less frequently) the non-salient areas during training without degrading performance.
```
jeremyshoemaker: This is a question more for my learning than anything else.

How do we determine salience? Like especially with a new unknown object, how would we know which parts are salient and which parts are non-salient _a priori_?

jeremyshoemaker (commented):
Sounds reasonable to me, but I'll leave it to others to approve or not.

rfcs/extended_goal_state_functionality.md
hlee9212: There are traditional (non-deep learning) methods to detect saliency, implemented in OpenCV. See [documentation](https://docs.opencv.org/3.4.2/d8/d65/group__saliency.html) here. (Not sure if these are what Scott will use but there are definitely some developed algorithms 😄 )

rfcs/extended_goal_state_functionality.md
vkakerbeck: Hm okay, I was thinking of this happening independent of which goal state was issued. I.e. if a sufficient a mount of the visual field is filled with the object, the next goal state issued will not be about moving toward the object anymore but it doesn't matter what the previous goal state was. What's the reason we need to consider the past goal state for informing the next one instead of just the current visual input?

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,72 @@
...
+
+The `GoalStateSelector` will receive any number of goal states derived from learning and sensor modules, and it will produce at most one goal state for consumption by the motor system. There are many potential strategies for choosing a goal state, but the simplest initial approach will be to select the goal state with the largest `confidence` value. In this way, it will act exactly as the `Monty` class currently does. More complex methods are outside the scope of this RFC.
+
+Many other Monty components will require modification. Below is a non-exhaustive list of anticipated changes.
+
+### `SensorModule`
+ - Will have an optional `gsg` attribute (i.e., it will be a `GoalStateGenerator` or `None`).
+ - Will call its `gsg`'s `step_gsg` method before exiting its own `step` method (if `gsg` is not `None`).
+ - Will have a `propose_goal_state` method that returns any number of goal states. Note that this differs from the learning module's `propose_goal_state` method which returns at most 1 goal state. Returing many goal states will enable planned work on salience-based targeting methods.
```
tristanls: **note**: This will work itself out in the pull requests, but I want to highlight the code invoking and expecting this method as a possible source of accidental coupling. This will be one aspect of design/implementation to review closely.

rfcs/extended_goal_state_functionality.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Long-Term Objectives
+
+Following discussions prior to this RFC, we have identified three objectives related to the functional and architectural changes proposed here.
+
+  1. Implement the functionality of `GetGoodView` for the distant agent within the proposed framework. Goal-state-generating SMs will provide the model-free signals used to orient the agent to on-object regions. These signals can be filtered out or overridden when learning modules emit goal states, and the filtering will be performed by a `GoalStateSelector`. Note this `GetGoodView` will be an inherent part of Monty, and can be called, for example, after a `JumpToGoalState` causes Monty to lose the object. This is thus distinct from the version of `GetGoodView` being extracted as a `PositioningProcedure`. Positioning Procedures cannot be called by Monty, and happen outside of the inference or learning episode.
```
tristanls: **nitpick**: I think we can call it `GetGoodView` positioning procedure now. It's been extracted already. https://github.com/thousandbrainsproject/tbp.monty/blob/9677cc918adeca9ae21233d957c0401e84f482ab/src/tbp/monty/frameworks/models/motor_policies.py#L443

**suggestion**: Call the new one: `GetGoodView` policy.

rfcs/extended_goal_state_functionality.md
nielsleadholm: Yeah I think that's an interesting question. In the long-term it could be useful for a GSG to explicitly check for whether a goal-state is successful or not after it outputs a goal state, based on the next sensory inputs. This would enable it to learn what goal-states are correlated with changes in the world. I agree with you that with the current setup, just checking the conditions for whether a goal-state should be generated would be functionality equivalent. 

For what it's worth, we already have a `_check_output_goal_state_achieved` method, so this wouldn't be adding anything.

nielsleadholm (approved):
Have previously reviewed while it was draft but just adding my approve, looks great!

scottcanoe:
# Motion For Final Comment Period (FCP)

Hear ye, hear ye! We will now hear final comments and tally votes from maintainers. Votes may be cast with a comment or with an emoji reaction to this announcement.

Thumbs up = MERGE
Thumbs down = NO MERGE
Eyes = ABSTAIN

Thank you all for taking the time and energy to provide excellent, thoughtful feedback throughout this process.

@vkakerbeck @nielsleadholm @tristanls @codeallthethingz @hlee9212 @jeremyshoemaker @ramyamounir

rfcs/extended_goal_state_functionality.md
scottcanoe: In biology, things like color opponency, higher (spatial) frequency areas, and motion contribute to salience. Basically, anything that makes some part of the visual field stand out. We can try and all methods in combination and see what shakes out. As @hlee9212 mentioned, there are some ready-to-go algorithms, but I plan on having a bit of fun with it and see if we can't come up with something that's simple and works well for our purposes.

PR #379, opened 2025-07-17
refactor: rename `step_gsg` to `step`

scottcanoe:
This PR renames the `GoalStateGenerator`'s `step_gsg` method to `step`. This change is intended to improving naming consistency.

### DIFF
--- a/src/tbp/monty/frameworks/models/abstract_monty_classes.py
+++ b/src/tbp/monty/frameworks/models/abstract_monty_classes.py
@@ -281,7 +281,7 @@ class GoalStateGenerator(metaclass=abc.ABCMeta):
         pass
 
     @abc.abstractmethod
-    def step_gsg(self):
+    def step(self):
         """Called on each step of the LM to which the GSG belongs."""
         pass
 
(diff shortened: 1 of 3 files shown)


src/tbp/monty/frameworks/models/abstract_monty_classes.py
```diff
@@ -281,7 +281,7 @@ def get_output_goal_state(self):
         pass
 
     @abc.abstractmethod
-    def step_gsg(self):
+    def step(self):
```
tristanls, line 284: **note**: It appears that this abstract method definition is incorrect, in that every GSG implementing it actually implements `def step(self, observations):`

tristanls:
@scottcanoe I updated the name to `refactor!`, wanted to do it real quick in case you hit the merge button really fast. But, now taking the time to comment 🙂. Since this is a public API change, I think it is a breaking change that should be marked with `!`.

src/tbp/monty/frameworks/models/abstract_monty_classes.py
scottcanoe: True. This guy isn't the only one. I've noticed that a handful of abstract methods don't have parameters though their implementations do. I'll update these as I find them.

PR #381, opened 2025-07-17
refactor!: Remove SensorModuleBase

scottcanoe:
## Summary
This PR addresses a few issues with the sensor module hierarchy.

- `SensorModule` inherits from `Callable`, but sensor modules aren't called.
    **Fix**: Don't inherit from `Callable`.

- `SensorModuleBase` is essentially a fake. Its implementations are dummies, so Inheriting from `SensorModuleBase` defeats the purpose of having abstract methods.
    **Fix 1**: Move `SensorModuleBase` into test code as `FakeSensorModule`. Use it in place of `SensorModuleBase` for tests.
**Fix 2**: Have `DetailedLoggingSM` inherit from `SensorModule` instead of `SensorModuleBase`. Add a couple of required attributes and methods that were handled by `SensorModuleBase`.

- `set_experiment_mode` is a de facto requirement of any sensor module class since Monty always calls it, but it isn't enforced or otherwise notated as required.
  **Fix**: Add an abstract method to `SensorModule`.


src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -15,7 +15,9 @@
 from scipy.spatial.transform import Rotation
 from skimage.color import rgb2hsv
 
-from tbp.monty.frameworks.models.monty_base import SensorModuleBase
+from tbp.monty.frameworks.models.abstract_monty_classes import SensorModule
+
+# from tbp.monty.frameworks.models.motor_system_state import AgentState
```
tristanls: **todo**: Remove commented out code.

src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -56,7 +58,8 @@ def __init__(
             weight_curvature: determines whether to use the "weighted" (True) or
                 "unweighted" (False) implementation for principal curvature extraction.
         """
-        super(DetailedLoggingSM, self).__init__(sensor_module_id)
```
tristanls, line 59: **issue**: Always opt into cooperative multiple inheritance, just in case, via:
```python
super().__init__()
```

This is the thing in our draft Project Code Organization Guidance under the section "You SHOULD always include call to `super().__init__()` in your `__init__` methods".

src/tbp/monty/frameworks/models/abstract_monty_classes.py
```diff
@@ -295,6 +294,10 @@ def state_dict(self):
         """
         pass
 
+    @abc.abstractmethod
+    def update_state(self, state):
+        pass
+
```
tristanls, line 300: **question (non-blocking)**: Just curious why move stuff around? 🙂

tristanls (approved):
Forgot to click the approve button 😬. Looks good, thank you for these changes.

tristanls:
@scottcanoe, I clicked approve, but forgot I made one **issue** that should be addressed, regarding the `super().__init__()` call. Also, please do the **todo** before merging.

tristanls:
@scottcanoe, also, since we are deleting a class, I think this should also be marked as a breaking change with `refactor!:` header.

src/tbp/monty/frameworks/models/sensor_modules.py
scottcanoe: Calling `super().__init__()` was problematic here. I dropped it after getting a bunch of errors like
```
FAILED tests/unit/graph_learning_test.py::GraphLearningTest::test_moving_off_object - TypeError: __init__() missing 1 required positional argument: 'noise_params'
```
I didn't think too hard about it, but it seemed like it must have been messing with the order of `__init__` calls.

I just tried `super(DetailedLoggingSM, self).__init__()` and got the same error, but `SensorModule.__init__(self)` works fine.

That's a bit of a bummer though. You never know whether subclasses will want to use mixins. I never could reason very well about `__init__` calls when multiple inheritance is involved, which is why I've always avoided it.

src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -56,7 +56,10 @@ def __init__(
...
+        SensorModule.__init__(self)
```
tristanls: **issue**: The goal is to opt into cooperative multiple inheritance in case someone does weird things with `DetailedLoggingSM` in the future. To do that, we should not ignore MRO and invoke `super().__init__()`.

Also, in modern Python we don't pass self to `super().__init__()` anymore. See code example in https://docs.python.org/3/library/functions.html#super:

```python
class C(B):
    def method(self, arg):
        super().method(arg)    # This does the same thing as:
                               # super(C, self).method(arg)
```

tristanls: **question**: Am I missing a specific use case you have in mind where `SensorModule.__init__(self)` is needed?

tristanls (changes requested):
Going to mark this as request for changes based on some recent commits for now.

src/tbp/monty/frameworks/models/sensor_modules.py
tristanls: Our comments passed each other. OK, I see. Let me take a look.

src/tbp/monty/frameworks/models/sensor_modules.py
tristanls: Resolving as this comment passed another comment and is no longer relevant.

scottcanoe: `super().__init__()` is always my go-to, but it results in 
```
FAILED tests/unit/graph_learning_test.py::GraphLearningTest::test_moving_off_object - TypeError: __init__() missing 1 required positional argument: 'noise_params'
```

src/tbp/monty/frameworks/models/sensor_modules.py
tristanls: @jeremyshoemaker can you advise here? `class HabitatDistantPatchSM(DetailedLoggingSM, NoiseMixin):` definition further below in the file along with its artisanal super invocations makes this complicated.

```python
        super(HabitatDistantPatchSM, self).__init__(
            sensor_module_id, save_raw_obs, pc1_is_pc2_threshold
        )
        NoiseMixin.__init__(self, noise_params)
```

src/tbp/monty/frameworks/models/abstract_monty_classes.py
scottcanoe: No good reason, really. It's a dumb story, but the TLDR is "mistake".

src/tbp/monty/frameworks/models/sensor_modules.py
tristanls: Also, `class FeatureChangeSM(HabitatDistantPatchSM, NoiseMixin):`, which means that `FeatureChangeSM` has `NoiseMixin` through inheritance via `HabitatDistancePatchSM` as well as directly mixing it in again.

tristanls: @scottcanoe, ok, I got it to work in a way that I think it gets us a bit closer to where we want to be:

```diff
diff --git a/src/tbp/monty/frameworks/models/sensor_modules.py b/src/tbp/monty/frameworks/models/sensor_modules.py
index 7204b64..9ade580 100644
--- a/src/tbp/monty/frameworks/models/sensor_modules.py
+++ b/src/tbp/monty/frameworks/models/sensor_modules.py
@@ -40,6 +40,7 @@ class DetailedLoggingSM(SensorModule):
         pc1_is_pc2_threshold=10,
         point_normal_method="TLS",
         weight_curvature=True,
+        **kwargs,
     ):
         """Initialize Sensor Module.
 
@@ -55,8 +56,9 @@ class DetailedLoggingSM(SensorModule):
                 value will raise an error.
             weight_curvature: determines whether to use the "weighted" (True) or
                 "unweighted" (False) implementation for principal curvature extraction.
+            **kwargs: Additional arguments to pass to the SensorModule base class.
         """
-        SensorModule.__init__(self)
+        super().__init__(**kwargs)
 
         self.sensor_module_id = sensor_module_id
         self.state = None
@@ -476,10 +478,12 @@ class HabitatDistantPatchSM(DetailedLoggingSM, NoiseMixin):
             gaussian_curvature and mean_curvature should be used together to contain
             the same information as principal_curvatures.
         """
-        super(HabitatDistantPatchSM, self).__init__(
-            sensor_module_id, save_raw_obs, pc1_is_pc2_threshold
+        super().__init__(
+            sensor_module_id,
+            save_raw_obs,
+            pc1_is_pc2_threshold,
+            noise_params=noise_params,
         )
-        NoiseMixin.__init__(self, noise_params)
         possible_features = [
             "on_object",
             "object_coverage",
@@ -591,7 +595,7 @@ class HabitatSurfacePatchSM(HabitatDistantPatchSM):
         self.on_object_obs_only = False  # parameter used in step() method
 
 
-class FeatureChangeSM(HabitatDistantPatchSM, NoiseMixin):
+class FeatureChangeSM(HabitatDistantPatchSM):
     """Sensor Module that turns Habitat camera obs into features at locations.
 
     Takes in camera rgba and depth input and calculates locations from this.
@@ -624,7 +628,7 @@ class FeatureChangeSM(HabitatDistantPatchSM, NoiseMixin):
                 False.
             noise_params: ?. Defaults to None.
         """
-        super(FeatureChangeSM, self).__init__(
+        super().__init__(
             sensor_module_id, features, save_raw_obs, noise_params=noise_params
         )
         self.delta_thresholds = delta_thresholds
```

tristanls: This turned out to be a good example of why we want to avoid stateful mixins and deep inheritance hierarchies.

scottcanoe: Agreed. Thanks for the changes, this worked. Feel free to approve, but I'm going to sanity check against a couple of benchmarks before merging.

tests/unit/frameworks/models/fakes/sensor_modules.py
```diff
@@ -0,0 +1,52 @@
...
+
+from tbp.monty.frameworks.models.abstract_monty_classes import SensorModule
+from tbp.monty.frameworks.models.states import State
+
+
+class FakeSensorModule(SensorModule):
+    """Dummy placeholder class used only for tests."""
+
+    def __init__(self, sensor_module_id: str):
+        self.sensor_module_id = sensor_module_id
```
tristanls, line 23: **todo**: One more `super().__init__()` here. I didn't catch this earlier.

tristanls (approved):
I think there's one more **todo**, but I think stuff looks good now. Thank you.

src/tbp/monty/frameworks/models/sensor_modules.py
jeremyshoemaker: > @jeremyshoemaker can you advise here?
> ...
> ```python
> super(HabitatDistantPatchSM, self).__init__(
>     sensor_module_id, save_raw_obs, pc1_is_pc2_threshold
> )
> NoiseMixin.__init__(self, noise_params)
> ```
That first version looks like the Python 2 version of a super call. I don't think it behaves differently in Python 3 but should probably be avoided.

I'm actually not sure what that second one will do. I wouldn't recommend it at all.

The changes you suggested in the diff above are the correct way to handle it. Often times if you have different constructors taking different arguments in the call hierarchy, you have to either pass in the expected arguments, or use `*args`, or `**kwargs` to pass through arguments.

Even in cases of diamond inheritance, you shouldn't ever have to call `super().__init__()` more than once. The calls through the parents get resolved in a fixed order and assuming all of them call super at some point, all of the constructors will end up being called.

scottcanoe:
@tristanls 

**question**: Your suggestion added `**kwargs` to `DetailedLogging.__init__` which is passed along via `super().__init__(**kwargs)`. However, `DetailedLoggingSM`'s base class (`SensorModule`) doesn't define an `__init__` method. Is this inclusion meant to future-proof in case `SensorModule.__init__` is defined later? Or is related to multiple inheritance? Looking for general guidance on whether this is a good idea in general. I wouldn't (and didn't) think to pass along `**kwargs` to a base class that doesn't define an `__init__` with default parameters or `**kwargs`.

tristanls:
@scottcanoe, if I recall correctly (@jeremyshoemaker might have more precise knowledge of this), the most general pattern is to always do:
```python
class MyClass:
    def __init__(self, *args, **kwargs):
```
This passes any positional and keyword arguments along to ancestors, no matter what type of inheritance structure is created.

For example, `super().__init__(**kwargs)` from `DetailedLoggingSM` passes `**kwargs` to `NoiseMixin`, when they are together in `class HabitatDistantPatchSM(DetailedLoggingSM, NoiseMixin)`.

I do not think this is a sustainable pattern. To make any changes, we have to understand every use and every hierarchy constructed, which rapidly becomes toilsome. If we get our code to the state advocated in #382, then there will only be one level of hierarchy, and mixins won't have any init arguments to worry about. This entire class of complexity will be gone, well.. at least much reduced.

PR #382, opened 2025-07-18
RFC: Code Organization Guidance

tristanls:
This proposal formalizes our approach to organizing `tbp.monty` code.

### DIFF
--- a/rfcs/0000_code_organization_guide.md
+++ b/rfcs/0000_code_organization_guide.md
@@ -1,0 +1,655 @@
+- Start Date: 2025-07-17
+- RFC PR: (leave this empty, it will be filled in after RFC is merged)
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14](https://www.rfc-editor.org/info/bcp14) [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) [RFC 8174](https://www.rfc-editor.org/rfc/rfc8174) when, and only when, they appear in all capitals, as shown here.
+
+This guidance does not dictate the only way to implement functionality. There are many ways to implement any particular functionality, each of which will work. This guidance _**establishes constraints**_ so that as more functionality is implemented and functionality is changed, it remains as easy as it was with the first piece of functionality.
+
+Please note that there are differences between research and platform requirements. While research needs speed and agility, the platform needs modularity and stability. These are different and can conflict. The guidance here is for the platform. _**If you are a researcher, you MAY ignore this guidance in your prototype and code in the way most effective for you and your task.**_ Later, if your prototype works and needs to be integrated into Monty, then we will refactor the prototype to correspond to the guidance here.
+
+# Guidance
+
+## Abstract classes MAY be used to specify interfaces without implementations:
+
+_Why?_ We want to move to [Protocols](https://typing.python.org/en/latest/spec/protocol.html#protocols) eventually. Keeping the abstract classes free of implementation makes it easier to transition to Protocols in the future. Whereas, having an abstract class with some implementation requires additional refactoring when we transition to Protocols.
+
+_Why do we want to move to Protocols eventually?_ We want to catch errors as early as possible, and using Protocols allows us to do this at type check time. Using abstract classes delays this until class instantiation, once the program runs.
+
+```python
+# ABCs raises errors during instantiation when/if constructor is called:
+class Monty(ABC):
+	def implemented(self):
+pass
+
+@abstractmethod
+def unimplemented(self):
+	pass
+
+class DefaultMonty(Monty):
+	pass
+
+monty= DefaultMonty() # runtime error
+
+# ---
+
+# Typical inheritance raises errors during runtime when monty.unimplemented() is called:
+class Monty:
+	def implemented(self):
+		pass
+
+	def unimplemented(self):
+		raise NotImplementedError
+
+class DefaultMonty(Monty):
+	pass
+
+monty = DefaultMonty() # ok
+monty.unimplemented() # runtime error
+
+# ---
+
+# Protocols raise errors during type check when monty.unimplemented() is written:
+class MontyProtocol(Protocol):
+	def implemented(self): ...
+	def unimplemented(self): ...
+
+class DefaultMonty:
+	def implemented(self):
+		pass
+
+monty: MontyProtocol = DefaultMonty() # ok
+monty.unimplemented() # fails type check
+```
+
+While abstract classes MAY be used, you SHOULD prefer Protocols.
+
+## Protocols SHOULD be preferred to document usage and expectations
+
+Protocols document a _behaves-like-a_ relationship.
+
+_Why_: We want to catch errors as early as possible, and using Protocols allows us to do this at type check time. Using abstract classes delays this until class instantiation, once the program runs.
+
+There is no material difference **in the context of usage and expectation documentation** between using Protocols and abstract classes. In other contexts, Protocols are favorable because they allow us to raise errors at type check time and, due to structural typing, do not require inheritance. Compare the code below to [Monty definition in abtract_monty_classes.py](https://github.com/thousandbrainsproject/tbp.monty/blob/c886e187c47aac6135a15b72052c71e34009a92b/src/tbp/monty/frameworks/models/abstract_monty_classes.py#L15).
+
+```python
+class Monty(Protocol):
+    ###
+    # Methods that specify the algorithm
+    ###
+    def _matching_step(self, observation):
+        """Step format for matching observations to graph.
+
+        Used during training or evaluation.
+        """
+        self.aggregate_sensory_inputs(observation)
+        self._step_learning_modules()
+        self._vote()
+        self._pass_goal_states()
+        self._pass_infos_to_motor_system()
+        self._set_step_type_and_check_if_done()
+        self._post_step()
+
+    def _exploratory_step(self, observation):
+        """Step format for adding data to an existing model.
+
+        Used only during training.
+        """
+        self.aggregate_sensory_inputs(observation)
+        self._step_learning_modules()
+        self._pass_infos_to_motor_system()
+        self._set_step_type_and_check_if_done()
+        self._post_step()
+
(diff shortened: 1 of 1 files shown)


tristanls:

@vkakerbeck, @nielsleadholm, @codeallthethingz, @hlee9212, @scottcanoe, @ramyamounir, @jeremyshoemaker please indicate your decision on the disposition to **merge**:
- approve: via approving this PR or commenting that you approve to merge
- abstain: by commenting that you abstain
- other: by commenting

ramyamounir (approved):

rfcs/0000_code_organization_guide.md
```diff
@@ -0,0 +1,655 @@
+- Start Date: 2025-07-17
+- RFC PR: (leave this empty, it will be filled in after RFC is merged)
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14](https://www.rfc-editor.org/info/bcp14) [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) [RFC 8174](https://www.rfc-editor.org/rfc/rfc8174) when, and only when, they appear in all capitals, as shown here.
```
jeremyshoemaker: We might want to indicate that the RFCs mentioned here are not our RFCs but rather IETF RFCs.

jeremyshoemaker (approved):
I left a comment but it looks fine to me.

rfcs/0000_code_organization_guide.md
```diff
@@ -0,0 +1,655 @@
...
+_Why_: We want to catch errors as early as possible, and using Protocols allows us to do this at type check time. Using abstract classes delays this until class instantiation, once the program runs.
+
+There is no material difference **in the context of usage and expectation documentation** between using Protocols and abstract classes. In other contexts, Protocols are favorable because they allow us to raise errors at type check time and, due to structural typing, do not require inheritance. Compare the code below to [Monty definition in abtract_monty_classes.py](https://github.com/thousandbrainsproject/tbp.monty/blob/c886e187c47aac6135a15b72052c71e34009a92b/src/tbp/monty/frameworks/models/abstract_monty_classes.py#L15).
+
+```python
+class Monty(Protocol):
+    ###
+    # Methods that specify the algorithm
+    ###
+    def _matching_step(self, observation):
```
scottcanoe: **suggestion (non-blocking)**: If I understand it correctly, these implemented methods won't be part of the required interface, so a duck-type could neglect to implement `_matching_step` and still pass type-checking. I suggest including some minimal guidance and/or gotchas on working with Protocols. It could be limited to if/when we should prefer that an implementation inherits from a Protocol and the above (to me) unexpected behavior surrounding implemented methods in Protocols

codeallthethingz:
I will abstain from this RFC, though comment scrolling through the code examples made me feel peaceful and happy. :)

scottcanoe (approved):
Thanks for doing this. It's very helpful.

**suggestion**: Please review code samples for indentation.

rfcs/0000_code_organization_guide.md
tristanls: The example is not a good Protocol example, so I'm going to remove it. I intended it to be a drop-in example for an abstract class in our current code, but it does not effectively communicate that concept.

tristanls:
> suggestion: Please review code samples for indentation.

Yikes! Thanks for pointing it out @scottcanoe 🙏.

tristanls:
Thanks @hlee9212!

I'll leave the type annotation to the upcoming type guidance RFC from @jeremyshoemaker.

I think instead of adding the configuration object, we can defer to enforcing the [PLR0913](https://docs.astral.sh/ruff/rules/too-many-arguments/) lint rule in the future.

I think that's good advice regarding the errors, and we should add it to the style guide. I want this RFC to focus on components vs. inheritance and how to assemble functionality, rather than being a general style guide for the code.

rfcs/0000_code_organization_guide.md
```diff
@@ -0,0 +1,555 @@
...
+class MontyProtocol(Protocol):
+	def implemented(self): ...
+	def unimplemented(self): ...
+
+class DefaultMonty:
+	def implemented(self):
+		pass
+
+monty: MontyProtocol = DefaultMonty() # ok
+monty.unimplemented() # fails type check
```
vkakerbeck: Sorry for the naive question but why wouldn't this fail the typecheck without the protocol? Shouldn't this also raise a type error:
```
class DefaultMonty:
	def implemented(self):
		pass

monty = DefaultMonty()
monty.unimplemented()
```
Why does the typechecker not check whether `DefaultMonty` adheres to `MontyProtocol` when we call `monty: MontyProtocol = DefaultMonty()`?

vkakerbeck: I tested this and when I cpy the protocol code example mypy already raises and error when initializing the DefaultMonty class (which is what I would want the protocol to help with). When I use the code without the protocol I also get a type error on the line where unimplemented is called.

rfcs/0000_code_organization_guide.md
```diff
@@ -0,0 +1,555 @@
...
+		self._area_computer = area_computer
+
+	@property
+	def area(self) -> float:
+		return self._area_computer(self._length, self._height)
+```
+
+## To share a functionality implementation that reads the state of the instance being mixed with, mixins MAY be used
+
+For sharing functionality, mixins only implement a shared behaves-like-a functionality. They add functionality, they do not add state. However, mixins SHALL NOT add state to the instance being mixed with. Every time you find yourself in need of state when working on a mixin, switch to composition instead.
```
vkakerbeck:
```suggestion
For sharing functionality, mixins only implement a shared behaves-like-a functionality. They add functionality, however, mixins SHALL NOT add state to the instance being mixed with. Every time you find yourself in need of state when working on a mixin, switch to composition instead.
```

rfcs/0000_code_organization_guide.md
```diff
@@ -0,0 +1,555 @@
...
+# class Child(Mixin, Parent) where we skip super().__init__() in Mixin
+
+class Parent:
+    def __init__(self) -> None:
+        super().__init__()
+        print("Parent init")
+
+class Mixin:
+    def __init__(self) -> None:
+	 	# super().__init__() skipped
```
vkakerbeck:
```suggestion
	 # super().__init__() skipped
```

rfcs/0000_code_organization_guide.md
```diff
@@ -0,0 +1,555 @@
...
+
+# Output
+# > Mixin init
+# > Child init
+
+# class Child(Parent, Mixin) where we skip super().__init__() in Parent
+
+class Parent:
+    def __init__(self) -> None:
+		# super().__init__() skipped
```
vkakerbeck:
```suggestion
	# super().__init__() skipped
```

vkakerbeck (approved):
Looks great, thanks for adding this!

rfcs/0000_code_organization_guide.md
tristanls: > Why does the typechecker not check whether DefaultMonty adheres to MontyProtocol when we call monty: MontyProtocol = DefaultMonty()?

~~The Python type system isn't smart enough to do this check.~~ It turns out it does.

> I tested this and when I cpy the protocol code example mypy already raises and error when initializing the DefaultMonty class (which is what I would want the protocol to help with). When I use the code without the protocol I also get a type error on the line where unimplemented is called.

Fair point, this happens because trying to use a non-existent method results in a type error.

Abstract classes will raise runtime errors on class instantiation. They will also raise type errors. But, they _have to_ specify the inheritance chain.

Protocols will raise runtime errors when trying to use the unimplemented method. They will also raise type errors. They do not require specifying the inheritance chain.

This set of examples may be clearer.

```python
# ABCs raises errors during instantiation when/if constructor is called:
class Monty(ABC):
    def implemented(self):
        pass

    @abstractmethod
    def unimplemented(self):
        pass

class DefaultMonty(Monty):
    pass

def invoke(monty: Monty):
    monty.unimplemented()

monty = DefaultMonty() # runtime error & fails type check
invoke(monty)  # OK, no type error

# Typical inheritance raises errors during runtime when monty.unimplemented() is called:
class Monty:
    def implemented(self):
        pass

    def unimplemented(self):
        raise NotImplementedError

class DefaultMonty(Monty):
    pass

def invoke(monty: Monty):
    monty.unimplemented()  # runtime error

monty = DefaultMonty() # OK, no type error
invoke(monty) # OK, no type error

# Protocols raise errors during type check when attempting to use it:
class MontyProtocol(Protocol):
    def implemented(self): ...
    def unimplemented(self): ...

class DefaultMonty:
    def implemented(self):
        pass

def invoke(monty: MontyProtocol):
    monty.unimplemented()  # runtime error

monty = DefaultMonty()  # OK
other: MontyProtocol = DefaultMonty() # fails type check
invoke(monty)  # fails type check
```

vkakerbeck: okay thanks, that makes sense

tristanls: Also, yes, it looks like this part of the example is wrong:
```python
monty: MontyProtocol = DefaultMonty() # ok
```
and should be
```python
monty: MontyProtocol = DefaultMonty() # fails type check
```

jeremyshoemaker: I think there's also a difference in the error you'll get from the type checker. One will be about the object not having a method, the other will be about the object not implementing the specified protocol. The latter is more helpful than the former, because it says _why_ it's a problem.

rfcs/0000_code_organization_guide.md
tristanls: What's the reason for this suggestion? The intent is that, in the context of allowable mixin use, they add functionality, but they do not add state.

jeremyshoemaker: That's what the sentence says after the "however". This change looks like an improvement to me.
> They add functionality, however, mixins SHALL NOT add state to the instance being mixed with.

tristanls: Ok, 2 to 1 takes it :)

vkakerbeck: yeah I didn't mean this to look like I think mixins should add state, I just removed it because it seems like a repetition of the next sentence.

tristanls:

PR #388, opened 2025-07-21
chore: error when attempting to use uv

tristanls:
With this change, when someone attempts `uv sync`, they will observe:

```
(tbp.monty) me@transient_atmospheric_phenomenon ~/tbp/tbp.monty % uv sync
Resolved 226 packages in 1ms
error: The current Python platform is not compatible with the lockfile's supported environments: `sys_platform == 'uv tool is not yet supported'`
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -336,3 +336,6 @@ namespaces = true
 [tool.setuptools.package-data]
 # habitat-sim resources
 'tbp.monty.simulators.resources' = ['*.json', '*.yml', '*.txt']
+
+[tool.uv]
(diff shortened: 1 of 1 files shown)


jeremyshoemaker:
I don't think this is a good idea, especially for something that's a "read the manual" problem, and a temporary one at that.

For one thing, it will needlessly break the uv tests. Even though they aren't required, I'm adding tests to them as I build the simulator and I'd prefer to keep them working. It also makes my work a little harder as I have to change this back in my branches, but remember to not commit it.

tristanls:
Yeah, fair points. I wanted to see what this method does, but I agree it's too disruptive.

PR #389, opened 2025-07-21
docs: Fix broken tutorial

scottcanoe:
The first tutorial broke due to recent refactoring of the sensor- and learning-module hierarchy.

The original tutorial used the most general classes possible, such as `MontyExperiment` and a now-deleted Monty config that used `MontyBase`, `SensorModuleBase`, and `LearningModuleBase`. The way I got this tutorial working again was to use more a more typical setup with `MontySupervisedObjectPretrainingExperiment`, `PatchViewFinderMountHabitatDatasetArgs`, etc.

It would have been nice to use the most generic versions of these classes, but it turned out to be a bit complicated. More specifically, our class interfaces aren't totally consistent, so one class can't really be swapped out without also swapping out others. For example, `MontyBase` calls `lm.pre_episode()` without a `primary_target` argument. `LearningModuleBase` was the only learning module that didn't accept/require `primary_target`, so `MontyBase` is now unusable with `DetailedLoggingSM` and its subclasses.

For what it's worth, I think the way this tutorial broke was pretty informative. There are undocumented dependencies between some of our classes, largely owing to the fact that our abstract methods don't have specific, required signatures. I don't want to get too into the weeds on this here though.

### DIFF
--- a/.vale/styles/config/vocabularies/TBP/accept.txt
+++ b/.vale/styles/config/vocabularies/TBP/accept.txt
@@ -180,3 +180,4 @@ Montymentous
 Viviane
 Inferencing
 LEGOs
+reimplement
--- a/docs/how-to-use-monty/tutorials/running-your-first-experiment.md
+++ b/docs/how-to-use-monty/tutorials/running-your-first-experiment.md
@@ -19,16 +19,20 @@ from dataclasses import asdict
 
 from benchmarks.configs.names import MyExperiments
 from tbp.monty.frameworks.config_utils.config_args import (
-    SingleCameraMontyConfig,
-    LoggingConfig
+    LoggingConfig,
+    PatchAndViewMontyConfig,
 )
 from tbp.monty.frameworks.config_utils.make_dataset_configs import (
     ExperimentArgs,
     get_env_dataloader_per_object_by_idx,
 )
 from tbp.monty.frameworks.environments import embodied_data as ED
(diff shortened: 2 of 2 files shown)


scottcanoe:
@codeallthethingz or @tristanls I'm getting ```Error: Vale.Spelling:Did you really mean 'reimplements'?```. Maybe I'm losing my mind, but I think the spelling here is correct. Is there a dictionary somewhere with allowed words? I vaguely remember something like that.

codeallthethingz:
> @codeallthethingz or @tristanls I'm getting `Error: Vale.Spelling:Did you really mean 'reimplements'?`. Maybe I'm losing my mind, but I think the spelling here is correct. Is there a dictionary somewhere with allowed words? I vaguely remember something like that.

Yes, its in accept.txt https://github.com/thousandbrainsproject/tbp.monty/blob/main/.vale/styles/config/vocabularies/TBP/accept.txt

scottcanoe:
@vkakerbeck I think the interfaces will get more consistent over time with the team's new guidelines. I just wanted to note the (to me) unexpected interface change between abstract/base classes and concrete implementations since it caught me off guard.

I did prefer this tutorial with the more general classes though. I'll try to keep this in mind and also more config-explanation text for tutorials though. I bet we can restore this tutorial to its more general form in the future too, after 1-2 small changes.

PR #366, opened 2025-07-08
rfc: intelligent resampling in monty

hlee9212:
Click [here](https://github.com/hlee9212/tbp.monty/blob/hlee9212/intelligence_resampling_rfc/rfcs/0000_intelligent_resampling.md) (thank you @codeallthethingz for pointing out the outdated link - has been updated!) to see the RFC rendered with images. :)

PR #390, opened 2025-07-22
rfc: dynamic adjustment for hypotheses resampling

ramyamounir:
This is an RFC for dynamic adjustment of hypothesis spaces with resampling.

### DIFF
--- a/rfcs/0000_resampling_dynamic_adjustment.md
+++ b/rfcs/0000_resampling_dynamic_adjustment.md
@@ -1,0 +1,126 @@
+- Start Date: 2025-07-22
+- RFC PR:
+
+# Summary
+
+This RFC extends [RFC-9](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0009_hypotheses_resampling.md), which introduced hypotheses resampling as a hypotheses updater function in `EvidenceGraphLM`.
+
+Instead of using fixed heuristics (e.g.,`hypotheses_count_multiplier` and `hypotheses_existing_to_new_ratio`) to
+determine the amount of resampling and hypotheses deletion at every step, this RFC proposes to have the
+resampling informed by more intelligent runtime heuristics, such as Monty's confidence in its prediction
+and step-wise prediction error. Dynamic adjustment of the hypothesis space will enable Monty to sample
+hypotheses only as needed, which is expected to improve Monty's run time as the hypothesis space can be smaller.
+
+Consider a scenario where, after a few steps of observing an object, Monty is confident in its prediction of the object
+id and its pose. In such case, continuing to resample will only slow down the recognition as it adds more computational
+overhead due to processing the unnecessarily resampled hypotheses at every step. However, if this high confidence is
+challenged with an unexpected observation (e.g., object swap), Monty should be smart enough to sample more hypotheses
+as it tries to make sense of this high prediction error.
+
+This RFC proposes the following:
+1. Redefine the resampling parameters to decouple the amount of resampled hypotheses from deleted hypotheses. At the moment,
+these are tightly coupled to satisfy the hypothesis space size multiplier parameter. More details in the [parameters decoupling](<0000_resampling_dynamic_adjustment#Resampling Parameters Decoupling>) section. 
+2. Allow resampling to be informed by Monty's step-wise performance metrics (e.g., confidence and prediction error), this will
+enable dynamic resizing of the hypothesis space. More details in the [Monty-informed resampling](<0000_resampling_dynamic_adjustment#Monty-Informed Resampling>) section.
+3. Optimize the `EvidenceSlopeTracker` to run faster. More specifically, implement these [notes](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/utils/evidence_matching.py?plain=1#L261-L271).
+
+
+## Resampling Parameters Decoupling
+
+The resampling procedure is currently defined by a few parameters that tightly couple resampling to hypotheses deletion.
+In other words, the user defines the `hypotheses_count_multiplier` parameter to communicate how the hypothesis space size
+should be scaled every step, and defines the `hypotheses_existing_to_new_ratio` to communicate how much of the existing
+hypotheses should be resampled. This design has several limitations:
+* We do not want fixed, user-defined resampling. These parameters fully define the resampling behavior, leaving no room for
+resampling to be informed by Monty. We want Monty to control these parameters during runtime.
+* Removing a multiplier parameter that requires the hypothesis space to be of a specific size (i.e., `hyp_space_size * multiplier`)
+allows for the possibility of these objects hyp. spaces to scale independently of each other. For instance, if deletion is
+based on a fixed evidence slope threshold, more hypotheses would be deleted from unlikely objects' spaces. This
+independent scaling is not possible with a global multiplier parameter applied to all hypothesis spaces.
+* While the existing parameters give us control over how to scale the hypothesis space, they cannot express all the desired resampling
+behaviors. For example, for a space of 100 hypotheses and a multiplier of 0.5 (desired
+hypothesis space of 50), we can use `hypotheses_existing_to_new_ratio` to control what portion of the 50 we want to be resampled
+vs. extracted from existing hypotheses, but we cannot e.g. choose to resample 80 and delete 130. Decoupling allows for defining those
+quantities independently.
+* A fixed user-defined multiplier (!= 1) is not ideal as it leads to continuous growth or decay of the hypothesis space size each step until
+it becomes quite large or diminishes to 0 hypotheses. The size will also vary exponentially because it's based off
+the size from the previous time step rather than a fixed value (i.e., number of nodes in the object graph).
+
+Therefore, we propose a new strategy for parameterizing the resampling procedure. Instead of controlling the hypotheses count multiplier, we
+can directly define parameters to specify how many hypotheses to be resampled and which hypotheses are to be deleted.
+Additionally, since this approach removes the concept of a fixed target hyp. space size, we must define minimum and maximum hypothesis
+space sizes to prevent unbounded growth or collapse.
+
+
+## Monty-Informed Resampling
+
+The resampling procedure consists of two main components: (1) which hypotheses to delete, and (2) how many new
+hypotheses to sample. Unlike the current implementation, we propose to have Monty control these details as follows:
+
+#### The deletion parameter
+
+This will likely be a fixed threshold on the evidence slope. A fixed threshold
+would allow different objects to scale their hypothesis spaces independently, which is desirable effect.
+That said, we could also consider more adaptive approaches. For example, the threshold could
+be adjusted dynamically to maintain a fixed number of top-performing hypotheses as a
+function of the object graph size (e.g., always keep the top 70 hypotheses for an object graph with
+100 nodes. These adaptive alternatives are worth exploring, but the simplest fixed-threshold option is the
+my preferred starting point.
+
+#### The resampling parameter
+
+This parameter would rely on Monty's confidence of the existing hypotheses and its prediction
+errors to determine how many new hypotheses are to be resampled. One useful idea is to combine
+these two quantities into a single metric; Monty’s surprise (placeholder name).
+This "surprise" metric captures how unexpected the current observation is, given Monty’s current belief:
+
+* A high prediction error in high-confidence hypotheses (i.e., high surprise)
+should trigger aggressive resampling.
+* A high prediction error in low-confidence hypotheses (i.e., low-to-moderate
+surprise) might require only mild resampling.
+* A low prediction error in high-confidence hypotheses could mean no resampling is
+needed.
+
+
+We can use existing `x_percent_threshold` logic to calculate confidence, whereas the prediction error can 
+easily be derived from the evidence score based on morphological and non-morphological features match. Note
+that the exact formula for calculating "Monty’s surprise" is considered an implementation detail and is
+therefore out of scope for this RFC.
+
+
+## Optimization Tricks
+
+This RFC also proposes addressing several performance improvements to the `EvidenceSlopeTracker`.
+These changes were discussed previously and are summarized below:
+
+#### Ring buffer for slope updates
+
+As suggested by Niels, we can replace the current method of shifting hypotheses indices by
+one at every step with a more efficient ring buffer structure. This avoids unnecessary
+array copying by using pointer-based indexing for inserting new data. The required changes
+are minimal:
+* Maintain a pointer to track where the next evidence entry should go.
+* Update the pointer each step, wrapping around the buffer as needed.
+
+#### Storing slopes instead of raw evidence
+
+Viviane suggested that instead of storing evidence values and computing their differences
+at each step, we can directly store the evidence slope values. This way we would not need
+to calculate the [`np.diff()`](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/utils/evidence_matching.py?plain=1#L380)
+of the evidence scores at every step. This might require some modifications especially
+because the first datapoint won't have a slope value, so it must still be stored in order
+to compute the first slope on the next step.
+
+# Plan of Work
+
+This is how I've planned the subtasks of this RFC, if approved.
+
+1. Implement the suggested code optimizations.
+    * Ensure that the performance didn't change from previous experiments.
+    * Measure the improvement in run time performance.
+2. Decouple resampling parameters
+    * Attempt to reproduce existing benchmarks by setting deletion and resampling to 10%
+3. Change deletion parameter to a fixed threshold and resampling parameter to Monty's surprise.
+    * Run unsupervised inference benchmarks and hyperparameter sweep on YCB benchmarks
+
+ 


rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,126 @@
...
+This RFC extends [RFC-9](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0009_hypotheses_resampling.md), which introduced hypotheses resampling as a hypotheses updater function in `EvidenceGraphLM`.
+
+Instead of using fixed heuristics (e.g.,`hypotheses_count_multiplier` and `hypotheses_existing_to_new_ratio`) to
+determine the amount of resampling and hypotheses deletion at every step, this RFC proposes to have the
+resampling informed by more intelligent runtime heuristics, such as Monty's confidence in its prediction
+and step-wise prediction error. Dynamic adjustment of the hypothesis space will enable Monty to sample
+hypotheses only as needed, which is expected to improve Monty's run time as the hypothesis space can be smaller.
+
+Consider a scenario where, after a few steps of observing an object, Monty is confident in its prediction of the object
+id and its pose. In such case, continuing to resample will only slow down the recognition as it adds more computational
```
nielsleadholm:
```suggestion
ID and its pose. In such a case, continuing to resample will only slow down the recognition as it adds more computational
```

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Monty-Informed Resampling
+
+The resampling procedure consists of two main components: (1) which hypotheses to delete, and (2) how many new
+hypotheses to sample. Unlike the current implementation, we propose to have Monty control these details as follows:
+
+#### The deletion parameter
+
+This will likely be a fixed threshold on the evidence slope. A fixed threshold
+would allow different objects to scale their hypothesis spaces independently, which is desirable effect.
```
nielsleadholm:
```suggestion
would allow different objects to scale their hypothesis spaces independently, which is a desirable effect.
```

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,126 @@
...
+That said, we could also consider more adaptive approaches. For example, the threshold could
+be adjusted dynamically to maintain a fixed number of top-performing hypotheses as a
+function of the object graph size (e.g., always keep the top 70 hypotheses for an object graph with
+100 nodes. These adaptive alternatives are worth exploring, but the simplest fixed-threshold option is the
```
nielsleadholm:
```suggestion
100 nodes). These adaptive alternatives are worth exploring, but the simplest fixed-threshold option is the
```

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,126 @@
...
+my preferred starting point.
+
+#### The resampling parameter
+
+This parameter would rely on Monty's confidence of the existing hypotheses and its prediction
+errors to determine how many new hypotheses are to be resampled. One useful idea is to combine
+these two quantities into a single metric; Monty’s surprise (placeholder name).
+This "surprise" metric captures how unexpected the current observation is, given Monty’s current belief:
```
nielsleadholm: **praise:** I like this idea! It's intuitive and nicely captures what we want. 

**thought:** I think discussing confidence vs. prediction error is also a useful way of framing our discussion the other day about the benefits of having both a shorter and a longer time-scale measure of evidence. 

The longer time-scale measure (the actual evidence value) can provide a measure of confidence, particularly when combined with something like x-percent-threshold as you suggest.

Prediction error is more instantaneous. In theory it could be from a single observation step, but the reality is that this would be quite noisy. That is why the slope-tracker is helpful - it looks at a medium length window to ensure that the prediction error is genuine. As such, we will likely want to continue using this. 

**Re. implementation of short-term prediction error**
Any form of shorter-context temporal integration would likely do (e.g. convolving the last few steps with an exponential kernel, calculating moving average, having an evidence value that just accumulates and decays more rapidly, etc.). For the last one, I think we could literally have the same concept as the current evidence scores, but just with more rapid integration/decay parameters; this would be conceptually simplest, and might tie in to some of @vkakerbeck 's pasts suggestions. Currently my view is that they would all be largely equivalent, so we might as well leave it as is, unless we see an obvious benefit like computational efficiency.

**Re. connection to voting**
When formulating the prediction error (e.g. via the slope-tracker), it could be nice to think how we would combine this with voting in the long term. For example, if it is quantified based on how many "inputs" there are to updating the evidence value, then these could be integrated over time (current slope-tracker), or space (voting basically passing a lot of information to slope tracker or an equivalent). From a neural point of view, this is equivalent to a neuron receiving many incoming signals over time, or all at once.

**Re. bio plausibility**
As bit of an aside, biological neurons definitely have different time constants for integration. As far as I recall, different neuron types can vary (e.g. inhibitory tend to integrate and spike much faster). In addition however, different types of synaptic channels (e.g. AMPA, NMDA) can have different temporal integration windows compared to one-another, as well as to the neuron as a whole. Finally, there is the added flexibility of dendritic compartments which might have different integration properties, or having a separate population of neurons integrating (pooling) a summary representation. All-in-all its both unsurprising that we are finding ourselves in need of at-least two measures of evidence at different time-scales, and this seems perfectly consistent with what biology could implement.

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,126 @@
...
+
+#### The deletion parameter
+
+This will likely be a fixed threshold on the evidence slope. A fixed threshold
+would allow different objects to scale their hypothesis spaces independently, which is desirable effect.
+That said, we could also consider more adaptive approaches. For example, the threshold could
+be adjusted dynamically to maintain a fixed number of top-performing hypotheses as a
+function of the object graph size (e.g., always keep the top 70 hypotheses for an object graph with
+100 nodes. These adaptive alternatives are worth exploring, but the simplest fixed-threshold option is the
+my preferred starting point.
```
nielsleadholm: **question:** Could we use the surprise metrics for deletion as well? If we have low surprise, then this could be a signal for pruning the hypothesis space, unless I'm overlooking something. That might also help unify these approaches.

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,126 @@
...
+* Update the pointer each step, wrapping around the buffer as needed.
+
+#### Storing slopes instead of raw evidence
+
+Viviane suggested that instead of storing evidence values and computing their differences
+at each step, we can directly store the evidence slope values. This way we would not need
+to calculate the [`np.diff()`](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/utils/evidence_matching.py?plain=1#L380)
+of the evidence scores at every step. This might require some modifications especially
+because the first datapoint won't have a slope value, so it must still be stored in order
+to compute the first slope on the next step.
```
nielsleadholm: **suggestion:** It could be worth adding a reference to Hojae's [in-progress RFC](https://github.com/thousandbrainsproject/tbp.monty/pull/366/files) along the lines of:

"Note this work is complementary to parallel work on resampling. In particular, RFC (ref) is working out the conceptual details of how prediction errors resulting from observations off an object can inform resampling. These observations correspond to seeing another object, or not sensing anything at all (e.g. the finger held in empty space), resulting in a high prediction error.

Notably, this parallel work is exploring how the CMP signal from a SM should be formulated and processed such that these kinds of observations result in a prediction error. The RFC here however considers how prediction errors should be integrated over time, and the resulting resampling strategies that follow from them."

Note to @hlee9212 - the above might also be useful context for you in terms of how your RFC ties in, and also a good way of formulating question 3 in your RFC as *dealing with SM outputs, and how these are processed*, not what to do with the prediction errors themselves.

nielsleadholm (commented):
Really nice first draft! I like the ideas you've presented and I think it's close to being ready for others to review, just a couple suggestions/discussion points.

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: Yes, these are all great points. I especially like how you framed confidence as a long-term quantity and prediction error as a short-term quantity (I didn't get it on Monday, but now I do), as well as the biological plausibility argument connecting the two. I've updated the text to reflect this.

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: Yes, we could. I'm unsure how this would affect the `x_percent_threshold` though. We may not need to rely on `x_percent_threshold` to come up with a confidence quantity, but if we do, then excessive pruning could introduce noise into the metric.

nielsleadholm: Ok yeah, I think for convergence, `x_percent_threshold` just requires at least two points at the moment (the MLH, and at least one other point that it is compared to). Assuming we deleted down to 1 hypothesis, then by definition, the MLH would be greater than `x_percent_threshold` above any "others". This would also be equivalent to maximal confidence. We could continue to just have one hypothesis (very efficient), unless we started getting prediction errors etc.

nielsleadholm (commented):
Nice! Thanks for those changes and for finding some sources for that neurobiology discussion. We can chat about the deletion options in a bit but otherwise I think the RFC is ready for a wider review.

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: Yeah, good point. Since we would also be defining a minimum hypothesis space size, we probably wont even let it go down to a single hypothesis anyway (if this is desirable).

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,168 @@
...
+space sizes to prevent unbounded growth or collapse.
+
+
+## Monty-Informed Resampling
+
+The resampling procedure consists of two main components: (1) how many new hypotheses to sample, and (2) which hypotheses
+to delete. Unlike the current implementation, we propose to have Monty control these details as follows:
+
+
+#### The resampling parameter
```
hlee9212: **question**: This may be getting a bit into the weeds, but are you thinking this parameter will be some integer number of hypotheses to resample? And that this value will be calculated based on Monty's "surprise"?

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,168 @@
...
+
+* A high prediction error in high-confidence hypotheses (i.e., high surprise)
+should trigger aggressive resampling.
+* A high prediction error in low-confidence hypotheses (i.e., low-to-moderate
+surprise) might require only mild resampling.
+* A low prediction error in high-confidence hypotheses could mean no resampling is
+needed.
+
+
+We can use existing `x_percent_threshold` logic to calculate confidence, whereas the prediction error can
```
hlee9212: **question**: This is more for me, but how are you thinking of using `x_percent_threshold` and cumulative evidence scores to compute confidence?

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,168 @@
...
+overhead due to processing the unnecessarily resampled hypotheses at every step. However, if this high confidence is
+challenged with an unexpected observation (e.g., object swap), Monty should be smart enough to sample more hypotheses
+as it tries to make sense of this high prediction error.
+
+This RFC proposes the following:
+1. Redefine the resampling parameters to decouple the amount of resampled hypotheses from deleted hypotheses. At the moment,
+these are tightly coupled to satisfy the hypothesis space size multiplier parameter.
+2. Allow resampling to be informed by Monty's step-wise performance metrics (e.g., confidence and prediction error), this will
+enable dynamic resizing of the hypothesis space.
+3. Optimize the `EvidenceSlopeTracker` to run faster. More specifically, implement these [notes](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/utils/evidence_matching.py?plain=1#L261-L271).
```
jeremyshoemaker: This URL should use a fixed commit hash instead of the `main` branch ref. If that file changes, the lines won't be the ones you meant at this point in time.

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,168 @@
...
+
+We can use existing `x_percent_threshold` logic to calculate confidence, whereas the prediction error can 
+easily be derived from the evidence score based on morphological and non-morphological features match. Note
+that the exact formula for calculating "Monty’s surprise" is considered an implementation detail and is
+therefore out of scope for this RFC.
+
+
+#### The deletion parameter
+
+This will likely be a fixed threshold on the evidence slope. A fixed threshold
```
hlee9212: **question**: Just for clarity, do you mean "delete" any hypotheses with evidence slop less than some value?

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,168 @@
...
+3. Optimize the `EvidenceSlopeTracker` to run faster. More specifically, implement these [notes](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/utils/evidence_matching.py?plain=1#L261-L271).
+
+
+## Resampling Parameters Decoupling
+
+The resampling procedure is currently defined by a few parameters that tightly couple resampling to hypotheses deletion.
+In other words, the user defines the `hypotheses_count_multiplier` parameter to communicate how the hypothesis space size
+should be scaled every step, and defines the `hypotheses_existing_to_new_ratio` to communicate how much of the existing
+hypotheses should be resampled. This design has several limitations:
+* We do not want fixed, user-defined resampling. These parameters fully define the resampling behavior, leaving no room for
```
jeremyshoemaker: This is probably just a personal Markdown style thing, but for bullet lists, I usually prefer to use `-` instead of `*` because the latter is usually used when making `**bold**` text. It really only matters when reading the raw text I suppose.

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,168 @@
...
+array copying by using pointer-based indexing for inserting new data. The required changes
+are minimal:
+* Maintain a pointer to track where the next evidence entry should go.
+* Update the pointer each step, wrapping around the buffer as needed.
+
+#### Storing slopes instead of raw evidence
+
+Viviane suggested that instead of storing evidence values and computing their differences
+at each step, we can directly store the evidence slope values. This way we would not need
+to calculate the [`np.diff()`](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/utils/evidence_matching.py?plain=1#L380)
```
jeremyshoemaker: Same as above, should use a fixed commit hash instead of the `main` branch ref.

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,168 @@
...
+As suggested by Niels, we can replace the current method of shifting hypotheses indices by
+one at every step with a more efficient ring buffer structure. This avoids unnecessary
+array copying by using pointer-based indexing for inserting new data. The required changes
+are minimal:
+* Maintain a pointer to track where the next evidence entry should go.
+* Update the pointer each step, wrapping around the buffer as needed.
+
+#### Storing slopes instead of raw evidence
+
+Viviane suggested that instead of storing evidence values and computing their differences
```
hlee9212: Not sure if this is the same idea, but my understanding of the slope calculation now is:
```
slope = mean([e1-e0, e2 - e1, ..., e_n - e_{n-1}])
```

Which I think is equivalent to just because the middle terms cancel out..? 
```
slope = (e_n - e_0) / n 
```

Also, I don't think memory is problem because window size is only 3, but exponential moving average (EMA) would be O(1) instead of O(window_size). 

```python
# recursive EMA
import numpy as np

def exponential_moving_average(stream, alpha):
    """
    Recursive EMA that handles np.nan values.
    
    Args:
        stream (iterable): Numeric stream, possibly containing np.nan.
        alpha (float): Smoothing factor (0 < alpha < 1).

    Yields:
        float: EMA value at each step. np.nan if initialization has not occurred.
    """
    ema = np.nan
    for x in stream:
        if np.isnan(x):
            yield ema  # retain the last EMA
        elif np.isnan(ema):
            ema = x  # first valid point initializes EMA
            yield ema
        else:
            ema = alpha * x + (1 - alpha) * ema
            yield ema
```
(With EMA we are still giving bias to most recent evidence values, just like windowed moving average we currently have. Technically you can also "window" EMA too by strictly forgetting old values completely, but...there really isn't any point because the updates are very efficient). 

If you want EMA to behave similar to current implementation, then setting `alpha = 2 / (window_size + 1)` would be approximate. Sadly it can't be exact because the windowed moving average is like a boxcar filter (sharp cutoff) and EMA is like a low-pass filter (exponential decay). I'm not sure if calculating the slope was the bottleneck, but in incremental average updating in general, you should only need current average, new value, and alpha (or window_size)?

Incremental average updating (because I love this trick so much):
$$\mu_n = \mu_{n-1} + \frac{x_n - \mu_{n-1}}{n}$$
^ blew my mind the first time when I realized you don't need to store all $n$ values to calculate average 🤣

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,168 @@
...
+
+This work is complementary to parallel work on resampling.
+In particular, The [RFC on Intelligent Resampling in Monty](https://github.com/thousandbrainsproject/tbp.monty/pull/366) is
+working out the conceptual details of how prediction errors resulting from observations off an object can inform resampling.
+These observations correspond to seeing another object, or not sensing anything at all (e.g. the finger
+held in empty space), resulting in a high prediction error.
+
+Notably, this parallel work is exploring how the CMP signal from a SM should be formulated and processed
+such that these kinds of observations result in a prediction error.
+The RFC here however considers how prediction errors should be integrated over time, and the resulting resampling
```
hlee9212: Awesome! 😄

rfcs/0000_resampling_dynamic_adjustment.md
tristanls: **suggestion (non-blocking)**: I think the concept described here as "surprise" is a good one and makes sense 👍. I dislike anthropomorphic terminology, though. ["We Are Obliged to Build Robots We Are not Obliged to"](https://www.ua-magazine.com/2019/03/25/artificial-intelligence-robots-joanna-bryson/). I would like to suggest "novelty", or anything less evocative. 

**note**: Paraphrasing things as Monty's confidence or surprise reads confusing to me, as these are singular learning module properties and not emergent or otherwise global Monty system properties. What will we call the confidence or novelty of a heterarchy of learning modules if not Monty's?

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,168 @@
...
+2-3 times faster[^1] than excitatory neurons.
+In addition, different types of synaptic channels (e.g. AMPA, NMDA) can have different temporal integration windows compared
+to one-another, as well as to the neuron as a whole.
+Finally, there is the added flexibility of dendritic compartments which might have different integration properties[^2] ,
+or having a separate population of neurons integrating (pooling) a summary representation.
+All-in-all its both unsurprising that we are finding ourselves in need of at-least two measures of evidence at different
+time-scales, and this seems perfectly consistent with what biology could implement.
+
+
+## Optimization Tricks
```
tristanls: **question**: What is the reason for optimizing things right now? Did you determine that this is the slowest part of the experiment? What is good enough?

I would expect optimizing code to achieve specific performance gains due to an identified bottleneck. Without determining this is the computational bottleneck, none of this work might impact the overall experimental runtime.

**note**: This seems unrelated to the rest of this RFC.

tristanls (commented):
**issue (non-blocking)**: This RFC introduces a new method of dynamic adjustment for hypotheses resampling, but it proposes (to me) unrelated performance work without identifying a specific performance problem. Furthermore, optimization work is proposed to be done first before introducing and testing the new proposed mechanism described in this RFC. @ramyamounir, I will point to our talk during the hackathon about which test failure to address first. I am attempting to present the same line of reasoning here. It is not that this won't speed things up. We can do many things to speed things up, but choosing these (and the speed up they deliver) feels arbitrary to me. 

**suggestion**: Separate out performance from this RFC.

**praise**: The non-optimization parts of the RFC sound like a good idea 👍.

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: Yeah, good point. I don't really have a strong opinion about what to name this quantity. It's a placeholder name for now, we can take a vote for it :slightly_smiling_face: 

Below is some context for why I chose "surprise", but it is based on my subjective mental model of what Monty does during inference and I don't expect us all to agree on this.

---
The term "surprise" was originally used in information theory as a way to describe the entropy of a system. It is the inverse of the probability of some event. If some event has a high probability, that means the surprise quantity resulting from the event occurring is low, and vice versa. It's described [here](https://en.wikipedia.org/wiki/Entropy_(information_theory)) better. This [video](https://www.youtube.com/watch?v=YtebGVx-Fxw&t=1s) also explains it really well. 

I know we do not model Monty as a probabilistic system, but when when Monty makes a prediction, I see it as assigning some high probability to a specific hypothesis over the others. We just greedily sample from this probability distribution the hypothesis with the highest evidence. The confidence is defining the probability distribution and how sharp the distribution is at this chosen hypothesis (aka the temperature of the distribution). We wait until the confidence is high enough (i.e., using `x_percent_threshold`) and then sample greedily. So when the confidence is high (more pronounced probability peak), a prediction error causes a higher "surprise". I know @hlee9212 is good at this stuff and may have some useful input, but we probably don't want to go too far into probabilistic models because that's probably not how the brain works.

hlee9212: Maybe just `confidence_error` could work? Could be interpreted in two ways: a combination of confidence and prediction error, and also a literal error in confidence, both of which works. Another could be something like `obs_belief_mismatch`, but I feel like the former makes it easier to remember what the combined parameter is...

ramyamounir: Yeah, I think `confidence_error` would work. Although "confidence" and "belief" may also fall under anthropomorphic terminology. We can discuss it on Monday..

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: Nice catch, thanks!

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: Yes, exactly. Monty's surprise will dictate how many new hypotheses to resample, which will be some integer number, but it can vary each step based on what it sees (prediction error) when compared to the accumulated evidence (confidence).

rfcs/0000_resampling_dynamic_adjustment.md
jeremyshoemaker: Honestly, I like the anthropomorphic terminology, but that's probably because I read John McCarthy's "Ascribing Mental Qualities to Machines" paper, years ago.

What came up for me when reading this was the theory from Jeff's TBT book about how predictions work, and unexpected input causing a flurry of neural activity compared to something matching a prediction causing minimal spiking. Surprise seems as good a term as any to me to describe it.

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: I'm not quite sure yet, but i'm thinking the difference in cumulative evidence between the MLH and the other hypotheses can give us some kind of confidence measure. In my mind, the confidence is somewhat like the temperature of the evidence distribution, i.e., a measure of its sharpness.

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: Yes, I would probably try this first to see if everything works as expected. Then maybe try more sophisticated heuristics like the surprise, as Niels suggested.

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: > Which I think is equivalent to just because the middle terms cancel out?

Yes, that would be equivalent and, on paper, faster to calculate than a `np.diff` and `np.mean`, but I would have to find the first non-nan values in each row to do this computation. Note that, with resampling, we add new hypotheses every step, so they don't all have the same history/age. With something like `np.diff`, numpy is very likely [manipulating the strides](https://numpy.org/doc/stable/reference/generated/numpy.lib.stride_tricks.html) of the array to do it all in one step, but if I try to do something hacky, it will probably end up being less efficient :sweat_smile: I was just looking for the fastest vector operation to do this, but I'm sure it can be sped up as you mentioned.

I did something similar for the incremental average update in [EvidenceSDRTargetOverlaps](https://github.com/thousandbrainsproject/tbp.monty/blob/9677cc918adeca9ae21233d957c0401e84f482ab/src/tbp/monty/frameworks/models/evidence_sdr_matching.py#L467), but I don't think this works well for a moving window because we would have to remove the points corresponding to the earlier evidence scores that get dropped out of the window, right? Am I missing something?

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: Fair point. These were mostly left over comments from implementing the `EvidenceSlopeTracker` and would be considered "nice to have" additions. I agree that there isn't any pressing need for these at the moment, especially since we haven't really profiled the experiments and identified these as efficiency bottlenecks. 

What do you think @vkakerbeck and @nielsleadholm? Should we focus on the accuracy performance for now and perhaps plan another RFC later that only optimizes the bottlenecks (assuming runtime turns out to be an issue)?

rfcs/0000_resampling_dynamic_adjustment.md
vkakerbeck: Just to add my two cents and do a bit more bikeshedding: When I would just read `confidence_error` I would have low confidence on what it means. I would find it most intuitive to just call it `prediction_error` or to be more precise, `confidence_scaled_prediction_error`. I also don't have an issue with `surprise` since I know it from information theory and it seems appropriate in this context.

rfcs/0000_resampling_dynamic_adjustment.md
vkakerbeck: I agree, it seems like a separate topic from this RFC (potentially not even needing an RFC). I think we do have a bit of a performance issue, but this may already be addressed by the ability to shrink the hypothesis space introduced in this RFC. If not, then I agree with Tristan that for more involved performance optimizations, it would be worth profiling the code and checking which function calls are actually the bottlenecks before spending significant time on optimizing. If the changes are relatively simple (like swapping one numpy function for another, it might be faster to just do that and compare runtimes. Either way, this seems like a very empirical task that doesn't require the in-depth discussions of an RFC (which doesn't mean it's not worth doing).

rfcs/0000_resampling_dynamic_adjustment.md
```diff
@@ -0,0 +1,168 @@
...
+
+This is how I've planned the subtasks of this RFC, if approved.
+
+1. Implement the suggested code optimizations.
+    - Ensure that the performance didn't change from previous experiments.
+    - Measure the improvement in run time performance.
+2. Decouple resampling parameters
+    - Attempt to reproduce existing benchmarks by setting deletion and resampling to 10%
+3. Change deletion parameter to a fixed threshold and resampling parameter to Monty's surprise.
+    - Run unsupervised inference benchmarks and hyperparameter sweep on YCB benchmarks
```
vkakerbeck: Earlier you said that the formula for surprise is out of the scope of this RFC (no problem!) but now I am curious. Do you plan on calculating surprise based on the MLH or based on all hypotheses?

rfcs/0000_resampling_dynamic_adjustment.md
nielsleadholm: Good points, thanks for bringing that up @tristanls . For what it's worth, there is a bit of a performance hit from the recent changes to hypothesis resampling, although it hasn't been closely profiled to determine the specific cause. I agree it makes most sense to start by implementing the general-algorithmic changes suggested here. As you point out Viviane, this might already address the general performance change. 

If there is one still outstanding, then we can follow this up with more detailed work (without needing to create an RFC). As such, it would be cleanest to remove the discussed optimization work from the RFC - we will still have it in the history for future reference if needs be.

rfcs/0000_resampling_dynamic_adjustment.md
nielsleadholm: Great discussion.

> Paraphrasing things as Monty's confidence or surprise reads confusing to me, as these are singular learning module properties and not emergent or otherwise global Monty system properties. What will we call the confidence or novelty of a heterarchy of learning modules if not Monty's?

I think this is a really interesting question. In the TBT, nothing in the cortex can exist "outside" of columns/LMs. Even if we have a heterarchy of LMs, all the information (i.e., which neurons are active and by how much) exists in each column; connectivity between columns just informs those representations. 

As such, I think its actually intuitive that the human-experienced analogue of surprise would basically be the "surprise" experienced across a heterarchy of many LMs. Similarly for confidence.

Put one way: a very simple sensorimotor animal like a fly might well experience a very basic form of surprise, while a mouse with many columns might experience a more interesting one closer to our phenomenological experience. I think it's more of a continuum based on the complexity of the system, but it's not that surprise as a concept jumps out once you have a sufficient number of LMs.

I think its a great point to bring up, but my vote would be for keeping surprise.

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: Ok nice, I'll remove this section from the RFC.

rfcs/0000_resampling_dynamic_adjustment.md
ramyamounir: Some of the metrics I described require only a single hypothesis statistics (e.g., accumulated evidence and prediction error), while others require us to look at the full hypothesis space for a relative measure (e.g., confidence with `x_percent_threshold`). 

My current thinking is that it is not enough to just look at the accumulated evidence for each hypothesis to calculate the surprise, in addition we want to look at this accumulated evidence relative to the others which gives us a better measure of confidence. For example, if we have 10 hypotheses and each one has accumulated 100 evidence points, I would think they all have a low confidence score, regardless of the absolute accumulated evidence scores.

So I think it's a combination of accumulated evidence, the relative accumulated evidence scores, and the new evidence score(s) for a measure of the prediction error. However, I'm not sure about the exact formula yet.

vkakerbeck: sounds good :) Just one thing to maybe keep in mind is that in the case of symmetrical objects we may have many hypotheses with similarly high evidence scores but still want to have a high confidence. But we can get into that more at a later time.

nielsleadholm (approved):
Looks great, thanks for the updates!

rfcs/0000_resampling_dynamic_adjustment.md
scottcanoe: Surprise is, as you mentioned, something out of information theory. If your usage is close enough to what it captures in an information-theoretic or probabilistic framework, then let's not reinvent the wheel w.r.t terminology to be more or less anthropomorphic. We could come up with a new name, but a lot of people might have to mentally translate it back into "surprise" or some other widely understood likelihood-based wording anyways.

ramyamounir:
> a model-free system might detect that an object is moving. This is a well-understood feature of model-free processing that would be a very useful signal that we now need to aggressively start reevaluating the object's pose.

@scottcanoe Yeah, interesting! I guess we can code it in a way that can be controlled by model-free polices that detect object motion (e.g., optical flow). It's also interesting to think of how environment change directly affects the surprise quantity. If the motion is initiated by the agent e.g., saccade, then the change in feature resulting from this displacement isn't much of a surprise. However, if the feature changes without the agent moving, then this results in high surprise.

What's even more interesting is whether we can design a policy to follow the object features in order to minimize "surprise". I wonder if it's related to infants face tracking which they learn pretty early in their development.

ramyamounir:
# Final Comment Period

@thousandbrainsproject/monty-maintainers We are now ready to take this RFC to the next step ( [RFC workflow - step 7](https://thousandbrainsproject.readme.io/docs/request-for-comments-rfc#overall-workflow) ). I would like to motion for FCP and invite all maintainers to indicate their final votes on the merging of this RFC. 

- 👍 for merge
- 👎 for object
- 👀 for abstain

@vkakerbeck @hlee9212 @nielsleadholm @scottcanoe @codeallthethingz @jeremyshoemaker @tristanls

PR #392, opened 2025-07-22
docs: code organization guide

tristanls:
This pull request incorporates [RFC 11 Code Organization Guide](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0011_code_organization_guide.md) into the code style guide in documentation.

### DIFF
--- a/docs/contributing/style-guide/code-style-guide.md
+++ b/docs/contributing/style-guide/code-style-guide.md
@@ -134,7 +134,7 @@ _Why_: We want to catch errors as early as possible, and using Protocols allows 
 There is no material difference **in the context of usage and expectation documentation** between using Protocols and abstract classes. In other contexts, Protocols are favorable because they allow us to raise errors at type check time and, due to structural typing, do not require inheritance.
 
 ```python
-# Protocols raise errors during type check when monty.unimplemented() is written:
+# Protocols raise errors during type check when attempting use:
 class MontyProtocol(Protocol):
 	def implemented(self): ...
 	def unimplemented(self): ...
--- a/docs/contributing/style-guide/code-style-guide.md
+++ b/docs/contributing/style-guide/code-style-guide.md
@@ -60,7 +60,7 @@ This guidance does not dictate the only way to implement functionality. There ar
 
 Please note that there are differences between research and platform requirements. While research needs speed and agility, the platform needs modularity and stability. These are different and can conflict. The guidance here is for the platform. _**If you are a researcher, you MAY ignore this guidance in your prototype and code in the way most effective for you and your task.**_ Later, if your prototype works and needs to be integrated into Monty, then we will refactor the prototype to correspond to the guidance here.
 
-### Abstract classes MAY be used to specify interfaces without implementations:
+### Abstract Classes MAY Be Used to Specify Interfaces Without Implementations
 
 _Why?_ We want to move to [Protocols](https://typing.python.org/en/latest/spec/protocol.html#protocols) eventually. Keeping the abstract classes free of implementation makes it easier to transition to Protocols in the future. Whereas, having an abstract class with some implementation requires additional refactoring when we transition to Protocols.
 
@@ -125,7 +125,7 @@ invoke(monty)  # fails type check
 
 While abstract classes MAY be used, you SHOULD prefer Protocols.
 
-### Protocols SHOULD be preferred to document usage and expectations
+### Protocols SHOULD Be Preferred to Document Usage and Expectations
 
 Protocols document a _behaves-like-a_ relationship.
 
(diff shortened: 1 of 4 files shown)


docs/contributing/style-guide/code-style-guide.md
```diff
@@ -0,0 +1,617 @@
...
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+```
+
+The `<YEARS>` is the year of the file's creation, and an optional sequence or range of years if the file has been modified over time. For example, if a file was created in 2024 and not modified again, the first line of the header should be `# Copyright 2024 Thousand Brains Project`. If the file has been modified in consecutive years between 2022 and 2024, the header should be `# Copyright 2022-2024 Thousand Brains Project`. If the file has been modified in multiple non-consecutive years in 2022, then in 2024 and 2025, the header should be `# Copyright 2022,2024-2025 Thousand Brains Project`.
+
+In other words, if you are creating a new file, add the copyright and license header with the current year. If you are modifying an existing file and the header does not include the current year, then add the current year to the header. You should never need to modify anything aside from the year in the very first line of the header.
```
jeremyshoemaker, line 50: I've been thinking about this a little bit, and I'm wondering if we might want to automate checking for modifications updating the year range.

We have [CPY001](https://docs.astral.sh/ruff/rules/missing-copyright-notice/) turned on to catch when we are missing a copyright notice, and I use a template to quickly generate that when I add a new file (it uses the current year), but we don't have any tooling to check if the year or year range includes the current year when modifying a file.

It looks like that CPY001 rule is using a regular expression to do the check, and much as I dislike throwing those around thoughtlessly, I think we could come up with a script or something to pull out the date or date range and then check if the current year (or maybe year of modification on the file though they should be the same usually) in a modified file is correct.

We'll probably only really start to notice this being a problem next year.

docs/contributing/style-guide/code-style-guide.md
```diff
@@ -0,0 +1,617 @@
...
+
+> [!NOTE]
+> While we deeply value and appreciate every contribution, the source code file header is reserved for essential copyright and license information and will not be used for contributor acknowledgments.
+
+## Code Organization Guide
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14](https://www.rfc-editor.org/info/bcp14), [IETF RFC 2119](https://www.rfc-editor.org/rfc/rfc2119), [IETF RFC 8174](https://www.rfc-editor.org/rfc/rfc8174) when, and only when, they appear in all capitals, as shown here.
```
jeremyshoemaker: I know I'm the reason we added these IETF prefixes, but after that change I went and checked. There are some RFCs in the "main" set that aren't from the Internet Engineering Task Force (IETF), and an example of that is RFC 2119, which is from the Network Working Group.

I don't know if there's a generic term for these official ones that we could use to disambiguate them from ours. Maybe we should just leave them as-is, "RFC 2119", etc. Our numbers are **much** lower than most of the official ones and if someone is in doubt they can just click the links.

docs/contributing/style-guide/code-style-guide.md
```diff
@@ -0,0 +1,617 @@
...
+# Protocols raise errors during type check when monty.unimplemented() is written:
+class MontyProtocol(Protocol):
+	def implemented(self): ...
+	def unimplemented(self): ...
+
+class DefaultMonty:
+	def implemented(self):
+		pass
+
+monty: MontyProtocol = DefaultMonty() # ok
```
jeremyshoemaker: Isn't this the line that fails the type check? We're assigning a `DefaultMonty` to a variable annotated as `MontyProtocol` and it doesn't implement `unimplemented` so it shouldn't match the Protocol and I'd expect this to fail type checking.

docs/contributing/style-guide/code-style-guide.md
```diff
@@ -0,0 +1,617 @@
...
+	def resize(self, new_radius: float) -> None:
+		self._radius = radius
+		self._resize_counter.increment()
+
+	@property
+	def resize_count(self) -> int:
+		return self._resize_counter.count
+```
+
+### You SHOULD Always Include Call to `super().__init__()` in Your `__init__` Methods
```
jeremyshoemaker:
```suggestion
### You SHOULD Always Include a Call to `super().__init__()` in Your `__init__` Methods
```

docs/contributing/style-guide/code-style-guide.md
tristanls: We could probably use the [`notice-rgx` config param](https://docs.astral.sh/ruff/settings/#lint_flake8-copyright_notice-rgx) for this. [We only use `author` right now](https://github.com/thousandbrainsproject/tbp.monty/blob/160d7a264cc9b9a01630e1feaec787ecc21fc484/pyproject.toml#L320).

tristanls: Oh right, but that doesn't help enforce the rule. Probably would need a GitHub Action to check that if a file is modified, the copyright includes the current year, as you suggested.

docs/contributing/style-guide/code-style-guide.md
tristanls: I will revert to the form without the "IETF" prefix.

docs/contributing/style-guide/code-style-guide.md
tristanls: You're right, it's a mistake I made in the RFC. I'll make sure to correct it here.

PR #394, opened 2025-07-23
chore: change point normal to surface normal

jeremyshoemaker:
This set of commits changes all references from "point normal" to "surface normal" in our code, comments, and documentation in order to match the nomenclature used in the DMC paper, as well as general 3D modelling and graphics terminology.

This **does not** change the video that defines point normals and pose curvatures, since that will need to be rerecorded.

---

I think I got all the references, though there might be some places where an abbreviation was used instead. This was basically the result of running the equivalent of `grep "point[ _-]normal"` on the code base and updating the results.

I split it into three commits to make review a little easier, focusing on comments, documentation, and code separately. There are also one or two grammatical and spelling corrections that I corrected since I was in the neighborhood.


docs/how-monty-works/observations-transforms-sensor-modules.md
```diff
@@ -36,11 +36,11 @@ For an overview of **which type of data processing belongs where please refer to
...
+In this implementation, some features are extracted using all of the information in the sensor patch (e.g. locations of all points in the patch for surface normal and curvature calculation) but then refer to the center of the patch (e.g. only the curvature and surface normal of the center are returned). At the moment all the feature extraction is predefined but in the future, one could also imagine some features being learned.
 
-## Point Normals and Principle Curvatures
-We can use the point normal and principal curvature to define the orientation of the sensor patch. The following video describes what those represent.
-[Point Normals and Principle Curvatures](https://res.cloudinary.com/dtnazefys/video/upload/v1731342526/point_normal.mp4)
+## Surface Normals and Principle Curvatures
+We can use the surface normal and principal curvature to define the orientation of the sensor patch. The following video describes what those represent.
```
codeallthethingz: Just because the video currently calls it point normal and that will be a point of confusion until we create a new video.

```suggestion
We can use the surface normal (previously referred to as point-normal) and principal curvature to define the orientation of the sensor patch. The following video describes what those represent.
```

codeallthethingz (changes requested):
Thank you for this!

One small doc suggestion.

docs/future-work/motor-system-improvements/model-based-exploration-policy.md
```diff
@@ -4,10 +4,10 @@ title: Model-Based Exploration Policy
...
 
-Such active exploration will also be important when dealing with compositional or multi-object settings, where establishing views of all sides of an object is not straightforward due to the presence of other objects or object-parts.
\ No newline at end of file
+Such active exploration will also be important when dealing with compositional or multi-object settings, where establishing views of all sides of an object is not straightforward due to the presence of other objects or object-parts.
```
tristanls, line 13: **question**: I'm assuming these are newline additions?

docs/how-monty-works/policy.md
```diff
@@ -88,12 +88,12 @@ The policies mentioned above are aimed at efficient inference. There is also a s
 
 
 | List of all policy classes         | Description                                                                                                                                                                                                                                                                                             |
-| ---------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| ---------------------------------- |---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
tristanls: **nitpick (inconsistency)**: You removed the spaces in the header separator for the `Description` column but left them in `List of all policy classes`.

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -1566,14 +1566,14 @@ def orienting_angle_from_normal(
         Returns:
             degrees that the agent needs to turn
         """
-        original_point_normal = self.processed_observations.get_point_normal()
+        original_surface_normal = self.processed_observations.get_surface_normal()
```
tristanls, line 1569: **nitpick**: Could have gone all the way to `.surface_normal()` renaming and get rid of `get` noise word. To me, this is in the same category as fixing spelling, which was done opportunistically in the docs.

src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -18,10 +18,10 @@
 from tbp.monty.frameworks.models.abstract_monty_classes import SensorModule
 from tbp.monty.frameworks.models.states import State
 from tbp.monty.frameworks.utils.sensor_processing import (
-    get_point_normal_naive,
-    get_point_normal_ordinary_least_squares,
-    get_point_normal_total_least_squares,
     get_principal_curvatures,
+    get_surface_normal_naive,
+    get_surface_normal_ordinary_least_squares,
+    get_surface_normal_total_least_squares,
```
tristanls: **nitpick**: `get_` 👀 ... 😭

src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -325,37 +325,37 @@ def observations_to_comunication_protocol(self, data, on_object_only=True) -> St
 
         return observed_state
 
-    def _get_point_normals(self, obs_3d, sensor_frame_data, center_id, world_camera):
-        if self.point_normal_method == "TLS":
+    def _get_surface_normals(self, obs_3d, sensor_frame_data, center_id, world_camera):
```
tristanls, line 328: **nitpick**: `_get_` 👀 ... 😭

src/tbp/monty/frameworks/models/states.py
```diff
@@ -132,23 +132,23 @@ def get_feature_by_name(self, feature_name):
...
+        When self.sender_type == "SM", the first pose vector is the surface normal and
+        the second and third are the curvature directions.
+        When self.sender_type == "LM", the pose vectors correspond to the rotation of
+        the object relative to the model learned of it.
         """
         return self.morphological_features["pose_vectors"][pose_vector_index]
 
-    def get_point_normal(self):
-        """Return the point normal vector.
+    def get_surface_normal(self):
```
tristanls, line 142: **nitpick**: `get_` 👀 ... 😭

src/tbp/monty/frameworks/utils/graph_matching_utils.py
```diff
@@ -168,7 +168,7 @@ def get_initial_possible_poses(initial_possible_pose_type) -> list[Rotation] | N
 
 
 def add_pose_features_to_tolerances(tolerances, default_tolerances=20) -> dict:
-    """Add point_normal and curvature_direction default tolerances if not set.
+    """Add surface_normal and curvature_direction default tolerances if not set.
```
tristanls: **issue**: This docstring doesn't seem to correctly describe what's happening here. Shouldn't this be "Add pose vector tolerances if not set?", or something along those lines?

src/tbp/monty/frameworks/utils/graph_matching_utils.py
```diff
@@ -271,18 +271,18 @@ def get_scaled_evidences(evidences, per_object=False):
...
-    """Calculate custom distances modulated by point normal and curvature.
+    """Calculate custom distances modulated by surface normal and curvature.
 
     Args:
         nearest_node_locs: locations of nearest nodes to search_locs.
             shape=(num_hyp, max_nneighbors, 3)
         search_locs: search locations for each hypothesis.
             shape=(num_hyp, 3)
-        search_pns: sensed point normal rotated by hypothesis pose.
+        search_pns: sensed surface normal rotated by hypothesis pose.
```
tristanls: **issue**: I think this should be `search_sns` and not `search_pns`. Maybe need another pass to check for plurals?

src/tbp/monty/frameworks/utils/sensor_processing.py
```diff
@@ -240,40 +240,40 @@ def get_point_normal_total_least_squares(
...
-#         sensor_location: location of sensor. Used to have the point normal
+#         sensor_location: location of sensor. Used to have the surface normal
 #             point towards the sensor.
 #
 #     Returns:
-#         Point normal at center_id
+#         Surface normal at center_id
 #     """
 #     if on_object_only and point_cloud[center_id, 3] <= 0:
 #         # center of sensor patch is not on object
```
tristanls: **issue**: This is commented out dead code. Let's remove it, it is in the repo history if we need it in the future.

tristanls (commented):

docs/future-work/motor-system-improvements/model-based-exploration-policy.md
jeremyshoemaker: That's what it looks like. I don't know why GitHub is so bad at displaying that change, when in the email I got notifying me of your review the diff clearly showed the old version missing the newline.

This is happening automatically because I have my editors set up to make sure files end with a newline.

tristanls: Sounds good, that's not an issue for me, just wanted to verify 👍.

docs/how-monty-works/policy.md
jeremyshoemaker: That must have been done by PyCharm when I saved the markdown file. I'll have to check and see what settings I need to turn off to get it to stop doing that, but I don't intend to make this change.

src/tbp/monty/frameworks/utils/graph_matching_utils.py
jeremyshoemaker: Yeah, I wasn't sure either, since at first I was thinking these were referring to specific dictionary keys or something due to the snake_case.

Your suggestion sounds reasonable to me.

src/tbp/monty/frameworks/utils/graph_matching_utils.py
jeremyshoemaker: Finding the abbreviated forms is harder. I'll fix this and double check for others.

src/tbp/monty/frameworks/utils/sensor_processing.py
jeremyshoemaker: You'll get no argument from me 😄

src/tbp/monty/frameworks/utils/graph_matching_utils.py
scottcanoe: I concur. `pose_vectors` contains a surface normal and two axes/directions of principal curvature, but writing out `surface_normal` and `curvature_directions` like variables names is misleading.
```suggestion
    """Add default `pose_vectors` tolerances if not set.
```

docs/future-work/motor-system-improvements/model-based-exploration-policy.md
jeremyshoemaker: For comparison, this is how Git renders it on the CLI:
```diff
-Such active exploration will also be important when dealing with compositional or multi-object settings, where establishing views of all sides of an object is not straightforward due to the presence of other objects or object-parts.
\ No newline at end of file
+Such active exploration will also be important when dealing with compositional or multi-object settings, where establishing views of all sides of an object is not straightforward due to the presence of other objects or object-parts.
```
I don't recall how we displayed it when I worked on Bitbucket, but I know we did. Take that, GitHub. 😛

docs/how-monty-works/policy.md
jeremyshoemaker: Well, I'm not sure how it changed in the first place, but when I looked at it again, PyCharm said the table was formatted incorrectly and when I had it correct it, it removed the spaces on both. So it's still a change.

src/tbp/monty/frameworks/models/motor_policies.py
jeremyshoemaker: This feels like the camel's nose. If I make that change, then I feel like it would be one small step to make this a `@property` instead of a method, and the rest of the codebase is still doing this getter approach.

If we're okay with some temporary inconsistency, I'd be fine with making this change.

src/tbp/monty/frameworks/models/sensor_modules.py
jeremyshoemaker: Some of these feel like they should be something like `calc_*` since they're doing actual calculations.

docs/how-monty-works/policy.md
tristanls: _(we're probably spending too much time on this 🙂)_. The latest diff still appears inconsistent. Seems like an arbitrary change.

Screenshot 2025-07-23 at 13 34 56

jeremyshoemaker: What is going on? 😕 This is what I get when I show that commit on the command line.

```diff
 | List of all policy classes         | Description                                                                                                                                                                                                                                                                                             |
-| ---------------------------------- |---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+|------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | **MotorPolicy**                    | Abstract policy class.
 ```
 
 And this is the commit itself in GitHub https://github.com/thousandbrainsproject/tbp.monty/pull/394/commits/d4c2698c75096c20a70072ebcb47d553157d6dd8
 
 The only difference I see is the lack of spaces and it added a couple of hyphens to make the ASCII version of the table line up.

File: src/tbp/monty/frameworks/models/sensor_modules.py (hunk 5 above)
Comment (tristanls) [reply src/tbp/monty/frameworks/models/sensor_modules.py]: I would reference the points I raised here, that `calc_` would probably extraneous too, using numpy as an example of people not being confused that a calculation is happening: https://github.com/thousandbrainsproject/tbp.monty/pull/325#discussion_r2155640306

Comment (jeremyshoemaker) [reply src/tbp/monty/frameworks/models/sensor_modules.py]: Well, either way, it's definitely not `get` because these are standalone functions, not getters on a class. I'll remove the `get` prefixes.

File: docs/how-monty-works/policy.md (hunk 3 above)
Comment (tristanls) [reply docs/how-monty-works/policy.md]: What you have now looks good. GitHub UI interface was taking me to see from 3 commits, but not all commits. Once I explicitly changed that to see all commits, I see what you see. 👍

**Comment by jeremyshoemaker** [comment]
> Looks good to me. As @vkakerbeck suggested in Slack, it's worth running a benchmark experiment if you haven't already to make sure there aren't any unexpected naming conflicts in the stored models. I **think** point/surface normals are stored as `norm` so it should be fine in theory.

I'll give it a run through the benchmarks before merging. Just want to address feedback first before doing that.

File: src/tbp/monty/frameworks/models/motor_policies.py (hunk 4 above)
Comment (tristanls) [reply src/tbp/monty/frameworks/models/motor_policies.py:1569]: It's a nitpick. I'm happy you're entertaining changing it at all 🙂.

Comment (jeremyshoemaker) [reply src/tbp/monty/frameworks/models/motor_policies.py:1569]: Well, you have to start somewhere. Otherwise a change like this will either be a large PR touching lots of separate parts of the codebase, or it will just never happen.

File: src/tbp/monty/frameworks/models/sensor_modules.py [hunk 11]
```diff
@@ -325,37 +325,37 @@ def observations_to_comunication_protocol(self, data, on_object_only=True) -> St
...
         # to use, uncomment lines below and import open3d
-        # elif self.point_normal_method == "open3d":
-        #     point_normal_alt = get_point_normal_open3d(
+        # elif self.surface_normal_method == "open3d":
+        #     surface_normal_alt = get_surface_normal_open3d(
         #         obs_3d,
         #         center_id,
         #         sensor_location=self.state["location"],
         #         on_object_only=True,
         #     )
```
tristanls: **issue**: Sorry, didn't see earlier. Dead code referencing infrastructure we no longer use.

jeremyshoemaker: I remembered seeing that, but couldn't find it when I went back and thought that maybe it as in that larger block I deleted.

tristanls (approved):
As far as I can tell, only a few nitpicks are left, so I approve.

Thanks for putting together this update, which should make things easier to understand.

codeallthethingz (approved):
LGTM

docs/how-monty-works/observations-transforms-sensor-modules.md
```diff
@@ -36,11 +36,13 @@ For an overview of **which type of data processing belongs where please refer to
...
 
-## Point Normals and Principle Curvatures
-We can use the point normal and principal curvature to define the orientation of the sensor patch. The following video describes what those represent.
-[Point Normals and Principle Curvatures](https://res.cloudinary.com/dtnazefys/video/upload/v1731342526/point_normal.mp4)
+## Surface Normals and Principle Curvatures
+The main feature we extract in the Sensor Module is the pose, which consist primarily of the _surface normal_ and the two _principle curvature_ vectors. These three vectors are orthogonal to each other, where the surface normal is the vector perpendicular to the surface and pointing away from the object, and the two principal curvature vectors point in the directions of the greatest and least curvature of the surface.
```
vkakerbeck:
```suggestion
Each Sensor Module needs to extract a pose from the sensory input it receives. This pose can be defined by the _surface normal_ and the two _principle curvature_ vectors. These three vectors are orthogonal to each other, where the surface normal is the vector perpendicular to the surface and pointing away from the object, and the two principal curvature vectors point in the directions of the greatest and least curvature of the surface.
```

vkakerbeck (approved):
Just a small suggested edit. Generally don't call the pose a feature so I reworded that part a little. Otherwise really great, thanks for making this change!

PR #395, opened 2025-07-23
chore: address PLE0605, invalid format for __all__

tristanls:
(I think) because of the invalid format of `__all__`, `MontyExperiment` does not appear in the API docs.

Either way, fix the format.

### DIFF
--- a/src/tbp/monty/frameworks/experiments/monty_experiment.py
+++ b/src/tbp/monty/frameworks/experiments/monty_experiment.py
@@ -46,7 +46,7 @@ from tbp.monty.frameworks.utils.dataclass_utils import (
     get_subset_of_args,
 )
 
-__all__ = {"MontyExperiment"}
+__all__ = ["MontyExperiment"]
 
 logger = logging.getLogger("tbp.monty")
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -165,6 +165,7 @@ explicit_package_bases = true
 # TODO: Remove global ignore and handle missing type stubs
 ignore_missing_imports = true
 mypy_path = "src"
+# strict = true
 warn_unused_configs = true
 
 [tool.ruff]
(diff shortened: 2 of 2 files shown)


src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -46,7 +46,7 @@
     get_subset_of_args,
 )
 
-__all__ = {"MontyExperiment"}
+__all__ = ["MontyExperiment"]
```
jeremyshoemaker, line 49: It's a little surprising to me that a set doesn't work, but looking at both the documentation for `__all__` and `set`, it makes sense.

> The public names defined by a module are determined by checking the module’s namespace for a variable named `__all__`; if defined, it must be a sequence of strings which are names defined or imported by that module.
~ [The import statement](https://docs.python.org/3/reference/simple_stmts.html#the-import-statement)

> A set is an unordered collection with no duplicate elements.
~ [Sets](https://docs.python.org/3/tutorial/datastructures.html#sets)

As to _why_ `__all__` needs to be a sequence, I don't know, but a sequence implies order, and sets are unordered.

It's also worth noting that `__all__` **only** matters if you're doing `from foo import *`, which I don't think we ever really do, and is generally discouraged.

jeremyshoemaker (approved):
Looks good to me

src/tbp/monty/frameworks/experiments/monty_experiment.py
tristanls: @jeremyshoemaker our documentation automation uses `__all__` to identify what to ... document. I haven't looked at doc configuration, but I observed that if an entity is not in `__all__` it's not in [the API docs](https://api-monty.thousandbrains.org/).

It is a loose correlation in my mind, so it might not be precisely as I described.

PR #399, opened 2025-07-24
chore: visualization for hypotheses out-of-reference frame 

hlee9212:
This PR adds in a visualization capability to Monty focusing on hypotheses (and their locations and rotations) for an object model in an episode.  I made a companion video that explains and points out the visual elements to help with the code review (posting on Slack due to size limit). Below is a quick screenshot showing all visual elements. 

CleanShot 2025-08-27 at 11 39 13

### DIFF
--- a/tools/github_readme_sync/readme.py
+++ b/tools/github_readme_sync/readme.py
@@ -38,7 +38,7 @@ regex_images = re.compile(r"!\[(.*?)\]\((.*?)\)")
 regex_image_path = re.compile(
     r"(\.\./){1,5}figures/((.+)\.(png|jpg|jpeg|gif|svg|webp))"
 )
-regex_markdown_path = re.compile(r"\(([\./]*)([\w\-/]+)\.md(#.*)?\)")
+regex_markdown_path = re.compile(r"\(([\./]*)([\w\-/]+)\.md(#.*?)?\)")
 regex_cloudinary_video = re.compile(
     r"\[(.*?)\]\((https://res\.cloudinary\.com/([^/]+)/video/upload/v(\d+)/([^/]+\.mp4))\)",
     re.IGNORECASE,
--- a/tools/github_readme_sync/tests/readme_test.py
+++ b/tools/github_readme_sync/tests/readme_test.py
@@ -408,14 +408,28 @@ This is a test document.""",
     @patch.dict(os.environ, {"IMAGE_PATH": "user/repo"})
     def test_correct_file_locations_markdown(self):
         """Test file location correction for Markdown file paths."""
-        base_expected = f"[File 1](/docs/slug#sub-heading)"
+        base_expected = (
+            "[File 1](/docs/slug#sub-heading) and [File 2](/docs/slug2#sub-heading)"
+        )
 
         # Test cases for Markdown file paths
         markdown_paths_with_deep_link = [
-            "[File 1](slug.md#sub-heading)",
-            "[File 1](contibuting/slug.md#sub-heading)",
-            "[File 1](../contibuting/slug.md#sub-heading)",
-            "[File 1](../../contibuting/slug.md#sub-heading)",
+            (
+                "[File 1](slug.md#sub-heading) and "  # fmt: skip
+                "[File 2](slug2.md#sub-heading)"
+            ),
+            (
+                "[File 1](contibuting/slug.md#sub-heading) and "
+                "[File 2](contibuting/slug2.md#sub-heading)"
+            ),
+            (
+                "[File 1](../contibuting/slug.md#sub-heading) and "
+                "[File 2](../contibuting/slug2.md#sub-heading)"
+            ),
+            (
+                "[File 1](../../contibuting/slug.md#sub-heading) and "
+                "[File 2](../../contibuting/slug2.md#sub-heading)"
+            ),
         ]
 
         markdown_paths_without_deep_link = [


tools/plot/data_models.py
```diff
@@ -0,0 +1,179 @@
...
+        """Return a detailed string representation of the ObjectModel."""
+        n_points = len(self.object_points_wrt_model)
+
+        feature_names = [
+            attr
+            for attr in dir(self)
+            if not attr.startswith("_") and not callable(getattr(self, attr))
+        ]
+
+        return f"ObjectModel(n_points={n_points}, features={feature_names})"
```
jeremyshoemaker: I'm pretty sure you can use this shortcut and get the same output, although it will say `feature_names=...` instead of `features=...` but you could change the variable to be `features` and get the original output.
```suggestion
        return f"ObjectModel({n_points=!s}, {feature_names=!s})"
```
F-strings have a lot of nice formatting options that aren't easy to discover. This one is [self-documenting expressions](https://docs.python.org/3/whatsnew/3.8.html#bpo-36817-whatsnew)

tools/plot/data_models.py
```diff
@@ -0,0 +1,179 @@
...
+    Returns:
+        Transformed orientations in world frame.
+    """
+    if not isinstance(target_rotation_wrt_world, Rotation):
+        target_rotation_wrt_world = Rotation.from_euler(
+            "xyz", target_rotation_wrt_world, degrees=True
+        ).as_matrix()
+
+    if row_vector_format:
+        # A @ R.T = B
```
jeremyshoemaker: This comment and the one below are superfluous re-statements of the lines of code they document.
```suggestion
```

tools/plot/data_models.py
```diff
@@ -0,0 +1,179 @@
...
+        return orientations_wrt_model @ target_rotation_wrt_world.T
+    else:
+        # R @ A = B
```
jeremyshoemaker:
```suggestion
```

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+    """
+    model_path = experiment_log_dir / "0" / "model.pt"
+
+    if not model_path.exists():
+        raise FileNotFoundError(f"Model file not found at {model_path}")
+
+    return model_path
+
+
+def deserialize_json_chunks_fast(json_file, episode_id=0):
```
jeremyshoemaker: This function is lacking typing hints.

Is "fast" important enough to be in the function name? A better name might just be `read_episode_data` since that's what it does. Deserialization is an implementation detail.

There are certain situations in which this implementation would not be fast. In particular if we were rendering multiple episodes from the same file. This algorithm behaves similarly to walking a linked list, starting at the beginning each time. Is the data large enough that loading it all into memory would be costly? Loading it all would result in a simpler implementation at the cost of additional load time and memory usage, but it might be worth it.

This function also makes certain assumptions about the `json_file` format that might not hold, e.g. it contains single JSON values separated by newlines.

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+        sm_timesteps_with_use_state_true = []
+        for sm_timestep, obs in enumerate(processed_obs):
+            if obs["use_state"]:
+                sm_timesteps_with_use_state_true.append(sm_timestep)
+
+        if len(sm_timesteps_with_use_state_true) == self.num_lm_steps:
+            self.lm_to_sm_mapping = sm_timesteps_with_use_state_true
+            logger.info("Successfully mapped LM timesteps to SM timesteps")
+        else:
+            raise ValueError(
```
jeremyshoemaker: Can we come up with a more specific error? My docstring here is vague to fit on a single line summary, but you could rephrase it. I'm also not sure about the name.

```python
class MappingMismatch(ValueError):
    """Raised when there is a mismatch between SM and LM timesteps."""
    pass
```

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+
+def compute_reference_frame_analysis(
+    object_model: ObjectModelForVisualization,
+    hypothesis_locations: np.ndarray,
+    hypothesis_rotations: np.ndarray,
+    sensed_orientation: np.ndarray,
+    max_abs_curvature: float,
+    max_nneighbors: int = 3,
+    max_match_distance: float = 0.01,
+) -> dict:
```
jeremyshoemaker: This `dict` type isn't specific enough. From the docstring of this function it sounds like it could be a TypedDict with two allowed keys.

```python
class RefFrameAnalysis(TypedDict):
    is_inside_reference_frame: list[bool]
    nearest_node_locs: ... # I don't know the type of this one.
```

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+        raise FileNotFoundError(f"Model file not found at {model_path}")
+
+    return model_path
+
+
+def deserialize_json_chunks_fast(json_file, episode_id=0):
+    # Load specific episode by line number (0-indexed)
+    with open(json_file, "r") as f:
+        for _ in range(episode_id + 1):
+            line = f.readline().strip()
```
jeremyshoemaker: I don't think this is needed. It seems like `json.loads` works fine if there's trailing or leading whitespace.
```suggestion
            line = f.readline()
```

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+    ]  # GraphObjectModel
+    points = np.array(data.pos, dtype=float)
+
+    feature_dict = {}
+    # data.feature_mapping contains feature names mapped to column indices in data.x.
+    # Available features include: 'node_ids', 'pose_vectors', 'pose_fully_defined',
+    # 'on_object', 'object_coverage', 'rgba', 'hsv', 'principal_curvatures',
+    # 'principal_curvatures_log', 'gaussian_curvature', 'mean_curvature',
+    # 'gaussian_curvature_sc', 'mean_curvature_sc'.
+    # Each feature maps to [start_index, end_index] for slicing the feature tensor.
```
jeremyshoemaker: This feels like it should be in a docstring. It reads like the requirements for the contents of the file pointed to by `model_path`.

Alternatively, you could create a type (perhaps a TypedDict?) for the `data` and annotate it on Line 84. That wouldn't guarantee that the values loaded from the file are correct, because the type checker can't check it at runtime, but it would document the expected structure.

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+
+        self.object_pointcloud = Points(
+            self.object_model.object_points_wrt_world,
+            c="gray",
+        )
+        self.object_pointcloud.point_size(8)
+        self.plotter.at(self.main_renderer_ix).add(self.object_pointcloud)
+
+    def _add_hypotheses_points(
+        self,
```
jeremyshoemaker: To get `ruff` to stop splitting this over lines, remove the trailing comma.
```suggestion
        self
```

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+        else:
+            self.random_hypothesis_neighbor_points.append(nearest_node_points)
+
+        self.plotter.at(self.main_renderer_ix).add(nearest_node_points)
+
+    def _add_sensed_pose_vector_arrows(
+        self,
+        location: np.ndarray,
+        hypothesis_rotation: np.ndarray,
+        is_heh: bool = False,
```
jeremyshoemaker, line 645: What does this parameter name mean? `is_heh` sounds like an interjection. Heh.

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+    def _add_summary_text(self) -> None:
+        """Create summary text for current timestep."""
+        num_hypotheses_in_reference_frame = sum(self.hypotheses_inside_reference_frame)
+        num_hypotheses_outside_reference_frame = sum(
+            ~self.hypotheses_inside_reference_frame
+        )
+
+        formatted_position = np.array2string(
+            np.array(self.data_loader.ground_truth_position),
+            precision=2,  # show up to 2 decimal places
```
jeremyshoemaker: Superfluous comment.
```suggestion
            precision=2,
```

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+                s=0.65,
+                font="Courier",
+                c=color,
+            )
+            self.plotter.at(self.main_renderer_ix).add(item)
+
+
+def plot_target_hypotheses(
+    exp_path: Path,
+    episode_id: int = 0,
```
jeremyshoemaker: This is really a personal preference, but I prefer if we rely on the formatter to decide when to break things out into separate lines. Removing the final trailing comma will do that.
```suggestion
    episode_id: int = 0
```

tools/plot/cli.py
```diff
@@ -17,6 +17,7 @@
 
 from tools.plot import (  # noqa: E402
     correct_percentage_per_episode,
+    hypothesis_oorf_visualizer,
```
ramyamounir, line 20: I've been using the prefix `interactive_` before in the filename to highlight that this is a Vedo interactive plot, not a simple seaborn/matplotlib. It's mostly a style thing and just a suggestion.

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+    parent_parser: argparse.ArgumentParser | None = None,
+) -> None:
+    """Add the hypothesis_oorf subparser to the main parser.
+
+    Args:
+        subparsers: The subparsers object from the main parser.
+        parent_parser: Optional parent parser for shared arguments.
+    """
+    parser = subparsers.add_parser(
+        "hypothesis_oorf",
```
ramyamounir: For all of the plots we have so far, the cli parser name has matched the filename (in this case it would be `hypothesis_oorf_visualizer`). This makes it easier to figure out what parser arguments you need to provide to `cli.py`. I suggest we continue to follow this style for consistency.

ramyamounir: For the same reason, I would also put the dependency files in a subdirectory with the same name as the plot filename (or derived from it). It's also a style suggestion but I feel that the top-level directory should be only the main plot files so that we don't clutter the directory with different files for each plot.

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+
+"""Interactive tool for visualizing hypotheses that are out of object's reference frame.
+
+This visualizer requires that experiments have been run with detailed logging
+enabled to generate detailed_run_stats.json files. To enable detailed logging,
+use DetailedEvidenceLMLoggingConfig in your experiment configuration.
```
ramyamounir: Should we also mention something about requiring `save_raw_obs` to be set to `True` for the first two SMs?

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+    is_outside = np.all(mask, axis=1)
+    return {
+        "is_inside_reference_frame": ~is_outside,
+        "nearest_node_locs": nearest_node_locs,
+    }
+
+
+class HypothesesOORFVisualizer:
+    """Interactive visualizer for hypotheses that are out of object's reference frame.
+
```
ramyamounir, line 152: I think this would be a good place to add a detailed description of all the features that this plot offers.

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+    Args:
+        json_path: Path to the detailed_run_stats.json file containing episode data.
+        model_name: Name of pretrained model to load object from.
+        max_match_distance: Maximum distance for matching.
+        max_nneighbors: Maximum number of nearest neighbors to consider.
```
ramyamounir: Looks like the args in the docstring don't match the arguments of the `__init__` function.

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+        self.all_mlh_orientations = []
+        self.all_mlh_graph_ids = []
+        self.all_sm0_locations = []
+        self.all_sm1_locations = []
+        self.all_sm0_rgba = []
+        self.all_sm1_rgba = []
+        self.max_abs_curvatures = []
+        self.sensed_orientations = []
+
+    def load_episode_data(self, episode_id: int = None) -> None:
```
ramyamounir: Should episode_id accept `int | None` since you're already passing a default of None?

tools/plot/episode_loader.py
ramyamounir: Is this function faster than the existing `deserialize_json_chunks` in [logging_utils](https://github.com/thousandbrainsproject/tbp.monty/blob/914454d8bb60fcb71a3a7898c49d6a0bbc643342/src/tbp/monty/frameworks/utils/logging_utils.py?plain=1#L100)? Could we just import and reuse that function instead of redefining it?

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+                highest_evidence_location.reshape(1, 3),
+                self.object_model.model_origin_wrt_world,
+                self.ground_truth_position,
+                self.ground_truth_rotation,
+            )
+            highest_evidence_orientation = transform_orientations_model_to_world(
+                highest_evidence_rotation.reshape(1, 3, 3),
+                self.ground_truth_rotation,
+                row_vector_format=False,
+            )
```
ramyamounir: isn't the following equivalent `highest_evidence_location=hyp_location[highest_evidence_index]`? and similar for orientation?

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+                match the number of LM timesteps.
+        """
+        logger.info("Finding LM to SM timestep mapping")
+
+        self.lm_to_sm_mapping = []
+        processed_obs = self.sm0_data["processed_observations"]
+
+        sm_timesteps_with_use_state_true = []
+        for sm_timestep, obs in enumerate(processed_obs):
+            if obs["use_state"]:
```
ramyamounir, line 264: I think that mapping is already being logged under `['LM_0']['lm_processed_steps']`, is it equivalent to just use this boolean mask as the mapping?

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+    def _find_lm_to_sm_mapping(self) -> None:
+        """Find mapping between LM timesteps and SM timesteps using use_state.
+
+        Raises:
+            ValueError: If the number of SM timesteps with use_state=True does not
+                match the number of LM timesteps.
+        """
+        logger.info("Finding LM to SM timestep mapping")
+
+        self.lm_to_sm_mapping = []
```
ramyamounir: This is probably just a nitpick, but why do we clear the list here? The visualization is for a single episode, do we ever load multiple episodes? If this is needed, my personal preference is to create a function that clears these variables on `load_episode_data`, you can also call it at the end of `__init__`. I could be missing something here.

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+        """Create interactive visualization with slider for timestep navigation."""
+        # Create plotter main view and 2 small overlays for sensor images
+        custom_view_areas = [
+            dict(bottomleft=(0.0, 0.0), topright=(1.0, 1.0)),  # Main view (full window)
+            dict(bottomleft=(0.73, 0.79), topright=(0.86, 0.99)),  # SM_0 (top-right)
+            dict(bottomleft=(0.86, 0.79), topright=(0.99, 0.99)),  # SM_1 (top-right)
+        ]
+
+        self.plotter = Plotter(
+            shape=custom_view_areas,
```
ramyamounir, line 269: Very cool! I wasn't aware we could define the renderers shapes manually like this.

ramyamounir (commented):
This looks really nice @hlee9212, also the code is well organized and documented. I didn't check every transform and math detail :sweat_smile: but that's probably not needed since the visualization looks correct. I just left a few minor comments about file names and their organization, as well as a request to add more documentation on the nice features of this visualization.

Also, you probably want to edit the title of the PR from `feat` to `chore` because we only use `feat` for the main codebase under `src/tbp/monty` according to [RFC10](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0010_conventional_commits.md#types)

tools/plot/episode_loader.py
jeremyshoemaker: I'd definitely prefer to reuse existing functions if we can. The main optimization I see in this version is that it looks for a single episode and stops looking as soon as it finds it. It also doesn't parse the intervening lines. Parsing them is going to be more robust in the face of malformed data.

There are details about that other version that if I were reviewing it today I'd probably ask questions about, but overall it seems fine. I don't think the performance difference is going to be enough to warrant having a separate function just for this visualizer.

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+            self.lm_to_sm_mapping = sm_timesteps_with_use_state_true
+            logger.info("Successfully mapped LM timesteps to SM timesteps")
+        else:
+            raise ValueError(
+                f"Mismatch: {len(sm_timesteps_with_use_state_true)} SM "
+                f"use_state=True vs {self.num_lm_steps} LM timesteps"
+            )
+
+    def _extract_max_abs_curvature(self) -> None:
+        """Extract sensed curvature values from sensor module data for each timestep."""
```
nielsleadholm, line 277: **suggestion:** Per the discussion on Slack, it probably makes sense to focus on one direction color always being most convex etc., rather than max absolute value.

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+            pose_vectors = np.array(morphological_features["pose_vectors"])
+            sensed_orientation_matrix = np.reshape(pose_vectors, (3, 3))
+            self.sensed_orientations.append(sensed_orientation_matrix)
+
+        logger.info(f"Extracted {len(self.sensed_orientations)} orientation matrices")
+        logger.info(f"Orientation matrices shape: {self.sensed_orientations[0].shape}")
+
+    def _extract_sensor_locations(self) -> None:
+        """Extract sensor locations from SM properties for each timestep.
+
```
nielsleadholm, line 317: **suggestion:** Can you add a bit more documentation for the purpose of this function? Is this to account for the fact that the SM can receive inputs, whereas the LM does not always?

If so, this appears to be necessary for most of these methods, so maybe to comment on at a higher level somewhere in this script.

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+        processed_obs = self.sm0_data["processed_observations"]
+
+        for lm_timestep in range(self.num_lm_steps):
+            sm_timestep = self.lm_to_sm_mapping[lm_timestep]
+            obs = processed_obs[sm_timestep]
+
+            morphological_features = obs["morphological_features"]
+            pose_vectors = np.array(morphological_features["pose_vectors"])
+            sensed_orientation_matrix = np.reshape(pose_vectors, (3, 3))
```
nielsleadholm: **question:** Can you please clarify what this reshape is doing? (maybe with a `# [D, D, D] --> [D', D', D']` style comment)

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+        logger.info(f"Target object for episode 0: {self.target_name}")
+        logger.info(
+            f"{self.target_name} is at position: {self.ground_truth_position} "
+            f"and rotation: {self.ground_truth_rotation}"
+        )
+
+    def _initialize_hypotheses_data(self) -> None:
+        """Extract and transform hypotheses into world frame.
+
+        Additional identify hypotheses with highest evidence for the target object.
```
nielsleadholm: **nitpick:** I think there's a typo in this sentence. "Additionally?"

tools/plot/episode_loader.py
```diff
@@ -0,0 +1,355 @@
...
+        This may be different from MLH, which considers hypotheses across all objects.
+        """
+        for lm_step in range(self.num_lm_steps):
+            hyp_locations_wrt_model = np.array(
+                self.lm_data["possible_locations"][lm_step][self.target_name]
+            )
+            hyp_object_orientations_wrt_model = np.array(
```
nielsleadholm, line 207: **question:** Aren't these hypothesized (object) orientations already in world coordinates? Why would they be affected by the ground-truth orientation of the object.

tools/plot/data_models.py
```diff
@@ -0,0 +1,179 @@
...
+        self.target_orientation_wrt_world = target_orientation_wrt_world
+
+        self.object_points_wrt_world = transform_locations_model_to_world(
+            self.object_points_wrt_model,
+            model_origin_wrt_world,
+            target_location_wrt_world,
+            target_orientation_wrt_world,
+        )
+
+        object_feature_orientations_wrt_model = np.reshape(
```
nielsleadholm, line 68: **suggestion:** Might be worth commenting a bit on this reshape here. How important is this? Does it relate to the observation you shared that we go from column-major to row-major when we go from the stored pose-features to the rotation matrices?

tools/plot/data_models.py
```diff
@@ -0,0 +1,179 @@
...
+class ObjectModelForVisualization:
+    def __init__(
+        self,
+        points: np.ndarray,
+        features: dict[str, np.ndarray],
+        model_origin_wrt_world: np.ndarray,
+        target_location_wrt_world: np.ndarray,
+        target_orientation_wrt_world: np.ndarray,
+    ):
+        """Initialize ObjectModel and transform from model to world coordinates.
```
nielsleadholm, line 28: **suggestion:** Maybe elaborate a bit on this, i.e. we are going to take the learned model points, and transform them to be in the location and orientation defined by the ground-truth pose of the object in the world.

tools/plot/data_models.py
```diff
@@ -0,0 +1,179 @@
...
+            by broadcasting.
+        Note that this is equivalent to transposing A first, multiplying, then
+            transposing back the result, i.e.(R @ A.T).T = A @ R.T
+
+    Args:
+        orientations_wrt_model: Orientations to transform, with shape
+            (n_points or n_hypotheses, 3, 3).
+        target_rotation_wrt_world: Target orientation (Euler angles in degrees
+            or Rotation object).
+        row_vector_format: Whether the orientations are in row vector format.
```
nielsleadholm: **note:** My understanding is that this assumes (probably rightly) that `target_rotation_wrt_world` is always column-major. It is just the input orientations that can switch between row and column major right?

tools/plot/data_models.py
```diff
@@ -0,0 +1,179 @@
...
+    row_vector_format: bool = False,
+) -> np.ndarray:
+    """Transform orientation matrices from model to world coordinates.
+
+    In linear algebra, to rotate matrix (or tensor) A by rotation matrix R:
+
+    1. If A consists of matrices composed of column vectors, then:
+        R (shape: (3, 3)) @ A (shape: (N, 3, 3)) = B (shape: (N, 3, 3))
+            by broadcasting.
+    2. If A consists of matrices composed of row vectors, then:
```
nielsleadholm, line 158: **question:** I'm a bit confused by this. Is the transpose of A "correcting" for the row vs. column-major format? Would a reshape be more appropriate, and then doing the same operation as in (1)? 

If it's not, then it's not clear to me that we can do `R @ A.T`?, i.e. the first part of the derivation below. Might be easiest for us to go through in a Huddle. Alternatively, if you can use the below notation to clarify what the transpose vs. row/column-major is changing, I would find that really helpful. 

Screenshot 2025-09-01 at 3 29 47 PM

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+    max_abs_curvature: float,
+    max_nneighbors: int = 3,
+    max_match_distance: float = 0.01,
+) -> dict:
+    """Compute reference frame analysis for hypotheses.
+
+    Args:
+        object_model: ObjectModel containing points and features.
+        hypothesis_locations: Array of hypothesis locations (n_hypotheses, 3).
+        hypothesis_rotations: Rotation matrix for each hypothesis (n_hypotheses, 3, 3).
```
nielsleadholm, line 97: **suggestion:** Clarify that this is the hypothesized rotation of the object.

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+        sensed_orientation: Sensed pose vectors (3, 3) to be transformed by hypotheses.
+            At each timestep, the Sensor Module extracts pose_vectors from RGBD image,
+            where pose_vectors represent row vectors of [surface_normal, pc1, pc2].
+        max_abs_curvature: Maximum absolute value of principal_curvature_logs.
```
nielsleadholm, line 102: **note:** See separate comment about max-absolute vs max-convex curvature.

tools/plot/hypothesis_oorf_visualizer.py
```diff
@@ -0,0 +1,843 @@
...
+        is_heh: bool = False,
+    ) -> None:
+        """Add ellipsoid around hypothesis based on sensed orientation."""
+        surface_normal, tangent1, tangent2 = (
+            self._transform_sensed_orientation_by_hypothesis(hypothesis_rotation)
+        )
+
+        stretch_factor = 1.0 / (np.abs(self.current_sensed_curvature) + 0.5)
+        semi_axis_tangent = self.max_match_distance
+        semi_axis_normal = self.max_match_distance / (1 + stretch_factor)
```
nielsleadholm, line 574: **question:** This seems different from the line below in `get_custom_distances`, is it possible to clarify the mismatch? Sorry if I'm overlooking something.

```python
    custom_nearest_node_dists = eucledian_dists + np.abs(dot_products) * (
        1 / (np.abs(search_curvature) + 0.5)
    )
```

tools/plot/cli.py
hlee9212: Sounds like a good convention - thanks Ramy! :) Note: Haven't updated in this PR because it will make all files appear as new changes since last review...will do so in a follow-up PR since it is less critical.

tools/plot/data_models.py
hlee9212: Added. :)

tools/plot/data_models.py
hlee9212: Hi @nielsleadholm, the `.reshape` here transforms `features["pose_vectors"]` from (n_points, 9) to (n_points, 3, 3). The original `features["pose_vectors"]` is row-major, as per how we define it in [sensor_modules.py](https://github.com/thousandbrainsproject/tbp.monty/blob/1ba1563ca4b9eb91077e02f3ef657336ea92db53/src/tbp/monty/frameworks/models/sensor_modules.py#L174-L183):
```python
morphological_features = {
    "pose_vectors": np.vstack(
        [
            surface_normal,
            dir1,
            dir2,
        ]
    ),
    "pose_fully_defined": pose_fully_defined,
}
```
The row-major is coming from the fact that [np.vstack concatenates arrays row-wise](https://numpy.org/devdocs/reference/generated/numpy.vstack.html). This is then flattened to make our stored (n_points, 9) `features["pose_vectors"]`, and then reshaped back into (n_points, 3, 3). Since we are flattening and reshaping in the same order (default argument for flatten/reshape is `order="C"`), it does not change the fact that how we dfine sensed pose features are row-major. So, this is not related to going from column-major to row-major - I just wanted to pass in a matrix for `transform_orientations_model_to_world()`. :) 


(I'm not sure how to write the above as a comment - do you have any suggestions?)

tools/plot/data_models.py
hlee9212: Thanks! I learned something. :) I kind of went back and forth on this, but I decided to keep the current code for a few reasons: 

(1) as you said, the formatting option may not be widely known, and despite being longer I think current code is easier to comprehend,
(2) I used to filter out a bit more than just `if not attr.startswith("_")` and the self documenting expression doesn't have this option,
(3) I used to have slight problems with `!r` (I guess it's not `!s`), kind of like [this Github issue](https://github.com/python/cpython/issues/131176?utm_source=chatgpt.com).

Out of curiosity, do you think Python 3.14's t-strings will affect the self-documenting expressions of f-strings?

tools/plot/data_models.py
hlee9212: Hi @nielsleadholm, this is hard stuff, so I'll try my best to explain more clearly. :) Let me know if you'd still like to Huddle via Slack. 

> question: I'm a bit confused by this. Is the transpose of A "correcting" for the row vs. column-major format? Would a reshape be more appropriate, and then doing the same operation as in (1)?

I am doing the same operation as in (1), which I tried to indicate via this comment:

```python
"""
Note that this is equivalent to transposing A first, multiplying, then
ransposing back the result, i.e.(R @ A.T).T = A @ R.T
"""
```

I'll try to give a simpler example. Let's say we have a matrix A of shape (3, 100). Think of these as 100 (column) vectors in 3D. Let's say we want to rotate A by certain amount. A rotation matrix in 3D always has shape (3, 3), and applying R @ A means we will be doing (3, 3) @ (3, 100), which will result in A' of shape (3, 100). Notice how A and A' have the same shape of (3, 100) representing 100 (column) vectors in 3D, just that A' is a rotated version of A.

Now let's say we have the same matrix but they are in rows, i.e. A has shape (100, 3). We can't just do R @ A because of the shape mismatch. So, what we do instead is transpose A so it is (3, 100), perform the rotation, then transpose back the result. This statement is equivalent to doing `(R @ A.T).T`. The `A.T` represents the transpose, the `(R @ A.T)` represents performing the rotation, and `(R @ A.T).T` represents transposing back the result. This "equation" can be simplified by distributing out the transpose inside the parentheses, i.e. `(X @ Y).T = Y.T @ X.T`, so `(R @ A.T).T = A @ R.T`, which is the formula for (2). Doing a quick shape check, we started with a matrix of (100, 3) where we have 100 (row) vectors in 3D, (R @ A.T).T = ((3, 3) @ (3, 100)).T = ((3, 100)).T = (100, 3). So, we end up with the same shape as the original of (100, 3), except these are now 100 (row) vectors that are rotated. 

P.S.: The general "formula" for rotating anything is `R @ whatever_matrix` and remembering that we are rotating the column vectors of `whatever_matrix`. One can remember that it is the column vectors when you think about normal matrix multiplication - in X @ Y, we are doing a bunch of dot products between row of X and columns of Y.

hlee9212: Part 2 using DMC notations :) 

> Alternatively, if you can use the below notation to clarify what the transpose vs. row/column-major is changing, I would find that really helpful.

Just FYI, it's not really about changing row/column-major, it's just that we use _both_ conventions in Monty. Our sensed features from SM, i.e. `features["pose_vectors"]` or `morphological_features["pose_vectors"]` is in row-major. The hypothesized object orientation in the [Hypotheses class](https://github.com/thousandbrainsproject/tbp.monty/blob/1ba1563ca4b9eb91077e02f3ef657336ea92db53/src/tbp/monty/frameworks/models/evidence_matching/hypotheses.py#L24), i.e. `hypothesis.poses` is in column-major. So, I had to multiply them appropriately, being aware of which is row-major and which is column-major. 

Previously, I treated them as all column-major (which is default), which was why I was getting such weird surface normals.

---
Moving onto DMC notations! :)

Also FYI, what I want for visualization is sensed orientations (i.e. `morphological_features["pose_vectors"]`) in _world_ coordinates to draw the surface normal, PC1, and PC2. We don't have a symbol for that in DMC since we are not dealing with _world_ coordinates. I also want stored orientations (i.e. `features["pose_vectors"]`) from object model in _world_ coordinates so I can draw the "iron maiden" surface normals for the object points. 

```math
{}^{B}_{S}R_{t=0} = \text{sensed orientation at initial timestep, i.e. morphological\_features["pose\_vectors"], in row-major format in tbp.monty} 
```
```math
{}^{B}_{M}R_k = \text{hypotheses.poses[k], in column-major format in tbp.monty} 
```
```math
{}^{M}_{S}R_i = \text{stored orientation for node i from object model, i.e. `features["pose\_vectors"]`, in row-major format in tbp.monty}
```

tools/plot/data_models.py
hlee9212: Yes, that is the correct assumption. :) If I am not mistaken, I think `target_rotation_wrt_world` is always stored as Euler angles in `detailed_run_stats.json`, e.g. [280, 260, 160].
```python
if not isinstance(target_rotation_wrt_world, Rotation):
        target_rotation_wrt_world = Rotation.from_euler(
            "xyz", target_rotation_wrt_world, degrees=True
        ).as_matrix()
```
The `as_matrix()` operation will create a rotation matrix that is column-major (because columns are default in linear algebra/numpy/scipy). 

P.S.: I'll update the code to only handle Euler angles for clarity. 

---
Edit: Because I answered these comments in order of how they appeared in my files, the following has been answered elsewhere:

> It is just the input orientations that can switch between row and column major right? 

Brief summary:

Not necessarily "switching", but that we use both row-major and column-major conventions in `tbp.monty`. Specifically, the "pose_vectors" for sensor module or stored in object model is row-major, while `hypotheses.poses` is column-major.

tools/plot/data_models.py
hlee9212: Removed.

tools/plot/data_models.py
hlee9212: Removed.

tools/plot/episode_loader.py
hlee9212: Ah yes, I was using `deseralize_json_chunks` in `logging_utils.py`, but got frustrated during developing/debugging because it felt so slow, which I why I wrote a "fast" version (though not sure if it were truly faster, I didn't profile). 😂 Forgot to change it back to that version, will do that.

tools/plot/episode_loader.py
hlee9212: **note:** this function has been deleted.

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Added.

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Thanks Ramy! Good catch.

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Nice, thank you! Also, does this mean we will keep `"COM812", # COM812: Trailing comma missing` in our `pyproject.toml`?

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Heh. 😂 It stands for "highest evidence hypothesis". I swear I explained the acronym somewhere, but I can't find it now. I'll update various parts of the code with something like `highest evidence hypothesis (HEH)`. Also in a function above I wrote in docstring how it is different from MLH:

```python
"""
The highest evidence hypothesis is the one with the highest evidence score
for the subset of hypotheses whose target is the same as the object. It can be
different from the MLH, which is the highest evidence hypothesis across entire
set of objects.
"""
```

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Removed.

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Removed final trailing comma.

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Hi @nielsleadholm, this is a great question! And actually, you reminded me that I forgot to add docstring about this because it's not actually obvious and I had to do some math. 😅

How I derived it:
1. Let's say we have a point that is some distance $d$ away in the direction of the **surface normal**. 
2. The Euclidean distance to this point is $d$, and `np.abs(dot_products)` will be $d$ as well. 
3. So for this point, we have `custom_nearest_node_dists = d + d * (1 / np.abs(search_curvature + 0.5))`.
I defined `stretch_factor = 1 / np.abs(search_curvature) + 0.5`, so the above becomes:
```
custom_nearest_node_dists = d + d * stretch_factor 
                          = d * (1 + stretch_factor)
```
4. Now, to draw the ellipsoid, we want to allow `custom_nearest_node_dists` up until `max_match_distance`, or:
```
custom_nearest_node_dists <= max_match_distance
d * (1 + stretch_factor) <= max_match_distance
d <= max_match_distance / (1 + stretch_factor)
```
So, the normal axis length for the Ellipsoid is `d = max_match_distance / (1 + stretch_factor)`.

tools/plot/episode_loader.py
hlee9212: Good catch!

tools/plot/episode_loader.py
hlee9212: Hi @nielsleadholm, sorry, I meant `hyp_object_rotation_wrt_model` (not orientation). From my understanding (please correct me if I am wrong), this represents how much we need to rotate from the learned object model to get the ground-truth orientation of the object, not the absolute orientation of object [1]. I think I called it "orientation" because our learned model is in "standard/canonical" orientation or [0, 0, 0] in Euler angles, so in this case the rotation = orientation (like how rotating from identity is just that rotation). What I wanted to calculate was `hyp_object_orientation_wrt_world`, which means these need to be rotated by the object's orientation during testing. 

A follow-up may be, why do we need `hyp_object_orientation_wrt_world`? It comes down to drawing those sensed feature orientations for each hypothesis. (FYI I'm not sure if you noticed, but I dropped `_wrt_world` if the reference frame is world to shorten variable names.)

To draw those sensed feature orienations (surface normal, PC1, PC2 arrows) for each hypothesis in world coordinates, I needed `hyp_object_rotation_wrt_world`. It's actually two rotations, roughly:

sensed_orientations @ hypothesized_rotations_wrt_model @ object's test rotation = A @ B @ C 

In code, what I did was precompute B @ C first, i.e. A @ (B @ C). The `hypothesized_orientation_wrt_world` represents B @ C. 

---
[1]: I honestly cannot remember why I thought it was `_wrt_model`, but I'll cite Viviane's diagram:
CleanShot 2025-09-01 at 15 03 38
(the light blue or blumenta seems to indicate that `hypotheses.poses` is the rotation of the object wrt. model)

tools/plot/episode_loader.py
hlee9212: Hi @nielsleadholm, I can technically update the colors of the PC1/PC2 arrows, but I'm not sure about updating the name here? It comes from `hypotheses_displacer.py`:
```python
max_abs_curvature = get_relevant_curvature(channel_features)
custom_nearest_node_dists = get_custom_distances(
    nearest_node_locs,
    search_locations,
    pose_transformed_features["pose_vectors"][:, 0],
    max_abs_curvature,
)
```
and `get_relevant_curvature()` does get `np.max(np.abs(...))`. (Also this is the scalar curvature value, not PC1/PC2 direction.)

tools/plot/episode_loader.py
hlee9212: Oh oops, I checked and `morphological_features["pose_vectors"]` is already in (3, 3) matrix format (row-major), so no need to reshape. I thought it was flattened to (9,) like how it is in pose_vector from object_model. 

**Note:** Stored pose_vectors from object model is flattened to shape (n_points, 9), but sensed pose_vector is just (3, 3).

tools/plot/episode_loader.py
hlee9212: I'm not sure if the comment location got displaced, but I think you are referring to `self.lm_to_sm_mapping`? If so, then yes - it accounts for the fact that FeatureChangeSM does not always send things to LM. The visualization's slider goes through LM Steps, but a sensor module may have taken several steps before an observation gets sent for a single LM step. The `self.lm_to_sm_mapping` exists so that I'm showing the correct SM location and RGB patch images - however, this does mean that the Sensor Module may seem to "jump" because it is skipping those intermediate SM steps where LM is not doing anything. 

**Note:** Added docstring to `_find_lm_to_sm_mapping()` method.

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Saw the comment, but not sure if I should update because it is the terminology used in Monty when calculating custom distances. Also, the scalar curvature value itself is fine for modulating distance. We probably actually need to update docs in tbp.monty to say that PC1 corresponds to max convex curvature, not max absolute curvature, somewhere around [here](https://github.com/thousandbrainsproject/tbp.monty/blob/1ba1563ca4b9eb91077e02f3ef657336ea92db53/src/tbp/monty/frameworks/utils/sensor_processing.py#L493-L495).

```python
""" 
(add comment like this) We are sorting eigenvalues algebraically in descending order. 
This means k1 and k2 could have values like: k1 = 0.2 and k2 = -3.4. 
In such cases, k2/pc2 is more curved, but k1/pc1 is still max_convex_curvature. 
"""
idx = eigval.argsort()[::-1] 
eigval_sorted = eigval[idx]
eigvec_sorted = eigvec[:, idx]

k1 = eigval_sorted[0] 
k2 = eigval_sorted[1]
```

tools/plot/episode_loader.py
nielsleadholm: Oh that's actually an interesting point. It feels like we want to use the actual (rather than absolute) values in the matching code as well? Did you ever have any visualizations from _concave_ surfaces and what the ellipsoid looks like? If we're trying to account for the surface, then in my mind concave curvature should actually cause us to search in a neighborhood like is shown below (i.e. a concave lens shape). I couldn't see any discussion of this situation [here](https://thousandbrainsproject.readme.io/docs/evidence-based-learning-module#search-radius-and-distance-weighting).

Screenshot 2025-09-02 at 10 31 24 AM

@vkakerbeck is this something you looked at before or have any thoughts on?

(fwiw even if we did want to change this, I'm not suggesting we do this now, but might be something to make a note of)

vkakerbeck: Since we are using a spherical search radius I don't think we can account for convex vs. concave. But yes, I did have it on my list that it would be nice to have the search radius look further in the direction of the point normal (away from surface) and less far in the other direction (into the surface), at which point it would be useful to account for convex vs. concave.

tools/plot/hypothesis_oorf_visualizer.py
nielsleadholm: Ok yeah, I guess there is a difference between how we do custom distance calculation (absolute value - being discussed separately), and how we label the PC1 direction. My main point is just that it would be worth documenting/making it clear that when we plot the PC1 direction, we are plotting the maximally convex direction, not the max absolute curvature. Without this context, it might look like the purple arrow is sometimes incorrect.

tools/plot/episode_loader.py
nielsleadholm: Ok yeah, it would probably be pretty involved to first return a series of points and then apply the concave lens-type neighborhood.

tools/plot/episode_loader.py
nielsleadholm: Ok yeah they should be `hyp_object_orientations_wrt_world`. I just chatted with Viviane and she agrees, so the voting figure you linked is being updated to `world rel. model` (note the only reason it isn't `model rel. world` is because at that point we are using the inverse). Here (hypothesized) object orientation in your notation = model orientation in Viviane's diagram.

I still don't see why these need to be transformed by the ground-truth orientation. We are trying to visualize how the LM thinks the object is oriented in the world. The impact of this global hypothesis on the sensed local poses should be visualized at each of the hypothesized locations, not the ground-truth location of the model. That is therefore independent of the ground-truth orientation of the object, isn't it?

tools/plot/data_models.py
nielsleadholm: Makes sense thanks, I think just adding a brief summary of what you've written (e.g. about flattening and reshaping in the same order) would be perfect.

tools/plot/data_models.py
nielsleadholm: Great thanks, I would maybe just add a comment to this effect, i.e. these are Euler angles, and so the choice around the `row_vector_format` bool does not affect `target_rotation_wrt_world`.

tools/plot/data_models.py
nielsleadholm: Thanks Hojae, let's resolve the other comment first re. model relative to world. It's not entirely clear to me we even need this `transform_orientations_model_to_world` function, as the object orientations (whether ground-truth or hypothesized) are always relative to the world. 

Also to clarify, in DMC, world = body (i.e. `B`), as they are arbitrary global coordinate systems. Obviously they would be different if the body started moving in the world, but this doesn't happen in DMC.

tools/plot/hypothesis_oorf_visualizer.py
nielsleadholm: Thanks Hojae, makes sense and nice derivation.

tools/plot/data_models.py
jeremyshoemaker: I probably should have been clearer, because I wasn't suggesting not using it, or that it wasn't common knowledge, just that it's hard to discover. `f"{foo=}"` is the current idiomatic way to do it. It's easy to look up when you see it, but hard to come up with if you haven't before. That's definitely changed over time. It took a while for people to switch from `"foo = %s" % foo` to `"foo = {}".format(foo)` to `f"foo = {foo}"` to `f"{foo=}"`.

The reason I suggested using `!s` was because I thought this was a `__str__` implementation, but it's `__repr__` so you can remove the `!s`. F-strings default to calling `__format__` on items it is printing, and I'm pretty sure that falls back to `__repr__` by default. Those details usually aren't important. Usually you just try it and see if it looks the way you want and then figure out if you need to change something.

> (2) I used to filter out a bit more than just if not attr.startswith("_") and the self documenting expression doesn't have this option,

I'm not sure what you mean here. The filtering in the list comprehension looks fine to me. The use of `dir` seems a bit odd to me, I haven't seen that often, but there's nothing inherently wrong with it.

As for t-strings, I haven't had a chance to look at them and see how they work. They might become the new idiom at some point.

tools/plot/episode_loader.py
hlee9212: Hi @nielsleadholm, I tried commenting out the following code that transforms to "world coordinates" and attached a screenshot of the results. 
```python
# hyp_locations = transform_locations_model_to_world(
            #     hyp_locations_wrt_model,
            #     self.object_model.model_origin_wrt_world,
            #     self.ground_truth_position,
            #     self.ground_truth_rotation,
            # )
            # hyp_object_orientations = transform_orientations_model_to_world(
            #     hyp_object_orientations_wrt_model,
            #     self.ground_truth_rotation,
            #     row_vector_format=False,
            # )
```

CleanShot 2025-09-02 at 11 59 13

I chose banana because it was easiest to see. In the visualization, the object model (gray points) is in its randomly rotated test orientation ([114.67, 334.6, 267.3]). However, the hypotheses points/locations are in the "standard/canonical" locations, as though the model was in [0, 0, 0]. I think this is why I thought the hypothesized locations are with respect to model, and for visualization need to be rotated by the ground-truth object orientation (and same reasoning for hypothesized object rotations).

tools/plot/data_models.py
hlee9212: Added. :)

tools/plot/data_models.py
hlee9212: Ah, I think I misunderstood you the first time. I thought the suggestion was to delete the entire list comprehension and just have the return statement. Now I believe you were referring to just updating the return statement - if so, updated. :)

tools/plot/data_models.py
hlee9212: Updated docstring and also updated the name to `orientations_in_row_vector_format` to make it clearer.

tools/plot/episode_loader.py
hlee9212: I have taken the first option and moved it to docstring.

tools/plot/episode_loader.py
hlee9212: Ah, actually, I think it shouldn't take the episode_id argument, since `self.episode_id` is defined (probably a refactoring error on my end - I think I had a global load_episode_data function with this argument..)

tools/plot/episode_loader.py
hlee9212: Good catch! I had separate calls when I was plotting MLH, but didn't update/refactor properly when I switched to plotting highest evidence hypothesis. 😂

tools/plot/episode_loader.py
hlee9212: Nice catch. :) I also agree / have the same preference (I have a `_clear_visualizations()` function that does this) - in this case I don't need the line since the visualization is for single episode only.

tools/plot/episode_loader.py
hlee9212: Created TimestepMappingError.

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Created TypedDict.

tools/plot/episode_loader.py
```diff
@@ -17,6 +17,10 @@
 import numpy as np
 import torch
 
+
+class TimestepMappingError(Exception):
```
jeremyshoemaker: This should inherit from ValueError, since it's a special case of that. This allows a direct or indirect caller further up the call stack to catch all ValueErrors and still catch this one, or be more specific and catch just this one. When it inherits from Exception, we lose that sort flexibility.
```suggestion
class TimestepMappingError(ValueError):
```

tools/plot/episode_loader.py
nielsleadholm: Thanks for checking that, I added some notes to our 1:1 which we can discuss over Slack when you're feeling better. But basically I think it comes down to deciding what we want to visualize, and being consistent about that (hypotheses as they exist in the external world, or hypotheses as they exist in the internal reference frame). Right now we're kind of doing a mix.

tools/plot/episode_loader.py
hlee9212: Updated. Thanks!

tools/plot/episode_loader.py
hlee9212: Hi @nielsleadholm, I checked your notes from the 1:1 document in Shortcut. If I am understanding you correctly, it sounds like we are going for visualizing hypotheses' locations super-imposed to internal reference frame. I think this is the current code, so I'll leave it as is. If this sounds good, feel free to mark conversation as resolved. :)

tools/plot/data_models.py
hlee9212: Hi @nielsleadholm, as per the [other comment](https://github.com/thousandbrainsproject/tbp.monty/pull/399#discussion_r2314023881), I think we established that we need this function to visualize hypotheses's locations super-imposed to locaitons in the internal reference frame. Feel free to mark this conversation as resolved if this sounds good. :)

tools/plot/episode_loader.py
hlee9212: Hi @ramyamounir, yes, the SM steps are equivalent to the indices where the boolean mask equals True. I think both ways are clear to me, but let me know if you think one is more clear than the other to update the code.

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: ~Note: Updated so that we check if we have the "raw_obs" in detailed_run_stats.json.~

Just realized 20250915 8:41 PM that `save_raw_obs` needs to be indeed `True`. Having this false doesn't save any sensor_module information, i.e. "SM_0" key is missing in detailed_run_stats.json. Will update tomorrow.

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Got it. :) I updated the docstring for `_add_sensed_pose_vector_arrows`.

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Added.

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Thanks Ramy! And agree, I'll create a subdirectory with the name of the plot filename.

hlee9212:
Hi @ramyamounir and @nielsleadholm, I think I have gone through all the comments again and made all the last minor updates. :) Please resolve comments as you go if they look good and let me know if there are any further changes I should make!

tools/plot/hypothesis_oorf_visualizer.py
hlee9212: Update: Added that we need `save_raw_obs` to be True.

tools/plot/data_models.py
nielsleadholm: Yeah per our discussion today, I would describe it as "hypothesized locations, super-imposed in the internal reference frame, which itself is super-imposed in the world space". This enables us to both i) visualize where the actual object is in the world, and where the sensor is actually present, relative to the object, as well as ii) what the hypothesis space looks like in the internal reference frame. I would clarify to users that the hypothesized rotations are *not* reflected by rotating multiple copies of the model and showing these in world coordinates. Instead, the movement in the internal reference frame that is visualized is influenced by the hypothesized rotation of the object.

hlee9212: Great, thanks! I added the note to `_initialize_hypotheses_data()` function.

ramyamounir (approved):
Great, thanks for making all these changes @hlee9212. Maybe before closing this PR, you can rename it from `feat` to `chore`. According to [RFC10](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0010_conventional_commits.md#types), we use `feat` only for the code under `src`.

When you move this PR into `tbp.plot`, the only change in code would be in how the plot is [registered with decorators](https://github.com/thousandbrainsproject/tbp.plot/blob/main/README.md#adding-new-plots). 
```python
# src/tbp/plot/plots/my_new_plot.py
import argparse
from tbp.plot.registry import register, attach_args

@register("my_new_plot", description="Example of a new plot")
def main(arg1: str, arg2: int) -> int:
    print(f"Running my_new_plot with arg1={arg1} and arg2={arg2}")
    # your plotting logic here...
    return 0

@attach_args("my_new_plot")
def add_arguments(p: argparse.ArgumentParser) -> None:
    p.add_argument("--arg1", required=True, help="The first argument.")
    p.add_argument("--arg2", type=int, default=100, help="The second argument.")
```

hlee9212:
Thanks @ramyamounir and @nielsleadholm! I'll be closing this PR and opening a new one in `tbp.plots` with the decorators @ramyamounir mentioned.

PR #396, opened 2025-07-24
feat: enable logging of resampling stats

ramyamounir:
This PR is part 1 of adding a visualization that allows us to inspect the hypothesis space at every step of every episode for a Monty unsupervised inference experiment.

In this PR, I'm adding the code for logging resampling statistics such as which hypotheses have been added or removed at every step, the evidence values, slopes and ages, rotations, pose errors, etc.

Since this is a lot of information that we're logging I added a `ResamplingHypothesesUpdater` flag named `save_stats` which must be set to true in the configs to save all this extra logging information. This is similar to `save_raw_obs` for the SMs.

In a follow up PR, I will add the interactive visualization to make use of this logged information.

PR #400, opened 2025-07-25
feat!: collect telemetry from hypotheses updaters

tristanls:
This is an alternate proposal to implement the functionality in #396.

The pull request is marked `feat!` due to a breaking change, which changed the `EvidenceSlopeTracker.window_size` parameter default.

The overall design for collecting telemetry from hypotheses updaters is to update the protocol to return a tuple and then, in the case of `ResamplingHypothesesUpdater`, include telemetry in the returned tuple if configured to do so.

This proposal attempts to minimize coupling between code.

The `HypothesesUpdater` protocol update now requires that _any_ updater may return some telemetry, with the `DefaultHypothesesUpdater` returning `None`.

The `EvidenceGraphLM`, that uses hypotheses updaters, now collects any telemetry emitted by them.
```python
        if hypotheses_update_telemetry is not None:
            self.hypotheses_updater_telemetry[graph_id] = hypotheses_update_telemetry
```
This way, no matter what implementation of hypotheses updater is used, if it emits telemetry, it will be collected.

The learning module collects telemetry because of the coupling between Monty and MontyExperiment, where telemetry is collected via `_add_detailed_stats` call. Additionally, we _do not want_ the learning module to assume that it has a specific type of hypotheses updater, nor do we want it to reach into that one specific type of hypotheses updater and then reach into its tracker to do some computation.

I attempted to minimize changes to `TheoreticalLimitLMLoggingMixin`. The difference in implementation here vs #396 is that `TheoreticalLimitLMLoggingMixin` does not assume a specific telemetry format of the channel telemetry (aside from the generic dictionary of Any). It accepts channel telemetry and only appends what is in its scope to append. There is some weak coupling here in that it is technically possible for `TheoreticalLimitLMLoggingMixin` to override keys in channel telemetry, but addressing that is beyond the scope of this pull request.

Instead of reaching into the internals of `EvidenceSlopeTracker`, it now has `hyp_ages` method that can be called. Similarly, `_calculate_slopes` has been renamed to `calculate_slopes` as it is now part of its public API and called outside the class.

I thought that adding `compute_pose_errors` and reusing it in `compute_pose_error` was a better way to offer the quite different functionality, instead of using a parameter toggle.

Lastly, the use of the term "telemetry" is intentional as a marker for future refactoring. I am hoping that one day we'll end up emitting telemetry in a similar way we do `logger.info()` calls, out of band of the main code logic. When we get there, anything with "telemetry" in it will need to be refactored, and so this makes it easier to find (possibly maybe).

### DIFF
--- a/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
@@ -10,7 +10,7 @@
 from __future__ import annotations
 
 import logging
-from typing import Literal, Protocol
+from typing import Any, Dict, Literal, Optional, Protocol
 
 import numpy as np
 from scipy.spatial.transform import Rotation
@@ -44,6 +44,8 @@ from tbp.monty.frameworks.utils.spatial_arithmetics import (
 
 logger = logging.getLogger(__name__)
 
+HypothesesUpdateTelemetry = Optional[Dict[str, Any]]
+HypothesesUpdaterTelemetry = Dict[str, Any]
 
 class HypothesesUpdater(Protocol):
     def update_hypotheses(
@@ -54,7 +56,7 @@ class HypothesesUpdater(Protocol):
         graph_id: str,
         mapper: ChannelMapper,
         evidence_update_threshold: float,
-    ) -> list[ChannelHypotheses]:
+    ) -> tuple[list[ChannelHypotheses], HypothesesUpdateTelemetry]:
         """Update hypotheses based on sensor displacement and sensed features.
 
         Args:
@@ -169,7 +171,7 @@ class DefaultHypothesesUpdater:
         graph_id: str,
         mapper: ChannelMapper,
         evidence_update_threshold: float,
-    ) -> list[ChannelHypotheses]:
+    ) -> tuple[list[ChannelHypotheses], HypothesesUpdateTelemetry]:
         """Update hypotheses based on sensor displacement and sensed features.
 
         Updates existing hypothesis space or initializes a new hypothesis space
@@ -243,7 +245,7 @@ class DefaultHypothesesUpdater:
 
             hypotheses_updates.append(channel_possible_hypotheses)
 
-        return hypotheses_updates
+        return hypotheses_updates, None
 
     def _get_all_informed_possible_poses(
         self, graph_id: str, sensed_channel_features: dict, input_channel: str
--- a/src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
@@ -9,9 +9,11 @@
 
 from __future__ import annotations
 
-from typing import Literal, Tuple, Type
+from dataclasses import asdict, dataclass
+from typing import Any, Literal, Tuple, Type
 
 import numpy as np
+import numpy.typing as npt
 from scipy.spatial.transform import Rotation
 
 from tbp.monty.frameworks.models.evidence_matching.feature_evidence.calculator import (
@@ -33,6 +35,7 @@ from tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer import (
     DefaultHypothesesDisplacer,
 )
 from tbp.monty.frameworks.models.evidence_matching.hypotheses_updater import (
+    HypothesesUpdateTelemetry,
     all_usable_input_channels,
 )
 from tbp.monty.frameworks.utils.evidence_matching import (
@@ -46,6 +49,24 @@ from tbp.monty.frameworks.utils.graph_matching_utils import (
 from tbp.monty.frameworks.utils.spatial_arithmetics import (
     align_multiple_orthonormal_vectors,
 )
+
+
+@dataclass
+class ChannelHypothesesResamplingTelemetry:
+    """Hypotheses resampling telemetry for a channel.
+
+    For a given input channel, this class stores which hypotheses were removed or
+    added at the current step.
+
+    Note:
+        TODO: Additional description here regarding availability of hypotheses
+        identified by `removed_ids`.
+    """
+
+    added_ids: npt.NDArray[np.int_]
+    ages: npt.NDArray[np.int_]
+    evidence_slopes: npt.NDArray[np.float64]
+    removed_ids: npt.NDArray[np.int_]
 
 
 class ResamplingHypothesesUpdater:
@@ -89,6 +110,7 @@ class ResamplingHypothesesUpdater:
         ),
         hypotheses_count_multiplier: float = 1.0,
         hypotheses_existing_to_new_ratio: float = 0.1,
+        include_telemetry: bool = False,
         initial_possible_poses: Literal["uniform", "informed"]
         | list[Rotation] = "informed",
         max_nneighbors: int = 3,
@@ -122,6 +144,9 @@ class ResamplingHypothesesUpdater:
             hypotheses_existing_to_new_ratio: Controls the proportion of the
                 existing vs. newly sampled hypotheses during resampling. Defaults to
                 0.0.
+            include_telemetry: Flag to control if we want to calculate and return the
+                resampling telemetry in the `update_hypotheses` method. Defaults to
+                False.
             initial_possible_poses: Initial
                 possible poses that should be tested for. Defaults to "informed".
             max_nneighbors: Maximum number of nearest neighbors to consider in the
@@ -147,6 +172,7 @@ class ResamplingHypothesesUpdater:
         self.feature_weights = feature_weights
         self.features_for_matching_selector = features_for_matching_selector
         self.graph_memory = graph_memory
+        self.include_telemetry = include_telemetry
         self.initial_possible_poses = get_initial_possible_poses(initial_possible_poses)
         self.tolerances = tolerances
         self.umbilical_num_poses = umbilical_num_poses
@@ -189,7 +215,7 @@ class ResamplingHypothesesUpdater:
         graph_id: str,
         mapper: ChannelMapper,
         evidence_update_threshold: float,
-    ) -> list[ChannelHypotheses]:
+    ) -> tuple[list[ChannelHypotheses], HypothesesUpdateTelemetry]:
         """Update hypotheses based on sensor displacement and sensed features.
 
         Updates existing hypothesis space or initializes a new hypothesis space
@@ -208,7 +234,17 @@ class ResamplingHypothesesUpdater:
             evidence_update_threshold: Evidence update threshold.
 
         Returns:
-            The list of hypotheses updates to be applied to each input channel.
+            A tuple containing the list of hypotheses updates to be applied to each
+            input channel and hypotheses update telemetry for analysis. The hypotheses
+            update telemetry is a dictionary containing:
+                - added_ids: IDs of hypotheses added during resampling at the current
+                    timestep.
+                - ages: The ages of the hypotheses as tracked by the
+                    `EvidenceSlopeTracker`.
+                - evidence_slopes: The slopes extracted from the `EvidenceSlopeTracker`.
+                - removed_ids: IDs of hypotheses removed during resampling. Note that
+                    these IDs can only be used to index hypotheses from the previous
+                    timestep.
         """
         # Initialize a `EvidenceSlopeTracker` to keep track of evidence slopes
         # for hypotheses of a specific graph_id
@@ -221,6 +257,7 @@ class ResamplingHypothesesUpdater:
         )
 
         hypotheses_updates = []
+        resampling_telemetry: dict[str, Any] = {}
 
         for input_channel in input_channels_to_use:
             # Calculate sample count for each type
@@ -233,7 +270,7 @@ class ResamplingHypothesesUpdater:
             )
 
             # Sample hypotheses based on their type
-            existing_hypotheses = self._sample_existing(
+            existing_hypotheses, remove_ids = self._sample_existing(
                 existing_count=existing_count,
                 hypotheses=hypotheses,
                 input_channel=input_channel,
@@ -275,10 +312,29 @@ class ResamplingHypothesesUpdater:
             )
             hypotheses_updates.append(channel_hypotheses)
 
+            if self.include_telemetry:
+                resampling_telemetry[input_channel] = asdict(
+                    ChannelHypothesesResamplingTelemetry(
+                        added_ids=(
+                            np.arange(len(channel_hypotheses.evidence))[
+                                -len(informed_hypotheses.evidence) :
+                            ]
+                            if len(informed_hypotheses.evidence) > 0
+                            else np.array([], dtype=np.int_)
+                        ),
+                        ages=tracker.hyp_ages(input_channel),
+                        evidence_slopes=tracker.calculate_slopes(input_channel),
+                        removed_ids=remove_ids,
+                    )
+                )
+
             # Update tracker evidence
             tracker.update(channel_hypotheses.evidence, input_channel)
 
-        return hypotheses_updates
+        return (
+            hypotheses_updates,
+            resampling_telemetry if self.include_telemetry else None,
+        )
 
     def _num_hyps_per_node(self, channel_features: dict) -> int:
         """Calculate the number of hypotheses per node.
@@ -383,7 +439,7 @@ class ResamplingHypothesesUpdater:
         input_channel: str,
         mapper: ChannelMapper,
         tracker: EvidenceSlopeTracker,
-    ) -> ChannelHypotheses:
+    ) -> tuple[ChannelHypotheses, npt.NDArray[np.int_]]:
         """Samples the specified number of existing hypotheses to retain.
 
         Args:
@@ -401,14 +457,16 @@ class ResamplingHypothesesUpdater:
         # Return empty arrays for no hypotheses to sample
         if existing_count == 0:
             # Clear all channel hypotheses from the tracker
+            remove_ids = np.arange(tracker.total_size(input_channel))
             tracker.clear_hyp(input_channel)
 
-            return ChannelHypotheses(
+            channel_hypotheses = ChannelHypotheses(
                 input_channel=input_channel,
                 locations=np.zeros((0, 3)),
                 poses=np.zeros((0, 3, 3)),
                 evidence=np.zeros(0),
             )
+            return channel_hypotheses, remove_ids
 
         keep_ids, remove_ids = tracker.calculate_keep_and_remove_ids(
             num_keep=existing_count,
@@ -419,12 +477,13 @@ class ResamplingHypothesesUpdater:
         tracker.remove_hyp(remove_ids, input_channel)
 
         channel_hypotheses = mapper.extract_hypotheses(hypotheses, input_channel)
-        return ChannelHypotheses(
+        maintained_channel_hypotheses = ChannelHypotheses(
             input_channel=channel_hypotheses.input_channel,
             locations=channel_hypotheses.locations[keep_ids],
             poses=channel_hypotheses.poses[keep_ids],
             evidence=channel_hypotheses.evidence[keep_ids],
         )
+        return maintained_channel_hypotheses, remove_ids
 
     def _sample_informed(
         self,
--- a/src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
+++ b/src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
@@ -7,6 +7,8 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
+from __future__ import annotations
+
 from typing import Any, Dict
 
 from scipy.spatial.transform import Rotation
@@ -14,7 +16,10 @@ from scipy.spatial.transform import Rotation
 from tbp.monty.frameworks.models.evidence_matching.learning_module import (
     EvidenceGraphLM,
 )
-from tbp.monty.frameworks.utils.logging_utils import compute_pose_error
+from tbp.monty.frameworks.utils.logging_utils import (
+    compute_pose_error,
+    compute_pose_errors,
+)
 
 
 class TheoreticalLimitLMLoggingMixin:
@@ -64,6 +69,54 @@ class TheoreticalLimitLMLoggingMixin:
             self._theoretical_limit_target_object_pose_error()
         )
         stats["target_object_pose_error"] = self._mlh_target_object_pose_error()
+        stats["resampling_telemetry"] = self._resampling_telemetry()
+        return stats
+
+    def _resampling_telemetry(self) -> dict[str, dict[str, dict[str, Any]]]:
+        """Compile resampling telemetry across all objects and input channels.
+
+        Returns:
+            A nested dictionary of {graph_id: {input_channel: resampling_telemetry}}.
+        """
+        stats = {}
+        for graph_id, graph_telemetry in self.hypotheses_updater_telemetry.items():
+            stats[graph_id] = {
+                input_channel: self._channel_resampling_telemetry(
+                    graph_id, input_channel, channel_telemetry
+                )
+                for input_channel, channel_telemetry in graph_telemetry.items()
+            }
+        return stats
+
+    def _channel_resampling_telemetry(
+        self, graph_id: str, input_channel: str, channel_telemetry: dict[str, Any]
+    ) -> dict[str, Any]:
+        """Assemble resampling telemetry for specific graph ID and input channel.
+
+        Args:
+            graph_id: The graph ID.
+            input_channel: The input channel.
+            channel_telemetry: Telemetry for the specific input channel.
+
+        Returns:
+            A dictionary containing telemetry in `channel_telemetry` and:
+                - evidence: The hypotheses evidence scores.
+                - rotations: Rotations of the hypotheses. Note that the buffer encoder
+                    will encode those as euler "xyz" rotations in degrees.
+                - pose_errors: Rotation errors relative to the target pose.
+        """
+        mapper = self.channel_hypothesis_mapping[graph_id]
+        channel_rotations = mapper.extract(self.possible_poses[graph_id], input_channel)
+        channel_rotations_inv = Rotation.from_matrix(channel_rotations).inv()
+        channel_evidence = mapper.extract(self.evidence[graph_id], input_channel)
+
+        stats: dict[str, Any] = channel_telemetry.copy()
+        stats["evidence"] = channel_evidence
+        stats["rotations"] = channel_rotations_inv
+        stats["pose_errors"] = compute_pose_errors(
+            channel_rotations_inv,
+            Rotation.from_quat(self.primary_target_rotation_quat),
+        )
         return stats
 
     def _theoretical_limit_target_object_pose_error(self) -> float:
--- a/src/tbp/monty/frameworks/utils/evidence_matching.py
+++ b/src/tbp/monty/frameworks/utils/evidence_matching.py
@@ -278,7 +278,7 @@ class EvidenceSlopeTracker:
         hyp_age: Maps channel names to hypothesis age counters.
     """
 
-    def __init__(self, window_size: int = 3, min_age: int = 5) -> None:
+    def __init__(self, window_size: int = 12, min_age: int = 5) -> None:
         """Initializes the EvidenceSlopeTracker.
 
         Args:
@@ -333,6 +333,14 @@ class EvidenceSlopeTracker:
             )
             self.hyp_age[channel] = np.concatenate((self.hyp_age[channel], new_age))
 
+    def hyp_ages(self, channel: str) -> npt.NDArray[np.int_]:
+        """Returns the ages of hypotheses in a channel.
+
+        Args:
+            channel: Name of the input channel.
+        """
+        return self.hyp_age[channel]
+
     def update(self, values: npt.NDArray[np.float64], channel: str) -> None:
         """Updates all hypotheses in a channel with new evidence values.
 
@@ -361,7 +369,7 @@ class EvidenceSlopeTracker:
         # Increment age
         self.hyp_age[channel] += 1
 
-    def _calculate_slopes(self, channel: str) -> npt.NDArray[np.float64]:
+    def calculate_slopes(self, channel: str) -> npt.NDArray[np.float64]:
         """Computes the average slope of hypotheses in a channel.
 
         This method calculates the slope of the evidence signal for each hypothesis by
@@ -442,7 +450,7 @@ class EvidenceSlopeTracker:
 
         # Retrieve valid slopes and sort them
         removable_mask = self.removable_indices_mask(channel)
-        slopes = self._calculate_slopes(channel)
+        slopes = self.calculate_slopes(channel)
         removable_slopes = slopes[removable_mask]
         removable_ids = total_ids[removable_mask]
         sorted_indices = np.argsort(removable_slopes)
--- a/src/tbp/monty/frameworks/utils/logging_utils.py
+++ b/src/tbp/monty/frameworks/utils/logging_utils.py
@@ -8,6 +8,8 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
+from __future__ import annotations
+
 import copy
 import json
 import logging
@@ -18,6 +20,7 @@ from pathlib import Path
 from sys import getsizeof
 
 import numpy as np
+import numpy.typing as npt
 import pandas as pd
 import quaternion
 import torch
@@ -388,11 +391,10 @@ def get_time_stats(all_ds, all_conditions) -> pd.DataFrame:
     time_stats.columns = ["model_type", "time", "step"]
     return time_stats
 
-
-def compute_pose_error(
+def compute_pose_errors(
     predicted_rotation: Rotation, target_rotation: Rotation
-) -> float:
-    """Computes the minimum angular pose error between predicted and target rotations.
+) -> npt.NDArray[np.float64] | float:
+    """Computes the angular pose errors between predicted and target rotations.
 
     Both inputs must be instances of `scipy.spatial.transform.Rotation`. The
     `predicted_rotation` may contain a single rotation or a list of rotations,
@@ -400,7 +402,7 @@ def compute_pose_error(
 
     The pose error is defined as the geodesic distance on SO(3) — the angle of the
     relative rotation between predicted and target. If `predicted_rotation` contains
-    multiple rotations, this function returns the minimum error among them.
+    multiple rotations, this function returns the errors among them.
 
     Note that the `.inv()` operation in this method is due to how geodesic distance
     between two rotations is calculated, not a side-effect of whether the target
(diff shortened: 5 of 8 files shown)


tristanls:
@ramyamounir, I asked for your review because this is the alternate proposal I mentioned. If you think this implements your prototype correctly, we can add others to the review.

hlee9212:
Just gonna mention myself @hlee9212 here to take a look because I'm also starting work on visualization.

**Update**: Looked through the implementation. Will the telemetry information be saved in `detailed_stats.json`? Or maybe the visualization will be "live" and one can just use the telemetry data instead of loading from a saved file and doing `DataExtractor`?

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -44,6 +44,9 @@
 
 logger = logging.getLogger(__name__)
 
+HypothesesUpdateTelemetry = Optional[Dict[str, Any]]
```
ramyamounir, line 47: For my understanding, would this be equivalent to `dict[str, Any] | None`? If so, when do I use which version? I'm new to type annotations.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -275,10 +312,29 @@ def update_hypotheses(
...
+                    ChannelHypothesesResamplingTelemetry(
+                        added_ids=(
+                            np.arange(len(channel_hypotheses.evidence))[
+                                -len(informed_hypotheses.evidence) :
+                            ]
+                            if len(informed_hypotheses.evidence) > 0
+                            else np.array([], dtype=np.int_)
+                        ),
+                        ages=tracker.hyp_ages(input_channel),
+                        evidence_slopes=tracker.calculate_slopes(input_channel),
```
ramyamounir, line 326: This is a very nice catch, I should have done that in my PR.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -401,14 +457,16 @@ def _sample_existing(
...
             tracker.clear_hyp(input_channel)
 
-            return ChannelHypotheses(
+            channel_hypotheses = ChannelHypotheses(
                 input_channel=input_channel,
                 locations=np.zeros((0, 3)),
                 poses=np.zeros((0, 3, 3)),
                 evidence=np.zeros(0),
             )
+            return channel_hypotheses, remove_ids
```
ramyamounir, line 470: Could you also update the `Returns` section of the docstring to include `remove_ids`? I must have missed it in my PR.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -275,10 +312,29 @@ def update_hypotheses(
             )
             hypotheses_updates.append(channel_hypotheses)
 
+            if self.include_telemetry:
+                resampling_telemetry[input_channel] = asdict(
```
ramyamounir, line 316: I'm curious about using the `asdict` here as opposed to later in the later in the logging mixin. Wouldn't it be better to keep it as the dataclass object for more specific type checking? We can later convert it to a dictionary to initialize the stats dict. Something similar to the code below?

```python
    def _channel_resampling_telemetry(
        self, graph_id: str, input_channel: str, channel_telemetry: ChannelHypothesesResamplingTelemetry
    ) -> dict[str, Any]:
        """Assemble resampling telemetry for specific graph ID and input channel.

        Args:
            graph_id: The graph ID.
            input_channel: The input channel.
            channel_telemetry: Telemetry for the specific input channel.

        Returns:
            A dictionary containing telemetry in `channel_telemetry` and:
                - evidence: The hypotheses evidence scores.
                - rotations: Rotations of the hypotheses. Note that the buffer encoder
                    will encode those as euler "xyz" rotations in degrees.
                - pose_errors: Rotation errors relative to the target pose.
        """
        channel_rotations = mapper.extract(self.possible_poses[graph_id], input_channel)
        channel_rotations_inv = Rotation.from_matrix(channel_rotations).inv()
        channel_evidence = mapper.extract(self.evidence[graph_id], input_channel)

        stats: dict[str, Any] = asdict(channel_telemetry).copy()
        stats["evidence"] = channel_evidence
        stats["rotations"] = channel_rotations_inv
        stats["pose_errors"] = compute_pose_errors(
            channel_rotations_inv,
            Rotation.from_quat(self.primary_target_rotation_quat),
        )
        return stats
```

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -48,6 +51,24 @@
...
+    """Hypotheses resampling telemetry for a channel.
+
+    For a given input channel, this class stores which hypotheses were removed or
+    added at the current step.
+
+    Note:
+        TODO: Additional description here regarding availability of hypotheses
+        identified by `removed_ids`.
+    """
+
```
ramyamounir, line 65: I see that you removed the `input_channel` field from here. Would it be better to mirror the `ChannelHypotheses` [design](https://github.com/thousandbrainsproject/tbp.monty/blob/290390d22d4378cd723501bf54b20f5dd36e7848/src/tbp/monty/frameworks/models/evidence_matching/hypotheses.py?plain=1#L27-L31), so that later in the `_update_hypotheses` function we return a
```python
-> tuple[list[ChannelHypotheses], list[ChannelHypothesesResamplingTelemetry]] 
OR
-> tuple[list[ChannelHypotheses], list[Any]] # since we do not want to assume a specific type of HypothesesTelemetry.
```
?

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -64,6 +69,54 @@ def _add_detailed_stats(self, stats: Dict[str, Any]) -> Dict[str, Any]:
...
+                - rotations: Rotations of the hypotheses. Note that the buffer encoder
+                    will encode those as euler "xyz" rotations in degrees.
+                - pose_errors: Rotation errors relative to the target pose.
+        """
+        mapper = self.channel_hypothesis_mapping[graph_id]
+        channel_rotations = mapper.extract(self.possible_poses[graph_id], input_channel)
+        channel_rotations_inv = Rotation.from_matrix(channel_rotations).inv()
+        channel_evidence = mapper.extract(self.evidence[graph_id], input_channel)
+
+        stats: dict[str, Any] = channel_telemetry.copy()
```
ramyamounir: Really nice, it looks so much cleaner. Thanks.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -278,7 +278,7 @@ class EvidenceSlopeTracker:
         hyp_age: Maps channel names to hypothesis age counters.
     """
 
-    def __init__(self, window_size: int = 3, min_age: int = 5) -> None:
+    def __init__(self, window_size: int = 12, min_age: int = 5) -> None:
```
ramyamounir: This is the breaking change you mentioned in the description. It could be better to discard it from here, I could open another small PR with this change and the effect of it on the benchmarks, docs, etc. We might have to update the unsupervised inference benchmark results because of this change. This shouldn't have been part of my previous PR, thanks for noticing that it's a breaking change.

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -333,6 +333,14 @@ def add_hyp(self, num_new_hyp: int, channel: str) -> None:
...
             self.hyp_age[channel] = np.concatenate((self.hyp_age[channel], new_age))
 
+    def hyp_ages(self, channel: str) -> npt.NDArray[np.int_]:
+        """Returns the ages of hypotheses in a channel.
+
+        Args:
+            channel: Name of the input channel.
+        """
+        return self.hyp_age[channel]
+
```
ramyamounir, line 343: Nice, there are a few instances of accessing the `hyp_age[channel]` directly in the [unit tests](https://github.com/thousandbrainsproject/tbp.monty/blob/290390d22d4378cd723501bf54b20f5dd36e7848/tests/unit/frameworks/utils/evidence_matching_test.py?plain=1#L277). We could probably update those too.

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -416,7 +440,7 @@ def compute_pose_error(
     Returns:
         The minimum angular error in radians.
     """
-    error = np.min((predicted_rotation * target_rotation.inv()).magnitude())
+    error = np.min(compute_pose_errors(predicted_rotation, target_rotation))
```
ramyamounir, line 443: Nice, I like this change :slightly_smiling_face:

ramyamounir (commented):
Really nice, thanks for taking the time to implement this change. 

I did consider this option of returning the logged statistics back to the parent LM vs. reaching into the composed functionality and extracting the logged statistics. I had two concerns with the former approach (one major and one minor):
- I assume that getting telemetry from a deeper composed class, such as the `HypothesesDisplacer`, would follow the same approach and return its telemetry to the Updater, and it gets included in the updater's telemetry (as a nested dictionary) then sent back to the parent LM. This works well if all the intermediate composed levels already follow the same protocol of sending back `Telemetry | None`. But if I have to implement this for the Displacer without it being already implemented in the Updater, I'd have to go implement it for the updater first, which may not sound much now, but it's definitely a lot of work for 4 or 5 levels of nested classes. This change would include updating the protocols and all the classes that follow these protocols for every level. To avoid this, we have to adopt this telemetry approach by default whenever extracting any functionality. I'm okay with this change btw.
- The current approach of sending back telemetry or None could break the hierarchical chain of telemetry. If I have `include_telemetry=False` for the Updater, but `include_telemetry=True` for the Displacer, the updater would not collect telemetry from it's child displacer even though it was defined in the configs. But this is probably an easy fix.

Note that I requested minor changes (e.g., unit tests or docstrings) in my review. Feel free to use the :eyes: to mean "do it yourself". I haven't tried it before, but I think maintainers are allowed to push commits to your PR branch if you've allowed it when creating the PR.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
tristanls: Correct, I tried to use `dict[str, Any] | None`, but using `|` in a type alias appears to require Python beyond 3.8 even with the `from __future__` import. So, I used `Optional` because other formulations did not work. Give it a try and see if you get an error, it's possible I made a mistake here.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
tristanls: I initially did what you suggested, but then I noticed it requires the Mixin (and therefore the learning module) to know that the `channel_telemetry` has the `ChannelHypothesesResamplingTelemetry` type. This means that the Mixin knows the specific hypotheses updater class being used, which it is not allowed to know. If it knows, then we have coupling, and when we swap out an updater, things will break unexpectedly. It will be unexpected because our protocol says that anything that conforms to the protocol can be used.

So, I used the dataclass to ensure I didn't forget to populate all the fields. Still, I wanted a generic return value, i.e., `dict[str, Any]` so that the Mixin (and therefore the learning module) would not have to be coupled to the specific hypotheses updater implementation, but could work with the protocol and telemetry, which could be substituted for any other one. This desire to work with any telemetry format (as long as it is `dict[str, Any]`) is why Mixin doesn't assume anything beyond dictionary structure or attempt to read it, since it can't know the field names.

tristanls: Another way to think about it, having `channel_telemetry: ChannelHypothesesResamplingTelemetry` would imply that the `HypothesesUpdater(Protocol)` be:
```python
class HypothesesUpdater(Protocol):
    def update_hypotheses(
        self,
        hypotheses: Hypotheses,
        features: dict,
        displacements: dict | None,
        graph_id: str,
        mapper: ChannelMapper,
        evidence_update_threshold: float,
    ) -> tuple[list[ChannelHypotheses], dict[str, ChannelHypothesesResamplingTelemetry]]:
        """Update hypotheses based on sensor displacement and sensed features."""
```
See the problem?

We are defining a protocol and dictating its implementation. Those are conflicting declarations. We want protocols to define generic interfaces. We want specific implementations to be swappable wherever a protocol is used.

UPDATE: Sorry, it should have been `dict[str, ChannelHypothesesResamplingTelemetry]` instead of `ChannelHypothesesResamplingTelemetry` in the example above, updated.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
tristanls: I was trying to reduce coupling, but couldn't get away from it, so I think your suggestion here would be equivalent and putting `input_channel` back into `ChannelHypothesesResamplingTelemetry` is fine.

The coupling problem (~~pre-existing, so~~ not addressed) is that the Mixin only works with one hypotheses updater :/ This hardcodes the Mixin to only work with resampling hypotheses updater. I didn't want to rewrite the mixin though:
```python
    def _resampling_telemetry(self) -> dict[str, dict[str, dict[str, Any]]]:
        """Compile resampling telemetry across all objects and input channels.

        Returns:
            A nested dictionary of {graph_id: {input_channel: resampling_telemetry}}.
        """
        stats = {}
        for graph_id, graph_telemetry in self.hypotheses_updater_telemetry.items():
            stats[graph_id] = {
                input_channel: self._channel_resampling_telemetry(
                    graph_id, input_channel, channel_telemetry
                )
                for input_channel, channel_telemetry in graph_telemetry.items()
            }
        return stats
```

tristanls: I think I have an idea 💡 ...

ramyamounir:
> Looked through the implementation. Will the telemetry information be saved in `detailed_stats.json`? Or maybe the visualization will be "live" and one can just use the telemetry data instead of loading from a saved file and doing `DataExtractor`?

@hlee9212 

The telemetry data will be available in the `EvidenceGraphLM` class, the `_add_detailed_stats` function adds them to the `FeaturesAtLocationBuffer` so that they can be logged with the `DetailedJSONHandler` as a json file. If you create a live visualization, you can use the telemetry data directly from LM class.

Warning: During the hackathon, we tried using vedo in a live visualization with HabitatSim and it gave us a segmentation fault. But we could still use it with our robot because there was no simulator.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
tristanls: My idea aside... I think I remember now why I did this.

Whatever telemetry `update_hypotheses` returns, we want it grouped by the input channel. Returning a list would work, but in that case, we would have to specify a protocol for the telemetry that we could use to interrogate what input channel the telemetry corresponds to:
```python
class InputChannelTelemetry(Protocol):
    @property
    def input_channel(self) -> str: ...
```
Then, we would have (gotta remember that `Optional`):
```python
HypothesesUpdateTelemetry = Optional[list[InputChannelTelemetry]]
```
The return value from `update_hypotheses` would still be:
```python
-> tuple[list[ChannelHypotheses], HypothesesUpdateTelemetry]
```
Then, `_resampling_telemetry` takes the shape of:
```python
    def _resampling_telemetry(self) -> dict[str, dict[str, dict[str, Any]]]:
        """Compile resampling telemetry across all objects and input channels.

        Returns:
            A nested dictionary of {graph_id: {input_channel: resampling_telemetry}}.
        """
        stats = {}
        for graph_id, graph_telemetry in self.hypotheses_updater_telemetry.items():
            stats[graph_id] = {
                channel_telemetry.input_channel: self._channel_resampling_telemetry(
                    graph_id, channel_telemetry
                )
                for channel_telemetry in graph_telemetry
            }
        return stats
```

It's pretty much equivalent in functionality. However, using a `dict[str, Any]` and declaring that the keys are input channels retains the same amount of coupling but lets us avoid writing the `InputChannelTelemetry` protocol. Maybe being explicit with the protocol is good.

tristanls: One reason to not include `input_channel` would be that in your original design in #396, you did not include `input_channel` in the `stats`:
```python
stats = {}
stats["remove_ids"] = channel_stats.removed_hypotheses_ids
stats["add_ids"] = channel_stats.added_hypotheses_ids
stats["evidence"] = channel_evidence
stats["evidence_slopes"] = tracker._calculate_slopes(
    channel_stats.input_channel
)
stats["ages"] = tracker.hyp_age[channel_stats.input_channel]
stats["rotations"] = channel_rotations_inv
stats["pose_errors"] = compute_pose_error(
    channel_rotations_inv,
    Rotation.from_quat(self.primary_target_rotation_quat),
    return_min=False,
)
```
The version of the code in this pull request would include `input_channel` in the stats due to:
```python
stats: dict[str, Any] = channel_telemetry.copy()
```

src/tbp/monty/frameworks/utils/evidence_matching.py
tristanls: The unit tests assign values to `hyp_age`, so another method that allows for assignment would be required. Now that I'm looking at those tests, they might be doing too much internal state inspection... but I'll leave them alone for now.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
ramyamounir: Yes, you're right. It wants 3.10 or newer.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
ramyamounir: That all makes sense now, it's because we're calling the `_channel_resampling_telemetry` for all updater types. We should probably change the function name and the stats key in the mixin to be something that doesn't mention resampling then. Maybe:

```python
        ...
        stats["hypotheses_updater_telemetry"] = self._hypotheses_updater_telemetry()
        return stats
```

One more nitpick: I ran it with `include_telemetry=False` and it stored a list of empty dictionaries for each episode - as expected. It's a nitpick but i'm not sure if we want to completely disable the logging or allow the empty dictionaries to be written. They don't take much space.

UPDATE: Scratch what I said before about the coupling in the mixin. Does the type checker still complain if we use Any for the protocol and a more specific type for the classes?

```python
class HypothesesUpdater(Protocol):
    def update_hypotheses(
        self,
        ...
        evidence_update_threshold: float,
    ) -> tuple[list[ChannelHypotheses], dict[str, Any] | None ]:
        """Update hypotheses based on sensor displacement and sensed features."""

class DefaultHypothesesUpdater:
    def update_hypotheses(
        self,
        ...
        evidence_update_threshold: float,
    ) -> tuple[list[ChannelHypotheses], None]:
        """Update hypotheses based on sensor displacement and sensed features."""

class ResamplingHypothesesUpdater:
    def update_hypotheses(
        self,
        ...
        evidence_update_threshold: float,
    ) -> tuple[list[ChannelHypotheses], dict[str, ChannelHypothesesResamplingTelemetry]]:
        """Update hypotheses based on sensor displacement and sensed features."""
```

It looks like you already have it this way and it's not working. I'm probably missing something about why this is not working :thinking:

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
ramyamounir: I was planning to just pop the input channel after we convert it from a `ChannelHypothesesResamplingTelemetry` to a `dict`. Even if this works, it's not better than what you have already implemented. I'm just commenting to check my understanding. What you have already looks great to me :+1: 

```python
    def _resampling_telemetry(self) -> dict[str, dict[str, dict[str, Any]]]:
        """Compile resampling telemetry across all objects and input channels.

        Returns:
            A nested dictionary of {graph_id: {input_channel: resampling_telemetry}}.
        """
        stats = {}
        for graph_id, graph_telemetry in self.hypotheses_updater_telemetry.items():
            stats[graph_id] = {
                channel_telemetry.input_channel: self._channel_resampling_telemetry(
                    graph_id, channel_telemetry
                )
                for channel_telemetry in graph_telemetry
            }
        return stats

    def _channel_resampling_telemetry(
        self, graph_id: str, channel_telemetry: dict[str, Any]
    ) -> dict[str, Any]:
        """Assemble resampling telemetry for specific graph ID and input channel.

        Args:
            graph_id: The graph ID.
            channel_telemetry: Telemetry for the specific input channel.

        Returns:
            A dictionary containing telemetry in `channel_telemetry` and:
                - evidence: The hypotheses evidence scores.
                - rotations: Rotations of the hypotheses. Note that the buffer encoder
                    will encode those as euler "xyz" rotations in degrees.
                - pose_errors: Rotation errors relative to the target pose.
        """
        stats: dict[str, Any] = asdict(channel_telemetry).copy()
        input_channel = stats.pop("input_channel")

        channel_rotations = mapper.extract(self.possible_poses[graph_id], input_channel)
        channel_rotations_inv = Rotation.from_matrix(channel_rotations).inv()
        channel_evidence = mapper.extract(self.evidence[graph_id], input_channel)

        stats["evidence"] = channel_evidence
        stats["rotations"] = channel_rotations_inv
        stats["pose_errors"] = compute_pose_errors(
            channel_rotations_inv,
            Rotation.from_quat(self.primary_target_rotation_quat),
        )
        return stats

```

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
tristanls: > Does the type checker still complain if we use Any for the protocol and a more specific type for the classes?

It should. This violates the Liskov Substitution Principle, which states that a protocol instance should be replaceable by any of its implementations.

`DefaultHypothesesUpdater` returning `tuple[list[ChannelHypotheses], None]` is fine, because `None` is allowed by the protocol.

`ResamplingHypothesesUpdater` is _**not substitutable**_ for the protocol because  `dict[str, ChannelHypothesesResamplingTelemetry]` is _**more specific**_ than `dict[str, Any] | None`.

The concrete protocol implementation can have a _less specific_ type, but not a more restrictive type.

Simpler example:
```python
from __future__ import annotations

from typing import Any, Protocol


class Something(Protocol):
    def accept(self, num: int | str) -> None: ...


class MyInt:
    def accept(self, num: int) -> None:
        print(f"MyInt: {num}")


class MyFloat:
    def accept(self, num: str) -> None:
        print(f"MyFloat: {num}")

class MyAny:
    def accept(self, num: Any) -> None:
        print(f"MyAny: {num}")

def check(acceptor: Something) -> None:
    pass

check(MyInt())  # fails type check
check(MyFloat())  # fails type check
check(MyAny()) # OK, can handle int, or str, or anything else
```

tristanls: > ```python
> stats["hypotheses_updater_telemetry"] = self._hypotheses_updater_telemetry()
> return stats
> ```
Yes, I think renaming it like this will clarify what's more specific and what's more generic. Updated here: https://github.com/thousandbrainsproject/tbp.monty/pull/400/commits/05bbb5cf4adc9f370472482bc339910de13370a6

tristanls: > One more nitpick: I ran it with include_telemetry=False and it stored a list of empty dictionaries for each episode - as expected. It's a nitpick but i'm not sure if we want to completely disable the logging or allow the empty dictionaries to be written. They don't take much space.

Addressed in https://github.com/thousandbrainsproject/tbp.monty/pull/400/commits/eb15bb26e0a82788a53da7a075f8686ee7d973b4

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
tristanls: > ```python
>     def _channel_resampling_telemetry(
>         self, graph_id: str, channel_telemetry: dict[str, Any]
>     ) -> dict[str, Any]:
>         stats: dict[str, Any] = asdict(channel_telemetry).copy()
>         input_channel = stats.pop("input_channel")
>         ...
> ```
This is not the appropriate place for this operation. Note that `channel_telemetry: dict[str, Any]`. There is no reason (from this code's perspective and the information passed into it) to assume at this point that `"input_channel"` is a valid key here. This is an information leak from a detailed implementation (`ChannelHypothesesResamplingTelemetry`) into a more abstract general code that should not have this detailed information (`dict[str, Any]`). If you wanted to pop `"input_channel"`, it should be done in the code that knows that key is there, i.e., the `ResamplingHypothesesUpdater` that constructs `ChannelHypothesesResamplingTelemetry`. But that's what I did by not including it in the first place.

This is also not a great example, because as I mentioned before, there is weak coupling in that we assume we can add `"evidence"`, `"rotations"`, and `"pose_errors"` without overriding what's in `channel_telemetry`. It might actually be better if we namespace things like so:
```python
        stats: dict[str, Any] = {}
        stats["channel_telemetry"] = channel_telemetry.copy()
        stats["evidence"] = channel_evidence
        stats["rotations"] = channel_rotations_inv
        stats["pose_errors"] = compute_pose_errors(
            channel_rotations_inv,
            Rotation.from_quat(self.primary_target_rotation_quat),
        )
        return stats
```
The above change is in https://github.com/thousandbrainsproject/tbp.monty/pull/400/commits/6010e84022a793526a98dd9f3f52bd4a7d6455cb

This is better, as we now treat `channel_telemetry` as a mostly opaque object that it is (`dict[str, Any]`).

src/tbp/monty/frameworks/utils/evidence_matching.py
tristanls: 👍 reverted in: https://github.com/thousandbrainsproject/tbp.monty/pull/400/commits/31dc592810bf6a3a67c78015b693826a9d6c9191

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
tristanls: **note**: I ended up with `stats["hypotheses_updater"] = channel_telemetry.copy()`. I think that's a better nested namespace as it points out where those channel stats came from.

tristanls:
@ramyamounir, requesting another review. I believe I addressed your comments. Please note that in the latest batch of changes, I nested the hypotheses updater telemetry one level deeper in the `stats` object. This will impact the use of the data for visualization. Reasons are in the comment thread above.
```python
        stats: dict[str, Any] = {}
        stats["hypotheses_updater"] = channel_telemetry.copy()  # <- extra nesting/namespacing here
        stats["evidence"] = channel_evidence
        stats["rotations"] = channel_rotations_inv
        stats["pose_errors"] = compute_pose_errors(
            channel_rotations_inv,
            Rotation.from_quat(self.primary_target_rotation_quat),
        )
        return stats
```

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
jeremyshoemaker: It bothers me how much we use dictionaries, especially nested ones all over the place. It makes the typing much harder to do. Improving it here is probably out of scope though.

**Edit**: I see the much larger conversation on this topic below.

tristanls: @jeremyshoemaker, I think I can improve here. I'll give it a shot.

tristanls: @jeremyshoemaker, updated: https://github.com/thousandbrainsproject/tbp.monty/pull/400/commits/0b45e8a0a812f6de93b845b190b9793bc248b713

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -7,14 +7,39 @@
...
+
+
+@dataclass
+class HypothesesUpdaterChannelTelemetry:
+    """Telemetry from HypothesesUpdater for a single input channel."""
+
+    hypotheses_updater: dict[str, Any]
+    evidence: npt.NDArray[np.float64]
+    rotations: npt.NDArray[np.float64]
+    pose_errors: npt.NDArray[np.float64] | float
```
tristanls: @ramyamounir, I think this could be fine without the `float`, but I want to verify with you.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
ramyamounir: One minor issue here is that we need to make sure `HypothesesUpdaterChannelTelemetry` is serializable before we add it to the dictionary for logging.

```python
        for graph_id, graph_telemetry in self.hypotheses_updater_telemetry.items():
            stats[graph_id] = {
                input_channel: self._channel_telemetry(
                    graph_id, input_channel, channel_telemetry
                ) # returns `HypothesesUpdaterChannelTelemetry`
                for input_channel, channel_telemetry in graph_telemetry.items()
            }
```

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
ramyamounir: Yes, I see the cast you added. Looks good!

ramyamounir:
Thanks for all these edits @tristanls, there is just one minor unresolved issue. I don't have any other concerns apart from the high-level bullet points in my original comment, but this is something we can discuss in one of the meetings.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
tristanls: @ramyamounir, I can see that being a concern. Is there specific code that fails as a result of this? Are we talking about a `BufferEncoder` test?

ramyamounir: Yes, sorry I should have led with that. I don't think the `BufferEncoder` knows how to handle it.

```bash
  File "/home/ramy/tbp/tbp.monty/benchmarks/run.py", line 44, in <module>
    main(all_configs=CONFIGS, experiments=cmd_args.experiments)
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/run.py", line 114, in main
    run(exp_config)
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/run.py", line 58, in run
    exp.evaluate()
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/experiments/monty_experiment.py", line 572, in evaluate
    self.run_epoch()
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/experiments/monty_experiment.py", line 526, in run_epoch
    self.run_episode()
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/experiments/object_recognition_experiments.py", line 45, in run_episode
    self.post_episode(last_step)
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/experiments/monty_experiment.py", line 497, in post_episode
    self.logger_handler.post_episode(self.logger_args)
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/loggers/exp_logger.py", line 52, in post_episode
    self.call_event("post_episode", logger_args)
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/loggers/exp_logger.py", line 77, in call_event
    getattr(logger, event)(
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/loggers/graph_matching_loggers.py", line 491, in post_episode
    self.log_episode(logger_args, output_dir, model)
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/loggers/graph_matching_loggers.py", line 485, in log_episode
    handler.report_episode(self.data, output_dir, episode, mode, **kwargs)
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/loggers/monty_handlers.py", line 87, in report_episode
    json.dump({total: output_data[total]}, f, cls=BufferEncoder)
  File "/home/ramy/.local/data/anaconda3/envs/tbp.monty/lib/python3.8/json/__init__.py", line 179, in dump
    for chunk in iterable:
  File "/home/ramy/.local/data/anaconda3/envs/tbp.monty/lib/python3.8/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/home/ramy/.local/data/anaconda3/envs/tbp.monty/lib/python3.8/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/home/ramy/.local/data/anaconda3/envs/tbp.monty/lib/python3.8/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/home/ramy/.local/data/anaconda3/envs/tbp.monty/lib/python3.8/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/home/ramy/.local/data/anaconda3/envs/tbp.monty/lib/python3.8/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/home/ramy/.local/data/anaconda3/envs/tbp.monty/lib/python3.8/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/home/ramy/.local/data/anaconda3/envs/tbp.monty/lib/python3.8/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/home/ramy/.local/data/anaconda3/envs/tbp.monty/lib/python3.8/json/encoder.py", line 438, in _iterencode
    o = _default(o)
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/models/buffer.py", line 680, in default
    return json.JSONEncoder.default(self, obj)
  File "/home/ramy/.local/data/anaconda3/envs/tbp.monty/lib/python3.8/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type HypothesesUpdaterChannelTelemetry is not JSON serializable
```

tristanls: @ramyamounir, I added code that registers a `HypothesesUpdaterChannelTelemetry` encoder with the `BufferEncoder` (thanks @scottcanoe for the easy mechanism). How do things look now?

```python
BufferEncoder.register(
    HypothesesUpdaterChannelTelemetry,
    lambda telemetry: asdict(telemetry),
)
```

ramyamounir: Looks and works great, thanks.

ramyamounir (approved):

tristanls:
I pressed the button too quickly 😕... I should have added a Co-Authored By trailer to indicate that you're the original author @ramyamounir. My apologies.

ramyamounir:
@tristanls That is totally fine, thanks for helping out with this telemetry issue 🙂

PR #402, opened 2025-07-28
ci: ignore uv test errors for now

jeremyshoemaker:


.github/workflows/monty.yml
```diff
@@ -339,6 +339,7 @@ jobs:
       - name: Run python tests
         if: ${{ needs.should_run_monty.outputs.should_run_monty == 'true' }}
         working-directory: tbp.monty
+        continue-on-error: true
```
tristanls: **issue**: I don't think this is the job you intended to modify.

jeremyshoemaker: Thanks. I wasn't expecting you to look at this yet since I'm on triage this week.

I'm also testing to see if this is needed on the next two steps (in the correct job).

tristanls: Oh sorry, it was not a draft so I assumed it was ready.

jeremyshoemaker: Old habits die hard.

tristanls (approved):
Thank you!

PR #403, opened 2025-07-28
ci: fix broken uv builds

jeremyshoemaker:
This involved a bit of work, but we need to specify the dependencies that Conda seems to include with PyTorch by default that the PyPI version doesn't.

To get it to work, we have to add an extra "find links" for the Pyg repository to pull in the prebuilt versions. Unfortunately, those prebuilt versions (at least for the versions we're pinning to) only exists for MacOS x86_64. Similar to the Conda environment, we need to force `uv` to use an x86_64 build of Python to get this to work. When creating the environment, we need to specify the exact version of Python we want.

```bash
uv venv --python cpython-3.9.22-macos-x86_64-none
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,9 @@ dependencies = [
     'sympy', # imported via conda (conda-forge::sympy)
     'torch==1.11.0', # imported via conda (pytorch::pytorch)
     'torchvision', # imported via conda (pytorch::torchvision)
-    'torch-geometric', # imported via conda (pyg::pyg)
+    'torch-geometric==2.1.0.post1', # imported via conda (pyg::pyg)
+    'torch-sparse==0.6.15', # imported via conda (pyg::pyg)
+    'torch-scatter==2.0.9', # imported via conda (pyg::pyg)
     'tqdm', # imported via conda (conda-forge::tqdm)
     'typing-extensions',
     'wandb', # imported via conda (conda-forge::wandb)
@@ -335,3 +337,7 @@ namespaces = true
 [tool.setuptools.package-data]
 # habitat-sim resources
 'tbp.monty.simulators.resources' = ['*.json', '*.yml', '*.txt']
+
+[tool.uv]
+# Extra index for PyTorch extensions
+find-links = ["https://data.pyg.org/whl/torch-1.11.0+cpu.html"]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -28,7 +28,9 @@ dependencies = [
     'sympy', # imported via conda (conda-forge::sympy)
     'torch==1.11.0', # imported via conda (pytorch::pytorch)
     'torchvision', # imported via conda (pytorch::torchvision)
-    'torch-geometric', # imported via conda (pyg::pyg)
+    'torch-geometric==2.1.0.post1', # imported via conda (pyg::pyg)
+    'torch-sparse==0.6.15', # imported via conda (pyg::pyg)
+    'torch-scatter==2.0.9', # imported via conda (pyg::pyg)
```
tristanls: **issue**: If these are pinned, then they also need to be pinned in the two environment YAML files: `environment.yml`, `environment_arm64.yml`. They'll also need to be named in the Conda way. For example:
```
# pyproject.toml
'torch==1.11.0`
```
```
# environment.yml
- pytorch::pytorch=1.11.0
```
Looks like they come from `pyg::`

tristanls:
Also, would you please add a `UV_PROTOTYPE.md` or similar that explains how to use `uv` with this work-in-progress setup?

jeremyshoemaker:
> Also, would you please add a `UV_PROTOTYPE.md` or similar that explains how to use `uv` with this work-in-progress setup?

That's a great idea.

pyproject.toml
jeremyshoemaker: When I do `conda list` I don't see `torch-geometric` or similar. The other two I can pin. I'm wondering if that one is being pulled in with something else.

jeremyshoemaker: Actually, now I'm wondering if I need to pin any of these, or if it will find the right ones because I changed the arch and added the find-link. I _did_ test, that find-link is needed.

jeremyshoemaker: Nope, the resolver fails if I unpin them.

jeremyshoemaker: Apparently `torch-geometric` is part of the Conda `pyg` package.

tristanls: The mapping is:
```
pyg::pyg -> torch-geometric
pyg::pytorch-scatter -> torch-scatter
pyg::pytorch-sparse -> torch-sparse
```
Yeah... `pyg` doesn't scream "torch geometric" at first glance 🙂.

Screenshot 2025-07-28 at 15 23 33

PR #404, opened 2025-07-28
fix: disable automatic changing of save_raw_obs with detailed loggers

ramyamounir:
This PR fixes one instance of configs coupling between detailed loggers and `save_raw_obs`. The current behavior for a Monty experiment is to automatically set `save_raw_obs=True` for all Sensor Modules if there exists a detailed logging handler. I've changed this to a warning.

**Original code**
```python
        if has_detailed_logger or self.show_sensor_output:
            # If we log detailed stats we want to save sm raw obs by default.
            for sm in self.model.sensor_modules:
                sm.save_raw_obs = True

```

**Proposed code**
```python
        if has_detailed_logger:
            for sm in self.model.sensor_modules:
                if hasattr(sm, "save_raw_obs") and not sm.save_raw_obs:
                    logger.warning(
                        "You are using a DETAILED logger with sensor module "
                        f"{sm.sensor_module_id} but 'save_raw_obs' is False. "
                        "Consider setting 'save_raw_obs' to True to log and visualize "
                        "the SM RGB raw values."
                    )
```


nielsleadholm (approved):
Looks good, thanks for updating this!

src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -407,10 +407,15 @@ def init_monty_data_loggers(self, logging_config: Dict[str, Any]) -> None:
         for lm in self.model.learning_modules:
             lm.has_detailed_logger = has_detailed_logger
 
-        if has_detailed_logger or self.show_sensor_output:
```
vkakerbeck, line 410: Did you check that the show_sensor_output check is not necessary? Does it still work if show_sensor_output is True and save_raw_obs is False?

vkakerbeck (commented):
Nice, thanks for making this change! Just a quick question on the removal of the show_sensor_output check

src/tbp/monty/frameworks/experiments/monty_experiment.py
ramyamounir: Yes, I checked. The option to `show_sensor_output` only depends on `observation` which already contains the RGB values from the sensor, it doesn't really need `save_raw_obs` to be set to True. This [here](https://github.com/thousandbrainsproject/tbp.monty/blob/c6199ca6d25094e039f7442bf841fad8a5135c47/src/tbp/monty/frameworks/experiments/object_recognition_experiments.py?plain=1#L133-L134) is the relevant code.

vkakerbeck (approved):

PR #405, opened 2025-07-28
rfc: RFC 14 Conducting Research While Building a Stable Platform

tristanls:
This RFC proposes a general strategy for how to reconcile our two goals of conducting research while building a stable platform.


tristanls:
@nielsleadholm, @vkakerbeck, I've discussed this idea with you, so I wanted to request that you take a first look and see if what I wrote down here reflects our conversations or whether I drifted into surprising territory.

rfcs/0000_conducting_research_while_building_a_stable_platform.md
```diff
@@ -0,0 +1,143 @@
...
+
+[The vision of the Thousand Brains Project is to develop a platform for building AI and robotics applications using the same principles as the human brain.](https://thousandbrainsproject.readme.io/docs/vision-of-the-thousand-brains-project) To achieve this vision, we pursue the platform work while we continue to conduct active research with [open questions still to be answered](https://thousandbrainsproject.readme.io/docs/open-questions) and already [planned future work](https://thousandbrainsproject.readme.io/docs/project-roadmap).
+
+We believe that the research and the platform are essential to our vision. We want to succeed at both.
+
+## Inherent challenge
+
+There lies an inherent challenge to achieving our vision. From the perspective of supply and demand competition in the marketplace, the activities of research and platform building differ in their characteristics. Some of what makes for good research, directly conflicts with what makes for a good platform, and vice versa.
+
+| Characteristic / Property | Research | Platform (today) | Platform (future) |
```
vkakerbeck: Is there some resource where these specific Characteristics and descriptions come from to read more about them? They sound quite specific and probably have a well defined meaning in that context but understanding all the classifications here without that context is a bit difficult. For example, what does it mean for "culture" to be a "decision driver"? Or what does the "writings about" characteristic refer to?

rfcs/0000_conducting_research_while_building_a_stable_platform.md
```diff
@@ -0,0 +1,143 @@
...
+
+For a platform, there is only so much change that platform users can absorb. While we intend for the platform to always stay up to date with the latest capabilities discovered through research, those capabilities need a level of stability before being incorporated into the platform. Even if we were to rapidly iterate on platform internals, any externally facing interface benefits from being stable. While significant improvements make changes easier to adopt, in general, the platform users are focused on doing their own research and/or solving their own business problems and want to minimize the cost of using our platform in toil, time, and treasure.
```
vkakerbeck: Just another case of unfamiliar vocabulary to me: What does cost in "toil, time, and treasure" refer to?

rfcs/0000_conducting_research_while_building_a_stable_platform.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Conflicting Goals
+
+So, our challenge is to navigate the conflicting goals of research and platform development. Research is the source of new capabilities. Platform is the stability on which solutions are built.
```
vkakerbeck:
```suggestion
So, our challenge is to navigate the conflicting goals of research and platform development. Research is the source of new capabilities. Platform is the ease-of-use and stability on which solutions are built.
```
Not sure how to best rephrase this, but just wanted to highlight that stability isn't the only purpose of the platform. It should also provide an easier to use and understand interface for a user that does not understand all of Monty's internal mechanics.

vkakerbeck (approved):
Nice writeup! I like the way you framed the tradeoff between platform and research goals and this looks like the best solution to it for now (although, like you mention, it comes with a few drawbacks as well).

rfcs/0000_conducting_research_while_building_a_stable_platform.md
tristanls: I'm sorry. This is just using [alliteration](https://en.wikipedia.org/wiki/Alliteration) to say "work, time, and money." I was trying to make it memorable 😬.

rfcs/0000_conducting_research_while_building_a_stable_platform.md
tristanls: 🤔, these are not as specific as they seem. They are intended as a general guide and there are many of them so that If one is difficult to understand or doesn't seem to make sense, then it can be skipped. There will be variance in interpretation of these.

To answer specifically, "culture" as a decision driver means that how we make decisions is driven by our organizational culture. We are building sensorimotor AI based on the neocortex, that will drive decisions about what we do, like research, read neuroscience papers, etc. This is in contrast to "previous experience", where we would make decisions based on what we did before that worked in that context. 

The "writings about" refers to the content people publish about the activity. So, what do people publish/write about TBP research, and about the platform? I renamed it from "publication types" trying to make it clearer.

I'll add a general disclaimer not to try to be super precise about all of these, but use them more as a guide to vector in on the general impression of these activities.

tristanls: update: https://github.com/thousandbrainsproject/tbp.monty/pull/405/commits/f5fd07e2b131aa574b4a9b7760cb01310fa948b2

rfcs/0000_conducting_research_while_building_a_stable_platform.md
tristanls: I tried `Platform, on which solutions are built, is usable and stable.`

rfcs/0000_conducting_research_while_building_a_stable_platform.md
vkakerbeck: okay, thanks!
I'm not sure if I would refer to the current decision driver as culture, I would think of it more as "evidence based", "neocortical organisation", or "TBT principals". But this is a minor thing not really relevant to the main topic of this RFC.

rfcs/0000_conducting_research_while_building_a_stable_platform.md
nielsleadholm: Haha ok yeah this confused me as well

rfcs/0000_conducting_research_while_building_a_stable_platform.md
```diff
@@ -0,0 +1,145 @@
...
+    IM -- Merge --> Monty(Monty main)
+```
+
+## Implementation Project
+
+The implementation project is a project because it may require major refactoring of the prototype or of Monty in order to integrate the new capability. It may be a single pull request, or it may be a long running effort as we reorganize all of Monty. The working prototype is helpful as it becomes a working specification of what needs to be implemented.
+
+This is the stage at which maintaining platform stability comes into consideration, not before.
+
+An additional benefit of an implementation project is that it functions as a knowledge exchange between researchers and engineers. Researchers catch up on the latest platform conventions, and engineers catch up on the latest capability and theory.
```
nielsleadholm: **praise:** I like that you highlight this point.

rfcs/0000_conducting_research_while_building_a_stable_platform.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Prototypes
+
+The main goal of a prototype is to provide feedback for the Thousand Brains Theory. By explicitly not merging prototypes, we shorten the feedback loop:
```
nielsleadholm:
```suggestion
The main goal of a prototype is to provide feedback for the Thousand Brains Theory. By explicitly not merging prototypes into the main `tbp.monty` repository, we shorten the feedback loop:
```

rfcs/0000_conducting_research_while_building_a_stable_platform.md
```diff
@@ -0,0 +1,145 @@
...
+
+The current code evolved through actual use solving real problems.
+
+We are a small team. Maintaining two major code bases will become challenging with our available personnel.
+
+It may be the case that a second system might succeed, but we defer that attempt to the community.
+
```
nielsleadholm: **suggestion:** It may be worth briefly mentioning the "third system", which is the status quo/what we have done until now. It doesn't need expounding on much, because it has the drawbacks you highlight earlier in the document, but maybe just say that if we find the new approach doesn't work, we technically have the option of reverting to the idea the research-code loop *does* factor in compatibility with the platform. 

That said I think you're right and the new approach will be better.

rfcs/0000_conducting_research_while_building_a_stable_platform.md
```diff
@@ -0,0 +1,145 @@
...
+### Second System
+
+An alternative approach would be to leave the Monty code (`tbp.monty`) as is and let it support primarily research needs. There would be less need for implementation projects, as the goal of the Monty code would be to support speed and agility. The platform would be created as a separate code base (`tbp.platform`), and it could pursue its goal of stability without interfering with research, absorbing any new `tbp.monty` capabilities as needed.
+
+The primary reason we should not write a new `tbp.platform` is because there are numerous examples of that approach failing. Rewriting code from scratch often seems like a good idea, but it comes with challenges that end in a long-term wasted effort on a code base that is never adopted. Joel Spolsky has a helpful blog post on the topic: [Things You Should Never Do, Part I](https://www.joelonsoftware.com/2000/04/06/things-you-should-never-do-part-i/), quoted liberally below:
```
nielsleadholm: **note:** I agree with this. I am also concerned that there would be *two* major headaches introduced by having two code-bases. One is porting over new features from `tbp.monty` to `tbp.platform`. The other is porting the stable, cleaned up version *back* to `tbp.monty` from `tbp.platform`. Depending on the cadence of changes to the research code-base, this might introduce entirely new, complex merge conflicts. I think by having a single main upstream that we all regularly sync from, we can minimize these issues.

nielsleadholm (approved):
**praise:** This is a great overview. I still like the ideas you have proposed and think it will be a good approach.

**suggestion:** I have started writing more specific guidance for researchers that makes these ideas even more concrete for that setting. I would ideally like for us to have documentation in a central, easy-to-access place that provides guidance on best code-practice for researchers, community members, and (assuming you agree) engineers.

As such, I think it would be helpful if the output of this RFC is to create an artefact/page in Readme.com that basically repeats the information you have written. That way I can open a future PR that edits that document, as well as links it to new ones. 

For context, below are some examples of the kinds of additions I'm thinking:
- Detailed guidance on how to manage Prototypes and Implementation Projects with Github.
- As the complexity of contributions from the community grows, we may want them to follow a similar approach to researchers (i.e. merge PRs into a branch on their own fork), before opening a PR that is the Implementation Project.
- At the same time, both researchers and community members MAY occasionally open PRs that go directly into tbp.monty, e.g. a useful refactor that is relatively independent of a new research idea.

rfcs/0000_conducting_research_while_building_a_stable_platform.md
tristanls: 2:1, I'll update to "work, time, and money".

rfcs/0000_conducting_research_while_building_a_stable_platform.md
tristanls: Oh, good point. I'll add it as an option.

tristanls: I added the status quo section. The worth highlighting bit is that it isn't the worst feedback loop example but a sort of combination of the two feedback loops described previously. More in the new section: https://github.com/thousandbrainsproject/tbp.monty/pull/405/commits/493dd5acdb6aab1f1374ada57e0abd717ae1f0c2

rfcs/0000_conducting_research_while_building_a_stable_platform.md
tristanls: Good point, I incorporated your note into the RFC: https://github.com/thousandbrainsproject/tbp.monty/pull/405/commits/8dca2e86598179db8f89e162c6d3aef46f2f037b

tristanls:
> **suggestion**: I have started writing more specific guidance for researchers that makes these ideas even more concrete for that setting. I would ideally like for us to have documentation in a central, easy-to-access place that provides guidance on best code-practice for researchers, community members, and (assuming you agree) engineers.
>
> As such, I think it would be helpful if the output of this RFC is to create an artefact/page in Readme.com that basically repeats the information you have written. That way I can open a future PR that edits that document, as well as links it to new ones.

@nielsleadholm, I think your suggestion is reasonable, and having a guide on code practices is helpful. We already have a section in the documentation for this information, the [Style Guide](https://thousandbrainsproject.readme.io/docs/style-guide). It needs to be expanded to include the distinctions you have in mind, but to me, that's the editable-via-PR document you bring up.

To me, this RFC is for a single decision: Should we adopt the strategy depicted in the illustration below?
Screenshot 2025-07-31 at 10 14 06

Much like I leave specifying the platform to some other document, I think detailed guidance on writing the code if we adopt this RFC's proposed strategy should also be elaborated elsewhere.

rfcs/0000_conducting_research_while_building_a_stable_platform.md
tristanls: I'm happy to change it. Would summarizing it as "research" work? As in `research / analysis & synthesis`. I updated it to that for now: https://github.com/thousandbrainsproject/tbp.monty/pull/405/commits/734d29de5889d5b8aa6cb55e5ff2404263482f4c

vkakerbeck: That works, thanks!

vkakerbeck:
Nice updates! Re. where to add this info to the docs I would tend towards adding it to https://thousandbrainsproject.readme.io/docs/pull-requests instead of the style guide as I wouldn't intuitively consult the style guide to find out about the process for contributing new features to Monty. But as you said, this RFC is just about the decision whether we adopt this process so we can defer this discussion to later (potentially restructuring the docs on this in general and incorporating more guidance on when and how to run benchmarks etc. as well).

nielsleadholm (approved):
Looks great, thanks for those changes! 

Agree about making a more general re-work of our documentation later, just that, as said, I like how you've worded a lot of this, so I think it would be a nice starting point for some of the guidance. Happy to help with actually porting it into the documentation.

tristanls:
Hi @codeallthethingz, @hlee9212, @scottcanoe, @ramyamounir, @jeremyshoemaker. Please look at this proposal to see what our general strategy should be going forward.

rfcs/0000_conducting_research_while_building_a_stable_platform.md
```diff
@@ -0,0 +1,176 @@
...
+
+# Conducting Research While Building a Stable Platform
+
+[The vision of the Thousand Brains Project is to develop a platform for building AI and robotics applications using the same principles as the human brain.](https://thousandbrainsproject.readme.io/docs/vision-of-the-thousand-brains-project) To achieve this vision, we pursue the platform work while we continue to conduct active research with [open questions still to be answered](https://thousandbrainsproject.readme.io/docs/open-questions) and already [planned future work](https://thousandbrainsproject.readme.io/docs/project-roadmap).
+
+We believe that the research and the platform are essential to our vision. We want to succeed at both.
+
+## Inherent challenge
+
+There lies an inherent challenge to achieving our vision. From the perspective of supply and demand competition in the marketplace, the activities of research and platform building differ in their characteristics. Some of what makes for good research, directly conflicts with what makes for a good platform, and vice versa.
```
jeremyshoemaker: The "supply and demand" supporting clause seems irrelevant to me, but maybe it's just me 😄. I think the activities of research and platform building simply differ, there's not any particular perspective that makes that so.

rfcs/0000_conducting_research_while_building_a_stable_platform.md
```diff
@@ -0,0 +1,176 @@
...
+    Monty -- Feedback --> T
+```
+
+With the status quo, the fast feedback loop already exists. Partial functionality is merged when it is deemed ready.
+
+This approach avoids the whole concept of an implementation project, and focuses on integrating the code one pull request at a time. Merging to Monty usually calls for additional design and refactoring. If the merged partial functionality will not change, then merging partial functionality one pull request at a time is the better approach. The challenge comes if the partial functionality changes too often, before the new capability is settled. This leads to multiple rounds of rework.
+
+This RFC proposes to avoid multiple rounds of rework so that researchers can spend what we assume would be most of their time in the prototype-theory feedback loop, followed by batching all the design and refactoring merge work into a single integration project.
+
+### Second System
```
jeremyshoemaker: I'm assuming this heading is a reference to [second-system effect or second-system syndrome](https://en.wikipedia.org/wiki/Second-system_effect). Would it make sense to change it to one of these to make the connection more explicit (for those familiar with it)?

jeremyshoemaker: Never mind, I think what you mean here is something like "Build a Second System". You do bring up the second-system effect, but I misread this for some reason.

jeremyshoemaker (approved):
Looks reasonable to me.

rfcs/0000_conducting_research_while_building_a_stable_platform.md
tristanls: I did name this section so that it "rhymes" with the effect and the syndrome for people familiar with it, but I did not want to argue over which one it is and the definition, so I left it loose like this.

jeremyshoemaker: For what it's worth, it looks like Fred Brooks coined it, and he used "effect". I tend to agree with that phrasing, though I hear the other one more often. To me a "syndrome" is usually a condition with lacks a known cause, though that might be an incorrect understanding of the term.

ramyamounir:
Thanks @tristanls, this is very well written. I don't have major objections to trying out this new approach, but I just wanted to complement your drawbacks section with some points that worry me as a researcher. I have already discussed many of them with @nielsleadholm.

- It's a lot easier to incorporate engineering reviews into the feature implementation when changes are introduced gradually during feature development rather than waiting for them to become a whole new project. This new integration project is not just expensive, it can also be demotivating for researchers to spend a few isolated weeks away from research. Once I finish implementing a feature, I am most excited about starting the next research idea. At that point, working on integrating an older research feature can easily become a burden and detract from research momentum.
- While this makes parallel, and rapid, prototyping for researchers easier, it can become difficult to integrate multiple projects from different researchers at the same time, especially if the prototypes are based on different Monty versions and target the same code.
- This assumes that research features are isolated in nature, but that's rarely the case. If researchers end up waiting until after the implementation project of other researchers to pull in each other’s code, we risk ideas and code fragmentation and degraded collaboration between researchers. For example, if Hojae didn't see the latest refactoring of the Hypotheses updaters and displacers, she would be building off of older code, which adds unnecessary work later during integration. Similarly, I want to regularly pull in Hojae's updates on removing hypotheses while working on my dynamic resizing feature.
- This unintentionally widens the gap between research and engineering. In my opinion, it is quite expected to have some friction between the rapid prototyping goals of researchers and the code stability goals of engineers. This friction is a learning signal for both sides; it is somewhat temporary and should significantly decrease with continuous interaction after a few projects with PR reviews. We should not avoid or postpone this learning opportunity.
- Monty is already designed as a modular architecture. Adding new, possibly incomplete, features should not introduce major stability issues. For example, a new `LearningModule` can be introduced without breaking any of the existing code. Some changes that introduce an entirely new component can be breaking (e.g., `GoalStateSelector`), but this may not happen very often. We could consider a special procedure for these cases, but that's a different issue.

vkakerbeck:
Hey @ramyamounir 

thanks for raising those concerns. Those are some thoughtful points and I won't argue against most of them. I think we will need to find a balance between our current approach and the issues that come with the proposed new approach and it might take us some experimentation over the next months to find the best workflow. Here is why I think it is worth trying this new approach:

- **We don't want to spend a lot of effort integrating incremental PRs on a research feature before this feature has proven itself**. The current state is that we make incremental PRs (for ease of reviewing) but then each incremental PR goes through a thorough review process. You mention "At that point, working on integrating an older research feature can easily become a burden and detract from research momentum." I worry that the current process does this, as it breaks the research momentum with each PR. Integration into Monty is an unavoidable cost that we will have at some point but I think it makes sense to delay this until we know that a research idea actually works and that we will want to integrate it into Monty. Otherwise, we would have wasted time from both research and engineering to get the research code up to Monty standards, and it would have been abandoned after all because the idea didn't turn out to work.
- **Maybe calling it an "implementation project" makes it sound larger than it should be**. We don't always have to wait until a giant milestone is completed before adding a feature into Monty. The main thing is that the feature should be experimentally proven before we dedicate the engineering effort to integrate it. However, this could be broken up into smaller incremental changes as well, depending on the type of project. I think this is where we need to figure out a good balance over time. 
- My current tendency would be to integrate new features as early as possible (and try to experimentally validate them as early as possible) so that, as you say, other researchers' code can use the new feature and the forks don't get too much out of sync. I think this is an important point you raise and we need to make sure to have good communication about the different work on our forks and how it may depend on each other.

ramyamounir:
@vkakerbeck Yes, seems like it will depend more on the size of the feature that gets added to Monty and finding this right balance, as you mentioned. I think some of this will be addressed more concretely in Niels' PR. 

The main point I was trying to make is that even when a research feature doesn’t make it into `tbp.monty`, the time spent getting engineering feedback isn’t completely wasted as long as researchers can generalize that knowledge and apply it to future features. Sure, it may slow down research at first, but it will save much time later on for both researchers and engineers. Personally, I’d like to eventually be writing PRs that pass engineering review with little to no friction, and I feel that the thorough and continuous feedback from Tristan and Jeremy is helping me get there a lot faster.

That said, I’m fully on board with trying out this new approach. I just wanted to voice some additional concerns early on, and maybe suggest we treat this as an experimental phase and iterate as we learn what works.

ramyamounir (approved):

rfcs/0000_conducting_research_while_building_a_stable_platform.md
scottcanoe:
```suggestion
There lies an inherent challenge to achieving our vision. From the perspective of supply and demand competition in the marketplace, the activities of research and platform building differ in their characteristics. Some of what makes for good research directly conflicts with what makes for a good platform, and vice versa.
```

scottcanoe (approved):
Nice work, I have no problem with any of this. As I'm sure we're all aware, the devil is going to be in the details, and the development and integration patterns may (should?) vary a lot depending on the project. For example, my current project right now is almost totally about extensions to Monty's architecture, and this is way different from, say, modifying how hypotheses might be resampled. So long as we're not too strict about how go about this (and this document isn't strict), I'm sure we can all figure out best practices moving forward in a context-dependent manner while keeping these goals and principles in mind.

tristanls:
@ramyamounir @vkakerbeck @scottcanoe, I wonder if it would be helpful to highlight that deciding when the prototype is stable and when to initiate integration (via PR or a larger effort) is mostly at the discretion of the researcher.

The strategy proposed here is mainly from the perspective of having a space for researchers to explicitly opt out of platform requirements. This is partly because I can easily imagine platform requirements ramping up (e.g., increased scrutiny on unit testing, schema requirements on telemetry data, stricter logging requirements).

tristanls:
> I wonder if it would be helpful to highlight that deciding when the prototype is stable and when to initiate integration (via PR or a larger effort) is mostly at the discretion of the researcher.

Added in https://github.com/thousandbrainsproject/tbp.monty/pull/405/commits/96121013feffed980843c15d0977e30ca73906d2

tristanls:

@vkakerbeck, @nielsleadholm, @codeallthethingz, @hlee9212, @scottcanoe, @ramyamounir, @jeremyshoemaker, I interpret your pull request approvals as confirmation to **merge**, but I will wait some time in case you have any final comments.

tristanls:

PR #407, opened 2025-07-30
feat: type hints and pathlib in setup_env.py

hlee9212:
Getting around to updating this file, which has previously caused a minor error when using `str` to represent path in the following:
```python
monty_models_dir = f"{monty_logs_dir}pretrained_models/"
```
(I think I had to add or get rid of a "/" in my `.zshrc` file, something like that).

Basically wanted to use `pathlib` and add some type hints for a while, then sort of became a slightly larger style change (but no change in functionality) while working on visualization PR and had a chance to revisit this file again. 

Some other changes:
1. Using logger instead of print. 
2. Returning a dictionary instead of None - I don't have much justification for this except if we were to add a unit test this would make it easier. Sometimes I see setters do this, but no strong opinion. 

Would be great if one of the engineering team members could go over this to make sure it adheres to the code style guide.

### DIFF
--- a/src/tbp/monty/frameworks/run_env.py
+++ b/src/tbp/monty/frameworks/run_env.py
@@ -8,39 +8,129 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
+"""Environment setup utilities for Monty framework."""
+
+import logging
 import os
+from pathlib import Path
+from typing import Dict, Union
+
+# Environment variable names
+MONTY_LOGS_VAR = "MONTY_LOGS"
+MONTY_MODELS_VAR = "MONTY_MODELS"
+MONTY_DATA_VAR = "MONTY_DATA"
+WANDB_DIR_VAR = "WANDB_DIR"
+
+# Type alias for path-like objects
+PathLike = Union[str, Path]
+
+# Configure logger
+logger = logging.getLogger(__name__)
 
 
-def setup_env(monty_logs_dir_default: str = "~/tbp/results/monty/"):
-    """Setup environment variables for Monty.
+def setup_env(
+    monty_logs_dir_default: PathLike = "~/tbp/results/monty/",
+) -> Dict[str, str]:
+    """Set up environment variables for Monty framework.
+
+    This function initializes required environment variables if they are not
+    already set, using sensible defaults for the Monty framework.
 
     Args:
-        monty_logs_dir_default: Default directory for Monty logs.
+        monty_logs_dir_default: Default directory path for Monty logs.
+            Can be a string or Path object. Defaults to "~/tbp/results/monty/".
+
+    Returns:
+        Dict[str, str]: A dictionary containing all configured environment
+            variables and their paths.
+
+    Side Effects:
+        Sets the following environment variables if not already present:
+        - MONTY_LOGS: Directory for Monty logs
+        - MONTY_MODELS: Directory for pretrained models (under MONTY_LOGS)
+        - MONTY_DATA: Directory for data
+        - WANDB_DIR: Directory for Weights & Biases logs (defaults to MONTY_LOGS)
+
+    Raises:
+        ValueError: If the provided path is invalid or cannot be processed.
+
+    Example:
+        >>> env_paths = setup_env()
+        >>> print(env_paths['MONTY_LOGS'])
+        /home/user/tbp/results/monty/
+
+        >>> env_paths = setup_env(Path.home() / "custom" / "logs")
+        >>> print(env_paths['MONTY_MODELS'])
+        /home/user/custom/logs/pretrained_models
     """
-    monty_logs_dir = os.getenv("MONTY_LOGS")
+    # Validate input
+    try:
+        monty_logs_dir_default = str(monty_logs_dir_default)
+    except Exception as e:
+        raise ValueError(f"Invalid path provided: {monty_logs_dir_default}") from e
 
-    if monty_logs_dir is None:
-        monty_logs_dir = monty_logs_dir_default
-        os.environ["MONTY_LOGS"] = monty_logs_dir
-        print(f"MONTY_LOGS not set. Using default directory: {monty_logs_dir}")
+    # Initialize environment paths dictionary
+    env_paths = {}
 
-    monty_models_dir = os.getenv("MONTY_MODELS")
+    # Set up MONTY_LOGS
+    monty_logs_dir = _get_or_set_env_var(
+        MONTY_LOGS_VAR, default_value=monty_logs_dir_default
+    )
+    env_paths[MONTY_LOGS_VAR] = monty_logs_dir
 
-    if monty_models_dir is None:
-        monty_models_dir = f"{monty_logs_dir}pretrained_models/"
-        os.environ["MONTY_MODELS"] = monty_models_dir
-        print(f"MONTY_MODELS not set. Using default directory: {monty_models_dir}")
+    # Set up MONTY_MODELS (depends on MONTY_LOGS)
+    monty_models_default = str(Path(monty_logs_dir) / "pretrained_models")
+    monty_models_dir = _get_or_set_env_var(
+        MONTY_MODELS_VAR, default_value=monty_models_default
+    )
+    env_paths[MONTY_MODELS_VAR] = monty_models_dir
 
-    monty_data_dir = os.getenv("MONTY_DATA")
+    # Set up MONTY_DATA
+    monty_data_dir = _get_or_set_env_var(MONTY_DATA_VAR, default_value="~/tbp/data/")
+    env_paths[MONTY_DATA_VAR] = monty_data_dir
 
-    if monty_data_dir is None:
-        monty_data_dir = os.path.expanduser("~/tbp/data/")
-        os.environ["MONTY_DATA"] = monty_data_dir
-        print(f"MONTY_DATA not set. Using default directory: {monty_data_dir}")
+    # Set up WANDB_DIR (uses MONTY_LOGS as default)
+    wandb_dir = _get_or_set_env_var(WANDB_DIR_VAR, default_value=monty_logs_dir)
+    env_paths[WANDB_DIR_VAR] = wandb_dir
 
-    wandb_dir = os.getenv("WANDB_DIR")
+    return env_paths
 
-    if wandb_dir is None:
-        wandb_dir = monty_logs_dir
-        os.environ["WANDB_DIR"] = wandb_dir
-        print(f"WANDB_DIR not set. Using default directory: {wandb_dir}")
+
+def _get_or_set_env_var(var_name: str, default_value: Union[str, Path]) -> str:
+    """Get environment variable or set it to default value.
+
+    This function always expands user paths (~) and resolves paths
+    to absolute paths for consistency.
+
+    Args:
+        var_name: Name of the environment variable.
+        default_value: Default value to use if variable is not set.
+            Can be a string or Path object.
+
+    Returns:
+        The value of the environment variable (either existing or newly set),
+        with user paths expanded.
+
+    Raises:
+        ValueError: If path processing fails.
+    """
+    value = os.getenv(var_name)
+
+    if value is None:
+        # Convert to Path for processing
+        try:
+            path = Path(str(default_value))
+            # Always expand user and resolve to absolute path
+            expanded_path = path.expanduser()
+            value = str(expanded_path)
+        except Exception as e:
+            raise ValueError(
+                f"Failed to process path for {var_name}: {default_value}"
+            ) from e
+
+        os.environ[var_name] = value
+        logger.info(f"{var_name} not set. Using default directory: {value}")
+    else:
+        logger.debug(f"{var_name} already set to: {value}")
+
+    return value
--- a/src/tbp/monty/frameworks/run_env.py
+++ b/src/tbp/monty/frameworks/run_env.py
@@ -16,10 +16,10 @@ from pathlib import Path
 from typing import Dict, Union
 
 # Environment variable names
-MONTY_LOGS_VAR = "MONTY_LOGS"
-MONTY_MODELS_VAR = "MONTY_MODELS"
-MONTY_DATA_VAR = "MONTY_DATA"
-WANDB_DIR_VAR = "WANDB_DIR"
+MONTY_LOGS = "MONTY_LOGS"
+MONTY_MODELS = "MONTY_MODELS"
+MONTY_DATA = "MONTY_DATA"
+WANDB_DIR = "WANDB_DIR"
 
 # Type alias for path-like objects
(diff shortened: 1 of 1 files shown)


jeremyshoemaker (approved):
Overall, this change looks fine to me.

That said, I know that the previous code already did this, but I just want to point out that **setting** environment variables from Python can cause problems. In particular, the docs have this to say about setting to `os.environ[...]`:

> On some platforms, including FreeBSD and macOS, setting environ may cause memory leaks. 
> ~ https://docs.python.org/3/library/os.html#os.environ

I generally try to avoid it if I can. There are a few places where it _might_ make sense, e.g. starting a subprocess with updated environment variables, but even in those cases, there are usually better alternatives.

The way frameworks like Django handle this is they have a centralize place where all the environment and settings are loaded and configured, and then the rest of the application just accesses that instead of the environment variables eliminating the need to set them. There's a Python package I've used in the past called `simple_settings` that does something similar that we might want to eventually consider, but that's probably part of a larger rethink of how we do configurations since there's probably some overlap there.

WANDB_DIR is the odd one out because it's used by a library we don't directly control. In that case, I'd probably look at having the code that looks at settings check for any default values or required values it might have and fail with an error if it isn't set correctly.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+# Environment variable names
+MONTY_LOGS_VAR = "MONTY_LOGS"
+MONTY_MODELS_VAR = "MONTY_MODELS"
+MONTY_DATA_VAR = "MONTY_DATA"
+WANDB_DIR_VAR = "WANDB_DIR"
+
+# Type alias for path-like objects
+PathLike = Union[str, Path]
+
+# Configure logger
```
tristanls, line 27: **issue**: Extraneous comment, please remove.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
 # https://opensource.org/licenses/MIT.
 
+"""Environment setup utilities for Monty framework."""
+
+import logging
 import os
+from pathlib import Path
+from typing import Dict, Union
+
+# Environment variable names
```
tristanls, line 18: **issue**: Extraneous comment, please remove.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+MONTY_LOGS_VAR = "MONTY_LOGS"
```
tristanls: **issue**: Please don't include `_VAR`, short for "variable," in a variable name.

jeremyshoemaker (changes requested):
@tristanls pointed out some issues that I missed.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+from pathlib import Path
+from typing import Dict, Union
+
+# Environment variable names
+MONTY_LOGS = "MONTY_LOGS"
+MONTY_MODELS = "MONTY_MODELS"
+MONTY_DATA = "MONTY_DATA"
+WANDB_DIR = "WANDB_DIR"
+
+# Type alias for path-like objects
```
tristanls, line 24: **issue**: Extraneous comment, please remove.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+    env_paths[MONTY_DATA] = monty_data_dir
+
+    # Set up WANDB_DIR (uses MONTY_LOGS as default)
+    wandb_dir = _get_or_set_env_var(WANDB_DIR, default_value=monty_logs_dir)
+    env_paths[WANDB_DIR] = wandb_dir
+
+    return env_paths
+
+
+def _get_or_set_env_var(var_name: str, default_value: Union[str, Path]) -> str:
```
tristanls, line 99: **issue**: `default_value` should be `PathLike`.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+    This function initializes required environment variables if they are not
+    already set, using sensible defaults for the Monty framework.
 
     Args:
-        monty_logs_dir_default: Default directory for Monty logs.
+        monty_logs_dir_default: Default directory path for Monty logs.
+            Can be a string or Path object. Defaults to "~/tbp/results/monty/".
+
+    Returns:
+        Dict[str, str]: A dictionary containing all configured environment
```
tristanls, line 44: **issue**: With the latest changes in #352, we no longer include the type in the Returns docstring.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+        ValueError: If the provided path is invalid or cannot be processed.
+
+    Example:
+        >>> env_paths = setup_env()
+        >>> print(env_paths['MONTY_LOGS'])
+        /home/user/tbp/results/monty/
+
+        >>> env_paths = setup_env(Path.home() / "custom" / "logs")
+        >>> print(env_paths['MONTY_MODELS'])
+        /home/user/custom/logs/pretrained_models
```
tristanls, line 64: **issue**: If I follow this example, the code will appear to not work. `setup_env()` only works on first run, after which it becomes a no-op. This is because it will set all environment variables on the first run, so when it runs again, it finds the variables set and will take no action.

**suggestion**: Update the example so it works if I paste it into an interactive shell; or a test would be a better example; or omit the example.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
     """
-    monty_logs_dir = os.getenv("MONTY_LOGS")
+    # Validate input
```
tristanls, line 66: **issue**: Extraneous comment, please remove.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+    try:
+        monty_logs_dir_default = str(monty_logs_dir_default)
+    except Exception as e:
+        raise ValueError(f"Invalid path provided: {monty_logs_dir_default}") from e
```
tristanls, line 70: **issue**: The type of `monty_logs_dir_default` is `PathLike`. I don't think there's a case where `str(str)` or `str(Path)` can raise an exception.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
 
-    if monty_logs_dir is None:
-        monty_logs_dir = monty_logs_dir_default
-        os.environ["MONTY_LOGS"] = monty_logs_dir
-        print(f"MONTY_LOGS not set. Using default directory: {monty_logs_dir}")
+    # Initialize environment paths dictionary
```
tristanls, line 72: **issue**: Extraneous comment, please remove.

src/tbp/monty/frameworks/run_env.py
tristanls: **nitpick**: I think the code is more readable if `_get_or_set_env_var` is replaced by `env_var`. For example:

```python
monty_logs_dir = env_var(MONTY_LOGS, default_value=monty_logs_dir_default)
```

Also, any reason to keep it "private" with `_` prefix?

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+    env_paths = {}
 
-    monty_models_dir = os.getenv("MONTY_MODELS")
+    # Set up MONTY_LOGS
```
tristanls, line 75: **issue**: Extraneous comment, please remove.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+    monty_logs_dir = _get_or_set_env_var(
+        MONTY_LOGS, default_value=monty_logs_dir_default
+    )
+    env_paths[MONTY_LOGS] = monty_logs_dir
 
-    if monty_models_dir is None:
-        monty_models_dir = f"{monty_logs_dir}pretrained_models/"
-        os.environ["MONTY_MODELS"] = monty_models_dir
-        print(f"MONTY_MODELS not set. Using default directory: {monty_models_dir}")
+    # Set up MONTY_MODELS (depends on MONTY_LOGS)
```
tristanls, line 81: **issue**: Extraneous comment, please remove.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+    monty_models_default = str(Path(monty_logs_dir) / "pretrained_models")
+    monty_models_dir = _get_or_set_env_var(
+        MONTY_MODELS, default_value=monty_models_default
+    )
+    env_paths[MONTY_MODELS] = monty_models_dir
 
-    monty_data_dir = os.getenv("MONTY_DATA")
+    # Set up MONTY_DATA
```
tristanls, line 88: **issue**: Extraneous comment, please remove.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+    monty_data_dir = _get_or_set_env_var(MONTY_DATA, default_value="~/tbp/data/")
+    env_paths[MONTY_DATA] = monty_data_dir
+
+    # Set up WANDB_DIR (uses MONTY_LOGS as default)
```
tristanls, line 92: **issue**: Extraneous comment, please remove.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+        # Convert to Path for processing
+        try:
+            path = Path(str(default_value))
+            # Always expand user and resolve to absolute path
+            expanded_path = path.expanduser()
+            value = str(expanded_path)
+        except Exception as e:
+            raise ValueError(
+                f"Failed to process path for {var_name}: {default_value}"
+            ) from e
```
tristanls, line 129: **issue (non-blocking)**: `default_value` should be a `PathLike`, so the only thing that could fail here is `path.expanduser()`. The only thing that raises is `RuntimeError` if the home directory cannot be resolved. I don't think it is worth wrapping with all this code.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+        ValueError: If path processing fails.
+    """
+    value = os.getenv(var_name)
 
-    if monty_data_dir is None:
-        monty_data_dir = os.path.expanduser("~/tbp/data/")
-        os.environ["MONTY_DATA"] = monty_data_dir
-        print(f"MONTY_DATA not set. Using default directory: {monty_data_dir}")
+    if value is None:
+        # Convert to Path for processing
```
tristanls, line 120: **issue**: Extraneous comment, please remove.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+        try:
+            path = Path(str(default_value))
+            # Always expand user and resolve to absolute path
```
tristanls, line 123: **issue**: Extraneous comment, please remove.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+            expanded_path = path.expanduser()
+            value = str(expanded_path)
+        except Exception as e:
+            raise ValueError(
+                f"Failed to process path for {var_name}: {default_value}"
+            ) from e
 
-    wandb_dir = os.getenv("WANDB_DIR")
+        os.environ[var_name] = value
+        logger.info(f"{var_name} not set. Using default directory: {value}")
```
tristanls, line 132: **issue**: I think the default logging level is `WARN`, so this doesn't show up like the previous `print` would.

**suggestion**: Change to `logger.warning`.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
+    value = os.getenv(var_name)
 
-    if monty_data_dir is None:
-        monty_data_dir = os.path.expanduser("~/tbp/data/")
-        os.environ["MONTY_DATA"] = monty_data_dir
-        print(f"MONTY_DATA not set. Using default directory: {monty_data_dir}")
+    if value is None:
+        # Convert to Path for processing
+        try:
+            path = Path(str(default_value))
```
tristanls, line 122: **issue**: `str` conversion is extraneous. `Path(Path) -> Path`.

```
>>> from pathlib import Path
>>> p = "~/path"
>>> pp = Path(p)
>>> ppp = Path(pp)
>>> p
'~/path'
>>> pp
PosixPath('~/path')
>>> ppp
PosixPath('~/path')
```

tristanls (commented):
Thank you for working on improving this code.

**note**: There are quite a few comments that I find extraneous. In the cases I highlighted, the code speaks for itself. Sometimes, the comment is longer than the code:
```python
# Initialize environment paths dictionary
env_paths = {}
```

**suggestion (non-blocking)**: Since @jeremyshoemaker mentioned that we should not set environment variables, maybe add a `# TODO` to make a note of it?

src/tbp/monty/frameworks/run_env.py
```diff
@@ -8,39 +8,129 @@
...
-    monty_data_dir = os.getenv("MONTY_DATA")
+    # Set up MONTY_DATA
+    monty_data_dir = _get_or_set_env_var(MONTY_DATA, default_value="~/tbp/data/")
+    env_paths[MONTY_DATA] = monty_data_dir
+
+    # Set up WANDB_DIR (uses MONTY_LOGS as default)
+    wandb_dir = _get_or_set_env_var(WANDB_DIR, default_value=monty_logs_dir)
+    env_paths[WANDB_DIR] = wandb_dir
+
+    return env_paths
```
tristanls, line 96: **issue**: While I understand that this might be used in a hypothetical scenario. It is not used either in test or by code using `setup_env`. That would make it dead code and should be removed, i.e. that's a reason not to assemble and return the dictionary from this function.

src/tbp/monty/frameworks/run_env.py
hlee9212: Removed example.

jeremyshoemaker: You've already removed it, so it doesn't matter, but we could also have the test suite run these as tests, which is what they're originally for.

jeremyshoemaker:
I think this PR is superseded by the work that `@carver` (trying to avoid notifying him unnecessarily) has been doing enabling Ruff checks for Path related stuff. Can we close it?

hlee9212:
Hi @jeremyshoemaker - yes, let's close this. I've been meaning to do it for a while (kept forgetting).

PR #408, opened 2025-07-31
fix: pin more package versions in conda

### LINKED ISSUE #401: NotImplementedError: Axes3D currently only supports the aspect argument 'auto'. You passed in 'equal'
### Describe the bug

Ok, so I've been getting more familiar with the code, and running different stuff. For this issue, I've run the plot code from the examples here: [Logging and Analysis
](https://thousandbrainsproject.readme.io/docs/logging-and-analysis)

This is the code I ran:
```
import os
from tbp.monty.frameworks.utils.logging_utils import load_stats
from tbp.monty.frameworks.utils.logging_utils import print_overall_stats
import matplotlib.pyplot as plt
from tbp.monty.frameworks.utils.plot_utils import plot_graph

pretrain_path = os.path.expanduser("~/tbp/results/monty/pretrained_models/")
pretrained_dict = pretrain_path + "pretrained_ycb_v10/surf_agent_1lm_10distinctobj/pretrained/"

log_path = os.path.expanduser("~/tbp/results/monty/projects/evidence_eval_runs/logs/")
exp_name = "randrot_10distinctobj_surf_agent"
exp_path = log_path + exp_name

train_stats, eval_stats, detailed_stats, lm_models = load_stats(
    exp_path,
    load_train=False, # doesn't load train csv
    load_eval=True, # loads eval_stats.csv
    load_detailed=False, # doesn't load .json
    load_models=True, # loads .pt models
    pretrained_dict=pretrained_dict,
)

plot_graph(lm_models['pretrained'][0]['mug']['patch'], rotation=120)
plt.show()
```

Running this generated this error:
```
...loading and checking eval statistics...
...loading LM models...
  File "benchmarks/see_statistics2.py", line 27, in <module>
    plot_graph(lm_models['pretrained'][0]['mug']['patch'], rotation=120)
  File "/mnt/c/dev/TBP-Monty/tbp.monty/src/tbp/monty/frameworks/utils/plot_utils.py", line 103, in plot_graph
    ax.set_aspect("equal")
  File "/home/melkor/anaconda3/envs/tbp.monty/lib/python3.8/site-packages/mpl_toolkits/mplot3d/axes3d.py", line 323, in set_aspect
    raise NotImplementedError(
NotImplementedError: Axes3D currently only supports the aspect argument 'auto'. You passed in 'equal'.
```
Image

So the error came from `plot_utils.py` on line 103: `ax.set_aspect("equal")`

At first, I thought my `matplotlib` package was not good, so I reinstalled it with: 
`conda install --force-reinstall conda-forge::matplotlib`

Then I checked my version:
```
>>> import matplotlib
>>> print(matplotlib.__version__)
3.3.2
```

I've checked the documentation for my version
[Matplotlib 3.3.2 Version](https://matplotlib.org/3.3.2/api/_as_gen/matplotlib.axes.Axes.set_aspect.html?highlight=axes%20set_aspect#matplotlib.axes.Axes.set_aspect)

And the `equal` value is not supported for this version.

But then I checked the latest `matplotlib` version and `equal` is supported.
[Matplotlib Latest Version](https://matplotlib.org/stable/api/_as_gen/mpl_toolkits.mplot3d.axes3d.Axes3D.set_aspect.html#mpl_toolkits.mplot3d.axes3d.Axes3D.set_aspect)

I've checked the codebase and haven't found a version specified for `matplotlib` in `environment.yml`. But the the latest `matplotlib` version is specified in the `uv.lock` file.

I understand that `uv.lock` is not used and is only experimental. But it seems someone is using updated packages from `uv.lock` and still contributing code, thus breaking stuff for people using the versions from `conda`.

I've manually updated `plot_utils.py` and set this to "auto": `ax.set_aspect("auto")`. When I did this, the code worked, and I got to see the cup plotted successfully. 
Image

Hmm, checking the git history of this file 
`gitk src/tbp/monty/frameworks/utils/plot_utils.py` 
it seems the original version of the file from the first commit from 2024 was using `ax.set_aspect("equal")`.

Ok. vclay made the first commit when the codebase was probably open sourced by Numenta. But that doesn't mean she's using the `uv.lock` latest packages and still contributing code.

Can we confirm if someone on the team is actually doing that, and if yes, can we ask them to stop?

jeremyshoemaker:
We've noticed that Conda is installing different versions of our dependencies on different platforms. In particular on Linux, we're installing older versions of `matplotlib` and `scipy`.

This change updates the `environment.yml` files to specify a minimum version for each of these.

Fixes #401


tristanls (approved):
Thank you for tracking this down and the fix.

tristanls (changes requested):
`pyproject.toml` needs to be updated with the same pins to avoid accidental overrides.

jeremyshoemaker:
> `pyproject.toml` needs to be updated with the same pins to avoid accidental overrides.

I keep forgetting we're `pip install -e`ing the project and using those values. I even thought of this and concluded I didn't need to update it because the `uv` locking would manage things correctly for us, but forgetting this fact.

tristanls:
Yeah, the joys of using two dependency sources...

tristanls (approved):

PR #409, opened 2025-07-31
rfc: Project Typing Guidelines

jeremyshoemaker:
This RFC proposes guidelines for how to use Python's type hinting feature with the `mypy` type checker.

### DIFF
--- a/rfcs/0000_project_typing_guidelines.md
+++ b/rfcs/0000_project_typing_guidelines.md
@@ -1,0 +1,164 @@
+- Start Date: 2025-07-31
+- RFC PR: (leave this empty, it will be filled in after RFC is merged)
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14](https://www.rfc-editor.org/info/bcp14) [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) [RFC 8174](https://www.rfc-editor.org/rfc/rfc8174) when, and only when, they appear in all capitals, as shown here.
+
+# Project Typing Guidelines
+
+Python initially introduced type hinting in version 3.5 after being defined in [PEP 484](https://peps.python.org/pep-0484/). The Python interpreter itself generally ignores type hints, so the main way in which they are used is via IDE/LSP support, and through type checking tools like `mypy`. The thinking was for this system to be optional. Any code that would run before would continue to run after the introduction of type hints. It is also in some ways gradual, in that the entire codebase doesn't have to be fully type hinted before some degree of benefit can be realized.
+
+In the future, we would like to utilize type hints in the Monty codebase. However, there are multiple approaches that can be taken to do this, so we want to provide some guidance to ensure we're adding type hints that are giving us the most benefits.
+
+## What does type hinting do for us?
+
+Python already has a _dynamic type_ system. Variables don't have types which restrict the values that can be assigned to them. The values themselves have types that determine what can be done with them, however any check of those types occur at runtime throwing errors when those checks fail.
+
+Using type hints with a type checker implements a _static type_ system. A static type system assigns types to variables, restricting which values can be assigned to them and which operations can be used on those variables and checking those types ahead of time without running the code.
+
+The main benefit of a static type checker enforcing constraints is that it can prevent errors by checking ahead of time for method calls or operations on a variable that would otherwise throw an error at runtime. Type hints also document what arguments are allowed in methods and what types they return. They can also encode logic into the type system to allow for proving certain properties of the code.
+
+## Type hints do nothing without a type checker
+
+One thing to be aware of, the Python interpreter DOES NOT care about type hints. Anything you could normally do in Python without type hints will still be possible at runtime with type hints. The only value they add is when a type checker is used to confirm that the operations being performed match what the type hints indicate is allowed.
+
+This especially applies to newtypes (see below for details). Newtypes do nothing at runtime. There’s a _slight_ performance hit for the call to the “constructor” of the newtype, but it’s negligible. It is, like the rest of type hinting, just a hint, that the Python interpreter ignores, but a type checker can use to ensure correctness.
+
+## Guidelines
+
+### Methods and functions SHOULD accept the broadest possible type
+
+The type assigned to arguments should be as abstract as possible when specifying the arguments to a method or function. For example, if a method needs some collection of items, instead of specifying the type as a `List`, use `Collection` instead.
+
+```python
+# Don't restrict the argument to a List
+def double_list(l: List[int]) -> List[int]
+    return [x * 2 for x in l]
+
+# Instead, use a collection type
+def double_coll(c: Collection[int]) -> List[int]
+    return [x * 2 for x in c]
+```
+
+Using the broadest type that provides the functionality needed allows for more flexibility when calling the function or method. The second example can be called with a set, while the first only works for lists.
+
+### Methods and functions SHOULD return the narrowest possible type
+
+The type returned from a method or function should be the most concrete type possible. Similar to the previous example, the return type could be `Collection[int]`, but that type wouldn't allow for calling list-specific methods on the returned value, even though we're returning a list.
+
+Returning the most concrete type gives more flexibility to the caller to use that value in ways that the function/method author might not have considered.
+
+### Structural typing SHOULD NOT be used when possible.
+
+_Structural typing_ is a type system in which the structure of the type is what matters for type checking. Two values with the same structure will type-check as the same type, regardless of any type aliases being used (see the Python glossary for a stricter definition of [structural types](https://typing.python.org/en/latest/spec/glossary.html#term-structural)).
+
+Basic types like `str`, `list`, and `dict[int, str]` are examples of structural types in Python. Nothing about a `dict[int, str]` indicates what those keys or values represent, and any dictionary that takes integers as keys and has strings as values would type-check against that type.
+
+Type aliases are structural types; they DO NOT lead to nominal types. From the type checker's perspective, they are replaced with the type that they alias. They are only a convenience for not having to write out long structural types.
+
+Structural types should not be used when possible because they don't define the concepts that the types represent, reducing their usefulness.
+
+```python
+# Type alias for a quaternion
+# This doesn't define a new type, it just allows the function
+# definiton below to be shorter.
+Quaternion = Tuple[float, float, float float]
+
+def normalize_quaternion(quat: Quaternion):
+    # Is it this?
+    w, x, y, z = quat
+    # Or is it this?
+    x, y, z, w = quat
+    # do some normalization
+    return normalized_quaterion
+
+quat: Quaternion = (1.0, 0.0, 0.0, 0.0) # this is just of type `tuple`
+norm = normalize_quaternion(quat) # is this giving valid results?
+```
+
+Alternative ways to model a quaternion are using newtypes or dataclasses, both forms of _nominal types_. Which to choose would depend on whether additional functionality is needed, or for easier compatibility with third-party libraries.
+
+```python
+@dataclass
+class Quaternion:
+    w: float
+    x: float
+    y: float
+    z: float
+
+# Or
+QuaternionWXYZ = NewType("QuaternionWXYZ", Tuple[float, float, float, float])
+```
+
+The dataclass approach forces the author to specify which coefficient they want to access, removing ambiguity, while the newtype still requires the author to be careful about which `float` in the `tuple` is the one they want, but the name helps indicate which it is.
+
+An exception to this guideline would be the types of **internal** fields of a class, like the `float`s in the Quaternion dataclass in the previous example. Oftentimes those don’t need to be anything more specific than the underlying structural type. An example might be a name field on a class, which could simply be a `str`. Unless there is some extra metadata, e.g. the `str` is untrusted user input that should be handled carefully, a simple `str` will suffice. This can also apply to things like lists, sets, and dictionaries.
+
+A rule of thumb in most cases would be to ask whether the type needs to be exposed outside of the class or function in which the variable is defined. Function or method arguments and return types should generally avoid structural types in favor of nominal types.
+
+### Nominal typing SHOULD be used to name concepts using types
+
+_Nominal typing_ is a type system in which the name of the type is what matters for type checking. Two values with the same structure but different names are distinct types in this system (see the Python glossary for a stricter definition of [nominal type](https://typing.python.org/en/latest/spec/glossary.html#term-nominal)).
+
+More complex types, like dataclasses and regular classes, are examples of nominal types in Python. Two dataclasses with the same fields but different names would type-check as different types. Classes and dataclasses (which are a convenience for defining certain kinds of classes) are more opaque than anything to do with the fields or methods they have. Different classes are different types and the only way instances of those classes can type check as each other is if one is a subclass of another. 
+
+Python also provides _newtype_s which can be used to define nominal types for basic types that would normally be structurally typed. An example of a newtype would be to define the concept of an integer ID for a user or agent in a system, and ensure that they can't be used in the wrong places.
+
+```python
+UserID = NewType('UserID', int)
+AgentID = NewType('AgentID', int)
+
+# Both UserID and AgentID are integers, but they cannot be swapped.
+def activate_agent(agent: AgentID, actor: UserID) -> None:
+    pass
+
+user_id = UserID(1)
+agent_id = AgentID(2)
+
+activate_agent(agent_id, user_id) # works fine
+# FAILS to typecheck: "UserID" is not "AgentID", etc.
+activate_agent(user_id, agent_id)
+# FAILS to typecheck: "int" is not "UserID"/"AgentID"
+activate_agent(1, 2)
+```
+
+Another example would be to codify the order of a quaternion (since there is disagreement between libraries whether to use WXYZ or XYZW) and then check usages.
+
+Newtypes should be used when **no additional functionality** beyond the underlying type is needed, but metadata about the type needs to be tracked. An example of this would be unsafe and safe strings in a web application. They shouldn’t be confused because they can lead to security vulnerabilities, but with newtypes we can help the author to think about which is being used.
+
+```python
+UnsafeString = NewType('UnsafeString', str)
+SafeString = NewType('SafeString', str)
+
+def render_with_content(template: SafeString, content: SafeString):
+    # This isn't the best way to do this but it's an example
+    return format(template, content)
+
+def sanitize_string(s: UnsafeString) -> SafeString:
+    # do some sanitization
+    return SafeString(new_s)
+
+unsafe_input: UnsafeString # comes from some user input
+template: SafeString = "Hello, {}"
+# FAILS to typecheck: "UnsafeString" is not "SafeString"
+render_with_content(template, unsafe_input)
+
+safe_input = sanitize_string(unsafe_input)
+# FAILS to typecheck: "SafeString" is not "UnsafeString"
+safe_input = sanitize_string(safe_input)
+```
+
+In this example, it becomes more difficult to accidentally pass an unsafe string to a rendering function that could cause security issues, and it’s also difficult to accidentally sanitize a safe string a second time.
+
+### Protocols SHOULD be used to define behaviour that would normally be duck typed
+
+_Duck typing_ is a form of structural typing in dynamically typed languages where two different value types can be substituted for each other if they have the same methods. The name comes from the duck test. “If it walks like a duck, swims like a duck, and quacks like a duck, then it probably is a duck.”
+
+Python’s type hinting library provides Protocols to allow defining the shared interface that types can implement in order to be considered the same type by a type checker. They allow defining interfaces that other types use without requiring those types to inherit from a base class to do it. A Protocol can be implemented by types elsewhere without those types even knowing about the Protocol’s existence.
+
+Protocols allow for defining abstract interfaces that types can satisfy to allow functions and methods to accept the broadest possible types. The abstract collection types, e.g. `Collection`, `Iterable`, etc., in the standard library can be thought of like protocols that various concrete collection types implement. This can be seen in the first example above.
+
+### Third-party libraries with poor type hinting SHOULD be isolated as much as possible.
+
+Some third-party libraries (e.g., Numpy/Scipy) provide poor type hinting, meaning the majority of the benefits that come from static type checking aren't available. To minimize this surface area, we should isolate code that uses these third-party libraries, wrapping them in functions or methods that provide the correct types that the rest of our code expects.
+
+This would be a good place to use nominal types to define things like quaternions in a particular order so that callers don't pass the wrong values into these external library functions.


rfcs/0000_project_typing_guidelines.md
```diff
@@ -0,0 +1,164 @@
...
+```python
+@dataclass
+class Quaternion:
+    w: float
+    x: float
+    y: float
+    z: float
+
+# Or
+QuaternionWXYZ = NewType("QuaternionWXYZ", Tuple[float, float, float, float])
```
vkakerbeck: Do I understand it right that the main difference between this and `Quaternion = Tuple[float, float, float float]` is that we are giving it a more meaningful name? Or is there something else that the use of `NewType` adds?

vkakerbeck: ah I think reading further, this is a nominal type now, right?

rfcs/0000_project_typing_guidelines.md
```diff
@@ -0,0 +1,164 @@
...
+render_with_content(template, unsafe_input)
+
+safe_input = sanitize_string(unsafe_input)
+# FAILS to typecheck: "SafeString" is not "UnsafeString"
+safe_input = sanitize_string(safe_input)
+```
+
+In this example, it becomes more difficult to accidentally pass an unsafe string to a rendering function that could cause security issues, and it’s also difficult to accidentally sanitize a safe string a second time.
+
+### Protocols SHOULD be used to define behaviour that would normally be duck typed
```
vkakerbeck: It could be worth to also link to our other guidance on Protocols here https://thousandbrainsproject.readme.io/docs/code-style-guide#protocols-should-be-preferred-to-document-usage-and-expectations (unless this will be integrated in the same place anyways and the reader has just read that)

vkakerbeck (approved):
Nice! Looks good to me :)

rfcs/0000_project_typing_guidelines.md
```diff
@@ -0,0 +1,164 @@
...
+
+Protocols allow for defining abstract interfaces that types can satisfy to allow functions and methods to accept the broadest possible types. The abstract collection types, e.g. `Collection`, `Iterable`, etc., in the standard library can be thought of like protocols that various concrete collection types implement. This can be seen in the first example above.
+
+### Third-party libraries with poor type hinting SHOULD be isolated as much as possible.
+
+Some third-party libraries (e.g., Numpy/Scipy) provide poor type hinting, meaning the majority of the benefits that come from static type checking aren't available. To minimize this surface area, we should isolate code that uses these third-party libraries, wrapping them in functions or methods that provide the correct types that the rest of our code expects.
```
ramyamounir: Do you mean if a function returns a numpy array, we should try to convert it to a `List` so that the numpy operations are isolated within the function? Would there be significant computational overhead to this approach (i.e., creating many new numpy arrays within isolated functions)?

Or did you mean we should create nominal types for arguments with 3rd-party libraries and use those types instead?

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: Yes, it makes it a nominal type. I found it tricky to talk about structural typing without mentioning nominal typing and vice versa, so maybe I could expand this example a little bit to make it clearer.

The benefit of this is that you can't just use a raw tuple in places where a `QuaternionWXYZ` are needed and have it type check. The Python interpreter won't complain, because at runtime it's still just a tuple, but the type checker will mark it as an error.

In use, you'd have to do something like what happens in the UserID/AgentID example below where you wrap the raw tuple with a call to `QuaternionWXYZ()`.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: The plan is to put them in the same place, and since the two documents complement each other, linking between them makes sense.

rfcs/0000_project_typing_guidelines.md
```diff
@@ -0,0 +1,164 @@
...
+
+A rule of thumb in most cases would be to ask whether the type needs to be exposed outside of the class or function in which the variable is defined. Function or method arguments and return types should generally avoid structural types in favor of nominal types.
+
+### Nominal typing SHOULD be used to name concepts using types
+
+_Nominal typing_ is a type system in which the name of the type is what matters for type checking. Two values with the same structure but different names are distinct types in this system (see the Python glossary for a stricter definition of [nominal type](https://typing.python.org/en/latest/spec/glossary.html#term-nominal)).
+
+More complex types, like dataclasses and regular classes, are examples of nominal types in Python. Two dataclasses with the same fields but different names would type-check as different types. Classes and dataclasses (which are a convenience for defining certain kinds of classes) are more opaque than anything to do with the fields or methods they have. Different classes are different types and the only way instances of those classes can type check as each other is if one is a subclass of another.
```
ramyamounir: In my mind, creating a new nominal type e.g., instead of a structural type, is effectively forcing us to use narrower types. A function that used to accept a simple float, can now accept only this nominal type even though the broader float type technically would still work. Does the use of nominal types contradict with the rule of "Methods and functions SHOULD accept the broadest possible type"?

rfcs/0000_project_typing_guidelines.md
vkakerbeck: okay that makes sense. Maybe adding a simple "(more on NewType in the next section)" can defer questions of the curious reader :D

rfcs/0000_project_typing_guidelines.md
ramyamounir: > Different classes are different types and the only way instances of those classes can type check as each other is if one is a subclass of another. 

Could you provide an example of this? Specifically, the cases where a class can type check for another. So in the below example, `Hypotheses` would type check for `ChannelHypotheses`?

```python
@dataclass
class Hypotheses:
    """Set of hypotheses consisting of evidence, locations, and poses.

    The three arrays are expected to have the same shape. Each index corresponds to a
    hypothesis.
    """

    evidence: np.ndarray
    locations: np.ndarray
    poses: np.ndarray


@dataclass
class ChannelHypotheses(Hypotheses):
    """A set of hypotheses for a single input channel."""

    input_channel: str
```

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: I can probably add some clarifications, or examples to this section.

A lot of libraries, especially ones that are implemented as extensions written in C, tend to not have type annotations, and so from the perspective of the type checker, they return some general type like `object`. I've run into this with the MuJoCo library as well.

So, the suggestion here is to isolate that code into functions or methods that are annotated to return the expected type, like `np.array` or whatever the actual type is.

We're not suggesting wrapping individual functions in the third-party library, but instead if there's a bit of code that does a bunch of related NumPy calculations, isolating those more and having them return a reasonable type.

This guidance might change as we try to put it into practice. Since I originally wrote it, I've run into some of this with the MuJoCo simulator work, and that might end up changing what this suggests.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: To your first comment, the goal of using a newtype instead of a raw `float` is that there is some additional meaning that the newtype provides to prevent mistakes. The UserID/AgentID example is one example, but something more related to what we might run into is if say we had some code that used radians and some code that used degrees. Using a newtype for each would make it so you can't mix those types of `float`s up. I don't think this conflicts with accepting the broadest type possible. `Radians(2.0)` is not the same as `Degrees(2.0)` even though the underlying datatype is just a `float`.

To your second comment, ~yes, that's correct~. You're the second person to ask about this, so maybe it's not as common knowledge as I assumed (and that's okay), and I should expand this section a bit.

Dataclasses are just a shorthand way of creating regular classes, so these just being classes, this falls back to object-oriented programming principles, in particular _subclass polymorphism_.

jeremyshoemaker: As I was writing an update for this section to provide an example, I found that NumPy actually has type hints in the version we're using, so it might not be the best example of this problem.

Generally this should be less and less of a problem as time goes on and more libraries add type hints.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: EDIT: I meant to put this comment on this thread, not the other one, so I moved it.

As I was writing an update for this section to provide an example, I found that NumPy actually has type hints in the version we're using, so it might not be the best example of this problem.

Generally this should be less and less of a problem as time goes on and more libraries add type hints.

EDIT: @tristanls pointed out that the NumPy type hints aren't great for other reasons.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: I made some updates in https://github.com/thousandbrainsproject/tbp.monty/pull/409/commits/73a969de64119858ee7157eb6d3e07d5e760fd96. Let me know what you think.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: Actually, I think what you asked is backwards. From your example code, you can use a `ChannelHypotheses` where something expects a `Hypotheses`, but not the other way around.

jeremyshoemaker: I updated this section a bit in https://github.com/thousandbrainsproject/tbp.monty/pull/409/commits/414c5b337af4bd3d1ea86b7e3a206927e5f05cc0. Let me know what you think.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: I added some clarification in https://github.com/thousandbrainsproject/tbp.monty/pull/409/commits/84bb2d143ebfde552996b7e4015bac8ae7e9be97. Let me know what you think.

rfcs/0000_project_typing_guidelines.md
```diff
@@ -0,0 +1,176 @@
...
+
+Using the broadest type that provides the functionality needed allows for more flexibility when calling the function or method. The second example can be called with a set, while the first only works for lists.
+
+### Methods and functions SHOULD return the narrowest possible type
+
+The type returned from a method or function should be the most concrete type possible. Similar to the previous example, the return type could be `Collection[int]`, but that type wouldn't allow for calling list-specific methods on the returned value, even though we're returning a list.
+
+Returning the most concrete type gives more flexibility to the caller to use that value in ways that the function/method author might not have considered.
+
+### Structural typing SHOULD NOT be used when possible.
```
tristanls: **issue**: This phrasing is problematic as Protocols, which we do recommend and for structural reasons, are structural types. Also, clicking on https://typing.python.org/en/latest/spec/glossary.html#term-structural displays: "A structural type (see e.g. Protocols, TypedDict)..."

tristanls: **suggestion**: Maybe "Structural typing SHOULD NOT be used unless explicitly recommended" or something along those lines? It seems like parsing the meaning starts to get complicated with this suggestion, though.

jeremyshoemaker: That's an interesting edge case. 

Thinking this out, my first thought is to say I feel like there's a distinction between Protocols and other structural types because most structural types have to do with... well... the structure of the thing, e.g. what kind of fields it has, in the case of `TypedDict` what keys it has, etc. With Protocols, they're more about behaviour. They _don't_ add any structure.

That said one could also say that part of structure is what you can do with a thing, in which case behaviour falls into that category.

When I wrote all of this, I didn't consider Protocols to be a form of structural typing... but now I'm not so sure.

tristanls: The title of [PEP 544](https://peps.python.org/pep-0544/) is "PEP 544 – Protocols: Structural subtyping (static duck typing)" 👀 😄.

jeremyshoemaker: Well, that's structural _subtyping_. There is some nuance there, but it's not exactly relevant here. I accept the argument that Protocols are structural.

How do you feel about having a line in this section mentioning that Protocols are a special case and to see the guidance below on protocols?

I basically treated _duck typing_ as its own separate thing in this document and didn't notice the overlap with _structural typing_ until you pointed it out.

tristanls: > How do you feel about having a line in this section mentioning that Protocols are a special case and to see the guidance below on protocols?

I think that should be enough. It captures the intent of my suggestion.

jeremyshoemaker: Added some clarification in https://github.com/thousandbrainsproject/tbp.monty/pull/409/commits/2ad12b5f100942628d08c2723a75504a94ca627d. Also added hyperlinks while I was at it, and cleared up some RFC keyword usage.

tristanls: LGTM

jeremyshoemaker:
> Thanks. I have a few more suggestions, though I'm not sure if some are too detailed for the RFC:
> 
>  1. **Optional**: I think it may be nice to add that using type hints would force one to think about better data structures. If there is something you are typing with some complex Union or other messy nest like `Dict[str, Dict[str, Dict[str, str]]]`, then maybe it should be a dataclass or something else. Type hints can lead to better design.
> 
>  2. **Optional**: This one may be too specific, but using `Generics` and `TypeVar`.
>     ...
>  3. This one I feel more strongly about: **Avoid Any Unless Absolutely Necessary**
>        In fact, Ruff [ANN401](https://docs.astral.sh/ruff/rules/any-type/) does not allow for `Any` type (though we currently allow it).

These are all good, but I'm not sure we need to include all of them in the typing guidelines. Going through each of them:

1. Type hints leading to better design is a _result_ of using them consistently, it's not exactly a guideline. Your example of the nested dictionaries vs using dataclasses matches the "Structural types SHOULD NOT be used when possible" guidance. I didn't discuss Union types, mostly because I don't think they occur that often outside of `something | None`. Anything more generally indicates that a function or method is doing too much and probably has a large `if/else` block to work with the various possible types in the Union.
2. I didn't discuss generics because I wasn't sure how often they'd come up in our codebase, and I didn't want information overload from talking about things that might not be relevant to us.
3. I could definitely add that one, but it's also something we can just enforce with the Ruff check. It might make sense to add it until we're in a state where we can activate the check. 

The trouble with a guideline for 3 is it would have to be something like "The `Any` type SHOULD NOT be used unless absolutely necessary", and then we'd need to provide guidelines for when it can be used, because we have places in the code where we _are_ using it.

I'll spend some time working on how to phrase that, to see if I can come up with something that works.

rfcs/0000_project_typing_guidelines.md
ramyamounir: This makes sense, thanks.

rfcs/0000_project_typing_guidelines.md
ramyamounir: This is clear now, thanks.

ramyamounir (approved):
Looks great, thanks for the changes!

nielsleadholm:
Unfortunately I won't be able to review this until I'm back next week. @jeremyshoemaker let me know if that's going to be an issue/blocker for you, and I can either see if I can figure something out, or abstain.

hlee9212:
> * Type hints leading to better design is a _result_ of using them consistently, it's not exactly a guideline. Your example of the nested dictionaries vs using dataclasses matches the "Structural types SHOULD NOT be used when possible" guidance. I didn't discuss Union types, mostly because I don't think they occur that often outside of `something | None`. Anything more generally indicates that a function or method is doing too much and probably has a large `if/else` block to work with the various possible types in the Union.
> * I didn't discuss generics because I wasn't sure how often they'd come up in our codebase, and I didn't want information overload from talking about things that might not be relevant to us.
> * I could definitely add that one, but it's also something we can just enforce with the Ruff check. It might make sense to add it until we're in a state where we can activate the check.

Sounds good, I changed from Review Requested to Approved. 

Regarding the use of `Any`, I think their few use cases may be in:
(1) wrappers or decorators
```python
from typing import Any, Callable

def my_decorator(func: Callable[..., Any]) -> Callable[..., Any]:
    def wrapper(*args: Any, **kwargs: Any) -> Any:
        print("Before call")
        return func(*args, **kwargs)
    return wrapper
```
(2) really truly dynamic data (from untyped sources), e.g. APIs, user inputs, configs 

For me just a general guideline of "SHOULD be used sparingly and deliberately" with the above two cases is fine. I'm not sure but it could be a while until we enforce the Any ruff rule.

jeremyshoemaker:
@hlee9212 I added some guidance in https://github.com/thousandbrainsproject/tbp.monty/pull/409/commits/b4cd16c3dd61658e34f78f7ea3f00b875f1c5169

@tristanls reminded me that we're not enumerating exceptions here, as those will be resolved on a case-by-case basis during code reviews, so I ended up not including any.

> (2) really truly dynamic data (from untyped sources), e.g. APIs, user inputs, configs

That's not nearly as dynamic as you might think. A great article I read that changed my mind on the topic is [No, dynamic type systems are not inherently more open](https://lexi-lambda.github.io/blog/2020/01/19/no-dynamic-type-systems-are-not-inherently-more-open/). I highly recommend giving it a read, but the gist of it is that dynamic code will end up doing things like checking for the existence of specific dictionary keys or the types of values, and all of that is type information that could be encoded in a type and checked statically.  And you can do all of this while also not having to write an exhaustive type that defines everything.

hlee9212:
Thanks @jeremyshoemaker for the update! LGTM.

> That's not nearly as dynamic as you might think. A great article I read that changed my mind on the topic is [No, dynamic type systems are not inherently more open](https://lexi-lambda.github.io/blog/2020/01/19/no-dynamic-type-systems-are-not-inherently-more-open/). I highly recommend giving it a read, but the gist of it is that dynamic code will end up doing things like checking for the existence of specific dictionary keys or the types of values, and all of that is type information that could be encoded in a type and checked statically. And you can do all of this while also not having to write an exhaustive type that defines everything.

And totally agree. Was going to add something like "(but configs and other things shouldn't be dynamic / there are schemas and things, etc.)" but didn't want to go in depth 🤣. Great resource, btw.

rfcs/0000_project_typing_guidelines.md
```diff
@@ -0,0 +1,184 @@
...
+
+One thing to be aware of, the Python interpreter DOES NOT care about type hints. Anything you could normally do in Python without type hints will still be possible at runtime with type hints. The only value they add is when a type checker is used to confirm that the operations being performed match what the type hints indicate is allowed.
+
+This especially applies to newtypes (see below for details). Newtypes do nothing at runtime. There’s a _slight_ performance hit for the call to the “constructor” of the newtype, but it’s negligible. It is, like the rest of type hinting, just a hint, that the Python interpreter ignores, but a type checker can use to ensure correctness.
+
+## Guidelines
+
+### Methods and functions SHOULD accept the broadest possible type
+
+The type assigned to arguments should be as abstract as possible when specifying the arguments to a method or function. For example, if a method needs some collection of items, instead of specifying the type as a `List`, use `Collection` instead.
```
scottcanoe: **suggestion (non-blocking)**: I think this example should probably use `Iterable` instead of `Collection` since I don't believe a `Collection` requires it can be iterated over (and this examples doesn't use `len(...)`.

scottcanoe (approved):
LGTM.

Side note: I'm hoping we can come at some issues from multiple angles. Type-checking yes, but also let's codify a standard quaternion representation for Monty, for example.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: You're right, it can be an `Iterable` because we aren't depending on the length. Those two are close enough that usually it won't matter but in the spirit of the rule, it should be changed to `Iterable`.

`Collection`s are "sized iterable containers" (from [the collections documentation](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes-detailed-descriptions)).

jeremyshoemaker:
> Side note: I'm hoping we can come at some issues from multiple angles. Type-checking yes, but also let's codify a standard quaternion representation for Monty, for example.

Definitely. I think the type checking can support the standardization.

For quaternions in particular, I kinda lean towards the dataclass example, with some supporting methods to convert to tuples or arrays or whatever is need, because with the dataclass it's clear which field is which coefficient.

That said, I don't know how often we convert things, or would need to transform representations, so I wouldn't want to add anything that could cause a big performance impact.

One option I didn't discuss, because I just forgot about it, is using a NamedTuple. It's somewhere between a custom class and a NewType. Not sure what it adds though in this case, and I'd just as soon prefer to settle on one approach for things like this.

rfcs/0000_project_typing_guidelines.md
```diff
@@ -0,0 +1,184 @@
...
+
+## What does type hinting do for us?
+
+Python already has a _dynamic type_ system. Variables don't have types which restrict the values that can be assigned to them. The values themselves have types that determine what can be done with them, however any check of those types occur at runtime throwing errors when those checks fail.
+
+Using type hints with a type checker implements a _static type_ system. A static type system assigns types to variables, restricting which values can be assigned to them and which operations can be used on those variables and checking those types ahead of time without running the code.
+
+The main benefit of a static type checker enforcing constraints is that it can prevent errors by checking ahead of time for method calls or operations on a variable that would otherwise throw an error at runtime. Type hints also document what arguments are allowed in methods and what types they return. They can also encode logic into the type system to allow for proving certain properties of the code.
```
nielsleadholm: **suggestion (non-blocking):** At some point we/someone else might implement certain components of Monty in a strongly typed language like C++  for optimization purposes. Not sure if this is actually true, but I would imagine in that case that having more explicit typing in our Python code would make that process much quicker.

rfcs/0000_project_typing_guidelines.md
```diff
@@ -0,0 +1,184 @@
...
+another_angle = Degrees(180.0)
+
+rads_to_degrees(right_angle) # works fine
+# FAILS to typecheck: "Degrees" is not "Radians", etc.
+rads_to_degrees(another_angle)
+# FAILS to typecheck: "float" is not "Radians", etc.
+rads_to_degrees(270.0)
+```
+
+Another example would be to codify the order of a quaternion (since there is disagreement between libraries whether to use WXYZ or XYZW) and then check usages (see the quaternion example in [the section on structural typing](#structural-typing-should-not-be-used-when-possible)).
```
nielsleadholm: **suggestion (non-blocking):** I think it would be helpful to re-visit briefly the quaternion example, and basically show how you cannot do:

```python
QuaternionWXYZ = Tuple[float, float, float, float]
QuaternionXYZW = Tuple[float, float, float, float]
```

As these are structural rather than nominal types, the type-checker won't be able to distinguish them (that, at least, is my understanding).

rfcs/0000_project_typing_guidelines.md
```diff
@@ -0,0 +1,184 @@
...
+
+Protocols allow for defining abstract interfaces that types can satisfy to allow functions and methods to accept the broadest possible types. The abstract collection types, e.g. `Collection`, `Iterable`, etc., in the standard library can be thought of like protocols that various concrete collection types implement. This can be seen in the first example above.
+
+### The `Any` type SHOULD NOT be used
+
+The `Any` type in Python causes the type checker to stop attempting to do type checking, since it has no information whatsoever to go on. This means we lose all the benefits of using a static type checker, and thus it SHOULD NOT be used.
+
+Another similar type is `object`. While it can be used with arbitrary typed values similar to `Any`, it is an ordinary static type so the type checker will reject most operations on it (i.e., those not defined for all objects), and so it MAY be used where appropriate (see the Mypy documentation on [Any vs. object](https://mypy.readthedocs.io/en/stable/dynamic_typing.html#any-vs-object) for more details).
```
nielsleadholm: **question:** Is there a concrete example that comes to mind where this might be appropriate in Monty? No worries if not, but just curious.

rfcs/0000_project_typing_guidelines.md
```diff
@@ -0,0 +1,184 @@
...
+
+```python
+def do_maths(input: ...) -> npt.NDArray[np.float64]:
+    # do lots of Numpy calls
+    return result
+```
+
+This might require the use of [`typing.cast()`](https://docs.python.org/3/library/typing.html#typing.cast) to satisfy the type checker.
+
+The granularity SHOULD NOT be individual NumPy functions, but rather whole operations in our code where we are using multiple NumPy functions in a row. We're not trying to make a wrapper for poorly typed libraries, but making sure when we return things into our code, store values on objects, etc. that we're giving them useful types.
```
nielsleadholm: **suggestion (non-blocking):** Related to Ramy's point, it might be worth explicitly calling out that this should be done in a way to minimize performance impacts. Typing is an issue if it requires expensive recreation of data-structures. My understanding is that this should be avoidable, but we should be mindful of this possibility and avoid it.

nielsleadholm (approved):
Really nice RFC. Just added a few non-blocking suggestions/questions.

rfcs/0000_project_typing_guidelines.md
tristanls: https://github.com/thousandbrainsproject/tbp.monty/blob/c27f1d57c6907dd7633352ad485443730cf40837/src/tbp/monty/frameworks/config_utils/config_args.py#L91

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: From what little work I've done writing Python extensions in Rust, and reading Python extensions written in C, all of the Python objects are very dynamically typed even there. I'm not sure that an extension would have access to the type hinting metadata or not. But even if it did, you couldn't trust it because at runtime Python will let you do things that your type hints say you can't do.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: Yeah, from the type checker's perspective, these are just `Tuple[float, float, float, float]` and thus the same. Making it a NewType or a dataclass would switch it to being nominally typed.

I'll take a look at that section again and see if I can make that more explicit.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: Using `object` for a type is usually going to be quite limiting, but unlike `Any` the type checker will make sure you don't try to do things with the value that you can't do to an `object`. Most of the examples I can come up with are contrived, e.g. a heterogeneous list where each element would need to be narrowed to their particular type with type checks in order to be used, but I wouldn't actually recommend doing something like that.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: I've been thinking about this one as I've been working on the MuJoCo simulator, and I might need to clarify it a bit.

One of the things I've been running into is that most, if not all of MuJoCo's functions and methods aren't typed, so I almost always have to specify the type when assigning the return value to a variable.

```python
worldbody: MjBody = spec.worldbody  # this would be of type Any otherwise
``` 

That's not something that's solved by wrapping it, because a few lines later I might need to use that object and get another value that would need to have a type hint.

rfcs/0000_project_typing_guidelines.md
nielsleadholm: I was more thinking if someone takes function x and wants to re-write it in C++, then it is helpful to be able to immediately see what the various data types are, rather than having to spend time reading and understanding the code. I know that a lot of types don't translate directly, so yeah might not be that helpful.

jeremyshoemaker: Ah, I was thinking of performance. Yeah, it would probably help with developing an extension, having an idea of what the values can be rather than tracing the code.

rfcs/0000_project_typing_guidelines.md
nielsleadholm: Thanks, fwiw @tristanls I meant about the `object` example, rather than `Any`. 

I asked ChatGPT and it suggested the following. If this is true, then it might be helpful to use this example.

> 
> What could be an “object is appropriate” example
> 
> You’d choose object when:
> - You truly don’t care about the specific type,
> - But you also want to prevent accidentally calling arbitrary methods or assuming structure.
> 
> Example:
> ```python
> from typing import Iterable
> 
> def first_item(items: Iterable[object]) -> object:
>     for item in items:
>         return item
>     raise ValueError("Empty iterable")
> ```
> 
> Here:
> - We don’t know what’s in the iterable.
> - We don’t want type checking to allow item.some_method().
> - Using Any would allow that, but object blocks it.
>

rfcs/0000_project_typing_guidelines.md
nielsleadholm: Ok, to make sure I follow, are you concerned that _would_ impact performance? My main point is just that we highlight that performance should generally be maintained when we make changes.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: The trouble with that example is that you can't do anything with the returned `object` unless you narrow it. That would actually be a place where generics would be useful. This is the second time generics have come up one way or another, so maybe it would make sense to include a section with guidance for that.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: I would be concerned about performance if we started doing things like:

```python
def get_worldbody(spec: MjSpec) -> MjBody:
    return cast(MjBody, spec.worldbody)
    
def some_other_method(...):
    ...
    worldbody = get_worldbody(spec)
    ...
```

Wrapping individual calls or interactions with a third party library that doesn't provide good typing would be excessive, because you're paying for extra function calls which aren't free, and as far as I know the Python interpreter doesn't do any sort of [inline expansion](https://en.wikipedia.org/wiki/Inline_expansion).

nielsleadholm: Ok thanks. Might be worth including this then as an example of excessive type-setting.

rfcs/0000_project_typing_guidelines.md
jeremyshoemaker: After some thought, I'm not sure what any generics guidance should be, aside from use them if they make sense.

I'm not sure where in our code base we'd be writing things that need to work with a collection of items without knowing their types, and therefore could make use of generics.

I'd rather not add a section on it without knowing how we'd want to use it.

jeremyshoemaker:
@nielsleadholm I made some edits in response to your comments. See https://github.com/thousandbrainsproject/tbp.monty/pull/409/commits/e24322e3ab17c8392bf87f15be4ca42f39535c02

rfcs/0000_project_typing_guidelines.md
```diff
@@ -177,8 +176,26 @@ def do_maths(input: ...) -> npt.NDArray[np.float64]:
...
+
+# Instead, just declare the types in the other method where needed
+def some_other_method(...) -> MjsBody:
+   ...
+   worldbody: MjsBody = spec.worldbody
+   # More MuJoCo calls on worldbody that may need type hints
+   ...
+```
+
+This would be also good place to use newtypes to define _input_ argument types like quaternion tuples that have a particular order so that callers don't pass the wrong values into these external library functions. See [the nominal type guidance](#nominal-typing-should-be-used-to-name-concepts-using-types) above.
```
nielsleadholm:
```suggestion
This would also be a good place to use newtypes to define _input_ argument types like quaternion tuples that have a particular order so that callers don't pass the wrong values into these external library functions. See [the nominal type guidance](#nominal-typing-should-be-used-to-name-concepts-using-types) above.
```

nielsleadholm (approved):
Looks great, thanks for the updates.

rfcs/0000_project_typing_guidelines.md
```diff
@@ -0,0 +1,201 @@
...
+    return result
+```
+
+This might require the use of [`typing.cast()`](https://docs.python.org/3/library/typing.html#typing.cast) or explicit type hints on variable declarations to satisfy the type checker.
+
+The granularity SHOULD NOT be individual NumPy functions, but rather whole operations in our code where we are using multiple NumPy functions in a row. We're not trying to make a wrapper for poorly typed libraries, but making sure when we return things into our code, store values on objects, etc. that we're giving them useful types.
+
+```python
+# Don't do something this granular
+def get_worldbody(spec: MjSpec) -> MjsBody:
```
tristanls: **issue**: Please, let's not provide examples of discouraged use of the `get` noiseword.

**suggestion**: I think a better example here would be one of `worldbody(...)`.

**note**: If that conflicts with the variable name, I'd suggest `wbody` for the variable name, as the examples have no code distance concerns.

jeremyshoemaker: Fair point. I figured it wouldn't matter in a bad code example, but you're right, a bad code example should show one bad thing and not multiple.

Fixed in ff81fbd

jeremyshoemaker:
## Motion for Final Comment Period

@thousandbrainsproject/monty-maintainers I would like to motion for FCP and invite all maintainers to indicate their final votes on the merging of this RFC.
👍 for merge
👎 for object
👀 for abstain

PR #418, opened 2025-08-07
docs: Fix for vale spell checking not catching misspellings

codeallthethingz:
Turns out that the accept file is odd.  It doesn't by default match against word boundaries. So very small sequences of letters in that file causes vale to skip valid mispellings.

In this case "Gu" in accept.txt was causing vale to not catch "distingush" in one of our documents.  This fix adds word boundaries to all the accept entries.


tristanls (approved):
👍

PR #419, opened 2025-08-09
refactor: remove unused MontyBase attributes `sm_ids` and `n_lms`.

scottcanoe:


jeremyshoemaker (approved):
LGTM

tristanls:
**note**: Updated title to `refactor!:` as removing existing attributes is a breaking change (even if unused by us).

scottcanoe:
@tristanls Thanks for that, I was unsure about the breaking-ness in cases like this.

tristanls:
@scottcanoe, no worries. In general, removing "public" things (attributes, properties, methods, functions, classes, etc.) should default to being a breaking change. The reason is that if someone wrote code that accessed `sm_ids` or `n_lms`, that code would now break. In Python in particular, it can be challenging to define what "public" is, though. We only have a convention to follow, so I'll treat anything starting with `_` as "private" (e.g., `_things`), and if that gets removed, then I wouldn't call that a breaking change, as the users shouldn't be relying on it.

Issue #422, opened 2025-08-12
chore: Change getters to properties using @property in simulator.py


annark:
Change getters to properties using `@property` in simulator.py per this [TODO](https://github.com/thousandbrainsproject/tbp.monty/blob/639618954454e85082783a77e6efbcc8b6721950/src/tbp/monty/simulators/simulator.py#L66).

tristanls:
Resolved in the scope of #424

PR #424, opened 2025-08-12
refactor!: Change getters to properties using @property in simulator.py. 

annark:
Change getters to properties using `@property` in simulator.py. Fixes issue #422

To note:
- `HabitatSim` has a similar set of getter methods to those changed in this PR, however it does not implement the `Simulator` protocol, so there are no changes to it here.
- In addition to looking for references to the changed methods across the codebase/documentation, I searched the `monty_labs` repo. No files there use the `Simulator` protocol.


tristanls:
**note**: I'm sorry. I re-read the description. Here's a bit more on why HabitatSim implements the protocol.

Protocol implementation does not need to be explicitly declared; it uses [structural subtyping](https://typing.python.org/en/latest/spec/protocol.html#terminology). As long as a class implements the methods in the protocol, it can be used where that protocol is expected. The [Defining a protocol section](https://typing.python.org/en/latest/spec/protocol.html#defining-a-protocol) has an example of this (the `Resource` class).

> A structural type (see e.g. [Protocols](https://typing.python.org/en/latest/spec/protocol.html#protocols), [TypedDict](https://typing.python.org/en/latest/spec/typeddict.html#typeddict)) defines a set of values not by their `__class__`, but by their properties (e.g. attributes, methods, dictionary key/value types). [Callable](https://typing.python.org/en/latest/spec/callables.html#callable) types are also structural; a callable type is a subtype of another callable type based on their signatures, not a subclass relationship. In contrast, see [nominal](https://typing.python.org/en/latest/spec/glossary.html#term-nominal) types.
> -- https://typing.python.org/en/latest/spec/glossary.html#term-structural

tristanls:
**note**: I updated the pull request to be `refactor!:` (A src/tbp/monty code change that neither fixes a bug nor adds a feature.) It is a breaking change because it changes the public Simulator API.

annark:
Thanks Tristan, that all makes sense. I'll also make the changes to `HabitatSim` as you describe.

tristanls:
Thank you for the updates @annark. Please re-request a review when you're ready for us to take another look.
Screenshot 2025-08-25 at 10 20 58

**issue**: I did notice edits to RFC 4. Please do not edit merged RFCs; we don't want to update those.

annark:
Thanks Tristan. I removed the edits to RFC 4, and re-requested review. I've updated my fork and branch with the latest monty repo changes, successfully run tests, and run the recommended checks:
```
(tbp.monty) ➜  tbp.monty git:(change-getters-to-properties) ✗ ruff check
All checks passed!
(tbp.monty) ➜  tbp.monty git:(change-getters-to-properties) ruff format
211 files left unchanged
```
However I'm seeing these [failed checks](https://github.com/thousandbrainsproject/tbp.monty/actions/runs/17226876425). I'm not sure what is causing them, if you had any insight?

tristanls:
@annark, the failures you're seeing are caused by a problem with how those checks are defined. I'll look into fixing them.

tristanls:
@annark #442 should resolve the problem with the checks. Once that is merged, it should help with the check failures encountered here.

tristanls:
@annark , I merged #442. I see that you are rebasing to update the branch, so I will rebase this branch on the latest main and see what happens to the checks.

annark:
simulator_test.py removed. Thanks!

tristanls (approved):
Looks great 🥳

PR #425, opened 2025-08-13
rfc: processing off-object observations

hlee9212:
Direct link to [View File](https://github.com/hlee9212/tbp.monty/blob/d976fe8d08c070568b3c2a8f5bdb5163517fbb5e/rfcs/0000_off_object_observations.md).


rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+Currently, `FeatureChangeSM` filters out observations where `on_object` is False. 
+
+```python
+# Current filtering behavior in sensor_modules.py (lines 680-684)
+if not observed_features.get_on_object():
+    logger.debug(f"No new point because not on object")
+    return False
+```
+
+This means that when a sensor moves off an object into empty space, the resulting off-object observation, which could provide valuable prediction error signals, is discarded before reaching the LM.
```
tristanls: **suggestion (non-blocking)**: Maybe it is worth highlighting that, in this case of off-object observation, because the observation is discarded, Monty never even calls `step()` on the LM.

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+- **Touch sensors**: Naturally detect "nothing" when not in contact with surface.
+- **Vision sensors**: Always detect something (even if just background), making the definition of "off-object" more complex and context-dependent.
+
+Taking this into consideration, we define "null features" as the absence of **morphological** features (surface normal and principal curvatures), since depending on the sensing modality, some **non-morphological features** may be present (e.g. background color) even when sensor is off-object.
+
+## Proposed Implementation
+
+The following implementation addresses the limitations identified above by enabling off-object observations to flow through the sensory processing pipeline and be used for hypothesis elimination. The approach involves two main steps: modifying the sensor module's filtering behavior and extending learning modules to process off-object observations.
+
+### Step 1: Update `FeatureChangeSM` to Send Null Features
```
tristanls: **question**: The proposed change line in `sensor_modules.py` is an update to the `DetailedLoggingSM`, not the `FeatureChangeSM`. Do you intend to change `DetailedLoggingSM` (and all its subclasses) or only the `FeatureChangeSM`?

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+        [np.nan, np.nan, np.nan],  
+        [np.nan, np.nan, np.nan],  
+        [np.nan, np.nan, np.nan],  
+    ]),
+    "pose_fully_defined": False,
+    "principal_curvatures": np.array[[np.nan, np.nan]]
+    "gaussian_curvature": np.nan
+    "mean_curvature": np.nan
+    ...
+}
```
tristanls: **issue**: This is a very complex sentinel value.

**question**: Can you help me understand why `morphological_features = None` or a simpler sentinel would not work?

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+
+# Existing code...
+# Proposed update in line 286 of sensor_modules.py
+else: 
+    morphological_features = NULL_MORPHOLOGICAL_FEATURES
+```
+
+We may also consider removing the use of `invalid_signals`, which is currently used to determine `use_state` in line 305 of `sensor_modules.py` (see code below). The current logic sets `invalid_signals` to `True` if surface normal or principal curvature directions are not defined (see `extract_and_add_features()` method in `sensor_modules.py`), but this check becomes less relevant when we process `NULL_MORPHOLOGICAL_FEATURES` for off-object observations.
```
tristanls: **question**: Could you explain why the check is less relevant, especially in the On-object --> On-object case? As far as I can tell from reading the code, it is possible to have surface normal or principal curvature invalid while still populating `morphological_features`. `morphological_features` are set without regard for validity.
```python
        morphological_features = {
            "pose_vectors": np.vstack(
                [
                    surface_normal,
                    dir1,
                    dir2,
                ]
            ),
            "pose_fully_defined": pose_fully_defined,
        }
```
If we removed the check for `not invalid_signals`, could it not be the case that `morphological_features["on_object"]` is true, while we don't have a well-defined surface normal or principal curvature? This would end up with `use_state=True` and LM processing `morphological_features`.

tristanls (commented):
**note**: I reviewed up to but not including `A. Transiition-Aware Logic in check_feature_change()`

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Question and Goal
+
+The question that we are addressing is: How can we send and process off-object observations in Monty?
+
+The goal of the RFC is to layout current state of Monty and implementation plan to process off-object observations. 
+
+## Summary
+
+This RFC proposes extending Monty's sensory processing pipeline to handle off-object observations (when a sensor has moved beyond an object's reference frame). Currently, these observations are filtered out and not passed to Learning Modules (LMs), limiting our ability to leverage prediction errors for penalizing hypotheses' evidence. By defining "null features" to represent the absence of morphological features, modifying the `FeatureChangeSM` to forward these observations, and updating the Learning Module to process off-object observations, we can enable more intelligent hypothesis management when sensors transition from on-object to off-object status. This enhancement will improve Monty's ability to recognize when it has moved off an object and reduce evidence for hypotheses that predict the sensor should still be detecting object features.
```
vkakerbeck:
```suggestion
This RFC proposes extending Monty's sensory processing pipeline to handle off-object observations (when a sensor has moved to a location in the environment where no features are sensed). Currently, these observations are filtered out and not passed to Learning Modules (LMs), limiting our ability to leverage prediction errors for penalizing hypotheses' evidence. By defining "null features" to represent the absence of morphological features, modifying the `FeatureChangeSM` to forward these observations, and updating the Learning Module to process off-object observations, we can enable more intelligent hypothesis management when sensors transition from on-object to off-object status. This enhancement will improve Monty's ability to recognize when it has moved off an object and reduce evidence for hypotheses that predict the sensor should still be detecting object features. 
```
I am confused by the use of "beyond an object's reference frame" here. So far we usually use this terminology when we talk about the RF of the object model inside the LM but here we are talking about the actual object in the world. Maybe the change I suggested clarifies this a bit more or feel free to reword it how you see best.

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+
+This RFC proposes extending Monty's sensory processing pipeline to handle off-object observations (when a sensor has moved beyond an object's reference frame). Currently, these observations are filtered out and not passed to Learning Modules (LMs), limiting our ability to leverage prediction errors for penalizing hypotheses' evidence. By defining "null features" to represent the absence of morphological features, modifying the `FeatureChangeSM` to forward these observations, and updating the Learning Module to process off-object observations, we can enable more intelligent hypothesis management when sensors transition from on-object to off-object status. This enhancement will improve Monty's ability to recognize when it has moved off an object and reduce evidence for hypotheses that predict the sensor should still be detecting object features. 
+
+<img src="0000_off_object_observations/off_object_observation.png" alt="Off-object observation" />
+
+_Figure 1_: Illustration of an off-object observation sensed by the Sensor Module. In this RFC, we will utilize this to penalize the evidence of hypotheses that incorrectly think they are still within the object's reference frame.
```
vkakerbeck:
```suggestion
_Figure 1_: Illustration of an off-object observation sensed by the Sensor Module (right). In this RFC, we will utilize this to penalize the evidence of hypotheses that incorrectly think they are still within the object's reference frame (left).
```

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Motivation
+
+In real-world scenarios, sensors frequently move off objects during exploration - whether transitioning between objects, reaching object boundaries, or moving through empty space. These off-object observations contain valuable information that Monty currently does not utilize. While [Question 2 of the Intelligent Resampling RFC](https://github.com/thousandbrainsproject/tbp.monty/pull/366) addresses hypotheses that have moved beyond object boundaries using path integration, this RFC focuses on a complementary approach: using sensory prediction errors to penalize hypotheses' evidence when the sensor has moved off the object.
```
vkakerbeck:
```suggestion
In real-world scenarios, sensors frequently move off objects during exploration - whether transitioning between objects, reaching object boundaries, or moving through empty space. These off-object observations contain valuable information that Monty currently does not utilize. While [Question 2 of the Intelligent Resampling RFC](https://github.com/thousandbrainsproject/tbp.monty/pull/366) addresses hypotheses that have moved beyond the internal model's boundaries using path integration, this RFC focuses on a complementary approach: using sensory prediction errors to penalize hypotheses' evidence when the sensor has moved off the object.
```

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+
+The core idea is to leverage the mismatch between predicted and observed features when sensors transition off objects. When a sensor moves off an object but active hypotheses still predict on-object features, this creates a strong prediction error signal that should penalize those incorrect hypotheses. Currently, off-object observations are filtered out by `FeatureChangeSM` and therefore not sent to the Learning Module, preventing the system from using this information.
```
vkakerbeck: "Currently, off-object observations are filtered out by `FeatureChangeSM`" 
Is this true? I thought the `DetailedLoggingSM` sets `use_state` to `False` if `on_object` is `False` and then `MontyBase` filters out the CMP messages where `use_state` is `False`?

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Current Architecture and Limitations
+
+### Overview of Current Sensory Processing Pipeline
+
+Monty's sensory processing pipeline uses `FeatureChangeSM` as a gatekeeper to determine which observations reach Learning Modules (LMs). Currently, this filtering excludes off-object observations entirely.
```
vkakerbeck: I don't think this is accurate. We also filter out off object observations when we don't use the FeatureChangeSM.

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+        center_row_col,
+        sensor_frame_data,
+        world_camera,
+    )
+else:
+    invalid_signals = True
+    morphological_features = {}  # Empty dictionary
+```
+
+This empty dictionary approach prevents meaningful comparison with predicted features, as the data structure becomes inconsistent between on-object and off-object states.
```
vkakerbeck: I'm not sure I understand. The code you copied above is not the output of `observations_to_comunication_protocol`. This is just an intermediate calculation and the empty dictionary is only used within that function. If you look at line 300 (right below that code snippet), this is turned into a CMP message (instance of State class) and the output of the function is that CMP message (`observed_state`), irrespective of whether it was on or off the object. If you want to recognize that an observation was off the object in the LM why not just check for the empty dictionary in the `morphological_features` field (or change it to None)?

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Sensor Modality Considerations
+
+The challenge of representing off-object observations varies by sensor type:
+
+- **Touch sensors**: Naturally detect "nothing" when not in contact with surface.
+- **Vision sensors**: Always detect something (even if just background), making the definition of "off-object" more complex and context-dependent.
+
+Taking this into consideration, we define "null features" as the absence of **morphological** features (surface normal and principal curvatures), since depending on the sensing modality, some **non-morphological features** may be present (e.g. background color) even when sensor is off-object.
```
vkakerbeck: Why would there be no morphological features if a vision sensor is looking at a different object in the background?

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+        # On-object to Off-object: significant change (prediction error signal)
+        logger.debug("Transition from on-object to off-object detected")
+        return True
+    elif current_on_object and not previous_on_object:
+        # Off-object to On-object: significant change (new features detected)
+        logger.debug("Transition from off-object to on-object detected")
+        return True
+    else:
+        # On-object to On-object: use existing feature comparison logic
+        return self._check_on_object_feature_changes(observed_features)
```
vkakerbeck: couldn't you just summarize the above lines into
```
if current_on_object != previous_on_object:
    return True
elif current_on_object:
   return self._check_on_object_feature_changes(observed_features)
else:
   return False
```

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+        self,
+        channel_displacement: np.ndarray,
+        channel_features: dict,
+        evidence_update_threshold: float,
+        graph_id: str,
+        possible_hypotheses: ChannelHypotheses,
+        total_hypotheses_count: int,
+    ) -> ChannelHypotheses:
+    # Check if this is an off-object obesrvation with null features
+    if_off_object = ...
```
vkakerbeck:
```suggestion
    is_off_object = ...
```

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+
+    # For off-object observation, apply evidence penalty 
+    if is_off_object:
+        evidence_penality = ... # may need to be an argument/parameter
+        new_evidence = possible_hypotheses.evidence - evidence_penalty
```
vkakerbeck: Why? This should only penalize hypotheses that predicted to be on the object after the displacement, not all of the hypotheses. If I have the correct hypothesis, it should correctly make no prediction when moving off the object (outside the model RF) and should not receive a penalty for this (if anything it should increase its evidence).

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+    channel_feature_order: list[str],
+    channel_feature_weights: dict,
+    channel_query_features: dict,
+    channel_tolerances: dict,
+    input_channel: str,  # noqa: ARG004
+) -> np.ndarray:
+    # Check if this is an off-object observation with null features 
+    is_off_object = ...
+
+    # For off-object observations, return zero evidence for all nodes
```
vkakerbeck: Why? What's the point of even processing the off object observations in that case?

rfcs/0000_off_object_observations.md
tristanls: **issue**: Inside this function, I do not see anything related to an observation that would allow us to calculate whether the observation is on or off the object. I think something like this would need to be done elsewhere, or a more significant refactor is needed.

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+def calculate(
+    channel_feature_array: np.ndarray,
+    channel_feature_order: list[str],
+    channel_feature_weights: dict,
+    channel_query_features: dict,
+    channel_tolerances: dict,
+    input_channel: str,  # noqa: ARG004
+) -> np.ndarray:
+    # Check if this is an off-object observation with null features 
+    is_off_object = ...
```
tristanls: **issue**: As with `DefaultHypothesesDisplacer`, as far as I can tell, this function does not contain observation information to determine whether things are on or off the object.

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+
+Besides adding new or updating existing unit tests for the above functions in `FeatureChangeSM`, `DefaultHypothesesDisplacer`, and `DefaultFeatureEvidenceCalculator` for correctness of implementation, the expectation for results on benchmarks are:
+
+- **Accuracy**: In best case scenarios, implementing this RFC will improve accuracy in object recognition to more quickly eliminate bad hypotheses, but it is not a guarantee. It should not decrease the performance, however.
+- **Steps to Convergence**: This is where this RFC will likely have the largest impact. By quickly removing bad hypotheses, Monty may converge faster.
+
+
+## Future Work
+
+Please see the [Open Question](https://github.com/hlee9212/tbp.monty/blob/hlee9212/intelligence_resampling_rfc/rfcs/0000_intelligent_resampling.md#when-were-learning-an-object-we-dont-have-a-complete-graph-model-how-do-we-deal-with-this) from [Previous RFC on Intelligent Resampling](https://github.com/hlee9212/tbp.monty/blob/hlee9212/intelligence_resampling_rfc/rfcs/0000_intelligent_resampling.md).
```
tristanls: **question**: Y'all have probably discussed this, but one thing comes to mind when deciding whether incorrect hypotheses should be eliminated, an incomplete model needs updating, or both. Would storing "off-object" locations on the graph itself help? This could perhaps change the problem from a difficult decision of "should nothing be there vs we've never been there?" to the maybe more tractable problem of "let's not store repeated off-object data points that seem to be going away from the object".

rfcs/0000_off_object_observations.md
hlee9212: Shoot, you are right. I was initially thinking I could somehow get it from `on_object` in morphological features in CMP and hoping it would be in `channel_features`, but from a quick debug run it's not. I'll think about it and come back to this.

rfcs/0000_off_object_observations.md
hlee9212: Ah sorry, when I was reading the docstring of `calculate()`:

```python
    def calculate(
        channel_feature_array: np.ndarray,
        channel_feature_order: list[str],
        channel_feature_weights: dict,
        channel_query_features: dict,
        channel_tolerances: dict,
        input_channel: str,  # noqa: ARG004
    ) -> np.ndarray:
        """Calculate the feature evidence for all nodes stored in a graph.

        Evidence for each feature depends on the difference between observed and stored
        features, feature weights, and distance weights.

        Evidence is a float between 0 and 1. An evidence of 1 is a perfect match, the
        larger the difference between observed and sensed features, the close to 0 goes
        the evidence. Evidence is 0 if the difference is >= the tolerance for this
        feature.

        If a node does not store a given feature, evidence will be nan.

        input_channel indicates where the sensed features are coming from and thereby
        tells this function to which features in the graph they need to be compared.

        Returns:
            The feature evidence for all nodes.
        """
```
I just assumed the evidence I should return is 0 because a off-object and on-object mistmatch should be the largest mismatch. My thought was:

  - Query: NaN/null morphological features (off-object)
  - All stored nodes: Real morphological features (on-object locations)
  - Feature difference: Always large (NaN vs real values) --> Return 0

I figured if I returned 0 then EvidenceSlopeTracker could still learn that slope is decreasing and the hypotheses that are predicting on-object incorrectly would get eliminated sooner. I may have over-thought this to the point of confusing myself. 😅

I'm not exactly sure, but did you mean returning negative evidence scores? (I think I just tried to give an example implementation without changing the definition of the method too much and more just updating to handle NaN values that may arise from off object observations).

vkakerbeck: I'm not sure it makes sense to even call this function if there are no features being sensed, but similar to my comment [here](https://github.com/thousandbrainsproject/tbp.monty/pull/425#discussion_r2276485301), if we do calculate evidence then we need to take into account the predictions that the hypotheses actually make. A correct hypothesis would correctly predict "off object" (aka no prediction) when the sensor moves off the object. This should not be penalized since it would be the correct prediction.

hlee9212: Okay! After some more thinking, how about something like below (and deleting the updates to `calculate()` and `displace_hypotheses_and_compute_evidence`)? 😄 

```python
def _calculate_evidence_for_new_locations(self, ...):
      # ... existing code through line 298 ...
          radius_evidence = (
              radius_evidence
              + hypothesis_radius_feature_evidence * self.feature_evidence_increment
          ) # line 298

      if is_off_object_observation:
          # Penalize hypotheses close to graph nodes (incorrectly predicting on-object)
          incorrect_on_object_predictions = ~mask  # mask == False means close to nodes
          radius_evidence[incorrect_on_object_predictions] = -1.0  # Configurable penalty
```

(So we already calculate whether hypotheses is within ref. frame or not in this method, and so if the input is off object but hypotheses are within ref. frame, then we can give penalty to these subsets?)

Re `is_off_object_observation`, as @tristanls pointed out, ~~we may need to update method signature to obtain this information. I was hoping for no major refactors for this RFC...but alas. Not sure if I can get into more details until I start coding, but I may have some plumbing work to do like Scott.~~

Update: We may be able to get this information by updating `_select_features_to_use` in `graph_matching.py`, possibly without large refactors:

```python
if (
      feature in self.tolerances[input_channel].keys()
      or feature == "pose_fully_defined"  
      or feature == "on_object" # Add this line
  ):
      features_to_use[input_channel][feature] = state.morphological_features[feature]
```

rfcs/0000_off_object_observations.md
hlee9212: My initial thought process was that I wanted to have at least the keys ("pose_vectors", etc.) in order to pass `all_usable_input_channels()` below:

```python
def all_usable_input_channels(
    features: dict, all_input_channels: list[str]
) -> list[str]:
    """Determine all usable input channels.

    Args:
        features: Input features.
        all_input_channels: All input channels that are stored in the graph.

    Returns:
        All input channels that are usable for matching.
    """
    # Get all usable input channels
    # NOTE: We might also want to check the confidence in the input channel
    # TODO S: Once we pull the observation class into the LM we could add this.
    return [ic for ic in features.keys() if ic in all_input_channels]
```
If this doesn't pass, then we trigger:
```python
if len(input_channels_to_use) == 0:
            logger.info(
                f"No input channels observed for {graph_id} that are stored in the "
                "model. Not updating evidence."
            )
            return []
```
Because I wanted to update hypotheses in some way even if we had null observation, I added in the keys. 

After reading your comment, I thought about switching to just `np.nan` or just `None` as the values. Then I randomly came across `_check_all_attributes()` in State class:
```python
    def _check_all_attributes(self):
        assert "pose_vectors" in self.morphological_features.keys(), (
            "pose_vectors should be in morphological_features but keys are "
        )
        # TODO S: may want to test length and angle between vectors as well
        assert self.morphological_features["pose_vectors"].shape == (
            3,
            3,
        ), "pose should be defined by three orthonormal unit vectors but pose_vectors "
    # more checks 
```
Which I think means we wouldn't pass these checks if I just set values to `np.nan` or `None`.

That said, I'm open to other ideas to represent what these "null observations" should be. Perhaps a better design in the end is to actually change or update `_check_all_attributes()`  and/or `all_usable_input_channels()` function rather than making it a complicated dictionary. 

I should also update in the RFC that the code here should be more treated as "pseudocode" or at least, for explanation or clarity. I think the idea that we need some representation for null observation is clear to the reader (or let me know if not), and I'm open to implementing it in a more principled manner while I code or during the integration project. I definitely cannot guarantee that the code will look exactly as shown in this RFC, and there will likely be other changes that I could not foresee.

rfcs/0000_off_object_observations.md
vkakerbeck: Yes, just reusing `mask` to determine how to update the evidence seems like the best way to go. I think in terms of where to do this in the code, there are better places than after the feature evidence calculation code (like as an else wrapped around line 264 (https://github.com/thousandbrainsproject/tbp.monty/blob/8c16a6393b221786a3b6357c7e71b615c4e68821/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py#L264). I also think that the whole `if self.use_features_for_matching[input_channel]:` block should be skipped in this case since we are not updating evidence based on features (since no features are sensed). 
I was also thinking we would increase the evidence for the hypotheses where mask == True (correctly predicted to be out of object's reference frame). I would actually think this may be more important than penalizing the other hypotheses (especially if the movement was small, we could get a lot of issues since there will still be points within the search radius even for a correct hypothesis so we don't want to penalize too harshly).

In terms of getting this information, I thought whether we are on or off object is directly contained in the CMP message? For example by having "morphological_features" == None.

hlee9212: Will get to replying fully, but just to answer the following question:

> In terms of getting this information, I thought whether we are on or off object is directly contained in the CMP message? For example by having "morphological_features" == None.

Yes, it is in CMP message, but `_calculate_evidence_for_new_locations(self, graph_id, input_channel, search_locations, channel_possible_poses, channel_features)` does not receive the full CMP message. 

CleanShot 2025-08-18 at 08 33 31

(Or at least, I believe this was what @tristanls was referring to in his [comment](https://github.com/thousandbrainsproject/tbp.monty/pull/425#discussion_r2276808325).)

hlee9212: Also, I believe if `morphological_features = {}` or `morphological_features = None` and `use_state = True`, then I do not think we will pass the `_check_all_attributes()` in State class:

CleanShot 2025-08-18 at 08 36 45

CleanShot 2025-08-18 at 08 37 01

This is my justification for the complex sentinel [here](https://github.com/thousandbrainsproject/tbp.monty/pull/425#discussion_r2277814053), though as noted in the comment I am open to a bigger refactor if we want to use `morphological_features = None` and refactor other places in the codebase, especially during integration phase. 🙂

vkakerbeck: Doesn't `_calculate_evidence_for_new_locations` get `channel_features` as input which would also be None (or "off_object"? It seems like it should be straightforward to get this information in that function but maybe I am missing something.

Re. the state function check, we definitely have to discuss the updates to our expectations of what a state should look like. Even if you pass nans, that doesn't change that we suddenly don't pass pose vectors which is going against a lot of Monty's assumptions. This should definitely be a discussion as part of this RFC. Maybe we talk about this more in the research meeting later.

tristanls:
**note**: @hlee9212, I assigned this PR to you as the Maintainer primary point of contact for this RFC.

codeallthethingz:
I'm going to abstain from this RFC.

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+The question that we are addressing is: How can we send and process off-object observations in Monty?
+
+The goal of the RFC is to layout current state of Monty and implementation plan to process off-object observations. 
+
+## Summary
+
+This RFC proposes extending Monty's sensory processing pipeline to handle off-object observations (when a sensor has moved beyond an object's reference frame). Currently, these observations are filtered out and not passed to Learning Modules (LMs), limiting our ability to leverage prediction errors for penalizing hypotheses' evidence. By defining "null features" to represent the absence of morphological features, modifying the `FeatureChangeSM` to forward these observations, and updating the Learning Module to process off-object observations, we can enable more intelligent hypothesis management when sensors transition from on-object to off-object status. This enhancement will improve Monty's ability to recognize when it has moved off an object and reduce evidence for hypotheses that predict the sensor should still be detecting object features. 
+
+<img src="0000_off_object_observations/off_object_observation.png" alt="Off-object observation" />
```
jeremyshoemaker: Is there a reason we're using an HTML literal for this? [Github Markdown supports image markup](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#images).

```suggestion
![Off-object observation](0000_off_object_observations/off_object_observation.png)
```

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+
+Currently, `FeatureChangeSM` filters out observations where `on_object` is False. 
+
+```python
+# Current filtering behavior in sensor_modules.py (lines 680-684)
+if not observed_features.get_on_object():
+    logger.debug(f"No new point because not on object")
+    return False
+```
+
```
jeremyshoemaker, line 44: It might make sense to include a link to the referenced code so someone reading the future can easily see it in context. If you decide to do this, make sure that the link uses the commit has for the "current" code as it exists now rather than a ref like `main` which moves over time.

```suggestion
[Source for `sensor_modules.py` lines 680-684](https://github.com/thousandbrainsproject/tbp.monty/blob/9677cc918adeca9ae21233d957c0401e84f482ab/src/tbp/monty/frameworks/models/sensor_modules.py#L680-L684)

```

rfcs/0000_off_object_observations.md
```diff
@@ -0,0 +1,237 @@
...
+        center_id,
+        center_row_col,
+        sensor_frame_data,
+        world_camera,
+    )
+else:
+    invalid_signals = True
+    morphological_features = {}  # Empty dictionary
+```
+
```
jeremyshoemaker: Same as above

```suggestion
[Source for `sensor_modules.py` lines 267-286](https://github.com/thousandbrainsproject/tbp.monty/blob/9677cc918adeca9ae21233d957c0401e84f482ab/src/tbp/monty/frameworks/models/sensor_modules.py#L267-286)
```

jeremyshoemaker:
🤦  I've been adding single comments instead of a batch for review.

rfcs/0000_off_object_observations.md
hlee9212: Added.

rfcs/0000_off_object_observations.md
hlee9212: I changed from `FeatureChangeSM` to `sensor_modules.py` in RFC. To answer your question: While this matters slightly less as we are not planning to implement this RFC anytime soon, I intended to change what was necessary (including `DetailedLoggingSM` and its subclasses) to enable off-object processing.

rfcs/0000_off_object_observations.md
hlee9212: As I am updating this RFC in the future with less focus on implementation, I have removed this code block.

rfcs/0000_off_object_observations.md
hlee9212: I have removed this section.

rfcs/0000_off_object_observations.md
hlee9212: I think your suggestion is clearer, so I have accepted it.

rfcs/0000_off_object_observations.md
hlee9212: Thanks, I was using HTML in previous RFC because I wanted to adjust width/height. Since I don't need to do that, I'll use the Markdown way.

rfcs/0000_off_object_observations.md
hlee9212: You are correct. I was using the language `FeatureChangeSM` in several places but didn't mean I was going to edit only that class, but any parent classes that might affect it (including `DetailedLoggingSM`, etc.). I have removed "by `FeatureChangeSM`", which also makes the RFC slightly less implementation-focused.

rfcs/0000_off_object_observations.md
hlee9212: Thanks. Updated the sentence.

rfcs/0000_off_object_observations.md
hlee9212: Thanks. Added.

rfcs/0000_off_object_observations.md
hlee9212: Thanks. Added.

rfcs/0000_off_object_observations.md
hlee9212: Yes, that works. I think we discussed a while ago that if we pass in an empty dict and set `use_state = True`, then it will not pass the assertions in `State._check_all_attributes()`. But, I think we decided that then we need to update the `State` class, rather than making a complex null morphological feature representation.

I think the RFC works without this section (the section was also implementation-focused), and has been removed for now.

rfcs/0000_off_object_observations.md
hlee9212: I have updated the sentence to make the definition of "null_features" more precise:

> Taking this into consideration, we treat "null features" as the absence of morphological information tied to the tracked object only when the center pixel’s semantic ID is zero (i.e. no object at that location).

rfcs/0000_off_object_observations.md
hlee9212: Yes, I was just being more explicit and writing out the four cases in code. However, since this is already written out right above the code, I have removed the code example entirely.

> In addition, the `use_state` variable will need to account for the four transitions are happening:
> 1. **On-object --> On-object**: Existing delta threshold feature comparison logic 
> 2. **On-object --> Off-object**: Treated as significant change; LM receives null features for hypothesis elimination
> 3. **Off-object --> On-object**: Treated as significant change; LM receives actual features for new processing
>4. **Off-object --> Off-object**: No change detected; observation filtered out

rfcs/0000_off_object_observations.md
hlee9212: This section as a whole has been updated. The most relevant section that addresses the above comment is:

> - **Adjust hypothesis evidence selectively**: `_calculate_evidence_for_new_locations` already builds a distance-based mask showing which hypotheses remained near stored nodes (`mask == False`) versus those that routed the sensor outside the object's reference frame (`mask == True`). During an off-object step we should penalize only the former group (since their predictions were wrong) and reward the latter group (for correctly predicting "off-object observation").

rfcs/0000_off_object_observations.md
hlee9212: This section has a whole has been updated to reflect several comments above.

rfcs/0000_off_object_observations.md
hlee9212: I will mark this as Resolved as this whole section has been updated and the code no longer exists.

rfcs/0000_off_object_observations.md
hlee9212: I will mark this as Resolved as this whole section has been updated and the code no longer exists.

rfcs/0000_off_object_observations.md
vkakerbeck: I'm not sure I understand this new sentence. Does it mean null features exist only when semantic ID is zero? What if there is no sematic ID (as in many experiments)? How can the sensor module take into account the tracked object when it has no representation of objects?

hlee9212: Hi Viviane, I simplified it to "absence of morphological information at current location". Hopefully it is clearer?

hlee9212:
Hi @nielsleadholm and @vkakerbeck, I have added the final summary of the open question/edge case regarding sensor moving just beyond object. After approval, this PR will be **closed** (not merged). 

(**Note**: If I saw that my reply got a thumbs-up emoji I marked the conversation as Resolved.)

rfcs/0000_off_object_observations.md
vkakerbeck: sounds good :)

nielsleadholm (commented):
This looks great, thanks for those updates @hlee9212 ! Happy for this to be **closed** (but in a good way haha).

hlee9212:
Got approval to close during 2025/09/29 standup meeting to close PR.

PR #61, opened 2024-11-22
Actions: Link checker

codeallthethingz:
This PR adds a check to the monty.yml.
It runs the docs link checker (tools.github_readme_sync check docs) if there are any changes to `/docs/*` and fails if there are broken links and outputs the broken links in the action output.

image

PR #429, opened 2025-08-18
chore: deploy docs on version change

codeallthethingz:
Currently the docs build is skipped when a PR for a version change comes in.  When `src/tbp/monty/__init__.py` is modified, a new docs deploy should happen.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,10 +7,10 @@ concurrency:
 on:
   push:
     branches:
-      - main2
+      - main
   pull_request:
     branches:
-      - main2
+      - main
 
 jobs:
   check_docs:


.github/workflows/docs.yml
```diff
@@ -121,7 +121,8 @@ jobs:
             echo $changed_file
             if [[ $changed_file == docs/* ]] ||
                [[ $changed_file == tools/github_readme_sync/* ]] ||
-               [[ $changed_file == .github/workflows/docs.yml ]]; then
+               [[ $changed_file == .github/workflows/docs.yml ]] ||
+               [[ $changed_file == src/tbp/monty/__init__.py ]]; then
```
tristanls: **nitpick**: This looks good to me, could you move it between docs & tools so it's alphabeticalish?

.github/workflows/docs.yml
```diff
@@ -135,7 +136,8 @@ jobs:
           while IFS= read -r changed_file
           do
             echo $changed_file
-            if [[ $changed_file == docs/* ]]; then
+            if [[ $changed_file == docs/* ]] ||
+               [[ $changed_file == src/tbp/monty/__init__.py ]]; then
```
tristanls: **issue**: I think both criteria should be the same. Should this list also include `.github/workflows/docs.yml` and `tools/github_readme_sync/*`? Did we decide not to do that for some specific reason? 🤔 Looking back at #61 I think I didn't catch that the two checks are not identical.

**question**: The `push` event happens when the PR is merged to `main`. The `pull_request` event happens while the PR is active. Any specific reason merge to `main` should have a different trigger? This would imply that merge to `main` would not run if `github_readme_sync` tool changed, or if `docs.yml` workflow changed.

codeallthethingz: I think you're right, this is an oversight and those triggers should be identical.  I'll open up a separate PR on my fork to fix that and and test that RFCs are actually getting linted and once thats working I'll merge that all in to this branch so there is one update to the vale files.

codeallthethingz:
@tristanls feel free to merge / set auto-merge if you're good with the updates.

PR #430, opened 2025-08-19
rfc: RFC 18 Guidance on Workflow for Research Code

nielsleadholm:
This RFC proposes some guidance on the workflow that TBP researchers (or community members doing research) can follow when contributing code to Monty.

### DIFF
--- a/rfcs/0018_code_guidance_for_researchers_and_community.md
+++ b/rfcs/0018_code_guidance_for_researchers_and_community.md
@@ -30,11 +30,11 @@ Below is a recommended workflow for researchers; however, it is guidance only, a
 
 When creating the fork, the owner should be set to the `thousandbrainsproject` organization:
 
-<img src="0000_code_guidance_for_researchers_and_community/create_new_Prototype_fork.png" alt="Create New Prototype Fork" width="500">
+<img src="0018_code_guidance_for_researchers_and_community/create_new_Prototype_fork.png" alt="Create New Prototype Fork" width="500">
 
 Give your fork a name like `feat.description_of_prototype` (for other examples of prefixes we use, see [RFC 10 - Conventional Commits](0010_conventional_commits.md); most Prototypes would best be described as `feat` or `perf` changes):
 
-<img src="0000_code_guidance_for_researchers_and_community/Prototype_fork_name.png" alt="Naming your Fork" width="500">
+<img src="0018_code_guidance_for_researchers_and_community/Prototype_fork_name.png" alt="Naming your Fork" width="500">
 
 If your Prototype is related to a particular [RFC](https://thousandbrainsproject.readme.io/docs/request-for-comments-rfc), it can be helpful to link to this RFC in the fork description, which will appear under the About heading.
 
@@ -62,13 +62,13 @@ If your Prototype is related to a particular [RFC](https://thousandbrainsproject
 
 The workflow is shown in the below diagram:
 
-<img src="0000_code_guidance_for_researchers_and_community/main_workflow.png" alt="Workflow with a Feature Fork" width="450">
+<img src="0018_code_guidance_for_researchers_and_community/main_workflow.png" alt="Workflow with a Feature Fork" width="450">
 
 Green arrows indicate pulling from upstream, blue arrows indicate PRs. Above the blue arrows, a suggestion is provided for whether a researcher or engineer would more suitable to review the PR. If you are contributing as a non-Maintainer, it will be the responsibility of the Maintainer you tag to determine who is most suitable to review the PR. NOTE that when implementing the IP, a researcher will pair with an engineer (i.e. it is not solely the responsibility of engineers to handle the IP process from start to finish).
 
 Below shows the potential follow-up workflow. In the example shown, the first Prototype (Prototype "A") did not demonstrate significant benefit, and the fork is archived without ever going through the IP process. A new fork is made to work on a new Prototype / feature.
 
-<img src="0000_code_guidance_for_researchers_and_community/follow_up_workflow.png" alt="Follow-up Workflow with a Second Fork" width="900">
+<img src="0018_code_guidance_for_researchers_and_community/follow_up_workflow.png" alt="Follow-up Workflow with a Second Fork" width="900">
 
 
 ### Regarding Code Quality Specified in Other Guideline
@@ -86,7 +86,7 @@ Below shows the potential follow-up workflow. In the example shown, the first Pr
 #### `dev` Branch as Intermediary
 - A slight modification to the above is to have an intermediary `dev` branch. This can serve as a useful end-point for more rapid PRs, where a single researcher might review these. Then, when the `dev` branch has reached a sufficient level of maturity, a PR to `main` can be opened, and feedback at that point sought from a broader set of the research team. This is shown diagrammatically below. TBP researchers are welcome to try this workflow to see if it provides a smoother process.
 
-<img src="0000_code_guidance_for_researchers_and_community/dev_workflow.png" alt="Using a `dev` Branch" width="450">
+<img src="0018_code_guidance_for_researchers_and_community/dev_workflow.png" alt="Using a `dev` Branch" width="450">
 
 #### `tbp.monty` as a Dependency
 - Another alternative is to have `tbp.monty` as a dependency, and simply subclass or overwrite any functions that need to be altered. See the [tbp.monty_project_template repository](https://github.com/thousandbrainsproject/tbp.monty_project_template) for an example of how to quickly set this up.
@@ -101,7 +101,7 @@ Below shows the potential follow-up workflow. In the example shown, the first Pr
 
 This approach is shown diagrammatically below:
 
-<img src="0000_code_guidance_for_researchers_and_community/single_fork_workflow.png" alt="Workflow with a Single Fork" width="900">
+<img src="0018_code_guidance_for_researchers_and_community/single_fork_workflow.png" alt="Workflow with a Single Fork" width="900">
 
 
 ### Discontinued Prototypes


rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Regarding Code Quality Specified in Other Guideline
+- When writing code for the RP, you may choose to follow the [official guidance for the platform](https://thousandbrainsproject.readme.io/docs/code-style-guide). 
+- The advantage of doing so while working on the RP is that it can reduce the amount of work required for the IP stage.
+- However, adhering to these guidelines is often not possible without first carrying out a significant refactor of the code, or through a time-consuming design process.
+- As such, researchers should not follow these guidelines if it will introduce significant delays. Rather, the guidance they provide (e.g., being mindful of how state is added or accessed in code) can provide general tips for writing high-quality code from the beginning.
+
+### Other Comments
+- Your new fork will not have any controls in place to prevent force pushes from other [Members of the TBP organization](https://github.com/orgs/thousandbrainsproject/people). We therefore ask that everyone exercise care in working with the forks of other team members.
```
nielsleadholm: @tristanls 

**note:** This is my understanding at least, but please feel free to correct me.

rfcs/0018_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,102 @@
...
+- Note that the IP process will often involve a series of atomic PRs as well, rather than a single major PR. However, each one will be constrained to follow the platform guidelines.
+
+### Regarding Code Quality Specified in Other Guideline
+- When writing code for the RP, you may choose to follow the [official guidance for the platform](https://thousandbrainsproject.readme.io/docs/code-style-guide). 
+- The advantage of doing so while working on the RP is that it can reduce the amount of work required for the IP stage.
+- However, adhering to these guidelines is often not possible without first carrying out a significant refactor of the code, or through a time-consuming design process.
+- As such, researchers should not follow these guidelines if it will introduce significant delays. Rather, the guidance they provide (e.g., being mindful of how state is added or accessed in code) can provide general tips for writing high-quality code from the beginning.
+
+### Other Comments
```
ramyamounir, line 80: One other thing I noticed is that the settings of the `tbp.monty` repository such as protected branches and commit default messages are not copied when we create a fork. For example, I had to go to settings -> general, and choose this as the default message to be consistent with what we do in `tbp.monty`

Screenshot 2025-08-19 at 6 49 06 AM

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,102 @@
...
+- The next step for any major code change is to then propose your idea in the form of [the RFC process](https://thousandbrainsproject.readme.io/docs/request-for-comments-rfc).
+
+Following the above guidance will ensure you do not spend a large amount of time on a code implementation that is never utilized.
+
+### Workflow for Contributing Code
+
+If you are a member of the community working on a research idea, then we would recommend you also follow the above guidance under [Workflow for TBP Researchers: from Research Prototype to Implementation Project](#workflow-for-tbp-researchers-from-research-prototype-to-implementation-project). The main additional points to highlight are:
+- When opening the fork, you should set yourself as the owner (it will not be possible to set the Thousand Brains Project as the owner).
```
ramyamounir: I think the process would be a bit different for community members because of the limitations (imposed by GitHub) on the number of personal forks they can have. Personal accounts can only have a single personal fork, therefore cannot work on multiple features simultaneously. Another disadvantage is that we also won't be able to archive useful code in our organization repo (if we decide not to merge the proposed feature now). This was partially possible before with closed PRs in `tbp.monty`.

Three alternatives I can think of:
- Community members can work from their personal forks similar to how @scottcanoe has been implementing his RFC, or as described in the alternative workflows section of this PR.
- When the community member reaches out to us through discourse, or other means, we create the feature fork for them and add them as a collaborator. We can add additional detail about how this would be set up in @codeallthethingz's Future Works RFC. I recommend this approach (despite it not being very scalable) because we can track the features being implemented, and perhaps add multiple collaborators to a single feature fork. It will also encourage collaborations between TBP and community researchers working on the same feature simultaneously.
- We allow regular PRs into `tbp.monty` from community members to make it easier for them to contribute code. Note that this is the widely accepted approach and many members will still be submitting PRs directly into `tbp.monty` despite all the guidelines we are trying to enforce.

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,102 @@
...
+- Please tag a Maintainer of the TBP when making atomic PRs into your feature branch. This will help get involvement from the TBP research team at an early stage of your work.
```
ramyamounir: How would they choose which maintainer is the best fit for reviewing the PR? I think we should control that. I would suggest we have a system where the community members can request a review from a maintainer and we can decide to assign a fitting TBP member to review the PR.

rfcs/0000_code_guidance_for_researchers_and_community.md
ramyamounir:
```suggestion
- When creating the fork, you should set yourself as the owner (it will not be possible to set the Thousand Brains Project as the owner).
```

rfcs/0000_code_guidance_for_researchers_and_community.md
tristanls: @nielsleadholm this matches my understanding as well 👍.

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,102 @@
...
+
+2. Clone the new fork, and then create a new development branch that is appropriately named.
+    - Your fork should now have a `main` branch, and a feature branch or branches where you will carry out active work.
+
+### Workflow
+- The aim is to progressively implement the Research Prototype via a series of PRs into the `main` branch of your RP fork.
+    - You should formulate PRs into the main branch as _atomic_ PRs for review by other researchers.
+    - Note that _atomic_ does not always mean small, although it is a common finding that PRs (especially by researchers!) are too large. Rather, atomic means that the PR represents a full, minimally self-contained change. The PR should not depend on code that will be added in the future, although context can be provided where helpful re. future changes.
+- When you open a PR into your fork's `main` branch, tag at least one other TBP researcher for review.
+    - You should NOT merge PR changes into the Research Prototype `main` branch without receiving a review from another researcher.
```
tristanls: **issue (non-blocking)**: Working via PRs may be helpful to provide structure; I'm not objecting to that. I worry that requiring a review from another researcher may lead to unnecessary delays. All code will ultimately be reviewed when merged to `tbp.monty`, so I think any review before that should have a very specific purpose instead of being the default, since the default, as described here, implies delay.

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,102 @@
...
+
+### PRs without RFCs
+- Researchers are reminded that not all research work requires an [RFC](https://thousandbrainsproject.readme.io/docs/request-for-comments-rfc) before it can be carried out. For example, a significant idea might require a single line of code to evaluate, in which case it is quicker to simply make the change and report the results, rather than begin with the RFC process.
+
+### Multiple Researchers Working on a Single Research Prototype
+- From time-to-time, two or more researchers may be working on ideas that could be considered elements of a single, over-arching Research Prototype. In this case, it is perfectly acceptable for them to use a shared RP fork.
+- As with typical Research Prototypes, this work should still regularly be synced with the upstream `tbp.monty`. The main difference is that the collaborating researchers can submit PRs into this shared fork.
```
tristanls: **suggestion (non-blocking)**: If two researchers are collaborating, they may also do pair programming, which can make a PR review extraneous. Perhaps highlighting that pair programming is an option and elaborating on that mode of cooperation would be useful here?

rfcs/0000_code_guidance_for_researchers_and_community.md
tristanls: **issue**: I don't recommend maintaining forks on behalf of external contributors. My concern is not scalability but security, as well as financial and reputational risk associated with hosting code as part of the thousandbrainsproject organization. **note**: I might not object if this were done for an external Maintainer, but I think we could revisit that decision once that Maintainer needs it.

rfcs/0018_code_guidance_for_researchers_and_community.md
nielsleadholm: Right. Re. protection, do you feel like this comment is insufficient? These forks are not like the main platform, where others are pulling from them and an accidental commit could be highly damaging. In the unlikely case that a commit ever happened by accident, it could always be reverted / the changes undone. 

Re. squash merging, that is mainly for keeping the history in `tbp.monty` concise, and any PRs that make their way into it will be squashed anyways, so again I'm not sure it's an issue? Feel free to say though if you're concerned.

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: Thanks Ramy for pointing that out, I hadn't considered the issue of the number of personal forks. I think my preference would be for option 1 of those three, so I could add a bit more detail on that. 3 isn't really a solution, but more as you say something that might happen time to time.

Before I update the wording and add details for how to do option 1, can I just check @tristanls , was your comment/issue re. option 2 that Ramy brought up, or re. the existing wording in the RFC?

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: Yeah exactly my thinking was they tag a maintainer, and then we internally decide on the best fit. I'll make that more clear/explicit.

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: My bias would be to try this way to begin with. Overall we are not putting in much structure into this workflow. Asking that PRs are atomic and are reviewed is pretty much the only structure, and I am concerned that without this, the process will become too uncontrolled. The undesirable default in research is often that someone works on a "PR" for ages, which turns into an enormous collection of code. When it's finally reviewed, it becomes clear that there was a misunderstanding or false direction that could have been avoided with earlier intervention / reviews. 

So as said my bias is to have this guidance for now. If we find it is slowing things down like you say, then we can revisit. Notably the guidance is for review by a researcher, not necessarily a manager. This was helpful during the DMC where Scott and Hojae would often review one-another's PRs related to the configs and visualization code. 

Curious for other's views including @vkakerbeck

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: Good point, I've added some text on this.

nielsleadholm:
@vkakerbeck there are a couple ongoing discussions, but I have addressed all comments with straightforward responses, so this RFC is now ready for your review.

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: I can potentially add some of the above reasoning as justification for this advice, but will wait for any follow-up discussion.

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,104 @@
...
+# Motivation
+
+Research benefits from rapidly exploring new ideas and features. At the same time, we want to ensure that Monty is a stable platform built on high-quality code. In [RFC 14](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0014_conducting_research_while_building_a_stable_platform.md), we established a workflow for integrating code into Monty that balances these two objectives. The intent of this RFC is to provide additional, concrete guidelines to realize this new workflow and ensure it is successful.
+
+When this RFC is merged, the intent is to transfer most of its contents into an updated page (or pages) in our documentation.
+
+# Definitions
+
+- *Research Prototype (RP)*: A rapidly implemented feature intended to evaluate a research idea. This can establish or refute the hypothesized benefits of the idea, as well as enable measuring other qualities associated with it.
+- *Implementation Project (IP)*: The process of refactoring a Research Prototype (as well as potentially `tbp.monty` itself) in order to merge a useful RP into the platform code-base (`thousandbrainsproject/tbp.monty`).
```
vkakerbeck:
```suggestion
- *Implementation Project (IP)*: The process of refactoring a Research Prototype (as well as potentially `tbp.monty` itself) in order to merge a feature that was proven useful by an RP into the platform code-base (`thousandbrainsproject/tbp.monty`).
```

rfcs/0000_code_guidance_for_researchers_and_community.md
vkakerbeck: Maybe we could have a middle ground where potentially controversial changes (mainly things that conceptually change how Monty works) should be reviewed or at least discussed as early as possible. 
But I also agree that we don't want to slow research prototypes down by requiring a review on every atomic change (which usually takes more than 24 hours). 

It's hard to suggest a concrete guideline for this as I think it is something we need to weigh on a case-by-case basis. Maybe just something like "RP work should be regularly reviewed and discussed with other researchers. Researchers can open PRs from branches on the feature fork to the feature main branch to receive early feedback on the execution of their idea. The start of an implementation project (PR to tbp.monty) should not be the first time a new feature is reviewed by other researchers. However, the researcher should use their own judgement to weigh the benefits of reviews on every atomic change against the benefit of quick iteration and experimentation."

rfcs/0018_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,104 @@
...
+- Your new fork will not have any controls in place to prevent force pushes from other [Members of the TBP organization](https://github.com/orgs/thousandbrainsproject/people). We therefore ask that everyone exercise care in working with the forks of other team members.
+- Similarly, GitHub Workflows (such as unit tests and linting) will not automatically run in your new fork. It is recommended that you regularly run linting and unit-tests in your local environment when developing. 
+
+### Alternative Workflows
+- We discussed using branches within existing personal forks as a way of managing Research Prototypes. However, this risks becoming overwhelming with the number of features. It also does not provide the same visibility to the community about existing prototypes that are reaching a mature stage of development.
+
```
vkakerbeck, line 85: One other alternative that I really liked was to use the `monty_project_template` https://github.com/thousandbrainsproject/tbp.monty_project_template to start a project that just uses Monty as a dependency. I enjoyed that workflow for our ultrasound project as it clearly isolated all the changes made in custom classes and it was clear to see just from the files in the repo what the customizations were (instead of requiring a diff view). It was also easy to pin to monty versions and decide when to upgrade the version and make related changes. Is there anything that speaks against that workflow?

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,104 @@
...
+- The next step for any major code change is to then propose your idea in the form of [the RFC process](https://thousandbrainsproject.readme.io/docs/request-for-comments-rfc).
+
+Following the above guidance will ensure you do not spend a large amount of time on a code implementation that is never utilized.
+
+### Workflow for Contributing Code
+
+If you are a member of the community working on a research idea, then we would recommend you also follow the above guidance under [Workflow for TBP Researchers: from Research Prototype to Implementation Project](#workflow-for-tbp-researchers-from-research-prototype-to-implementation-project). The main additional points to highlight are:
+- When creating the fork, you should set yourself as the owner (it will not be possible to set the Thousand Brains Project as the owner).
+- Please tag a Maintainer of the TBP when making atomic PRs into your feature branch. This will help get involvement from the TBP research team at an early stage of your work. Once you have tagged a Maintainer, we will triage and determine the most appropriate member of the team to review the PR.
```
vkakerbeck: This could just be a recommendation. I am not sure we need to require review of every atomic PR

vkakerbeck (commented):
Nice, thanks for writing out these guidelines. My main two high-level comments on content I was missing here:
- What about the monty-as-dependency approach (see more details in my comment on this)
- Will our cloud compute infrastructure work with this approach (i.e. can I run benchmark experiments from this feature fork the same way I do with my tbp.monty fork or do I need to make any updates to the cli tool?

rfcs/0018_code_guidance_for_researchers_and_community.md
nielsleadholm: Yeah good question. I think it kind of depends on the nature of the change. If you are creating an entirely new class / new child etc., then it's true that this can be cleaner. However I feel like if you are largely overwriting existing modules/code, then the diff is actually harder to see - the PR will show everything as new, even when many lines of code are the same as in the original. With the forking approach, the researcher has the flexibility to modify existing code directly, or subclass where appropriate, so I feel like it can enable both options.

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: Fair point, changed

nielsleadholm:
Thanks @vkakerbeck , good question about AWS infrastructure, I will check.

Re. the monty-as-dependency, I've written some thoughts, let me know what you think.

The other comments I have changed the text to address.

rfcs/0018_code_guidance_for_researchers_and_community.md
ramyamounir: I see what you mean. Makes sense, there is no reason for the feature fork to follow the same settings as `tbp.monty`. I choose to match the default commit message for consistency. 

My comment was more of a general observation that the fork doesn't automatically copy the same settings of the upstream repo, so the maintainers may have to do that manually if they wish to keep it consistent (especially for things like preventing a `main` commit push without a PR).

rfcs/0000_code_guidance_for_researchers_and_community.md
ramyamounir: Ok yeah I misunderstood then. One potential issue to maybe watch out for is that community members don't have the option to tag or request a review from maintainers unless they add this maintainer as a collaborator on their personal fork. Alternatively, they can reach out to us (e.g., through discourse) with a link to their PR for review.

Screenshot 2025-08-21 at 6 31 51 AM

rfcs/0018_code_guidance_for_researchers_and_community.md
vkakerbeck: Yes I agree it depends on the type of change. But maybe it would still be worth to list as an alternative workflow for changes that won't affect existing Monty infrastructure. For example implementing the 2DSurfaceSM would be a project that I would personally prefer that approach for as it would just involve creating a new SM class and wouldn't touch any of Monty's existing algorithm.

vkakerbeck:
Thanks for the updates! I left a comment on the monty-as-dependency question. The changes all look good to me :) Re. the AWS infrastructure, we don't need to add this info into this RFC but I think it is an important consideration for deciding our default setup.

rfcs/0000_code_guidance_for_researchers_and_community.md
tristanls: @nielsleadholm, I was responding to Ramy's option 2.

rfcs/0000_code_guidance_for_researchers_and_community.md
tristanls: While the markdown tag/mention autocomplete might not work, I believe you can still tag/mention people via @ramyamounir. https://github.blog/news-insights/mention-somebody-they-re-notified/

No need for Disourse or elsewhere, an `@` in the markdown description or comment should do it.

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Setup
+1. Begin a new Research Prototype by creating a fork of [tbp.monty](https://github.com/thousandbrainsproject/tbp.monty/):
+
+
+When creating the fork, the owner should be set to the `thousandbrainsproject` organization:
+
+<img src="0000_code_guidance_for_researchers_and_community/create_new_RP_fork.png" alt="Create New Research Prototype Fork" width="500">
+
+Give your fork a name like `feat.description_of_prototype`:
```
hlee9212: **question/suggestion:** I remember there was some discussion about whether to put RFC number instead of `feat` or something similar. I'll tag @ramyamounir because I think he may have brought it up and may remember better. :) I am agnostic/fine with keeping the fork name as `feat.description_of_prototype`. However, I think it may be useful to specify in the description of the project and/or URL to link to an RFC the feature is related for organization. This should be specified as Optional since not all RP may have an associated RFC.

rfcs/0000_code_guidance_for_researchers_and_community.md
hlee9212: **question/suggestion:** Would it be beneficial to add something like other Members of the TBP organization should set `git remote set-url upstream --push "readonly"` (change `upstream` to whatever name you give to each other's fork), especially when collaborating?

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Non-Prototype PRs from Researchers
+- Researchers may also open PRs directly into `thousandbrainsproject/tbp.monty`, such as a refactor that is useful for them or other researchers, and which is not specific to their current feature project.
+- This will often be desirable if it constrains the complexity of a Research Prototype, and improves the quality of the platform. 
+- However, such PRs should be of general benefit if they are to be opened in this way. In other words, even if a Research Prototype is ultimately abandoned and archived, PRs directly into `tbp.monty` should still have been useful on their own merits.
+- When contemplating such PRs, it is important to consider when a change is suitable for a researcher to work on, vs. when it would benefit from an engineer's expertise. For example, major refactors, or features that are primarily about infrastructure tooling, are likely to be a poorer fit for a researcher's focus.
```
hlee9212: **suggestion (non-blocking):** While I understand the intent and that it is for TBP researchers only (note: you may wish to specify as TBP Researchers in heading title instead of just Researchers), I think this bullet point in particular may unintentionally feel discouraging to community members. My suggestion is to just remove this last bullet point.

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,104 @@
...
+    - However, researchers should use their own judgement to weigh the benefits of reviews on every atomic change against the benefit of quick iteration and experimentation.
+- During your work, ensure you regularly update your fork and its branches with any upstream changes to `thousandbrainsproject/tbp.monty`. This will reduce the chance that a longer RP-project is associated with significant merge conflicts when you later wish to merge it into the primary Monty code-base.
+- When a research prototype has been fully implemented and its merits established through evaluations and visualizations, the `main` branch is ready to form the basis of an Implementation Project.
+- The Implementation Project (IP) involves any necessary refactoring of the Research Prototype to ensure it adheres to the [guidelines of the platform](https://thousandbrainsproject.readme.io/docs/code-style-guide) before merging into `tbp.monty`. For more information, see [RFC 14 Conducting Research While Building a Stable Platform](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0014_conducting_research_while_building_a_stable_platform.md).
```
tristanls: **suggestion (non-blocking)**: I think it'd be helpful to link directly to [the Implementation Project section](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0014_conducting_research_while_building_a_stable_platform.md#implementation-project) here.

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,104 @@
...
+- Note that the IP process will often involve a series of atomic PRs as well, rather than a single major PR. However, each one will be constrained to follow the platform guidelines.
```
tristanls: **suggestion**: I think this might be better as a direct quote from RFC 14:

> The implementation project is a project because it may require major refactoring of the prototype or of Monty in order to integrate the new capability. It may be a single pull request, or it may be a long running effort as we reorganize all of Monty. The working prototype is helpful as it becomes a working specification of what needs to be implemented.
>
> This is the stage at which maintaining platform stability comes into consideration, not before.
>
> -- [RFC 14](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0014_conducting_research_while_building_a_stable_platform.md#implementation-project)

I think the summary here is overly restrictive by highlighting a series of atomic PRs.

Also, the note from RFC 14 may also be helpful:

> Please note that the decision as to when a prototype capability is stable, what constitutes a merge-worthy capability, and when to initiate the implementation project is at the discretion of the researcher.
>
> -- [RFC 14](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0014_conducting_research_while_building_a_stable_platform.md#rapid-prototyping-followed-by-platform-integration)

nielsleadholm:
Thanks for the feedback all, I made several updates which at a high level introduce more flexibility in the guidance. As part of this, I've added some diagrams to give a bit more clarity about the different workflows that can be used. It should be ready for review again. 

@hlee9212 re. your question about the IP process, I added a bit more detail on this, although I'm hesitant to add too much, as I feel like it will expand the scope significantly. It's also something that will come a lot down to @tristanls 's preferences, so is difficult for me to write out, and I think we can revisit this when we have an actual IP to go through. Hope that makes sense.

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: FYI I have added some clearer guidance for the option (1) that would work if someone does not have the ability to make multiple forks.

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: This has been clarified in the text now.

rfcs/0018_code_guidance_for_researchers_and_community.md
nielsleadholm: Ah sorry, yes definitely worth mentioning it. I've added more discussion of alternative workflows, and tried to be less prescriptive in which one to follow (rather, just highlighting pros and cons).

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: Ok yeah I've added about the description, I don't recall either the specific guidance on RFC numbers for PRs, but agree it could be interesting. Will see what Ramy says.

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: Ok thanks, the heading should already have TBP in it but I've added TBP in the text itself as well. I still think it is a useful thing to highlight and don't see it discouraging a community member.

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -36,20 +36,41 @@ Give your fork a name like `feat.description_of_prototype`:
...
+- As a brief summary of the IP process, please see the below extract:
+
+> The implementation project is a project because it may require major refactoring of the prototype or of Monty in order to integrate the new capability. It may be a single pull request, or it may be a long running effort as we reorganize all of Monty. The working prototype is helpful as it becomes a working specification of what needs to be implemented.
+>
+> This is the stage at which maintaining platform stability comes into consideration, not before.
+>
+> -- [RFC 14](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0014_conducting_research_while_building_a_stable_platform.md)
+
+- When you believe the RP is ready for the IP process, please tag (`@`) or otherwise discuss with the TBP lead engineer, [Tristan Slominski](https://github.com/tristanls).
```
nielsleadholm: **thought:** @tristanls In order to provide some more concrete guidance on how to kick-off the IP process, I thought this would be helpful. However, one issue I've noticed is that we have organizational structure within the TBP that doesn't clearly map onto the structure in the open-source world. For example, only @vkakerbeck has a different distinction as the Lead Maintainer. I appreciate that, as we move towards more and more community involvement, external concepts such as Maintainer will become more important. I wonder if to navigate this, we need more fine-grained designations, or we just leave it as is for now?

rfcs/0000_code_guidance_for_researchers_and_community.md
ramyamounir: If I remember correctly, the discussion was about adding metadata to the repository to provide information that relates a feature fork to a specific RFC, or just describe what feature this fork is trying to implement. And we ended up agreeing that we can put this information in the about section of the repo on GitHub. 

I think `feat.description_of_prototype` is reasonable, but maybe worth noting if there are other options available (and use cases for each) or if `feat` is the only prefix we should use for RPs.

ramyamounir (approved):
This looks great, thank you for adding in those changes!

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: @tristanls per our discussion just now, I have removed this line.

@hlee9212 I spoke to Tristan about this and I think we can simplify things by not adding too much guidance on IPs at the moment. It should be pretty straightforward for internal researchers (just discuss with the team when you think it's ready, we'll take it from there). I added a brief line at the end for community members advising they can tag a maintainer.

tristanls: I just remembered we have an example of how this played out in the community. Collin [shared data on how GPU can accelerate computation](https://thousandbrains.discourse.group/t/preliminary-gpu-acceleration-data-indicates-significant-speedup-of-hypothesis-update-operations/750). In there, we talked about next steps. I created a [separate topic where we discussed the prototype and what to do next](https://thousandbrains.discourse.group/t/prototype-hypothesesupdater-operations-on-cuda-gpus/761). So, the approach of discussing it with the team and figuring out what to do next also seems to translate to community interactions, at least in this case.

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,146 @@
...
+- When you open a PR into your fork's `main` branch, it is generally recommended that you tag at least one other TBP researcher for review.
+    - The aim is that RP work should be regularly reviewed and discussed with other researchers; getting reviews early, particularly for complex or conceptually important changes, reduces the chance that significant effort will be spent on a change that might be in a suboptimal direction.
+    - However, researchers should use their own judgement to weigh the benefits of reviews on every atomic change against the benefit of quick iteration and experimentation.
+- During your work, ensure you regularly update your fork and its branches with any upstream changes to `thousandbrainsproject/tbp.monty`. This will reduce the chance that a longer RP-project is associated with significant merge conflicts when you later wish to merge it into the primary Monty code-base.
```
tristanls: **issue (non-blocking)**: I'm not sure this will be helpful. What comes to mind is something like a full refactor of some functionality that landed in `tbp.monty`, which will cause problems when attempting to merge into a fork that's in progress and relied on previous functionality. I understand that this is trying to avoid merge conflicts later, but dealing with merge conflicts is something I recommend deferring to the Implementation Project, and not something to worry about while getting a prototype to work. Resolving merge conflicts is wasted work if the prototype is discarded.

rfcs/0018_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,146 @@
...
+> The implementation project is a project because it may require major refactoring of the prototype or of Monty in order to integrate the new capability. It may be a single pull request, or it may be a long running effort as we reorganize all of Monty. The working prototype is helpful as it becomes a working specification of what needs to be implemented.
+>
+> This is the stage at which maintaining platform stability comes into consideration, not before.
+>
+> -- [RFC 14](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0014_conducting_research_while_building_a_stable_platform.md)
+
+The workflow is shown in the below diagram:
+
+<img src="0000_code_guidance_for_researchers_and_community/main_workflow.png" alt="Workflow with a Feature Fork" width="450">
```
tristanls: **praise**: Very nice graphic, summarizes the process well.

**issue**: This and the following graphics are missing the expected Researcher involvement in the Implementation Project. As such, I worry that it sets the wrong expectations.

>The general approach going forward is to enable rapid research prototypes without imposing platform stability requirements.
>
> 1. Researchers will create prototypes on their Monty forks and iterate rapidly there.
> 2. Once the prototypes demonstrate stable capability, **researchers will pair with engineers on an implementation project.**
> 3. The implementation project results in merging of the capability to Monty main.
>
> -- [RFC 14](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0014_conducting_research_while_building_a_stable_platform.md#rapid-prototyping-followed-by-platform-integration)

tristanls (commented):
I re-read the entire RFC.

One thing that strikes me about the RFC is that it is mainly about TBP Researchers. I struggle to find anything that a community member would find actionable. I think this is OK. 

**suggestion**: Make this an RFC for TBP Researchers only. The last two paragraphs mentioning the community could be folded into the existing documentation. I think this would be clearer for readers. The curious community members could still read through to see how we organize our research and extract what they want from it. But, more importantly, community members wouldn't feel that they have to do anything suggested here to contribute to research. Instead, they'd follow the normal GitHub norms (as much as there are any 🙂)

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: Hmmm yeah I would maybe suggest a middle-of-the-road approach. The reality is most upstream changes are fine, but if left for long (and RPs do sometimes drag on), they can accumulate to a bit of a nightmare. I might soften the language to suggest that it is advised to regularly pull from upstream, but if this ever introduces breaking changes, then rather than try to resolve them the, the recommendation is to revert to the older commit and deal with it later during the IP.

rfcs/0018_code_guidance_for_researchers_and_community.md
nielsleadholm: Ok, the graphic is meant to show who is reviewing. I've added some description trying to make this clearer, let me know if you would like it changed further / still isn't clear.

nielsleadholm:
Thanks for the re-reviews. @tristanls re. your high-level comment, my personal preference would be to leave some discussion of community guidance in this RFC. As you say, it's light touch, but my hope is that it will encourage others to feel like they can contribute to the research-side of things, i.e. that we don't assume that only TBP researchers will be doing this. 

I agree with "The last two paragraphs mentioning the community could be folded into the existing documentation.", this is what I was getting at (maybe not very clearly) with "...the aim will be to make this guidance a bit clearer, including the motivation)" --> i.e. that I will update that documentation once this RFC has merged, rather than creating a new document. How does that sound?

rfcs/0000_code_guidance_for_researchers_and_community.md
tristanls: @nielsleadholm, that sounds good 👍

tristanls:
> but my hope is that it will encourage others to feel like they can contribute to the research-side of things, i.e. that we don't assume that only TBP researchers will be doing this.

@nielsleadholm, I honestly don't think that this RFC is a good mechanism to encourage outside researchers. I think there ought to be better ways of doing it. Perhaps we could get an outside researcher to read this RFC and see if they are encouraged? I guess that this RFC will have the opposite effect by highlighting a lot of complexity that outside researchers don't need to deal with or care about.

rfcs/0018_code_guidance_for_researchers_and_community.md
tristanls: Without the graphic itself changing, I believe the issue remains. 

**suggestion**:
Screenshot 2025-08-28 at 11 40 52

vkakerbeck: I agree, the graphic should at least say "reviewed by ..." to make this clear.

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,147 @@
...
+>
+> -- [RFC 14](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0014_conducting_research_while_building_a_stable_platform.md)
+
+The workflow is shown in the below diagram:
+
+<img src="0000_code_guidance_for_researchers_and_community/main_workflow.png" alt="Workflow with a Feature Fork" width="450">
+
+Green arrows indicate pulling from upstream, blue arrows indicate PRs. Above the blue arrows, a suggestion is provided for whether a researcher or engineer would more suitable to review the PR. If you are contributing as a non-Maintainer, it will be the responsibility of the Maintainer you tag to determine who is most suitable to review the PR. NOTE that when implementing the IP, a researcher will pair with an engineer (i.e. it is not solely the responsibility of engineers to handle the IP process from start to finish).
+
+Below shows the potential follow-up workflow. In the example shown, the first RP (RP "A") did not demonstrate significant benefit, and the fork is archived without ever going through the IP process. A new fork is made to work on a new RP / feature.
```
vkakerbeck: This is maybe a personal issue so don't feel like you need to make a change if noone else notices that but I keep reading PR instead of RP. Could we just call it a prototype instead? That would make it a bit more encompassing in general (also including engineering prototypes etc.) and intuitive to understand (no abbreviation).

rfcs/0018_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,147 @@
...
+
+# Summary
+
+This RFC proposes some guidance on the workflow that TBP researchers (or community members doing research) should follow when contributing code to Monty.
+
+# Motivation
+
+Research benefits from rapidly exploring new ideas and features. At the same time, we want to ensure that Monty is a stable platform built on high-quality code. In [RFC 14](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0014_conducting_research_while_building_a_stable_platform.md), we established a workflow for integrating code into Monty that balances these two objectives. The intent of this RFC is to provide additional, concrete guidelines to realize this new workflow and ensure it is successful.
+
+When this RFC is merged, the intent is to transfer most of its contents into an updated page (or pages) in our documentation.
```
vkakerbeck, line 12: A lot of this only applies to us (like making a fork under the TBP org). The public docs may not be the best place for this. Maybe we just update the ways to contribute page https://thousandbrainsproject.readme.io/docs/contributing with a short form of the last section of this and then link to this RFC.

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Alternative Workflows
+
+#### `dev` Branch as Intermediary
+- A slight modification to the above is to have an intermediary `dev` branch. This can serve as a useful end-point for more rapid PRs, where a single researcher might review these. Then, when the `dev` branch has reached a sufficient level of maturity, a PR to `main` can be opened, and feedback at that point sought from a broader set of the research team. This is shown diagrammatically below. TBP researchers are welcome to try this workflow to see if it provides a smoother process.
+
+<img src="0000_code_guidance_for_researchers_and_community/dev_workflow.png" alt="Using a `dev` Branch" width="450">
+
+#### `tbp.monty` as a Dependency
+- Another alternative is to have `tbp.monty` as a dependency, and simply subclass or overwrite any functions that need to be altered.
```
vkakerbeck: It could be useful to link our template repo for this here https://github.com/thousandbrainsproject/tbp.monty_project_template

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,147 @@
...
+- This can clearly demarcate what parts of the code have changed, and might be particularly suitable when creating a new child-class, or an entirely new function. Examples of this might be implementing a new data-loader for a particular dataset, or a new type of sensor module.
+- This approach is less well suited when significant parts of the existing code need to be altered in-place. While still doable, it can be challenging with this approach and a standard `diff` to see what is genuinely new.
+
+#### Feature Branches (Not Feature Forks)
+- Another possibility is using branches within existing personal forks as a way of managing Research Prototypes.
+- In this case, a new pair of branches is made in your personal fork whenever starting work on a Research Prototype.
+- We have decided that for internal use-cases, this risks becoming overwhelming with the number of RPs/features that we explore, so is generally best avoided.
+- It also does not provide the same visibility to the community about existing prototypes that are reaching a mature stage of development, as these forks are often private.
```
vkakerbeck: Why would those be private? Aren't our forks public by default?

rfcs/0000_code_guidance_for_researchers_and_community.md
```diff
@@ -0,0 +1,147 @@
...
+- However, if you are a member of the community doing research and you are unable to create multiple forks on your profile, this can be a good approach to adopt.
+
+This approach is shown diagrammatically below:
+
+<img src="0000_code_guidance_for_researchers_and_community/single_fork_workflow.png" alt="Workflow with a Single Fork" width="900">
+
+
+### Discontinued Research Prototypes
+- Sometimes a research idea will not bear fruit. This is fine, indeed we expect this to happen from time to time, and it is the primary reason we expect the proposed workflow to be faster than merging all changes into `tbp.monty`. When an RP does not demonstrate a fundamental improvement, you should archive the fork in question.
```
vkakerbeck: It might be worth adding a note on updating the README or other documentation on why it did not work un case we ever get back to it.

vkakerbeck (commented):
Nice, thanks for the updates! Looks great :) I left a few more small comments.
Did you have a chance to check whether our AWS setup works out of the box with the RP forks?

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: Oh you're right, not sure why I thought they would be private. Must have been disassociating to Numenta days...

rfcs/0000_code_guidance_for_researchers_and_community.md
nielsleadholm: Good point, have updated that.

rfcs/0018_code_guidance_for_researchers_and_community.md
nielsleadholm: Ok thanks for the feedback, I've made these changes.

nielsleadholm:
Thanks @tristanls and @vkakerbeck for your comments. Most of these should be addressed now. 

Re. your point Tristan that the community guidance risks introducing undesirable overhead - I agree asking researchers in the community is a good suggestion. I will post in Discourse for some feedback. Once we have a decision on that point, I expect that this RFC will be ready for the Final Comment Period.

Re. your question Viviane about AWS, yes this is possible now with the `tbp.cli` tool, copied below the guidance from Tristan. I tested this with `feat.compositional_testbed` - while the model wasn't available, the correct code (and therefore config) was accessed. Syncing models and datasets easily is a broader question that I'm currently digging into. 

> Running on AWS Infrastructure from a tbp.monty fork can be done by specifying --monty-dir flag pointing to the fork directory. It will be rsync'd and the code will be used as is on the fork. One caveat is that the conda environment will be from tbp.monty main, so any changes in environment.yml or pyproject.toml will not be available and would require a different approach (creating a PR to tbp.monty main, having an AWS instance image pre-built, and then using it with --pr flag).

nielsleadholm:
@tristanls it's now been nearly a week and the one feedback we have had from a community member was in support of this guidance (along with a good suggestion about how we should present it when moving it to the actual documentation). I therefore think it is appropriate to leave the community guidance in.

nielsleadholm:
@thousandbrainsproject/monty-maintainers Thank you all for your feedback. I would like to motion for FCP and invite all maintainers to indicate their final votes on the merging of this RFC.

👍 for merge
👎 for object
👀 for abstain

PR #428, opened 2025-08-15
rfc: centralized process to update the future work artifacts

codeallthethingz:

PR #432, opened 2025-08-19
chore: add functionality to github_readme_sync tool to build an index file from the documentation Frontmatter.

codeallthethingz:
This PR adds a command to the github readme sync tool that generates an index.json file containing frontmatter found in markdown documents in a source directory.

The primary use-case for this tool is as part of #428 and will be run against the `/docs` directory.

The generated data file will be used by a downstream consumer.

For each markdown file found in the target directory (presumably `/docs`), extract the frontmatter, and build a JSON array for each markdown file containing all the values in the frontmatter, as well as including the unprocessed markdown text.

This tool also generates the following fields in the data file.
- path - A path to the markdown file found. `/docs/future-work/oss/some-markdown-file.md`
- path1 - the first path segment after the target directory `future-work`
- path2 - the second path segment after the target directory `oss`
- pathN - ...

Example output data file

```javascript
[
  {
    "title": "Re-Anchor Hypotheses for Robustness to Noise and Distortions",
    "slug": "re-anchor-hypotheses",
    "path": "docs/future-work/learning-module-improvements/re-anchor-hypotheses.md",
    "text": "One aspect that we believe may...",
    "path1": "future-work",
    "path2": "learning-module-improvements"
  },
// ...
]
```

### DIFF
--- a/tools/github_readme_sync/tests/index_test.py
+++ b/tools/github_readme_sync/tests/index_test.py
@@ -34,7 +34,9 @@ class TestGenerateIndex(unittest.TestCase):
         with open(md_file_path, "w", encoding="utf-8") as f:
             f.write(content)
 
-        index_file_path = generate_index(self.temp_dir, f"{self.temp_dir}/index.json")
+        index_file_path = generate_index(
+            self.temp_dir, str(Path(self.temp_dir) / "index.json")
+        )
 
         self.assertTrue(Path(index_file_path).exists())
 
@@ -99,8 +101,8 @@ class TestGenerateIndex(unittest.TestCase):
                     generate_index(actual_docs_dir, actual_output_file)
 
     def test_generate_index_nonexistent_docs_folder(self):
-        nonexistent_dir = f"{self.temp_dir}/nonexistent"
-        output_file_path = f"{self.temp_dir}/index.json"
+        nonexistent_dir = str(Path(self.temp_dir) / "nonexistent")
+        output_file_path = str(Path(self.temp_dir) / "index.json")
 
         with self.assertRaises(ValueError) as context:
             generate_index(nonexistent_dir, output_file_path)


tools/github_readme_sync/cli.py
```diff
@@ -16,18 +15,19 @@
...
     create_hierarchy_file,
 )
-from tools.github_readme_sync.readme import ReadMe  # noqa: E402
-from tools.github_readme_sync.upload import upload  # noqa: E402
+from tools.github_readme_sync.index import generate_index
+from tools.github_readme_sync.readme import ReadMe
+from tools.github_readme_sync.upload import upload
+
+monty_root = Path(__file__).resolve().parent.parent.parent
+sys.path.append(str(monty_root))
```
tristanls: **issue**: Rearranging this order breaks some tool invocation patterns and you have to be in a specific directory to invoke the tool.

From original pull request in the archive:

> This pull request enables the github_readme_sync tool to run standalone, e.g.:
> - `monty_root_dir$ python tools/github_readme_sync/cli.py -h`
> - `monty_root_dir/tools$ python github_readme_sync/cli.py -h`
> - `monty_root_dir/tools/github_readme_sync$ python cli.py -h`
> 
> The `# noqa: E402` directives are there to ignore import sort order as the code must be added before importing `from tools.*`.

tools/github_readme_sync/file.py
```diff
@@ -10,6 +10,10 @@
 
 import os
 from pathlib import Path
+from typing import List, Optional
+
+DEFAULT_IGNORE_DIRS = [".pytest_cache", ".github", ".git", "figures", "snippets"]
```
tristanls: **issue**: Anything starting with `.` should be ignored.

tools/github_readme_sync/file.py
```diff
@@ -18,3 +22,47 @@ def get_folders(file_path: str) -> list:
         for name in os.listdir(file_path)
         if Path(file_path).joinpath(name).is_dir()
     ]
+
+
+def find_markdown_files(
+    folder: str,
+    ignore_dirs: Optional[List[str]] = None,
+    ignore_files: Optional[List[str]] = None,
```
tristanls: **nitpick**: `list[str] | None` instead of `Optional[List[str]]`

tools/github_readme_sync/file.py
```diff
@@ -18,3 +22,47 @@ def get_folders(file_path: str) -> list:
...
+    ignore_dirs = ignore_dirs or DEFAULT_IGNORE_DIRS
+    ignore_files = ignore_files or DEFAULT_IGNORE_FILES
+
+    md_files = []
+    for root, _, files in os.walk(folder):
+        if any(ignore_dir in root for ignore_dir in ignore_dirs):
+            continue
+
+        md_files.extend(
+            os.path.join(root, file)
```
tristanls: **issue (non-blocking)**: use `pathlib.Path` instead of `os.path`.

**note**: I'm surprised `ruff` didn't complain 🤔.

tools/github_readme_sync/file.py
```diff
@@ -18,3 +22,47 @@ def get_folders(file_path: str) -> list:
...
+    Args:
+        folder: Root directory to search
+        ignore_dirs: List of directory names to exclude (uses defaults if None)
+        ignore_files: List of file names to exclude (uses defaults if None)
+
+    Returns:
+        List of full paths to markdown files
+    """
+    ignore_dirs = ignore_dirs or DEFAULT_IGNORE_DIRS
+    ignore_files = ignore_files or DEFAULT_IGNORE_FILES
```
tristanls: **issue**: If the caller passes in an empty list (so as to not ignore any dirs or files), this will replace the empty list with the default list.

**suggestion**: Should be `DEFAULT_INGORE_DIRS if ignore_dirs is None else ignore_dirs`

tools/github_readme_sync/index.py
```diff
@@ -0,0 +1,123 @@
...
+
+    Args:
+        docs_dir: The directory containing markdown files to scan.
+        output_file_path: Path where to write the output file.
+
+    Returns:
+        Path to the generated output file.
+    """
+    output_file = output_file_path
+    logging.info(f"Scanning docs directory: {CYAN}{docs_dir}{RESET}")
```
tristanls: **issue**: Here and rest of file should be `logger.info`.

**suggestion**: `logger = logging.getLogger(__file__)` at top of the module.

tools/github_readme_sync/index.py
```diff
@@ -0,0 +1,123 @@
...
+    """Generate index.json file from docs directory.
+
+    Args:
+        docs_dir: The directory containing markdown files to scan.
+        output_file_path: Path where to write the output file.
+
+    Returns:
+        Path to the generated output file.
+    """
+    output_file = output_file_path
```
tristanls: **issue**: This looks like an extraneous assignment. Any reason `output_file_path` cannot be used directly?

tools/github_readme_sync/index.py
```diff
@@ -0,0 +1,123 @@
...
+
+    Continues if there are errors in the markdown files.
+
+    Returns:
+        List of dictionaries containing extracted front-matter and the body text.
+
+    Raises:
+        ValueError: If directory doesn't exist.
+    """
+    if not os.path.exists(docs_dir):
```
tristanls: **issue (non-blocking)**: `pathlib.Path` over `os.path`.

tools/github_readme_sync/index.py
```diff
@@ -0,0 +1,123 @@
...
+        raise ValueError(f"Directory {docs_dir} does not exist")
+
+    entries = []
+    docs_path = Path(docs_dir)
+    folder_name = os.path.basename(docs_dir)
```
tristanls: **issue**: Since you're using `Path`, then, for consistency, `os.path` should not be used.

tools/github_readme_sync/index.py
```diff
@@ -0,0 +1,123 @@
...
+
+    for md_file_path in find_markdown_files(docs_dir):
+        md_file = Path(md_file_path)
+        logging.info(f"Processing: {CYAN}{md_file.relative_to(docs_path)}{RESET}")
+
+        try:
+            content = read_file_content(md_file_path)
+            frontmatter = parse_frontmatter(content)
+        except (OSError, UnicodeDecodeError):
+            logging.exception(f"Error reading {md_file_path}")
```
tristanls: **nitpick**: Should be `"Error reading %s", md_file_path"`. Avoid f-strings in log messages. [G004](https://docs.astral.sh/ruff/rules/logging-f-string/)

tools/github_readme_sync/index.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    if not os.path.exists(docs_dir):
+        raise ValueError(f"Directory {docs_dir} does not exist")
+
+    entries = []
+    docs_path = Path(docs_dir)
+    folder_name = os.path.basename(docs_dir)
+
+    for md_file_path in find_markdown_files(docs_dir):
+        md_file = Path(md_file_path)
```
tristanls: **suggestion (non-blocking)**: This usage suggests that `find_markdown_files()` should return `list[Path]` instead of `list[str]`.

tools/github_readme_sync/index.py
```diff
@@ -0,0 +1,123 @@
...
+            "title": frontmatter.get("title", ""),
+            "slug": slugify(md_file.stem),
+            "path": f"{folder_name}/{relative_path}",
+            "text": body_content.strip(),
+        }
+
+        entry.update(
+            {
+                field: value
+                for field, value in frontmatter.items()
```
tristanls: **issue**: Missing validation of [keys and values allowed to be in the index](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0015_future_work.md#new-front-matter-values).

tools/github_readme_sync/index.py
codeallthethingz: shame: no.

tools/github_readme_sync/index.py
codeallthethingz: 🤔 I agree there should be some validation here to stop PRs exploding the index / trying json injection attacks, however, I was keeping all the specific future work business logic in the next tool that checks values / keys etc...

How about:
- we set a maximum length for the field name / value and if it exceeds that we log a warning and exclude it.
- sanitize both the key and value using nh3.

tristanls: Sounds reasonable to me.

tools/github_readme_sync/file.py
```diff
@@ -7,14 +7,65 @@
...
+    Args:
+        folder: Root directory to search
+        ignore_dirs: List of directory names to exclude (uses defaults if None)
+        ignore_files: List of file names to exclude (uses defaults if None)
+
+    Returns:
+        List of Path objects to markdown files
+    """
+    ignore_dirs = DEFAULT_IGNORE_DIRS if ignore_dirs is None else ignore_dirs
+    ignore_files = ignore_files or DEFAULT_IGNORE_FILES
```
tristanls: **nitpick**: This should be the same pattern as `ignore_dirs` above, but I'm leaving it as a nitpick since, in practice, the fail case of the present pattern probably won't trigger.

tools/github_readme_sync/index.py
```diff
@@ -0,0 +1,197 @@
...
+    if _is_empty(docs_dir):
+        raise ValueError("docs_dir cannot be empty")
+    if _is_empty(output_file_path):
+        raise ValueError("output_file_path cannot be empty")
+
+    logger.info(f"Scanning docs directory: {CYAN}{docs_dir}{RESET}")
+
+    entries = process_markdown_files(docs_dir)
+
+    os.makedirs(os.path.dirname(output_file_path), exist_ok=True)
```
tristanls: **issue**: `Path` instead of `os.path`.

tools/github_readme_sync/index.py
tristanls: **note**: As I'm re-reviewing, I see I communicated poorly. I wasn't thinking about security, although the improvement is good 👍. I was thinking of making a typo in the markdown docs, and the field doesn't show up in the widget, or new mistyped fields start showing up in the widget. The use case I was thinking of is that since this is where we parse these fields, this is where I should get feedback if I made a mistake.

The feedback loop with the code here would be: commit -> deploy -> see mistake -> another PR
The feedback loop with validation here would be: commit -> check error -> fix commit -> deploy

With validation here, I am assuming the feedback loop to the author would be faster.

tools/github_readme_sync/tests/file_test.py
```diff
@@ -39,5 +39,56 @@ def test_get_folders(self):
...
+        (regular_dir / "guide.md").write_text("# Guide")
+        (dot_git_dir / "config.md").write_text("# Git config")
+        (dot_vscode_dir / "settings.md").write_text("# VSCode settings")
+        (nested_dot_dir / "secret.md").write_text("# Secret doc")
+
+        (regular_dir / "regular.txt").write_text("Not a markdown file")
+
+        result = find_markdown_files(self.temp_dir)
+
+        result_basenames = [os.path.basename(path) for path in result]
```
tristanls: **issue (non-blocking)**: `Path` instead of `os.path`.

tools/github_readme_sync/tests/index_test.py
```diff
@@ -0,0 +1,134 @@
...
+    ) -> list:
+        subdir_path = Path(self.temp_dir) / subdir
+        subdir_path.mkdir(parents=True, exist_ok=True)
+
+        content = f"---\ntitle: test doc\n{frontmatter_fields}\n---\n"
+        md_file_path = subdir_path / "test-doc.md"
+        with open(md_file_path, "w", encoding="utf-8") as f:
+            f.write(content)
+
+        output_file_path = os.path.join(self.temp_dir, "index.json")
```
tristanls: **issue (non-blocking)**: `Path` instead of `os.path` here and the rest of the file.

tools/github_readme_sync/tests/index_test.py
```diff
@@ -0,0 +1,134 @@
+# Copyright 2025 Thousand Brains Project
+# Copyright 2024 Numenta Inc.
```
tristanls: **issue**: Nope.

tools/github_readme_sync/index.py
```diff
@@ -0,0 +1,197 @@
+# Copyright 2025 Thousand Brains Project
+# Copyright 2024 Numenta Inc.
```
tristanls: **issue**: Nope.

tristanls (commented):
Thank you for the changes. Almost there.

Just got a few copyright header issues.

I commented on `Path` vs `os.path` in tests, but not gonna block based on that. But there is one more in `index.py` for consistency.

codeallthethingz:
@tristanls I can't respond to this comment so inlining here

> note: As I'm re-reviewing, I see I communicated poorly. I wasn't thinking about security, although the improvement is good 👍. I was thinking of making a typo in the markdown docs, and the field doesn't show up in the widget, or new mistyped fields start showing up in the widget. The use case I was thinking of is that since this is where we parse these fields, this is where I should get feedback if I made a mistake.
> The feedback loop with the code here would be: commit -> deploy -> see mistake -> another PR
The feedback loop with validation here would be: commit -> check error -> fix commit -> deploy
> With validation here, I am assuming the feedback loop to the author would be faster.

Not how I'm thingking about it: This tool, as well as the subsequest widget generation tool, will all be called within one github action (using the artifacts/tmp dir to store the index.json file) so the feedback for the developer will be within a single PR cycle.

tools/github_readme_sync/tests/index_test.py
```diff
@@ -96,14 +93,14 @@ def test_generate_index_invalid_parameters(self):
...
+                    actual_output_file = str(Path(self.temp_dir) / "index.json")
 
                 with self.assertRaises(expected_exception):
                     generate_index(actual_docs_dir, actual_output_file)
 
     def test_generate_index_nonexistent_docs_folder(self):
-        nonexistent_dir = os.path.join(self.temp_dir, "nonexistent")
-        output_file_path = os.path.join(self.temp_dir, "index.json")
+        nonexistent_dir = f"{self.temp_dir}/nonexistent"
+        output_file_path = f"{self.temp_dir}/index.json"
```
tristanls: **issue**: Without `os.path.join` I think this should be like line L99R96 above:
```python
nonexistent_dir = str(Path(self.temp_dir) / "nonexistent")
output_file_path = str(Path(self.temp_dir) / "index.json")
```

codeallthethingz:
> One more `Path` thing. `os.path.join` was more robust than the f-string.
> 
> Regarding the feedback loop (non-blocking)...
> 
> Maybe I'm missing something. You have the whole system in mind, so perhaps there is validation happening somewhere that I am unaware of. From my current understanding, I don't think it matters if everything is happening in a Github Action. If there is no key validation, then I don't have any automation telling me I made a mistake until I go to the deployed website and happen to notice that I spelled something incorrectly. A scenario I have in mind:
> 
> 1. I try to set difficult to "easy" and type in "eassy."
> 2. GitHub Action accepts everything and deploys future work, which now has a `easy` tag and an `eassy` tag.
> 3. One epoch later...
> 4. I happen to go on the website and notice "eassy". Oh noes, let me make a PR to fix it.
> 
> If there is validation in the code, then the scenario becomes:
> 
> 1. I try to set difficult to "easy" and type in "eassy."
> 2. GitHub Action runs the check and `github_readme_sync` tool fails with an error that "eassy" is not a valid key.
> 3. I'm still in the PR, make another commit, and fix it.

Yes, the validation is happening somewhere else (in the next PR). I wanted to separate that logic out from github_readme_sync tool as it shouldn't live there.

PR #434, opened 2025-08-20
chore: future work tool, front matter, widget

codeallthethingz:


codeallthethingz:
@vkakerbeck tagging you here as there is a new documentation page that you might want to have a look at - https://thousandbrainsproject.readme.io/v0.11-codeallthethingz-future-work-tool-and-front-matter/docs/future-work-widget-metadata.

docs/contributing/documentation/future-work-widget-metadata.md
```diff
@@ -0,0 +1,41 @@
...
+
+!snippet[../../snippets/future-work-status.md]
+
+# RFC
+
+Does this work item required an RFC? (These values are processed in the `validator.py` code) and can be of the form:
+
+`https://github\.com/thousandbrainsproject/tbp\.monty/.*` `required` `optional` `not-required` `unknown`
+
+# Owner
```
tristanls: **issue**: [Per RFC 15](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0015_future_work.md#new-front-matter-values), these are "contributors", here and everywhere else.

docs/contributing/documentation/future-work-widget-metadata.md
```diff
@@ -0,0 +1,41 @@
+---
+title: Future Work Widget Metadata
+---
+
+The future work documents have special Frontmatter metadata that is used to power the future-work widget.  The following fields are validated against allow lists defined in snippet files to ensure consistency and quality.
```
vkakerbeck: Could you include a template/example of the special front matter that people can copy when making a new page?

docs/future-work/oss-communication-improvements/make-more-condensed-videos-about-the-project-monty.md
```diff
@@ -1,3 +1,6 @@
 ---
 title: Make More Condensed Videos About the Project & Monty
+rfc: required
+estimated-scope: medium
+skills: research, thousand brains theory, video editing, content marketing, content creation
```
vkakerbeck: Why does this item require 'research' and 'content marketing'? It seems like those are not directly related to creating the content. Also `thousand brains theory` may not be required (i.e. someone could consult with researchers on the content but coordinate the production)

docs/future-work/oss-communication-improvements/organize-start-podcast-series.md
```diff
@@ -1,3 +1,6 @@
 ---
 title: Organize & Start Podcast Series
+rfc: required
+estimated-scope: large
+skills: podcasting, interviewing, research, thousand brains theory, writing, video editing, content marketing
```
vkakerbeck: similarly here, I think some of those skills are not required to 'organize' a podcast series.

docs/snippets/future-work-skills.md
```diff
@@ -0,0 +1 @@
+`CSS` `HTML` `JS` `S3` `content marketing` `github-actions` `github_readme_sync tool` `interviewing` `podcasting` `python` `research` `thousand brains theory` `video editing` `writing` `content creation`
```
vkakerbeck: is there a reason why github-actions uses '-' and the sync tool '_'? Can we make the use of -, _ and space consistent?

docs/snippets/future-work-tags.md
```diff
@@ -0,0 +1 @@
+`accuracy` `pose` `numsteps` `speed` `noise` `learning` `multiobj` `generalization` `compositional` `deformations` `featsandmorph` `scale` `realworld` `dynamic` `goalpolicy` `abstract` `adversarial` `transfer` `infrastructure` `oss`
```
vkakerbeck: I know these are taken from the tags in our current future work table but since we are not really tight on space here, should we write these out (like featsandmorph -> features_and_morphology or 'features and morphology') Similarly with 'goalpolicy' maybe we should add a space or _ between the words.

tools/future_work_widget/README.md
```diff
@@ -0,0 +1,94 @@
...
+**Usage:** `?columns=column1,column2,column3`
+
+**Available columns:**
+- `title` - The title of the future work item with edit/view links
+- `estimated-scope` - Size estimation (small, medium, large)
+- `rfc` - Related RFC link or reference
+- `status` - Current status with owner avatars
+- `tags` - Categorization tags
+- `skills` - Required skills for the work
+
```
vkakerbeck, line 71: What happened to `improved-metric` (might be related to `tags`) and `output-type` from the RFC?

tools/future_work_widget/app/images/favicon.png
vkakerbeck, line 1: Haha nice!

docs/future-work/oss-communication-improvements/make-more-condensed-videos-about-the-project-monty.md
codeallthethingz: These are just placeholder values, I think these conversations about what should be in these lists will wait for the next PR!  I'll reduce these down to uncontestable values for now.

docs/future-work/oss-communication-improvements/organize-start-podcast-series.md
codeallthethingz: as above.

docs/snippets/future-work-skills.md
codeallthethingz: no particular reason.  

Having thought about it a bit, I've adopted kebab-case as the standard.

docs/snippets/future-work-tags.md
codeallthethingz: expanded. But these can all change in the next PR when we actually go through all the documents and apply the Frontmatter.

tools/future_work_widget/README.md
codeallthethingz: I added in improved-metric and output-type.

docs/snippets/future-work-skills.md
vkakerbeck: haha never knew that's what it is called :D

vkakerbeck (approved):
Nice, thanks for the updates

codeallthethingz:
@tristanls I updated from the RFC content.

tools/future_work_widget/README.md
```diff
@@ -0,0 +1,96 @@
...
+```
+pytest -n 0 tools/future_work_widget
+```
+
+## Running Locally
+
+To try the tool out simply run the following command from the tbp.monty directory:
+
+```
+./tools/future_work_widget/run-local.sh
```
tristanls: **issue**: Running this command results in an error:
```
~/tbp/tbp.monty % ./tools/future_work_widget/run-local.sh
python -m tools.github_readme_sync.cli generate-index docs /tmp/index.json
Scanning docs directory: docs
Processing: contributing/comment-on-issues.md
...
Processing: future-work/cmp-hierarchy-improvements/figure-out-performance-measure-and-supervision-in-heterarchy.md
Generated index with 139 entries: /tmp/index.json
python -m tools.future_work_widget.cli /tmp/index.json tools/future_work_widget/app
{
  "success": true,
  "processed_items": 53,
  "total_items": 139
}
./tools/future_work_widget/run-local.sh: line 17: http-server: command not found
```

**note**: The above error happens after I followed the installation instructions
```
pip install -e '.[future_work_widget_tool]'
```

tools/future_work_widget/tests/validator_test.py
```diff
@@ -0,0 +1,163 @@
...
+            "pose",
+            "learning",
+            "multiobj",
+        ]
+        expected_skills = [
+            "python",
+            "github-actions",
+            "JS",
+            "HTML",
+        ]
```
tristanls: **nitpick**: Feels like it'd be DRY'er to use these lists when writing to the files in the setup above.

tools/future_work_widget/tests/validator_test.py
```diff
@@ -0,0 +1,163 @@
...
+
+        expected_values = [
+            "simple-word",
+            "accuracy",
+            "learning",
+        ]
+
+        self.assertEqual(
+            sorted(validator.exact_values["tags"]), sorted(expected_values)
+        )
```
tristanls: **question**: How does this test word boundary wrapping?

tools/future_work_widget/validator.py
```diff
@@ -0,0 +1,388 @@
...
+
+        for file_path in future_work_files:
+            field_name = file_path.stem.replace("future-work-", "")
+
+            if field_name in self.CUSTOM_VALIDATION_FIELDS:
+                logger.debug(
+                    f"Skipping {file_path.name} - using custom validation "
+                    f"logic for '{field_name}'"
+                )
+                continue
```
tristanls: **issue**: This seems like it should be an error. The configuration is wrong, and it is being ignored.

**suggestion**: Fail here and force the user to remove the erroneous file.

tools/future_work_widget/validator.py
```diff
@@ -0,0 +1,388 @@
...
+            record: The record to validate
+
+        Returns:
+            Tuple of (record or None if there are errors, list of validation errors)
+        """
+        if record.get("path1") != "future-work" or "path2" not in record:
+            return None, []
+
+        record_copy = record.copy()
+        file_path = record.get("path")
```
tristanls: **issue**: Because you're using `.get(...)`, the `file_path` could be `None`. However, none of the methods below declares nor handles that this could be the case.
```python
>>> a = {}
>>> b = a.get("path")
>>> b is None
True
````

Image

File: tools/future_work_widget/validator.py [hunk 13]
```diff
@@ -0,0 +1,388 @@
...
+
+        errors = []
+        self._validate_comma_separated_fields(record_copy, file_path, errors)
+        self._validate_required_fields(record_copy, file_path, errors)
+        self._validate_custom_fields(record_copy, file_path, errors)
+        self._validate_field_values(record_copy, file_path, errors)
+        return (None, errors) if errors else (record_copy, errors)
```
tristanls: **nitpick**: I think this is clearer:
```python
return (None if errors else record_copy, errors)
```

tools/future_work_widget/validator.py
```diff
@@ -0,0 +1,388 @@
...
+        Returns:
+            Tuple of (record or None if there are errors, list of validation errors)
+        """
+        if record.get("path1") != "future-work" or "path2" not in record:
+            return None, []
+
+        record_copy = record.copy()
+        file_path = record.get("path")
+
+        errors = []
```
tristanls: **nitpick**: Missing type annotation
```python
errors: list[ValidationError] = []
```

tools/future_work_widget/validator.py
```diff
@@ -0,0 +1,388 @@
...
+    ) -> None:
+        """Process comma-separated fields by validating and transforming them.
+
+        Args:
+            record: The record being transformed (modified in place)
+            file_path: Path to the source file for error reporting
+            errors: List to append validation errors to
+        """
+        for field in self.COMMA_SEPARATED_FIELDS:
+            if field in record and isinstance(record[field], str):
```
tristanls: **question**: What is the purpose of the `isinstance` check here? Why would it not be a string? Why would it be valid to skip if it is in `COMMA_SEPARATED_FIELDS` but somehow not a string?

tools/future_work_widget/validator.py
```diff
@@ -0,0 +1,388 @@
...
+            return None, []
+
+        record_copy = record.copy()
+        file_path = record.get("path")
+
+        errors = []
+        self._validate_comma_separated_fields(record_copy, file_path, errors)
+        self._validate_required_fields(record_copy, file_path, errors)
+        self._validate_custom_fields(record_copy, file_path, errors)
+        self._validate_field_values(record_copy, file_path, errors)
```
tristanls: **question**: What is the reason for using side-effects?

**note**: It's python. It doesn't feel to me like optimizing via side-effects is worth the code clarity.

**suggestion (non-blocking)**: a more functional approach
```python
record, errors = self._validate_comma_separated_fields(record_copy, file_path)
(record, errs) = self._validate_required_fields(record, file_path))
errors.append(errs)
(record, errs) = self._validate_custom_fields(record, file_path))
errors.append(errs)
(record, errs) = self._validate_field_values(record, file_path))
errors.append(errs)
```

tools/future_work_widget/validator.py
```diff
@@ -0,0 +1,388 @@
...
+        items = [item.strip() for item in field_value.split(",")]
+        if len(items) > self.MAX_COMMA_SEPARATED_ITEMS:
+            self._add_validation_error(
+                f"{field_name} field cannot have more than "
+                f"{self.MAX_COMMA_SEPARATED_ITEMS} items. "
+                f"Got {len(items)} items: {', '.join(items)}",
+                file_path,
+                field_name,
+                errors,
+            )
```
tristanls: **question**: why the side-effects?

**suggestion (non-blocking)**: I feel this would read a lot clearer without side-effects
```python
errors.append(self._validation_error(...))
```

tools/future_work_widget/cli.py
```diff
@@ -0,0 +1,51 @@
...
+
+from __future__ import annotations
+
+import argparse
+import json
+import logging
+import sys
+from pathlib import Path
+
+from .build import build
```
tristanls: **note**: This makes the `cli` not runnable from `future_work_widget` directory:
```
~/tbp/tbp.monty/tools/future_work_widget % python ./cli.py --help
  File "./cli.py", line 18, in <module>
    from .build import build
ImportError: attempted relative import with no known parent package
```
or
```
~/tbp/tbp.monty/tools/future_work_widget % python -m cli --help
  File "/Users/tslominski/miniconda3/envs/tbp.monty/lib/python3.8/runpy.py", line 194, in _run_module_as_main
  File "/Users/tslominski/miniconda3/envs/tbp.monty/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/tslominski/tbp/tbp.monty/tools/future_work_widget/cli.py", line 18, in <module>
    from .build import build
ImportError: attempted relative import with no known parent packag
```
this is why other tools have this bit at the top and no relative imports:
```python
monty_root = Path(__file__).resolve().parent.parent.parent
sys.path.append(str(monty_root))

# Note: this tool requires this specific import order so don't remove
# the noqa: E402 comments
from tools.github_readme_sync.colors import RED, RESET  # noqa: E402
...
```

tools/future_work_widget/cli.py
```diff
@@ -0,0 +1,51 @@
...
+        description="Build the data and package the future work widget."
+    )
+    parser.add_argument("index_file", help="The JSON file to validate and transform")
+    parser.add_argument(
+        "output_dir", help="The output directory to create and save data.json"
+    )
+    parser.add_argument(
+        "--docs-snippets-dir",
+        help="Optional path to a snippets directory for validation files",
+        default="docs/snippets",
```
tristanls: **suggestion (non-blocking)**: You can use `Path` here
```python
default=Path("docs/snippets")
```

tools/future_work_widget/build.py
```diff
@@ -0,0 +1,132 @@
...
+        output_dir: Path to the output directory to create and save data.json
+        docs_snippets_dir: Path to docs/snippets directory for validation files
+
+    Returns:
+        Dict with keys:
+        - success: bool indicating if build was successful
+        - processed_items: int number of items processed
+        - total_items: int total number of items found
+        - errors: list of error dicts with file/line/message info
+        - error_message: str summary error message (only if success=False)
```
tristanls: **issue**: If you know all the keys, then this shouldn't be a generic dictionary. At the very least, it should be a `TypedDict`, next up a `Dataclass`, and lastly, a custom class. Name suggestion: `BuildResult`.

tools/future_work_widget/build.py
```diff
@@ -0,0 +1,132 @@
...
+
+def _validate_params(
+    index_file: Path,
+    output_dir: Path,
+    docs_snippets_dir: Path,
+) -> dict[str, Any] | None:
+    """Validate input paths and setup output directory.
+
+    Returns:
+        None if all validations pass, otherwise error dict with success=False
```
tristanls: **issue**: Same as above, if the keys are known, then this should not be a generic dictionary.

tools/future_work_widget/build.py
```diff
@@ -0,0 +1,132 @@
...
+    Returns:
+        Dict with keys:
+        - success: bool indicating if build was successful
+        - processed_items: int number of items processed
+        - total_items: int total number of items found
+        - errors: list of error dicts with file/line/message info
+        - error_message: str summary error message (only if success=False)
+    """
+    try:
+        validation_error = _validate_params(index_file, output_dir, docs_snippets_dir)
```
tristanls: **praise**: I like the functional pattern here.

**note**: This is not empty praise as per https://conventionalresponses.org/ 😉. It was a nice contrast to see this after seeing side-effect use in the `validator.py`.

tools/future_work_widget/build.py
```diff
@@ -0,0 +1,132 @@
...
+        if validation_error is not None:
+            return validation_error
+
+        with open(index_file, "r", encoding="utf-8") as f:
+            data = json.load(f)
+
+        if not isinstance(data, list):
```
tristanls: **note**: `isinstance`... ewww... but anything else feels like too much 🫤 .. `pydantic` would probably help with validation here. I'll add a note to `validator.py` about it.

tools/future_work_widget/validator.py
```diff
@@ -0,0 +1,388 @@
+# Copyright 2025 Thousand Brains Project
```
tristanls, line 1: **question**: Have you tried using [Pydantic](https://docs.pydantic.dev/latest/) for this? Specifically, [`@field_validator`](https://docs.pydantic.dev/latest/concepts/validators/#field-validators)? `validator.py` appears to be a partial implementation of Pydantic functionality and has the largest number of non-test lines of code added in this PR (only the entire JavaScript app comes close).

**note**: It seems that the maintainers of the future work section should be capable of adding allowlisted values in a `Pydantic` field validator code instead of markdown files.

tools/future_work_widget/tests/build_test.py
```diff
@@ -0,0 +1,387 @@
...
+        """Create snippet files with given configurations.
+
+        Args:
+            snippet_configs: Dict mapping filename to content
+
+        Returns:
+            Path to snippets directory
+        """
+        snippets_dir = self.temp_path / "snippets"
+        snippets_dir.mkdir(exist_ok=True)
```
tristanls, line 39: **note**: Oh, interesting regarding `exist_ok=True`.. I would have thought that `tempfile.mkdtemp()` would not reuse names.

tools/future_work_widget/tests/build_test.py
```diff
@@ -0,0 +1,387 @@
...
+            snippets_dir = self._create_snippets({})
+
+        result = build(index_file, self.temp_path, snippets_dir)
+
+        if not result["success"]:
+            raise ValueError(result["error_message"])
+
+        data_file = self.temp_path / "data.json"
+        with open(data_file, "r", encoding="utf-8") as f:
+            return json.load(f)
```
tristanls, line 101: **issue**: As far as I can tell, this is a duplicate of and subsumes `_run_build_test_with_snippets` function above.

tools/future_work_widget/tests/build_test.py
```diff
@@ -0,0 +1,387 @@
...
+        with open(index_file, "w", encoding="utf-8") as f:
+            json.dump(input_data, f)
+
+        if snippets_dir is None:
+            snippets_dir = self._create_snippets({})
+
+        result = build(index_file, self.temp_path, snippets_dir)
+
+        self.assertFalse(result["success"])
+        self.assertIn(expected_error_fragment, result["error_message"])
```
tristanls: **issue**: Duplicating code.

**suggestion**: This should probably call a `_run_build` and expect the exception from `raise ValueError(result["error_message"])`.

tools/future_work_widget/tests/build_test.py
```diff
@@ -0,0 +1,387 @@
...
+        result = build(index_file, self.temp_path, snippets_dir)
+
+        if not result["success"]:
+            raise ValueError(result["error_message"])
+
+        data_file = self.temp_path / "data.json"
+        with open(data_file, "r", encoding="utf-8") as f:
+            return json.load(f)
+
+    def _create_test_item(self, **overrides) -> dict[str, Any]:
```
tristanls: **suggestion**: Rename to `_create_future_work_item`. Makes it clearer in context like:
```python
input_data = [
    self._create_future_work_item(
        path2="voting-improvements",
        title="Improve voting mechanism",
        content="Test content for voting",
    ),
    {
        "path1": "how-monty-works",
        "path2": "learning-modules",
        "title": "Learning modules overview",
        "content": "Test content for learning",
    },
]
```

docs/contributing/documentation.md
```diff
@@ -170,4 +170,4 @@ The documentation [Style Guide](style-guide.md#documentation-style-guide)
 
 # Versioning
 
-The Monty documentation uses the first two parts of semantic versioning (semver), as there is nothing to document for patch changes.  You can read about semver here <https://semver.org/>.
+The Monty documentation uses the first two parts of semantic versioning (semver), as there is nothing to document for patch changes.  You can read about semver here <https://semver.org/>.
```
tristanls: **note**: I hope this adds a newline at the end of the file and does not remove it.

tools/future_work_widget/app/js/future-work-widget.js
```diff
@@ -0,0 +1,323 @@
...
+      this.setupSearch(table);
+    } catch (error) {
+      console.error('Failed to initialize Future Work Widget:', error);
+      this.showError('Failed to load data - see the console for more details or refresh the page to try again.');
+    }
+  },
+
+
+  async loadData() {
+    const response = await fetch(`data.json?t=${Date.now()}`, {
```
tristanls: **question**: Is `t` intended to be a cache buster? (I might have to set up the infrastructure to care about it)

tools/future_work_widget/app/js/future-work-widget.js
```diff
@@ -0,0 +1,323 @@
...
+    if (!response.ok) {
+      throw new Error(`HTTP error! status: ${response.status}, body: ${await response.text()}`);
+    }
+    const data = await response.json();
+    return Array.isArray(data)
+      ? data.slice().sort((a, b) =>
+          (a.path2 || '').localeCompare(b.path2 || '', undefined, { sensitivity: 'base' }) ||
+          (a.title || '').localeCompare(b.title || '', undefined, { sensitivity: 'base' })
+        )
+      : data;
```
tristanls: **question**: Why would the response come in different formats?

tools/future_work_widget/app/js/future-work-widget.js
```diff
@@ -0,0 +1,323 @@
...
+      groupBy: 'path2'
+    });
+  },
+
+
+  setupSearch(table) {
+    const searchInput = document.getElementById('searchInput');
+    if (!searchInput) {
+      console.warn('Search input not found');
+      return;
```
tristanls: **question**: Not throwing (which would show error "page") if can't set up search? Is this intended?

tools/future_work_widget/app/js/future-work-widget.js
```diff
@@ -0,0 +1,323 @@
...
+        ]
+          .filter(Boolean)
+          .join(' ')
+          .toLowerCase();
+
+        return trimmedTerm
+          .split(/\s+/)
+          .every(word => searchableText.includes(word));
+      });
+    };
```
tristanls, line 250: **praise**: Oh interesting. I didn't realize you had to implement the search yourself. Noice.

tools/future_work_widget/app/js/future-work-widget.js
```diff
@@ -0,0 +1,323 @@
...
+            copyUrlLink.textContent = '📋';
+            copyUrlLink.classList.remove('success');
+          }, 1500);
+        } catch (err) {
+          console.error('Failed to copy URL to clipboard:', err);
+          copyUrlLink.textContent = '❌';
+          setTimeout(() => {
+            copyUrlLink.textContent = '📋';
+          }, 1500);
+        }
```
tristanls: **note**: My brain struggles so hard without a framework 😆

tools/future_work_widget/app/js/future-work-widget.js
```diff
@@ -0,0 +1,323 @@
...
+      });
+    }
+
+    document.addEventListener('click', (e) => {
+      const searchValue = e.target.dataset.searchValue;
+      if (searchValue) {
+        addToSearch(searchValue);
+      }
+    });
```
tristanls, line 283: **question**: What user interaction is this for?

tristanls (commented):
Lots of good stuff here. I hope the number of comments I left does not diminish the quality of what you put together.

Since I left plenty comments, I'm not going to try to summarize them all here.

Also, I'd like to run the app locally before finally approving, so I left a comment about that.

tools/future_work_widget/README.md
codeallthethingz: ah, that'll be because http-server is an npm package.

I think run-local should probably be a python script, and it runs a simple http server.

tools/future_work_widget/tests/validator_test.py
codeallthethingz: ah dangling test from a previous version that used to put boundaries around things. Removed.

tools/future_work_widget/validator.py
codeallthethingz: i mean techincally its getting parsed as Yaml so it could be any json type object. But for our purpsoses it should always be string. I'll just convert to string instead of checking and skipping and then the validation should catch everything else.

tristanls-tbp (commented):
I think I am waiting for more commits. I'm commenting here so you can request a re-review if the content is ready for another review.

tools/future_work_widget/validator.py
codeallthethingz: more about this in a later summary comment as this introduced a lot of code change.
High level, I did both, I wanted the snippets so that I could have a central place to update the docs and the validation.

docs/contributing/documentation.md
codeallthethingz: 😊

tools/future_work_widget/app/js/future-work-widget.js
codeallthethingz: yes.  Cache busting.

tools/future_work_widget/app/js/future-work-widget.js
codeallthethingz: should not.

tools/future_work_widget/app/js/future-work-widget.js
codeallthethingz: I don't think this will ever happen so removed entirely.

tools/future_work_widget/app/js/future-work-widget.js
codeallthethingz: this is for clicking on the tags in the table and adding them to the search.

codeallthethingz:
Fairly large refactor moving the validation code to Pydantic. 

Interesting notes, I had to add a shim library to get future annotations working with pydantic.  See the toml file for that. I added a note there that it can be removed when we hit Python 3.10.

`class FutureWorkRecord(BaseModel)` is the main annotated class that defines all the validation logic.

tools/future_work_widget/app/js/future-work-widget.js
```diff
@@ -0,0 +1,317 @@
...
+
+  input.dispatchEvent(new Event('input', { bubbles: true }));
+}
+
+
+function updateUrlSearchParam(searchTerm) {
+  const url = new URL(window.location);
+
+  if (searchTerm.trim()) {
+    url.searchParams.set('q', searchTerm.trim());
```
tristanls-tbp: **nitpick**: Unconventional double `.trim()` call.

tools/future_work_widget/app/js/future-work-widget.js
```diff
@@ -273,46 +268,45 @@ const FutureWorkWidget = {
...
+        if (className) element.classList.remove(className);
+      }, 1500);
+    };
 
+    try {
+      await navigator.clipboard.writeText(window.location.href);
+      setTemporaryState('✅', 'success');
+    } catch (err) {
+      console.error('Failed to copy URL to clipboard:', err);
+      setTemporaryState('❌');
```
tristanls-tbp, line 302: **praise**: Much clearer :)

tools/future_work_widget/validator.py
```diff
@@ -11,395 +11,476 @@
...
-        Args:
-            record: The record being transformed (modified in place)
-            file_path: Path to the source file for error reporting
-            errors: List to append validation errors to
-        """
-        for field in self.COMMA_SEPARATED_FIELDS:
-            if field in record:
-                field_value = str(record[field])
+        sanitized = nh3.clean(v).strip()
+        if sanitized not in allowed_values:
```
tristanls-tbp: **issue**: This code allows values to be specified dangerously and incorrectly. If this code changes, those invalid/dangerous values could leak onto the Internet.

**suggestion**: Check if input is in `allowed_values` before sanitizing it. Input should not be hostile.

tools/future_work_widget/validator.py
```diff
@@ -11,395 +11,476 @@
...
         rfc_patterns = [
             r"https://github\.com/thousandbrainsproject/tbp\.monty/.*",
             r"required",
             r"optional",
             r"not-required",
         ]
 
-        if not any(re.fullmatch(pattern, sanitized_value) for pattern in rfc_patterns):
+        if not any(re.fullmatch(pattern, value) for pattern in rfc_patterns):
             valid_options = "a GitHub URL, 'required', 'optional', or 'not-required'"
```
tristanls-tbp, line 342: **issue**: Shouldn't `required`, `optional`, `not-required` come from `info.allowed_values` ?

tools/future_work_widget/build.py
```diff
@@ -29,104 +46,94 @@ def build(
...
-            }
+        build_result = _validate_params(index_file, output_dir, docs_snippets_dir)
+        if build_result is not None:
+            return build_result
+
+        with open(index_file, encoding="utf-8") as f:
+            raw_data = json.load(f)
+
+        try:
+            index = FutureWorkIndex.model_validate(raw_data)
```
tristanls-tbp: **suggestion**: Use `.model_validate_json`

```
index = FutureWorkIndex.model_validate_json(f)
```

tools/future_work_widget/app/js/future-work-widget.js
codeallthethingz: I tried adding in I really tiny framework (alpine) but it still addded 45kb so I went back and just made this a bit cleaner instead.

tools/future_work_widget/validator.py
```diff
@@ -11,395 +11,476 @@
...
-        """Validate that all required fields are present and not empty."""
-        for field in self.REQUIRED_FIELDS:
-            if field not in record:
-                self._add_validation_error(
-                    f"Required field '{field}' is missing", file_path, field, errors
-                )
-                continue
+    @field_validator("contributor", mode="before")
+    @classmethod
+    def parse_contributor(cls, v: Any) -> list[str] | None:
```
tristanls-tbp: **issue**: I think this should be named `def validate_contributor(...)`.

tools/future_work_widget/cli.py
```diff
@@ -43,8 +48,8 @@ def main():
 
     result = build(index_file, output_dir, docs_snippets_dir)
 
-    logger.info(json.dumps(result, indent=2))
-    sys.exit(0 if result["success"] else 1)
+    logger.info(json.dumps(result.model_dump(exclude_none=True), indent=2))
```
tristanls-tbp: **suggestion**: `logger.info(result.model_dump_json(...))`

tools/future_work_widget/validator.py
```diff
@@ -11,395 +11,476 @@
...
+    file: str
+    line: int
+    field: str
+    level: str
+    title: str
+    annotation_level: str
+
+
+class FutureWorkIndex(RootModel):
+    root: list[dict[str, Any]]
```
tristanls-tbp: **issue**: I think this should be `root: list[FutureWorkRecord]`.

**note**: This would allow you to get rid of most of the `build` and `ResultValidator` code. I'm guessing something like:
```python
allowed_values = _load_validation_files(docs_snippets_dir)
try:
  index = FutureWorkIndex.model_validate_json(f, context={"allowed_values": allowed_values})
except PydanticValidationError as e:
  ...
else:
  data_file = output_dir / "data.json"
  with open(data_file, "w", encoding="utf-8") as f:
    index.model_dump_json(indent=2, ensure_ascii=False)
```

tools/future_work_widget/validator.py
```diff
@@ -11,395 +11,476 @@
...
-    def _validate_field_values(
-        self, record: dict[str, Any], file_path: str, errors: list[ValidationError]
+    def _add_error_detail(
+        self, message: str, file_path: str, field: str, errors: list[ErrorDetail]
     ) -> None:
-        """Validate field values against exact matches and regex patterns.
+        """Add an error detail to the errors list."""
+        errors.append(self._create_error_detail(message, file_path, field))
+
+    def _load_validation_files(self, docs_snippets_dir: Path) -> None:
```
tristanls-tbp: **issue**: Is there still, knowing what we now know about the implementation, a compelling reason to keep all of these validation files separate?

**note**: I assume the primary reason for the file to have the content is so that everything is in `docs` folder and the tool does not need to be updated when we edit allowed values.

**suggestion**: Consolidate on a single key/value file to load.

tools/future_work_widget/validator.py
codeallthethingz: In this case, no.  I initially had these values being loaded like the other ones, but it introduced a lot of complexity dealing with regex's in the .md files.  So I opted that special field values only get validated in the validator code so it made sense to have them all located right in this validation code.

tools/future_work_widget/build.py
codeallthethingz: You can't pass in a f, but you can read the bytes, and pass that in so that rust is doing the json parsing.

tools/future_work_widget/cli.py
codeallthethingz: oh, cool.

tools/future_work_widget/validator.py
codeallthethingz: Yes, well, I'm not sure about compelling but there are reasons.

When the build action runs it could output which snippet file you should edit if you want to add this skill/tag/whatever.

`
"message": "Value error, Invalid skills value 'Python'. Valid values are: content marketing, content-creation, css, github-actions, github-readme-sync-tool, html, interviewing, javascript, podcasting, python, research, s3, thousand-brains-theory, video-editing, writing. To add a new value, edit docs/snippets/future-work-skills.md",
`

These are also displayed in the docs here which I think is useful and would be more cumbersome if the file was consolidated. 

https://thousandbrainsproject.readme.io/v0.12-codeallthethingz-future-work-tool-and-front-matter/docs/future-work-widget-metadata

tools/future_work_widget/validator.py
codeallthethingz: **much nicer**

tools/future_work_widget/build.py
```diff
@@ -0,0 +1,163 @@
...
+    except (OSError, PermissionError) as e:
+        return BuildResult(
+            success=False,
+            error_message=f"Failed to create output directory {output_dir}: {e}",
+        )
+
+    return None
+
+
+def _return_error_result(
```
codeallthethingz, line 113: @tristanls-tbp when you run the entire set of records in one pydantic parse, to then get the filename out requires this ugly logic.  I need the filename for the github action annotation.

tools/future_work_widget/validator.py
goingmaymay: Off-topic: I would love to see pydantic used in the core project, not just the future work tool.

@tristanls-tbp Have you identified any places which could especially benefit from pydantic? So far, it seems to me that it could help with the config code.

tristanls-tbp: @mf-maymay, @jeremyshoemaker and I, will make a 3-day focused attempt on Nov 5-7, at refactoring all of Monty's configuration to use Hydra + Pydantic along the lines of https://github.com/ptomecek/pycon2025/blob/main/PyCon%202025.pptx

goingmaymay: Excellent. If I can help in any way, please let me know. I would be happy to review.

tools/future_work_widget/app/js/future-work-widget.js
tristanls-tbp: I've been thinking about this more. Since `data.json` never changes except during app deployment, your app deployment process should invalidate `data.json` cache within our infrastructure on each deployment. This is not needed.

**question**: Do you have a different scenario in mind than app deployment to invalidate the cache for here?

tools/future_work_widget/build.py
```diff
@@ -0,0 +1,163 @@
...
+            success=False,
+            error_message=f"Unexpected error during build: {e}",
+        )
+
+
+def _validate_params(
+    index_file: Path,
+    output_dir: Path,
+    docs_snippets_dir: Path,
+) -> BuildResult | None:
```
tristanls-tbp: **issue**: This is a confusing return signature. This function does not build anything, so `BuildResult` is an inappropriate result.

**suggestion**: `-> str | None` (where `str` is the error message). Then, in usage it is:
```python
        error_message = _validate_params(index_file, output_dir, docs_snippets_dir)
        if error_message is not None:
            return BuildResult(success=False, error_message=error_message)
```

tools/future_work_widget/validator.py
```diff
@@ -0,0 +1,371 @@
...
+        list[str] | None,
+        Field(
+            default=None,
+            description="GitHub usernames of contributors",
+            max_length=10,
+        ),
+    ]
+
+    @classmethod
+    def _get_allowed_values(
```
tristanls-tbp: **nitpick**: Noise word, prefer `def _allowed_values(`

tristanls-tbp (commented):
**praise**: I got to see the widget in action 🎊 . Good work, thank you for doing this. I'm looking forward to seeing it deployed.

Almost there 🙂.

```
(tbp.monty) me@scar_glamour ~/tbp/tbp.monty % pip install -e '.[future_work_widget_tool]'
...
(tbp.monty) me@scar_glamour ~/tbp/tbp.monty % python tools/future_work_widget/run_local.py
  File "tools/future_work_widget/run_local.py", line 22, in <module>
    from tools.github_readme_sync.index import generate_index  # noqa: E402
  File "/Users/me/tbp/tbp.monty/tools/github_readme_sync/index.py", line 19, in <module>
    from slugify import slugify
ModuleNotFoundError: No module named 'slugify'
```

**issue** The README should include `pip install -e '.[github_readme_sync_tool]'` as part of the Running Locally section.

codeallthethingz:
@tristanls-tbp incorporated those changes, feel free to merge if it passes review.

PR #436, opened 2025-08-22
docs: restructure "how monty works" section

vkakerbeck:
I restructures the how monty works section so that all the LM topics are grouped in one section and separated out more clearly.


codeallthethingz (approved):
LGTM.

PR #438, opened 2025-08-22
docs: Add documentation around reference frame transforms in Monty

vkakerbeck:
This PR adds documentation around the reference frames document we discussed in past research meetings into the documentation. I added a bit more structure & code references to it since then.


docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,56 @@
+---
+title: Reference Frame Transforms
```
codeallthethingz, line 2: suggestion: link here from the glossary - https://thousandbrainsproject.readme.io/v0.10-vkakerbeck-rf-transform-docs/docs/glossary#reference-frame

.vale/styles/config/vocabularies/TBP/accept.txt
```diff
@@ -193,4 +193,5 @@
 \bYoon\b
 \byoutube\b
 \bYouTube\b
-\bzsh\b
\ No newline at end of file
+\bzsh\b
+\bDiscretizing\b
```
codeallthethingz: suggestion: case insensitive

```suggestion
\b[Dd]iscretizing\b
```

docs/contributing/documentation.md
```diff
@@ -151,6 +151,9 @@ To check the links, [activate the conda environment](../how-to-use-monty/getting
 python -m tools.github_readme_sync.cli check docs
 ```
 
+> [!NOTE] 
+> See the [readme sync tool documentation](../../tools/github_readme_sync/README.md) for more details on how to use it and how to install the additional dependencies for it.
```
codeallthethingz: note: this was an issue in a separate PR so I assume that change will carry over here. Just highlighting in case it's lost during the merge.

docs/figures/how-monty-works/sensor_moves.gif
codeallthethingz, line 1: praise: this is so great it makes me want to learn powerpoint (almost)

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,56 @@
...
+---
+In a sensorimotor learning setup one naturally encounters several different reference frames in which information can be represented. Additionally, Monty has it's internal reference frames in which is learns models of objects. Those can get confusing to wrap your head around and keep track of so here is a brief overview of all the reference frames involved in a typical Monty setup.
```
codeallthethingz: todo: commas, typo.

```suggestion
In a sensorimotor learning setup, one naturally encounters several different reference frames in which information can be represented. Additionally, Monty has internal reference frames in which it learns models of objects. Those can get confusing to wrap your head around and keep track of, so here is a brief overview of all the reference frames involved in a typical Monty setup.
```

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,56 @@
...
+
+# Reference Frames in a Typical Monty Experiment
+![](../../figures/how-monty-works/reference_frames_overview.png)
+
+### 🟢 Object rel. World
+The orientation of the object in the world (unknown to Monty). For instance, if we use a simulator, this would be the configuration that specifies where the object is placed in the environment (like we do [here](https://github.com/thousandbrainsproject/tbp.monty/blob/4844ef17a4cadce455acb8d852fe3ed7038a298f/src/tbp/monty/frameworks/config_utils/make_dataset_configs.py#L229)). In the real world we don't have a specific origin but still, all objects are placed in a common reference frame with relative displacements between each other.
```
codeallthethingz: suggestion: unknown implies it can be learned.  Maybe 'withheld' which indicates its never sent to monty?



```suggestion
The orientation of the object in the world (withheld from Monty). For instance, if we use a simulator, this would be the configuration that specifies where the object is placed in the environment (like we do [here](https://github.com/thousandbrainsproject/tbp.monty/blob/4844ef17a4cadce455acb8d852fe3ed7038a298f/src/tbp/monty/frameworks/config_utils/make_dataset_configs.py#L229)). In the real world we don't have a specific origin but still, all objects are placed in a common reference frame with relative displacements between each other.
```

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,56 @@
...
+### 🟢⚫️ Feature rel. Object
+The pose of all the features on the object (also unknown to the Monty). For instance, if you use a 3D simulator, this could be defined in the object meshes.
```
codeallthethingz, line 12: suggestion: withheld


```suggestion
The pose of all the features on the object (also withheld from Monty). For instance, if you use a 3D simulator, this could be defined in the object meshes.
```

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,56 @@
...
+### 🟡 Sensor rel. World/Body
+The sensor’s location and orientation in the world can be calculated from proprioceptive information and motor efference copies. Basically, as the system moves it's sensors, it can use this information to update the locations of those sensors in a common reference frame. For the purposes of Monty it doesn't matter where the origin of this RF is (it could be the agent's body or an arbitrary location in the environment) but it matters that all sensor locations are represented in this common RF.
```
codeallthethingz: todo: apostrophe.

```suggestion
The sensor’s location and orientation in the world can be calculated from proprioceptive information and motor efference copies. Basically, as the system moves its sensors, it can use this information to update the locations of those sensors in a common reference frame. For the purposes of Monty it doesn't matter where the origin of this RF is (it could be the agent's body or an arbitrary location in the environment) but it matters that all sensor locations are represented in this common RF.
```

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,56 @@
...
+### 🟠 Feature rel. Sensor
+This is the feature orientation relative to the sensor. For instance in Monty, we often use the surface normal and curvature direction to define the sensed pose, which are extracted from the depth image of the sensor ([code](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/models/sensor_modules.py#L161-L167)).
```
codeallthethingz: question: should this be "the feature's orientation"?

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,56 @@
...
+### 🟢⚪️ Feature rel. World/Body
+[bright green]
+The estimated orientation of the sensed feature in the world (sensor_rel_world * feature_rel_sensor). In Monty this currently happens to the depth values in [DepthTo3DLocations](https://github.com/thousandbrainsproject/tbp.monty/blob/4844ef17a4cadce455acb8d852fe3ed7038a298f/src/tbp/monty/frameworks/environment_utils/transforms.py#L220) while the surface normal and curvature extraction then happens in the SM, but in the end you get the same results.
```
codeallthethingz: todo: comma 

```suggestion
The estimated orientation of the sensed feature in the world (sensor_rel_world * feature_rel_sensor). In Monty, this currently happens to the depth values in [DepthTo3DLocations](https://github.com/thousandbrainsproject/tbp.monty/blob/4844ef17a4cadce455acb8d852fe3ed7038a298f/src/tbp/monty/frameworks/environment_utils/transforms.py#L220) while the surface normal and curvature extraction then happens in the SM, but in the end you get the same results.
```

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,56 @@
...
+### 🔵 Object rel. Model
+This is the hypothesized orientation of the learned model relative to the object (in the world). This orientation needs to be inferred by the LM based on its sensory inputs. There are usually multiple hypotheses which Monty learning modules keeps track of in [`self.possible_poses`](https://github.com/thousandbrainsproject/tbp.monty/blob/4844ef17a4cadce455acb8d852fe3ed7038a298f/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L227)
+### 🔵⚫️ Feature rel. Model
+This is the rotation of the currently sensed feature relative to the object model in the LM (feature_rel_world * object_rel_model, see code [here](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py#L141-L142)). The learning module uses it's pose hypothesis to transform the feature relative to the world into it's objects reference frame so that it can recognize the object in any orientation and location in the world, independent of where it was learned.
```
codeallthethingz: todo: apostrophes

```suggestion
This is the rotation of the currently sensed feature relative to the object model in the LM (feature_rel_world * object_rel_model, see code [here](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py#L141-L142)). The learning module uses its pose hypothesis to transform the feature relative to the world into its object's reference frame so that it can recognize the object in any orientation and location in the world, independent of where it was learned. 
```

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,56 @@
...
+
+If the object is in a different rotation than how it was learned, all the features on the object will be sensed in different orientations and location in the world. This needs to be compensated by the light blue projection of the pose hypothesis.
+
+![](../../figures/how-monty-works/object_moves.gif)
+
+What is not shown in this image is that the same rotation transform is also applied to the movement vector (displacement) that is applied to update the hypothesized location on the object.
+
+# Reference Frame Transforms for Voting
+Voting in Monty happens in object space. We directly translate between the object’s RF of the sending LM to the object RF of the receiving LM. This relies on the assumption that both LMs learned the object at the same time, and hence their reference frames line up in orientation and displacement (since we receive features rel. world, which will automatically line up if the LMs learn the object at the same time). Otherwise, we could store one displacement between their reference frames and apply that in addition.
```
codeallthethingz: suggestion: slight rephrasing

```suggestion
Voting in Monty happens in object space. We directly translate between the object’s RF of the sending LM to the object RF of the receiving LM. This relies on the assumption that both LMs learned the object at the same time, and hence their reference frames line up in orientation and displacement (since we receive features rel. world, which will automatically line up if the LMs learn the object at the same time). Otherwise, we could store one displacement between their reference frames and apply that as well.
```

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,56 @@
...
+
+The two LMs receive input from two sensors that sense different locations and orientations in space. They receive that as a pose in a common coordinate system (rel. world in the image below). Since the sensors are at different locations on the object and our hypotheses are “locations of sensor rel. model” we can’t just vote on the hypotheses directly but have to account for the relative sensor displacement. So the sensor that is sensing the handle of the cup needs to incorporate the offset to the sensor that senses the rim to be able to use its hypotheses.
+
+This offset can be easily calculated from the difference of the two LM’s inputs as those poses are in a common coordinate system. The sending LM attaches its sensed pose in the world to the vote message it sends out (along with its hypotheses about the locations on the mug) and the receiving LM compares it with its own sensed pose and applies the difference to the vote hypotheses.
```
codeallthethingz: todo: commas

```suggestion
This offset can be easily calculated from the difference of the two LM’s inputs, as those poses are in a common coordinate system. The sending LM attaches its sensed pose in the world to the vote message it sends out (along with its hypotheses about the locations on the mug), and the receiving LM compares it with its own sensed pose and applies the difference to the vote hypotheses.
```

docs/overview/further-reading.md
```diff
@@ -303,4 +303,4 @@ Below is a great introductory video of Jeff Hawkins presenting an overview of ou
   "href": "https://www.youtube.com/watch?v=owRC8sLSb64",
   "typeOfEmbed": "youtube"
 }
-[/block]
\ No newline at end of file
+[/block]
```
codeallthethingz: suggestion: revert this file as it was not changed.

codeallthethingz (changes requested):
This is such a great reference, the animations are 🔥 !

A few minor typos / grammar suggestions.

docs/how-monty-works/monty/reference-frame-transformations.md
vkakerbeck: hm technically Monty can infer this (it just doesn't know the ground truth). Withheld sounds kind of like we always have access to this information but usually (when you are not running controlled experiments in a simulator) we also don't know about this info (if it was available there would be no application for Monty) and so it is not really withheld but just not available. Do you feel strongly about this? (mostly asking bc I would also have to update the animations & image with that wording but if it is confusing how it is I can do that)

codeallthethingz: ah, ok, that was my misunderstanding, please disregard.

vkakerbeck:
@codeallthethingz I added some more visualizations and text based on our discussion yesterday. Lmk if you think this is helpful of if anything is missing.

@nielsleadholm this is ready for review now.

Pretty view is here: https://thousandbrainsproject.readme.io/v0.10-vkakerbeck-rf-transform-docs/docs/reference-frame-transforms

codeallthethingz:
> @codeallthethingz I added some more visualizations and text based on our discussion yesterday. Lmk if you think this is helpful of if anything is missing.

looks good! 👍

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,83 @@
...
+![](../../figures/how-monty-works/vote_loc_transform.png#width=400px)
+
+The displaced votes are overlaid onto the existing location hypotheses of the receiving LM (orange dots). Only a few points will be consistent with the incoming votes (in this case just one, circled in green). When the right LM sends votes to the left one, it's location hypotheses will have to be displaced in the opposite direction (towards the left). 
+
+## Example: Sensors Sense Different Locations and Orientations
+Lastly, if the two LMs also sense different orientations in the world, this needs to be applied to the vote location displacement. This is the example animation shown above where the incoming vote locations are both shifted towards the right on the mug and rotated. 
+
+![](../../figures/how-monty-works/vote_loc_rot_transform.png#width=400px)
+
+After this displacement is applied, the vote locations (red dots) can again be overlaid onto the existing hypothesis space (orange dots). In this example, the right LM received a pose and features that were consistent with the top or bottom of the handle. After incorporating the votes, the top of the handle is the most likely location (circled in green).
```
tristanls: **issue (non-blocking)**: I've tried to understand this for a while but still do not understand why this makes sense. The graphic shows me that one learning module thinks it has found a point on the cup's rim, and it shows the transform that takes that point from the rim and maps it onto the handle.

**question (non-blocking)**: Why is it valid to map the cup's rim onto any other part of the cup? It is not clear to me what the subject of the vote is. What is being voted on? The source learning module sent data associated with "rim of the cup" in its model. Why is it valid for the destination learning module to map any of those points onto the "cup handle" in its model? Are the votes only communicating "my model exists at these coordinates" and nothing more?

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,83 @@
+---
+title: Reference Frame Transforms
+---
+In a sensorimotor learning setup, one naturally encounters several different reference frames in which information can be represented. Additionally, Monty has internal reference frames in which it learns models of objects. Those can get confusing to wrap your head around and keep track of, so here is a brief overview of all the reference frames involved in a typical Monty setup.
+
+# Reference Frames in a Typical Monty Experiment
+![](../../figures/how-monty-works/reference_frames_overview.png)
+
+### 🟢 Object rel. World
+The orientation of the object in the world (unknown to Monty). For instance, if we use a simulator, this would be the configuration that specifies where the object is placed in the environment (like we do [here](https://github.com/thousandbrainsproject/tbp.monty/blob/4844ef17a4cadce455acb8d852fe3ed7038a298f/src/tbp/monty/frameworks/config_utils/make_dataset_configs.py#L229)). In the real world we don't have a specific origin but still, all objects are placed in a common reference frame with relative displacements between each other.
```
nielsleadholm:
```suggestion
The orientation of the object in the world (unknown to Monty). For instance, if we use a simulator, this would be the configuration that specifies where the object is placed in the environment (like we do [here](https://github.com/thousandbrainsproject/tbp.monty/blob/4844ef17a4cadce455acb8d852fe3ed7038a298f/src/tbp/monty/frameworks/config_utils/make_dataset_configs.py#L229)). In the real world we don't have a specific origin, but all objects are still in a common reference frame with relative displacements between each other.
```

docs/how-monty-works/monty/reference-frame-transformations.md
scottcanoe: Answering since I'm writing a reply to a community question right now. @vkakerbeck let me know if this is correct.

My understanding is that the voting is this:
- Column A says: "My assumption/hypothesis is that I'm on the rim of this cup, which is oriented in a particular way."
- Column A thinks: "If my hypothesis is correct, then moving my sensor over to where column B's sensor is _would_ yield an observation of the handle."
- Columns A/B check whether this is actually true using column B's observation as a proxy for what would happen if column A looked at column B's location.

vkakerbeck: Yes, that is correct how you describe it @scottcanoe Maybe one thing to highlight is that the vote doesn't contain any kind of labels like "rim" or "handle", I just refer to those in the text to make it easier to identify the different locations I am talking about. 
The votes themselves only contain locations. Those locations are unique to a specific object (in Monty, the vote dictionary identifies the object ID that the location votes are associated with). 
We work on the assumption that the object was learned in both columns at the same time and that therefore the RFs correspond to each other. So if you look at example 1, where both sensors are sensing the same location in the world, no displacement needs to be applied, and the location votes from LM1 can be directly applied to LM2's RF. If there is a sensor displacement, then the voted locations need to be shifted by that amount. 
If we didn't do this, then the LM that currently thinks it is on the rim would also add evidence for the locations on the rim to LM2, but that LM is sensing a different location on the mug, so it should not expect to be on the rim. Again, "rim" and "handle" are never explicitly part of the votes, but the locations that are sent to LM2 are displaced such that they end up on those locations on the model in this example. Does that make sense?

tristanls: Thank you both.

FWIW, I didn't understand @scottcanoe's description, but I understood @vkakerbeck's "votes only contain locations." I think my sticking point in @scottcanoe's description was "observation", which I interpret differently from "location," since I think of observations as features at poses.

So, translating for my understanding: 

Setup: 
- LM 1 perceives object P at location A
- LM 2 perceives object P at location B

Dynamics:
Learning Modules only vote on locations. When a Learning Module sends out a vote, the vote contains information that an object ID exists at a location. So, in my setup, LM 1 would send out a vote "object P @ location A". When another Learning Module receives a vote, the vote is transformed so that the receiving Learning Module gets the information that the object ID should exist at the location it is sensing. So, in my setup, LM 2 would receive a vote "object P @ location A" transformed into "object P @ location B". And, at the same time, LM 2 sends out a vote "object P @ location B", which LM 1 would receive transformed into "object P @ location A". So, LM 1 would get reinforcement that object P is at location A from itself and from the vote of LM 2. Also, LM 2 would get reinforcement that object P is at location B from itself and from the LM 1 vote.

The above makes sense to me.

Follow-up question.

Looking at the code, along with the location, we send a "pose" in `morphological_features` (although it seems to me this should maybe be "orientation", since ["pose" == "location" + "orientation"](https://thousandbrainsproject.readme.io/docs/glossary#pose)):

https://github.com/thousandbrainsproject/tbp.monty/blob/fbed64f11bd70eda31fe3e79cb5d945a55ae519d/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L373-L390

I notice that we _don't use_ the orientation:

https://github.com/thousandbrainsproject/tbp.monty/blob/fbed64f11bd70eda31fe3e79cb5d945a55ae519d/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L864-L866

So, in practice, we only vote on locations, but it is a bit confusing seeing orientations sent out in the votes. Our docs state that we don't do it, while the code does.

I noticed we say "Currently we do not send out pose hypotheses when we are voting" in [Use Pose for Voting improvement doc](https://thousandbrainsproject.readme.io/docs/use-pose-for-voting). Aren't we, though?

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,83 @@
...
+### 🟢⚫️ Feature rel. Object
+The pose of all the features on the object (also unknown to the Monty). For instance, if you use a 3D simulator, this could be defined in the object meshes.
+### 🟡 Sensor rel. World/Body
+The sensor’s location and orientation in the world can be calculated from proprioceptive information and motor efference copies. Basically, as the system moves its sensors, it can use this information to update the locations of those sensors in a common reference frame. For the purposes of Monty it doesn't matter where the origin of this RF is (it could be the agent's body or an arbitrary location in the environment) but it matters that all sensor locations are represented in this common RF.
```
nielsleadholm:
```suggestion
The sensor’s location and orientation in the world can be estimated from proprioceptive information and motor efference copies. Basically, as the system moves its sensors, it can use this information to update the locations of those sensors in a common reference frame. For the purposes of Monty it doesn't matter where the origin of this RF is (it could be the agent's body or an arbitrary location in the environment) but it matters that all sensor locations are represented in this common RF. Lastly, it's worth noting that in simulation environments, this pose estimation can be perfect / error-free if desired. 
```

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,83 @@
...
+### 🟠 Feature rel. Sensor
+This is the feature's orientation relative to the sensor. For instance in Monty, we often use the surface normal and curvature direction to define the sensed pose, which are extracted from the depth image of the sensor ([code](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/models/sensor_modules.py#L161-L167)).
+### 🟢⚪️ Feature rel. World/Body
+[bright green]
```
nielsleadholm:
```suggestion
```

nielsleadholm: Not sure if this was meant to be here or was meant to change some text? At the moment it's just rendered as normal text.

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,83 @@
...
+The estimated orientation of the sensed feature in the world (sensor_rel_world * feature_rel_sensor). In Monty, this currently happens to the depth values in [DepthTo3DLocations](https://github.com/thousandbrainsproject/tbp.monty/blob/4844ef17a4cadce455acb8d852fe3ed7038a298f/src/tbp/monty/frameworks/environment_utils/transforms.py#L220) while the surface normal and curvature extraction then happens in the SM, but in the end you get the same results.
```
nielsleadholm, line 18:
```suggestion
The estimated orientation of the sensed feature in the world (sensor_rel_world * feature_rel_sensor). In Monty, the depth values are currently processed in [DepthTo3DLocations](https://github.com/thousandbrainsproject/tbp.monty/blob/4844ef17a4cadce455acb8d852fe3ed7038a298f/src/tbp/monty/frameworks/environment_utils/transforms.py#L220) while the surface normal and curvature extraction then happens in the SM, but in the end you get the same results.
```

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,83 @@
...
+### 🔵 Object rel. Model
+This is the hypothesized orientation of the learned model relative to the object (in the world). This orientation needs to be inferred by the LM based on its sensory inputs. There are usually multiple hypotheses which Monty learning modules keeps track of in [`self.possible_poses`](https://github.com/thousandbrainsproject/tbp.monty/blob/4844ef17a4cadce455acb8d852fe3ed7038a298f/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L227)
+### 🔵⚫️ Feature rel. Model
+This is the rotation of the currently sensed feature relative to the object model in the LM (feature_rel_world * object_rel_model, see code [here](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py#L141-L142)). The learning module uses its pose hypothesis to transform the feature relative to the world into its object's reference frame so that it can recognize the object in any orientation and location in the world, independent of where it was learned. 
+
```
nielsleadholm, line 28: **suggestion:** This might be a good place to link to the DMC paper, and note that the Methods section contains a mathematical description of these same transformations.

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,83 @@
+---
+title: Reference Frame Transforms
+---
+In a sensorimotor learning setup, one naturally encounters several different reference frames in which information can be represented. Additionally, Monty has internal reference frames in which it learns models of objects. Those can get confusing to wrap your head around and keep track of, so here is a brief overview of all the reference frames involved in a typical Monty setup.
+
+# Reference Frames in a Typical Monty Experiment
+![](../../figures/how-monty-works/reference_frames_overview.png)
+
+### 🟢 Object rel. World
```
nielsleadholm: **suggestion (non-blocking):** This might be too much work to justify, but you could make these colors a bit more aligned with our general style-guide. E.g. the processed feature could be blue (feed-forward), while the current blue elements could be a shade of golden (for thalamic transformations). The SM and LM boxes don't necessarily need any color, although you could make the bottom of the SM box square to align with the way it is represented elsewhere.

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,83 @@
...
+
+This is one of the key definitions of the [CMP](../observations-transforms-sensor-modules.md#cmp-and-the-state-class): The pose sent out from all SMs is in a common reference frame. In Monty, we use the [DepthTo3DLocations](../../../src/tbp/monty/frameworks/environment_utils/transforms.py) transform for this calculation and report locations and orientations in an (arbitrary) world reference frame.
+
+![](../../figures/how-monty-works/sensor_moves.gif)
+
+## Applying the Pose Hypothesis as the Object Rotates
+
+If the object changes its pose in the world, the pose hypothesis in the LM comes into play. The LM has a hypothesis on how the object is rotated in the world relative to its model of the object. This is essentially the rotation that needs to be applied to rotate the incoming features and movements into the model’s reference frame.
+
+The plot below only shows one pose hypothesis, but in practice, Monty has many of them, and it needs to infer this pose from what it is sensing (dark green orientations are unknown).
```
nielsleadholm: **suggestion:** I would expand this with a sentence or two to emphasize that in the animation shown, the hypothesis is correct (and updated quickly to remain correct), such that the sensed pose is stable. However, having the correct object pose is dependent on inference.

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,83 @@
...
+
+What is not shown in the visualizations above is that the same rotation transform is also applied to the movement vector (displacement) of the sensor. In Monty we calculate how much the sensor has moved in the world by taking the difference between two successive location and orientation inputs to the LM ([code](https://github.com/thousandbrainsproject/tbp.monty/blob/a408bf6063852323b98e009da5e1373d097beb73/src/tbp/monty/frameworks/models/graph_matching.py#L1052)). The sensor movement is applied to update the hypothesized location on the object. 
+
+Before we can apply the movement to update our hypothesized location on the object, the movement needs to be transformed from a movement in the world to movement in the object's reference frame. To do this, the hypothesized object orientation (light blue) is applied, the same way it is applied to the incoming features. So for example, if object is rotated by 90 degrees to the left and the sensor moved right to left on the object (like in the animation shown below), then the orientation transform will update the location on the object model (red dot) to change from bottom to top of the object.
```
nielsleadholm: **note (non-blocking):** Re. colors again, I think it would make sense to use pink rather than red, consistent with the heterarchy paper.

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Reference Frame Transforms for Voting
+Voting in Monty happens in object space. We directly translate between the object’s RF of the sending LM to the object RF of the receiving LM. This relies on the assumption that both LMs learned the object at the same time, and hence their reference frames line up in orientation and displacement (since we receive features rel. world, which will automatically line up if the LMs learn the object at the same time). Otherwise, we could store one displacement between their reference frames and apply that as well.
+
+The two LMs receive input from two sensors that sense different locations and orientations in space. They receive that as a pose in a common coordinate system (rel. world in the image below). Since the sensors are at different locations on the object and our hypotheses are “locations of sensor rel. model” we can’t just vote on the hypotheses directly but have to account for the relative sensor displacement. So the sensor that is sensing the handle of the cup needs to incorporate the offset to the sensor that senses the rim to be able to use its hypotheses.
```
nielsleadholm:
```suggestion
The two LMs receive input from two sensors that sense different locations and orientations in space. They receive that as a pose in a common coordinate system (rel. world in the image below). Since the sensors are at different locations on the object and our hypotheses are “locations of sensor rel. model” we can’t just vote on the hypotheses directly, but have to account for the relative sensor displacement. For example, the sensor that is sensing the handle of the cup needs to incorporate the offset to the sensor that senses the rim to be able to use its hypotheses.
```

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -0,0 +1,83 @@
...
+
+Check out our evidence LM documentation for [more details on voting](../learning-module/evidence-based-learning-module.md#voting-with-evidence).
+
+## Example: Both Sensors Sense Same Location
+The simplest example is if both learning modules receive input from the same location in the world (i.e. the two sensors "look" at the same point). In this case, no transformation needs to be applied to the incoming votes and they can directly be overlaid onto the receiving LM's reference frame.
+![](../../figures/how-monty-works/vote_same_loc.png#width=400px)
+
+However, voting in this case does not help reduce ambiguity as both LMs receive the same input.
+
+## Example: Sensors Sense Different Locations
```
nielsleadholm, line 78: **suggestion:** For "if you are over there, and you think you are at this location on the mug, then I should be x amount to the right of that", I think it would be helpful to add an "A" and a "B" or the like to the diagrams, and replace here/there/that with these specific labels.

docs/how-monty-works/monty/reference-frame-transformations.md
nielsleadholm: Here's my understanding, which might also address some of the points you've raised @tristanls 

For voting, in principle the receiving LM should transform the incoming displacement in world coordinates by its hypothesized rotation of the _object_. This would be a different transforming rotation for each receiving hypothesis, as we currently do for displacements that occur due to movement. The main difference between voting and movements is that the displacement is in the opposite direction / has a negative sign, because it gives "where you should be given I the sending LM am somewhere else", rather than "where you should have moved to, given you were somewhere else a moment ago". 

If we don't do the above rotation, consider two location hypotheses on the same location on the rim, but one where the cup is hypothesized to be upright, and one where it is hypothesized to be upside down. Without performing a rotation, these hypotheses will be at the exact same location in the receiving LM's reference frame after the displacement. However, based on the movement in the world, and the position of the cup in the world, only one of these could be consistent with the true end location (e.g. one on the handle, and one now off of the cup). 

I don't think it makes sense to do any transform based on the sensed poses of the features, as this says nothing about the pose of the object, and since we are not voting on local features/poses, but just location. Similarly, there is no reason that the local pose on the rim would tell us anything about the local pose on the handle. These could be totally different, depending on the shape of the object.

However, like Tristan says, I don't think we do any actual rotating at the moment. I think we discussed this a few months ago @vkakerbeck when I was trying to formalize the algorithm for DMC, and we decided to leave it for now.

nielsleadholm (commented):
**praise:** Beautiful animations and great descriptions.

Just a couple minor comments, and then adding to the discussion around voting.

docs/how-monty-works/monty/reference-frame-transformations.md
vkakerbeck: @tristanls how you describe it all sounds correct. One thing to highlight is that we don't say "the object exist at this location (in the world)" but we say "I am on this location on the object" (object centric).
You are right that we currently don't use the sent out pose_vectors. This has been a future work item for a long time but we just haven't gotten around to it (https://thousandbrainsproject.readme.io/docs/use-pose-for-voting)

@nielsleadholm the poses we send out with the vote correspond to the pose hypothesis of the whole object for that location, not the local feature sensed there, so I do think we can make use of this during voting (not by applying it to anything but by comparing to the pose hypothesis in the other LM).

I agree it is a bit confusing to send out rotation votes if we don't use them. I think I wrote this code and didn't think that I would get side tracked with other things for such a long time :D I also think I added the placeholder partially bc the State class requires pose_vectors to be defined. It's not a great reason though.

nielsleadholm: In that case I think it would be worth updating the description and diagram, which current implies the orientation is used.

Re. longer-term, I think the advantage of applying the rotation is that it would enable similar (but not exactly the same) rotations, or symmetric rotations, to correctly update the location in a similar way. But yeah this is something we can revisit later.

vkakerbeck: No, the diagram shows applying the rotation difference between the two sensory inputs, which we definitely do do here: https://github.com/thousandbrainsproject/tbp.monty/blob/a408bf6063852323b98e009da5e1373d097beb73/src/tbp/monty/frameworks/models/evidence_matching/model.py#L77 (what I also link in this doc)

What I was talking about in my previous comment are the rotation hypotheses of the sending LM.

docs/how-monty-works/monty/reference-frame-transformations.md
vkakerbeck: That would be a lot of work (updating colors in every frame of the animations). Do you mind if we leave them as they are for now? I tried being kind of consistent (using the blue and the different shades of it for the LM and its inputs and outputs) but since this figure includes the thalamus in the LM box it doesn't translate directly to how we use the colors in the neuroscience figures.

docs/how-monty-works/monty/reference-frame-transformations.md
vkakerbeck: oh, right, that was accidentally left in there

docs/how-monty-works/monty/reference-frame-transformations.md
vkakerbeck: Here I was specifically talking about the transformation from rel. sensor to rel. world, not the depth value processing.

docs/how-monty-works/monty/reference-frame-transformations.md
vkakerbeck: You are right, that would have been better... I'll note it down as a change to make in a batch of those animation updates

docs/how-monty-works/monty/reference-frame-transformations.md
vkakerbeck: sorry! After a huddle Niels and I clarified some points and I actually got part of this wrong. We actually apply to pose hypotheses here: https://github.com/thousandbrainsproject/tbp.monty/blob/a408bf6063852323b98e009da5e1373d097beb73/src/tbp/monty/frameworks/models/evidence_matching/model.py#L106 I will make another voting example that shows voting when the object orientation is different from it's learned rotation to visualize this.
Also, the figure it example 3 is wrong and shouldn't show the point cloud rotate, only displace. I will update that.
The thing that we actually don't use is the `sensor_rotation_disp` (we rotate the morphological features by it here: https://github.com/thousandbrainsproject/tbp.monty/blob/a408bf6063852323b98e009da5e1373d097beb73/src/tbp/monty/frameworks/models/evidence_matching/model.py#L109 but when we receive votes we don't look at morphological features. I think Niels is right that it also wouldn't make sense to look at the morphological features so we could remove that item from future work and also remove the rotation of morphological features line. I will double check on that tomorrow again.

vkakerbeck:
@nielsleadholm I updated the figures with the different colors like we discussed. I also added a new example on voting to show how the rotation hypothesis is applied to the sensor displacement. Let me know what you think.

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -70,14 +76,16 @@ The simplest example is if both learning modules receive input from the same loc
...
 
-## Example: Sensors Sense Different Locations and Orientations
-Lastly, if the two LMs also sense different orientations in the world, this needs to be applied to the vote location displacement. This is the example animation shown above where the incoming vote locations are both shifted towards the right on the mug and rotated. 
+When voting, the sending LM does not only send it's location hypotheses but also the [hypothesized pose for each of those locations](https://github.com/thousandbrainsproject/tbp.monty/blob/87159d54e9fbef7ca602b5d4a3940be7f8f65e0a/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L380). The receiving LM then used these poses to [rotate the sensor displacement](https://github.com/thousandbrainsproject/tbp.monty/blob/87159d54e9fbef7ca602b5d4a3940be7f8f65e0a/src/tbp/monty/frameworks/models/evidence_matching/model.py#L106) (thick green arrow). This is similar to how sensor movement in the world needs to be rotated into the model's reference frame.
```
nielsleadholm:
```suggestion
When voting, the sending LM does not only send it's location hypotheses but also the [hypothesized pose of the model, where there will be one such pose for each location hypothesis sent](https://github.com/thousandbrainsproject/tbp.monty/blob/87159d54e9fbef7ca602b5d4a3940be7f8f65e0a/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L380). The receiving LM then uses these poses to [rotate the sensor displacement](https://github.com/thousandbrainsproject/tbp.monty/blob/87159d54e9fbef7ca602b5d4a3940be7f8f65e0a/src/tbp/monty/frameworks/models/evidence_matching/model.py#L106) (thick green arrow). This is similar to how sensor movement in the world needs to be rotated into the model's reference frame.
```

docs/figures/how-monty-works/reference_frames_overview.png
nielsleadholm, line 1: **question:** Not sure how much of an issue it is (@codeallthethingz ), but some of these are PNGs with transparent backgrounds, while some have white backgrounds. In my browser with darkmode, the transparent one's get filled in with black, which makes them hard to view. This may just happen in GitHub, but mentioning it in case it could be an issue for the Readme docs (which are dark in my browser).

docs/how-monty-works/monty/reference-frame-transformations.md
```diff
@@ -36,28 +40,30 @@ This is one of the key definitions of the [CMP](../observations-transforms-senso
...
+This offset can be easily calculated from the difference of the two LM’s inputs, as those poses are in a common coordinate system. The sending LM [attaches its sensed pose in the world to the vote message](https://github.com/thousandbrainsproject/tbp.monty/blob/a408bf6063852323b98e009da5e1373d097beb73/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L394) it sends out (along with its hypotheses about the locations on the mug), and the receiving LM [compares it with its own sensed pose](https://github.com/thousandbrainsproject/tbp.monty/blob/a408bf6063852323b98e009da5e1373d097beb73/src/tbp/monty/frameworks/models/evidence_matching/model.py#L77)(green arrow) and [applies the displacement to the vote hypotheses](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/models/evidence_matching/model.py#L106-L112) (shifting the pink dots).
```
nielsleadholm, line 66: **suggestion (non-blocking):** For the next figure, maybe just mention that the green "arrow" is a dot because there is no displacement?

docs/figures/how-monty-works/reference_frames_overview.png
vkakerbeck: oh good catch! @codeallthethingz did you have some tool for adding white background? Otherwise I'll reexport them

vkakerbeck: nvm, I just updated it :)

codeallthethingz: @nielsleadholm good catch! There is dark mode on our docs too so its helpful to always have white backgrounds on our images.  

@vkakerbeck I did have a tool, but I think it only existed as a part of the now deleted code that did the initial import from the PDFs.

PR #440, opened 2025-08-25
feat: `config_to_dict` harder

scottcanoe:
### Summary
`config_to_dict` is meant to recursively convert any dataclass instances in an experiment config to regular dictionaries. It's essentially `dataclasses.as_dict` except it can take a dictionary argument. However, its current implementation can "miss" dataclass instances located within lists or tuples, particularly when the argument to `config_to_dict` is a dictionary.

Here's an example:
```python
from dataclasses import dataclass
from tbp.monty.frameworks.utils.dataclass_utils import config_to_dict

@dataclass
class SomeConfig:
    name: str

config_a = SomeConfig(name="a")
config_b = SomeConfig(name="b")

# Dataclasses are in a dictionary. They'll be converted to dicts as expected.
configs_in_dict = {"configs": {"config_a": config_a, "config_b": config_b}}
print(config_to_dict(configs_in_dict))
# Prints: {'configs': {'config_a': {'name': 'a'}, 'config_b': {'name': 'b'}}}

# Dataclasses are in a list. They'll be missed by `config_to_dict`.
configs_in_list = {"configs": [config_a, config_b]}
print(config_to_dict(configs_in_list))
# Prints: {'configs': [SomeConfig(name='a'), SomeConfig(name='b')]}
```

Unlike `config_to_dict`, `dataclasses.asdict` _does_ handle dataclass instances stored in lists and tuples by via the recursive function `dataclasses._asdict_inner`. We can reuse that function's logic to gain the list/tuple functionality. Since `_asdict_inner` isn't public, I just copied it into `dataclass_utils` as `_config_to_dict_inner`. I had to do the same for `dataclasses._is_dataclass_instance` since `_asdict_inner` relies on it.

### Tests
- I added a suite of `config_to_dict` tests that cover the new dataclass conversion functionality.
- Most of the existing `config_to_dict` basically just tested `dataclasses.asdict` -- very few tests used dictionary input. The tests I added should improve on this situation. I didn't delete the old ones, though they probably could be removed.
- I also added tests for `is_dataclass_instance` and `is_config_like` (which now uses `is_dataclass_instance`).

### Note
I stumbled on this behavior by accident, but I think it explains why configs are always defined in dictionaries, even when lists or tuples might make more sense. For example, a typical monty config will have something like
```python
    sensor_module_configs=dict(
        sensor_module_0=patch_config,
        sensor_module_1=view_finder_config,
    )
```
One might think the names `sensor_module_0` and `sensor_module_1` are meaningful in some way, but they aren't. Those names are ignored during model set-up, and sensor modules will end up in a list anyways. The same goes for learning modules and their configs.

Personally, I'd prefer to avoid this confusion by avoiding dictionaries in this context. It'd be trivial to modify model set-up code to allow for SM/SM lists instead of dictionaries. Merging this PR would make that possible.


scottcanoe:
Tagging @hlee9212 since this involves configs and the multiple-handles confusion for SMs and LMs.

hlee9212:
Thanks @scottcanoe for the mention. 

This is just a high-level idea for the future, but after writing the RFC for use of Hydra package for config management, doing a short presentation on Pydantic for config validation (pre-Jeremy), and [short comment from Jeremy about setting environment may causing memory leaks](https://github.com/thousandbrainsproject/tbp.monty/pull/407#pullrequestreview-3072223671), my latest recommendation is to use [pydantic-settings](https://github.com/pydantic/pydantic-settings). Some capabilities that solves the above three aforementioned issues:

1. It can automatically dump configurations to dict/JSON.
2. Validation.
3. Support for dot-env. 

I admit that writing an RFC for the above is not high on my list of to-dos, however, I just wanted to share the idea publicly in case anyone has any feedback(s).

scottcanoe:
@hlee9212 pydantic-settings sounds good to me. I only recently started playing with OmegaConf, so I'm unfamiliar with all the bells and whistles you get with pydantic and hydra.

However, I did look into Hydra's instantiation scheme recently. Very nice. I'd love to move away from custom configs with keys like"sensor_module_class" + "sensor_module_args" vs the common "_target_" + "_args_". It's a simple pattern that would make development and customization quite a bit cleaner and simpler.

hlee9212:
Hey @scottcanoe - that's awesome!! And yes, I also like the `target` and `args` in Hydra as well. Just FYI, Hydra is built on top of `OmegaConf` -- so if you know OmegaConf, you are probably an expert in Hydra under-the-hood. 😅 I think @jeremyshoemaker said he used Hydra before, and like him, I only barely knew Hydra because it's just updating some YAML (very low learning curve to start using). I only learned that it was built on OmegaConf and had an actual syntax when I was digging into it more for the RFC. 

I think the possible benefits of using pydantic/pydantic-settings over YAML like hydra are:
1. We give up YAML to write configs in python code, which allows us to specify `numpy` array in args (can only do lists in hydra); there's probably some other data types that we can't easily pass via YAML. 
2. `pydantic` was built for validation of configs (i.e. making sure we can't specify invalid configs), and during the RFC process this turned out to be what we were actually more interested on back then.

Note: I don't think hydra and pydantic are mutually exclusive (e.g. use Hydra to specify, pydantic to validate). Though I think pydantic can do both, and it's very `dataclasses`-like (but better 😉).

tristanls:
I'll return to this pull request tomorrow, but I'll leave the suggestion below in case you'd like to consider it.

**suggestion**: 🤔, a while back, I created `class Dataclass(Protocol):` in https://github.com/thousandbrainsproject/tbp.monty/blob/a408bf6063852323b98e009da5e1373d097beb73/src/tbp/monty/frameworks/config_utils/config_args.py#L85-L93

If we were to add `@runtime_checkable` annotation to that, it would enable `isintance(something, Dataclass)` checks:

```python
from dataclasses import dataclass
from typing import Any, ClassVar, Dict

from typing_extensions import Protocol, runtime_checkable


@runtime_checkable
class Dataclass(Protocol):
    """A protocol for dataclasses to be used in type hints.

    The reason this exists is because dataclass.dataclass is not a valid type.
    """

    __dataclass_fields__: ClassVar[Dict[str, Any]]
    """Checking for presence of __dataclass_fields__ is a hack to check if a class is a
    dataclass."""

@dataclass
class MyDataclass:
    a: int

my_dataclass = MyDataclass(a=1)

print(isinstance(my_dataclass, Dataclass))  # prints True
```

scottcanoe:
@tristanls I'm down for that.

src/tbp/monty/frameworks/utils/dataclass_utils.py
```diff
@@ -162,19 +167,60 @@ def config_to_dict(config: DataclassInstance | Dict[str, Any]) -> Dict[str, Any]
     We want to convert configs composed of mixed dataclass and dict elements to
     pure dicts without dataclasses for backward compatibility.
 
-    TODO: Remove once all other configs are converted to dict only
+    Like `dataclasses.asdict` (and `dataclasses._asdict_inner`), objects that
+    are not or do not contain dataclass instances are deep-copied and returned.
 
     Args:
-        config: Config to convert to dict.
+        config: dict or dataclass instance to convert to dict.
```
tristanls, line 170: **nitpick**: "Dictionary or dataclass instance..."

tests/unit/dataclass_utils_test.py
```diff
@@ -47,6 +48,17 @@ class DeepNestedDataclass:
     field3: NestedDataclass
 
 
+@dataclasses.dataclass
```
tristanls, line 51: **nitpick**: `@dataclass` instead of `@dataclasses.dataclass`.

tests/unit/dataclass_utils_test.py
```diff
@@ -340,6 +352,252 @@ def test_dict_with_dataclass(self):
             },
         )
 
+    # -------------------------------------------------------------------------
+    # - Basic (not-nested)
+    # -------------------------------------------------------------------------
+
```
tristanls: **nitpick**: This warrants a different test case, e.g., `ConfigToDictNonNestedTest`. We don't need artisanal comments.

tests/unit/dataclass_utils_test.py
```diff
@@ -340,6 +352,252 @@ def test_dict_with_dataclass(self):
...
+        dict_result = dataclass_utils.config_to_dict(dict_config)
+        dataclass_result = dataclass_utils.config_to_dict(dataclass_config)
+        self.assertDictEqual(dict_result, expected)
+        self.assertIsInstance(dict_result["value"], FakeNamedTuple)
+        self.assertDictEqual(dataclass_result, expected)
+        self.assertIsInstance(dataclass_result["value"], FakeNamedTuple)
+
+    # -------------------------------------------------------------------------
+    # - Nested
+    # -------------------------------------------------------------------------
```
tristanls: **nitpick**: This warrants a different test case, e.g., `ConfigToDictNestedTest`. We don't need artisanal comments.

tristanls (commented):
**praise**: Thanks for fixing this!

It looks like I introduced `Dataclass(Protocol)` and the `DataclassInstance` import from typeshed.

Here's what I propose.

**suggestion**: Use the `Dataclass(Protocol)` from `config_args.py` by adding `@runtime_checkable` to it and moving it to `dataclass_utils.py`. (config args should import from utils, not the other way around). Then, you'll be able to remove `DataclassInstance` use and import and replace it with `Dataclass`. This will allow for removal of `_DATACLASS_FIELDS` as well as `is_dataclass_instance` method (and its tests). Instead, you should be able to `isinstance(thing, Dataclass)`.

**note**: Reason we want to use our `Dataclass` and not the one from typeshed is ours will have the `@runtime_checkable`.

tests/unit/dataclass_utils_test.py
tristanls: I see that stuff is everywhere in this file already, so never mind.

jeremyshoemaker: I'd argue that it should be changed while we're here, though I'm fine if we decide it's out of scope. `@dataclass` is such a common Python thing that using a qualified name for it is just line noise.

tests/unit/dataclass_utils_test.py
jeremyshoemaker: I agree. We should really break up a lot of these large test case classes to make them easier to run separately, as well as keep related tests together and not all jumbled in a huge test case class.

This style of comment is more common in C or C++ codebases, and not so much in Python, at least in my experience.

scottcanoe:
@tristanls `isinstance(obj, Dataclass)` evaluates to `True` for classes/types. I'm not sure if there's a way to make a runtime-checkable protocol that disallows type/class objects.

tristanls:
> @tristanls isinstance(obj, Dataclass) evaluates to True for classes/types. I'm not sure if there's a way to make a runtime-checkable protocol that disallows type/class objects.

@scottcanoe, this is also true for `is_dataclass_instance`, e.g.,:

```python
@dataclass
class MyDataclass:
    a: int

my_dataclass = MyDataclass(a=1)

print(hasattr(MyDataclass, "__dataclass_fields__"))  # True
print(hasattr(my_dataclass, "__dataclass_fields__"))  # True
```

scottcanoe:
@tristanls My implementation , which is the same as `dataclasses._is_dataclass_instance`, uses
```python
 return hasattr(type(obj), "__dataclass_fields__")
```
which works as expected/desired.

tristanls:
> Prefixing fields and methods with a `_` to indicate that they are private is just a convention. There's nothing in the language that makes that private, so I'm not sure I agree with the argument here for why we're copying code from the standard library. We're not being forced to copy it by the language.

@jeremyshoemaker, while this is only a convention, it is a convention that says "do not rely on this method being unchanged." I do not want anyone using Monty to rely on anything that starts with `_` for the same reason. Yes, the language allows us to ignore the convention, but by that argument, the language allows us to ignore architecture, software patterns, etc. So, I think vendoring is the appropriate thing here. An alternative to vendoring would be to make a pull request to (somewhere) and expose that method without the private `_` prefix. If it's stable, they shouldn't mind.

jeremyshoemaker:
> So, I think vendoring is the appropriate thing here. An alternative to vendoring would be to make a pull request to (somewhere) and expose that method without the private `_` prefix. If it's stable, they shouldn't mind.

I feel like it's different when we're vendoring part of the standard library. In this case it's probably fine, but that always seems like a boundary best not to cross.

tristanls (changes requested):
> I feel like it's different when we're vendoring part of the standard library. In this case it's probably fine, but that always seems like a boundary best not to cross.

@jeremyshoemaker, that's a fair point. Why can't the standard library follow the standard conventions 😢.

So, I went down the path of what pull request would I need to make. Here's something interesting from the official documentation:

Screenshot 2025-08-26 at 18 29 45

-- https://docs.python.org/3/library/dataclasses.html#dataclasses.is_dataclass

So, the implementation should be:
```python
def is_dataclass_instance(obj):
    return is_dataclass(obj) and not isinstance(obj, type)
```

Does this pass the tests @scottcanoe ?

jeremyshoemaker:
> So, I went down the path of what pull request would I need to make. Here's something interesting from the official documentation:

I had the feeling that we've had a similar discussion before, and it turns out, we have in [a slightly different context](https://github.com/thousandbrainsproject/tbp.monty/pull/342#discussion_r2164388467). In fact, when I saw the code in that PR, I was similarly confused about it not doing what we're expecting it to do, looked up the documentation, and seeing that code snippet just assumed we'd copied it before.

In that context though, it made sense to change the check to bail early when we had a type instead of an instance, but I don't think that will make sense here.

src/tbp/monty/frameworks/utils/dataclass_utils.py
```diff
@@ -28,6 +29,10 @@
 # Keeps track of the dataclass type in a serializable dataclass dict
 _DATACLASS_TYPE = "__dataclass_type__"
 
+# The name of an attribute on the class where Field objects are stored.
+# Primarily used here to check if a class is a dataclass.
+_DATACLASS_FIELDS = "__dataclass_fields__"
+
```
tristanls: **issue**: This is no longer used.

**suggestion**: Remove.

scottcanoe:
@tristanls and @jeremyshoemaker thanks for the feedback. I've made a couple of changes.

1. I'm using the implementation of `is_dataclass_instance` suggested by Jeremy/Tristan/the docs.
2. I broke up tests. The tests I added are now in `ConfigToDictBasicTest` and `ConfigToDictNestedTest`. All previous tests are still in `ConfigToDictTest`. Those are probably no longer needed, but I'm wary of deleting any tests.

As for vendoring code (and thanks for the terminology, I didn't know what that was called), I had originally just used the built-in functions `_config_to_dict_inner` and `_is_dataclass_instance`, but just wasn't sure if that was kosher or not, so what you see here is just what I settled on. I don't have any opinions or preferences on this matter.

src/tbp/monty/frameworks/utils/dataclass_utils.py
```diff
@@ -162,19 +163,60 @@ def config_to_dict(config: DataclassInstance | Dict[str, Any]) -> Dict[str, Any]
...
-        config: Config to convert to dict.
+        config: dict or dataclass instance to convert to dict.
 
     Returns:
         Pure dict version of config.
+
+    Raises:
+        TypeError: If the object is not a dict or dataclass instance
+    """
+    if isinstance(config, dict) or is_dataclass_instance(config):
```
tristanls: **issue**: this should be `if is_config_like(config):`.

tristanls (approved):
Thanks for going on this journey with us 😅.

tristanls:
**note**: @scottcanoe, while the below note is true, `sensor_module_configs` is the _only_ instance where the placeholder name is unused.

> ## Note
>
>I stumbled on this behavior by accident, but I think it explains why configs are always defined in dictionaries, even when lists or tuples might make more sense. For example, a typical monty config will have something like
>
>```
>sensor_module_configs=dict(
>    sensor_module_0=patch_config,
>    sensor_module_1=view_finder_config,
>)
>```
>
>One might think the names sensor_module_0 and sensor_module_1 are meaningful in some way, but they aren't. Those names are ignored during model set-up, and sensor modules will end up in a list anyways. The same goes for learning modules and their configs.
>
>Personally, I'd prefer to avoid this confusion by avoiding dictionaries in this context. It'd be trivial to modify model set-up code to allow for SM/SM lists instead of dictionaries. Merging this PR would make that possible.

I found that for learning modules, we override the learning module ID with the dictionary key:

https://github.com/thousandbrainsproject/tbp.monty/blob/1ba1563ca4b9eb91077e02f3ef657336ea92db53/src/tbp/monty/frameworks/experiments/monty_experiment.py#L121-L130

I do agree that we should probably avoid dictionaries in the case of sensor modules and learning modules config. This would require adding `learning_module_id` prop to `learning_module_args` (like sensor modules have).

PR #442, opened 2025-08-26
ci: install monty if failed to find monty cache

tristanls:
As activity in the repo increases, the odds of the cached Monty installation disappearing from the cache increase. When this happens, checks that assume a cached Monty installation exists fail as follows:

Screenshot 2025-08-26 at 12 46 00

This pull request extracts the Monty installation into an `install_monty` action and updates the checks to install Monty if the cache is not found.

### DIFF
--- a/src/tbp/monty/frameworks/config_utils/config_args.py
+++ b/src/tbp/monty/frameworks/config_utils/config_args.py
@@ -29,7 +29,6 @@ from typing import (
 import numpy as np
 import wandb
 from scipy.spatial.transform import Rotation
-from typing_extensions import runtime_checkable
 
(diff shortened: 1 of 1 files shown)


src/tbp/monty/frameworks/config_utils/config_args.py
jeremyshoemaker, line 1: This seems like an unrelated change that shouldn't be in this PR, unless I'm missing something.

tristanls: Thank you for catching this. I was sloppy with the code. I will remove this random change.

PR #444, opened 2025-08-27
refactor: update get_edge_index to use vector operations

### LINKED ISSUE #443: Update graph_utils.py get_edge_index()
Came across this [function](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/environment_utils/graph_utils.py#L14-L31) with a TODO: comment stating there must be an easier way.

```python
def get_edge_index(graph, previous_node, new_node):
    mask = (graph.edge_index[0] == previous_node) & (graph.edge_index[1] == new_node)
    edge_ids = mask.nonzero(as_tuple=False).view(-1)
    return edge_ids[0].item() if len(edge_ids) > 0 else None
```
Replacing the for loop with tensor operations gives a slight performance improvement in benchmark experiments.

Lvidal2santos:
closes #443 

# Summary
[get_edge_index](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/environment_utils/graph_utils.py#L14-L31) has a TODO: comment stating there must be an easier way to do this. 

# Changes
Use a boolean mask and nonzero vector operations to get all possible edges. We return the first one as in the original implementation. Returning all possible edges might come handy if we need to handle multigraphs

# Experiment
Experiments that use the DisplacementGraphLM should show an improvement. 
You can see the results for the randrot_noise_10distincobj_5lm_dist_agent experiment [here](https://api.wandb.ai/links/lucasvidaldossantos-developer/cph7mg7d)

### DIFF
--- a/src/tbp/monty/frameworks/environment_utils/graph_utils.py
+++ b/src/tbp/monty/frameworks/environment_utils/graph_utils.py
@@ -21,5 +21,5 @@ def get_edge_index(graph, previous_node, new_node):
         edge ID between the two nodes
     """
     mask = (graph.edge_index[0] == previous_node) & (graph.edge_index[1] == new_node)
-    edge_ids = mask.nonzero(as_tuple=False).view(-1)
+    edge_ids = mask.nonzero().view(-1)
     return edge_ids[0].item() if len(edge_ids) > 0 else None
--- a/src/tbp/monty/frameworks/environment_utils/graph_utils.py
+++ b/src/tbp/monty/frameworks/environment_utils/graph_utils.py
@@ -9,6 +9,7 @@
 # https://opensource.org/licenses/MIT.
 from typing import Optional
 
+
 def get_edge_index(graph, previous_node, new_node) ->  Optional[int]:
     """Get the edge index between two nodes in a graph.
 
--- a/src/tbp/monty/frameworks/environment_utils/graph_utils.py
+++ b/src/tbp/monty/frameworks/environment_utils/graph_utils.py
@@ -8,13 +8,9 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
-import numpy as np
-
 
 def get_edge_index(graph, previous_node, new_node):
     """Get the edge index between two nodes in a graph.
-
-    TODO: There must be an easier way to do this!
 
     Args:
         graph: torch_geometric.data graph
@@ -24,8 +20,6 @@ def get_edge_index(graph, previous_node, new_node):
     Returns:
         edge ID between the two nodes
     """
-    edges_of_node = np.where(graph.edge_index[0] == previous_node)[0]
-    for i in range(len(edges_of_node)):
-        possible_next_node = graph.edge_index[1][edges_of_node[i]]
-        if possible_next_node == new_node:
-            return edges_of_node[i]
+    mask = (graph.edge_index[0] == previous_node) & (graph.edge_index[1] == new_node)
+    edge_ids = mask.nonzero(as_tuple=False).view(-1)
+    return edge_ids[0].item() if len(edge_ids) > 0 else None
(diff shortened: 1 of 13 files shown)


tristanls:
Hi, the description looks promising 🙂.

**note**: You won't be able to update the benchmark tables as we run benchmarks on our own specific infrastructure. However, sharing what you've seen locally on your machine is helpful to set our expectations.

Lvidal2santos:
Great. I edited the description to include an experiment run report.
Let me know if I'm missing something. First contribution here

src/tbp/monty/frameworks/environment_utils/graph_utils.py
```diff
@@ -24,8 +20,6 @@ def get_edge_index(graph, previous_node, new_node):
     Returns:
         edge ID between the two nodes
     """
-    edges_of_node = np.where(graph.edge_index[0] == previous_node)[0]
-    for i in range(len(edges_of_node)):
-        possible_next_node = graph.edge_index[1][edges_of_node[i]]
-        if possible_next_node == new_node:
-            return edges_of_node[i]
+    mask = (graph.edge_index[0] == previous_node) & (graph.edge_index[1] == new_node)
+    edge_ids = mask.nonzero(as_tuple=False).view(-1)
```
jeremyshoemaker: From the [PyTorch docs](https://docs.pytorch.org/docs/1.11/generated/torch.nonzero.html), the default for `as_tuple` is False, so I don't think we need to add it here.
```suggestion
    edge_ids = mask.nonzero().view(-1)
```

src/tbp/monty/frameworks/environment_utils/graph_utils.py
```diff
@@ -24,8 +20,6 @@ def get_edge_index(graph, previous_node, new_node):
...
+    return edge_ids[0].item() if len(edge_ids) > 0 else None
```
jeremyshoemaker: I was going to suggest changing this to just `if edge_ids`, but it looks like Tensors don't behave enough like other containers to make it work.

Lvidal2santos: thanks for the feedback
What do you think about this?

```python
from typing import Optional

def get_edge_index(graph, previous_node, new_node) ->  Optional[int]:
    mask = (graph.edge_index[0] == previous_node) & (graph.edge_index[1] == new_node)
    if mask.any():
        return mask.nonzero().view(-1)[0].item()
```

`mask` is a boolean mask so we can use [mask.any()](https://docs.pytorch.org/docs/stable/generated/torch.any.html). It returns a boolean tensor that we can use in an if statement.

vkakerbeck:
Hi @Lvidal2santos 
thanks a lot of picking out this TODO and fixing it! You solution looks much better than the loop version I came up with. I don't see any functional issues with the change (it should still produce the same output). I just started the unit test runs to make sure those all still pass.

Regarding the benchmark experiment results, this change should not affect any of them since we currently use the `EvidenceGraphLM` for all our benchmarks. To test the timing impacts of this change, I would recommend starting with a config from our unit tests that uses the DisplacementGraphLM (like here: https://github.com/thousandbrainsproject/tbp.monty/blob/fbed64f11bd70eda31fe3e79cb5d945a55ae519d/tests/unit/graph_learning_test.py#L201) We have more configs with that DisplacementGraphLM in our monty_lab repository [here](https://github.com/thousandbrainsproject/monty_lab/blob/main/experiments/configs/graph_experiments.py), but those are quite old and may not work anymore with the current Monty version. In general, I don't think we have to check the impact on benchmarks for this PR. Thanks for making this nice wandb report, it is a great sanity check to make sure the accuracy and other benchmark metrics are unaffected.

src/tbp/monty/frameworks/environment_utils/graph_utils.py
jeremyshoemaker: My knowledge of Pytorch is rather limited still, but that seems reasonable to me. Though we should add a `return None` after the `if`. I don't think we actually enforce this with a lint at the moment ([RET503](https://docs.astral.sh/ruff/rules/implicit-return/)), but we probably should.

jeremyshoemaker (approved):

PR #446, opened 2025-08-28
refactor: delete unassigned statement

hlee9212:
Delete unassigned statement.

---
Below is a preservation of old description when change was:
```python
pn_evidence[np.logical_not(use_cd)] *= 2 # added "="
```
And how that affected benchmark performance. 

- @vkakerbeck I'll be re-running the benchmark experiments to see if this causes any differences in performance. 
- @nielsleadholm I don't think this affects learning, so I think just running inference is okay. 
- @jeremyshoemaker FYI if you want to update "pn" as "surface_normal". 

---
## Shorter Experiments with 10 Objects

| Experiment | Correct (%) | Num Match Steps | Rotation Error (degrees) | 
|------------| ------------ | ------------------ | ------------------------ | 
| base_config_10distinctobj_dist_agent | 98.57 (-1.43%) | 45 (+8) | 14.53 (-1.67) | 
| base_config_10distinctobj_surf_agent | 100 (no diff)| 34 (+6) | 18.27 (+4.26) |
| randrot_noise_10distinctobj_dist_agent | 98 (-2%) | 60 (+5) | 19.83 (-2.66) |
| randrot_noise_10distinctobj_dist_on_distm | 98 (-1%) | 59 (+19) | 12.00 (-0.15) |
| randrot_noise_10distinctobj_surf_agent | 100 (no diff) | 39 (+10) | 24.30 (+ 3.88) | 
| randrot_noise_10distinctobj_5lms_dist_agent | 100 (no diff) | 55 (-2) | 44.87 (-0.61) | 
| base_10simobj_surf_agent | 95 (+1.43%) | 78 (+8) | 15.44 (+1.46) |
| randrot_noise_10simobj_dist_agent | 81 (-3%) | 226 (-11) | 29.96 (-5.05) |
| randrot_noise_10simobj_surf_agent | 92 (-1%) | 163 (-9) | 26.06 (+5.12) | 
| randomrot_rawnoise_10distinctobj_surf_agent | 66 (+2%) | 17 (+1) | 115.07 (+ 8.56) | 
| base_10multi_distinctobj_dist_agent | 85 (+ 5.71%) | 31 (no diff) | 26.88 (+7.04) | 

Observations:
- Generally seems to improve rotation error for `_dist_agent` (except for the last experiment) but worsen for `_surf_agent`. 
- Generally seem to increase `num_match_steps`.
- Small accuracy drops. 

---
## Longer Experiments with all 77 YCB Objects

| Experiment | Correct (%) | Num Match Steps | Rotation Error (degrees) | 
|------------| ------------ | ------------------ | ------------------------ | 
| base_77obj_dist_agent | 92.64 (-0.87%) | 122 (+14 ) | 16.27 (+0.41) |
| base_77obj_surf_agent | 99.13 (+0.43%) | 61 (+5) | 11.40 (-1.54) |
| randrot_noise_77obj_dist_agent | 86.58 (-3.03%) | 174 (+22) | 35.15 (-2.32) |
| randrot_noise_77obj_surf_agent | 93.94 (no diff) | 133 (+18) | 34.14 (-0.66) |
| randrot_noise_77obj_5lms_dist_agent |  89.61 (+1.30%) | 83 (+13) | 52.74 (-5.51) |

---
Overall, I think randrot+noise+dist agent benefitted across the board in terms of rotation error.


vkakerbeck:
great catch! Looking forward to the benchmark results. I would expect a change in performance there but hopefully not a negative one (otherwise we can just delete this line)

jeremyshoemaker (approved):
Assuming everything works correctly still, this looks fine to me.

I'd also be fine with you renaming the variable to `sn_*` while you're here. I figured there's going to be a few of those that I missed, and we should just fix them as we find them.

vkakerbeck:
Thanks for running all these benchmarks so quickly! The results don't look strongly/consistently impacted in either direction. 

Just thinking through the logic behind this line I would tend towards removing it. The original idea behind it was that also when no curvature directions are stored at nodes (like when cd1==cd2 as on a flat surface) the added evidence will still be in a [-1, 1] range. However, I don't think we need it to be. It seems totally fine to say that if curvature directions are not defined, we have less confidence and only increment in [-0.5, 0.5] range. @nielsleadholm Do you see any issues with that?

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
```diff
@@ -381,7 +381,7 @@ def _get_pose_evidence_matrix(
             # nodes where pc1==pc2 receive no cd evidence but twice the pn evidence
             # -> overall evidence can be in range [-1, 1]
             cd1_evidence = cd1_evidence * use_cd
-            pn_evidence[np.logical_not(use_cd)] * 2
+            pn_evidence[np.logical_not(use_cd)] *= 2
```
codeallthethingz, line 384: unasked for opinion: as this is a breaking change anyway, maybe we should rename this to `surface_normal_evidence`

nielsleadholm:
Yeah @vkakerbeck I think that makes sense. Just to check my understanding, when is this line / the comment "nodes where pc1==pc2 receive no cd evidence but twice the pn evidence" actually relevant. If they were the same, wouldn't we have `query_features["pose_fully_defined"]=False`, and we wouldn't enter this `else` block in the first place?

vkakerbeck:
This one looks at the `node_features` (what is stored in the nearest neighbor nodes in the object model), not the `query_features` (what is currently sensed). So at each step, some nodes will have that and others don't. However, I think you touch on a good point, that if the query features have their pose_fully_defined but the nodes don't, then this is likely a wrong hypothesis anyways and should be receiving boosted (2x) evidence if the point normal matches well.

nielsleadholm:
Ok thanks, makes sense about stored nodes. Just to check, your last sentence - is that meant to be "should" or "shouldn't"? Under the condition that "the query features have their pose_fully_defined but the nodes don't, then this is likely a wrong hypothesis" implies to me that we would want to add _less_ evidence for these hypotheses, as they are probably considering different parts of an object.

If we wanted to do any up-scaling, it would be in the `if not query_features["pose_fully_defined"]:` block if the stored nodes are also `pc1==pc2` (e.g. modify `pn_weight`, although I'm not suggesting we need to add this now).

vkakerbeck:
yes, sorry, I meant to say "shouldn't". Good point, maybe we should test removing the line in this PR and instead adding the *2 multiplication in that block when the query features have pose features undefined. That would make more sense

hlee9212:
Hi @vkakerbeck and @nielsleadholm, I removed the unused line and updated the PR title. I'll create two new issues, one for updating `pn` to `surface_normal` and one for the if block. I have a slight preference for creating issues over writing TODO in code (slightly more visible to public, easier to track/account, technically TODO is not encouraged by Ruff), but let me know if this is not preferred. 

(I'll merge once I get an approval from either of you).

vkakerbeck:
Nice! Good idea about opening an issue. Could we add the TODO comment in addition? I usually find it useful to have that information in context as well. Like if I am reading a function and see the TODO I immediately know what we were thinking of something and what we were still planning to do but I wouldn't have a working memory of all the open issues and which lines of code they relate to.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
```diff
@@ -360,6 +360,9 @@ def _get_pose_evidence_matrix(
             cd1_weight = 0
             # Only calculate curv dir angle if sensed curv dirs are meaningful
             cd1_evidence = np.zeros(pn_error.shape)
+            # TODO: Test whether we should boost (2x) the evidence if the surface normal 
+            # matches, i.e. move use_cd from else block and set
+            # pn_evidence[np.logical_not(use_cd)] *= 2 (see PR#446)
```
vkakerbeck:
```suggestion
            # TODO: Test whether we should double the SN evidence if no 
            # curvatures are sensed and pose_fully_defined == False at node.
            # i.e. move use_cd from else block and set
            # pn_evidence[np.logical_not(use_cd)] *= 2 (see PR#446)
```

tristanls:
I renamed the PR to [`refactor:` A **src/tbp/monty** code change that neither fixes a bug nor adds a feature.](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0010_conventional_commits.md#types)

tristanls:
@hlee9212 , could you update the description? The description still highlights benchmarks and changes from before, but removing a line that was a noop should have no impact on benchmarks.

nielsleadholm (approved):
Looks good, thanks Hojae!

And good idea about opening an issue - for context (probably too much for the comment, which is I'm guessing why you left it out), it would be worth highlighting that we would do the `*= 2` because it would basically indicate a match between the sensation and the stored point that we are at an umbilical point. In other words, it's not just a surface-normal match, but also a match in the curvature information (flat point).

Issue #426, opened 2025-08-14
Combine Dataloader and Dataset class into EnvironmentInterface


annark:
Change the `Dataset` and `Dataloader` naming and setup to something that more accurately reflects that Monty learns in an environment and not from a static dataset.

annark:
I'm currently working on this issue.

tristanls:
**note**: Additional context is also available in the [Combine Dataloader and Dataset class into EnvironmentInterface thread on Discourse](https://thousandbrains.discourse.group/t/combine-dataloader-and-dataset-class-into-environmentinterface/751).

vkakerbeck:
Can we mark this issue as closed?

PR #448, opened 2025-09-01
refactor!: Move Dataset functionality into EnvironmentDataloader

annark:
### Context

Addresses issue #426 

This PR addresses the first two steps for combining `Dataloader` and `Dataset` classes into `EnvironmentInterface`, as outlined [here](https://thousandbrains.discourse.group/t/hello-anna-from-victoria/744/2?u=annark):
1. Move the four functions implemented in `EnvironmentDataset` into `EnvironmentDataLoader`.
2. Update the calls to those functions to reference the `EnvironmentDataLoader` instead.

The remaining steps will be completed in a subsequent PR.

Completing these steps also required the following changes:
- Updating the instantiation and class hierarchy of `EnvironmentDataloader` and its child classes.
- Moving simulator environment instantiation and lifecycle management into the `MontyExperiment` class.

Context and reasoning for these changes [here](https://thousandbrains.discourse.group/t/combine-dataloader-and-dataset-class-into-environmentinterface/751/18?u=annark).

### Questions
(copied over from [forum thread](https://thousandbrains.discourse.group/t/combine-dataloader-and-dataset-class-into-environmentinterface/751/7?u=annark)):

Given the following class hierarchy, defined in embodied_data.py:
```
class EnvironmentDataLoader:…
class EnvironmentDataLoaderPerObject(EnvironmentDataLoader):…
class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):…
class OmniglotDataLoader(EnvironmentDataLoaderPerObject):…
class SaccadeOnImageDataLoader(EnvironmentDataLoaderPerObject):…
class SaccadeOnImageFromStreamDataLoader(SaccadeOnImageDataLoader):…
```

I’ve refactored the child classes to call their super class `__init__` methods, since the `EnvironmentDataset` that was previously required as an initialization param for those child classes is now integrated into `EnvironmentDataLoader`, the super class. However there are a few parts that aren’t straightforward with the refactoring.

**Question 1:** `OmniglotDataLoader`’s super class `EnvironmentDataLoaderPerObject` expects `object_names` and `object_init_sampler` as init params, and calls `create_semantic_mapping` based on those. `OmniglotDataLoader` needs the environment to be initialized in order to define its object_names list -
```
object_names = [
    str(self.env.alphabet_names[alphabets[i]])
    + “_”
    + str(self.characters[i])
    for i in range(n_objects)
]
```
but the environment isn’t initiated until its super class runs. I handled this by letting `object_init_sampler` and `object_names` be optional params to `EnvironmentDataLoaderPerObject`, so that the child class can decide what to do with them. This also means `create_semantic_mapping` is not called if `object_names` is not provided. Does this approach align with how you’d like this handled?

**Question 2:** In order to update `SaccadeOnImageFromStreamDataLoader`, child class of `SaccadeOnImageDataLoader` with these changes, I had to do something about the scenes and versions params that `SaccadeOnImageDataLoader` expects on `__init__`. I chose to make them optional on the super class, but again wanted to check whether this approach of using optional parameters in this way fits the conventions of this codebase?

### Benchmarks
Running `python benchmarks/run.py -e base_config_10distinctobj_dist_agent` on main and my branch gives nearly identical results.

<details>
  <summary>Results from main</summary>

```
wandb: Run summary:
wandb: LM_0/episode/individual_ts_rotation_error 0.0892
wandb:       LM_0/episode/steps_to_individual_ts 28
wandb:                          episode/confused 0
wandb:                      episode/confused_mlh 0
wandb:                           episode/correct 1
wandb:                       episode/correct_mlh 0
wandb:           episode/goal_state_success_rate 0.66667
wandb:             episode/goal_states_attempted 3
wandb:                          episode/lm_steps 28
wandb:          episode/mean_lm_steps_to_indv_ts 28.0
wandb:              episode/monty_matching_steps 28
wandb:                       episode/monty_steps 49
wandb:                          episode/no_match 0
wandb:                     episode/pose_time_out 0
wandb:                    episode/rotation_error 0.0892
wandb:                          episode/run_time 0.74988
wandb:                 episode/symmetry_evidence 0.0
wandb:                          episode/time_out 0
wandb:           episode/used_mlh_after_time_out 0
wandb:              overall/avg_episode_run_time 1.23476
wandb:                  overall/avg_num_lm_steps 35.20714
wandb:      overall/avg_num_monty_matching_steps 35.20714
wandb:               overall/avg_num_monty_steps 83.85
wandb:                overall/avg_rotation_error 0.30123
wandb:                      overall/num_episodes 140
wandb:                  overall/percent_confused 1.42857
wandb:              overall/percent_confused_mlh 0.0
wandb:           overall/percent_confused_per_lm 1.42857
wandb:                   overall/percent_correct 98.57143
wandb:               overall/percent_correct_mlh 0.0
wandb:            overall/percent_correct_per_lm 98.57143
wandb:                  overall/percent_no_match 0.0
wandb:           overall/percent_no_match_per_lm 0.0
wandb:             overall/percent_pose_time_out 0.0
wandb:      overall/percent_pose_time_out_per_lm 0.0
wandb:                  overall/percent_time_out 0.0
wandb:           overall/percent_time_out_per_lm 0.0
wandb:    overall/percent_used_mlh_after_timeout 0.0
wandb:                          overall/run_time 172.86672
```
</details>

<details>
  <summary>Results from my branch</summary>

```
wandb: Run summary:
wandb: LM_0/episode/individual_ts_rotation_error 0.0892
wandb:       LM_0/episode/steps_to_individual_ts 28
wandb:                          episode/confused 0
wandb:                      episode/confused_mlh 0
wandb:                           episode/correct 1
wandb:                       episode/correct_mlh 0
wandb:           episode/goal_state_success_rate 0.66667
wandb:             episode/goal_states_attempted 3
wandb:                          episode/lm_steps 28
wandb:          episode/mean_lm_steps_to_indv_ts 28.0
wandb:              episode/monty_matching_steps 28
wandb:                       episode/monty_steps 49
wandb:                          episode/no_match 0
wandb:                     episode/pose_time_out 0
wandb:                    episode/rotation_error 0.0892
wandb:                          episode/run_time 0.75012
wandb:                 episode/symmetry_evidence 0.0
wandb:                          episode/time_out 0
wandb:           episode/used_mlh_after_time_out 0
wandb:              overall/avg_episode_run_time 1.23719
wandb:                  overall/avg_num_lm_steps 35.20714
wandb:      overall/avg_num_monty_matching_steps 35.20714
wandb:               overall/avg_num_monty_steps 83.85
wandb:                overall/avg_rotation_error 0.30123
wandb:                      overall/num_episodes 140
wandb:                  overall/percent_confused 1.42857
wandb:              overall/percent_confused_mlh 0.0
wandb:           overall/percent_confused_per_lm 1.42857
wandb:                   overall/percent_correct 98.57143
wandb:               overall/percent_correct_mlh 0.0
wandb:            overall/percent_correct_per_lm 98.57143
wandb:                  overall/percent_no_match 0.0
wandb:           overall/percent_no_match_per_lm 0.0
wandb:             overall/percent_pose_time_out 0.0
wandb:      overall/percent_pose_time_out_per_lm 0.0
wandb:                  overall/percent_time_out 0.0
wandb:           overall/percent_time_out_per_lm 0.0
wandb:    overall/percent_used_mlh_after_timeout 0.0
wandb:                          overall/run_time 173.20699
```
</details>


src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -54,111 +51,64 @@
...
-        PanTiltZoomCamera.translation_step
-
-    Note:
-        This one on its own won't work.
-    """
-
-    def __init__(self, dataset: EnvironmentDataset, motor_system: MotorSystem, rng):
-        assert isinstance(dataset, EnvironmentDataset)
         if not isinstance(motor_system, MotorSystem):
-            f"motor_system must be an instance of MotorSystem, got {motor_system}"
```
annark, line 151: This was previously a no-op as there was no `raise`.

src/tbp/monty/frameworks/environments/embodied_environment.py
```diff
@@ -114,6 +114,3 @@ def close(self):
         Any call to any other environment method may raise an exception
         """
         pass
-
-    def __del__(self):
-        self.close()
```
annark, line 119: I removed this  `__del__` method as part of addressing the segfault around destructing the simulator environment, due to Python's non-deterministic garbage collection timing. I didn't find any other uses of it, and moving the destructor of the env to monty_experiment's `__exit__` context manager protocol seems to address this resource management issue well.

src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -200,56 +200,47 @@ def init_model(self, monty_config, model_path=None):
...
+    def init_env(self, env_init_func, env_init_args):
+        self.env = env_init_func(**env_init_args)
+        assert isinstance(self.env, EmbodiedEnvironment)
+
+    def load_dataloaders(self, config):
         # Initialize everything needed for dataloader
-        dataset_class = config["dataset_class"]
-        dataset_args = config["dataset_args"]
-        self.dataset = self.load_dataset(dataset_class, dataset_args)
+        env_interface_args = config["dataset_args"]
```
annark, line 213: To contain the scope of changes in this PR, I left `config['dataset_args']` variables as they are, and intend to change them in the follow up PR where all `dataloader` references are changed to `env_interface`.

tests/unit/embodied_data_test.py
```diff
@@ -318,6 +310,41 @@ def check_two_d_movement(self, prev_loc, current_loc, action):
                 "Agent should not have moved",
             )
 
+    def test_omniglot_data_loader(self):
```
annark, line 316: I added this test as there were necessary changes to the `OmniglotDataLoader`, but no tests for it yet.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -202,10 +177,6 @@ def pre_epoch(self):
     def post_epoch(self):
         pass
 
-    def finish(self):
```
annark, line 205: Removed as this was unused, and closing of the environment now happens in `MontyExperiment`.

tristanls:
**note**: I added `!` to `refactor!:` as there is lots of breaking changes throughout.

src/tbp/monty/frameworks/environments/embodied_data.py
tristanls: 😬

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -175,18 +127,41 @@ def __next__(self):
         else:
             action = self.motor_system()
             self._action = action
-            self._observation, proprioceptive_state = self.dataset[action]
+            self._observation, proprioceptive_state = self.__getitem__(action)
```
tristanls: **issue**: As long as we have `__getitem__`, this should be `self[action]`. This comment applies here and to the rest of the file.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -819,40 +794,22 @@ def __init__(
...
-        if not isinstance(motor_system, MotorSystem):
-            raise TypeError(
-                f"motor_system must be an instance of MotorSystem, got {motor_system}"
-            )
-        self.dataset = dataset
-        self.motor_system = motor_system
-        self._observation, proprioceptive_state = self.dataset.reset()
-        self.motor_system._state = (
-            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
+        super(OmniglotDataLoader, self).__init__(
```
tristanls: **issue**: It should be `super().__init__(...)`. However, please see my general review comment about not making the hierarchy work this way.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -898,10 +855,8 @@ class SaccadeOnImageDataLoader(EnvironmentDataLoaderPerObject):
 
     def __init__(
         self,
-        scenes,
-        versions,
-        dataset: EnvironmentDataset,
-        motor_system: MotorSystem,
+        scenes=None,
+        versions=None,
```
tristanls: **issue**: As far as I can tell, if these are allowed to be `None`, then we get an error the first time `change_object_by_idx()` is called.

**question**: Why is it OK for these to be `None`?

src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -624,8 +614,16 @@ def load_state_dict(self, load_dir):
             setattr(self, k, exp_state_dict[k])
 
     def close(self):
-        if isinstance(self.dataset, EnvironmentDataset):
-            self.dataset.close()
+        dataloader = getattr(self, "dataloader", None)
+        if dataloader is not None and isinstance(
+            self.dataloader, EnvironmentDataLoader
+        ):
+            self.dataloader = None
```
tristanls: **question**: How necessary is it to set `dataloader` to `None`? After an experiment is closed, is there anything that happens aside from program exit?

tests/unit/habitat_data_test.py
```diff
@@ -143,35 +126,45 @@ def test_dataset_dist(self, mock_simulator_class, mock_agent_class):
...
+        action_space_dist.rng = rng
+        self.assertIsInstance(action_space_dist, ActionSpace)
+        self.assertCountEqual(action_space_dist, EXPECTED_ACTIONS_DIST)
+        self.assertIn(action_space_dist.sample(), EXPECTED_ACTIONS_DIST)
+
+        # Check if env interface is getting observations from simulator
         mock_sim_dist.get_sensor_observations.side_effect = self.mock_observations
         for i in range(1, DATASET_LEN):
-            obs_dist, _ = dataset_dist[motor_system_dist()]
+            obs_dist, _ = env_interface_dist.__getitem__(motor_system_dist())
```
tristanls: **issue**: Here and elsewhere, this should be `env_interface_dist[motor_system_dist()]`

tristanls (commented):
Thank you for diving into the data loaders and trying to make them work better. Untangling all of these is not straightforward, so thank you for making these improvements. Thank you for adding tests as well 🙏.

Overall, the details look great, but some high level comments...

To answer **Question 1**. I believe that the code here uses inheritance hiearchy to reuse functionality, and not to implement an "is a" relationship. This is made more apparent from the constructors not calling their respective supers. I think that instead of trying to make the inheritance hierarchy coherent by attempting to forge an "is a" relationship between classes, instead, we should accept that "is a" relationship does not exist. In this case, I believe the solution should be to duplicate the `OmniglotDataLoader`-specific constructor code, much like it was before, and not call `super()`. The current approach makes required `object_names` and `object_init_sampler` optional, which, I feel, is trading one approach (inheritance-for-code-reuse) for another (required-params-not-required). Eventually, neither are desirable, so I'd rather not shuffle them around to manufacture "is a" where there isn't one. Similar for **Question 2**, and making `scenes` and `versions` optional, since there isn't "is a" relationship, I don't think they should call `super()`.

I've seen `obj.__getitem__(param)` instead of `obj[param]` in many places. Please use the latter.

src/tbp/monty/frameworks/environments/embodied_data.py
annark: Ah, this is related to Viviane's comment [here](https://thousandbrains.discourse.group/t/combine-dataloader-and-dataset-class-into-environmentinterface/751/2?u=annark):
>Since this is already the way we interact with self.env in the EnvironmentDataset class (observation = self.env.step(action)) I agree that we can just strip it down to that call (+ transform application & state retrieval) and strip away any of the __getitem__ and torch.utils.data.Dataset related code.

I had planned to change the name of the `__getitem__` method to `step` as part of the `EnvironmentDataloader` class renaming step (which originally I anticipated being in the same PR). Unless I'm misunderstanding the intent, I'll change the name of `__getitem__` to `step` now, to address this issue and move towards Viviane's suggested approach.

src/tbp/monty/frameworks/experiments/monty_experiment.py
annark: Not in the `dataloader`, no, since the environment lifecycle is handled in `MontyExperiment`. I'll remove this.

src/tbp/monty/frameworks/environments/embodied_data.py
annark: Fixed with refactoring to no longer call super class `__init__` methods.

tests/unit/habitat_data_test.py
annark: Per my comment above, I've renamed `__getitem__` to `step()`.

src/tbp/monty/frameworks/environments/embodied_data.py
annark: Thank you for the review!

Regarding the refactoring of the Dataloader classes, your point about not trading one non-ideal approach for another makes sense 👍 . I've returned the classes to how they were previously, modulo the necessary changes for this task.

I originally went the inheritance route in order to bring transforms and the random number generator params (for transforms that need them) from the `EnvironmentDataset` class into  `EnvironmentDataloader`, along with the methods that were moved. This PR now adds those params into `EnvironmentDataloader` and each of its subclasses, plus copies the setup logic into each subclass's `__init__`.

Ideally the setup logic for all these common attributes would be handled only once in the highest common superclass, or use composition instead of a brittle inheritance and manual initialization chain. Out of curiosity, what is the longer term plan for this part of the framework, if known?

The TODO on [this line](https://github.com/thousandbrainsproject/tbp.monty/blob/d6565950206b998c3a9297049fa984a6c8504baa/src/tbp/monty/frameworks/environments/embodied_data.py#L1014) of the file suggested this inheritance approach. Would you like me to remove/modify that line?

annark:
The tests pass on my machine, but started failing after I updated the branch from the upstream main with the error
```
coverage.exceptions.DataError: Couldn't use data file '/home/runner/work/tbp.monty/tbp.monty/tbp.monty/.coverage.runnervmdmv4e.4905.XMxoZbgx': no such table: meta
```

Copilot is suggesting removing coverage files in github actions steps between runs:
```
  - name: Clean up coverage files
    run: rm -f .coverage*
  - name: Run tests with coverage
    run: pytest --cov
```

src/tbp/monty/frameworks/environments/embodied_data.py
tristanls: Thank you for the updates.

> Out of curiosity, what is the longer term plan for this part of the framework, if known?

In the long term, we aim for the code to be compliant with the [RFC 11 Code Organization Guide](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0011_code_organization_guide.md). Specifically, for data loaders, even after renaming, they will eventually be extracted from tbp.monty, when we extract all Experiment logic from the platform. But, this will be a while yet. Below is a small snapshot of the related Platform Future Reality Tree (I'm hoping to find time to publish it this week).

Screenshot 2025-09-29 at 13 09 04


Regarding the TODO, it appears to be outdated and can be removed.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -175,18 +122,41 @@ def __next__(self):
...
+            observation = transform(observation, state)
+        return observation
+
+    def step(self, action: Action):
+        observation = self.env.step(action)
+        state = self.env.get_state()
+        if self.transform is not None:
+            observation = self.apply_transform(self.transform, observation, state)
+        return observation, ProprioceptiveState(state) if state else None
+
```
tristanls, line 154: **note (non-blocking)**: I'm noticing that all but the first line of `reset()` and `step()` are the same.

tests/unit/habitat_data_test.py
```diff
@@ -103,7 +100,7 @@ def setUp(self):
 
     @mock.patch("habitat_sim.Agent", autospec=True)
     @mock.patch("habitat_sim.Simulator", autospec=True)
-    def test_dataset_dist(self, mock_simulator_class, mock_agent_class):
+    def test_env_interface_dist(self, mock_simulator_class, mock_agent_class):
```
tristanls, line 103: **note**: 🙂 forward thinking test name I see...

tristanls (approved):
All benchmark ran, no changes. Thank you again.

annark:
Thanks Tristan, exciting! 

I appreciate the resources you linked as well - between that RFC and the future reality tree it seems there’s a really clearly defined path to iteratively improving and building out the platform. Neat!

I’ll get started on the follow up PR in the coming days (and remove the obsolete TODO with that work).

Thanks again to you, Viviane and Jeremy for your support on this task.

vkakerbeck:
Thank you, Anna, for the huge effort. This is amazing to see! Really nice that you also added more tests (I am happy that the OmniglotDataLoader finally has test coverage!) :)

PR #449, opened 2025-09-02
chore: update column key in make_detailed_follow_up_configs.py

hlee9212:


tristanls:
~~I updated the PR name to `fix!:` since this is a breaking change to the "API" of interacting with collected stats.~~

tristanls:
Correction to my previous message, I updated this to `chore:` instead. [`fix:` is reserved for "Fix to a bug in the **src/tbp/monty** codebase."](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0010_conventional_commits.md#types) This is a change to `benchmarks` folder.

PR #459, opened 2025-09-05
Fixed path assumptions and use pathlib.Path for portability

pepedocs:
While running https://thousandbrainsproject.readme.io/docs/getting-started for the first time and following all the set up requirements, I encountered an error and also noticed that dir path manipulation is not os independent.

What are the changes?

1. Used pathlib.Path to allow os independent path manipulation. I have only tested this in Linux. For example using `expanduser()` to expand user paths (e.g. ~/tbp) and make this portable across oses.
2. Used pathlib.Path `/` operator when concatenating paths so that it will be portable. Previously it was simply using a string and it has some assumption about the existence of a `/` at the end of the `MONTY_LOGS_DIR` env var which is not conventional/usual. This caused an error while running the experiment.

I'm not sure if this PR is valid because I can't see it in the Contributions category, but just thought I'd fix this so that the next person wouldn't get stuck with debugging the missing `/` issue plus the possible path manipulation that are not os independent.

### DIFF
--- a/src/tbp/monty/frameworks/run_env.py
+++ b/src/tbp/monty/frameworks/run_env.py
@@ -10,6 +10,7 @@
 
 import os
 
+from pathlib import Path
 
 def setup_env(monty_logs_dir_default: str = "~/tbp/results/monty/"):
     """Setup environment variables for Monty.
@@ -17,30 +18,29 @@ def setup_env(monty_logs_dir_default: str = "~/tbp/results/monty/"):
     Args:
         monty_logs_dir_default: Default directory for Monty logs.
     """
-    monty_logs_dir = os.getenv("MONTY_LOGS")
 
-    if monty_logs_dir is None:
-        monty_logs_dir = monty_logs_dir_default
-        os.environ["MONTY_LOGS"] = monty_logs_dir
-        print(f"MONTY_LOGS not set. Using default directory: {monty_logs_dir}")
+    dir_path = os.getenv("MONTY_LOGS")
+    if dir_path is None:
+        dir_path = str(Path(monty_logs_dir_default).expanduser())
+        os.environ["MONTY_LOGS"] = dir_path
+        print(f"MONTY_LOGS not set. Using default directory: {dir_path}")
 
-    monty_models_dir = os.getenv("MONTY_MODELS")
+    monty_logs_dir = Path(dir_path)
+    
+    dir_path = os.getenv("MONTY_MODELS")
+    if dir_path is None:
+        dir_path = str(Path("~/tbp/results/monty").expanduser() / "pretrained_models")
+        os.environ["MONTY_MODELS"] = dir_path
+        print(f"MONTY_MODELS not set. Using default directory: {dir_path}")
 
-    if monty_models_dir is None:
-        monty_models_dir = f"{monty_logs_dir}pretrained_models/"
-        os.environ["MONTY_MODELS"] = monty_models_dir
-        print(f"MONTY_MODELS not set. Using default directory: {monty_models_dir}")
+    dir_path = os.getenv("MONTY_DATA")
+    if dir_path is None:
+        dir_path = str(Path("~/tbp/data").expanduser())
+        os.environ["MONTY_DATA"] = dir_path
+        print(f"MONTY_DATA not set. Using default directory: {dir_path}")
 
-    monty_data_dir = os.getenv("MONTY_DATA")
-
-    if monty_data_dir is None:
-        monty_data_dir = os.path.expanduser("~/tbp/data/")
-        os.environ["MONTY_DATA"] = monty_data_dir
-        print(f"MONTY_DATA not set. Using default directory: {monty_data_dir}")
-
-    wandb_dir = os.getenv("WANDB_DIR")
-
-    if wandb_dir is None:
-        wandb_dir = monty_logs_dir
-        os.environ["WANDB_DIR"] = wandb_dir
-        print(f"WANDB_DIR not set. Using default directory: {wandb_dir}")
+    dir_path = os.getenv("WANDB_DIR")
+    if dir_path is None:
+        dir_path = str(monty_logs_dir)
+        os.environ["WANDB_DIR"] = dir_path
+        print(f"WANDB_DIR not set. Using default directory: {dir_path}")


tristanls:
Thank you for updating this @pepedocs.

The overall update to pathlib.Path makes sense to me.

> This caused an error while running the experiment.

Could you share what caused this error? I assume you set `MONTY_LOG` to a non-default value? Are you setting `MONTY_LOG` and `MONTY_MODELS` to independent paths (one not nested in another)?

I'm still thinking through 3), because it removes the coupling between `MONTY_LOG` and `MONTY_MODELS`. I want to double-check that this change doesn't effect our infrastructure. (this won't be anything for you to do, just something I'll need to fix on our side before merging)

pepedocs:
> Could you share what caused this error? I assume you set `MONTY_LOG` to a non-default value? Are you setting `MONTY_LOG` and `MONTY_MODELS` to independent paths (one not nested in another)?

So I did not set MONTY_LOGS variable to anything so it took the default value based on the [code](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/run_env.py#L22). I also have not set MONTY_MODELS variable therefore again based on the [code](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/run_env.py#L30C9-L30C65), it concatenated MONTY_LOGS directory with the pretrained models directory. This concatenation has an assumption that the logs directory has a `/` which in my experience is not standard or at least can't be guaranteed. Users may define `/my/logs/dir` rather than `/my/logs/dir/`, regardless the pathlib usage will accept any form and still concatenates these correctly.

>I'm still thinking through 3), because it removes the coupling between MONTY_LOG and MONTY_MODELS. I want to double-check that this change doesn't effect our infrastructure. (this won't be anything for you to do, just something I'll need to fix on our side before merging)

No worries, the reason why I modified the path is because the code was looking into `MONTY_LOGS/pretrained-models` but the getting started [document](https://thousandbrainsproject.readme.io/docs/getting-started#42-download-pretrained-models) implied the pretrained-models are in `~/tbp/results/monty/pretrained_models/`. Therefore when I followed this doc, this caused another error. 

PS I only followed the getting started doc and left any optional requirements. I encountred two errors abovementioned and debug the code.

BTW @tristanls how can I run the linting locally so I don't have to commit my changes in order to see if my changes comply with code format?

tristanls:
> I encountred two errors abovementioned and debug the code.

Could you copy and paste the errors that you observed when running the code?

> BTW @tristanls how can I run the linting locally so I don't have to commit my changes in order to see if my changes comply with code format?

Yes, you can use the commands from the job locally: https://github.com/thousandbrainsproject/tbp.monty/blob/fc72a1738b6cb97ba1bdf7c067f620d644cf97cf/.github/workflows/monty.yml#L208-L209

pepedocs:
@tristanls 

I got quite distracted with the errors and debugging while following the getting started guide for the first time, please bear with me. I tried following this guide again from scratch and below are these errors I am trying to fix in this PR. 

Running `python benchmarks/run.py -e base_77obj_dist_agent`

1.  `FileNotFoundError: [Errno 2] No such file or directory: '/home/pepedocs/tbp/logspretrained_models/pretrained_ycb_v10/surf_agent_1lm_77obj/pretrained/model.pt'`
    - This is the first error while setting `MONTY_LOGS` to `"/home/pepedocs/tbp/logs"` it incorrectly concatenates the path `logspretrained_models`
    - I use pathlib to fix this so it is agnostic with slashes, and also bonus for different platforms (e.g. windows etc.)
2. `ModuleNotFoundError: No module named 'tbp'`
    - I had to set `PYTHONPATH=$PWD/src` to fix this. We might need to add this in the docs?
3. `FileNotFoundError: [Errno 2] No such file or directory: '/home/pepedocs/tbp/logs/pretrained_models/pretrained_ycb_v10/surf_agent_1lm_77obj/pretrained/model.pt'`
    - The error is due to the default model dir is `MONTY_LOGS/pretrained_models` which is inconsistent with the document it is saved in [~/tbp/results/monty/pretrained_models/](https://thousandbrainsproject.readme.io/docs/getting-started#42-download-pretrained-models)
    - The fix was to set it to the directory mentioned in the docs.
    - If this is not fixed, then I have to relocate my models to `MONTY_LOGS/pretrained_models` which is not intuitive (logs is confusing) and not mentioned in the docs.

tristanls:
@pepedocs, thank you for the elaboration. 

**praise**: Changes look good in general.

I have some edits I'd like to make that would be easier if I committed them to your branch instead of trying to get you to make them via comments. Are you OK with that?

src/tbp/monty/frameworks/run_env.py
```diff
@@ -17,30 +18,27 @@ def setup_env(monty_logs_dir_default: str = "~/tbp/results/monty/"):
...
-        os.environ["MONTY_DATA"] = monty_data_dir
-        print(f"MONTY_DATA not set. Using default directory: {monty_data_dir}")
-
-    wandb_dir = os.getenv("WANDB_DIR")
-
-    if wandb_dir is None:
-        wandb_dir = monty_logs_dir
-        os.environ["WANDB_DIR"] = wandb_dir
-        print(f"WANDB_DIR not set. Using default directory: {wandb_dir}")
+    dir_path = os.getenv("MONTY_LOGS")
```
tristanls: **issue**: I don't find the generic `dir_path` reuse more readable. I don't see a reason to get rid of `monty_logs_dir`, `monty_models_dir`, `monty_data_dir`, nor `wandb_dir` variables.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -17,30 +18,27 @@ def setup_env(monty_logs_dir_default: str = "~/tbp/results/monty/"):
...
+    if dir_path is None:
+        dir_path = str(Path(monty_logs_dir_default).expanduser())
+        os.environ["MONTY_LOGS"] = dir_path
+        print(f"MONTY_LOGS not set. Using default directory: {dir_path}")
+
+    monty_logs_dir = Path(dir_path)
+    dir_path = os.getenv("MONTY_MODELS")
+    if dir_path is None:
+        dir_path = str(Path("~/tbp/results/monty").expanduser() / "pretrained_models")
```
tristanls: **issue**: `~/tbp/results/monty` is equal to `monty_logs_dir_default`.

**suggestion**: introduce `monty_models_dir_default = "~/tbp/results/monty"` as an additional `setup_env()` function argument.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -9,6 +9,7 @@
 # https://opensource.org/licenses/MIT.
 
 import os
+from pathlib import Path
 
 
 def setup_env(monty_logs_dir_default: str = "~/tbp/results/monty/"):
```
tristanls: **suggestion (non-blocking)**: Since this commit now uses `Path`, then might as well make the function arguments be `Path`s instead of `str`ings.

src/tbp/monty/frameworks/run_env.py
```diff
@@ -17,30 +18,27 @@ def setup_env(monty_logs_dir_default: str = "~/tbp/results/monty/"):
...
+    monty_logs_dir = Path(dir_path)
+    dir_path = os.getenv("MONTY_MODELS")
+    if dir_path is None:
+        dir_path = str(Path("~/tbp/results/monty").expanduser() / "pretrained_models")
+        os.environ["MONTY_MODELS"] = dir_path
+        print(f"MONTY_MODELS not set. Using default directory: {dir_path}")
+
+    dir_path = os.getenv("MONTY_DATA")
+    if dir_path is None:
+        dir_path = str(Path("~/tbp/data").expanduser())
```
tristanls: **suggestion (non-blocking)**: While we're adding defaults, might as well add `monty_data_dir_default = "~/tbp/data"` to `setup_env()` function arguments and use it here.

tristanls (commented):
As I mentioned in my previous comment, overall changes look good, thank you.

I can make the suggested changes by committing to your branch if that's OK. However, I ended up writing up the details here if you'd rather do it yourself.

src/tbp/monty/frameworks/run_env.py
pepedocs: agreed and added

src/tbp/monty/frameworks/run_env.py
pepedocs: agreed and added

src/tbp/monty/frameworks/run_env.py
pepedocs: I didn't add this, I believe leaaving it as `str` is much simpler for a user of this function compared to a `Path`. Simply because `str` has no assumptions, it's just a string.

src/tbp/monty/frameworks/run_env.py
pepedocs: agreed and added

pepedocs:
@tristanls I've added most of your requests. Please don't hesitate to let me know of any changes you think are necessary, happy to help.

tristanls (approved):
Thank you, look good!

tristanls:
**note**: I added `refactor:` prefix to the PR title so I don't forget it in the comment message when I merge this.

tristanls:
**note**: Regarding the `check-style-monty` failure, running `ruff format src/tbp/monty/frameworks/run_env.py` should resolve it.

pepedocs:
hi @tristanls , I've updated the code by running the command above you mentioned. However I'm still getting the same error.

It seems it's not related to my changes, rather something like the coverage file is corrupted?

https://github.com/thousandbrainsproject/tbp.monty/actions/runs/17941029210/job/51017925785?pr=459#annotation:6:82

PR #467, opened 2025-09-12
feat!: Add compositional experiments & metrics

vkakerbeck:
Adding configs & new evaluation measures for learning and recognizing compositional models.

I will rerun pre-training and inference on AWS and update the benchmark results table after the general review of this PR.

### DIFF
--- a/docs/future-work/environment-improvements/make-dataset-to-test-compositional-objects.md
+++ b/docs/future-work/environment-improvements/make-dataset-to-test-compositional-objects.md
@@ -3,7 +3,7 @@ title: Make Dataset to Test Compositional Objects
 ---
 
 > [!TIP] We have implemented a first dataset to test compositional modeling!
-> You can find the dataset descriptions and current performance on it in out [benchmark documentation](../../overview/benchmark-experiments.md#compositional-datasets).
+> You can find the dataset descriptions and current performance on it in our [benchmark documentation](../../overview/benchmark-experiments.md#compositional-datasets).
 > Nevertheless, we will eventually want to implement a more complex dataset as described below to test scene level representations and generalization.
 
 To test compositional objects, we would like to develop a minimal dataset based on common objects (such as mugs and bowls) with logos on their surfaces. This will enable us to learn on the component objects in isolation, while moving towards a more realistic setting where the component objects must be disentangled from one another. The logo-on-surface setup also enables exploring interesting challenges of object distortion, and learning multiple location-specific associations, such as when a logo has a 90 degree bend half-way along its length.


src/tbp/monty/frameworks/utils/live_plotter.py
```diff
@@ -147,7 +148,7 @@ def show_view_finder(
             self.ax[0].add_patch(square)
         if hasattr(first_learning_module, "current_mlh"):
             mlh = first_learning_module.get_current_mlh()
-            if mlh is not None:
+            if mlh is not None and mlh["graph_id"] != "no_observations_yet":
```
tristanls: **issue**: Seems like instead of the `"no_observations_yet"` sentinel (I just now became aware of it), we should use `None`. This way, this extra check is not needed, and `self.current_mlh` is allowed to be `None`. As I read through the code, I don't think `self.current_mlh` can ever be `None`.

src/tbp/monty/frameworks/utils/live_plotter.py
```diff
@@ -147,7 +148,7 @@ def show_view_finder(
             self.ax[0].add_patch(square)
         if hasattr(first_learning_module, "current_mlh"):
```
tristanls: **issue**: Noticing now that this check isn't directly related to the line below. It should be `if hasattr(first_learning_module, "get_current_mlh"):`

tristanls (commented):
Sorry,  I didn't get a chance to review anything really. But I did stumble on `mlh` never being `None`. So I wanted to add these comments.

Perhaps a separate PR that changes `mlh` sentinel to be `None` instead of checking if "graph_id" property is set to "no_observations_yet" would be an improvement.

src/tbp/monty/frameworks/utils/live_plotter.py
vkakerbeck: Yeah that's a good point. I think the reason we set it to "no_observations_yet" (and also set defaults for the other current_mlh fields) is so we have all these fields available for logging, even when there were no observations (i.e. put "no_observations_yet" into the mlh column in .csv file is an LM didn't receive observations). But we could also set those defaults in the Logger instead and check for `current_mlh is None` there. I would do this in a separate PR though since it is not related to the compositional testbed (and might involve a bunch more unrelated changes).

tristanls:
I'm removing myself as an assignee in favor of @jeremyshoemaker.

benchmarks/configs/infer_compositional_objects.py
```diff
@@ -0,0 +1,319 @@
...
+    EnvInitArgsTwoLMDistantStackedMount,
+    TwoLMStackedDistantMountHabitatDatasetArgs,
+)
+
+# 14 unique rotations that give good views of the object. Same rotations used
+# for supervised pretraining.
+test_rotations_all = get_cube_face_and_corner_views_rotations()
+# test_rotations_all = [[0.0, 0.0, 0.0]]
+
+model_path_monolithic_models_lvl1 = os.path.join(
```
jeremyshoemaker, line 63: Do we want to add some documentation about what the various levels are? It's only been a month and I don't remember what they mean.

benchmarks/configs/infer_compositional_objects.py
```diff
@@ -0,0 +1,319 @@
...
+                min_post_goal_success_steps=5,
+                x_percent_scale_factor=0.75,
+                desired_object_distance=0.03,
+            ),
+        ),
+    ),
+    learning_module_1=dict(
+        learning_module_class=EvidenceGraphLM,
+        learning_module_args=dict(
+            max_match_distance=0.01,  # TODO: C - Scale with receptive field size
```
jeremyshoemaker: Do we want to address this TODO now?

benchmarks/configs/infer_compositional_objects.py
```diff
@@ -0,0 +1,319 @@
...
+for lm_id in ["learning_module_0", "learning_module_1"]:
+    two_stacked_constrained_resampling_lms_inf_config[lm_id]["learning_module_args"][
+        "hypotheses_updater_class"
+    ] = ResamplingHypothesesUpdater
+    two_stacked_constrained_resampling_lms_inf_config[lm_id]["learning_module_args"][
+        "evidence_threshold_config"
+    ] = "all"
+    two_stacked_constrained_resampling_lms_inf_config[lm_id]["learning_module_args"][
+        "object_evidence_threshold"
+    ] = 1
```
jeremyshoemaker: Could we do something like this instead and reduce the noise?
```suggestion
for lm_config in two_stacked_constrained_resampling_lms_inf_config.values():
    lm_args = lm_config["learning_module_args"]
    lm_args["hypotheses_updater_class"] = ResamplingHypothesesUpdater
    lm_args["evidence_threshold_config"] = "all"
    lm_args["object_evidence_threshold"] = 1
```
I'm pretty sure this will work, but it needs to be tested.

benchmarks/configs/infer_compositional_objects.py
```diff
@@ -0,0 +1,319 @@
...
+    ),
+)
+
+
+infer_parts_with_part_models = copy.deepcopy(infer_comp_base_config)
+infer_parts_with_part_models.update(
+    experiment_args=EvalExperimentArgs(
+        model_name_or_path=model_path_part_models,
+        n_eval_epochs=len(test_rotations_all),
+        # show_sensor_output=True,
```
jeremyshoemaker: Probably can remove this comment, since we can easily add it back if we need to test it.

benchmarks/configs/infer_compositional_objects.py
```diff
@@ -0,0 +1,319 @@
...
+        parent_to_child_mapping=PARENT_TO_CHILD_MAPPING,
+    ),
+)
+
+
+infer_comp_lvl1_with_comp_models = copy.deepcopy(infer_comp_base_config)
+infer_comp_lvl1_with_comp_models.update(
+    experiment_args=EvalExperimentArgs(
+        model_name_or_path=model_path_compositional_models_lvl1,
+        n_eval_epochs=len(test_rotations_all),
```
jeremyshoemaker: We're doing `len(test_rotations_all)` a lot here. The performance implications don't matter that much, but it might make sense to make this a top-level constant so we don't forget to change one of them if we change the rest.

Something like:
```python
NUM_EPOCHS = len(test_rotations_all)
```

benchmarks/configs/learn_compositional_objects.py
```diff
@@ -0,0 +1,388 @@
...
+]["gsg_args"] = dict(
+    goal_tolerances=dict(
+        location=0.015,
+    ),
+    elapsed_steps_factor=10,
+    min_post_goal_success_steps=20,
+    x_percent_scale_factor=0.75,
+    desired_object_distance=0.03,
+    wait_growth_multiplier=1,  # Since learning, enable more frequent jumps
+)
```
jeremyshoemaker, line 132: This is another place where assigning the "learning_module_args" dictionary to a temporary variable and doing the updates through that variable would be a lot less noisy and easier to read and understand.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -309,6 +320,7 @@ def __call__(self):
 class EnvironmentDataloaderPerObjectArgs:
     object_names: List
     object_init_sampler: Callable
+    parent_to_child_mapping: Dict = None
```
jeremyshoemaker: The type here should be `Optional[Dict]` or in the newer style `dict | None` (but I think we still need to import `from __future__ import annotations` for the Python version we're on). 

I haven't been calling out using the new style yet because we still have a lot of code using the old style, and I figure we'll update them all once we're on a version of Python that supports it directly.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -336,6 +347,16 @@ def change_object_by_idx(self, idx):
...
+        if self.parent_to_child_mapping is not None:
+            if self.primary_target["object"] in self.parent_to_child_mapping:
+                self.consistent_child_objects = self.parent_to_child_mapping[
+                    self.primary_target["object"]
+                ]
+            else:
+                logger.warning(
+                    f"target object {self.primary_target['object']} not in",
+                    " parent_to_child_mapping",
+                )
```
jeremyshoemaker: Does it make sense to run this code in the event that `self.parent_to_child_mapping` is an empty dictionary, i.e. `{}`? Currently it will be run in that case, but if it doesn't make sense (I could see it either way), it would make the conditional simpler to just do `if self.parent_to_child_mapping` and that would exclude both `None` and `{}`.

src/tbp/monty/frameworks/environments/logos_on_objs.py
```diff
@@ -0,0 +1,87 @@
...
+# https://opensource.org/licenses/MIT.
+
+"""There are 4 levels of difficulty for testing compositional models.
+
+Levels of Difficulty for Testing Compositional Models:
+1. Logos on flat surfaces
+2. Logos on curved surfaces
+3. Logos at different rotations
+4. Logos with bends
+"""
```
jeremyshoemaker, line 17: Ah, here's the definition of the levels I was looking for in the other file. I wonder if it belongs more in that file than this one, since this looks like it's just a bunch of constants?

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
```diff
@@ -67,21 +67,11 @@ def run_episode(self):
...
-            lm.buffer.stats["detected_location_on_model"] = (
-                self.first_epoch_object_location[target["object"]]
-            )
-            lm.buffer.stats["detected_location_rel_body"] = np.array(target["position"])
-            lm.buffer.stats["detected_rotation"] = target["euler_rotation"]
-            lm.detected_rotation_r = Rotation.from_quat(target["quat_rotation"]).inv()
-            lm.buffer.stats["detected_scale"] = target["scale"]
+        # Save compute if we are providing labels to all models, so don't need to
+        # perform matching parts of LM updates (default is matching_step)
+        if self.supervised_lm_ids == "all":
```
jeremyshoemaker: It's been bothering me that the type of this field is basically `str | list[str]` because strings and lists respond to a lot of the same method calls (they can both be iterated and indexed into), and so the potential exists to accidentally treat one like the other.

My initial thought on how to fix this, if we want to keep the "all" option, would be to accept that in the constructor but when setting `self.supervised_lm_ids` expand it out into the list of all the LMs in that case. That way all the code here can just treat it like a list of LM ids without having to deal with the special cases (it simplifies lines 104-105 below). But, I don't know how to deal with this conditonal here. Maybe move this setting of the `step_type` up into the constructor where the list of ids is set would solve that?

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
```diff
@@ -139,16 +164,16 @@ def pre_episode(self):
         if self.show_sensor_output:
             self.live_plotter.initialize_online_plotting()
 
+    def pre_epoch(self):
+        super().pre_epoch()
+
```
jeremyshoemaker: If we aren't adding anything here, we can remove this completely because then any calls to `pre_epoch` would automatically go to the parent class, similar to what `super().pre_epoch()` is doing here.
```suggestion
```

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
```diff
@@ -139,16 +164,16 @@ def pre_episode(self):
...
     def post_epoch(self):
```
jeremyshoemaker, line 165: Is there any reason why we're not calling `super().post_epoch()` somewhere in this method? Or are we completely overriding its behaviour?

src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
```diff
@@ -255,6 +260,17 @@ def update_overall_stats(self, mode, episode, episode_steps, monty_matching_step
...
+            if "episode_avg_prediction_error" in episode_stats.keys():
+                stats["episode_avg_prediction_error"].append(
+                    episode_stats["episode_avg_prediction_error"]
+                )
+
+            stats["num_correct_child_or_parent"] += (
+                int(performance == "consistent_child_obj")
+                or int(performance == "correct")
+                or int(performance == "correct_mlh")
+            )
```
jeremyshoemaker: This code seems too clever. Since these values are mutually exclusive, then the only possible values we're adding to this stat is 1 or 0. A better way to write this would be something like:
```suggestion
            if performance in {"consistent_child_obj", "correct", "correct_mlh"}:
                stats["num_correct_child_or_parent"] += 1
```

We could also use a list here by changing the `{}` to `[]`, but I think a set is faster for membership checks like this.

src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
```diff
@@ -456,6 +491,13 @@ def get_formatted_overall_stats(self, mode, episode):
             overall_stats["episode/lm_performances"] = wandb.Histogram(
                 episode_lm_performances
             )
+            # filter out prediction errors that are nan
+            prediction_errors = stats["episode_avg_prediction_error"][-len(self.lms) :]
+            valid_prediction_errors = [e for e in prediction_errors if not np.isnan(e)]
+            if len(valid_prediction_errors) > 0:
```
jeremyshoemaker: Since `valid_prediction_errors` is a list, the conditional should check whether it has contents like this:
```suggestion
            if valid_prediction_errors:
```

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -219,6 +226,10 @@ def update_hypotheses(
             return []
 
         hypotheses_updates = []
+        telemetry: dict[str, Any] = {}
```
jeremyshoemaker, line 229: Is `Any` the most specific we can get for the values in this dictionary?

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -445,18 +447,20 @@ def get_output(self):
...
             morphological_features={
                 "pose_vectors": pose_features,
                 "pose_fully_defined": not self._enough_symmetry_evidence_accumulated(),
+                # on_object is also same as sensor input to higher LM (assuming they are
+                # colocated) so not used.
                 "on_object": self.buffer.get_currently_on_object(),
             },
             non_morphological_features={
                 "object_id": object_id_features,
                 # TODO H: test if this makes sense to communicate
```
jeremyshoemaker: Should we address this TODO now? I'm not sure what it's suggesting.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -445,18 +447,20 @@ def get_output(self):
...
-                "location_rel_model": mlh["location"],
+                # "location_rel_model": mlh["location"],
```
jeremyshoemaker: Commented out code. Just remove it.
```suggestion
```

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -1183,6 +1189,46 @@ def _add_votes_to_buffer_stats(self, vote_data):
...
+                "channel_hypothesis_displacer_telemetry",
+            ) and hasattr(
+                channel_telemetry.channel_hypothesis_displacer_telemetry,
+                "mlh_prediction_error",
+            ):
+                displacer_telemetry = (
+                    channel_telemetry.channel_hypothesis_displacer_telemetry
+                )
+                channel_prediction_error = displacer_telemetry.mlh_prediction_error
+                prediction_errors.append(channel_prediction_error)
```
jeremyshoemaker: The `hasattr` function used to be a constant source of bugs in Python 2. It's also got some edge cases in Python 3, so I've always been cautious and tried to avoid its use. 

I think we can simplify this code in one of two ways. The first would be make sure whatever `channel_telemetry` is here, it always has some reasonable values for the attributes we're checking for.

The other way to handle this would be to just do the accesses as if the attributes are there, and catch any AttributeErrors and act accordingly. Like this:
```suggestion
            try:
                displacer_telemetry = (
                    channel_telemetry.channel_hypothesis_displacer_telemetry
                )
                channel_prediction_error = displacer_telemetry.mlh_prediction_error
                prediction_errors.append(channel_prediction_error)
            except AttributeError:
                # channel_telemetry was missing needed attributes, 
                # so skip adding prediction errors
                pass
```

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -1183,6 +1189,46 @@ def _add_votes_to_buffer_stats(self, vote_data):
...
+
+        if len(prediction_errors) > 0:
```
jeremyshoemaker: Another instance of checking whether a list contains items.
```suggestion
        if prediction_errors:
```

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -347,11 +354,14 @@ def update_hypotheses(
                         removed_ids=remove_ids,
                     )
                 )
+            else:
+                # Still return prediction error.
+                # TODO: make this nicer like dependent on log_level.
```
jeremyshoemaker, line 359: Should we go ahead and address this TODO now?

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -347,11 +354,14 @@ def update_hypotheses(
...
+                resampling_telemetry[input_channel] = ChannelHypothesesUpdateTelemetry(
+                    channel_hypothesis_displacer_telemetry=channel_hypothesis_displacer_telemetry
+                )
 
-        return (
-            hypotheses_updates,
-            resampling_telemetry if self.include_telemetry else None,
-        )
+        return (hypotheses_updates, resampling_telemetry)
```
jeremyshoemaker: This is a nitpick, so feel free to ignore it.

This doesn't need the parenthesis. The tuple is created by the `,` not the `()` around it. There are a few places where you have to use parenthesis to get a tuple instead of something else, but this isn't one of them.

I'm kinda surprised we don't have a formatting lint for this.

src/tbp/monty/frameworks/models/object_model.py
```diff
@@ -534,7 +535,13 @@ def _initialize_and_fill_grid(
         # The offset is set such that the first observed location starts at the
         # center of the grid. To preserve the relative locations, the offset is
         # applied to all following locations.
-        self._initialize_location_mapping(start_location=locations[0])
+        # Sort out nan locations. For some reason the first location is sometimes nan.
+        # TODO - C: Why is there a nan location?
```
jeremyshoemaker: 😄 Well, why is there a NaN location? Is this something we can easily fix now, or is it better to wait?

src/tbp/monty/frameworks/models/object_model.py
```diff
@@ -794,20 +801,31 @@ def _get_new_voxel_features(
...
+                avg_feat = get_most_common_value(feats)
             else:
                 avg_feat = np.mean(feats, axis=0)
             # Only take average if there was a feature stored here before.
             # since self._observation_count already includes the new obs
             # this needs to be > the number of new feature obs in the voxel.
-            if self._observation_count[voxel[0], voxel[1], voxel[2], 0] > len(feats):
+            num_obs_in_voxel = self._observation_count[voxel[0], voxel[1], voxel[2], 0]
+            num_new_obs = len(feats)
+            if num_obs_in_voxel > num_new_obs:
```
jeremyshoemaker, line 808: I like that the two parts were separated out into named variables, as it makes this easier to understand.

src/tbp/monty/frameworks/models/object_model.py
```diff
@@ -794,20 +801,31 @@ def _get_new_voxel_features(
                 avg_feat = get_most_common_bool(feats)
                 # NOTE: object_id may need its own most common function until
                 # IDs actually represent similarities
+            elif feature == "object_id":
+                avg_feat = get_most_common_value(feats)
```
jeremyshoemaker, line 800: I backtracked `avg_feat` up here because down on line 825 there's a comparison between `avg_feat` and `previous_average`, and thinking these were floats, that `if` would almost always be True.

Looking at this line, it seems (going off the function name) that these are actually not averages. I'm pretty sure this is the statistical _mode_, which could probably used to shorten the name of this function. Why do all the variables here mention "average" instead?

src/tbp/monty/frameworks/models/object_model.py
```diff
@@ -794,20 +801,31 @@ def _get_new_voxel_features(
...
                     if avg_feat is None:
                         avg_feat = previous_average
                     elif use_cds_to_update is False:
                         avg_feat[3:] = previous_average[3:]
+                elif feature == "object_id":
+                    # TODO: Figure out a more nuanced way to take into account past obs
+                    if avg_feat != previous_average:
+                        if num_old_obs > num_new_obs:
+                            avg_feat = previous_average
+                        avg_feat = previous_average
```
jeremyshoemaker: This conditional seems superfluous. We set `avg_feat = previous_average` no matter what on line 828.
```suggestion
                        avg_feat = previous_average
```

src/tbp/monty/frameworks/models/object_model.py
```diff
@@ -794,20 +801,31 @@ def _get_new_voxel_features(
...
                 # NOTE: could weight these
                 avg_feat = (avg_feat + previous_average) / 2
```
jeremyshoemaker, line 826: Okay, now it definitely looks like `avg_feat` is a float, so back on line 825, when is `avg_feat` ever going to be equal to `previous_average` except when we explicitly set it like we do inside that conditional?

src/tbp/monty/frameworks/utils/live_plotter.py
```diff
@@ -107,7 +107,8 @@ def show_observations(
             is_saccade_on_image_data_loader,
         )
         self.show_patch(first_sensor_depth)
-        self.show_mlh(mlh, mlh_model)
+        if mlh_model is not None:
```
jeremyshoemaker: If `mlh_model` is an object most of the time and only None occasionally, this can be simplified similar to checking if a list is empty. We're already doing something similar down on line 175.
```suggestion
        if mlh_model:
```

src/tbp/monty/frameworks/utils/live_plotter.py
jeremyshoemaker: If we don't end up making the changes @tristanls-tbp suggested, then I would suggest this change to tighten up the conditional. This is the more typical way you'd see this written in Python.
```suggestion
            if mlh and mlh["graph_id"] != "no_observations_yet":
```

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -809,6 +814,15 @@ def add_evidence_lm_episode_stats(lm, stats):
             ),
             4,
         )
+    # Check if the most likely object is a consistent child object
+    # Don't do this if the episode timed out, we had no match, or the detected object
+    # was already an exact match with the label.
+    if (
+        stats["primary_performance"] in ["confused", "confused_mlh"]
+        and consistent_child_objects is not None
```
jeremyshoemaker: The argument `consistent_child_objects` seems to be a collection from it's use in this function. I think in that case it being None and it being empty should be handled the same and thus this can be simplified to:
```suggestion
        and consistent_child_objects
```

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -857,6 +871,99 @@ def target_data_to_dict(target):
...
+        )
+        print(
+            f"LM_{lm_id} accuracy: {compositional_object_accuracy}% correct",
+            "(or correct_mlh)",
+        )
+        print(f"LM_{lm_id} consistent child accuracy: {consistent_child_accuracy}%")
+        print(
+            f"LM_{lm_id} average prediction error: ",
+            f"{np.mean(eval_stats_for_lm['episode_avg_prediction_error'])}",
+        )
```
jeremyshoemaker, line 952: Are these `print` calls left over from debugging, or are they meant to stay here? I think in the latter case they should be logging statements instead.

benchmarks/configs/infer_compositional_objects.py
vkakerbeck: That documentation is at the top of logos_on_objects.py right now but I will add a comment here as well.

benchmarks/configs/infer_compositional_objects.py
vkakerbeck: I missed removing this comment. We have a separate list of different hyperparameter configurations we want to test through here now: https://docs.google.com/document/d/17xM-NfAZQAQCaB9icW9qz1g8AI99qSIYJxrvGcI-se0/edit?tab=t.0#heading=h.e5dqr88419bn I think we would want to do this as a separate project

benchmarks/configs/infer_compositional_objects.py
vkakerbeck: tested it and it works :) much nicer to read!

benchmarks/configs/learn_compositional_objects.py
vkakerbeck: its slightly different from above since we don't loop over all lms here but I still made it a bit more readable by using .update on the dict one. Lmk what you think.

src/tbp/monty/frameworks/environments/embodied_data.py
vkakerbeck: Yeah I hadn't thought about that case (why would someone provide {} when the default is None?) but I agree covering both with `if self.parent_to_child_mapping` is safer

src/tbp/monty/frameworks/environments/logos_on_objs.py
vkakerbeck: I think it is useful here since it explains why the different objects are in the different lists. Also there are several config files that use these levels so we would have to list those at the top of each of them otherwise. I added a comment referring to this file to the configs instead. Lmk if you'd prefer having the level explanations in each file instead

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
vkakerbeck: good catch! I feel like I removed this in some commit but maybe I just imagined that :D

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
vkakerbeck: I think we just overwrite everything bc the supervised pretraining is a bit of a special experiment (i.e. we don't do any eval epochs and we don't want to save state dicts at every epoch) I see how this could lead to unexpected behavior though. But it's not anything we changed for this PR and it has worked for us for a while now.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
vkakerbeck: I'm not sure, I just copied this from https://github.com/thousandbrainsproject/tbp.monty/blob/f0e181c4f8f66103d941df89f622db29aee840fc/src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py#L277

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
vkakerbeck: I think this is something that we will get back to later when improving performance on these benchmarks

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
vkakerbeck: This is what the TODO refers to so I will update the TODO to be more specific instead.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
vkakerbeck: makes sense, thanks

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
vkakerbeck: I'm not sure how we would do this at the moment. I think it would require the whole telemetry infrastructure to take into account log level which sounds like a bigger change. Or can you think of an easier way to do this?

src/tbp/monty/frameworks/models/object_model.py
vkakerbeck: I am not sure :D I thought I had figured it out but then it resurfaced I think.I think it has something to do with the way the buffer fills in nan values when a channel appears at a later step. I can take a closer look again

src/tbp/monty/frameworks/models/object_model.py
vkakerbeck: for most features it is the average. But some types of features are treated differently like books (using get_most_common_bool) and object_id (getting the mode). But since for a lot of features it is the average I think it would be confusing to change the name to mode. I'm not sure I can think of a good alternative that would encompass all. Maybe `accumulated_feat`?

src/tbp/monty/frameworks/models/object_model.py
vkakerbeck: nice catch! But I think we would want to keep the conditional and remove the line below it (i.e. if we have more new than old obs, we do want to use avg_feat and not overwrite it

src/tbp/monty/frameworks/models/object_model.py
vkakerbeck: you're right, I updated the conditional to:
```
if num_old_obs > num_new_obs:
                            avg_feat = previous_average
                        else:
                            previous_average = avg_feat
```

src/tbp/monty/frameworks/utils/logging_utils.py
vkakerbeck: This is a helper function that is only called in analysis scripts (like in our jupyter notebook). So the print statements should be there to output the summary statistics (similar to some of the other functions in this file). I guess it's another question whether those type of functions should stay in tbp.monty or move to monty_lab or some other place

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
vkakerbeck: I took a crack at it, lmk what you think

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
vkakerbeck: I realized when I add `from __future__ import annotations` in that file ruff starts finding a bunch of unrelated type issues. Should I just fix all of them in the PR or stick with the old way of doing Optional[Dict]?

vkakerbeck (commented):
Nice, thanks for all the feedback and suggestions!

src/tbp/monty/frameworks/models/object_model.py
vkakerbeck: This is causing one of the unit tests to fail so debugging it right now

benchmarks/configs/learn_compositional_objects.py
jeremyshoemaker: I had thought about suggesting using `.update` to simplify and combine it a bit too. That's another good approach to clear up the noise. My main concern is the repetition of `two_stacked_constrained_lms_config_with_resampling["learning_module_0"]["learning_module_args"]` makes updating it a little harder because a change in one place might not get done in other places. And it's just a lot of noise to need to parse to see what's actually being changed.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
jeremyshoemaker: I think sticking with the old way for now is fine. We have a lot of places we need to change, and I don't think it makes sense to clutter up this PR with those unrelated changes.

src/tbp/monty/frameworks/environments/logos_on_objs.py
jeremyshoemaker: I'd rather not duplicate it if we can avoid it, just because then we have to keep them in sync. I think adding a comment to the other file referencing this one is a good approach.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
jeremyshoemaker: If it would involve a larger change to the telemetry code, we can leave it like this for now. That seems like an unrelated change.

src/tbp/monty/frameworks/models/object_model.py
jeremyshoemaker: I wasn't sure what the correct change was, so my suggestion was just keeping the current behaviour. If it makes sense to remove the line outside the `if` then that also works.

src/tbp/monty/frameworks/models/object_model.py
jeremyshoemaker: Yeah, I can't think of a better name either, given that extra context. I guess we can leave it as-is for now.

vkakerbeck:
A couple of things I still need to do (besides addressing remaining comments):

- [x] Figure out why nan's sometimes show up in the buffer locations.
- [x] Merge main & update configs after the recent refactors.
- [x] Check that tutorials still work.
- [x] Run all training and evaluations on AWS.
- [x] Update the benchmarks table with those results.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -49,6 +59,7 @@ class ExperimentArgs:
     model_name_or_path: str = ""
     min_lms_match: int = 1
     seed: int = 42
+    supervised_lm_ids: Union[List[str], Literal["all"]] = field(default_factory=list)
```
jeremyshoemaker, line 62: This looks fine to me. I forgot about setting defaults in a dataclass like this. I was thinking of the case of setting a default in a function argument list and the problems with using list or dictionary literals in those places. This is a bit different, and using the factory approach sidesteps the issue I was mentioning in our huddle.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -239,6 +248,8 @@ def __init__(self, object_names, object_init_sampler, *args, **kwargs):
         self.episodes = 0
         self.epochs = 0
         self.primary_target = None
+        self.consistent_child_objects = None
+        self.parent_to_child_mapping = parent_to_child_mapping
```
jeremyshoemaker: This is more like what I was mentioning in our huddle. We're using `None` as the default in the argument list, because using `{}` would break things in weird ways. 

But I don't think it makes sense to propagate that `None` to the rest of the code, so one way around that would be to change this to set it to an empty dictionary in the case where the argument is `None` or falsey.

```suggestion
        self.parent_to_child_mapping = parent_to_child_mapping if parent_to_child_mapping else {}
```

Then, down on line 350, we don't need to check if `self.parent_to_child_mapping` is empty or not, because the key lookup on 351 will work as expected.

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
```diff
@@ -67,21 +67,11 @@ def run_episode(self):
...
-            lm.buffer.stats["detected_location_on_model"] = (
-                self.first_epoch_object_location[target["object"]]
-            )
-            lm.buffer.stats["detected_location_rel_body"] = np.array(target["position"])
-            lm.buffer.stats["detected_rotation"] = target["euler_rotation"]
-            lm.detected_rotation_r = Rotation.from_quat(target["quat_rotation"]).inv()
-            lm.buffer.stats["detected_scale"] = target["scale"]
+        # Save compute if we are providing labels to all models, so don't need to
+        # perform matching parts of LM updates (default is matching_step)
+        if len(self.supervised_lm_ids) == len(self.model.learning_modules):
```
jeremyshoemaker: I think in most cases this will be fine, but we're not actually confirming that these two lists have the same elements. I think the quickest way to do this would be to turn them both into sets and compare the sets.
```suggestion
        if set(self.supervised_lm_ids) == set(self.model.learning_modules):
```

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
```diff
@@ -104,17 +94,49 @@ def run_episode(self):
...
                 if i == 0:
                     first_pos = self.sensor_pos[0]
                 else:
                     lm_offset = self.sensor_pos[i] - first_pos
-
-                    lm.buffer.stats["detected_location_rel_body"] += lm_offset
-                    # Rotate offset into model RF
-                    lm_offset_model_rf = lm.detected_rotation_r.apply(lm_offset)
-                    lm.buffer.stats["detected_location_on_model"] += lm_offset_model_rf
+                    if lm.detected_object is not None:
```
jeremyshoemaker: If `target["object"]` (which is what this gets set to on line 104) is an object and not a collection like a list or dict, then we can probably safely simplify this.
```suggestion
                    if lm.detected_object:
```

src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
```diff
@@ -255,6 +260,14 @@ def update_overall_stats(self, mode, episode, episode_steps, monty_matching_step
             )
             stats["monty_steps"].append(episode_steps)
             stats["monty_matching_steps"].append(monty_matching_steps)
+            # older LMs don't have prediction error stats
+            if "episode_avg_prediction_error" in episode_stats.keys():
```
jeremyshoemaker: This shouldn't need the `keys()` call. Using `in` on a dictionary checks for whether the value is a key in the dictionary.
```suggestion
            if "episode_avg_prediction_error" in episode_stats:
```

src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
```diff
@@ -441,6 +469,10 @@ def get_formatted_overall_stats(self, mode, episode):
             overall_stats[f"{lm}/episode/individual_ts_rotation_error"] = lm_stats[
                 "individual_ts_rotation_error"
             ]
+            if "episode_avg_prediction_error" in lm_stats.keys():
```
jeremyshoemaker: Same unneeded `keys()` call.
```suggestion
            if "episode_avg_prediction_error" in lm_stats
```

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -445,18 +447,21 @@ def get_output(self):
...
+                # colocated) so not used.
                 "on_object": self.buffer.get_currently_on_object(),
             },
             non_morphological_features={
                 "object_id": object_id_features,
-                # TODO H: test if this makes sense to communicate
-                "location_rel_model": mlh["location"],
+                # TODO H: test if it makes sense to communicate mlh["location"] as a
+                # non-morphological feature as well (would be kind of like the inverse
+                # of top-down connections).
```
jeremyshoemaker, line 463: This is much clearer than the previous TODO comment was. 👍

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -718,6 +724,7 @@ def _update_possible_matches(self, query):
         # NOTE: would not need to do this if we are still voting
         # Call this update in the step method?
         self.possible_matches = self._threshold_possible_matches()
+        self.previous_mlh = copy.deepcopy(self.current_mlh)
         self.current_mlh = self._calculate_most_likely_hypothesis()
```
jeremyshoemaker: Do we need the `deepcopy` here? On line 728, we're assigning a new value to `current_mlh`, so after these two lines (without the `deepcopy`) we'd have:

self.previous_mlh pointing to the old value of current_mlh
self.current_mlh pointing to the new value created inside _calculate_most_likely_hypothesis

I think the `deepcopy` call is just extra unnecessary work here.
```suggestion
        self.previous_mlh = self.current_mlh
        self.current_mlh = self._calculate_most_likely_hypothesis()
```

tests/unit/hierarchy_test.py
```diff
@@ -0,0 +1,419 @@
...
+            ),
+        )
+
+        two_stacked_lms_eval_config = copy.deepcopy(two_stacked_constrained_config)
+        two_stacked_lms_eval_config.update(
+            experiment_args=ExperimentArgs(
+                do_train=False,
+                min_lms_match=1,
+                n_eval_epochs=2,
+                model_name_or_path=self.compositional_save_path + "/pretrained",
```
jeremyshoemaker: This should use `os.path.join` or the newer `pathlib` equivalents to avoid any path separator differences between platforms. I've put off suggesting changing everything over to `pathlib` since there's a lot of the older style in the codebase, so I'm fine with using the `os.path` approach.
```suggestion
                model_name_or_path=os.path.join(self.compositional_save_path, "pretrained"),
```

tests/unit/hierarchy_test.py
```diff
@@ -0,0 +1,419 @@
...
+        )
+
+        two_stacked_semisupervised_lms_config = copy.deepcopy(
+            two_stacked_constrained_config
+        )
+        two_stacked_semisupervised_lms_config.update(
+            experiment_args=SupervisedPretrainingExperimentArgs(
+                supervised_lm_ids=["learning_module_1"],
+                min_lms_match=2,
+                model_name_or_path=self.compositional_save_path + "/pretrained",
```
jeremyshoemaker: Same as the other one
```suggestion
                model_name_or_path=os.path.join(self.compositional_save_path, "pretrained"),
```

jeremyshoemaker (commented):
Not sure how much what I'm commenting on this time is new, or just stuff I missed the first time around. Looks good for the most part, with some minor things left.

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
vkakerbeck: good idea

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
vkakerbeck: yeah that's probably true. Let me double check, I'm super paranoid about this :D

vkakerbeck: ok looks like its fine :)

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -336,6 +349,17 @@ def change_object_by_idx(self, idx):
             "semantic_id": self.semantic_label_to_id[self.object_names[idx]],
             **self.object_params,
         }
+        if self.primary_target["object"] in self.parent_to_child_mapping:
+            self.consistent_child_objects = self.parent_to_child_mapping[
+                self.primary_target["object"]
+            ]
+        elif self.parent_to_child_mapping.keys():
```
jeremyshoemaker: Checking to see if a dictionary is empty is more idiomatically done this way:
```suggestion
        elif self.parent_to_child_mapping:
```
The comment also should be updated or removed. I'm not sure it's needed after this change.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -718,6 +723,7 @@ def _update_possible_matches(self, query):
         # NOTE: would not need to do this if we are still voting
         # Call this update in the step method?
         self.possible_matches = self._threshold_possible_matches()
+        self.previous_mlh = self.current_mlh
         self.current_mlh = self._calculate_most_likely_hypothesis()
```
jeremyshoemaker, line 739: This is more of an FYI, since I don't think it makes sense to do it here with such long variable and function names, but another way to write this that you'll see often in Python code is to do both assignments at the same time.
```suggestion
        self.previous_mlh, self.current_mlh = self.current_mlh, self._calculate_most_likely_hypothesis()
```

jeremyshoemaker (commented):
Fewer comments this time 😄

jeremyshoemaker (approved):
Unless there are other changes you need to make, this looks good to me.

vkakerbeck:
Okay, I think I figured out why the performance changed after merging main. This PR https://github.com/thousandbrainsproject/tbp.monty/pull/503 added this line: https://github.com/thousandbrainsproject/tbp.monty/blob/5ab1f91aa97f24eec3285b2f924322f1263f8041/src/tbp/monty/frameworks/models/sensor_modules.py#L803 which resets `self._last_sent_n_steps_ago` on the first step of an episode. That previously didn't happen but it should happen. Since this seems like a bug in the previous code, I will just update our benchmark results here. It probably didn't show up in the existing benchmarks since there, we have lower `delta_thresholds` for `n_steps` (while the higher level LM in the comp config sets this quite high at 100 steps) so we may have never run into the case where a carry over from the previous episode causes the n_steps condition to trigger sending an observation to the LM. 
After adding the first step reset line to our pre-merge code, the results of post-merge are exactly replicated: https://wandb.ai/thousand-brains-project/Monty/runs/uralisi3?nw=nwuserviviane_clay
FYI @tristanls

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -517,7 +517,7 @@ def _add_attr_to_feature_buffer(self, input_channel, attr_name, attr_value):
             # sure the same index in different feature array corresponds to
             # the same time step and location.
             self.features[input_channel][attr_name] = np.full(
-                (len(self.locations), attr_shape), np.nan
+                (len(self) + 1, attr_shape), np.nan
```
jeremyshoemaker, line 520: I'm missing some context here to know why this was causing the NaNs and how this fixes it. Can you explain it to me a little?

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
```diff
@@ -71,7 +71,7 @@ def run_episode(self):
         # perform matching parts of LM updates (default is matching_step)
         all_lm_ids = [lm.learning_module_id for lm in self.model.learning_modules]
         if set(self.supervised_lm_ids) == set(all_lm_ids):
-            self.model.step_type = "exploratory_step"
+            self.model.switch_to_exploratory_step()
```
jeremyshoemaker, line 74: Didn't realize we had this function, or I'd have suggested it. I like this better than setting the step type directly.

src/tbp/monty/frameworks/models/buffer.py
vkakerbeck: Sure! When we are getting the first observation from a new channel (like when the HLLM gets the first observation from the LLLM we initialize an array of nans with the length of the buffer https://github.com/thousandbrainsproject/tbp.monty/blob/5ab1f91aa97f24eec3285b2f924322f1263f8041/src/tbp/monty/frameworks/models/buffer.py#L539 For example, the buffer might already be storing 10 observations from the sensor patch when it gets the first input from the learning module. For the array indices to match up correctly we initialize an array with 10 nans and then add the new location from the LLLM at the 11th position. 

This works correctly for locations but we were doing it wrong for the features (looking at the lenths of locations, which is 0 when initializing a new channel, instead of the length of the buffer). The append for features was correct though so we would get arrays that looked like `[nan, 1, nan, nan, nan, nan, 1, 1, 1]` instead of `[nan, nan, nan, nan, nan, nan, 1, 1, 1]`. I.e. there was already a stray value at index 1. This would then mess up the indexing when we used this function: https://github.com/thousandbrainsproject/tbp.monty/blob/5ab1f91aa97f24eec3285b2f924322f1263f8041/src/tbp/monty/frameworks/models/graph_matching.py#L1509 to extract entries with data from the buffer to update models, leading to it returning a nan value for the first location (since the location array correctly stores nan at index 1 but we used a feature array to determine `missing_features`.

I know its a bit complicated. Lmk if that makes sense

jeremyshoemaker: That makes sense. Thanks!

vkakerbeck:
Okay, I ran all experiments + pretraining on AWS and updated the benchmark table. In the process of that I found another small bug in the GSG (only showed up on rare occasions so wasn't caught in the runs locally) and updated the wandb logging to work when run in parallel.

This should be all for the first version of the new benchmark. I made a list of next steps I can look into next: https://docs.google.com/document/d/1hN-R33GLnLKDR98PIXJH9nxyxhcBCABmiFn8Q7xjLew/edit?usp=sharing

docs/future-work/environment-improvements/make-dataset-to-test-compositional-objects.md
```diff
@@ -2,6 +2,10 @@
 title: Make Dataset to Test Compositional Objects
 ---
 
+> [!TIP] We have implemented a first dataset to test compositional modeling!
+> You can find the dataset descriptions and current performance on it in out [benchmark documentation](../../overview/benchmark-experiments.md#compositional-datasets).
```
jeremyshoemaker: Typo.
```suggestion
> You can find the dataset descriptions and current performance on it in our [benchmark documentation](../../overview/benchmark-experiments.md#compositional-datasets).
```

src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
```diff
@@ -412,6 +412,10 @@ def get_formatted_overall_stats(self, mode, episode):
             "episode/confused_mlh": stats["episode_confused_mlh"],
             "episode/pose_time_out": stats["episode_pose_time_out"],
             "episode/time_out": stats["episode_time_out"],
+            "episode/consistent_child_obj": stats["episode_consistent_child_obj"],
+            "episode/consistent_child_or_parent": stats["episode_consistent_child_obj"]
+            or stats["episode_correct"]
+            or stats["episode_correct_mlh"],
```
jeremyshoemaker: Does this work correctly? It looks wrong to me, usually because Python will assume the end of a statement if there isn't something to indicate that it continues onto the next line. Maybe it works because it's expecting a `,` but the formatting of this looks wrong to me. I'm surprised that `ruff format --check` is okay with this.

jeremyshoemaker (commented):
Found a typo, and also have a question about some code that looks weird to me.

src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
vkakerbeck: autoformat broke it down like that to me. We are already doing this on some other lines further up but I can add some brackets just to be safe. Rerunning the experiment again to double check I get the same results.

jeremyshoemaker: That definitely looks better and is easier to understand. I was originally going to ask you if we wanted this entry and the previous one to both be `stats["episode_consistent_child_obj"]` or if that was a typo until I noticed the `or` on the next line, so this helps with that confusion.

vkakerbeck: yes I agree, that looks much nicer

jeremyshoemaker: There seems to be some debate in the `black` community, and by extension `ruff`, about whether the formatter should wrap long dictionary values in parenthesis or not, and it's not exactly cut and dry because of some odd edge cases. There doesn't seem to be a lint for it either.

vkakerbeck: good to know! fwiw it doesn't seem to be any different functionally, I still get the exact same results

jeremyshoemaker: It was something to do with removing parentheses sometimes changes the behaviour and they're reluctant to make the formatter do any changes that can affect program behaviour.

I don't expect anything to change here with adding the parentheses, aside from making it easier to understand.

PR #457, opened 2025-09-04
chore: upgrade pytorch and remove torchvision

ramyamounir:
This PR upgrades the package `torch` from `1.11.0` to `1.13.1` to resolve compatibility issues with `glibc>2.40`. We also remove the  unneeded dependency of `torchvision` (currently only required in `monty_lab`).

# Benchmark Results

I did not notice any change in accuracy or rotation error, but looks like there are some variations in runtime. The runs are on WandB with the tag `upgrade_torch`.

I've used the following tolerances:
```yaml
            tolerances:
                percent_correct: 0.1
                rotation_error: 0.01
                match_steps: 0.1
                runtime: 0.05
                used_mlh: 0.1
```

### base_10simobj_surf_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 93.57 | 93.57 | 0.00 | ✅ |
| rotation_error (deg) | 13.98 | 13.98 | -0.00 | ✅ |
| match_steps | 70.00 | 70.00 | 0.00 | ✅ |
| runtime (min) | 6.00 | 5.00 | -1.00 | 👍 |
| used_mlh | 8.57 | 8.57 | 0.00 | ✅ |

### base_config_10distinctobj_dist_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100.00 | 100.00 | 0.00 | ✅ |
| rotation_error (deg) | 16.20 | 16.20 | -0.00 | ✅ |
| match_steps | 37.00 | 37.00 | 0.00 | ✅ |
| runtime (min) | 2.00 | 2.00 | 0.00 | ✅ |
| used_mlh | 0.00 | 0.00 | 0.00 | ✅ |

### base_config_10distinctobj_surf_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100.00 | 100.00 | 0.00 | ✅ |
| rotation_error (deg) | 14.01 | 14.01 | 0.00 | ✅ |
| match_steps | 28.00 | 28.00 | 0.00 | ✅ |
| runtime (min) | 2.00 | 2.00 | 0.00 | ✅ |
| used_mlh | 0.00 | 0.00 | 0.00 | ✅ |

### randomrot_rawnoise_10distinctobj_surf_agent

| Metric | Baseline | Proposed | Δ | Result |
[...]

PR #468, opened 2025-09-14
fix! : hypothesis space initialization bugs for compositional tasks

ramyamounir:
This PR fixes a bug in the channel mapper when a new input channel is introduced during an episode (not at the start). This is a special case that happens during compositional learning. I've introduced some simplifications to the logic to make it more readable. The changes are:
- If the input channel is not in the mapper, we now [add a channel to the mapper](https://github.com/thousandbrainsproject/tbp.monty/blob/29d8afdbc9704e282d15085f7359e544c281af63/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L832-L835) regardless of how many channels already exist. The bug here was that we only added a channel when `len(mapper.channels) == 0`, which isn't enough for the compositional learning scenario.
- Fixed the size mismatch error, by initializing the [new channel with 0 hypotheses](https://github.com/thousandbrainsproject/tbp.monty/blob/29d8afdbc9704e282d15085f7359e544c281af63/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L835), then using the `mapper.update` to slice in the new data. This required initializing the data with [empty arrays of correct shape](https://github.com/thousandbrainsproject/tbp.monty/blob/29d8afdbc9704e282d15085f7359e544c281af63/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L820-L823). The mapper will later [resize](https://github.com/thousandbrainsproject/tbp.monty/blob/29d8afdbc9704e282d15085f7359e544c281af63/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L859) those 0 hypotheses to the correct number after slicing in the data.
- Fixed a hypotheses updater protocol violation that returned the possible locations as a [list instead of an array](https://github.com/thousandbrainsproject/tbp.monty/blob/29d8afdbc9704e282d15085f7359e544c281af63/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py?plain=1#L336-L338). This only happens during uniform hypotheses initialization.

### Benchmark Runs
__Coming soon__

### DIFF
--- a/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
@@ -333,6 +333,9 @@ class DefaultHypothesesUpdater:
                     )
                     initial_possible_channel_rotations.append(rotation.as_matrix())
 
+            initial_possible_channel_locations = np.array(
+                initial_possible_channel_locations
+            )
             initial_possible_channel_rotations = np.array(
                 initial_possible_channel_rotations
             )
--- a/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
@@ -829,8 +829,10 @@ class EvidenceGraphLM(GraphLM):
                 current_mean_evidence = np.mean(self.evidence[graph_id])
                 new_evidence = new_evidence + current_mean_evidence
 
-            # Add a mapper channel to be updated with the new data
-            mapper.add_channel(new_hypotheses.input_channel, len(new_evidence))
+            # Add a mapper channel to be updated with the new data. The mapper will
+            # be later resized to `len(new_evidence)` after we update the hypothesis
+            # space.
+            mapper.add_channel(new_hypotheses.input_channel, 0)
 
         # The mapper update function calls below automatically resize the
         # arrays they update. Afterward, we must update the channel indices
--- a/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
@@ -812,24 +812,25 @@ class EvidenceGraphLM(GraphLM):
         """
         # Extract channel mapper
         mapper = self.channel_hypothesis_mapping[graph_id]
-
         new_evidence = new_hypotheses.evidence
 
-        # Add a new channel to the mapping if the hypotheses space doesn't exist
         if new_hypotheses.input_channel not in mapper.channels:
+            # If there are currently no channels in the mapper, initialize the
+            # space with empty arrays of the correct shapes.
             if len(mapper.channels) == 0:
-                self.possible_locations[graph_id] = np.array(new_hypotheses.locations)
-                self.possible_poses[graph_id] = np.array(new_hypotheses.poses)
-                self.evidence[graph_id] = np.array(new_evidence)
+                self.possible_locations[graph_id] = np.empty((0, 3))
+                self.possible_poses[graph_id] = np.empty((0, 3, 3))
+                self.evidence[graph_id] = np.empty((0,))
 
-                mapper.add_channel(new_hypotheses.input_channel, len(new_evidence))
-                return
-
-            # Add current mean evidence to give the new hypotheses a fighting
-            # chance.
+            # If there exists other channels, add current mean evidence to give the
+            # new hypotheses a fighting chance.
             # TODO H: Test mean vs. median here.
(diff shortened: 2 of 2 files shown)


ramyamounir:
@nielsleadholm @vkakerbeck @jeremyshoemaker  

Could you confirm that this version fixes the problems you were seeing when running the compositional learning task? I plan to run all the benchmarks before merging but I want to make sure that this works first. Thanks!

vkakerbeck:
This is what we changed to fix it: https://github.com/thousandbrainsproject/tbp.monty/commit/b3ff42a0e0b9607b5db18c5a7494b3b9f25b5f9e

I am not sure if your change is sufficient since we also had to update the if statement in evidence_matching.py to `if (self.channel_sizes[channel] == data.shape[0]) and original.shape[0] > end:` Or is this fixed by the added line for `initial_possible_locations`? Also, why do you now initialize the arrays differently when `len(mapper.channels) == 0:`? This seems unrelated to the compositional object issue. Is that just because you noticed this would be more efficient?

ramyamounir:
```python
            initial_possible_channel_locations = np.array(
                initial_possible_channel_locations
            )
```

The above line fixes a separate issue that happens only during uniform hypothesis space initialization. We were getting the locations as a list of arrays instead of an array. This is a violation of the protocol for hypotheses updaters, but not related to the compositional issue.

---

It seems that the size mismatch is caused by [this line](https://github.com/thousandbrainsproject/tbp.monty/blob/b3ff42a0e0b9607b5db18c5a7494b3b9f25b5f9e/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L829). When we added a new channel with `new_evidence` size before actually adding the data, it caused the mapper and the actual data (e.g., `self.evidence`) to be out of sync during the `mapper.update` call. I now initialize an empty mapper channel and let the `mapper.update` take care of replacing the data (returns a view) or slicing the data in (returns a copy). Then the mapper channel size is updated afterwards [here](https://github.com/thousandbrainsproject/tbp.monty/blob/715c90f8dcc25a618d21c9b596278391af771887/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L859).

Is it straightforward to rerun the compositional code to make sure that it runs fine without any errors? The fix is only inside `_set_hypotheses_in_hpspace`, changes to `ChannelMapper.update` should be reverted for this test.

vkakerbeck:
I just applied your fix to our fork (https://github.com/thousandbrainsproject/feat.compositional_testbed/commit/8e86d67aa0e1998f50b0be5aef2df60857105454) and it replicated the results we got before without error (tested pertaining & inference). Seems to work :)

ramyamounir:
Awesome, thanks @vkakerbeck. I will rerun benchmarks for this fix and then mark the PR as ready for review.

ramyamounir:
@vkakerbeck I ran the benchmarks for this fix so that we can update them to the latest (given the discussion [here](https://github.com/thousandbrainsproject/tbp.monty/pull/457)). Please take a look at the updated description with new benchmark results.

vkakerbeck:
Why does this change here affect the benchmarks (besides slight runtime variations)? I thought the change in benchmarks discussed on the other PR was likely caused by this commit https://github.com/thousandbrainsproject/tbp.monty/commit/0ab205c9a9ae12be1a5ba62028a79064c2c01bb6? In that case, I would make a separate PR for the benchmark update and not associate it with the change here.

ramyamounir:
@vkakerbeck We never updated the benchmarks after merging in the [upgrade pytorch PR](https://github.com/thousandbrainsproject/tbp.monty/pull/457). The reported benchmarks here look similar to the ones we got during the pytorch upgrade. This fix here doesn't really affect the benchmarks but I thought it would be a good chance to update the docs to the latest results after running the new AWS images.

That said, I can update the benchmarks in a separate PR and then this fix will show no change in the benchmarks results. Let me know if you prefer that.

Note that I changed the symbols meaning in the comparison tables to what @nielsleadholm suggested.
- No change: ⬜
- Positive change: ✅
- Negative change: ❌

vkakerbeck:
Yes, I think a separate PR for the benchmark update would be better. We want to be able to associate which code changes lead to changes in benchmark results. If the code changes in this PR don't change benchmarks, then we shouldn't include the update in this PR.

vkakerbeck:
btw, I am happy to approve this PR already after you remove the benchmark results here. If this change doesn't cause a change in benchmarks I don't see a reason to wait with integrating it.

ramyamounir:
Nice, thanks. I've updated the description of this PR to show the deltas with [PR#485](https://github.com/thousandbrainsproject/tbp.monty/pull/485). The changes seem to be all runtime related, nothing too concerning in my opinion.

PR #469, opened 2025-09-15
fix: Broken doc links when 2+ links per line

AgentRev:
I noticed many broken links on multiple doc pages. I tracked it down to a regex pattern in `github_readme_sync` behaving incorrectly when multiple links are on the same line.

[Example](https://thousandbrainsproject.readme.io/docs/long-term-goals-and-principles):

image

### DIFF
--- a/tools/github_readme_sync/README.md
+++ b/tools/github_readme_sync/README.md
@@ -20,7 +20,7 @@ First, ensure you setup Monty. See [Getting Started - 2. Set up Your Environment
 Next, from the root Monty directory, install this tool's dependencies:
 
 ```
-pip install -e '.[dev,github_readme_sync_tool]'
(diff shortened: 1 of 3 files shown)


tristanls:
Thank you @AgentRev.

**note**: I renamed the PR to `chore:` as [`fix:` is reserved for `src/tbp/monty` fixes](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0010_conventional_commits.md#types).

codeallthethingz:
I had a quick look at this and the code looks right, and the new tests correctly simulate the issue, but the preview documentation links are still broken.  I'm going to run the build again and see if it works now.

AgentRev:
Confirmed working, links are fixed 🙂

vkakerbeck:
Awesome, thanks a lot for catching and fixing this @AgentRev !

PR #470, opened 2025-09-15
[DO NOT MERGE] Experimenting with `radius_evidence[mask] = -100`

hlee9212:
Update 2025-09-18: Decided to re-visit eliminating OORF hypotheses once the compositional benchmark is available. Closing PR.

---

This is an experiment to investigate what happens when we set the `radius_evidence[mask] = -100` (instead of -1) in `hypotheses_displacer.py`. This significantly penalizes the hypotheses that are considered out-of-reference frame, equivalent to eliminating hypotheses that are out of reference frame. This idea originates from PR#366, Question 2. 

This draft PR serves only to convey modified code and results, and should not be merged to `tbp.monty`. 

Screenshot of results comparing `base_77obj_surf_agent` with two different penalty values:
CleanShot 2025-09-17 at 11 33 15

Screenshot of results comparing `randrot_noise_77obj_surf_agent` with two different penalty values:
CleanShot 2025-09-17 at 11 37 22
CleanShot 2025-09-17 at 11 37 57
Accuracy drops from 95% to 74%. 

---
Note: This PR contains a slight modification to DetailedLogging to save selected episodes via detailed_run_stats_episode<EPISODE>.json. This was done for two reasons: (1) saving all episodes is too large (ran out of disk space), and (2) make_detailed_follow_up_configs.py may not be exactly faithful with GSG.


src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
```diff
@@ -262,6 +262,7 @@ def _calculate_evidence_for_new_locations(
         )
         # Set the evidences which are too far away to -1
         radius_evidence[mask] = -1
+        print("Penalizing -1")
```
vkakerbeck, line 265: not important to address this now, but in general we should use variables for this to reduce the chance of error. Like
```
oorf_penalty = -1
radius_evidence[mask] = oorf_penalty
print(f"Penalizing oorf with {oorf_penalty})
```

PR #474, opened 2025-09-22
fix: Add missing log stat to wandb

vkakerbeck:
When adding metric logging for the compositional object benchmark we noticed that we were tracking per lm performance stats but forgot to add them to the wandb logs. This adds them.

### DIFF
--- a/src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
+++ b/src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
@@ -445,6 +445,11 @@ class BasicGraphMatchingLogger(BaseMontyLogger):
                 "individual_ts_rotation_error"
             ]
 
+        for p in self.performance_options:
+            overall_stats[f"overall/percent_{p}_per_lm"] = (
+                stats[f"num_{p}_per_lm"] / (stats["num_episodes"] * len(self.lms))
+            ) * 100
+
         if len(self.lms) > 1:  # add histograms when running multiple LMs
             overall_stats["episode/rotation_error_per_lm"] = wandb.Histogram(episode_re)
             overall_stats["episode/steps_per_lm"] = wandb.Histogram(


src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
```diff
@@ -445,6 +445,11 @@ def get_formatted_overall_stats(self, mode, episode):
                 "individual_ts_rotation_error"
             ]
 
+        for p in self.performance_options:
```
nielsleadholm: **question:** It seems like there is some redundancy between this and lines 353-375 (copied below):

```python
            "overall/percent_correct_per_lm": (
                (stats["num_correct_per_lm"] + stats["num_correct_mlh_per_lm"])
                / (stats["num_episodes"] * len(self.lms))
            )
            * 100,
            "overall/percent_no_match_per_lm": (
                stats["num_no_match_per_lm"] / (stats["num_episodes"] * len(self.lms))
            )
            * 100,
            "overall/percent_confused_per_lm": (
                (stats["num_confused_per_lm"] + stats["num_confused_mlh_per_lm"])
                / (stats["num_episodes"] * len(self.lms))
            )
            * 100,
            "overall/percent_pose_time_out_per_lm": (
                stats["num_pose_time_out_per_lm"]
                / (stats["num_episodes"] * len(self.lms))
            )
            * 100,
            "overall/percent_time_out_per_lm": (
                stats["num_time_out_per_lm"] / (stats["num_episodes"] * len(self.lms))
            )
            * 100,
```

In particular, it seems like our proposed change fixes the absence of some of the below performance options from the above lines, but it also means that some of these results are calculated twice, and overwritten.

In some cases I think this may also change how we calculate things, e.g. for percent_correct, the manual calculation considers `correct` and `correct_mlh` as correct, but this isn't captured in the existing for-loop.

It might be worth removing where we calculate them outside the for-loop, unless we explicitly need to because of custom logic - but then we want to be careful we don't overwrite. This could maybe captured with an `if` branch.

```python
self.performance_options = [
            "patch_off_object",
            "no_label",
            "pose_time_out",
            "time_out",
            "confused_mlh",
            "correct_mlh",
            "no_match",
            "confused",
            "correct",
        ]
```

vkakerbeck: very good catch! Lmk what you think of this fix

vkakerbeck:
Yes, here is the 5LM experiment from this branch: https://wandb.ai/thousand-brains-project/Monty/runs/pwgiahmq?nw=nwuserviviane_clay and here from main on Sept. 3rd: https://wandb.ai/thousand-brains-project/Monty/runs/bpclgc93?nw=nwuserviviane_clay All of the graphs look exactly the same (except for some slight run time variance)

Here an example screenshot (also showing the metrics on the left that were previously not logged)
Screenshot 2025-09-23 at 4 26 17 PM

PR #475, opened 2025-09-22
fix: Supervised learning when object is translated

vkakerbeck:
So far we always kept the object at the same position and rotated it when doing supervised learning so we never noticed this bug. However, when translating the object between episodes, this translation was not correctly taken into account.
This fix only affects supervised learning since this is where we provide the model with the object position.

### DIFF
--- a/docs/hierarchy.md
+++ b/docs/hierarchy.md
@@ -35,6 +35,7 @@
   - [multiple-learning-modules](how-to-use-monty/tutorials/multiple-learning-modules.md)
   - [using-monty-in-a-custom-application](how-to-use-monty/tutorials/using-monty-in-a-custom-application.md)
   - [using-monty-for-robotics](how-to-use-monty/tutorials/using-monty-for-robotics.md)
+  - [running-an-experiment-from-a-different-repository](how-to-use-monty/tutorials/running-an-experiment-from-a-different-repository.md)
 - [common-issues-and-how-to-fix-them](how-to-use-monty/common-issues-and-how-to-fix-them.md)
 
 # how-monty-works: How Monty Works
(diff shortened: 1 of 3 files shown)


src/tbp/monty/frameworks/experiments/pretraining_experiments.py
```diff
@@ -122,6 +124,15 @@ def post_epoch(self):
         self.train_epochs += 1
         self.train_dataloader.post_epoch()
 
+    def pre_epoch(self):
+        super().pre_epoch()
+        # if it's the first epoch, save the primary target position. This is needed to
```
nielsleadholm: **thought:** This seems to assume that all of the objects within the first epoch are placed at the same location. I.e., even as `target["object"]` will change for a given episode, within the first epoch, `first_epoch_object_location` will not. I think this is a fair assumption given how our data-loaders/experiments work, but I wonder if it's worth adding a check for this?

vkakerbeck: That's a good point. What do you think of this solution: https://github.com/thousandbrainsproject/tbp.monty/pull/475/commits/8bfec1e09db789cb534b544f3ba97f314041ca97 (I'll still need to test it with the translated pre-training.)

vkakerbeck: Ok all still works as expected but now it is a lot more flexible so that new objects can appear whenever, wherever :)

vkakerbeck:
Since this change only affects pretraining there wouldn't be any effect on the benchmarks without pre-training. Even if I do pre-training again, we never have the case of translated objects in our benchmarks. So I reran model retraining for level 1 of our compositional dataset + inference on the level 1 models (after [this update](https://github.com/thousandbrainsproject/feat.compositional_testbed/commit/a022f858c1f957481e522a98bb1e28e592a8e009)). It replaced the results exactly: https://wandb.ai/thousand-brains-project/Monty/runs/j8ejojm5 (original: https://wandb.ai/thousand-brains-project/Monty/runs/flwatjrs)

PR #476, opened 2025-09-22
feat: Live plotting option for all MontyExperiment instances

vkakerbeck:
Previously the `show_sensor_output` option only worked for instances of `MontyObjectRecognitionExperiment`. However, there is nothing that prevents us from using the same feature in other `MontyExperiment` instances such as `MontySupervisedObjectPretrainingExperiment`. I moved the plotting code to the parent class and added the corresponding init and plot calls to `MontySupervisedObjectPretrainingExperiment`.

### DIFF
--- a/src/tbp/monty/frameworks/experiments/monty_experiment.py
+++ b/src/tbp/monty/frameworks/experiments/monty_experiment.py
@@ -44,7 +44,7 @@ from tbp.monty.frameworks.utils.dataclass_utils import (
     config_to_dict,
     get_subset_of_args,
 )
-from tbp.monty.frameworks.utils.plot_utils import add_patch_outline_to_view_finder
+from tbp.monty.frameworks.utils.live_plotter import LivePlotter
 
 # turn interactive plotting off -- call plt.show() to open all figures
 plt.ioff()
@@ -73,6 +73,9 @@ class MontyExperiment:
         self.config = config
 
         self.unpack_experiment_args(config["experiment_args"])
+
+        if self.show_sensor_output:
+            self.live_plotter = LivePlotter()
 
     def setup_experiment(self, config: Dict[str, Any]) -> None:
         """Set up the basic elements of a Monty experiment and initialize counters.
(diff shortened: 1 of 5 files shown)


src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -655,3 +663,113 @@ def __exit__(self, exc_type, exc_value, exc_traceback) -> Literal[False]:
         """
         self.close()
         return False  # don't silence exceptions inside the with block
+
+    ####
+    # Live plotting
+    ####
```
tristanls: **issue**: This should be a separate class that a `MontyExperiment` can be configured with if `show_sensor_output` is true.

```python
def __init__(...):
    ...
    if self.show_sensor_output:
        self.plotter = LivePlotter()
        self.plotter.initialize_online_plotting() # could be part of LivePlotter __init__

def ...
    ...
    if self.show_sensor_output:
        self.plotter.show_observations(observation, num_steps)
```

vkakerbeck:
I was thinking about this. Right now, it is true for all of our experiments to the same degree because, besides the `MontySupervisedObjectPretrainingExperiment` they all inherit from `MontySupervisedObjectPretrainingExperiment`. However, it is not guaranteed for any of them that those keys will be part of the observations. 
If I isolate the plotter into a separate class and explicitly state those assumptions, would that be okay? Otherwise, I think I will go down a deep rabbit hole since this live plotting code has a lot of things that are not great and very hacky about it. For now I just wanted to move it so that it is accessible from all the experiments that can be configured with the `show_sensor_output` argument. But while I am here I can fix some of it, I am just not sure where I should stop :D

tristanls:
> Right now, it is true for all of our experiments to the same degree

OK 👍 

> If I isolate the plotter into a separate class and explicitly state those assumptions, would that be okay?

Yes, moving the code into a separate class while still relying on the same hardcoded things will still be an improvement, as it will structurally highlight that it is completely isolated code. I think you should also be able to move `plt.ioff()` in there. I won't block the PR because of existing hardcoding. 🙂

vkakerbeck:
Nice, thanks for updating that! That all makes sense and looks good, sorry I didn't do it myself.

PR #481, opened 2025-09-26
feat: Add mlh to live plot + fix text

vkakerbeck:
Fixes a bug in the live plotter update that still assumed `current_mlh` to be an attribute of `graph_memory` instead of the LM. Also adds a 3D plot of the current MLH (for LMs that have an mlh).
The live plot now looks like this:

https://github.com/user-attachments/assets/451c7dba-fd25-4103-baaa-85f612f35f29

As before, you can see this plot by setting `show_sensor_output=True` in the `experiment_args`.


src/tbp/monty/frameworks/utils/live_plotter.py
```diff
@@ -59,15 +58,31 @@ def hardcoded_assumptions(self, observation, model):
         """
         first_learning_module = model.learning_modules[0]
         first_sensor_module_raw_observations = model.sensor_modules[0].raw_observations
-        patch_depth = observation[model.motor_system._policy.agent_id]["patch"]["depth"]
+        first_sensor_module_id = model.sensor_modules[0].sensor_module_id
+        patch_depth = observation[model.motor_system._policy.agent_id][
```
tristanls: **issue (non-blocking)**: `patch_depth` name no longer reflects what the contents are.

**suggestion (non-blocking)**: Rename to `first_sensor_depth` here and in following functions that use this argument.

PR #482, opened 2025-09-28
docs: update Bluesky profile path

Twintails:
- fix vale regex issue
- update SM link for Bluesky


codeallthethingz:
Thanks @Twintails this is great!

PR #486, opened 2025-10-01
refactor!: ensure update of "all" hypotheses with ResamplingHypothesesUpdater

ramyamounir:
Until now we've allowed `ResamplingHypothesesUpdater` to use any `evidence_threshold_config`. This should not be allowed when resampling. During resampling, hypotheses are deleted instead of being excluded from updating. This PR ensures that the config specifies `evidence_threshold_config="all"` when the `ResamplingHypothesesUpdater` is being used. Otherwise, `InvalidEvidenceThresholdConfig` exception is raised.


tristanls (approved):
LGTM

tristanls:
**note**: I feel this is more of a `refactor!`, as we're changing what exists, no new functionality is being added, in fact, it is being removed/restricted.

PR #487, opened 2025-10-02
feat!: Implement multi-action steps

scottcanoe:
This PR expands `EnvironmentDataLoader.step` by allowing it to perform several actions instead of just one. It's useful for triggering multiple actuators "at the same", like when saccading to a location that requires both a yawing and pitching action.

### DIFF
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -7,10 +7,12 @@
 # Use of this source code is governed by the MIT
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
+from __future__ import annotations
 
 import copy
 import logging
 from pprint import pformat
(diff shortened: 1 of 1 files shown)


src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -145,8 +147,13 @@ def apply_transform(self, transform, observation, state):
             observation = transform(observation, state)
         return observation
 
-    def step(self, action: Action):
-        observation = self.env.step(action)
+    def step(
+        self,
+        action: Action | Iterable[Action],
+    ) -> tuple[dict, ProprioceptiveState | None]:
```
scottcanoe, line 153: Please suggest an alternative to `dict` for the observation return type. I'm not sure where we are with types for that.

tristanls:
We need to make this change throughout all of Monty. I'll put together a PR based on the changes you proposed here.

PR #490, opened 2025-10-06
feat!: framework handles sequence of actions instead of single action

tristanls:
This pull request changes all `step(self, action: Action)` methods into `step(self, actions: Sequence[Action])`.

The Simulator protocol has been updated to include `apply_actions(self, actions: Sequence[Action])`.

After the refactor, it became clear that `self._action` in the data loaders is only used in tests. Therefore, this pull request removes `self._action` from data loaders. The test using `self._action` was extracted from the data loader test and moved into a unit test of the environment to which it belonged.

Policy internals are unaffected.
Simulator internals are unaffected.

Documentation is updated.

### DIFF
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -449,6 +449,7 @@ class InformedEnvironmentDataLoader(EnvironmentDataLoaderPerObject):
 
                 if (
                     not attempting_to_find_object
+                    and actions
                     and actions[0].name != OrientVertical.action_name()
                 ):
                     # We are not attempting to find the object, which means that we


src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -128,22 +128,22 @@ def set_experiment_mode(self, mode: Literal["train", "eval"]) -> None:
...
-            The action to take.
+            The actions to take.
         """
         if self.is_predefined:
             action: Action | None = self.predefined_call()
         else:
             action = self.dynamic_call(state)
         self.post_action(action, state)
-        return action
+        return [action] if action else []
```
tristanls, line 146: @scottcanoe, please take note here. This implementation still has policy code only return one action, which is then turned into a list. I assume you'll be interested in changing this next.

tristanls:
@scottcanoe, adding you to the PR since this should be enabling follow on PRs in our implementation project.

@vkakerbeck, adding you to the PR to verify that I'm still testing the thing you wanted to test with saccade 2D motion.

@jeremyshoemaker, adding you to the PR as I'm updating the simulator protocol and want you to see the nature of the changes.

tristanls:
On second thought @jeremyshoemaker, just an FYI 🙂

src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -331,51 +335,53 @@ def add_object(self, *args, **kwargs):
...
-                    "sensor_frame_data": sensor_frame_patch,
-                    "world_camera": self.world_camera,
-                    "pixel_loc": query_loc,  # Save pixel loc for plotting
-                },
-                "view_finder": {
-                    "depth": self.current_depth_image,
-                    "rgba": self.current_rgb_image,
-                },
+        obs = {}
+        for action in actions:
```
scottcanoe, line 357: **question**: does it ever worry you to have both the singular and plural version of a name on the board, like `action` and `actions`? Or is that just me. I pretty much assume I'll misread or mistype the `s` suffix at some point in the future.

src/tbp/monty/frameworks/models/motor_system.py
```diff
@@ -8,7 +8,7 @@
 # https://opensource.org/licenses/MIT.
 
 
-from typing import Literal, Optional
+from typing import List, Literal, Optional
```
scottcanoe: **question:** are we preferring `typing.List[...]` over `list[...]`? (assuming `from future import annotations`)?

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -120,9 +120,8 @@ def __next__(self):
             self._counter += 1
             return self._observation
         else:
-            action = self.motor_system()
-            self._action = action
-            self._observation, proprioceptive_state = self.step(action)
+            actions = self.motor_system()
+            self._observation, proprioceptive_state = self.step(actions)
```
scottcanoe, line 124: **thought:** I have yet to figure out when a simulator ever would, or ever does, return a `None`-valued `proprioceptive_state`. Assuming it could be, as the code implies, has been a pain w.r.t. typing.

scottcanoe (approved):
**praise:** This is much more comprehensive than I was expecting. Much thanks for the thoroughness and nice documentation changes.

**thought:** We're probably not too far off from needing to step the environment without performing any actions. Like when we passively view a stapler head move for a few frames (for object behavior experiments). "no action" is pretty naturally represented by an empty list, so that's a future bonus of switching to sequences of actions.

That said, lots of things will break right now if there isn't at least one action in `actions`. One of your .md files does say `actions` should contain one-or-more actions. I don't know how you feel about having an explicit check for this somewhere, at least until we can support an empty list of actions.

src/tbp/monty/frameworks/models/motor_system.py
tristanls: No, I wanted `list`, but did it wrong. I will update.

src/tbp/monty/frameworks/environments/two_d_data.py
tristanls: If I were to mistype it with an `s`, my IDE would highlight the problem for me, tests would fail, and the type checker would complain.

tristanls:
@vkakerbeck @scottcanoe please take another look. I wanted to ensure that given an empty actions sequence, the environment returns the current observations. Change here: https://github.com/thousandbrainsproject/tbp.monty/pull/490/commits/7244d848921181a719ae476a16a5260c9115619d

src/tbp/monty/frameworks/environments/embodied_data.py
tristanls: I agree that `None` should not be returned from the environment. Should be an easy future fix, but I'll keep it out of this PR.

scottcanoe (approved):
LGTM. Thanks again.

vkakerbeck (approved):
LGTM

PR #491, opened 2025-10-07
feat!: `DetailedJSONHandler` can save subset of episodes

hlee9212:
This PR adds capabilities to `DetailedJSONHandler` so that users can specify a list of ints (episode_ids) to save in `detailed_run_stats.json`. It also has an option to save as individual files per episode. 

---
<details>
<summary>
Previous description for Ramy and Scott
</summary>

Hey @ramyamounir, this is the updated DetailedJSONHandler that I used to save specific episodes. I included a `my_experiments.py` so you can test it out - the only annoying thing was that configuration was a bit annoying (I had to create the `CustomDetailedJSONLoggingConfig` class). Was this similar to what you were thinking about? 

After running, this is what I see in my results folder:
CleanShot 2025-10-07 at 14 01 40

Instead of `detailed_run_stats.json`, I now saved a JSON file for each episode I specified. 

I'll keep this as Draft for now - feel free to take it and create a PR with any other modifications you may need. Or if it looks good enough, I could clean up / remove the `my_experiments.py` and `names.py` example and open as a PR. 

Also tagging @scottcanoe because I think he was interested in some functionality like this too?
</details>


benchmarks/configs/my_experiments.py
```diff
@@ -7,17 +7,45 @@
...
+    )
+    save_per_episode: bool = True
+    save_consolidated: bool = False
+    episodes_to_save: List[int] = field(default_factory=list)
 
 
+custom_logging = copy.deepcopy(base_config_10distinctobj_surf_agent)
+custom_logging[
+    "logging_config"
+].run_name = "custom_logging"
```
ramyamounir: I know you will delete this so my comment here doesn't really matter, but I don't think this sets the `run_name` because you reinitialize `CustomDetailedJSONLoggingConfig` below.

src/tbp/monty/frameworks/loggers/monty_handlers.py
```diff
@@ -55,8 +55,36 @@ def log_level(self):
...
+        episodes_to_save=None,
+        parallel_episode_index=None,
+    ):
+        """Initialize the DetailedJSONHandler.
+
+        Args:
+            save_per_episode (bool): If True, save each episode as a separate file
+                in an 'episodes/' subdirectory. Defaults to True.
+            save_consolidated (bool): If True, also maintain the original behavior
+                of appending to a single consolidated file. Defaults to False.
```
ramyamounir: I think these two are doing the same job, I doubt we will ever need to save both at the same time. I suggest just using `save_per_episode` with default set to `False` (indicating consolidated file). When set to `True`, we only save individual episodes. Did you ever need to save both?

src/tbp/monty/frameworks/loggers/monty_handlers.py
```diff
@@ -55,8 +55,36 @@ def log_level(self):
...
+        self.save_per_episode = save_per_episode
+        self.save_consolidated = save_consolidated
+        self.parallel_episode_index = (
+            int(parallel_episode_index) if parallel_episode_index is not None else None
+        )
+        self.episodes_to_save = (
+            None
+            if episodes_to_save is None
+            else {int(ep) for ep in episodes_to_save}
+        )
```
ramyamounir: This is an OCD-style nitpick :sweat_smile:, but I would rearrange this ternary operator to match the one above it.
```python
        self.episodes_to_save = (
            {int(ep) for ep in episodes_to_save}
            if episodes_to_save is not None
            else None
        )
```

src/tbp/monty/frameworks/loggers/monty_handlers.py
```diff
@@ -70,30 +98,54 @@ def report_episode(self, data, output_dir, episode, mode="train", **kwargs):
...
-        maybe_rename_existing_file(save_stats_path, ".json", self.report_count)
+        output_data[global_total] = copy.deepcopy(stats)
+        output_data[global_total].update(data["DETAILED"][local_total])
 
-        with open(save_stats_path, "a") as f:
-            json.dump({total: output_data[total]}, f, cls=BufferEncoder)
-            f.write(os.linesep)
+        # Per-episode saving
+        if self.save_per_episode:
+            episodes_dir = os.path.join(output_dir, "episodes")
```
ramyamounir: We could name this directory `detailed_run_stats` or `detailed_stats` for clarity.

src/tbp/monty/frameworks/loggers/monty_handlers.py
```diff
@@ -70,30 +98,54 @@ def report_episode(self, data, output_dir, episode, mode="train", **kwargs):
...
+            os.makedirs(episodes_dir, exist_ok=True)
 
-        print("Stats appended to " + save_stats_path)
+            episode_file = os.path.join(episodes_dir, f"episode_{global_total:06d}.json")
+            with open(episode_file, "w") as f:
```
ramyamounir: It looks like we lost the ability to rename existing logs with this. Logs will be overwritten. I suggest looking into renaming the directory with `maybe_rename_existing_file` or similar. This way we still follow the same logging convention of Monty.

src/tbp/monty/frameworks/run_parallel.py
```diff
@@ -227,13 +271,23 @@ def post_parallel_eval(configs: List[Mapping], base_dir: str) -> None:
...
         if issubclass(handler, DetailedJSONHandler):
-            filename = "detailed_run_stats.json"
-            filenames = [os.path.join(pdir, filename) for pdir in parallel_dirs]
-            outfile = os.path.join(base_dir, filename)
-            post_parallel_log_cleanup(filenames, outfile, cat_fn=cat_files)
+            if save_per_episode and not collected_episode_logs:
+                collect_detailed_episode_logs(parallel_dirs, base_dir)
+                collected_episode_logs = True
+
+            if save_consolidated:
```
ramyamounir: This would also change if we remove the `save_consolidated` parameter.

src/tbp/monty/frameworks/run_parallel.py
```diff
@@ -227,13 +271,23 @@ def post_parallel_eval(configs: List[Mapping], base_dir: str) -> None:
...
-    for handler in configs[0]["logging_config"]["monty_handlers"]:
+    for handler in logging_config["monty_handlers"]:
         if issubclass(handler, DetailedJSONHandler):
-            filename = "detailed_run_stats.json"
-            filenames = [os.path.join(pdir, filename) for pdir in parallel_dirs]
-            outfile = os.path.join(base_dir, filename)
-            post_parallel_log_cleanup(filenames, outfile, cat_fn=cat_files)
+            if save_per_episode and not collected_episode_logs:
+                collect_detailed_episode_logs(parallel_dirs, base_dir)
+                collected_episode_logs = True
```
ramyamounir: I'm curious why we need this boolean flag.. Do we call this function multiple times or expect to have multiple `DetailedJSONHandler` instances?

src/tbp/monty/frameworks/run_parallel.py
```diff
@@ -215,6 +219,46 @@ def move_reproducibility_data(base_dir, parallel_dirs):
         )
 
 
+def collect_detailed_episode_logs(parallel_dirs, base_dir):
+    """Move per-episode detailed logs into a shared episodes directory."""
+
+    episodes_root = Path(base_dir) / "episodes"
```
ramyamounir: This directory name would also need to change if we rename it to `detailed_run_stats` or `detailed_stats`

src/tbp/monty/frameworks/run_parallel.py
```diff
@@ -215,6 +219,46 @@ def move_reproducibility_data(base_dir, parallel_dirs):
...
+    if episodes_root.exists() and any(episodes_root.iterdir()):
+        timestamp = int(time.time())
+        backup_dir = episodes_root.with_name(f"{episodes_root.name}_{timestamp}")
+        counter = 1
+        while backup_dir.exists():
+            backup_dir = episodes_root.with_name(
+                f"{episodes_root.name}_{timestamp}_{counter:02d}"
+            )
+            counter += 1
+        episodes_root.rename(backup_dir)
```
ramyamounir: I see you're doing the backup here, which would be equivalent to the functionality of `maybe_rename_existing_file`, correct? If so, then disregard my previous comment.

ramyamounir (commented):
This is awesome @hlee9212! Thanks a lot for putting it together, super useful!

I just left a couple of nitpicks, but yeah we should definitely merge it if possible. Let me know if you'd like me to help addressing the comments by pushing a few commits to your branch and cleaning up the configs. I won't create a new PR. Otherwise, if you can do it, that'd be great.

hlee9212:
Wow, thanks for the super fast feedback @ramyamounir! I'll address your comments and open the PR.

benchmarks/configs/my_experiments.py
hlee9212: Oooh, thank you! I think you may have resolved a weird bug/annoyance I had before related to this.

src/tbp/monty/frameworks/loggers/monty_handlers.py
hlee9212: I never saved both. :) I like your idea. Actually, I will remove both flags and just use `episodes_to_save` - if this is not `None`, then the behavior is that we want to save some list of episodes individually. I put a warning if the list is empty.

ramyamounir: Ah, I see. So:
- For consolidated, it would be `None`
- For some episodes saved individually, it would be, e.g., [1, 3, 5]
- For all episodes saved individually, it would be an empty list or a string, e.g., "*" ?

I like the idea of separating the option of which episodes to save from whether you want them consolidated or separate. I think you could for example want a few episodes in one file or all episodes in separate files.

src/tbp/monty/frameworks/loggers/monty_handlers.py
hlee9212: Naturally resolved by updating the Type hint to pass in a list[int] so I don't have to "int"-ify it. 😂

src/tbp/monty/frameworks/loggers/monty_handlers.py
hlee9212: Nice, I chose `detailed_run_stats`.

src/tbp/monty/frameworks/run_parallel.py
hlee9212: Yes - now it is just an "else" statement.

src/tbp/monty/frameworks/loggers/monty_handlers.py
hlee9212: Oh, I changed it so that if you set `episodes_to_save=[]` (empty list), then it won't save anything. I think if you want to save all episodes individually, passing something like `episodes_to_save=list(range(max_episode_number))` would work? 

I didn't think about the case of consolidating a few episodes in one file - it doesn't have that capability, but definitely useful. Let me add that in.

src/tbp/monty/frameworks/loggers/monty_handlers.py
hlee9212: Updated to call `maybe_rename_existing_file` for individual files.

src/tbp/monty/frameworks/run_parallel.py
hlee9212: It was for renaming the previously aggregated result.

src/tbp/monty/frameworks/loggers/monty_handlers.py
hlee9212: Added. Now you can specify which episodes to save, e.g. `[1, 3, 5]`, and whether you want to save that individually or in a single `detailed_run_stats.json` file.

src/tbp/monty/frameworks/run_parallel.py
hlee9212: I guess technically there is no stopping us in passing multiple DetailedJSONHandler (or another class that might inherit from DetailedJSONHandler)....but this unnecessarily complicates code so removed.

ramyamounir (approved):
Looks great, thanks. Maybe just add the default fields `episodes_to_save` and `save_per_episode` to the `DetailedEvidenceLMLoggingConfig` [here](https://github.com/thousandbrainsproject/tbp.monty/blob/880799ee75f803dfcda1da7fc81d2e1b560611ca/src/tbp/monty/frameworks/config_utils/config_args.py?plain=1#L227-L237)? 

I think we can keep the consolidated file as a default behavior (`save_per_episode=False` and `episodes_to_save=None`), then this would not be a breaking change and we can rename the PR back to `feat` (?)

ramyamounir:
I've also been thinking that it might be better to rename `save_per_episode` to `detailed_save_per_episode` and `episodes_to_save` -> `detailed_episodes_to_save`. This way we make it clear that these fields in the logging configs only affect the detailed handlers. The way it is named now, maybe someone could get confused thinking it would also apply to `BasicCSVStatsHandler` and `ReproduceEpisodeHandler`, or even `wandb_handlers`. What do you think?

```python
@dataclass
class DetailedEvidenceLMLoggingConfig(EvalEvidenceLMLoggingConfig):
    monty_handlers: List = field(
        default_factory=lambda: [
            BasicCSVStatsHandler,
            DetailedJSONHandler,
            ReproduceEpisodeHandler,
        ]
    )
    wandb_handlers: list = field(default_factory=list)
    monty_log_level: str = "DETAILED"
    detailed_save_per_episode: bool = False
    detailed_episodes_to_save: List[int] | None = None

```

src/tbp/monty/frameworks/loggers/monty_handlers.py
```diff
@@ -70,24 +96,77 @@ def report_episode(self, data, output_dir, episode, mode="train", **kwargs):
...
+            maybe_rename_existing_file(
+                episode_file, ".json", 0 if self.saved_episode_count == 0 else 1
+            )
+            with open(episode_file, "w") as f:
+                json.dump(
+                    {global_total: output_data[global_total]},
+                    f,
+                    cls=BufferEncoder,
+                    indent=2,
+                )
```
ramyamounir: Also just a note, I started using this PR already :slightly_smiling_face: and realized that `deserialize_json_chunks` will fail if we don't write a `os.linesep` after dumping the dictionary with json. I guess it's looking for the `\n`. We don't really need to use the deserialize function because we can just load the dictionary with `json.load` in this case but this is just one thing I noticed.

src/tbp/monty/frameworks/loggers/monty_handlers.py
```diff
@@ -55,8 +57,32 @@ def log_level(self):
 class DetailedJSONHandler(MontyHandler):
     """Grab any logs at the DETAILED level and append to a json file."""
 
-    def __init__(self):
+    def __init__(
+        self,
+        episodes_to_save: list[int] | None = None,
```
tristanls-tbp: **suggestion (non-blocking)**: If you make the default an empty list, it'll save you writing `if is not None` everywhere.

tristanls-tbp: Nevermind... I see the default is `all`, bad suggestion.

tristanls-tbp: **suggestion (non-blocking)**: Could we make the default be a more explicit sentinel, like `Literal["all"]`? I find `episodes_to_save=None` being translated to "save everything" confusing.

hlee9212: Updated.

src/tbp/monty/frameworks/loggers/monty_handlers.py
hlee9212: Thanks Ramy, I added this in.

hlee9212:
Hey @ramyamounir and @tristanls-tbp, I briefly converted this back to draft and re-opened again for review to do a bit of refactoring and address the above comments. I'm not really sure how to justify the refactoring except that previously it took me some "mental energy" to understand the code, so I tried to make it simpler for my brain. It follows the KISS principle for me, and also not striving for perfection but making it easier for future updates (hopefully). Hopefully I didn't open up a rabbit hole. 

~~Update: Back to Draft to address some failing pytests.~~

The difficult part of this update is not about saving a subset of episodes, but determining the correct `episode_id` when saving. There are two complications:

1. To work with `run_parallel.py`, I supplied a `episode_id_parallel` (otherwise all workers seems to have episode id of 0). 
2. It seems like `data["BASIC"]` and `data["DETAILED"]` have different ways of enumerating episode_id. It seems like "BASIC" stores episode_ids as {0, 1, .., N_train - 1} and {0, 1, ..., M_eval - 1} while "DETAILED" stores episode_ids as {0, 1, .., N_train + M_eval}. This seems to be coming from `graph_matching_loggers.py`. 

The current implementation works with `run.py`, but I haven't extensively tested on `run_parallel.py`, because my computer often hangs / never exits. 

I wish I could make it clearer but I think this problem is outside the scope of updating `DetailedJSONHandler`. @ramyamounir let me know how it goes - if something unexpected happens then I think it will be best to capture that in another PR with a unit test and updated solution, since current set of tests are passing.

src/tbp/monty/frameworks/loggers/monty_handlers.py
```diff
@@ -55,40 +59,152 @@ def log_level(self):
...
+            detailed_episodes_to_save: Iterable of episodes to save or
+                the string ``"all"`` (default) to include every episode.
+            detailed_save_per_episode: Whether to save individual episode files or
+                consolidate into a single detailed_run_stats.json file.
+                Defaults to False.
+            episode_id_parallel: Episode id associated with current run,
+                used to identify the episode when using run_parallel.
+        """
+        self.already_renamed = False
+        self.detailed_episodes_to_save = detailed_episodes_to_save
```
scottcanoe, line 80: **suggestion**: Based on its usage `detailed_save_per_episode` is of the `Container` type. I didn't try it, but `range` objects should also work (and `set`, `pandas.Index`, etc.) .

scottcanoe (approved):
**praise:** Nice, looks great!

**suggestion (non-blocking)**: When saving per-episode, it seems to me like the cleanest way to handle existing files is to rename the entire `detailed_run_stats` directory to `detailed_run_stats_old` and start fresh with a new directory. Having old and new json files living side-by-side seems dangerous on the loading side.

More specifically, if someone looks for a file that starts with `"episode_000005"`, they might end up loading `"episode_000005_old.json"` instead of `"episode_000005.json"`.

src/tbp/monty/frameworks/loggers/monty_handlers.py
ramyamounir: Aren't `range` and `set` of type `Iterable` as well? Is `Container` a more general type?

ramyamounir:
@hlee9212 There were a couple of bugs preventing `run_parallel` from storing the logs. I took a stab at it and did a few other things in the `run_parallel.py` file, such as reusing the `post_parallel_log_cleanup` function and renaming the `detailed_run_stats` directory as @scottcanoe suggested.

Feel free to revert the commits if you find a better way. I'll wait for you to review, edit and merge it since I changed some stuff.

src/tbp/monty/frameworks/loggers/monty_handlers.py
scottcanoe: Hm yeah, `range` and `set` are iterable. But the code only requires that `detailed_save_per_episode` implements `__contains__`, so, strictly speaking, `Container` is the right protocol/type.

I definitely should have marked this as a **nitpick** though.

ramyamounir: Ah I see, yeah that makes sense. Updated in [bcf85bf1dd0e](https://github.com/ramyamounir/tbp.monty/commit/bcf85bf1dd0e07abcc75d45cf39d1a0818db7e81).

ramyamounir (approved):

hlee9212:
Hey @ramyamounir, thanks for the changes! Looks good to me. I'll update branch and merge.

PR #493, opened 2025-10-08
docs: Update behavior theory page + add associated future work pages

vkakerbeck:
The theory page update has been edited and reviewed by Jeff. Afterwards I added some future work pages with loose descriptions. I can add more detail if you think it would be useful but didn't want to be too specific since detail may change and are probably better in an RFC.

### DIFF
--- a/benchmarks/results/ycb_10objs.csv
+++ b/benchmarks/results/ycb_10objs.csv
@@ -9,5 +9,5 @@ randrot_noise_10distinctobj_5lms_dist_agent,100.00,0.00,52,44.71,5,40
 base_10simobj_surf_agent,93.57,8.57,70,12.34,5,27
 randrot_noise_10simobj_dist_agent,84.00,40.00,234,31.08,11,89
 randrot_noise_10simobj_surf_agent,93.00,34.00,172,22.81,14,120
-randomrot_rawnoise_10distinctobj_surf_agent,64.00,74.00,15,101.1,4,7
+randomrot_rawnoise_10distinctobj_surf_agent,67.00,77.00,15,117.12,4,7
 base_10multi_distinctobj_dist_agent,80.71,14.29,30,19.60,3,1
--- a/benchmarks/results/ycb_77objs.csv
+++ b/benchmarks/results/ycb_77objs.csv
@@ -1,6 +1,6 @@
 Experiment,Correct (%)|align right,Used MLH (%)|align right,Num Match Steps|align right,Rotation Error (degrees)|align right,Run Time (mins)|align right,Episode Run Time (s)|align right
 base_77obj_dist_agent,93.51,12.99,107,16.46,20,68
 base_77obj_surf_agent,98.70,6.49,56,11.38,12,34
-randrot_noise_77obj_dist_agent,89.18,22.00,147,37.37,32,117
+randrot_noise_77obj_dist_agent,89.18,22.08,147,37.37,32,117
 randrot_noise_77obj_surf_agent,93.94,23.81,115,36.70,36,130
 randrot_noise_77obj_5lms_dist_agent,87.01,0.00,64,60.61,10,95


vkakerbeck:
@codeallthethingz I am not sure why the docs check fails. Maybe I am missing an obvious typo but the link looks correct to me in all files and when I run `python -m tools.github_readme_sync.cli check docs` locally it all passes. Is it an issue with using ../../ relative links?

codeallthethingz:
no, its because the filename is case-sensitive I think.

vkakerbeck:
Perfect, thanks! That solved it

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
+Learning the behaviors of objects is an essential component of any intelligent system. If an intelligent system is going to generate novel behaviors, it first needs to learn the behaviors of objects and then enact those behaviors in both familiar and novel situations.
 
-# The Solution
+The Thousand Brains Theory (TBT) currently explains how an intelligent system learns the static structure of the world.  Here, we describe how to extend the TBT to learn the dynamic structure of the world. In the TBT, static objects are learned as a set of features at poses (location and orientation) in a reference frame. There is an assumption that the features are not changing or moving, therefore, the existing theory and implementation work well for representing static structure.
```
nielsleadholm: **nitpick:** There's something about the ordering of this paragraph that I keep tripping over. It's a combination I think of repetition and jumping between two concepts. I think the below might flow better? (getting rid of first sentence, and reversing the rest of the paragraph)

> In the TBT, static objects are learned as a set of features at poses (location and orientation) in a reference frame. There is an assumption that the features are not changing or moving, therefore, the existing theory and implementation work well for representing static structure. Here, we describe how to extend the TBT to learn the _dynamic_ structure of the world.

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
 
-![Behavior models (purple) use an analogous mechanism to morphology models (green). The main difference is that they store changes instead of static features and have a temporal dimension.](../../figures/theory/behavior_models.png#width=600px)
+To learn and represent object behaviors, we use the same mechanism as we use for learning static structure. The main difference is what the input features represent. In the static models, the features are also static, such as the edge of a mug. In the dynamic models, the features represent changes, such as the moving edge of a stapler top as it opens. Static features are stored at locations in static/morphology models, and changing features are similarly stored at locations in behavior models. Another difference is that behaviors occur over time. As a stapler opens, the locations where moving edges occur change over time. Therefore, behavior models are sequences of changes at locations.
```
nielsleadholm: **suggestion:** 

"Another difference is that behaviors occur over time. As a stapler opens, the locations where moving edges occur change over time. Therefore, behavior models are sequences of changes at locations"

- I wonder if we want to reword this emphasis? Morphology models can also have a time component, and indeed might be where we learn a "behavior" like a stoplight suddenly changing from red to green and back again. Maybe we can emphasize that we are adding time as a component, which can be used by all models. In practice, it is particularly important for behavior models, for the reasons you describe.

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
 
-## Implementation \- Illustrated on the Example of the Neocortex
+Static models and dynamic models are learned in separate reference frames, but they share sensor patches and how the sensor patch is moving. Therefore, when we observe a stapler, we can simultaneously learn both the morphology of the stapler and how that morphology changes over time. But because the behavioral model has its own reference frame, it exists independently of the stapler. Now imagine we see a new object that doesn’t look like a stapler. If this new object starts to open like a stapler, then the stapler’s behavior model will be recognized and we will predict the new object behaves like a stapler. This method is very general and applies to every type of behavior we observe in the world.
```
nielsleadholm, line 19:
```suggestion
Static models and dynamic models are learned in separate cortical columns, but they can share sensor patches and how the sensor patch is moving. Therefore, when we observe a stapler, we can simultaneously learn both the morphology of the stapler and how that morphology changes over time. But because the behavioral model has its own column with its own reference frame, it exists independently of the stapler. Now imagine we see a new object that doesn’t look like a stapler. If this new object starts to open like a stapler, then the stapler’s behavior model will be recognized and we will predict the new object behaves like a stapler. This method is very general and applies to every type of behavior we observe in the world.
```

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
+
+The figure below illustrates how state and time relate to the two types of models. The behavioral model on the left stores where changes occur over time. It does not represent the morphology of the object exhibiting the behavior. It represents behaviors independent of any particular object. The morphology model on the right stores where static features occur for a particular object. It may have learned several morphology models for the object, such as an open and closed stapler, but on its own, it doesn’t know how the stapler moves. The concept of state applies to both models. On the left, state represents where the model is in a sequence. On the right, state represents different learned morphologies for an object.
+
+![Both behavior (left) and morphology models can be state-conditioned. This means that depending on the state, different features/changes are expected at different locations. State might be traversed through passing time or by applying actions.](../../figures/theory/state_conditioning.png#width=600px)
```
nielsleadholm, line 33: **suggestion:** For the morphology visualization on the right, it might be worth reversing the order of the two staplers shown. I think this would correspond better to the movement on the left example (stapler closing, rather than opening).

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
 [6] Hawkins, J. C., & Ahmad, S. (2024). *U.S. Patent No. 12,094,192*. **Inferencing and learning based on sensorimotor input data.** Washington, DC: U.S. Patent and Trademark Office.  
 [7] Clay, V., Leadholm, N., and Hawkins, J. (2024). **The thousand brains project: A new paradigm for sensorimotor intelligence.** URL: [https://arxiv.org/abs/2412.18354](https://arxiv.org/abs/2412.18354).  
 [8] Hawkins, J. C., Ahmad, S., Clay, V., & Leadholm, N. (2025). **Architecture and Operation of Intelligent System.** *U.S. Patent Application No. 18/751,199*.  
 [9] **Monty code:** [https://github.com/thousandbrainsproject/tbp.monty](https://github.com/thousandbrainsproject/tbp.monty)  
 [10] **TBP documentation**: [https://thousandbrainsproject.readme.io/docs/welcome-to-the-thousand-brains-project-documentation](https://thousandbrainsproject.readme.io/docs/welcome-to-the-thousand-brains-project-documentation) 
+[11] Hawkins, J., Leadholm, N. and Clay, V. (2025) **Hierarchy or Heterarchy? A Theory of Long-Range Connections for the Sensorimotor Brain** URL: [https://arxiv.org/pdf/2507.05888](https://arxiv.org/pdf/2507.05888)
```
nielsleadholm, line 135: **issue:** For some reason these two references aren't appearing on their own lines.

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
+ 
 As we have previously demonstrated [2, 7, 10], cortical columns can vote to quickly reach a consensus of the static object being observed. The same mechanism applies to behaviors. Multiple columns can vote to quickly infer behaviors.  
+
+Analogous to object models, behavior models can be recognized in any location, orientation, and scale by transforming the physical movement vector into the behavior's reference frame [11]. This allows for recognizing a behavior at different locations on an object in varying orientations and scales, and therefore represents a flexible way to apply and recognize behaviors in novel situations. Notably, the behavior can be recognized independently of the object on which it was learned.
+
+## Associating Morphology and Behavior Models
+
+Any particular object may exhibit multiple independent behaviors. For example, the top of a stapler can be raised or lowered and, independently, the stapler deflection plate can be rotated. A coffee maker may have a power switch, a lid to add water, and a grinds basket that swings out. Each of these parts exhibits its own behaviors.
```
nielsleadholm:
```suggestion
Any particular object may exhibit multiple independent behaviors. For example, the top of a stapler can be raised or lowered and, independently, the stapler deflection plate can be rotated. A coffee maker may have a power switch, a lid to add water, and a filter basket that swings out. Each of these parts exhibits its own behaviors.
```

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
+
+Note that the parent object, on the right in the figure, does not know if the child object is a morphology or behavioral object.
+
+![Analogous to learning compositional models of static objects, behaviors can be associated with locations on morphology models using hierarchical connections between two LMs with co-located inputs. This way, a detected behavior can bias the object (bottom-up) and a detected object can inform expected behaviors (top-down). ](../../figures/theory/associating_behavior_with_morphology.png#width=500px)
+
+The above examples illustrate that the two modeling systems, morphology/static and behavior/dynamic are similar and share many mechanisms and attributes. This commonality makes it easier to understand and implement the two systems. 
+
+## Sequence Learning and Representation of Time
+
+In many ways, the two modeling systems are similar. As mentioned, a major difference is that behavioral models require a temporal dimension. Behaviors are high-order sequences, and the time between sequence elements is often important.
```
nielsleadholm: **suggestion:** Again, I would maybe soften the distinction, e.g. "... behavioral models require a temporal dimension, whereas only some morphology models will make use of time." 

This also fits with the universality of columns. Otherwise we are creating an artificial distinction if only some types of columns need time.

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
  
+Previous work at Numenta showed how any layer of neurons can learn high-order sequences [1]. This mechanism will work well for learning behavioral sequences. In addition, there is a need for learning the time between sequence elements.
+
+Matrix cells in the thalamus could encode the time passed since the last event and broadcast this signal widely across the neocortex. Matrix cells project to L1, where they form synapses on the apical dendrites of L3 cells, allowing the behavioral model to encode the timing between behavioral states. The static model does not require time but could still learn state/context-conditioned models using the same mechanism.
```
nielsleadholm: **suggestion:** It might be worth mentioning the stoplight example here?

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
+
+![In addition to the local sensory and movement input, the column receives a more global time input to L1. This signal represents the amount of time passed since the last event. It is sent widely/globally and therefore synchronizes learning between many columns. A detected event in one learning module can reset the interval timer for all of them.](../../figures/theory/timing_input.png#width=500px)
 
-### Details
+During inference, the model can be used to speed up of slow down the global time signal. If input features arrive earlier than expected, the interval timer is sped up. If they arrive later than expected, the interval timer is slowed down. This allows for recognizing the same behavior at different speeds.
```
nielsleadholm:
```suggestion
During inference, the model can be used to speed up or slow down the global time signal. If input features arrive earlier than expected, the interval timer is sped up. If they arrive later than expected, the interval timer is slowed down. This allows for recognizing the same behavior at different speeds.
```

docs/theory/recent-progress/object-behaviors.md
nielsleadholm:
```suggestion
Matrix cells in the thalamus could encode the time passed since the last event and broadcast this signal widely across the neocortex. Matrix cells project to L1, where they form synapses on the apical dendrites of L3 and L5 cells, allowing the behavioral model to encode the timing between behavioral states. The static model does not require time but could still learn state/context-conditioned models using the same mechanism.
```

It's clear to me that the behavior *ID* (e.g. "opening/closing hinge") would be at L3, but I think it's still unclear whether the current state ("start of opening sequence" etc) is in L3 or L5, or indeed somewhere else.

docs/theory/recent-progress/object-behaviors.md
nielsleadholm: **suggestion:** When I first read this, I was left feeling that it would benefit from a more concrete description of the time proposal, but I see you provided that later in e.g. `global-interval-timer.md`. I wonder whether it's worth linking to that, or just saying something like "A more concrete description of our proposal for modelling time is provided in our Future Work sections, links of which can be found at the bottom of this page."

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
  
+For behavior models, we propose using the same mechanism. The main difference is that the LM modeling behaviors only receives input when changes are detected. This could be a local movement (e.g., a moving edge) or a feature change (e.g., color changing). It therefore encodes changes at locations. We use the same mechanism of iteratively testing hypotheses when inferring a behavior. Analogous to the object recognition mechanism, we apply a pose hypothesis to the sensed sensor displacement to transform it into the behavior model’s reference frame. We can then compare the sensed change to the stored change at that location.
```
nielsleadholm, line 72: **note:** I think we had some discussion about how a color change would be represented. Seeing green turn to red isn't really a smooth change that be generalized e.g. if the initial color is anything other than green. I think it might be best to leave out from here for now?

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
+![As the modeling mechanism of behaviors is analogous to that of static morphology, the main update to Monty would be to introduce a new change-detecting sensor module (SM). The LM that receives input from this new SM will learn behavior models.](../../figures/theory/monty_implementation_change_SM.png)
 
-Any particular object may exhibit multiple independent behaviors. For example, the top of a stapler can be raised or lowered and, independently, the stapler deflection plate can be rotated. A coffee maker may have a power switch, a lid to add water, and a coffee grounds basket that swings out. Each of these parts exhibit their own behaviors.
+Additionally, behavior models require a temporal input as well as conditioning the state of the model based on this input. This could be implemented as multiple graphs or sets of points in the same reference frame that are traversed as time passes. There are many possible ways to achieve this in code. The important thing is that the state in the temporal sequence can condition the changes to expect at a location.
```
nielsleadholm:
```suggestion
Additionally, models require a temporal input as well as conditioning the state of the model based on this input. This could be implemented as multiple graphs or sets of points in the same reference frame that are traversed as time passes. There are many possible ways to achieve this in code. The important thing is that the state in the temporal sequence can condition the changes to expect at a location.  
```

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -1,87 +1,133 @@
...
-Additionally, behavior models have a temporal dimension. This could be implemented as multiple graphs or sets of points in the same reference frame that are traversed as time passes. In code, there are many possible ways to achieve this. The important thing is that the timing in the temporal sequence can condition the changes to expect at a location.  
-Recognizing a behavior model in practice will likely depend more strongly on voting between multiple learning modules (at least for more complex behaviors where one sensor patch may not be able to observe a lot of the behavior fast enough on its own). The voting process for behavior models would work analogously to the voting process already implemented for object models. It would additionally require a shared temporal signal (the same as used internally in an LM) to keep the learning modules in sync.
+We can start with 1 and 2 for the first prototype. 3 should fall out of that solution (+ work on modeling compositional objects in general). 4 and 5 still have many unresolved questions and can be added in a second step.
```
nielsleadholm:
```suggestion
We can start with 1 and 2 for the first prototype. 3 should fall out of that solution (+ work on modeling compositional objects in general). 4 and 5 still have many unresolved questions and can be added in a second phase. 
```

docs/future-work/cmp-hierarchy-improvements/global-interval-timer.md
```diff
@@ -0,0 +1,48 @@
...
+description: Add a (semi) global interval timer that provides information about time elapsed since the last event to LMs.
+---
+
+This item relates to the broader goal of [modeling object behaviors in Monty](../../theory/recent-progress/object-behaviors.md#implementation-in-monty).
+
+We want to have an interval timer that exists separately from the current Monty components and provides input to a large group of LMs (or all).
+
+The interval timer is a bit like a stop watch that counts up time from a previous event and broadcasts this elapsed time widely. The time can be reset by significant events that are detected in LMs (or SMs).
+
+In the brain, this elapsed time might be encoded with time cells (e.g. see Kraus, 2013) or a similar mechanism, where different cells fire at different temporal delays from the start of an interval. Together they tile the temporal space (up to a certain max duration and with higher resolution for shorter durations). This means, than the LM receives different neural input from the timer, depending on how much time has passed since the start of this state.
```
nielsleadholm:
```suggestion
In the brain, this elapsed time might be encoded with time cells (e.g. see Kraus, 2013) or a similar mechanism, where different cells fire at different temporal delays from the start of an interval. Together they tile the temporal space (up to a certain max duration and with higher resolution for shorter durations). This means that the LM receives different neural input from the timer, depending on how much time has passed since the start of this state.
```

docs/future-work/cmp-hierarchy-improvements/global-interval-timer.md
```diff
@@ -0,0 +1,48 @@
...
+
+In the conceptual figure below, this mechanism is illustrated as the circle with little purple dots on it. Each significant event resets the circle to have the top neuron (darkest purple) active. Then, as time passes, the neurons on the circle become active in a clockwise direction. The ID of the active time neuron is the input to L1 of the cortical column.
+
+There is a default speed at which the timer goes through the different active states. This is the timing information stored during learning. During inference, however, the column can tell the interval timer to spin faster or slower, which means that it will receive the same neural input to L1 at different absolute times from the last event.
```
nielsleadholm: **thought:** In order for this to work in Monty, I think that like in the brain, the "timer" needs to have discrete points, as these points will form the basis of associative connections. Given the increasing intervals/lower resolution, this might look a bit like a Fibonachi sequence. Then we would associate these discrete points of very early, early, mid, late, very late, etc. with the behavioral states. 

In other words, in Monty it's not sufficient to just broadcast an actual time-since-event signal (e.g. 2.34 seconds), because this doesn't really mean anything on its own. However, with the default tempo, you could interpret the discrete time-stamps as actual seconds elapsed.

The above is described pretty clearly in the context of biology, but I think it would be worth clarifying this for the actual Monty implementation as well.

docs/future-work/cmp-hierarchy-improvements/global-interval-timer.md
```diff
@@ -0,0 +1,48 @@
...
+
+The clock is reset and starts ticking through the different neural activations as before. However, as we activate different time cells, we receive different inputs in L4 and associate these different inputs with the different time cells. This allows us to represent feature inputs at different offsets after the last significant event. When a significant event occurs (i.e. stapler reaches the bottom position and tacks some paper), the clock is reset and we move one forward in the element in the sequence (L5b)
+
+TODO: include video
+
+One other change here is that we are sensing in 2D (or 3D) space, so in addition to our position in the sequence, we have a location on the object (L6a, shown as grid here). This location is updated using sensor movement (our existing mechanism). In the example here, the sensor doesn’t move so we just store sequence information at that location but if we explore a behavior over and over again at different locations, this would become a richer representation of the entire object behavior.
```
nielsleadholm:
```suggestion
One other change here is that we are sensing in 3D space (2D in the example diagram), so in addition to our position in the sequence, we have a location on the object (L6a, shown as a grid here). This location is updated using sensor movement (our existing mechanism). In the example here, the sensor doesn’t move so we just store sequence information at that location but if we explore a behavior over and over again at different locations, this would become a richer representation of the entire object behavior.
```

docs/future-work/environment-improvements/object-behavior-test-bed.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Level 5:**
+- Behaviors that involve a change in features instead of movement
+
+
+## Initial Experiment Setup
+- 1 sensor patch that gets processed by 2 SMs (one for static and one for moving features) and sent to 2 LMs.
+- Supervision during learning -> provide object ID, pose & state in sequence at each step.
+- Evaluate accuracy & speed of recognizing behavior & morphology ID, pose & state (+visualize).
+- Could start with 2D environment to make the state dimension easier to visualize (keep in mind potential difficulty to extract movement in 3D from depth image).
```
nielsleadholm:
```suggestion
- Could start with 2D environment to make the state dimension easier to visualize (keep in mind potential difficulty in extracting movement from a 3D from depth image).
```

docs/future-work/environment-improvements/object-behavior-test-bed.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Level 3:**
+- One object has multiple behaviors at different locations
+
+**Level 4:**
+- Same behavior at varying speeds
+
+**Level 5:**
+- Behaviors that involve a change in features instead of movement
+
```
nielsleadholm, line 26: **suggestion:** Do we maybe want one on learning without the state in the sequence being provided as a supervisory signal?

docs/future-work/learning-module-improvements/event-detection-to-reset-timer.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Event Detection to Reset Timer
+description: Add ability to detect significant events which are used as signals to reset the global interval timer.
+---
+
+This item relates to the broader goal of [modeling object behaviors in Monty](../../theory/recent-progress/object-behaviors.md#implementation-in-monty). For a broader overview of the (semi)-global interval timer, see [this future work page](../cmp-hierarchy-improvements/global-interval-timer.md).
+
+Any LM should have the ability to detect significant events (such as the attack of a new note, an object reaching a canonical state, or a significant change in input features). If one such event is detected, the LM send a signal to the interval timer (outside the LM) which is then reset. The interval timer provides input to a large group of LMs and hence a detected event in one LM resets the input signal to many LMs.
```
nielsleadholm:
```suggestion
Any LM should have the ability to detect significant events (such as the attack of a new note, an object reaching a canonical state, or a significant change in input features). If one such event is detected, the LM (or LMs) send a signal to the interval timer (outside the LM) which is then reset. The interval timer provides input to a large group of LMs and hence a detected event in one LM can reset the input signal to many LMs.
```

I imagine there will be some threshold for resetting, and getting a signal from multiple LMs will tip the balance in favor of resetting. Without this, it is likely to be rather chaotic if a single LM can reset the global clock.

docs/future-work/learning-module-improvements/include-state-in-hypotheses.md
```diff
@@ -0,0 +1,18 @@
...
+description: Infer state of object, in addition to its pose.
+---
+
+This item relates to the broader goal of [modeling object behaviors in Monty](../../theory/recent-progress/object-behaviors.md#implementation-in-monty).
+
+Once we [integrate state into the models learned in LMs](include-state-in-models.md), we will also need to add state to our hypotheses to infer the state the object is in.
+
+To do this, a hypothesis needs to include which of the states/intervals we are in. Hence, they are represented as (l, r, s), where l is the location on the object, r is the rotation of the object, and s is the state in the sequence. In Monty, we would add a `states` variable to the `Hypotheses` class (in addition to the existing `locations` and `poses` arrays).
+
+Analogous to testing possible locations and possible poses, we would test possible states. Depending on sensory input, the evidence for different location-rotation-state combinations would be inceremented/decremented and hypotheses would be narrowed down.
```
nielsleadholm, line 12: **suggestion:** It might be worth adding a reference to `use-models-with-fewer-points.md`, noting that the additional dimension in hypothesis space means we need to significantly constrain the number of points in models in order to prevent an explosion in the hypothesis space size.

docs/future-work/learning-module-improvements/include-state-in-models.md
```diff
@@ -0,0 +1,18 @@
...
+description: Add a 'state' dimension to the models learned in Monty that conditions which features to expect at what locations.
+---
+
+This item relates to the broader goal of [modeling object behaviors in Monty](../../theory/recent-progress/object-behaviors.md#implementation-in-monty).
+
+Any model in an LM can be state conditioned. For instance a stapler might be open or closed, and an object behavior is represented as a sequence of states. Depending on the state of an object, the model expects to see different features at different locations.
+
+![Both behavior (left) and morphology models can be state-conditioned. This means that depending on the state, different features/changes are expected at different locations. State might be traversed through passing time or by applying actions.](../../figures/theory/state_conditioning.png#width=600px)
+
+In Monty, state can be interpreted as a sub-object ID. Depending on the state of an object Monty will look at the model stored for this state. (also see [include state in hypotheses](include-state-in-hypotheses.md)).
```
nielsleadholm:
```suggestion
In Monty, state can be interpreted as a sub-object ID. Depending on the state of an object Monty will look at the model stored for this state (also see [include state in hypotheses](include-state-in-hypotheses.md)). 
```

docs/future-work/learning-module-improvements/include-state-in-models.md
```diff
@@ -0,0 +1,18 @@
...
+
+Any model in an LM can be state conditioned. For instance a stapler might be open or closed, and an object behavior is represented as a sequence of states. Depending on the state of an object, the model expects to see different features at different locations.
+
+![Both behavior (left) and morphology models can be state-conditioned. This means that depending on the state, different features/changes are expected at different locations. State might be traversed through passing time or by applying actions.](../../figures/theory/state_conditioning.png#width=600px)
+
+In Monty, state can be interpreted as a sub-object ID. Depending on the state of an object Monty will look at the model stored for this state. (also see [include state in hypotheses](include-state-in-hypotheses.md)). 
+
+One alternative is to represent state as a 4th dimension in the models such that every location on the object is represented as (x, y, z, s). However, it seems like interpolation along the state/time dimension is likely limited, which may be hard to represent this way.
```
nielsleadholm:
```suggestion
One alternative is to represent state as a 4th dimension in the models such that every location on the object is represented as (x, y, z, s). However, it seems like interpolation along the state/time dimension is likely limited. Furthermore, measuring distance along the state dimension is likely quite different from along the three spatial dimensions. Neither of these points would be reflected well if using a continuous 4D model. 
```

docs/future-work/learning-module-improvements/include-state-in-models.md
```diff
@@ -0,0 +1,18 @@
...
+
+States are learned as an ordered sequence. The model of an object can include an ordered sequence of states as well as the temporal duration between states in the sequence. The interval length between two states is provided by the timer input to the LM and can be stores in the models. States might also be traversed by applying actions, however we don't have a concrete proposal for this yet.
+
+Both behavior & morphology models can have different states and sequences and both can be driven by time or other factors (there is no difference between the LMs besides their input).
```
nielsleadholm:
```suggestion
Both behavior & morphology models can have different states and sequences and both can be driven by time or other factors. In other words, there is no difference between the LMs besides their input.
```

docs/future-work/sensor-module-improvements/change-detecting-sm.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Change Detecting SM
+description: Add a new type of sensor module that detects local changes and output's those as CMP messages.
+---
+
+This item relates to the broader goal of [modeling object behaviors in Monty](../../theory/recent-progress/object-behaviors.md#implementation-in-monty). It also build on [detecting local and global flow](detect-local-and-global-flow.md) in the SM.
```
nielsleadholm:
```suggestion
This item relates to the broader goal of [modeling object behaviors in Monty](../../theory/recent-progress/object-behaviors.md#implementation-in-monty). It also builds on [detecting local and global flow](detect-local-and-global-flow.md) in the SM.
```

docs/future-work/sensor-module-improvements/change-detecting-sm.md
```diff
@@ -0,0 +1,18 @@
...
+
+As outlined in the [theory section on modeling object behaviors](../../theory/recent-progress/object-behaviors.md), the idea is that we can use the same LM for modeling object behaviors as we do for modeling static objects. The only difference is what kind of input the LM receives. An LM that receives input from the change detecting SM outlined here would be learning behavior models.
+
+![The introduction of the change detecting LM means that the LM that receives input from this new SM will learn behavior models.](../../figures/theory/monty_implementation_change_SM.png)
```
nielsleadholm, line 10: **suggestion:** I think it could be clearer to show the two components comprising the "morphological" feature of the change SM as analogous to the morphological (location + pose) feature given to the static SM. Right now, it looks like the orientation and direction of the change are non-morphological, optional features, and only the location is the key morphological feature. 

Screenshot 2025-10-14 at 1 54 45 PM

docs/theory/recent-progress/object-behaviors.md
vkakerbeck: Yeah we talk about this a bit more later but I can reword it here to sound less like this is only for behavior models

docs/theory/recent-progress/object-behaviors.md
vkakerbeck: Is there a specific reason you want to emphasize this? The main intend here is to explain how it's important for them to be in separate reference frames, whether they are in separate columns is not functionally important for that so I worry it would confuse people.

docs/theory/recent-progress/object-behaviors.md
vkakerbeck: Where are you seeing this? They look like they are all on their own line to me...
Screenshot 2025-10-15 at 11 56 04 AM

docs/theory/recent-progress/object-behaviors.md
vkakerbeck: Hm I am not sure the stoplight is the best example since it's morphology doesn't change, only it's features. Does that image of the two morphological staplers not help?

docs/theory/recent-progress/object-behaviors.md
vkakerbeck: I just tried to keep it as general as possible for now since I am not sure we completely ruled out change detection in feature space. We can always make it more constrained later.

docs/future-work/environment-improvements/object-behavior-test-bed.md
vkakerbeck: yes definitely some time. But I wasn't thinking of this as part of the first test setup. This is just an outline to test the plumbing of learning and inferring behavior models but with max supervision. Removing supervision during learning would be a next step but I think it still requires answering some conceptual questions like how do we recognize when a sequence repeats.

docs/future-work/sensor-module-improvements/change-detecting-sm.md
vkakerbeck: I thnk that could be confusing since we still want to send a location signal (i.e. where is the change detected). So we want to send "change@location" instead of "feature@location". I guess I could update the top part to show location and orientation separately and just remove the other features?

docs/theory/recent-progress/object-behaviors.md
nielsleadholm: I thought it was helpful since people who have read the previous version or watched our videos will have seen the idea that both reference frames are in the same column. Since we are not proposing that here, I thought it would be worth being explicit, as the first sentence is ambiguous in this regard.

docs/theory/recent-progress/object-behaviors.md
nielsleadholm: Ok interesting, on GitHub if I go to the `...` option and then "View File" to get a Markdown preview, I see the attached, but I agree if I go to the Readme .com preview that it is rendered correctly. Probably fine then. 

Screenshot 2025-10-15 at 2 57 43 PM

docs/theory/recent-progress/object-behaviors.md
```diff
@@ -52,15 +52,18 @@ The above examples illustrate that the two modeling systems, morphology/static a
 
 ## Sequence Learning and Representation of Time
 
-In many ways, the two modeling systems are similar. As mentioned, a major difference is that behavioral models require a temporal dimension. Behaviors are high-order sequences, and the time between sequence elements is often important.
+In many ways, the two modeling systems are similar. As mentioned, a major difference is that behavioral models require a temporal dimension, whereas our morphology models could do without it so far, although also those may make use of time and represent different states for the same object. Behaviors are high-order sequences, and the time between sequence elements is often important.
```
nielsleadholm:
```suggestion
In many ways, the two modeling systems are similar. As mentioned, a major difference is that behavioral models require a temporal dimension, whereas our morphology models could do without it so far, although also those may make use of time and represent different states for the same object. Behaviors are high-order sequences, and the time between sequence elements is often important.
```
```suggestion
In many ways, the two modeling systems are similar. As mentioned, a major difference is that behavioral models require a temporal dimension, whereas our morphology models could do without it so far, although they may also make use of time and represent different states for the same object. Behaviors are high-order sequences, and the time between sequence elements is often important.
```

docs/theory/recent-progress/object-behaviors.md
nielsleadholm: Ok yeah, happy to use the stapler example, I would maybe just call it out as a specific example again. I.e. that this is a model that, over time, can switch back and forth between two different states.

docs/future-work/environment-improvements/object-behavior-test-bed.md
nielsleadholm: Sounds good

docs/future-work/sensor-module-improvements/change-detecting-sm.md
nielsleadholm: Yeah I think that would be clearer.

docs/theory/recent-progress/object-behaviors.md
vkakerbeck: We were leaving it a bit ambiguous on purpose since this claim is less settled yet. It's a simplification for Monty and would be the simplest way to imagine it in the brain yet, but the sure thing that this paragraph focuses on is that separate reference frames are required. Maybe there would be cortical columns (defined by same sensor patch input) that can represent two RFs at the same time and get both static and dynamic input to the respective areas. That shouldn't invalidate the general idea presented here.

docs/future-work/sensor-module-improvements/change-detecting-sm.md
vkakerbeck: updated :)

PR #494, opened 2025-10-08
feat: GoalStateGenerator is optional for Learning Modules

tristanls-tbp:
This pull request makes GoalStateGenerator optional for a LearningModule. @scottcanoe mentioned that it is helpful during development to have only specific learning (or sensor) modules emit the goal state for easier evaluation and experimentation.

By making the LearningModule GoalStateGenerator optional, experimental configuration can essentially "turn off" goal state generation as desired.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -6,7 +6,7 @@
 
 - [@vkakerbeck](https://github.com/vkakerbeck)
 - [@nielsleadholm](https://github.com/nielsleadholm)
-- [@tristanls](https://github.com/tristanls)
+- [@tristanls-tbp](https://github.com/tristanls-tbp)
 - [@codeallthethingz](https://github.com/codeallthethingz)
 - [@scottcanoe](https://github.com/scottcanoe)
 - [@hlee9212](https://github.com/hlee9212)


src/tbp/monty/frameworks/models/monty_base.py
```diff
@@ -284,7 +284,9 @@ def _pass_goal_states(self):
         # Currently only use GSG outputs at inference
         if self.step_type == "matching_step":
             for lm in self.learning_modules:
-                self.gsg_outputs.append(lm.propose_goal_state())
+                goal_state = lm.propose_goal_state()
+                if goal_state is not None:
+                    self.gsg_outputs.append(goal_state)
```
tristanls-tbp, line 289: **note**: I assumed we don't want `None`s in `gsg_outputs`, but this is only an assumption. Please let me know if that's wrong.

tests/unit/evidence_lm_test.py
```diff
@@ -729,7 +729,7 @@ def get_elm_with_fake_object(
         self,
         fake_obs,
         initial_possible_poses="informed",
-        gsg_class=GraphGoalStateGenerator,
+        gsg_class=None,
```
tristanls-tbp, line 729: **note**: I changed this test to demonstrate that the code works without a goal state generator when it isn't needed. When a goal state generator is necessary, it is set elsewhere in the test file.

src/tbp/monty/frameworks/models/monty_base.py
scottcanoe: No, there's no reason to hang on to `None` goal states that I'm aware of. Note, however, that I decided to leave that up to the `GoalStateSelector` in my implementation.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -186,8 +186,9 @@ def __init__(
         )
         if gsg_args is None:
             gsg_args = {}
-        self.gsg = gsg_class(self, **gsg_args)
-        self.gsg.reset()
+        self.gsg = gsg_class(self, **gsg_args) if gsg_class is not None else None
```
scottcanoe: **suggestion:** I recommend allowing `gsg_args` to be `None` when `gsg_class` is non-`None`. For sensor modules, I did
```python
        if gsg_class is not None:
            gsg_args = gsg_args or {}
            self.gsg = gsg_class(self, **gsg_args)
        else:
            self.gsg = None
```
I forgot to add `self.gsg.reset()` though. No idea if it's actually expected or required though.

tristanls-tbp: Your suggestion looks better, I'll update.

tristanls-tbp:
Since @vkakerbeck is away this week, I'm removing her from Assignees.

Also, FYI @nielsleadholm .

PR #495, opened 2025-10-09
docs: clarify to install Windows version of VSCode on Windows

tristanls-tbp:
Received a report that it must be the Windows version of VS Code. Adding clarification to the docs.


tristanls-tbp:
Asking you @scottcanoe for review since you were there 😬.

PR #496, opened 2025-10-09
refactor!: extract snapshot telemetry tracking from DetailedLoggingSM

tristanls-tbp:
This pull request extracts tracking snapshots of raw observations and their poses from `DetailedLoggingSM` into a separate `SnapshotTelemetry` component.

Sensor modules no longer invoke `DetailedLoggingSM.step()` to snapshot the observations.

FYI @vkakerbeck @nielsleadholm

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -6,7 +6,7 @@
 
 - [@vkakerbeck](https://github.com/vkakerbeck)
 - [@nielsleadholm](https://github.com/nielsleadholm)
-- [@tristanls](https://github.com/tristanls)
+- [@tristanls-tbp](https://github.com/tristanls-tbp)
 - [@codeallthethingz](https://github.com/codeallthethingz)
 - [@scottcanoe](https://github.com/scottcanoe)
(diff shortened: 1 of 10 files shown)


src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -63,62 +107,33 @@ def __init__(
...
-                            sm_location=np.array(self.state["location"]),
-                        )
-                    )
+            self._snapshot_telemetry.raw_observation(
+                data,
+                self.state["rotation"],
+                self.state["location"]
+                if "location" in self.state.keys()
+                else self.state["position"],
+            )
```
tristanls-tbp, line 132: **note**: As we previously asserted `assert len(self.sm_properties) == len(self.raw_observations)`, it means that `if self.state is not None` was never a valid check, as it was invalid to append to `raw_observations` without appending to `sm_properties`. The updated code expects one of location or position to be always present, if not, it will crash, which is what the assertion wants anyway.

PR #499, opened 2025-10-09
refactor!: extract HabitatObservationProcessor from sensor module hierarchy

tristanls-tbp:
This pull request extracts the `HabitatObservationProcessor` component from the sensor module hierarchy. [According to our code style guide, we prefer to use composition over inheritance for reusing functionality](https://thousandbrainsproject.readme.io/docs/code-style-guide#composition-should-be-used-to-share-a-functionality-implementation-that-needs-its-own-state).

Notably, the hierarchy-related complexity is significantly reduced after this and #496. I'll be eliminating the hierarchy next.

FYI @nielsleadholm @vkakerbeck

### DIFF
--- a/docs/contributing/style-guide.md
+++ b/docs/contributing/style-guide.md
@@ -5,6 +5,10 @@ title: Style Guide
 
 See the [Code Style Guide](./style-guide/code-style-guide.md).
 
+# Typing Guide
+
+See the [Typing Guide](./style-guide/typing-guide.md).
+
 # GitHub Actions
 
 We use GitHub Actions to run our continuous integration workflows.
--- a/docs/contributing/style-guide/typing-guide.md
+++ b/docs/contributing/style-guide/typing-guide.md
@@ -1,0 +1,200 @@
+---
+title: Typing Guide
+---
+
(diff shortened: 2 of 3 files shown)


src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -566,10 +627,16 @@ def step(self, data):
...
+        observed_state, telemetry = self._habitat_observation_processor.process(
             data, on_object_only=self.on_object_obs_only
         )
 
+        if not self.is_exploring:
+            self.processed_obs.append(telemetry.processed_obs.__dict__)
+            self.states.append(self.state)
+            self.visited_locs.append(telemetry.visited_loc)
+            self.visited_normals.append(telemetry.visited_normal)
+
```
tristanls-tbp, line 638: **note**: This used to happen inside `observations_to_communication_protocol()`; however, I updated the pattern here so that the processor returns `telemetry` and we don't need to pollute the processor with `self.is_exploring`, `self.state`, `self.processed_obs`, `self.states`, `self.visited_locs`, or `self.visited_normals`, which are sensor module-specific concepts. These sensor module things will end up getting bundled into another telemetry helper soon, at the sensor module level (not the processor).

PR #502, opened 2025-10-13
feat: scipy rotation/quaternion conversions with `scalar_first` argument

scottcanoe:
Scipy added a `scalar_first` argument to `Rotation.as_quat` and `Rotation.from_quat` in version [1.14.0](https://scipy.github.io/devdocs/release/1.14.0-notes.html). It's a very useful option that eliminates the need to keep shuffling between wxyz or xyzw formats.

Until we can use scipy version >= 1.14.0, we can use the functions added in this PR.

This is one of several PRs targeting improvements to the spatial arithmetic and transformation utilities.


tristanls-tbp:
@scottcanoe, what error do you get attempting to upgrade to `scipy >= 1.14.0`?

scottcanoe:
@tristanls-tbp It's Python 3.8. We need >= 3.10.

```
Could not solve for environment specs
The following packages are incompatible
├─ pin on python 3.8.* =* * is installable and it requires
│  └─ python =3.8 *, which can be installed;
└─ scipy =1.14.1 * is not installable because there are no viable options
   ├─ scipy 1.14.1 would require
   │  └─ python >=3.10,<3.11.0a0 *, which conflicts with any installable versions previously reported;
   ├─ scipy 1.14.1 would require
   │  └─ python >=3.11,<3.12.0a0 *, which conflicts with any installable versions previously reported;
   ├─ scipy 1.14.1 would require
   │  └─ python >=3.12,<3.13.0a0 *, which conflicts with any installable versions previously reported;
   └─ scipy 1.14.1 would require
      └─ python >=3.13,<3.14.0a0 *, which conflicts with any installable versions previously reported.

Pins seem to be involved in the conflict. Currently pinned specs:
 - python=3.8
```

tristanls-tbp:
steve-carell-the-office

PR #503, opened 2025-10-13
refactor!: remove Sensor Module inheritance hierarchy

tristanls-tbp:
This pull request eliminates the Sensor Module hierarchy.

We are now only left with `Probe` and `HabitatSM`.

### DetailedLoggingSM
`DetailedLoggingSM` is renamed to `Probe`.  A probe can be inserted into Monty in place of a sensor module. It will track raw observations for logging, and can be used by experiments for positioning procedures, visualization, etc. What distinguishes a probe from a sensor module is that it does not process observations and does not emit a Cortical Message.

### HabitatDistantPatchSM
`HabitatDistantPatchSM` is renamed to `HabitatSM`.

### HabitatSurfacePatchSM
`HabitatSurfacePatchSM` is removed and replaced with `is_surface_sm` constructor argument to `HabitatSM`. So, previously, if you had `HabitatSurfacePatchSM`, you'll now instantiate `HabitatSM(is_surface_sm=True,...)`.

### FeatureChangeSM
`FeatureChangeSM` is removed and replaced with a `FeatureChangeFilter` that is contained within `HabitatSM`. So, previously, if you had `FeatureChangeSM(delta_thresholds={...},...)`, you'll now instantiate `HabitatSM(delta_thresholds={...},...)`. If `delta_thresholds` is `None` or omitted, then no filter is configured. Without a filter, `HabitatSM` behaves like `HabitatDistantPatchSM` or `HabitatSurfacePatchSM`, depending on `is_surface_sm`.

### NoiseMixin
`NoiseMixin` is removed and replaced with a `DefaultMessageNoise` that is contained within `HabitatSM`. To configure, pass in `noise_params` to `HabitatSM` as usual.

### Benchmarks

```diff
Experiment,Correct (%)|align right,Used MLH (%)|align right,Num Match Steps|align right,Rotation Error (degrees)|align right,Run Time (mins)|align right,Episode Run Time (s)|align right
base_config_10distinctobj_dist_agent,100.00,0.00,34,13.31,2,8
base_config_10distinctobj_surf_agent,100.00,0.00,26,15.03,2,12
randrot_noise_10distinctobj_dist_agent,100.00,3.00,52,22.41,3,22
randrot_noise_10distinctobj_dist_on_distm,99.00,1.00,37,12.27,2,17
randrot_noise_10distinctobj_surf_agent,100.00,1.00,29,20.12,3,20
randrot_10distinctobj_surf_agent,100.00,0.00,28,19.36,2,10
randrot_noise_10distinctobj_5lms_dist_agent,100.00,0.00,52,44.71,5,40
base_10simobj_surf_agent,93.57,8.57,70,12.34,5,27
randrot_noise_10simobj_dist_agent,84.00,40.00,234,31.08,11,89
randrot_noise_10simobj_surf_agent,93.00,34.00,172,22.81,14,120
-randomrot_rawnoise_10distinctobj_surf_agent,64.00,74.00,15,101.1,4,7
+randomrot_rawnoise_10distinctobj_surf_agent,67.00,77.00,15,117.12,4,7
base_10multi_distinctobj_dist_agent,80.71,14.29,30,19.60,3,1
```
Change in `randomrot_rawnoise_10distinctobj_surf_agent` is not due to this pull request, as running benchmark from current `main` outputs the new numbers as well: https://wandb.ai/thousand-brains-project/Monty/runs/wf72619g?nw=nwusertbp_tslominski

```diff
Experiment,Correct (%)|align right,Used MLH (%)|align right,Num Match Steps|align right,Rotation Error (degrees)|align right,Run Time (mins)|align right,Episode Run Time (s)|align right
base_77obj_dist_agent,93.51,12.99,107,16.46,20,68
base_77obj_surf_agent,98.70,6.49,56,11.38,12,34
-randrot_noise_77obj_dist_agent,89.18,22.00,147,37.37,32,117
+randrot_noise_77obj_dist_agent,89.18,22.08,147,37.37,32,117
randrot_noise_77obj_surf_agent,93.94,23.81,115,36.70,36,130
randrot_noise_77obj_5lms_dist_agent,87.01,0.00,64,60.61,10,95
```
Change in `randrot_noise_77obj_dist_agent` is a typo, original benchmark was `22.08` https://wandb.ai/thousand-brains-project/Monty/runs/ib26lo7a?nw=nwusertbp_tslominski

### DIFF
--- a/docs/how-monty-works/observations-transforms-sensor-modules.md
+++ b/docs/how-monty-works/observations-transforms-sensor-modules.md
@@ -19,7 +19,7 @@ The transformed, **raw input is then sent to the sensor module and turned into t
 | --------------------------------- | --------------- |
 | **SensorModule**                  | Abstract sensor module class. |
 | **HabitatSM**                     | Sensor module for HabitatSim. Extracts pose and features in CMP format from an RGBD patch. Keeps track of agent and sensor states. Also checks if observation is on object and should be sent to LM. Can be configured to add feature noise. |
-| **Probe**                   | A probe that can be inserted into Monty instead of a sensor module. It will track raw observations for logging, and can be used by experiments for positioning procedures, visualization, etc. What distinguishes a probe from a sensor module is that it does not process observations and does not emit a Cortical Message. |
+| **Probe**                   | A probe that can be inserted into Monty in place of a sensor module. It will track raw observations for logging, and can be used by experiments for positioning procedures, visualization, etc. What distinguishes a probe from a sensor module is that it does not process observations and does not emit a Cortical Message. |
 
 ## Noise
 
--- a/src/tbp/monty/frameworks/models/sensor_modules.py
+++ b/src/tbp/monty/frameworks/models/sensor_modules.py
@@ -382,7 +382,7 @@ class HabitatObservationProcessor:
 
 
 class Probe(SensorModule):
-    """A probe that can be inserted into Monty instead of a sensor module.
+    """A probe that can be inserted into Monty in place of a sensor module.
 
     It will track raw observations for logging, and can be used by experiments
     for positioning procedures, visualization, etc.


nielsleadholm:
Beautiful! I think `Probe` (rather than some variant of ...SM) was a good choice naming-wise for `DetailedLoggingSM`. Maybe `SensorProbe` would be more specific, but I can see the argument that if there are no other kinds of probes, then better to keep it short.

jeremyshoemaker:
> Beautiful! I think `Probe` (rather than some variant of ...SM) was a good choice naming-wise for `DetailedLoggingSM`. Maybe `SensorProbe` would be more specific, but I can see the argument that if there are no other kinds of probes, then better to keep it short.

It's worth remembering that it's located inside the `tbp.monty.frameworks.models.sensor_modules` Python module, so the fact that it's related to sensor modules is already encoded in its fully qualified name. That's kinda the point of namespaces like Python modules, you don't need to put all the details in the name to disambiguate something like you'd have to do in a language that doesn't have namespaces (e.g. C).

tristanls-tbp:
Note:

```diff
-randomrot_rawnoise_10distinctobj_surf_agent,64.00,74.00,15,101.1,4,7
+randomrot_rawnoise_10distinctobj_surf_agent,67.00,77.00,15,117.12,4,7
```

This was caused by https://github.com/thousandbrainsproject/tbp.monty/pull/490, as demonstrated by https://wandb.ai/thousand-brains-project/Monty/runs/jv2s32pa?nw=nwusertbp_tslominski

nielsleadholm:
@tristanls-tbp  Good to know, my guess is that would have been just a change in the effective seed due to the raw-noise which is added being sampled within the new `simulator.py` for-loop over actions.

PR #509, opened 2025-10-16
docs: update missed benchmark change due to #490

tristanls-tbp:
While working on #503, we discovered that #490 caused a benchmark change that was not included in the pull request.

The purpose of this pull request is to update benchmark results and identify the cause for the benchmark change.

The benchmark run https://wandb.ai/thousand-brains-project/Monty/runs/jv2s32pa/overview demonstrates the change as caused by 4a4ccdcb2faddb39736bede9146139d886fc05ec.

@nielsleadholm [comments](https://github.com/thousandbrainsproject/tbp.monty/pull/503#issuecomment-3410240333):
> @tristanls-tbp Good to know, my guess is that would have been just a change in the effective seed due to the raw-noise which is added being sampled within the new simulator.py for-loop over actions.


vkakerbeck (approved):
LGTM. Thanks for catching and updating!

scottcanoe (approved):
LGTM

PR #515, opened 2025-10-21
chore: Target ruff rules F841, PIE807, SIM910, and UP008

mf-maymay:
- [F841](https://docs.astral.sh/ruff/rules/unused-variable/): unused-variable
- [PIE807](https://docs.astral.sh/ruff/rules/reimplemented-container-builtin/): reimplemented-container-builtin
- [SIM910](https://docs.astral.sh/ruff/rules/dict-get-with-none-default/): dict-get-with-none-default
- [UP008](https://docs.astral.sh/ruff/rules/super-call-with-parameters/): super-call-with-parameters

These changes are intended to improve the clarity and simplicity of the code.

F841 simply asks for unused variables to be removed. There was only one instance of this rule being violated.

PIE807 asks for `lambda: []` and `lambda: {}` to be replaced with the built-in factory functions, `list` and `dict`.

SIM910 asks for the default `None` default to be omitted from `dict.get`. There were no instances of this rule being violated.

UP008 asks for legacy `super(MyClass, self)` calls to be replaced with the modern `super()` style. (I have seen the legacy style lead to bugs in cargo cult code, when the class name was not updated or the `self` was omitted by accident. We can avoid these problems and also have prettier code by enforcing this rule.)

### DIFF
--- a/docs/how-to-use-monty/tutorials/multiple-learning-modules.md
+++ b/docs/how-to-use-monty/tutorials/multiple-learning-modules.md
@@ -44,6 +44,7 @@ from tbp.monty.frameworks.environments import embodied_data as ED
 from tbp.monty.frameworks.experiments import (
     MontySupervisedObjectPretrainingExperiment,
 )
+from tbp.monty.frameworks.models.motor_policies import NaiveScanPolicy
 from tbp.monty.simulators.habitat.configs import (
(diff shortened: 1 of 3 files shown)


tristanls-tbp:
Thank you for addressing these 🙏.

PR #517, opened 2025-10-23
style!: Use PEP 604 union syntax

mf-maymay:
## Summary

This PR replaces `typing.Union[X, Y]` with the PEP 604 `X | Y` syntax throughout the codebase. These changes make for prettier code that is better aligned with newer versions of Python.

As a side effect of the required `from __future__ import annotations` import, type annotations in affected modules are now stored as strings (e.g., `"int"` instead of `int`). This affects calls to `create_dataclass_args`, hence why I am marking this as a breaking change.

## Changes Made

- Added `from __future__ import annotations` to modules that used `typing.Union`
- Replaced instances of `typing.Union[X, Y]` with `X | Y` per [UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/)
- Moved standard library imports used only for type hints to `if TYPE_CHECKING` blocks per [TC003](https://docs.astral.sh/ruff/rules/typing-only-standard-library-import/)
- Updated `CreateDataclassArgsTest` unit tests to reflect that dataclass field types are now strings (when not inferred from defaults)
    Screenshot 2025-10-25 at 1 19 24 PM
- Reorganized `CreateDataclassArgsTest` for improved clarity and code locality

## Breaking Change: `create_dataclass_args`

The `create_dataclass_args` function relies on runtime inspection of type annotations. With `from __future__ import annotations`, these annotations are now stored as strings rather than type objects. If any code is calling `create_dataclass_args` on functions defined in the modified modules, it will receive stringified types instead of actual type objects.

That said, dataclass field types don't actually matter at runtime (unlike Pydantic models), so the practical impact should be minimal.

## Why These Changes Are Safe

The ruff rule [UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/) marks this fix as "unsafe" because it can cause runtime errors with libraries like Pydantic that rely on runtime type annotations. However, this warning doesn't apply here: we don't use runtime type inspection in ways that would break with stringified annotations.

The `X | Y` syntax only works with valid types, so we now have errors in places where we already had errors (e.g., `dataclass | Dict` was already invalid because `dataclass` isn't a type).

Regarding the imports moved to `if TYPE_CHECKING` blocks: this is done to reduce runtime overhead. If we don't like these changes, we can add `TC003` to the ruff ignore list.

## Additional Context

**Other modernizations not applied:** Technically, the `annotations` feature also allows for PEP 585 syntax (`list[str]` instead of `List[str]`), but ruff isn't picking up on this despite the `__future__` import.

**Future plans:** One of the rules we currently ignore, [FA100](https://docs.astral.sh/ruff/rules/future-rewritable-type-annotation/), asks that we add `from __future__ import annotations` wherever we could benefit from newer syntax. This PR addresses the majority of violations, so we could consider un-ignoring it soon.

### DIFF
--- a/src/tbp/monty/__init__.py
+++ b/src/tbp/monty/__init__.py
@@ -8,4 +8,4 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
-__version__ = "0.11.0"
+__version__ = "0.12.0"


tests/unit/dataclass_utils_test.py
```diff
@@ -188,45 +170,56 @@ def test_deep_nested_dataclass(self):
...
+
+        self.assertTrue(dataclasses.is_dataclass(dc))
+        self.assertEqual(dc.__name__, "test1")
+
         expected = {
-            "field1": (str, dataclasses.MISSING),
-            "field2": (int, dataclasses.MISSING),
+            "field1": ("str", dataclasses.MISSING),
+            "field2": ("int", dataclasses.MISSING),
         }
```
goingmaymay, line 184: This diff reveals the breaking change. `tests/unit/dataclass_utils_test.py` now has a `from __future__ import annotations` import, resulting in `sample_function`'s annotations being stored as strings, not types.

This should be inconsequential. The types for dataclass fields are merely metadata. `dc` can still be instantiated without any problems.

tests/unit/dataclass_utils_test.py
```diff
@@ -188,45 +170,56 @@ def test_deep_nested_dataclass(self):
 
 class CreateDataclassArgsTest(unittest.TestCase):
     def test_simple_functions(self):
-        # Test simple function
+        def sample_function(field1: str, field2: int):
+            pass
+
```
jeremyshoemaker, line 175: I'm definitely a fan of moving these into the tests that use them when they aren't shared. 😁

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -1195,7 +1198,7 @@ def make_multi_lm_monty_config(
     sensor_module_args: Optional[Mapping],
     motor_system_class: type,
     motor_system_args: Optional[Mapping],
-    monty_args: Optional[Union[Mapping, MontyArgs]],
+    monty_args: Optional[Mapping | MontyArgs],
```
tristanls-tbp, line 1201: **suggestion (non-blocking)**: Seems this would be better suited by `Mapping | MontyArgs | None`, but we can save that for later `Optional[...]` to `... | None` change.

jeremyshoemaker: I almost mentioned that myself, but thought it would be better done in a separate PR, since we have several other places where we use `Optional`.

goingmaymay: Yes, I was thinking myself that it was weird that ruff didn't automatically switch from `Optional[...]` to `... | None`, but it also made for fewer changes in this PR.

I'll be looking to make this change, as well as the one from `List[str]` to `list[str]`, in future PRs.

PR #518, opened 2025-10-25
refactor!: remove unused union_of_possible_matches

ramyamounir:
In `MontyForGraphMatching._vote()` we have a call to a function that sets a `union_of_possible_matches` state. It doesn't look like this state is doing anything. Terminal condition checks do not use it and 5lm experiments run the same without it. 

I'm proposing to remove the function and other references to this state. I'm marking it as a breaking change just in case someone is using it in their fork or code outside of `tbp.monty`, but I'm not sure if it is a breaking change.

### DIFF
--- a/src/tbp/monty/frameworks/models/graph_matching.py
+++ b/src/tbp/monty/frameworks/models/graph_matching.py
@@ -399,13 +399,13 @@ class MontyForGraphMatching(MontyBase):
                 self.update_stats_after_vote(self.learning_modules[i])
 
         # Log possible matches
-        for i, lm in enumerate(self.learning_modules):
+        for lm in self.learning_modules:
             pm = (
                 lm.get_possible_matches()
                 if lm.buffer.get_num_observations_on_object()
                 else []
             )
-            logger.info(f"Possible matches for LM {i}: {pm}")
+            logger.info(f"Possible matches for {lm.learning_module_id}: {pm}")
 
     def _pass_infos_to_motor_system(self):
         """Pass input observations to the motor system.


ramyamounir:
Thanks @vkakerbeck. [Issue#527](https://github.com/thousandbrainsproject/tbp.monty/issues/527) is opened, and the logging is added in [95367f43871c](https://github.com/ramyamounir/tbp.monty/commit/95367f43871ced252b92990f66493a6ed1d1b040)

Note that I kept the logging in the same place after voting to preserve the same behavior since the possible matches are [updated during voting](https://github.com/thousandbrainsproject/tbp.monty/blob/2893df319b4a1fc0a463bd0fbd94773bc469a326/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L325).

vkakerbeck:
Nice, thanks for adding this! I think since we are not looking at unions here you don't need to convert to sets. To simplify it further you could also just add something like `logger.info(f"Possible matches for {self.learning_module_id}: {pm}")` in the `_threshold_possible_matches` function instead of looping over lms in the Monty class. It wouldn't exactly replicate the log from before but still provide the same information to the user and be a bit simpler. But either way works for me.

ramyamounir:
I tried to avoid logging in the `_threshold_possible_matches` because we call it twice; once [after stepping the LMs](https://github.com/thousandbrainsproject/tbp.monty/blob/95367f43871ced252b92990f66493a6ed1d1b040/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L737) and another time [after voting](https://github.com/thousandbrainsproject/tbp.monty/blob/95367f43871ced252b92990f66493a6ed1d1b040/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L325). I'm not really sure we need to call it after stepping the LMs, I think calling it after voting is enough since we only use this for checking TS. But maybe I missed something.

> you don't need to convert to sets

You're right, I will get rid of the sets.

vkakerbeck:
Yes, good point. But I guess you could move it into the previous loop over LMs or even the `update_stats_after_vote` call to avoid a second loop over lms.

ramyamounir:
I was also trying to avoid that since these are [only called](https://github.com/thousandbrainsproject/tbp.monty/blob/95367f43871ced252b92990f66493a6ed1d1b040/src/tbp/monty/frameworks/models/graph_matching.py?plain=1#L388) if the `lm_to_lm_vote_matrix` is defined, that's why I kept it outside. We can do some refactoring in another PR, I think we can do a single call to update the possible matches before the TS check, then we would be able to put the logging there.

vkakerbeck:
Okay, sounds good, it's not the most important thing anyways where that call is :D sorry about all the back and forth!

src/tbp/monty/frameworks/models/graph_matching.py
```diff
@@ -427,8 +398,14 @@ def _vote(self):
...
-        # Update IoPM, needed for checking terminal condition
-        self.union_of_possible_matches = self._get_union_of_possible_matches()
+        # Log possible matches
+        for i, lm in enumerate(self.learning_modules):
+            pm = (
+                set(lm.get_possible_matches())
+                if lm.buffer.get_num_observations_on_object()
+                else set()
+            )
+            logger.info(f"Possible matches for LM {i}: {pm}")
```
tristanls-tbp: **issue**: `i` is not a valid learning module identifier. Please use `learning_module_id`.

ramyamounir: fixed in [ab01f6e66eed](https://github.com/ramyamounir/tbp.monty/commit/ab01f6e66eede65406e6aca7810905ac90c1aa75)

PR #521, opened 2025-10-27
refactor!: AgentID newtype

tristanls-tbp:
This pull request introduces `AgentID` newtype to clarify the code and reduce the likelihood of errors.

This work is part of the preparation for the motor system to swap out policies based on received goals.

## Benchmarks

Screenshot 2025-10-28 at 10 13 16

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -207,7 +207,6 @@ ignore = [
     # Original inherited flake8 ignores
     "D1", # D1XX: Missing Docstrings
     "ERA001", # ERA001: Found commented out code
-    "F541", # F541: f-string is missing placeholders
     "FIX001", # FIX001: Line contains FIXME, consider resolving the issue
     "FIX002", # FIX002: Line contains TODO, consider resolving the issue
     "N812", # N812: lowercase imported as non lowercase. Allow "import torch.nn.functional as F"
@@ -262,7 +261,6 @@ ignore = [
     "PLR1714", # PLR1714: Consider merging multiple comparisons: {expression}. Use a set if the elements are hashable.
     "PLR1730", # PLR1730: Replace `if` statement with {replacement}
     "PLR2004", # PLR2004: Magic value used in comparison, consider replacing {value} with a constant variable
-    "PLW0127", # PLW0127: Self-assignment of variable
     "PT009", # PT009: Use a regular `assert` instead of unittest-style `assertEqual`
     "PT018", # PT018: Assertion should be broken down into multiple parts
     "PT027", # PT027: Use `pytest.raises` instead of unittest-style `assertRaises`
@@ -278,8 +276,6 @@ ignore = [
     "RET503", # RET503: Missing explicit `return` at the end of function able to return non-`None` value
     "RET504", # RET504: Unnecessary assignment to {name} before return statement
     "RET505", # RET505: Unnecessary {branch} after return statement
-    "RET506", # RET506: Unnecessary `else` after `raise` statement
-    "RET508", # RET508: Unnecessary `else` after `break` statement
     "RSE102", # RSE102: Unnecessary parentheses on raised exception
     "RUF005", # RUF005: Consider {expression} instead of concatenation
     "RUF010", # RUF010: Use explicit conversion flag
--- a/src/tbp/monty/frameworks/models/displacement_matching.py
+++ b/src/tbp/monty/frameworks/models/displacement_matching.py
@@ -518,7 +518,7 @@ class DisplacementGraphMemory(GraphMemory):
             graph_id: Name of the object.
             input_channel: ?
         """
-        logger.info(f"Adding a new graph to memory.")
+        logger.info("Adding a new graph to memory.")
 
         model = GraphObjectModel(
             object_id=graph_id,
--- a/src/tbp/monty/frameworks/models/evidence_matching/graph_memory.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/graph_memory.py
@@ -130,7 +130,7 @@ class EvidenceGraphMemory(GraphMemory):
             graph_id: name of new graph.
             input_channel: ?
         """
-        logger.info(f"Adding a new graph to memory.")
+        logger.info("Adding a new graph to memory.")
 
         model = GridObjectModel(
             object_id=graph_id,
--- a/src/tbp/monty/frameworks/models/graph_matching.py
+++ b/src/tbp/monty/frameworks/models/graph_matching.py
@@ -1375,7 +1375,7 @@ class GraphMemory(LMMemory):
             graph_id: name of new graph.
             input_channel: ?
         """
-        logger.info(f"Adding a new graph to memory.")
+        logger.info("Adding a new graph to memory.")
         model = GraphObjectModel(
             object_id=graph_id,
         )
--- a/src/tbp/monty/frameworks/models/motor_policies.py
+++ b/src/tbp/monty/frameworks/models/motor_policies.py
@@ -1017,11 +1017,10 @@ class NaiveScanPolicy(InformedPolicy):
             # python loop statements using iterators.
             # See https://docs.python.org/3/library/exceptions.html#StopIteration
             raise StopIteration()
-        else:
-            self.check_cycle_action()
-            action = self._naive_scan_actions[self.current_action_id]
+
+        self.check_cycle_action()
         self.step_on_action += 1
-        return action
+        return self._naive_scan_actions[self.current_action_id]
 
     def pre_episode(self):
         super().pre_episode()
@@ -1274,7 +1273,7 @@ class SurfacePolicy(InformedPolicy):
             # This is determined by some logic in embodied_data.py, in particular
             # the next method of InformedEnvironmentDataLoader
 
-        elif self.last_surface_policy_action is None:
+        if self.last_surface_policy_action is None:
             logger.debug(
                 "Object coverage good at initialization: "
                 + str(
--- a/src/tbp/monty/frameworks/models/sensor_modules.py
+++ b/src/tbp/monty/frameworks/models/sensor_modules.py
@@ -726,7 +726,7 @@ class FeatureChangeFilter(StateFilter):
         if not state.get_on_object():
             # Even for the surface-agent sensor, do not return a feature for LM
             # processing that is not on the object
-            logger.debug(f"No new point because not on object")
+            logger.debug("No new point because not on object")
             return False
 
         for feature in self._delta_thresholds:
--- a/src/tbp/monty/frameworks/models/states.py
+++ b/src/tbp/monty/frameworks/models/states.py
@@ -82,7 +82,7 @@ class State:
                         repr_string += f"           {vector}\n"
                 else:
                     repr_string += f"       {feature}: {feat_val}\n"
-        repr_string += f"   Non-Morphological Features: \n"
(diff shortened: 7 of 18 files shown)


src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -509,47 +514,51 @@ def get_omniglot_eval_dataloader(
 
 @dataclass
 class SensorAgentMapping:
-    agent_ids: List[str]
+    agent_ids: list[AgentID]
     sensor_ids: List[str]
-    sensor_to_agent: Dict
+    sensor_to_agent: dict[str, AgentID]
```
jeremyshoemaker, line 519: It might be out of scope, but do these `str` keys represent something that we might want a newtype for? It might be part of a larger refactor though, so I'm fine with leaving it off for now.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
```diff
@@ -932,12 +941,14 @@ def make_multi_sensor_mount_config(
         converted to lists.
 
     """
+    if agent_id is None:
```
jeremyshoemaker:
```suggestion
    if not agent_id:
```
This will also catch the case of an empty string.

src/tbp/monty/frameworks/config_utils/policy_setup_utils.py
```diff
@@ -153,6 +155,8 @@ def make_base_policy_config(
     Returns:
         BasePolicyConfig instance
     """
+    if agent_id is None:
```
jeremyshoemaker: Same here.
```suggestion
    if not agent_id:
```

src/tbp/monty/frameworks/config_utils/policy_setup_utils.py
```diff
@@ -201,6 +205,8 @@ def make_informed_policy_config(
     Returns:
         InformedPolicyConfig instance
     """
+    if agent_id is None:
```
jeremyshoemaker:
```suggestion
    if not agent_id:
```

src/tbp/monty/frameworks/config_utils/policy_setup_utils.py
```diff
@@ -227,6 +233,8 @@ def make_naive_scan_policy_config(step_size: float, agent_id="agent_id_0"):
     Returns:
         NaiveScanPolicyConfig instance
     """
+    if agent_id is None:
```
jeremyshoemaker:
```suggestion
    if not agent_id:
```

src/tbp/monty/frameworks/config_utils/policy_setup_utils.py
```diff
@@ -276,6 +284,8 @@ def make_surface_policy_config(
     Returns:
         SurfacePolicyConfig instance
     """
+    if agent_id is None:
```
jeremyshoemaker:
```suggestion
    if not agent_id:
```

src/tbp/monty/frameworks/config_utils/policy_setup_utils.py
```diff
@@ -335,6 +345,8 @@ def make_curv_surface_policy_config(
     Returns:
         SurfaceCurveInformedPolicyConfig instance
     """
+    if agent_id is None:
```
jeremyshoemaker:
```suggestion
    if not agent_id:
```

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -603,7 +604,7 @@ def get_good_view_with_patch_refinement(self) -> bool:
         """
         self.get_good_view("view_finder")
         for patch_id in ("patch", "patch_0"):
-            if patch_id in self._observation["agent_id_0"].keys():
+            if patch_id in self._observation[AgentID("agent_id_0")].keys():
```
jeremyshoemaker: We don't need to call `keys()` when checking if a dictionary contains a specific key.
```suggestion
            if patch_id in self._observation[AgentID("agent_id_0")]:
```

src/tbp/monty/frameworks/experiments/data_collection_experiments.py
```diff
@@ -54,7 +54,7 @@ def pass_features_to_motor_system(self, observation, step):
             None
             if self.model.motor_system._policy.action is None
             else (
-                f"{self.model.motor_system._policy.action.agent_id}."
+                f"{str(self.model.motor_system._policy.action.agent_id)}."
```
jeremyshoemaker: Does this not work without the `str` call? I'd expect that the representation of a newtype is the representation of the underlying type.

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -186,7 +186,7 @@ def report_episode(self, data, output_dir, episode, mode="train", **kwargs):
                     else:
                         o[key] = value
                 actions[i][0] = {
-                    f"{a.agent_id}": {"action": a.name, "params": json.dumps(o)}
+                    f"{str(a.agent_id)}": {"action": a.name, "params": json.dumps(o)}
```
jeremyshoemaker: Same question about the `str` call.

src/tbp/monty/simulators/habitat/actuator.py
```diff
@@ -63,7 +64,7 @@ def action_name(self, action: Action) -> str:
 
         The action name is prefixed by the agent ID.
         """
-        return f"{action.agent_id}.{action.name}"
+        return f"{str(action.agent_id)}.{action.name}"
```
jeremyshoemaker: Same question about the `str` call.

src/tbp/monty/simulators/habitat/agents.py
```diff
@@ -173,7 +176,7 @@ def get_action_space(self, spec: AgentConfiguration) -> AgentConfiguration:
 
         spec.action_space = {}
         for action in action_space:
-            spec.action_space[f"{self.agent_id}.{action[0]}"] = ActionSpec(
+            spec.action_space[f"{str(self.agent_id)}.{action[0]}"] = ActionSpec(
```
jeremyshoemaker: Same question about the `str` call.

src/tbp/monty/simulators/tacto/agents.py
```diff
@@ -79,16 +80,16 @@ def __init__(
     def get_spec(self):
         spec = super().get_spec()
         spec.action_space = {
-            f"{self.agent_id}.move_forward": ActionSpec(
+            f"{str(self.agent_id)}.move_forward": ActionSpec(
```
jeremyshoemaker: More calls to `str` in f-strings in this file that I'm not sure are needed.

tests/unit/habitat_data_test.py
```diff
@@ -34,34 +36,34 @@
...
 EXPECTED_STATES = np.random.rand(DATASET_LEN, 64, 64, 1)
 EXPECTED_ACTIONS_DIST = (
-    f"{AGENT_ID}.look_down",
-    f"{AGENT_ID}.look_up",
-    f"{AGENT_ID}.move_forward",
-    f"{AGENT_ID}.turn_left",
-    f"{AGENT_ID}.turn_right",
-    f"{AGENT_ID}.set_agent_pose",
-    f"{AGENT_ID}.set_sensor_rotation",
+    f"{str(AGENT_ID)}.look_down",
```
jeremyshoemaker: More `str` calls in f-strings.

tests/unit/habitat_sim_test.py
```diff
@@ -298,13 +300,13 @@ def test_states(self):
         sensor_rot = qt.from_rotation_vector(sensor_spec.orientation)
 
         # Compute rotation quartenions
-        turn_left_spec = agent_config.action_space[f"{agent_id}.turn_left"]
+        turn_left_spec = agent_config.action_space[f"{str(agent_id)}.turn_left"]
```
jeremyshoemaker: More `str` calls in f-strings.

jeremyshoemaker (approved):
There a few minor changes I highlighted.

The one about the `str` calls inside f-strings, from what I can tell testing in a REPL, these aren't needed, because at runtime all the NewTypes don't really exist. So we're basically calling `str` on a string.

That brings me to my second question. How do we define a breaking change? These changes only affect the type checking, not the runtime behaviour of the code. Does that count as a breaking change or not? I could see it either way, but since we aren't enforcing type checking at the moment, it doesn't seem like it would break anything. I'm fine with leaving it as a breaking change, I just wanted to raise the question.

src/tbp/monty/frameworks/experiments/data_collection_experiments.py
jeremyshoemaker: I wrote this before I tested it. It seems to work like I expected, so the calls to `str` aren't needed.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
tristanls-tbp: Yes, these will eventually become `SensorID`.

src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
tristanls-tbp: The intent is for this to be a default parameter, but it serves as a workaround for the problem that `AgentID("agent_id_0")` cannot be a default parameter because it is technically a function. It is not intended as a validation.

src/tbp/monty/frameworks/environments/embodied_data.py
tristanls-tbp: Feels unrelated to this pull request, but I can update it since I'm touching the line.

src/tbp/monty/frameworks/experiments/data_collection_experiments.py
tristanls-tbp: Good point, I'll remove these.

src/tbp/monty/frameworks/environments/embodied_data.py
jeremyshoemaker: Yeah, it could wait until we implement the lint for it. I'm fine either way.

tristanls-tbp:
> That brings me to my second question. How do we define a breaking change? These changes only affect the type checking, not the runtime behaviour of the code. Does that count as a breaking change or not? I could see it either way, but since we aren't enforcing type checking at the moment, it doesn't seem like it would break anything. I'm fine with leaving it as a breaking change, I just wanted to raise the question.

@jeremyshoemaker, the type signatures of several things have changed; the new types are not interchangeable, as the expectations are now stricter. This is a breaking change. [Semver defines a breaking change as "incompatible API changes"](https://semver.org/#summary). A type change is an incompatible API change. The contract has changed. It is not about breaking the runtime; it is about breaking the contract with the users.

A non-breaking change would be: `AgentID: TypeAlias = str` or `AgentID = Annotated[str, "AgentID"]`.

tristanls-tbp:
With the current `ruff` version, doing the following results in a lint issue:

```python
def __init__(self, agent_id: AgentID = AgentID("agent_id_0"))
```

This is resolved in `ruff >= 0.14.2`. 

For now, I added `# noqa: B008` to ignore, as a `str` newtype is safe. As a follow-up pull request, I will update `ruff`, at which point I'll remove the lint ignores.

PR #522, opened 2025-10-28
chore: Target ruff rules F541, PLW0127, RET506, and RET508

mf-maymay:
- [F541](https://docs.astral.sh/ruff/rules/f-string-missing-placeholders/): f-string-missing-placeholders
- [PLW0127](https://docs.astral.sh/ruff/rules/self-assigning-variable/): self-assigning-variable
- [RET506](https://docs.astral.sh/ruff/rules/superfluous-else-raise/): superfluous-else-raise
- [RET508](https://docs.astral.sh/ruff/rules/superfluous-else-break/): superfluous-else-break

F541 asks that all f-strings have placeholders. No `f"hello world"` allowed, only `f"hello, {name}"`.
While addressing this rule, I noticed a case where two strings were separated by commas when it seemed that they should instead be one string. I opted to concatenate these strings, as I suspect was intended. (See my first commit.)

PLW0127 asks that we don't say `x = x`, since that doesn't accomplish anything.
There was only one instance of this rule being violated.

RET506 and RET508 both try to reduce unnecessary indentation/nesting due to superfluous `else` blocks:
- RET506 asks to omit `else` when the preceding `if` block raises an exception.
- RET508 asks to omit `else` when the preceding `if` block ends in a `break`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -276,7 +276,6 @@ ignore = [
     "RET503", # RET503: Missing explicit `return` at the end of function able to return non-`None` value
     "RET504", # RET504: Unnecessary assignment to {name} before return statement
     "RET505", # RET505: Unnecessary {branch} after return statement
-    "RET506", # RET506: Unnecessary `else` after `raise` statement
     "RSE102", # RSE102: Unnecessary parentheses on raised exception
     "RUF005", # RUF005: Consider {expression} instead of concatenation
     "RUF010", # RUF010: Use explicit conversion flag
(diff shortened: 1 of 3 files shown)


jeremyshoemaker (commented):
Honestly, I'd like to turn on all the RET* lints, but that might be a larger set of changes. Did you consider turning on [RET505](https://docs.astral.sh/ruff/rules/superfluous-else-return/#superfluous-else-return-ret505) since it's similar to the two you're doing here? I know there are a few places that will need to change for that one.

goingmaymay:
@jeremyshoemaker Yes, I want to target all of those rules.

The RET505 rule is really nice, but it has 59 violations, so I decided not to address it in this PR. I'm trying to keep the diffs readable and small enough to avoid merge conflicts.

I think targeting the rest of the RET* rules in a next PR would be nice and clean. That would cover 171 violations.

jeremyshoemaker (approved):

PR #523, opened 2025-10-28
feat: allow episodes selection in `run_parallel`

ramyamounir:
After running a full experiment with all episodes, it’s common to notice a specific episode or range of episodes in the W&B logs that show unusual or interesting performance. Until now, rerunning or debugging those specific episodes required rerunning the entire experiment and marking them for detailed logging.

This PR introduces the ability to select specific episodes (or ranges of episodes) to rerun directly. This makes targeted debugging and analysis faster and more reproducible, and it also enables running subsets of episodes for specialized tasks such as visualization or plotting.

I’ve validated that this approach reproduces the exact same performance as when the selected episodes were run as part of the full experiment.

### CLI Examples

```zsh
# Run all episodes (default behavior)
python run_parallel.py -e my_experiment

# Run a single episode
python run_parallel.py -e my_experiment -i "3"

# Run multiple specific episodes
python run_parallel.py -e my_experiment -i "0,2,5"

# Run a closed range (end exclusive)
python run_parallel.py -e my_experiment -i "3:7"

# Run an open-ended range (from start or to end)
python run_parallel.py -e my_experiment -i ":5"    # episodes 0–4
python run_parallel.py -e my_experiment -i "7:"    # episodes 7–end

# Combine multiple selections
python run_parallel.py -e my_experiment -i "0,3,5:8"
```

### DIFF
--- a/src/tbp/monty/frameworks/models/graph_matching.py
+++ b/src/tbp/monty/frameworks/models/graph_matching.py
@@ -171,7 +171,7 @@ class MontyForGraphMatching(MontyBase):
 
     def reset(self):
         """Reset monty status."""
-        self.union_of_possible_matches = None
+        pass
 
     # ------------------ Getters & Setters ---------------------
 
@@ -265,35 +265,6 @@ class MontyForGraphMatching(MontyBase):
                 self.learning_modules[i].stepwise_targets_list.append(
                     self.learning_modules[i].stepwise_target_object
                 )
-
-    def _get_union_of_possible_matches(self):
-        """Take union of matches between LMs.
-
-        Update the union of possible matches returned by each learning module.
-        This is used to check the terminal condition that possible_matches is 0 or 1.
-
-        Returns:
-            Union of possible matches.
-        """
-        union_of_pm = None
-        for i, lm in enumerate(self.learning_modules):
-            if lm.buffer.get_num_observations_on_object() > 0:
-                pm = set(lm.get_possible_matches())
-            else:  # LM didn't get any observations yet -> don't make predictions.
-                # if we would use all memory IDs then time outs occur if the patch
-                # never gets on the object because it keeps the union of possible
-                # matches large.
-                # TODO: This LM may already have some IDs narrowed down by using
-                # incoming voted. Account for that.
-                pm = set()
-            logger.info(f"Possible matches for LM {i}: {pm}")
-            if union_of_pm is None:
-                union_of_pm = pm
-            else:
-                union_of_pm = set.union(union_of_pm, pm)
-        if len(self.learning_modules) > 1:
-            logger.info(f"Union of matches: {union_of_pm}")
-        return union_of_pm
 
     def _combine_votes(self, votes_per_lm):
         """Combine outgoing votes using lm_to_lm_vote_matrix matrix.
@@ -427,8 +398,14 @@ class MontyForGraphMatching(MontyBase):
                 self.send_vote_to_lm(self.learning_modules[i], i, combined_votes)
                 self.update_stats_after_vote(self.learning_modules[i])
 
-        # Update IoPM, needed for checking terminal condition
-        self.union_of_possible_matches = self._get_union_of_possible_matches()
+        # Log possible matches
+        for lm in self.learning_modules:
+            pm = (
+                lm.get_possible_matches()
+                if lm.buffer.get_num_observations_on_object()
+                else []
+            )
+            logger.info(f"Possible matches for {lm.learning_module_id}: {pm}")
 
     def _pass_infos_to_motor_system(self):
         """Pass input observations to the motor system.


ramyamounir:
FYI, @thousandbrainsproject/monty-maintainers

tristanls-tbp:
**question**: Is there a reason this option is omitted from `run.py`?

ramyamounir:
Unit tests added in [11ab20f8b891](https://github.com/ramyamounir/tbp.monty/commit/11ab20f8b8914357c0153233e9836acb5c82aae8)

> Is there a reason this option is omitted from `run.py`?

Yes, `run.py` is trickier to rerun specific episodes with reproducibility. The episodes run back to back without resetting the `rng`, so we cannot skip the earlier episodes because every `rng` call changes its internal state for the subsequent episodes. I think it's best to make sure the `rng` is always reset in `pre_episode` with the same seed, this way it would be easier to run specific episodes and expect the same results.

tests/unit/frameworks/run_parallel_test.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class ParseEpisodeSpecTest(unittest.TestCase):
+    def test_selects_all(self):
+        self.assertEqual(parse_episode_spec(None, total=5), [0, 1, 2, 3, 4])
+        self.assertEqual(parse_episode_spec("", total=3), [0, 1, 2])
+        self.assertEqual(parse_episode_spec(":", total=3), [0, 1, 2])
+        self.assertEqual(parse_episode_spec("all", total=4), [0, 1, 2, 3])
+
+    # Single index tests
```
tristanls-tbp: **issue**: extraneous comment, please remove

**note**: test names being `test_single_index_` obviate the need for the comment.

tests/unit/frameworks/run_parallel_test.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def test_single_index_negative_raises(self):
+        with self.assertRaisesRegex(ValueError, "not a valid index"):
+            parse_episode_spec("-1", total=5)
+
+    def test_single_index_out_of_bounds_raises(self):
+        with self.assertRaisesRegex(ValueError, "not a valid index"):
+            parse_episode_spec("5", total=5)
+
+    # Closed range tests
```
tristanls-tbp: **issue**: extraneous comment, please remove

**note**: test names being `test_closed_range_` obviate the need for the comment.

tests/unit/frameworks/run_parallel_test.py
```diff
@@ -0,0 +1,123 @@
...
+            parse_episode_spec("5:3", total=10)
+
+    def test_closed_range_end_equals_total_ok(self):
+        self.assertEqual(parse_episode_spec("0:4", total=4), [0, 1, 2, 3])
+
+    def test_closed_range_end_greater_than_total_raises(self):
+        with self.assertRaisesRegex(ValueError, "not a valid range"):
+            parse_episode_spec("0:6", total=5)
+
+    # Open range tests
```
tristanls-tbp: **issue**: extreaneous comment, please remove

**note**: test names being `test_open_` obviate the need for the comment.

tests/unit/frameworks/run_parallel_test.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def test_open_right_start_equals_total_invalid(self):
+        with self.assertRaisesRegex(ValueError, "not a valid range"):
+            parse_episode_spec("5:", total=5)
+
+    def test_open_right_start_negative_invalid(self):
+        with self.assertRaisesRegex(ValueError, "not a valid range"):
+            parse_episode_spec("-1:", total=5)
+
+    # Multiple selection and other edge cases
```
tristanls-tbp: **issue**: extraneous comment, please remove

src/tbp/monty/frameworks/run_parallel.py
```diff
@@ -469,6 +471,102 @@ def run_episodes_parallel(
...
+    if episode_spec is None:
+        return list(range(total))
+    s = episode_spec.strip().lower()
+    if s in ("", "all", ":"):
+        return list(range(total))
+
+    re_open_left = re.compile(r"^:([+-]?\d+)$")  # ":N"
+    re_open_right = re.compile(r"^([+-]?\d+):$")  # "N:"
+    re_closed = re.compile(r"^([+-]?\d+)\s*:\s*([+-]?\d+)$")  # "A:B"
+    re_single = re.compile(r"^[+-]?\d+$")  # "N"
```
tristanls-tbp: **question**: Why are `+` or `-` allowed? There is no use case for those.

**suggestion**: Remove

tristanls-tbp: **issue (non-blocking)**: I think these should be constants: `RE_OPEN_LEFT`, `RE_OPEN_RIGHT`, etc. and compiled outside of this function.

src/tbp/monty/frameworks/run_parallel.py
```diff
@@ -646,6 +745,7 @@ def main(
             from command line as the config is selected from `all_configs`.
         experiment: Name of experiment to run. Not required if running
             from command line.
+        episodes: The episodes ids to run. Default is "all".
```
tristanls-tbp: **issue**: [Our convention is "Defaults to ..."](https://github.com/search?q=repo%3Athousandbrainsproject%2Ftbp.monty%20%22Defaults%20to%22&type=code).

src/tbp/monty/frameworks/run_parallel.py
ramyamounir: I needed to catch the `+` and `-` and include them in the group so I can turn it in to `int` with the sign and apply the right checks/tests.

For example, if the user inputs `-i ":-5"` I want to include the `-` in the extracted group so I can apply the correct check and raise the correct error (incorrect range). If a user inputs `-i "+5:+8"`, I want the + to be included so that it can be later ignored when I cast it to `int`, otherwise it won't match any of the patterns defined.

tests/unit/frameworks/run_parallel_test.py
ramyamounir: Extraneous comments removed in [af17fd405101](https://github.com/ramyamounir/tbp.monty/commit/af17fd405101e8036c9f7c0e4821a53c88acb5e0)

src/tbp/monty/frameworks/run_parallel.py
ramyamounir: Regex statements became constants outside the function in [342e4ab68008](https://github.com/ramyamounir/tbp.monty/commit/342e4ab680089dc2348175c3717d422980aa9ef4)

src/tbp/monty/frameworks/run_parallel.py
ramyamounir: Fixed in [ad37a1df3ad3](https://github.com/ramyamounir/tbp.monty/commit/ad37a1df3ad31c3860f19a6c931fd6917d8054db)

vkakerbeck:
Cool! This is a really useful option to have :) Also an interesting idea to just reset the rng in every pre-episode call. I hadn't thought of this before but it sounds like it would save us a lot of reproducibility headaches that we tried to work around with the follow-up configs script.

ramyamounir:
@vkakerbeck , thanks! Yes, resetting `rng` at the beginning of each episode is now on my non-urgent Todo list. I'll get to it soon and then we could probably delete the follow-up configs script if we can reproduce specific episodes the same way with `run.py`.

src/tbp/monty/frameworks/run_parallel.py
tristanls-tbp: > For example, if the user inputs -i ":-5" I want to include the - in the extracted group so I can apply the correct check and raise the correct error (incorrect range). If a user inputs -i "+5:+8", I want the + to be included so that it can be later ignored when I cast it to int, otherwise it won't match any of the patterns defined.

**issue**: Allowing both signed and unsigned formats increases complexity and potential for bugs without apparent user demand. None of the examples or tests so far use negative or signed values, such as `+5:+8`. Furthermore, `+5:+8` is not idiomatic, and I see no reason for its inclusion. Since users generally know the total episode count and all episodes are numbered with positive integers, requiring positive integers simplifies both usage and maintenance. Please remove support for `+` and `-`.

ramyamounir: In [34502a535590](https://github.com/ramyamounir/tbp.monty/commit/34502a5355904499548069cfd2b704b0d7faddee)
- Support for `+` and `-` removed. 
- Unit tests `test_single_index_negative_raises` and `test_open_right_start_negative_invalid` removed.

tristanls-tbp (approved):
Thank you.

src/tbp/monty/frameworks/run_parallel.py
jeremyshoemaker: Putting the default in the docstring is redundant, similar to putting the type in the docstring. Does the generated documentation not include the defaults in the function signatures?

ramyamounir: Yeah, good point. I think, in another style PR, we can remove all `Defaults to ...` from the docstrings. The default value is easy to look up in the signature like you said.

That said, I think we're copying other documentation conventions where they specify the default value in both places (e.g., [numpy](https://numpy.org/devdocs/reference/generated/numpy.array.html#numpy-array))

tristanls-tbp: We are using the Google docstring convention. https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

The `Defaults to` comes from Sphinx (what we use to generate our docs) example of Google docstrings: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

PR #524, opened 2025-10-28
ci!: ruff update

tristanls-tbp:
This pull request updates `ruff` to `0.14.2` and addresses lints that resulted from the update:
- [PD901](https://docs.astral.sh/ruff/rules/pandas-df-variable-name/) was removed, so it is removed from ignore list
- [F401](https://docs.astral.sh/ruff/rules/unused-import/) unused import
- [FURB122](https://docs.astral.sh/ruff/rules/for-loop-writes/) for loop writes
- [PLC0415](https://docs.astral.sh/ruff/rules/import-outside-top-level/) import outside top level
- [PLW0177](https://docs.astral.sh/ruff/rules/nan-comparison/) nan comparison
- [PLW1641](https://docs.astral.sh/ruff/rules/eq-without-hash/) eq without hash
- [RUF028](https://docs.astral.sh/ruff/rules/invalid-formatter-suppression-comment/) invalid formatter suppression comment
- [RUF059](https://docs.astral.sh/ruff/rules/unused-unpacked-variable/) unused unpacked vars
- [RUF100](https://docs.astral.sh/ruff/rules/unused-noqa/) unused noqa
- [UP045](https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/) don't use Optional

The pull request is marked as a breaking change because it addresses PLC0415 lints, and plot utilities have been split into three separate files.

Interestingly, the goal was to remove `# noqa: B008` from flagging using `AgentID(...)` newtype in default arguments as a problem, but that does not seem to have worked out of the box, and instead, needed to register `tbp.monty.frameworks.agents.AgentID` as immutable for the `ruff` linter to recognize.

FYI @mf-maymay 

@thousandbrainsproject/monty-maintainers, you'll want to update your local `ruff` after this. Also, please note that most `plot_utils.py` moved to `plot_utils_dev.py` and `plot_utils_analysis.py` to separate optional dependency imports.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -304,6 +304,9 @@ ignore = [
     "UP006", # UP006: Use `type` instead of `Type` for type annotation
 ]
 
+[tool.ruff.lint.flake8-bugbear]
+extend-immutable-calls = ["tbp.monty.frameworks.agents.AgentID"]
+
 [tool.ruff.lint.flake8-comprehensions]
 # Our configurations use this all the time
 allow-dict-calls-with-keyword-arguments = true
--- a/src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
+++ b/src/tbp/monty/frameworks/config_utils/make_dataset_configs.py
@@ -899,7 +899,7 @@ def make_sensor_positions_on_grid(
 
 def make_multi_sensor_mount_config(
     n_sensors: int,
-    agent_id: AgentID = AgentID("agent_id_0"),  # noqa: B008
+    agent_id: AgentID = AgentID("agent_id_0"),
     sensor_ids: Sequence[str] | None = None,
     height: Number = 0.0,
     position: ArrayLike = (0, 1.5, 0.2),  # agent position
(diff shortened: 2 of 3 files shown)


tristanls-tbp:
@jeremyshoemaker I hear you.

There were three main places where I had to add a bunch of PLC0415.

In the `benchmarks/configs/load.py`, they need to be where they are, because that's where we do conditional module loading depending on what experiment we run. If I were to move those to the top of the file, we would be coupled to all the simulators at once.

The other place is `src/tbp/monty/framework/utils/plot_utils.py`. But looking at it now... I think I can clean that up a bit...

tristanls-tbp:
@jeremyshoemaker asking for another look. I moved all the unused plot utils into `dev_plot_utils.py` so that the two utilities left in `plot_utils.py`, when they are used, they will not attempt to import all of the optional dependencies.

This way, we no longer import optional dependencies inside the functions/methods, but at the top of the file.

jeremyshoemaker (approved):

Issue #506, opened 2025-10-16
Simulator protocol should be reduced to the bare minimum


AgentRev:
This is a continuation of [the topic](https://thousandbrains.discourse.group/t/2025-09-brighton-retreat-final-presentations/808/4?u=agentrev) discussed on the forums by @tristanls-tbp and I;

> My recommendation would be to first integrate the changes that reduce the Simulator protocol to the bare minimum. Those changs are all contained in [GitHub - thousandbrainsproject/prototype.simulator: Monty prototype using a simulator via a socket.](https://github.com/thousandbrainsproject/prototype.simulator) While the separation made it very clear what is the minimum required, reducing the protocol surface is a good thing even prior to/without the separation. There’s less to implement in any simulator, including MuJoCo.
>
> With the minimal protocol, I assume it should be easier to try out different serialization/communication approaches.

So, I tried manually merging `simulators/simulator.py`, `habitat/simulator.py`, and `mujoco/simulator.py`. These changes also require merging `embodied_environment.py`, `abstract_monty_classes.py`, `motor_system_state.py`, which in turn require merging other files, fixing tests, and so on.

You also mentioned _not_ to merge the whole [prototype.simulator](https://github.com/thousandbrainsproject/prototype.simulator) repo as-is, since it contains a lot of experimental code (which is totally fair since this was bascially a hackathon 😉)

I guess I'm just trying to figure out where to draw the line in the sand as to what to keep or not. Should I just keep merging things until most errors are gone and try to fix whatever remains? Is that what you had in mind?

tristanls-tbp:
I would suggest the strategy of doing a pull request for each `Simulator` protocol change. For example, first I would suggest removing `primary_target_bb` parameter from `Simulator.add_object`, commits df595b2bf6a57ddba482866ba43613b809bb995c and 719ae42fc19e2aca4b6bdb9227dd100bf524e7d5. This would be one PR.

Next, I would suggest extending the `add_object` return type with semantic ID, from commit 4fcab969bac696497e00d2eae918fef40bee5e37 up to and including 94fa4aee8c5aa89d1683b77afeada41d92488b6f. Note that `ObjectID` and `SemanticID` newtypes are orthogonal changes and should be their own separate PRs, if desired.

And so on...

The above approach enables us to quickly review obviously correct changes while keeping the destination in mind and referencing the prototype code for clarity.

I would not recommend one PR that changes everything at once.

FYI @jeremyshoemaker

[...]

PR #525, opened 2025-10-28
refactor!: integrate ObjectID, SemanticID, and ObjectInfo

AgentRev:
- Added ObjectID & SemanticID newtypes, and ObjectInfo dataclass;
- Adjusted associated function parameters, comments, and return values;
- Replaced `get_bounding_corners` with `_bounding_corners`.

### DIFF
--- a/src/tbp/monty/frameworks/environments/embodied_environment.py
+++ b/src/tbp/monty/frameworks/environments/embodied_environment.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 import abc
 import collections.abc
 from dataclasses import dataclass
-from typing import Any, Dict, NewType, Optional, Sequence, Tuple
+from typing import Any, Dict, NewType, Sequence, Tuple
 
 from tbp.monty.frameworks.actions.actions import Action
 
@@ -68,7 +68,7 @@ class EmbodiedEnvironment(abc.ABC):
         rotation: QuaternionWXYZ = (1.0, 0.0, 0.0, 0.0),
         scale: VectorXYZ = (1.0, 1.0, 1.0),
         semantic_id: SemanticID | None = None,
-        enable_physics: Optional[bool] = False,
+        enable_physics: bool | None = False,
         object_to_avoid=False,
         primary_target_object: ObjectID | None = None,
     ) -> ObjectID:


AgentRev:
This PR implements the following changes from issue #506 as part of the Simulator protocol refactor:

> * introduce ObjectID newtype (specifically `ObjectID(int)`)
> * replace `primary_target_bb=None` parameter with `primary_target_object: ObjectID | None = None` in `Simulator.add_object(...)`
> * introduce SemanticID newtype
> * change `Simulator.add_object(...)` return type to `tuple[ObjectID, SemanticID]`

I had to combine all of these changes into a single PR, since they had to be all brought in together for all unit tests to pass.

Instead of dealing with `tuple[ObjectID, SemanticID]`, I took the liberty of creating an `ObjectInfo` dataclass, which encapsulates `ObjectID` and `SemanticID`. This simplifies any potential future addition of new parameters.

I implemented the changes based on the `prototype.simulator` repo's `main` branch: https://github.com/thousandbrainsproject/prototype.simulator/tree/main

In order to get a single-page diff of all changes made to that branch during the focus week for easy reference, I fused them in a single commit here : https://github.com/AgentRev/tbp.monty/commit/4329d52a3e9cc73b1ede002bcb1839c150432a66

I have also attempted to merge `prototype.simulator` directly into the latest `tbp.monty` in an isolated branch, in order to better visualize what changes remain to be implemented as issue #506 progresses, which I will keep updated accordingly: https://github.com/AgentRev/tbp.monty/commits/protosim-merge-latest/

tristanls-tbp:
Hi @AgentRev, thank you for making the changes.

**note**: This pull request combines four independent changes into one pull request (ObjectID, SemanticID, get_bounding_corners, apply_actions). As such, it may take extra time to review in full due to increased cognitive load. Another concern is that if benchmarks don't match up, it won't be clear which of the four changes caused the benchmark drift.

AgentRev:
@tristanls-tbp It was a trade-off I had to make, since ObjectID & SemanticID are very closely intertwined, and to avoid having to create (and later revert) temporary code necessary to split them into multiple PRs (while also allowing the unit tests to still pass), which would have increased the total cognitive load for everyone involved regardless, in addition to extra PR overhead.

tristanls-tbp:
@AgentRev, maybe some more context will help. If you implement all the changes and end up with the code as in `prototype.simulator`, you will discover that Monty no longer works 🙂. So, somewhere along the chain of all those changes, I made a mistake and broke something. This is why I advocate for one-at-a-time changes as much as possible, because it will be easier to find the error I made.

AgentRev:
When you say "Monty no longer works" in that branch, do you have more details? Were the unit tests failing, or something more severe?

Also, I mistakenly said earlier that I changed `apply_actions` to `step`, that's not within this PR, I was getting mixed up with my other merge experiments. I corrected the PR description.

tristanls-tbp:
> When you say "Monty no longer works" in that branch, do you have more details? Were the unit tests failing, or something more severe?

Everything ran without errors, but Monty wasn't recognizing or learning things. The most likely culprit was something with serialization/deserialization of data and the data not being passed correctly. But, I don't know for sure, ran out of time to investigate.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -40,8 +45,6 @@
     ProprioceptiveState,
 )
 
-from .embodied_environment import EmbodiedEnvironment
```
tristanls-tbp, line 43: Thanks, good catch!

src/tbp/monty/frameworks/environments/embodied_environment.py
```diff
@@ -7,24 +7,43 @@
...
 
 
+@dataclass
+class EnvObject:
+    """Contains the identifying information of an object created in the environment."""
+
+    object_id: ObjectID
+    semantic_id: SemanticID | None
+
+
```
tristanls-tbp, line 46: **issue (non-blocking)**: There won't be more parameters, so I think this is preparing too much for things that won't happen. I prefer the tuple.

src/tbp/monty/simulators/simulator.py
```diff
@@ -58,9 +63,13 @@ def add_object(
...
-            primary_target_bb: If not None, this is a list of the min and
-              max corners of a bounding box for the primary object, used to prevent
-              obscuring the primary object with the new object.
+            primary_target_object: ID of the primary target object. If not None, the
+                added object will be positioned so that it does not obscure the initial
+                view of the primary target object (which avoiding collision alone cannot
+                guarantee). Used when adding multiple objects. Defaults to None.
+
+        Returns:
+            EnvObject instance of the added object.
```
tristanls-tbp: **issue (non-blocking)**: This reads too much like `-> EnvObject`. Also, the type information will be included in the docs by automation.

**suggestion**: The IDs associated with the added object. (or something similar)

**note**: The awkwardness of this sentence highlights why it's not really an "object" being returned (and why I prefer the tuple).

tests/unit/habitat_sim_test.py
```diff
@@ -183,35 +187,40 @@ def test_move_and_get_agent_state(self):
...
+                name="cylinderSolid", position=(-0.2, 1.5, -0.2)
+            ).semantic_id
+
+            cube = sim.add_object(
+                name="cubeSolid", position=(0.6, 1.5, -0.6)
+            ).semantic_id
 
             # Check if observations include both objects
-            expected = {cylinder.semantic_id, cube.semantic_id}
+            expected = {cylinder, cube}
```
tristanls-tbp, line 199: **question**: Why make the change at all? Since an object is being returned, doesn't the original code work here?

AgentRev: This change was originally made by Will, I simply copied it over: https://github.com/thousandbrainsproject/prototype.simulator/commit/f5d0aefd9c8d86133f979906fc20f3048ae7986d#diff-281bcf23c761803946b79d5e90b464520bcdcfd756318c32fdb0102cdb70a2b5

src/tbp/monty/simulators/simulator.py
AgentRev: I guess "The added object's metadata / information" could be enough, although I didn't want to mention "ID" outright in case EnvObject is ever extended to contain other metadata than just IDs.

src/tbp/monty/frameworks/environments/embodied_environment.py
AgentRev: I usually prefer dataclasses to tuples when the elements are closely related and likely to stay together, since the tuple is a generic type that can't communicate its origin or purpose. Although, this is more of a personal choice. It also mimics the format of the previous RigidObject return type, in a sense. If you want me to replace it with tuples, please say so.

src/tbp/monty/simulators/simulator.py
AgentRev: On second thought, I don't really like `EnvObject` as a name either, I think `ObjectInfo` would be better.

tristanls-tbp:
I ran the benchmarks, and everything looks good with no differences. However, one of them did not terminate. I'll see if restarting resolves it or if I need to debug further.

Screenshot 2025-10-29 at 15 00 39

tristanls-tbp:
Rerunning worked. I'm going to 🤷 about why it didn't terminate.

Screenshot 2025-10-29 at 15 10 41

I see there are two straightforward conflicts. Once you resolve those, I'm ready to approve and merge.

Please re-request a review so that I know it's ready.

Screenshot 2025-10-29 at 15 13 05

If you'd like to rename to `ObjectInfo`, then please do so before re-requesting a review, otherwise, I might miss that commit by merging quickly.

I still prefer the tuple over a dataclass in this case, but this PR doesn't need to wait for that. If it bothers me enough, I can always make a future PR myself 😬.

tests/unit/habitat_sim_test.py
tristanls-tbp: They're not quite the same. Destructuring vs property access. Looks like Will's change was necessary due to the tuple return type, but with object being returned, the property access is identical, hence no need for the change.

AgentRev: I first copied Will's change, then I later refactored the tuple away, but I kept the rest of Will's lines intact, because it's cleaner without those multiple `semantic_id` 😉

PR #526, opened 2025-10-29
chore: Target ruff rules RET501, RET504, and RET505

mf-maymay:
- [RET501](https://docs.astral.sh/ruff/rules/unnecessary-return-none/): unnecessary-return-none
- [RET504](https://docs.astral.sh/ruff/rules/unnecessary-assign/): unnecessary-assign
- [RET505](https://docs.astral.sh/ruff/rules/superfluous-else-return/): superfluous-else-return

RET501 asks that we only use `return`, not `return None`, when the function only ever returns `None`.

RET504 asks that we do not assign a variable only to return it, i.e. we say `return 5 * x` instead of `y = 5 * x; return y`.
I adjusted some fixes to reposition comments.

RET505 asks that we don't follow `if: [...] return` with an `else` (or `elif`), since the `return` makes the `else` unnecessary.
I added some line breaks while addressing this. (The code is cramped in some places, imo.)

I chose not to address [RET503](https://docs.astral.sh/ruff/rules/implicit-return/) in this PR, because there is at least one place where the fix would require (slightly) delicate care, and this diff could obfuscate that.

### DIFF
--- a/src/tbp/monty/frameworks/models/goal_state_generation.py
+++ b/src/tbp/monty/frameworks/models/goal_state_generation.py
@@ -179,7 +179,7 @@ class GraphGoalStateGenerator(GoalStateGenerator):
         a GoalState object with a None value for the location, morphological features,
         etc, or some variation of this.
         """
-        return None
+        return
 
     def _generate_goal_state(self, observations):
         """Generate a new goal-state to send out to other LMs and/or motor actuators.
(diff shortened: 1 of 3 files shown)


src/tbp/monty/frameworks/models/goal_state_generation.py
```diff
@@ -898,7 +862,29 @@ def _compute_goal_state_for_target_loc(
...
+                "on_object": 1,
+            },
+            non_morphological_features=None,
+            confidence=goal_confidence,
+            use_state=True,
+            sender_id=self.parent_lm.learning_module_id,
+            sender_type="GSG",
+            goal_tolerances=None,
+            info=info,
+        )
```
tristanls-tbp, line 887: **nitpick**: Reversing the code vs comment order changes the meaning slightly. "Consider also..." (also? but I haven't seen the not-also thing yet).

goingmaymay: Yes, I see what you mean. I could have tweaked the comment to make it less awkward.

PR #529, opened 2025-10-30
chore: update torch versions for uv environment

jeremyshoemaker:
The `uv` tests are breaking in various environments because the versions of `torch-scatter` and `torch-sparse` we pinned aren't available for the newer version of `torch` we're using.

This changes the version spec in `pyproject.toml` to allow newer versions and updates the lock file with a version that works.

I don't think we're actually testing the codepaths that use `torch` at the moment in the `uv` environment, so it might be broken anyway, but at least the tests we do have will run with this change.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -5031,6 +5031,12 @@ dev = [
     { name = "unittest-parametrize", version = "1.4.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.9'" },
     { name = "unittest-parametrize", version = "1.6.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.9'" },
 ]
+future-work-widget-tool = [
+    { name = "eval-type-backport" },
+    { name = "nh3" },
+    { name = "pydantic", version = "2.10.6", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.9'" },
+    { name = "pydantic", version = "2.11.5", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.9'" },
+]
 generate-api-docs-tool = [
     { name = "docutils", version = "0.20.1", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.9'" },
     { name = "docutils", version = "0.21.2", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.9'" },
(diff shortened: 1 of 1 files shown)


jeremyshoemaker:
Per our discussions, I've updated the `pyproject.toml` file to isolate the version specs for 3.8 (the Conda environment) and 3.9 (the UV environment).

Just to test it, I set one of the 3.8 dependencies to a non-existent version number to check that Conda uses only that specification and got this error:
```
Pip subprocess error:
ERROR: Could not find a version that satisfies the requirement torch-scatter==3.0.9; python_version == "3.8" (from tbp-monty) (from versions: 0.3.0, 1.0.2, 1.0.3, 1.0.4, 1.1.0, 1.1.1, 1.1.2, 1.2.0, 1.3.0, 1.3.1, 1.3.2, 1.4.0, 2.0.2, 2.0.3, 2.0.4, 2.0.5, 2.0.6, 2.0.7, 2.0.8, 2.0.9, 2.1.0, 2.1.1, 2.1.2)
ERROR: No matching distribution found for torch-scatter==3.0.9; python_version == "3.8"
```

tristanls-tbp:
**note**: I updated the PR/commit to `chore:` since [`fix:` is reserved for `src/tbp/monty` changes](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0010_conventional_commits.md#types).

jeremyshoemaker:
> **note**: I updated the PR/commit to `chore:` since [`fix:` is reserved for `src/tbp/monty` changes](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0010_conventional_commits.md#types).

Thanks. I thought of it as a fix because something was broken, but had forgotten we reserved it for Monty specific fixes.

tristanls-tbp:
And to double-check. Installing the environment using `conda` from scratch works with this configuration, yes?

jeremyshoemaker:
> And to double-check. Installing the environment using `conda` from scratch works with this configuration, yes?

Let me triple check, because right now I'm having issues getting the uv environment to work in Github Actions, and it feels like locally I keep getting different inconsistent results for unknown reasons.

jeremyshoemaker:
> Installing the environment using `conda` from scratch works with this configuration, yes?

Yes, I checked again that it works and installs the expected versions of the libraries. I was pretty sure I already checked this before you asked, but I've been doing enough `conda env remove` and `conda env create` and similar to be sure of what I have or haven't done.

PR #531, opened 2025-10-31
refactor!: SemanticID in GetGoodView positioning procedure

tristanls-tbp:
`GetGoodView` positioning procedure now uses `SemanticID` newtype instead of `int`.

I also changed the `0` sentinel to `"any"` in `get_perc_on_obj_semantic(...)` to make it more explicit as to what's happening.

## Benchmarks

No changes

Screenshot 2025-10-31 at 18 45 01

### DIFF
--- a/src/tbp/monty/frameworks/models/motor_policies.py
+++ b/src/tbp/monty/frameworks/models/motor_policies.py
@@ -17,16 +17,7 @@ import logging
 import math
 import os
 from dataclasses import dataclass, field
-from typing import (
-    Any,
-    Dict,
-    List,
-    Literal,
-    Mapping,
-    Tuple,
-    Type,
-    cast,
-)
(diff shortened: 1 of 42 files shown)


src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -1612,7 +1613,9 @@ def write_action_file(actions: list[Action], file: str) -> None:
         )
 
 
-def get_perc_on_obj_semantic(semantic_obs, semantic_id=0):
+def get_perc_on_obj_semantic(
+    semantic_obs, semantic_id: SemanticID | Literal["any"] = "any"
```
jeremyshoemaker, line 1617: Definitely like this, making the alternative explicit and not overloading the 0. 👍

PR #532, opened 2025-11-02
refactor!: remove enable_physics and object_to_avoid params

AgentRev:
As part of issue #506 for the Simulator protocol refactor, this PR removes the `enable_physics` and `object_to_avoid` params from all applicable functions.

### DIFF
--- a/src/tbp/monty/simulators/habitat/simulator.py
+++ b/src/tbp/monty/simulators/habitat/simulator.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 
 from collections import defaultdict
 from pathlib import Path
-from typing import Any, Dict, List, Sequence
+from typing import Any, Sequence
 
 import habitat_sim
 import magnum as mn
@@ -107,7 +107,7 @@ class HabitatSim(HabitatActuator):
 
     def __init__(
         self,
-        agents: List[HabitatAgent],
+        agents: list[HabitatAgent],
         data_path: str | None = None,
         scene_id: str | None = None,
         seed: int = 42,
@@ -510,7 +510,7 @@ class HabitatSim(HabitatActuator):
         agent_index = self._agent_id_to_index[agent_id]
         return self._sim.get_agent(agent_index)
 
-    def apply_actions(self, actions: Sequence[Action]) -> Dict[str, Dict]:
+    def apply_actions(self, actions: Sequence[Action]) -> dict[str, dict]:
         """Execute given actions in the environment.
 
         Args:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -239,7 +239,6 @@ ignore = [
     "EM101", # EM101: Exception must not use a string literal, assign to variable first
     "EM102", # EM102: Exception must not use an f-string literal, assign to variable first
     "F403", # F403: from {name} import * used; unable to detect undefined names
-    "FA100", # FA100: Add `from __future__ import annotations` to simplify
     "FBT001", # FBT001: Boolean-typed positional argument in function definition
     "FBT002", # FBT002: Boolean default positional argument in function definition
     "FBT003", # FBT003: Boolean positional value in function call
@@ -298,7 +297,6 @@ ignore = [
     "TD004", # TD004: Missing colon in TODO
     "TD005", # TD005: Missing issue description after `TODO`
     "TRY003", # TRY003: Avoid specifying long messages outside the exception class
-    "UP006", # UP006: Use `type` instead of `Type` for type annotation
 ]
 
 [tool.ruff.lint.flake8-bugbear]
--- a/src/tbp/monty/frameworks/config_utils/config_args.py
+++ b/src/tbp/monty/frameworks/config_utils/config_args.py
@@ -13,14 +13,7 @@ import copy
 import os
 from dataclasses import dataclass, field
 from itertools import product
-from typing import (
-    TYPE_CHECKING,
-    Callable,
-    Dict,
-    Iterable,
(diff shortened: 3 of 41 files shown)


tristanls-tbp:
@jeremyshoemaker one of the benchmarks changed as the result of this PR, please confirm the change and identify the cause
Screenshot 2025-11-03 at 16 04 13

fwiw.. it's a nice _improvement_

I restarted the changed benchmark as a sanity check.

tristanls-tbp:
I confirmed that the benchmark change occurred no later than this commit, but I don't know whether an earlier commit caused it.

Screenshot 2025-11-03 at 16 37 39

src/tbp/monty/simulators/habitat/simulator.py
```diff
@@ -298,7 +287,7 @@ def add_object(
 
         # Update object-specific physics
         # This physics setting will persist into the environment e.g. during inference
-        if enable_physics:
+        if primary_target_object is not None:
             obj.motion_type = habitat_sim.physics.MotionType.DYNAMIC
```
AgentRev, line 291: I suspect the benchmark changes are probably a result of this edit

jeremyshoemaker: Wait, isn't this conditional just duplicating part of what the conditional that starts on 277 does, specifically line 280?

That said, if we're removing `enable_physics` and making it always on, shouldn't this conditional just go away because it would be the same as `if True`?

I haven't had a chance to sit down and run the benchmarks on the commit right before this to see if it also has the changed results, so it might end up not being this commit that changed it at all.

jeremyshoemaker: It's definitely something in this commit that changed the benchmark results, and I suspect you're right about it being this code since there's not really any other good candidate.

I think what's happened is that this code branch used to not be taken, pretty much ever, because `enable_physics` at the `add_object` level was defaulting to `False`. The code was confusing because it had the enable physics stuff at the simulator level too.

I'm running an experiment with this conditional changed to just be the `else` half of it, since that's what would have been happening before to see if it matches the previous results.

jeremyshoemaker: So, that's what changed. Before the code was always doing the `else` branch, and now it's not whenever we provide a `primary_target_object` (which currently only happens in the `base_10multi_distinctobj_dist_agent` experiment).

If we want to change this back to the old behaviour, we'd need to just remove this `if` section and use the contents of the `else` branch. When I do that and run the benchmark on it, the results return the what they were before.

tristanls-tbp: For consistency, let's always go with the `else` branch. Tweaking a benchmark result because `DYNAMIC` vs `KINEMATIC` doesn't seem much to do with Monty, so I'd rather not game the benchmark.

@vkakerbeck, let us know if you don't want to always go with the `else` branch.

jeremyshoemaker: I opened #538 to fix the logic.

vkakerbeck: Yes I took a closer look at the code again and it seems like `enable_physics` is never set to True. I agree that for now it makes most sense to preserve the current behavior and just keep going with the else branch. As far as I remember we used to set `enable_physics` to True when we set up the multi object environment because this was required to check for object collisions (at least that's what it says here https://github.com/thousandbrainsproject/tbp.monty/blame/1f10388cd68e9b0c670dda9e5e5eb23fac85ad42/src/tbp/monty/simulators/habitat/simulator.py#L275) But it may still be correct since, as far as I can see, also after this refactor we temporarily set this to true here https://github.com/thousandbrainsproject/tbp.monty/blob/66a09e91a8151f163c4e09a7bbcb618961c537d9/src/tbp/monty/simulators/habitat/simulator.py#L279) 

The idea is we temporarily turn physics on to check for collisions when we add multiple objects. Then we turn physics off again before the experiment starts so the objects don't fall down into the void (since there is no table or similar for them to rest on at the moment). The benchmark change was likely because physics was not turned off again before the experiment and so objects were falling/disappearing. 

So TL;DR the fix @jeremyshoemaker is proposing looks like the right way to preserve the previous behavior. However, removing the if branch and the parameter will make it harder in the future to run experiments with physics enabled. But it also doesn't seem like it would be too difficult to add that feature back in at that point.

jeremyshoemaker: This code is kinda in the weeds of the details of HabitatSim, so I'm not sure it would make sense to keep it explicitly here (vs just looking it up in Git history) if we're aren't planning on keeping Habitat. From what I've seen, MuJoCo handles physics differently enough that we might not have this problem, or we'll have different problems.

tristanls-tbp: > The benchmark change was likely because physics was not turned off again before the experiment and so objects were falling/disappearing.

@vkakerbeck , the benchmark change was an _improvement_. So I don't think any objects fell or disappeared.

vkakerbeck: Yes I am not sure at all what happened. But for example, if for some reason physics were just enabled for the distractor objects then the task would have become much easier and performance would have improved (not saying this is what happened).

tristanls-tbp: @vkakerbeck , oh I see now, thank you; it is a per-object specification, so distractors could just disappear. Good to know 👍 .

PR #533, opened 2025-11-02
refactor!: deprecate ActionSpace

AgentRev:
As part of issue #506 for the Simulator protocol refactor, this PR deprecates `ActionSpace`. It's a cherry pick of 2 commits from the [prototype.simulator](https://github.com/thousandbrainsproject/prototype.simulator) repo. The only edits I made are conflict resolution and Ruff format.


src/tbp/monty/simulators/habitat/agents.py
```diff
@@ -120,71 +121,80 @@ def process_observations(self, agent_obs) -> dict:
...
-        )
-        surface_agent_action_space = (
-            ["move_forward", self.translation_step, None],
-            ["move_tangentially", self.translation_step, None],
-            ["orient_horizontal", self.rotation_step, None],
-            ["orient_vertical", self.rotation_step, None],
-            ["set_agent_pose", [[0.0, 0.0, 0.0], qt.one], None],
-            ["set_sensor_rotation", [[qt.one]], None],
+def action_space(
+    action_space_type: Literal["absolute_only", "distant_agent", "surface_agent"],
```
jeremyshoemaker: We use this Literal type in several places below, so it might make since to make a type alias for it. That way we aren't repeating the definition everywhere. If it's just limited to this file, then the definition can just go at the top somewhere, but otherwise, I'm not sure if we have a good place for it yet.

Something like (though I don't like the "Type" in this name, needs some thought):
```python
ActionSpaceType = Literal["absolute_only", "distant_agent", "surface_agent"]
```

Alternatively, we could make it an Enum instead. I'm not sure if we need the values to be strings or not, but I haven't looked at this code in a while, so I'm not sure.

src/tbp/monty/frameworks/environments/embodied_environment.py
```diff
@@ -36,6 +38,7 @@
 QuaternionWXYZ = Tuple[float, float, float, float]
 
 
+@deprecated("Use `ActionSampler` instead.")
```
tristanls-tbp: **issue**: Pretty sure this is the wrong place for this annotation.

AgentRev: Oops you're right, I clicked the wrong button in my merge tool

src/tbp/monty/simulators/habitat/agents.py
AgentRev: The comment for `action_space_type` args is "name of action space", so how about `ActionSpaceName`?

Enums can be strings (e.g. `class MyEnum(str, Enum)`), but it wouldn't be a drop-in replacement here, because there are 46 action_space_type literals across the codebase that would have to be replaced with enum values, so a generic `Literal` is more convenient for now in my opinion.

jeremyshoemaker: I think an Enum would probably be a bit much here, or rather it would be its own set of changes since it changes the values as well as the types. But a type alias seems reasonable.

TIL: string Enums. I worked on a Python 2.7 project for so long I'm still discovering Python 3 things that I didn't know about.

jeremyshoemaker:
I just want to link these two together so the context isn't lost. In [the comment @tristanls-tbp is referring to](https://github.com/thousandbrainsproject/tbp.monty/pull/525), he said:
> If you implement all the changes and end up with the code as in `prototype.simulator`, you will discover that Monty no longer works 🙂. So, somewhere along the chain of all those changes, I made a mistake and broke something.

jeremyshoemaker:
Benchmark results are unchanged from the previous values:
image

AgentRev:
I'll change the Literal to an alias later today

jeremyshoemaker:
> I'll change the Literal to an alias later today

Sounds good. I'll hold off on merging then.

jeremyshoemaker:
Looks like there are some conflicts that need resolving. Also, FYI @tristanls-tbp and I are doing a focus mini-week for the rest of this week so we'll be slower to respond to PRs.

PR #536, opened 2025-11-04
ci: actions to build and validate the future work widget

codeallthethingz:
Add GitHub Actions workflow to build and deploy the Future Work widget to S3

# future_work_widget_validate.yml

Runs on pull requests targets, checks out the PR branch, builds the widget from PR docs, captures validation results, and annotates PR files with GitHub error annotations if validation fails, failing the workflow if the build is invalid.

image

# future_work_widget_deploy.yml

Runs on merges to main - builds the widget from the main branch docs, syncs the built widget files to S3 (two syncs to prevent downtime during deployment).


.github/workflows/future_work_widget_validate.yml
```diff
@@ -0,0 +1,111 @@
...
+      - name: Checkout tbp.monty
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          lfs: true
+          path: tbp.monty
+      - name: Checkout PR branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
tristanls-tbp, line 60: **note**: I'm surprised this worked. I would have expected `github.even.pull_request_target.head.sha`, but it's been a while since I looked at the details, and if it works, it works.

.github/workflows/future_work_widget_validate.yml
```diff
@@ -0,0 +1,111 @@
...
+        if: always()
+        working-directory: tbp.monty
+        run: |
+          if [ -f "${{ runner.temp }}/build_result.json" ]; then
+            success=$(jq -r '.success // false' "${{ runner.temp }}/build_result.json")
+            if [ "$success" != "true" ]; then
+              echo "Build failed"
+              exit 1
+            fi
+          fi
```
tristanls-tbp, line 116: **issue**: This should also fail if the `build_result.json` file doesn't exist. I see `Build future work widget` ends with `|| true`, which might succeed despite a failure.

.github/workflows/future_work_widget_validate.yml
codeallthethingz: Yeah, seems to be the payload for the event uses pull_request regardless of whether its a PRT or PR.

tristanls-tbp (approved):
Thank you for the updates, looks good.

PR #538, opened 2025-11-05
fix: set physics motion type correctly

jeremyshoemaker:
Previously, we had an argument to the `add_object` function, called `enable_physics` that defaulted to False, which we never set to True at any of the call sites.

When that argument was removed, the logic was mistakenly changed so that where we were previously setting the physics MotionType to KINEMATIC, now we were sometimes setting it to DYNAMIC.

This fixes the logic to do what the code previously did and consistently set the MotionType to KINEMATIC.

### DIFF
--- a/src/tbp/monty/simulators/habitat/simulator.py
+++ b/src/tbp/monty/simulators/habitat/simulator.py
@@ -285,16 +285,11 @@ class HabitatSim(HabitatActuator):
                 primary_obj_bb=primary_target_bb,
             )
 
-        # Update object-specific physics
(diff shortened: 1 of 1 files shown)


jeremyshoemaker:
I'm working on getting benchmark numbers to confirm that they go back to the previous values.

tristanls-tbp (approved):
Thank you for the discovery and the follow-up.

Approving, conditional on the benchmarks turning out as expected.

jeremyshoemaker:
> conditional on the benchmarks turning out as expected.

Results match the previous results from PR #531 

image

image

PR #539, opened 2025-11-05
chore!: Target ruff rules FLY002, PLE0302, PLR1714, RUF012, RUF013, and SIM103

mf-maymay:
- [FLY002](https://docs.astral.sh/ruff/rules/static-join-to-f-string/): Join strings with f-strings, not `str.join`
- [PLE0302](https://docs.astral.sh/ruff/rules/unexpected-special-method-signature/): Use the correct signatures when defining special (dunder) methods
- [PLR1714](https://docs.astral.sh/ruff/rules/repeated-equality-comparison/): Use set membership checks in place of repeated equality checks
- [RUF012](https://docs.astral.sh/ruff/rules/mutable-class-default/): Explicitly annotate class vars with `typing.ClassVar`
- [RUF013](https://docs.astral.sh/ruff/rules/implicit-optional/): Do not rely on the "implicit optional" for function arg types
    - Example: `x: int = None` should be `x: int | None = None`
- [SIM103](https://docs.astral.sh/ruff/rules/needless-bool/): Return a boolean expression directly instead of using an equivalent if-block

I opted to address PLE0302 by removing the `BaseBuffer` class entirely. It seems to be meant to define the core interface for buffers, but it's only used as the base class for `FeatureAtLocationBuffer`. I suggest that we wait to implement a common buffer interface (as a protocol) until we know we will use it.

In a couple of places, I opted to rewrite a dict definition using the curly brace syntax. `{"x": 1, "y": "two"}` is more conventional than `dict(x=1, y="two")`, and it's better at announcing its dict-ness, too. (I wish there was a ruff rule to require the curly brace syntax...)

### DIFF
--- a/tools/github_readme_sync/req.py
+++ b/tools/github_readme_sync/req.py
@@ -60,4 +60,4 @@ def delete(url: str, headers=None):
     headers["Authorization"] = f"Basic {os.getenv('README_API_KEY')}"
     response = requests.delete(url, headers=headers, timeout=REQUEST_TIMEOUT_SECONDS)
     logging.debug("delete %s %s", url, response.status_code)
-    return not (response.status_code < 200 or response.status_code >= 300)
+    return 200 <= response.status_code < 300


jeremyshoemaker:
Thanks for continuing to work on this, I love that we're getting these enabled. FYI @tristanls-tbp and I are doing a focus mini-week for the rest of this week so we'll be slower to respond to PRs.

jeremyshoemaker:
> I wish there was a ruff rule to require the curly brace syntax...

There's a Ruff setting we've set to explicitly allow using those because the experiment configurations rely heavily on the `dict(foo=bar)` syntax to reduce the amount of quotes.

See #274 and in particular [this section of the configuration](https://github.com/thousandbrainsproject/tbp.monty/pull/274/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R305-R308).

src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -288,10 +288,10 @@ def is_motor_only_step(self, state: MotorSystemState) -> bool:
         """
         agent_state = self.get_agent_state(state)
 
-        if "motor_only_step" in agent_state.keys() and agent_state["motor_only_step"]:
-            return True
-
-        return False
+        # FIXME: "motor_only_step" is not a valid AgentState key (based on type).
+        return bool(
+            "motor_only_step" in agent_state.keys() and agent_state["motor_only_step"]
```
jeremyshoemaker: If `agent_state` is a dictionary (I don't know if it is here), then `keys()` isn't needed to check if the key is in the dictionary.
```suggestion
            "motor_only_step" in agent_state and agent_state["motor_only_step"]
```

goingmaymay:
@jeremyshoemaker 
Oh wow, so there used to be even more `dict()` calls? Thanks for the link, and thanks for addressing the ones you did :)

src/tbp/monty/frameworks/models/motor_policies.py
goingmaymay: True, good catch. Addressed.
The result resulted in another ruff complaint for RUF019, so I addressed that, too.

I will soon target the ruff rule that forbids unnecessary `.keys()` calls.

tools/github_readme_sync/colors.py
```diff
@@ -19,17 +19,13 @@
...
-    if sys.platform == "win32":
-        return True  # Colorama handles color support on Windows
+    win32 = sys.platform == "win32"
 
-    if hasattr(sys.stdout, "isatty") and sys.stdout.isatty():
-        return True  # Non-Windows platforms with a TTY
+    isatty = hasattr(sys.stdout, "isatty") and sys.stdout.isatty()
 
-    return False
+    return not running_in_ci and (win32 or isatty)
```
tristanls-tbp, line 28: **note**: I see this is more than needed to address SIM103, but ok.

tools/github_readme_sync/req.py
```diff
@@ -60,6 +60,4 @@ def delete(url: str, headers=None):
     headers["Authorization"] = f"Basic {os.getenv('README_API_KEY')}"
     response = requests.delete(url, headers=headers, timeout=REQUEST_TIMEOUT_SECONDS)
     logging.debug("delete %s %s", url, response.status_code)
-    if response.status_code < 200 or response.status_code >= 300:
-        return False
-    return True
+    return not (response.status_code < 200 or response.status_code >= 300)
```
tristanls-tbp: **issue**: let's avoid boolean negation when possible
```suggestion
    return 200 <= response.status_code < 300
```

goingmaymay: Good call.

tristanls-tbp (approved):

PR #540, opened 2025-11-05
refactor!: use Hydra for configuring Monty

jeremyshoemaker:
> [!NOTE]
> Description written by @tristanls-tbp on behalf of @jeremyshoemaker and myself.

> _"We do this not because it is easy... but because we thought it was going to be easy..."_

This pull request converts our configuration approach from a previous custom-built code solution to using [Hydra](https://github.com/facebookresearch/hydra). Please note that a Hydra tutorial is beyond the scope of this pull request description.

## The goal

The goal of this pull request is not to improve configurations. The goal is to move it to Hydra so that configurations can be more easily improved in the future. The eventual goal is for all configuration to be specified in yaml, and the way to instantiate a new experiment is to do:
```python
@hydra.main(config_path="../../../conf", config_name="experiment", version_base=None)
def main(cfg: DictConfig):
  ...
  with experiment:
    exp.train()
    exp.evaluate()  
```
Eventually, instead of our custom-built init functions, all the work of creating things will be delegated to `hydra.utils.instantiate(...)`.

## Overview

The diff may be challenging to review, but here is the overview of changes.

All benchmarks are removed from the `benchmarks` folder and are now in form of yaml configuration files in `conf`. The only things remaining in the `benchmarks` folder are the CSV files containing our benchmark numbers.

The translation from benchmark configuration class inheritance hierarchy into Hydra yaml configuration was quite literal, in that the inheritance hierarchy is largely maintained. This means that the hydra configuration can be simplified and refactored in the future, but for now, the goal was to make it as familiar as possible while retaining somewhat of the original shape for easier debugging and review.

The `conf` folder is as follows:
```
|- conf
    |- benchmarks
    |- experiment
    |- test
    |- _init_.py
    |- experiment.yaml
    |- test.yaml
    |- validate.py
```

Within `conf/benchmarks`, you'll find some common values used in the experiment configurations. This will need to be cleaned up in the future. It comes from `benchmarks/defaults.py` file.

Within `conf/experiment` are all experiments. There are tutorial configurations within `conf/experiment/tutorial`. There are (for reasons) test configurations for testing `run.py` and `run_parallel.py` in `conf/experiment/test`.

The rest of the configuration hierarchy is in `conf/experiment/config` and matches the previous configuration hierarchy we had.

To run an experiment, we now do:
```
./run.py experiment=experiment_name
```
or
```
./run_parallel.py experiment=experiment_name
```

The main configuration entrypoint file `conf/experiment.yaml` contains additional command-line parameters that can be provided. For example, 
```yaml
# conf/experiment.yaml
defaults:
  - benchmarks: defaults
  # Experiment name
  - experiment: MISSING
  - _self_

# Which episodes to run (zero-based). Examples: "all" | "3" | "0,3,5:8" | "4:" | ":8".
episodes: all
# How many episodes to run in parallel
num_parallel: 16
# Don't run an experiment; just print out the config for visual inspection
# TODO: Replace with validate or similar
print_cfg: false
# Set logging levels in habitat to quiet
quiet_habitat_logs: true
```
means that you can run
```
./run.py experiment=experiment_name num_parallel=1 quiet_habitat_logs=false
```

You will find a temporary utility `conf/validate.py`, which let's you run
```
python conf/validate.py experiment=experiment_name
```
to quickly validate that the config can instantiate an experiment. This will be improved upon in future pull requests.

The configuration entrypoint file `conf/test.yaml` is for use by tests and points to `conf/test/*` for the various test configurations.

### logging_config renamed to logging

During our initial enthusiasm, we changed `logging_config` to `logging`. However, for all other renames, we decided to wait until a later PR to simplify reviewing the changes and translating the configs.

In the future, we'll remove all the `_config`, `_args`, and similar suffixes as they are redundant in YAML configurations now that they're no longer in code.

### We no longer allow multiple agents to be specified.

We can revisit this in the future when we work on having multiple agents. It was done for expediency to meet the project timeline.

### Follow up configs are gone

Follow-up configs can be fully subsumed by "run only an episode" functionality already present in the current configs. The only obstacle to that is that we do not reset random seed prior to each episode when we run serially. The fix for that should be resetting the random seed prior to each episode. As a result and a motivation, the follow-up config code is deleted instead of rewritten.

### Default output_dir for training

The new defaults will output training into a folder called `my_trained_models` as to not conflict with any shared pretrained models in our infrastructure. This allows for defaults to be safe and require deliberate action to move a trained model into a shared (readonly) space.

## Deletions form `src`

Lots of configuration and utility classes and functions that became dead code are removed. There are a few remaining stragglers that are still used in tests, but they can be wrangled in follow-on PRs.

## Benchmarks

We replicated all the inference benchmarks. See `pr540` tag in WandB.

Additionally, for training, for YCB we replicated created bit-by-bit identical `model.pt` files. For compositional objects, we trained using the new yaml configurations and then replicated exact inference benchmark numbers based on those trained models. While the `model.pt` files were not identical bit-by-bit in this case, their model contents were and did not affect inference results.

## TODO (but could be a follow on PR):
- [ ] update tutorials
- [ ] update documentation

## Future work

We'd like to provide a snapshot validation utility for configurations. Due to their inherited nature, it is easy to unintentionally change configs one did not intend to change. 

This pull request only moved the configurations to Hydra. We still need to add Pydantic validation to all objects.

Futhermore, there are numerous configuration improvements once we can more safely change Hydra configurations, in particular, switching to dependency injection instead of the `*_class` + `*_args` pattern of custom initialization.

### DIFF
--- a/conf/experiment/README.md
+++ b/conf/experiment/README.md
@@ -41,4 +41,4 @@ To generate models for the objects with logos experiments, run the following pre
 - `./run_parallel.py experiment=supervised_pre_training_objects_with_logos_lvl4_comp_models`
 
 
-Fro more details, see [Running Benchmarks](https://thousandbrainsproject.readme.io/docs/running-benchmarks) and [Benchmark Experiments](https://thousandbrainsproject.readme.io/docs/benchmark-experiments) in the documentation.
+For more details, see [Running Benchmarks](https://thousandbrainsproject.readme.io/docs/running-benchmarks) and [Benchmark Experiments](https://thousandbrainsproject.readme.io/docs/benchmark-experiments) in the documentation.
--- a/conf/experiment/base_config_10distinctobj_dist_agent.yaml
+++ b/conf/experiment/base_config_10distinctobj_dist_agent.yaml
@@ -16,7 +16,7 @@ config:
   n_eval_epochs: ${benchmarks.rotations_all_count}
   logging:
     wandb_group: benchmark_experiments
-    # Comment in for quick debugging (turns of wandb and increases logging)
+    # Comment in for quick debugging (turns off wandb and increases logging)
     # wandb_handlers: []
     # python_log_level: DEBUG
     # FIXME?: we want to actually concatentate with logging/defaults.yaml output_dir
--- a/conf/experiment/config/defaults.yaml
+++ b/conf/experiment/config/defaults.yaml
@@ -7,7 +7,7 @@ max_eval_steps: 500
 # episode steps that can be taken before timing out, regardless of e.g. whether LMs
 # receive sensory information and therefore perform a true matching step (due to
 # e.g. being off the object)
-# heuristic is 4 * max_train_steps + max_eval_steps
+# heuristic is 4 * (max_train_steps + max_eval_steps)
 max_total_steps: 6000
 n_train_epochs: 3
 n_eval_epochs: 3
--- a/conf/experiment/config/environment/init_args/agents/patch_and_view_finder.yaml
+++ b/conf/experiment/config/environment/init_args/agents/patch_and_view_finder.yaml
@@ -1,6 +1,6 @@
 # Config using view finder to find the object before starting the experiment.
 #
-# A common default for Viviane's experiments that use the view finder to navigate
+# A common default for experiments that use the view finder to navigate
 # so the object is in view before the real experiment happens.
 agent_id: ${monty.agent_id:agent_id_0}
 sensor_ids:
--- a/conf/experiment/config/environment/init_args/agents/surface_and_view_finder.yaml
+++ b/conf/experiment/config/environment/init_args/agents/surface_and_view_finder.yaml
@@ -1,7 +1,7 @@
 # Use surface agent and view finder to find the object before experiment start.
 #
-# Adaptation of Viviane's code that use the view finder to navigate so
-# the object is in view before the real experiment happens + surface-agent sensor
+# Adaptation of distant agent code for the surface agent to use the view finder to 
+# navigate so the object is in view before the real experiment happens.
 defaults:
   - patch_and_view_finder
 
--- a/conf/experiment/config/infer_comp.yaml
+++ b/conf/experiment/config/infer_comp.yaml
@@ -20,7 +20,7 @@ n_eval_epochs: ${benchmarks.rotations_all_count}
 min_lms_match: 1
 logging:
   wandb_group: benchmark_experiments
-  # Comment in for quick debugging (turns of wandb and increases logging)
+  # Comment in for quick debugging (turns off wandb and increases logging)
   # wandb_handlers: []
   # python_log_level: INFO
 monty_config:
--- a/conf/experiment/randrot_noise_10distinctobj_dist_agent.yaml
+++ b/conf/experiment/randrot_noise_10distinctobj_dist_agent.yaml
@@ -14,7 +14,7 @@ defaults:
   - config/environment_interface/clear_object_init_sampler@config.eval_env_interface_args
 
 config:
-  mode_name_or_path: ${benchmarks.pretrained_dir}/surf_agent_1lm_10distinctobj/pretrained/
+  model_name_or_path: ${benchmarks.pretrained_dir}/surf_agent_1lm_10distinctobj/pretrained/
   n_eval_epochs: 10 # number of random rotations to test for each object
   logging:
     run_name: randrot_noise_10distinctobj_dist_agent
--- a/conf/test/config/monty/learning_modules/default_evidence_lm.yaml
+++ b/conf/test/config/monty/learning_modules/default_evidence_lm.yaml
@@ -8,7 +8,7 @@ learning_module_args:
   # recognizing the cube as a sphere during learning.
   # The current implementation isn't optimized for few,
   # incomplete models in memory. In this case the slightly
-  # curved surface of the spere is too similar to the flat
+  # curved surface of the sphere is too similar to the flat
   # surface of the cube to lead to negative evidence and the
   # sphere is not excluded from possible matches.
   max_match_distance: 0.001
--- a/conf/validate.py
+++ b/conf/validate.py
@@ -19,7 +19,7 @@ from tbp.monty.hydra import register_resolvers
 
 @hydra.main(config_path=".", config_name="experiment", version_base=None)
 def validate(cfg: DictConfig):
-    # Force interpolation of the config to get errors
+    # Force resolving of the config to get errors
     OmegaConf.to_object(cfg)
     print(OmegaConf.to_yaml(cfg))
 


hlee9212:
Not sure if this has been fixed since [a25fb03](https://github.com/thousandbrainsproject/tbp.monty/pull/540/commits/a25fb03436cdc0bcda30b1b63d0869337b412d2b) but I tried running all benchmark experiments in AWS but all the wandb experiment names are `base_config_10distinctobj_dist_agent`. 

CleanShot 2025-11-13 at 16 19 22

tristanls-tbp:
@hlee9212 the run name thing should be resolved now.

conf/experiment/randrot_noise_10distinctobj_dist_agent.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - /benchmarks/default_all_noisy_sensor_module@config.monty_config.sensor_module_configs.sensor_module_0
+  - config/monty/args/clear_monty_args@config.monty_config
+  - config/monty/args/defaults@config.monty_config.monty_args
+  - config/environment/clear_eval_env_interface_args@config
+  - config/environment_interface/per_object@config.eval_env_interface_args
+  - config/environment_interface/ycb/distinct_objects@config.eval_env_interface_args.object_names
+  - config/environment_interface/clear_object_init_sampler@config.eval_env_interface_args
+
+config:
+  mode_name_or_path: ${benchmarks.pretrained_dir}/surf_agent_1lm_10distinctobj/pretrained/
```
hlee9212: **typo**: I think it should be `model_name_or_path`

conf/experiment/config/defaults.yaml
```diff
@@ -0,0 +1,17 @@
+do_train: true
+do_eval: true
+show_sensor_output: false
+max_train_steps: 1000
+max_eval_steps: 500
+# Total number of
+# episode steps that can be taken before timing out, regardless of e.g. whether LMs
+# receive sensory information and therefore perform a true matching step (due to
+# e.g. being off the object)
+# heuristic is 4 * max_train_steps + max_eval_steps
```
hlee9212: **nitpick**: Put parentheses around, i.e. `4 * (max_train_steps + max_eval_steps)` = 4 * (1000 + 500) = 4 * 1500 = 6000.

conf/experiment/config/environment/two_lm_stacked_distant_mount_habitat.yaml
```diff
@@ -2,3 +2,20 @@ defaults:
   - multi_lm_mount_habitat
   - init_args/clear_env_init_args@
   - init_args/two_lm_distant_stacked_mount@env_init_args
+
+env_interface_config:
```
tristanls-tbp: **issue**: This won't work because you'll end up with `env_interface_config.env_interface_config`.

tristanls-tbp: **note**: Updated in https://github.com/thousandbrainsproject/tbp.monty/pull/540/commits/229c0611c6b1f0e9ba6b514dcffd08a4a9eb01fd

tristanls-tbp: **note**: We have a `validate` utility that might be helpful: `python conf/validate.py experiment=...` which tries to create the experiment class.

tristanls-tbp: **note**: Although, the `validate` utility wouldn't have helped in this case. I tried, it didn't error.

run.py
tristanls-tbp, line 1: **note**: GitHub change detection is being weird; this file wasn't moved, it's a small new file.

run_parallel.py
tristanls-tbp, line 1: **note**: GitHub change detection is being weird; this file wasn't moved, it's a small new file.

benchmarks/README.md
vkakerbeck, line 1: Is the content of this file preserved anywhere?

conf/benchmarks/default_all_noisy_sensor_module.yaml
vkakerbeck, line 1: if the `conf/benchmarks/` folder only contains defaults, not actual benchmark experiments, why not call it `conf/defaults/`?

conf/experiment/README.md
```diff
@@ -0,0 +1,44 @@
...
+- `./run_parallel.py experiment=supervised_pre_training_objects_with_logos_lvl1_comp_models_resampling`
+
+### Phase 4
+
+- `./run_parallel.py experiment=supervised_pre_training_objects_with_logos_lvl2_comp_models`
+- `./run_parallel.py experiment=supervised_pre_training_objects_with_logos_lvl3_comp_models`
+- `./run_parallel.py experiment=supervised_pre_training_objects_with_logos_lvl4_comp_models`
+
+
+Fro more details, see [Running Benchmarks](https://thousandbrainsproject.readme.io/docs/running-benchmarks) and [Benchmark Experiments](https://thousandbrainsproject.readme.io/docs/benchmark-experiments) in the documentation.
```
vkakerbeck:
```suggestion
For more details, see [Running Benchmarks](https://thousandbrainsproject.readme.io/docs/running-benchmarks) and [Benchmark Experiments](https://thousandbrainsproject.readme.io/docs/benchmark-experiments) in the documentation.
```

conf/experiment/base_config_10distinctobj_dist_agent.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - config/environment_interface/per_object@config.eval_env_interface_args
+  - config/environment_interface/ycb/distinct_objects@config.eval_env_interface_args.object_names
+
+_target_: tbp.monty.frameworks.experiments.object_recognition_experiments.MontyObjectRecognitionExperiment
+config:
+  model_name_or_path: ${benchmarks.pretrained_dir}/surf_agent_1lm_10distinctobj/pretrained/
+  n_eval_epochs: ${benchmarks.rotations_all_count}
+  logging:
+    wandb_group: benchmark_experiments
+    # Comment in for quick debugging (turns of wandb and increases logging)
```
vkakerbeck:
```suggestion
    # Comment in for quick debugging (turns off wandb and increases logging)
```

conf/experiment/config/environment/init_args/agents/patch_and_view_finder.yaml
```diff
@@ -0,0 +1,15 @@
+# Config using view finder to find the object before starting the experiment.
+#
+# A common default for Viviane's experiments that use the view finder to navigate
```
vkakerbeck: I think this is a pretty old comment and is now a default for pretty much all our experiments
```suggestion
# A common default for experiments that use the view finder to navigate
```

conf/experiment/config/environment/init_args/agents/surface_and_view_finder.yaml
```diff
@@ -0,0 +1,17 @@
+# Use surface agent and view finder to find the object before experiment start.
+#
+# Adaptation of Viviane's code that use the view finder to navigate so
+# the object is in view before the real experiment happens + surface-agent sensor
```
vkakerbeck:
```suggestion
# Adaptation of distant agent code for the surface agent to use the view finder to 
# navigate so the object is in view before the real experiment happens.
```

conf/experiment/config/infer_comp.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - environment_interface/per_object@eval_env_interface_args
+  - environment_interface/logos_on_objs/objects_with_logos_lvl1@eval_env_interface_args.object_names
+  - environment_interface/clear_parent_to_child_mapping@eval_env_interface_args
+  - environment_interface/logos_on_objs/parent_to_child_mapping@eval_env_interface_args.parent_to_child_mapping
+
+n_eval_epochs: ${benchmarks.rotations_all_count}
+min_lms_match: 1
+logging:
+  wandb_group: benchmark_experiments
+  # Comment in for quick debugging (turns of wandb and increases logging)
```
vkakerbeck:
```suggestion
  # Comment in for quick debugging (turns off wandb and increases logging)
```

conf/experiment/config/infer_comp.yaml
vkakerbeck, line 1: is it possible to call this `infer_compositional.yaml`? No worries if that's a bigger change, it might just make it a bit more intuitive to understand what this config is for

conf/experiment/config/logging/defaults.yaml
```diff
@@ -0,0 +1,16 @@
+monty_log_level: DETAILED
+monty_handlers:
+  - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.BasicCSVStatsHandler}
+  - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.DetailedJSONHandler}
+  - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.ReproduceEpisodeHandler}
```
vkakerbeck, line 5: If we are not keeping the make_follow_up_config code, we also don't need the ReproduceEpisodeHandler anymore.

conf/benchmarks/defaults.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - [35, 45, 0]
+  - [325, 45, 0]
+  - [35, 315, 0]
+  - [325, 315, 0]
+  - [35, 135, 0]
+  - [325, 135, 0]
+  - [35, 225, 0]
+  - [325, 225, 0]
+
+rotations_3_count: 3 # Surprise!
```
vkakerbeck, line 47: :D is it not possible to use syntax similar to `${np.ones:3}` below? like `n_eval_epochs=${len(rotations_3)}`?

conf/experiment/randrot_noise_10distinctobj_dist_agent.yaml
vkakerbeck: Yes, good catch! I am surprised this worked with the typo...

conf/experiment/test/run.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - /experiment/config/environment/single_ptz_habitat@config.env_interface_config
+
+_target_: tbp.monty.frameworks.experiments.monty_experiment.MontyExperiment
+config:
+  max_train_steps: 10
+  max_eval_steps: 5
+  n_train_epochs: 2
+  n_eval_epochs: 1
+  logging:
+    monty_log_level: TEST
```
vkakerbeck, line 14: Is this a new option?

conf/test/config/monty/learning_modules/default_evidence_lm.yaml
```diff
@@ -0,0 +1,14 @@
...
+  - feature_weights/defaults@learning_module_args.feature_weights
+  - tolerances/defaults@learning_module_args.tolerances
+
+learning_module_class: ${monty.class:tbp.monty.frameworks.models.evidence_matching.learning_module.EvidenceGraphLM}
+learning_module_args:
+  # Need to set max_match_distance low here to avoid
+  # recognizing the cube as a sphere during learning.
+  # The current implementation isn't optimized for few,
+  # incomplete models in memory. In this case the slightly
+  # curved surface of the spere is too similar to the flat
```
vkakerbeck:
```suggestion
  # curved surface of the sphere is too similar to the flat
```

conf/validate.py
```diff
@@ -0,0 +1,40 @@
...
+import hydra
+from omegaconf import DictConfig, OmegaConf
+
+from tbp.monty.frameworks.run_env import setup_env
+from tbp.monty.hydra import register_resolvers
+
+
+@hydra.main(config_path=".", config_name="experiment", version_base=None)
+def validate(cfg: DictConfig):
+    # Force interpolation of the config to get errors
```
vkakerbeck: What does this mean?

src/tbp/monty/frameworks/run.py
```diff
@@ -56,60 +53,4 @@ def run(config):
...
-        # Update run_name and output dir with experiment name
-        # NOTE: wandb args are further processed in monty_experiment
-        if not exp_config["logging_config"]["run_name"]:
-            exp_config["logging_config"]["run_name"] = experiment
-        exp_config["logging_config"]["output_dir"] = os.path.join(
-            exp_config["logging_config"]["output_dir"],
-            exp_config["logging_config"]["run_name"],
-        )
-        # If we are not running in parallel, this should always be False
-        exp_config["logging_config"]["log_parallel_wandb"] = False
```
vkakerbeck, line 104: Where does this now happen?

src/tbp/monty/frameworks/utils/follow_up_configs.py
```diff
@@ -28,26 +32,104 @@ def recover_output_dir(config, config_name):
...
-    return config["logging_config"]["wandb_id"]
+    return config["logging"]["wandb_id"]
+
+
+def create_eval_episode_hydra_cfg(
+    parent_config: DictConfig, episode: int
+) -> DictConfig:
+    """Returns a new Hydra config that runs only that episode with detailed logging.
+
+    This doesn't support the `update_run_dir` option that the function below does.
```
vkakerbeck, line 62: What does this mean? Is this an issue?

src/tbp/monty/frameworks/utils/follow_up_configs.py
vkakerbeck, line 1: Does this actually work? I thought we were deprecating this?

conf/validate.py
jeremyshoemaker: Some values in the config are calculated using what OmegaConf calls _resolvers_, and they don't get resolved until you do something like this to force resolution. I think this also catches things like a `_target_` value that doesn't point to an actual class or function, like if the module path is wrong.

run.py
jeremyshoemaker: It's actually at the Git level. Rename detection is difficult because Git doesn't track changes directly, just the before and after state, and so it's heuristics based.

src/tbp/monty/frameworks/utils/follow_up_configs.py
jeremyshoemaker: The tests that needed a new version of this function didn't need that functionality, so I left it out since I think we're planning to remove the rest of it.

benchmarks/README.md
jeremyshoemaker: It's in the history of the repository so if we need it we can always recover it.

conf/benchmarks/default_all_noisy_sensor_module.yaml
tristanls-tbp: I think we'll refactor this further when we redesign the configurations.

conf/benchmarks/defaults.yaml
tristanls-tbp: No, we're not able to derive properties like this. Things like this complicate the configuration. This shouldn't be a configuration option; if it is derived, it (`len(rotations)`) should happen in the code inside the class initialized with rotations.

conf/experiment/config/logging/defaults.yaml
tristanls-tbp: Sounds good. We'll do it in a follow-up PR.

conf/experiment/config/defaults.yaml
tristanls-tbp:
```suggestion
# heuristic is 4 * (max_train_steps + max_eval_steps)
```

conf/experiment/config/infer_comp.yaml
tristanls-tbp: 😄 I didn't know what that meant until your comment 😬. Yes, happy to change it in a follow-on PR (if that's alright).

conf/experiment/test/run.yaml
tristanls-tbp: Just copying test for now https://github.com/thousandbrainsproject/tbp.monty/blob/11881b3f0fead9df998b4e6b138e42b9c3f397d4/tests/unit/run_test.py#L121

conf/experiment/randrot_noise_10distinctobj_dist_agent.yaml
tristanls-tbp:
```suggestion
  model_name_or_path: ${benchmarks.pretrained_dir}/surf_agent_1lm_10distinctobj/pretrained/
```

conf/validate.py
tristanls-tbp:
```suggestion
    # Force resolving of the config to get errors
```

src/tbp/monty/frameworks/utils/follow_up_configs.py
tristanls-tbp: We got rid of follow_up_configs. This is the leftover code that is still used by `graph_learning_test.py`, specifically in the `test_reproduce_single_episode` test. We will eventually clean this up as we clean up the tests.

src/tbp/monty/frameworks/run.py
tristanls-tbp: In all but one config, we made the default `false`:
```yaml
# conf/experiment/config/logging/defaults.yaml
...
log_parallel_wandb: false
```
In `run_parallel.py`, we set it to `True` if there are WandB handlers present:
```python
            if len(new_experiment["config"]["logging"]["wandb_handlers"]) > 0:
                new_experiment["config"]["logging"]["wandb_handlers"] = []
                new_experiment["config"]["logging"]["log_parallel_wandb"] = True
                new_experiment["config"]["logging"]["experiment_name"] = name
            else:
                new_experiment["config"]["logging"]["log_parallel_wandb"] = False
```
The only config that sets `log_parallel_wandb: true` is `parallel_evidence_lm.yaml`.

Overall, this doesn't seem like it should be specified in a configuration file, but only set dynamically. We'll address that in follow-on PRs.

tristanls-tbp:
@vkakerbeck , I updated the tutorials, we fixed the test, and I believe we addressed your comments.

Would you please take another look.

Screenshot 2025-11-21 at 10 03 49

benchmarks/README.md
tristanls-tbp: We did add a `conf/experiments/README.md` which includes pretraining details. So, we could add something similar to `conf/README.md`. I deleted this because most of it didn't apply `# Benchmarks` section nor `## Follow-up Configs` section, and the `## Types of Benchmarks` section was out of date.

vkakerbeck: OKay that makes sense. When updating documentation it may be worth adding a conf/README file that links to our benchmark docs so people know where to look for more information on these experiments.

conf/benchmarks/default_all_noisy_sensor_module.yaml
vkakerbeck: ok it just seems a bit confusing right now since this is the folder where I would look in first to find the benchmark experiments. But if you have that on your list of next changes already I am happy with having it in a separate PR.

benchmarks/README.md
tristanls-tbp: Sounds good, adding to the list.

conf/benchmarks/defaults.yaml
vkakerbeck: okay makes sense. Generally number of epoch wouldn't always be equal to len of rotations so we don't code it in a different place. It is just derive that way in the experiments here but we still want it to be configurable more flexibly.

conf/experiment/test/run.yaml
vkakerbeck: ah okay, wasn't aware we did that

vkakerbeck (approved):
Nice, thanks for the updates and answers! Happy to defer the outstanding changes to future PRs

PR #541, opened 2025-11-06
refactor!: follow ruff rule ARG002

carver:
~This isn't ready for review, just a heads-up that I'm working on it.~ Okay, I think this is ready for a first pass review.

When encountering unused parameters, there were a couple of possible options. By taking a quick look at the context, I made a best guess for the preferred solution. Some different solutions used throughout:

1. Delete the argument, when it seems that it doesn't break the API in an unrecoverable way
2. If the API depends on the argument staying, but it's not used, then prepend the argument with an underscore, to indicate that it's known that the variable is unused
3. Raise an exception when an unexpected value is passed (thereby reading the unused argument, and warning the caller if they are trying to use an unimplemented feature)
4. Ignore whole files that prefer unused arguments extensively (eg~ test mocks)

Let me know if you want me to switch up any of the solutions in the PR!

## Process Note
I have a habit of opening PRs very early, marking them as Draft and/or [Work in Progress]. I personally find it useful on both sides of the project (maintainer and contributor). For example: it's nice to run against CI, especially when I haven't built out my local runtime scripts to check linting. Or maybe someone will notice I'm headed in the wrong direction, like maybe there's a good reason to not start with ARG002.

Please let me know if you prefer that I wait until it's more complete, next time.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -307,6 +307,13 @@ allow-dict-calls-with-keyword-arguments = true
 [tool.ruff.lint.flake8-copyright]
 author = "Thousand Brains Project"
 
+[tool.ruff.lint.isort]
+# wandb creates a folder called 'wandb' during local runs (not logged in)
+# this needs to be added to prevent isort from incorrectly sorting
+# It can be removed when a fix for this is released:
+# https://github.com/astral-sh/ruff/issues/10519
+known-third-party = ["wandb"]
+
 [tool.ruff.lint.mccabe]
 max-complexity = 18
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# wandb local configs
+wandb/
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/src/tbp/monty/frameworks/models/evidence_matching/features_for_matching/selector.py
+++ b/src/tbp/monty/frameworks/models/evidence_matching/features_for_matching/selector.py
@@ -36,8 +36,11 @@ class DefaultFeaturesForMatchingSelector:
             ):
                 use_features[input_channel] = False
             else:
-                feature_weights_provided = (
-                    len(feature_weights[input_channel].keys()) > 2
-                )
+                non_morphological_features = {
+                    k
+                    for k in feature_weights[input_channel].keys()
+                    if k not in ["pose_vectors", "pose_fully_defined"]
+                }
+                feature_weights_provided = len(non_morphological_features) > 0
                 use_features[input_channel] = feature_weights_provided
         return use_features
--- a/src/tbp/monty/frameworks/models/salience/sensor_module.py
+++ b/src/tbp/monty/frameworks/models/salience/sensor_module.py
@@ -54,6 +54,10 @@ class HabitatSalienceSM(SensorModule):
         # TODO: Goes away once experiment code is extracted
         self.is_exploring = False
 
+    @property
+    def sensor_module_id(self) -> str:
+        return self._sensor_module_id
+
     def state_dict(self):
         return self._snapshot_telemetry.state_dict()
 


carver:
```sh
$ ruff check
All checks passed!
```

Next step is to set up my local environment and run the tests locally. After that, I think it should be ready for a proper review.

carver:
```sh
$ pytest --ff -x -f
...
================= 330 passed, 17 skipped, 1 warning in 10.82s ==================
```

Next up is to figure out what test-tools is doing, and then run the type checker. I tried a naive `mypy` run, but that was showing a bunch of errors on my checkout from `main`.

carver:
Ah, I suppose both mypy and `pytest tools` require conda, and I've been running with `uv` (I know, against advice :P ). Well, if you're up for it, I'll just let CI tell me what else I'm missing.

tests/unit/frameworks/actions/action_samplers_test.py
```diff
@@ -155,18 +155,6 @@ def setUp(self) -> None:
             translation_distance=self.translation_distance,
         )
 
-    def test_accepts_arbitrary_additional_init_params(self) -> None:
-        ConstantSampler(
```
carver, line 159: I doubt that deleting this test (and the one below it) is the preferred solution. I don't have enough information to make a better suggestion. What do you think should be done here?

It seems strange that this test implies that it's so important to keep the `**kwargs`, and yet the keyword args are completely ignored by `ConstantSampler` and `UniformlyDistributedSampler`.

jeremyshoemaker:
> I have a habit of opening PRs very early, marking them as Draft and/or [Work in Progress]. I personally find it useful on both sides of the project (maintainer and contributor). For example: it's nice to run against CI, especially when I haven't built out my local runtime scripts to check linting. Or maybe someone will notice I'm headed in the wrong direction, like maybe there's a good reason to not start with ARG002.
>
> Please let me know if you prefer that I wait until it's more complete, next time.

Our usual approach is to open a Draft PR if we're still working on things, and only move it to a regular one when it's ready for review. That way it doesn't show up on our triage views before it's ready for one of us to look at.

src/tbp/monty/frameworks/models/graph_matching.py
```diff
@@ -570,7 +570,7 @@ def __init__(self, initialize_base_modules=True):
 
         if initialize_base_modules:
             self.graph_memory = GraphMemory(k=None, graph_delta_thresholds=None)
-            self.gsg = GraphGoalStateGenerator(self, gsg_args=None)
```
carver, line 574: I'm flying pretty blind here, but it seems like it's okay to drop this. It is passing a None value kwarg that is ignored by this specific class.

carver:
Ah darn, I was hoping once you approved a particular PR for running in CI, that at least followup changes in that same PR would automatically get run, but it looks like they require your permission every time. That's pretty inconvenient.

src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
```diff
@@ -522,13 +524,15 @@ def __init__(self, handlers):
...
-        kwargs = dict(
-            train_episodes_to_total=self.train_episodes_to_total,
-            eval_episodes_to_total=self.eval_episodes_to_total,
-        )
+
+        # TODO: start logging
+        # (
+        #    train_episodes_to_total=self.train_episodes_to_total,
+        #    eval_episodes_to_total=self.eval_episodes_to_total,
+        # )
```
tristanls-tbp: **issue**: Please do not add commented-out code.

src/tbp/monty/frameworks/models/salience/strategies.py
```diff
@@ -18,5 +18,5 @@ def __call__(self, rgba: np.ndarray, depth: np.ndarray) -> np.ndarray: ...
 
 
 class UniformSalienceStrategy(SalienceStrategy):
-    def __call__(self, rgba: np.ndarray, depth: np.ndarray) -> np.ndarray:
+    def __call__(self, _rgba: np.ndarray, depth: np.ndarray) -> np.ndarray:
```
tristanls-tbp: **note**: Oh nice, I'm glad that protocol recognition works with different argument names.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -154,7 +157,17 @@ def __init__(
...
+        Raises:
+            InvalidEvidenceThresholdConfig: If `evidence_threshold_config` is not
+                set to "all".
+
         """
+        if evidence_threshold_config != "all":
+            raise InvalidEvidenceThresholdConfig(
+                "evidence_threshold_config must be 'all' for `DefaultHypothesesUpdater`"
+            )
+
```
tristanls-tbp: **issue**: This is not true. If `evidence_threshold_config` is unused and ignored, then it can be anything.

**issue**: When I ran benchmarks, most of them failed with:
```
tbp.monty.frameworks.utils.evidence_matching.InvalidEvidenceThresholdConfig: evidence_threshold_config must be 'all' for `DefaultHypothesesUpdater`
```

src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -393,7 +393,7 @@ class Probe(SensorModule):
 
     def __init__(
         self,
-        rng,
+        rng,  # noqa: ARG002
```
tristanls-tbp, line 396: **question**: Any specific reason `_rng` doesn't work here?

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -81,6 +81,9 @@ def add_object(
 
         # TODO: reinitialize agents?
 
+        if primary_target_object is not None:
+            raise NotImplementedError("MuJoCo's primary_target_object is not supported")
+
```
tristanls-tbp: **issue**: Please do not raise `NotImplementedError`s.

**suggestion**: `_primary_target_object`.

tests/unit/frameworks/actions/action_samplers_test.py
tristanls-tbp: It is worth highlighting here that tests in the `tests` folder are insufficient to test Monty, and that benchmarks are also part of our end-to-end test suite. Additionally, due to the code's historical path, it may very well be the case that `**kwargs` are unnecessary. However, until we run all the benchmarks, we won't know for sure. There may have been a historical reason that no longer exists. I wrote the samplers, so I copied over whatever the pattern was from before. I will be glad to be rid of `**kwargs` here if they are not needed 👍.

tests/unit/frameworks/models/fakes/sensor_modules.py
```diff
@@ -18,7 +18,11 @@
 class FakeSensorModule(SensorModule):
     """Dummy placeholder class used only for tests."""
 
-    def __init__(self, rng, sensor_module_id: str):
+    def __init__(
+        self,
+        rng,  # noqa: ARG002
```
tristanls-tbp, line 23: **question**: Why not `_rng`?

src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
```diff
@@ -200,9 +200,11 @@ def post_episode(self, logger_args, output_dir, model):
         self.update_episode_data(logger_args, output_dir, model)
         self.log_episode(logger_args, output_dir, model)
 
-    def update_episode_data(self, logger_args, output_dir, model):
+    def update_episode_data(self, logger_args, _output_dir, model):
```
tristanls-tbp: **issue**: This parameter should be deleted from `update_episode_data` signature. It is unused here, and subclasses only accept this parameter to call `super().update_episode_data` with it, which ultimately remains unused.

src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
```diff
@@ -522,13 +524,15 @@ def __init__(self, handlers):
...
+
+        # TODO: start logging
+        # (
+        #    train_episodes_to_total=self.train_episodes_to_total,
+        #    eval_episodes_to_total=self.eval_episodes_to_total,
+        # )
 
         for handler in self.handlers:
-            handler.report_episode(self.data, output_dir, episode, mode, **kwargs)
+            handler.report_episode(self.data, output_dir, episode, mode)
```
tristanls-tbp: **issue**: Removing `kwargs` here is incorrect. See https://github.com/thousandbrainsproject/tbp.monty/blob/8cdf7a48ea57c57fd0db1eb6d56134fc3d374b13/src/tbp/monty/frameworks/loggers/monty_handlers.py#L111

src/tbp/monty/frameworks/models/evidence_matching/graph_memory.py
```diff
@@ -161,7 +161,6 @@ def _extend_graph(
         object_location_rel_body,
         location_rel_model,
         object_rotation,
-        object_scale,
```
tristanls-tbp, line 164: **issue**: This is insufficient. Looking at the code, it appears that `object_scale` should be removed from `_extend_graph()` signature entirely.

src/tbp/monty/frameworks/models/goal_state_generation.py
```diff
@@ -730,7 +727,11 @@ def _get_target_loc_info(self, target_loc_id):
         }
 
     def _compute_goal_confidence(
```
tristanls-tbp: **issue**: Looks like this function is just returns `lm_output_confidence` and therefore can be removed.

pyproject.toml
```diff
@@ -320,6 +318,12 @@ author = "Thousand Brains Project"
 [tool.ruff.lint.mccabe]
 max-complexity = 18
 
+[tool.ruff.lint.per-file-ignores]
+# Ignore `ARG002` (unused argument violations) in all test files that use
+# mocking: mocking adds arguments that aren't directly accessed.
+"tools/github_readme_sync/tests/*_test.py" = ["ARG002"]
```
tristanls-tbp: **issue**: No, please don't do this. For example, I took a look at cli_test.py and none of those tests need `@patch` annotation, hence all unused arguments can be removed. Similarly, the `colors_test.py`, you can get rid of unused parameters by using pytest fixtures and usefixtures annotation, for example:
```python
    @pytest.fixture
    def isatty_false(self):
        with mock.patch("sys.stdout.isatty", return_value=False):
            yield

    @mock.patch.dict(os.environ, {"CI": "false"})
    @mock.patch("sys.platform", "linux")
    @pytest.mark.usefixtures("isatty_false")
    def test_colors_disabled_on_linux_without_tty(self):
        # Reload colors module after changing the environment
        importlib.reload(colors)
        self.assertFalse(colors._supports_color())
        self.assertEqual(colors.RED, "")
        self.assertEqual(colors.GRAY, "")
        self.assertEqual(colors.GREEN, "")
        self.assertEqual(colors.WHITE, "")
        self.assertEqual(colors.CYAN, "")
        self.assertEqual(colors.BLUE, "")
        self.assertEqual(colors.RESET, "")
```

pyproject.toml
```diff
@@ -320,6 +318,12 @@ author = "Thousand Brains Project"
...
+"tests/unit/frameworks/models/salience/sensor_module_test.py" = ["ARG002"]
```
tristanls-tbp: **issue**: This can also likely be fixed with `@pytest.fixture` & `@pytest.mark.usefixtures` when needed.

src/tbp/monty/frameworks/utils/plot_utils_dev.py
```diff
@@ -1189,13 +1189,15 @@ def plot_up_to_step(self, total_steps=None):
...
+    def plot_animation(self, zoom=None, view=None):
         """Plot an animation of the episode's full action policy."""
         self.plot_core_object()
         self.derive_policy_details()
 
         self.ax.set_aspect("equal")
-        # self.ax.set_box_aspect((6, 6, 6), zoom=zoom)
+        if zoom is not None:
+            # zoom defaults to 1.0
+            self.ax.set_box_aspect((6, 6, 6), zoom=zoom)
```
tristanls-tbp: **issue**: This is utility code, so I'm not sure of the original intent here, but I don't think we should override the default 4:4:3 box aspect with 6:6:6 since it was commented out before.

tristanls-tbp:
**note**: I changed the PR name to `refactor!`. I marked it as a refactor since removing (even if unusued) parameters is refactoring some method and constructor signatures. Additionally, I marked it breaking `!` because the removal of parameters from methods and constructors breaks the API (even if the parameters are unused).

src/tbp/monty/frameworks/models/sensor_modules.py
carver: Some callers pass `rng` in as a keyword argument that gets mapped to this named argument. That mapping fails if it is renamed to `_rng` (as found in some test failures).

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
carver: Ok, so I guess there will be a natural tension between the codebase and this ARG002 rule. As I understand it, the design is to intentionally pass in variables that are ignored by some classes. A couple alternatives here:
- `noqa: arg002` on the argument, of classes that chose to ignore it
- a bigger refactor: switch from kwargs to a configuration value as a positional argument. A class that only uses part of the config will still pass arg002. A class that doesn't use it at all can prepend `_`. (because we don't rely on passing it in as a keyword arg)

src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
carver: ~In #548 now~ nevermind, I just merged this change into this PR

src/tbp/monty/frameworks/models/evidence_matching/graph_memory.py
carver: ~now in #549~ nevermind, folded in

src/tbp/monty/frameworks/models/goal_state_generation.py
carver: In #550

src/tbp/monty/simulators/mujoco/simulator.py
carver: I *think* this is passed in as a kwarg sometimes. I can't tell if Simulators are ever used as Environments. Environments are called this way:
https://github.com/thousandbrainsproject/tbp.monty/blob/5176bba40767426c9e85452cba683c934624b602/src/tbp/monty/frameworks/environments/embodied_data.py#L395-L398

I also see a lot of dictionary manipulation in the code around primary_target_object, which makes me wonder if it's being passed in as a kwarg somewhere. But it's tough to tell with just some grepping around.

~Anyhow, I can try it and we can see if the testing reveals any issues. (It does technically deviate from the Simulator protocol with this change, but maybe that's okay?)~

Edit to add (ETA): the `typing.override()` solution I mentioned elsewhere seems to fit this scenario nicely. I'll try that instead.

tests/unit/frameworks/models/fakes/sensor_modules.py
carver: Same as the other sensor module: callers pass it in as `sensor_module(rng=...)`, which then causes a `TypeError`.

carver:
I also have to revert the `_` prepend in the arguments to all the logger functions, because of errors like this:

```log
____________________________ MontyRunTest.test_run _____________________________
[gw0] linux -- Python 3.8.18 /home/runner/miniconda/envs/tbp.monty/bin/python

config = {'env_interface_config': SinglePTZHabitatEnvInterfaceConfig(env_init_func=<class 'tbp.monty.simulators.habitat.environ...s=6000, n_train_epochs=2, n_eval_epochs=1, model_name_or_path='', min_lms_match=1, seed=42, supervised_lm_ids=[]), ...}

    def run(config):
        with config["experiment_class"](config) as exp:
            # TODO: Later will want to evaluate every x episodes or epochs
            # this could probably be solved with just setting the logging freqency
            # Since each trainng loop already does everything that eval does.
            if exp.do_train:
                print("---------training---------")
>               exp.train()

src/tbp/monty/frameworks/run.py:54: 

self = <tbp.monty.frameworks.experiments.monty_experiment.MontyExperiment object at 0x7674d6f12220>

    def train(self):
        """Run n_train_epochs."""
>       self.logger_handler.pre_train(self.logger_args)

src/tbp/monty/frameworks/experiments/monty_experiment.py:570: 

self = <tbp.monty.frameworks.loggers.exp_logger.LoggingCallbackHandler object at 0x7674d6e812b0>
logger_args = {'eval_episodes': 0, 'eval_epochs': 0, 'total_eval_steps': 0, 'total_train_steps': 0, ...}

    def pre_train(self, logger_args):
>       self.call_event("pre_train", logger_args)

src/tbp/monty/frameworks/loggers/exp_logger.py:61: 

self = <tbp.monty.frameworks.loggers.exp_logger.LoggingCallbackHandler object at 0x7674d6e812b0>
event = 'pre_train'
logger_args = {'eval_episodes': 0, 'eval_epochs': 0, 'total_eval_steps': 0, 'total_train_steps': 0, ...}

    def call_event(self, event, logger_args):
        for logger in self.loggers:
>           getattr(logger, event)(
                logger_args=logger_args, output_dir=self.output_dir, model=self.model
            )
E           TypeError: pre_train() got an unexpected keyword argument 'logger_args'

src/tbp/monty/frameworks/loggers/exp_logger.py:77: TypeError
```

carver:
I did a bit of reading on one of the core issues that kept coming up (subclass implementations that don't use one of the arguments). I found this interesting: https://github.com/astral-sh/ruff/issues/15952

One of the key recommendations was to use the `typing.override()` decorator to indicate to ruff that it's a subclass definition. Basically another way to `noqa: ARG002` but perhaps a little prettier, and only for particular contexts. It seems to work fine for me when fixing the logger issue mentioned just above.

The downside is that it's less clear what problem it's solving, when a dev reads the decorator for the first time.

carver:
Thanks for the deep review! I think this is ready for the tests to run again. If they pass, then it is ready for the next look.

I did put one change in a separate PR: https://github.com/thousandbrainsproject/tbp.monty/pull/550 -- I'm happy to merge in either order and handle the rebasing. Or to just fold those changes in here.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -125,6 +125,7 @@ def __init__(
                 other options set a fixed threshold that does not take MLH evidence into
                 account. In [int, float, '[int]%', 'mean', 'median', 'all',
                 'x_percent_threshold']. Defaults to 'all'.
+                Value is currently ignored.
```
tristanls-tbp: **suggestion (non-blocking)**: "Ignored by the DefaultHypothesesUpdater. Required for compatibility with the HypothesesUpdater protocol."

src/tbp/monty/typing.py
tristanls-tbp, line 1: **praise**: Nice solution, thank you.

**suggestion (non-blocking)**: Rename to `typing.py`. `_compat` suffix is unnecessary as there's no other context for this file right now and an import will be `from tbp.monty.typing`, which is fine.

**question**: Any specific need to keep this behind `TYPE_CHECKING`? Why can't we `from typing_extensions import override` everywhere?

tests/unit/frameworks/models/salience/sensor_module_test.py
```diff
@@ -67,9 +73,10 @@ def test_step_snapshots_raw_observation_as_needed(
         else:
             self.sensor_module._snapshot_telemetry.raw_observation.assert_not_called()  # type: ignore[attr-defined]
 
-    def test_step_returns_no_percept(self, on_object_observation: MagicMock) -> None:
+    def test_step_returns_no_percept(self) -> None:
         self.assertIsNone(self.sensor_module.step(MagicMock()))
 
+    @patch("tbp.monty.frameworks.models.salience.sensor_module.on_object_observation")
     def test_step_proposes_goals_properly(
```
tristanls-tbp, line 80: **issue**: I think it should be sufficient to `@patch` here (like you have) and remove `@pytest.mark.usefixtures` from the class. If I'm reading this correctly, the rest of the class does not use `on_object_observation` and that scaffolding can be removed.

src/tbp/monty/frameworks/utils/plot_utils_dev.py
```diff
@@ -1189,13 +1189,16 @@ def plot_up_to_step(self, total_steps=None):
         for current_step in range(total_steps):
             self.plot_movement_step(current_step)
 
-    def plot_animation(self, zoom=1.0, view=None):
-        """Plot an animation of the episode's full action policy."""
+    def plot_animation(self, _zoom=1.0, view=None):
+        """Plot an animation of the episode's full action policy.
+
+        zoom value is currently not ignored.
```
tristanls-tbp: **issue**: I'm confused by this statement. It seems counter to the `_zoom` naming convention and the fact that it _is_ currently being ignored (in commented out code). Is this sentence as intended? Is it necessary?

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -111,9 +112,10 @@ class BasicWandbTableStatsHandler(WandbHandler):
     def log_level(cls):
         return "BASIC"
 
-    def report_episode(self, data, output_dir, episode, mode="train", **kwargs):
+    def report_episode(self, data, _episode, mode="train", **_kwargs):
         ###
         # Log basic statistics
+        # Ignore the episode value
```
tristanls-tbp, line 119: **nitpick**: I think comments like these are unnecessary. The convention of prefixing a variable with `_` means exactly this.

tristanls-tbp: **issue**: Shouldn't this be `@override` instead?, like on line ~253 below?

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -97,7 +98,7 @@ def post_init(self):
     def log_level(cls):
         return ""
 
-    def report_episode(self, data, output_dir, mode="train", **kwargs):
+    def report_episode(self, data, mode="train", **_kwargs):
```
tristanls-tbp: **note**: So odd that it would complain about `kwargs` being unused but not `data` nor `mode` 🤔.

src/tbp/monty/frameworks/loggers/monty_handlers.py
```diff
@@ -223,7 +223,8 @@ def __init__(self):
         """
         self.reports_per_file = {}
 
-    def report_episode(self, data, output_dir, episode, mode="train", **kwargs):
+    def report_episode(self, data, output_dir, _episode, mode="train", **_kwargs):
```
tristanls-tbp: **issue**: `@override`?

tests/unit/frameworks/models/salience/sensor_module_test.py
carver: I tried it that way first. At least one (two, IIRC) of the other tests fail in that approach, because `on_object_observation()` is called somewhere deeper inside the test. It raises an exception with a message like "cannot compare Mock and int with <" from inside an `on_object_observation()` call. Reintroducing the class-wide patch solved the problem.

I have developed an aversion to monkey-patch mocks, and this is a good example why. It's hard to reason about when and how the patch will work or break, or that it's testing what you think it is. But I'll just muddle through and do my best with the current approach, rather than try to rebuild the test a different way.

src/tbp/monty/typing.py
carver: Great, moved to `typing`

I seem to remember running into a failure somewhere with a global `typing_extensions` but I can't remember what. I'll try it again to see if I can reproduce. If it does work everywhere then there's not really any reason for the `tbp.monty.typing` anyway.

Actually, it's possible I was doing something silly like trying `from typing import override` and then seeing it was [not available until py3.12](https://docs.python.org/3/library/typing.html#typing.override), and forgetting to switch to `typing_extensions` before I went down the compatibility module rabbit-hole.

ETA: yeah I think that was it. 🤦🏻‍♂️ Ok, switched to use `typing_extensions` everywhere, and dropped the compatibility module.

src/tbp/monty/frameworks/utils/plot_utils_dev.py
carver: 😆 yeah I think this was a "not used" mixed up with "ignored". Anyhow, we use the _ prepend enough places to know that it means the field is not used.

Also, I thought I was being clever by changing the `zoom` to `_zoom` down in the commented out code. But if it gets uncommented, then the `_zoom` should lose it's prepend anyway, so I reverted that.

src/tbp/monty/frameworks/loggers/wandb_handlers.py
carver: Hm, yeah I will try to standardize on `@override`.

On a further review I like the solution less, but it still seems like the best one we have, without a larger refactor.

```py
class MontyHandler(metaclass=abc.ABCMeta):
    @abc.abstractmethod
    def report_episode(self, **kwargs):
        pass

class WandbHandler(MontyHandler):
    """Parent class for wandb loggers."""
    def report_episode(self, data, mode="train", **_kwargs):
        pass

class BasicWandbTableStatsHandler(WandbHandler):
    """Log LM episode stats to wandb as tables."""

    def report_episode(self, data, _episode, mode="train", **_kwargs):
        ...
```

In fact I would say these classes are pretty clearly breaking the [LSV](https://en.wikipedia.org/wiki/Liskov_substitution_principle) principle. An `@override` isn't super satisfying for classes that fail that.

src/tbp/monty/frameworks/loggers/wandb_handlers.py
carver: Hm, no, I think I got a little trigger-happy with a search/replace macro. I'll revert.

carver:
I think this is good to go once CI turns green.

I preferred `@override` in even more places, wherever I could justify it. The main places I didn't were because:
1) there was no superclass with the same method name, or 
2) `__init__` which is a special case that standard inheritance doesn't apply for

tests/unit/frameworks/models/salience/sensor_module_test.py
tristanls-tbp: Ah, ok 👍.

src/tbp/monty/frameworks/loggers/wandb_handlers.py
tristanls-tbp: You may notice that there are multiple instances in the code where subclasses modify the signature of the overridden method. We'll work our way through those eventually.

tristanls-tbp (approved):
Thank you for the updates. Nice to have this rule in place.

tristanls-tbp:
I'm going to run some benchmarks to verify that kwargs weren't a factor in action samplers.

tristanls-tbp (changes requested):
Something in this PR is breaking the one benchmark that we initiate with `run.py` (instead of `run_parallel.py`).

The benchmark `python benchmarks/run.py base_10multi_distinctobj_dist_agent` finishes too early and without data.

I ran that same benchmark for `main@c2d3a46` and it's working as intended in that we can see that it is progressing beyond the point where this PR's benchmark already stopped.

Image

carver:
I re-added the kwargs to the action samplers. I don't see how removing them would fail silently, though. If unrecognized keyword args were passed in, I would expect a `TypeError` crash.

I looked around with an eye to silent failures, and the wandb logger changes are the only things I can imagine causing a change in behavior without a crash. So I reverted those changes and depend entirely on `@override`. I think with those two changes it is worth running the benchmarks again.

tristanls-tbp:
@carver, FYI, you can run benchmarks and compare between `main` and this PR as well: `python benchmarks/run.py base_10multi_distinctobj_dist_agent`. The numbers won't be the same between your runs and our infrastructure, so we won't be able to see if the benchmarks changed, but if you run a benchmark on `main` on your machine and get a set of numbers, then that set of numbers should be the same if you run a benchmark on this PR branch on your machine. Does that make sense?

The only difference between your machine and our infrastructure should be the actual number values, and for final verification, we'll only use our infrastructure numbers, but if it doesn't run for you, it's unlikely to run on our infrastructure.

carver:
Ah, it _was_ an exception. Just not visible in that screenshot above:
```
  File "benchmarks/run.py", line 63, in <module>
    main(all_configs=CONFIGS, experiments=cmd_args.experiments)
  File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/run.py", line 114, in main
    run(exp_config)
  File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/run.py", line 58, in run
    exp.evaluate()
  File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/experiments/monty_experiment.py", line 583, in evaluate
    self.run_epoch()
  File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/experiments/monty_experiment.py", line 537, in run_epoch
    self.run_episode()
  File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/experiments/object_recognition_experiments.py", line 41, in run_episode
    self.post_episode(last_step)
  File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/experiments/monty_experiment.py", line 508, in post_episode
    self.logger_handler.post_episode(self.logger_args)
  File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/loggers/exp_logger.py", line 54, in post_episode
    self.call_event("post_episode", logger_args)
  File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/loggers/exp_logger.py", line 79, in call_event
    getattr(logger, event)(
  File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/loggers/graph_matching_loggers.py", line 201, in post_episode
    self.log_episode(logger_args, output_dir, model)
  File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/loggers/graph_matching_loggers.py", line 183, in log_episode
    handler.report_episode(self.data, output_dir, episode, mode)
TypeError: report_episode() takes from 3 to 4 positional arguments but 5 were given
```
Local tests confirm that the benchmark can run again after reverting the argument changes in `report_episode()` and using `@override` instead.

carver:
> you can run benchmarks and compare between `main` and this PR

~Is the thing I should be comparing `~/tbp/results/monty/projects/evidence_eval_runs/logs/base_10multi_distinctobj_dist_agent/eval_stats.csv`?~ I see [these instructions for results](https://thousandbrainsproject.readme.io/docs/running-benchmarks#where-to-find-the-results) which might be the same thing. I'll give it a try.

I already got it to stop crashing, but I haven't done both full runs yet.

tristanls-tbp:
I think that with your latest changes, this PR is good to go. However, I'm also running numerous benchmarks for #540. I haven't forgotten this PR, we'll get this merged, I just need to find the time to shepherd benchmarks for this PR again.

carver:
I ran `base_10multi_distinctobj_dist_agent` against main and this PR. In the summary columns on wandb, the only values that varied were the runtime by a small %, so I think everything else looks good.

So I will be hands-off on this PR until you ask for something new (or if I notice that the branch cannot be cleanly merged into main).

tristanls-tbp:
I ran the benchmarks and there's no changes, as expected

Screenshot 2025-11-17 at 11 43 22

PR #544, opened 2025-11-06
chore: follow ruff rule PTH208

carver:
Replace `os.listdir(pathstr)` with `Path(pathstr).iterdir()`.

Earlier versions of this made broader Path-related improvements, like replacing some nearby `join`s. They have mostly been moved out, except where it was especially awkward.

The code will get cleaner later, when every path variable is expected to be a `Path` object.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -234,8 +234,6 @@ ignore = [
     "ANN204", # ANN204: Missing return type annotation for special method
     "ANN206", # ANN206: Missing return type annotation for classmethod
     "ANN401", # ANN401: Dynamically typed expressions (typing.Any) are disallowed in {name}
-    # TODO: Address ARG002 right away
-    "ARG002", # ARG002: Unused method argument
     "COM812", # COM812: Trailing comma missing
     "DTZ005", # DTZ005: `datetime.datetime.now()` called without a `tz` argument
     "EM101", # EM101: Exception must not use a string literal, assign to variable first
--- a/src/tbp/monty/frameworks/actions/action_samplers.py
+++ b/src/tbp/monty/frameworks/actions/action_samplers.py
@@ -9,7 +9,7 @@
 # https://opensource.org/licenses/MIT.
 from __future__ import annotations
 
-from typing import Any, Callable
+from typing import Callable
 
 import quaternion as qt
 from numpy import cos, pi, sin, sqrt
@@ -98,7 +98,6 @@ class ConstantSampler(ActionSampler):
         rotation_degrees: float = 5.0,
         rotation_quat: QuaternionWXYZ | None = None,
         translation_distance: float = 0.004,
-        **kwargs: Any,  # Accept arbitrary keyword arguments for compatibility
     ) -> None:
         super().__init__(actions=actions, rng=rng)
         self.absolute_degrees = absolute_degrees
@@ -192,7 +191,6 @@ class UniformlyDistributedSampler(ActionSampler):
         max_translation: float = 0.05,
         min_translation: float = 0.05,
         rng: Generator | None = None,
-        **kwargs: Any,  # Accept arbitrary keyword arguments for compatibility
     ):
         super().__init__(actions=actions, rng=rng)
         self.max_absolute_degrees = max_absolute_degrees
--- a/src/tbp/monty/frameworks/environment_utils/transforms.py
+++ b/src/tbp/monty/frameworks/environment_utils/transforms.py
@@ -51,7 +51,7 @@ class MissingToMaxDepth:
         self.threshold = threshold
         self.needs_rng = False
 
-    def __call__(self, observation, state=None):
+    def __call__(self, observation, _state=None):
         """Replace missing depth values with max_depth.
 
         Args:
@@ -83,7 +83,7 @@ class AddNoiseToRawDepthImage:
         self.sigma = sigma
         self.needs_rng = True
 
-    def __call__(self, observation, state=None):
+    def __call__(self, observation, _state=None):
         """Add gaussian noise to raw sensory input.
 
         Args:
@@ -136,7 +136,7 @@ class GaussianSmoothing:
         self.kernel = self.create_kernel()
         self.needs_rng = False
 
-    def __call__(self, observation, state=None):
+    def __call__(self, observation, _state=None):
         """Apply gaussian smoothing to depth images.
 
         Args:
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -804,8 +804,8 @@ class OmniglotEnvironmentInterface(EnvironmentInterfacePerObject):
         rng,
         transform=None,
         parent_to_child_mapping=None,
-        *args,
-        **kwargs,
+        *_args,
+        **_kwargs,
     ):
         """Initialize environment interface.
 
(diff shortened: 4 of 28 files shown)


benchmarks/configs/follow_ups/__init__.py
```diff
@@ -17,11 +16,10 @@
...
-files = [file for file in os.listdir(current_dir) if file.endswith(".pkl")]
-names = [file.split(".")[0] for file in files]
 
 CONFIGS = {}
-for file, name in zip(files, names):
-    with open(os.path.join(current_dir, file), "rb") as f:
+for path in current_dir.glob("*.pkl"):
+    with path.open("rb") as f:
         config = pickle.load(f)
+        name = path.stem
```
carver, line 24: You may notice that the behavior changes on files with multiple `.`s in the name. I took a best guess that this new behavior is better.

The old behavior, with these two files:
- `executive.officer.for.the.week.pkl`
- `executive.assistant.pkl`

would be for one of the configs (arbitrary choice) to be saved to `CONFIGS['executive']`

The new behavior with the same two files, would be to save two different configs under:
- `CONFIGS['executive.officer.for.the.week']`
- `CONFIGS['executive.assistant']`

Thoughts?

src/tbp/monty/frameworks/config_utils/make_env_interface_configs.py
```diff
@@ -469,41 +465,35 @@ def get_omniglot_eval_env_interface(
...
         OmniglotEnvironmentInterfaceArgs for evaluation.
     """
-    if data_path is None:
-        data_path = os.path.join(os.environ["MONTY_DATA"], "omniglot/python/")
-    if os.path.exists(data_path):
-        alphabet_folders = [
-            a for a in os.listdir(data_path + "images_background") if a[0] != "."
-        ]
-    else:
+    data_path = make_data_path(data_path, "omniglot" / "python")
```
carver: bug

src/tbp/monty/frameworks/config_utils/make_env_interface_configs.py
```diff
@@ -469,41 +465,35 @@ def get_omniglot_eval_env_interface(
...
-                )
+                versions_of_char = list(char_path.iterdir())
                 num_versions = len(versions_of_char) - start_at_version
 
-            for v_idx in range(num_versions + start_at_version):
-                if v_idx >= start_at_version:
-                    all_alphabet_idx.append(a_idx)
-                    all_character_idx.append(c_idx + 1)
-                    all_version_idx.append(v_idx + 1)
+            for v_idx in range(start_at_version, num_versions + start_at_version):
```
carver: This change is unrelated to the `Path` changes. I was just doing a [broken window](https://blog.codinghorror.com/the-broken-window-theory/) fix. I can move it to another commit, or another PR, if preferred. Or you may prefer another variation, like:

```py
for version in range(num_versions):
  v_idx = start_at_version + version
```

carver (commented):
In the process of working on `get_omniglot_train_env_interface()`, I found a bug that never showed up in testing. When I looked further, I couldn't find any invocations of the method at all (but maybe there's some dark magic invocation that I'm not finding). Should I just delete the method as unused?

Same for `get_omniglot_eval_env_interface()`.

src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -212,23 +210,25 @@ def reset(self):
...
-        img_char_dir = os.path.join(
-            self.data_path,
-            "images_background",
-            self.current_alphabet,
-            "character" + str(self.character_id).zfill(2),
+        img_char_dir = (
+            self.data_path
+            / "images_background"
+            / self.current_alphabet
+            / f"character{self.character_id:02}"
```
carver: If I don't update this to an f-string, then ruff tries to reformat the line to something particularly confusing/ugly, like:
```suggestion
            self.data_path / "images_background" / self.current_alphabet / "character"
            + str(self.character_id).zfill(2)
```
It's worth avoiding that split of the name of the last path element onto two lines. The best option I found is this f-string update.

carver:
I'm going to minimize this a bit, down to just the listdir changes as much as possible, and leave the [windows broken](https://blog.codinghorror.com/the-broken-window-theory/).

benchmarks/configs/follow_ups/__init__.py
tristanls-tbp: **praise**: Nice, much better.

src/tbp/monty/frameworks/config_utils/make_env_interface_configs.py
```diff
@@ -426,24 +427,19 @@ def get_omniglot_train_env_interface(num_versions, alphabet_ids, data_path=None)
...
     for a_idx in alphabet_ids:
-        alphabet = alphabet_folders[a_idx]
-        characters_in_a = list(os.listdir(data_path + "images_background/" + alphabet))
-        for c_idx, character in enumerate(characters_in_a):
-            versions_of_char = list(
-                os.listdir(
-                    data_path + "images_background/" + alphabet + "/" + character
-                )
-            )
+        alphabet_children = alphabet_folders[a_idx].iterdir()
```
tristanls-tbp: **issue**: This should be `characters_in_a` (as in, "characters in an alphabet").

src/tbp/monty/frameworks/config_utils/make_env_interface_configs.py
```diff
@@ -478,25 +474,21 @@ def get_omniglot_eval_env_interface(
...
         return OmniglotEnvironmentInterfaceArgs()
+
     all_alphabet_idx = []
     all_character_idx = []
     all_version_idx = []
     for a_idx in alphabet_ids:
-        alphabet = alphabet_folders[a_idx]
-        characters_in_a = list(os.listdir(data_path + "images_background/" + alphabet))
-        for c_idx, character in enumerate(characters_in_a):
+        alphabet_children = alphabet_folders[a_idx].iterdir()
```
tristanls-tbp: **issue**: `characters_in_a` again please.

src/tbp/monty/frameworks/experiments/profile.py
```diff
@@ -119,15 +120,10 @@ def evaluate(self):
...
         if len(self.wandb_handlers) > 0:
-            profile_files = os.listdir(self.profile_dir)
-            profile_paths = [
-                os.path.join(self.profile_dir, file) for file in profile_files
-            ]
-            csv_files = [i for i in profile_paths if i.endswith(".csv")]
-
-            for csv in csv_files:
+            profile_path = Path(self.profile_dir)
+            for csv in profile_path.glob("*.csv"):
```
tristanls-tbp, line 124: **praise**: Much nicer.

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -90,7 +87,8 @@ def load_models_from_dir(exp_path, pretrained_dict=None):
             pretrained_models = state_dict["lm_dict"][lm_id]["graph_memory"]
             lm_models["pretrained"][lm_id] = pretrained_models
 
-    for folder in os.listdir(exp_path):
+    for child in Path(exp_path).iterdir():
+        folder = child.name
```
tristanls-tbp, line 91: **suggestion (non-blocking)**:
```python
for folder in [f.name for f in Path(exp_path).iterdir()]:
   ...
```

carver: Soon, the `os.path.join` will be updated to `child / "model.pt"`, so we will want access to the full path.

I'll leave it this way unless you change your mind and decide it's blocking.

carver:
Force-pushed to work around #556

PR #545, opened 2025-11-10
refactor!: follow ruff rule PTH118

carver:
Switch to the Path-style "/" joining of paths instead of `os.path.join()`.

There are probably some places where path arguments are guaranteed to be `Path` already, and they are being recast to `Path()` anyway. Since this PR is so big, I wanted to be able to reason about it *locally* as much as possible, and then someone can do follow-up work to skip unnecessary casts.

### DIFF
--- a/src/tbp/monty/frameworks/models/evidence_sdr_matching.py
+++ b/src/tbp/monty/frameworks/models/evidence_sdr_matching.py
@@ -76,7 +76,7 @@ class LoggerSDR:
         """
         if hasattr(self, "path"):
             np.save(
-                self.path / f"episode_{str(self.episode).zfill(3)}.npy",
+                self.path / f"episode_{self.episode:03d}.npy",
                 data,
             )
             self.episode += 1


src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -81,11 +81,11 @@ def __init__(self, patch_size=10, data_path=None):
...
         self.state = 0
-        self.data_path = data_path
-        if self.data_path is None:
-            self.data_path = os.path.join(os.environ["MONTY_DATA"], "omniglot/python/")
-        data_path = Path(self.data_path)
-        alphabet_path = data_path / "images_background"
+        if data_path is None:
+            self.data_path = Path(os.environ["MONTY_DATA"]) / "omniglot/python"
+        else:
+            self.data_path = Path(data_path)
```
carver, line 87: I started making a utility function to replace these 4 lines, but then decided that it belongs in another PR because this one is already so giant.

I imagine something like:
```py
self.data_path = make_data_path(data_path, "omniglot/python")
```

src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -213,17 +213,17 @@ def reset(self):
...
-        img_char_dir = os.path.join(
-            self.data_path,
-            "images_background",
-            self.current_alphabet,
-            "character" + str(self.character_id).zfill(2),
+        img_char_dir = (
+            self.data_path
+            / "images_background"
+            / self.current_alphabet
+            / f"character{self.character_id:02d}"
```
carver, line 220: ruff format splits "character" and the ID to different lines unless I combine them like this.

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -82,15 +81,15 @@ def load_models_from_dir(exp_path, pretrained_dict=None):
...
+        state_dict = torch.load(Path(pretrained_dict) / "model.pt")
         for lm_id in list(state_dict["lm_dict"].keys()):
             pretrained_models = state_dict["lm_dict"][lm_id]["graph_memory"]
             lm_models["pretrained"][lm_id] = pretrained_models
 
     for child in Path(exp_path).iterdir():
         folder = child.name
         if folder.isnumeric():
-            state_dict = torch.load(os.path.join(exp_path, folder, "model.pt"))
+            state_dict = torch.load(child / "model.pt")
```
carver, line 92: I'm open to more useful names than `child` here. I don't know the folder structure well enough to make a good guess. Maybe just `experiment`?

tests/unit/run_test.py
```diff
@@ -158,49 +158,43 @@ def test_main_without_experiment(self):
         sys.argv = ["monty", "-e", "test_1"]
         main(all_configs=self.CONFIGS)
 
-        output_dir = os.path.join(
-            self.CONFIGS["test_1"]["logging_config"].output_dir, "test_1"
-        )
-        with open(os.path.join(output_dir, "fake_log.pkl"), "rb") as f:
+        output_dir = Path(self.CONFIGS["test_1"]["logging_config"].output_dir)
+
+        with open(output_dir / "test_1" / "fake_log.pkl", "rb") as f:
```
carver, line 163: Notice that `test_1` is not added to output_dir anymore, and is joined during the `open()` instead. `output_dir` is not used elsewhere, so there's no effective change to the test.

tests/unit/run_test.py
```diff
@@ -213,8 +207,9 @@ def test_importing_existing_configs(self):
             # We want the _last_ "tests" index
             base_repo_idx += 1
         base_dir = pathlib.Path(*current_file_path.parts[:base_repo_idx])
-        exp_dir = os.path.join(str(base_dir), "benchmarks")
-        sys.path.append(exp_dir)
+        exp_dir = base_dir / "benchmarks"
+        assert (exp_dir / "configs").exists()
+        sys.path.append(str(exp_dir))
```
carver, line 212: This had a bug earlier, that was causing the following import to fail. I added an `assert` for my own sanity. Then I found out that `sys.path.append` ignores any `append()` calls that aren't strings. Let me know if you'd prefer that I drop the assert from this PR.

src/tbp/monty/frameworks/environments/two_d_data.py
jeremyshoemaker: What's the alternative look like? This looks reasonable to me, but I'm curious what we're avoiding.

benchmarks/configs/infer_compositional_objects.py
```diff
@@ -60,34 +61,35 @@
 
 # For an explanation of the different levels of difficulty, see logos_on_objs.py
 
-model_path_monolithic_models_lvl1 = os.path.join(
-    pretrained_dir,
-    "supervised_pre_training_objects_with_logos_lvl1_monolithic_models/pretrained/",
+model_path_monolithic_models_lvl1 = (
+    pretrained_dir
+    / "supervised_pre_training_objects_with_logos_lvl1_monolithic_models/pretrained/"
```
jeremyshoemaker: Could you take the `/`s out of the string and use the operator to join them? I know the original didn't do this, but the original was also problematic since `/` isn't guaranteed to be a path separator on all platforms.
```suggestion
    / "supervised_pre_training_objects_with_logos_lvl1_monolithic_models"
    / "pretrained"
```
Same with the others in this file.

benchmarks/configs/learn_compositional_objects.py
```diff
@@ -184,9 +183,8 @@
     experiment_args=SupervisedPretrainingExperimentArgs(
         n_train_epochs=len(LOGO_POSITIONS) * len(LOGO_ROTATIONS),
         supervised_lm_ids=["learning_module_0"],
-        model_name_or_path=os.path.join(
-            fe_pretrain_dir,
-            "supervised_pre_training_flat_objects_wo_logos/pretrained/",
+        model_name_or_path=(
+            fe_pretrain_dir / "supervised_pre_training_flat_objects_wo_logos/pretrained"
```
jeremyshoemaker: Same here with the `/` in the string instead of using the operator. And below as well.

benchmarks/configs/monty_world_experiments.py
```diff
@@ -52,9 +51,8 @@
 therefore ensure you omit the -m flag when running these
 """
 
-model_path_numenta_lab_obj = os.path.join(
-    pretrained_dir,
-    "surf_agent_1lm_numenta_lab_obj/pretrained/",
+model_path_numenta_lab_obj = (
+    pretrained_dir / "surf_agent_1lm_numenta_lab_obj/pretrained"
```
jeremyshoemaker: Same here with using a `/` in the string instead of the operator.

benchmarks/configs/monty_world_habitat_experiments.py
```diff
@@ -42,9 +41,8 @@
 
 test_rotations_one = [[0, 0, 0]]
 
-model_path_numenta_lab_obj = os.path.join(
-    pretrained_dir,
-    "surf_agent_1lm_numenta_lab_obj/pretrained/",
+model_path_numenta_lab_obj = (
+    pretrained_dir / "surf_agent_1lm_numenta_lab_obj/pretrained"
```
jeremyshoemaker: Same here with using a `/` in the string instead of the operator.

benchmarks/configs/pretraining_experiments.py
```diff
@@ -61,14 +62,12 @@
...
 
-fe_pretrain_dir = os.path.expanduser(
-    os.path.join(monty_models_dir, "pretrained_ycb_v10")
-)
+fe_pretrain_dir = monty_models_dir.expanduser() / "pretrained_ycb_v10"
 
-pre_surf_agent_visual_training_model_path = os.path.join(
-    fe_pretrain_dir, "supervised_pre_training_all_objects/pretrained/"
+pre_surf_agent_visual_training_model_path = (
+    fe_pretrain_dir / "supervised_pre_training_all_objects/pretrained/"
```
jeremyshoemaker: Same here with using a `/` in the string instead of the operator.

benchmarks/configs/ycb_experiments.py
```diff
@@ -115,34 +114,22 @@
 # runs with all 77 YCB objects.
 test_rotations_3 = test_rotations_all[:3]
 
-model_path_10distinctobj = os.path.join(
-    pretrained_dir,
-    "surf_agent_1lm_10distinctobj/pretrained/",
-)
+model_path_10distinctobj = pretrained_dir / "surf_agent_1lm_10distinctobj/pretrained/"
```
jeremyshoemaker: Same here with using a `/` in the string instead of the operator. And below as well.

benchmarks/make_detailed_follow_up_configs.py
```diff
@@ -94,7 +94,7 @@
     # Pickle these configs in the follow_ups directory
     follow_up_dir = pathlib.Path(__file__).parent / "configs/follow_ups"
     follow_up_name = f"{experiment}_{follow_up_suffix}"
-    file_path = os.path.join(follow_up_dir, follow_up_name + ".pkl")
+    file_path = follow_up_dir / follow_up_name + ".pkl"
```
jeremyshoemaker: We could use `with_suffix` here, or alternatively an f-string. I think we should probably avoid the string concatination here. The `pathlib` use of the `/` operator while sensible could be confused for math with the `+` here too. I'm also not sure if operator precedence could cause trouble here.

Looking at both options, I think I prefer the f-string since it avoids parentheses.
```suggestion
    file_path = follow_up_dir / f"{follow_up_name}.pkl"
```

benchmarks/make_detailed_follow_up_configs.py
```diff
@@ -94,7 +94,7 @@
     # Pickle these configs in the follow_ups directory
     follow_up_dir = pathlib.Path(__file__).parent / "configs/follow_ups"
```
jeremyshoemaker: Same here with using a `/` in the string instead of the operator.

src/tbp/monty/frameworks/config_utils/config_args.py
```diff
@@ -88,9 +89,7 @@ class LoggingConfig:
     python_log_level: str = "INFO"
     python_log_to_file: bool = True
     python_log_to_stderr: bool = True
-    output_dir: str = os.path.expanduser(
-        os.path.join(monty_logs_dir, "projects/monty_runs/")
-    )
+    output_dir: Path = monty_logs_dir / "projects/monty_runs"
```
jeremyshoemaker: Same here with using a `/` in the string instead of the operator. And below as well.

src/tbp/monty/frameworks/config_utils/make_env_interface_configs.py
```diff
@@ -91,30 +91,24 @@ class SupervisedPretrainingExperimentArgs(ExperimentArgs):
 # Data-set containing RGBD images of real-world objects taken with a mobile device
 @dataclass
 class EnvInitArgsMontyWorldStandardScenes:
-    data_path: str = os.path.join(
-        os.environ["MONTY_DATA"], "worldimages/standard_scenes/"
-    )
+    data_path = Path(os.environ["MONTY_DATA"]) / "worldimages/standard_scenes/"
```
jeremyshoemaker: Same here with using a `/` in the string instead of the operator. And below as well.

src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -81,11 +81,11 @@ def __init__(self, patch_size=10, data_path=None):
         self.rotation = qt.from_rotation_vector([np.pi / 2, 0.0, 0.0])
         self.step_num = 0
         self.state = 0
-        self.data_path = data_path
-        if self.data_path is None:
-            self.data_path = os.path.join(os.environ["MONTY_DATA"], "omniglot/python/")
-        data_path = Path(self.data_path)
-        alphabet_path = data_path / "images_background"
+        if data_path is None:
+            self.data_path = Path(os.environ["MONTY_DATA"]) / "omniglot/python"
```
jeremyshoemaker: Same here with using a `/` in the string instead of the operator. And below as well.

src/tbp/monty/frameworks/environments/two_d_data.py
jeremyshoemaker: It's a good idea, and I appreciate you deciding to separate it out into a separate change. 😀

src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -328,7 +332,7 @@ def init_loggers(self, logging_config: dict[str, Any]) -> None:
         self.python_log_level = logging_config["python_log_level"]
         self.log_to_file = logging_config["python_log_to_file"]
         self.log_to_stderr = logging_config["python_log_to_stderr"]
-        self.output_dir = logging_config["output_dir"]
+        self.output_dir = Path(logging_config["output_dir"])
         self.run_name = logging_config["run_name"]
 
         if not os.path.exists(self.output_dir):
```
jeremyshoemaker, line 338: Isn't there a `pathlib` version of this method on Path objects? Could fix it while you're here.

src/tbp/monty/frameworks/utils/logging_utils.py
jeremyshoemaker: Another option could be to change `child` to `folder` and replace the uses of `folder` below with `child.name`. That seems reasonable to me. But I'm also fine with leaving this as-is for now.

src/tbp/monty/simulators/habitat/configs.py
```diff
@@ -82,7 +83,7 @@ class EnvInitArgs:
     )
     scene_id: int | None = field(default=None)
     seed: int = field(default=42)
-    data_path: str = os.path.join(os.environ["MONTY_DATA"], "habitat/objects/ycb")
+    data_path: str = Path(os.environ["MONTY_DATA"]) / "habitat/objects/ycb"
```
jeremyshoemaker: Same here with using a `/` in the string instead of the operator.

tests/unit/embodied_data_test.py
```diff
@@ -364,10 +360,8 @@ def test_saccade_on_image_stream_env_interface(self):
         patch_size = 48
         expected_keys = ["depth", "rgba", "pixel_loc"]
 
-        data_path = "./resources/dataloader_test_images/"
-        data_path = os.path.join(
-            Path(__file__).parent,
-            "resources/dataloader_test_images/0_numenta_mug/",
+        data_path = (
+            Path(__file__).parent / "resources/dataloader_test_images/0_numenta_mug"
```
jeremyshoemaker: Same here with using a `/` in the string instead of the operator.

tests/unit/run_test.py
jeremyshoemaker: Looks fine to me. We're assuming it exists, might as well assert it since it's a test.

tools/github_readme_sync/hierarchy.py
```diff
@@ -68,13 +70,15 @@ def write_document(file, parent_slug, doc, indent_level):
 
 
 def check_hierarchy_file(folder: str):
+    folder = Path(folder)
     hierarchy = []
 
-    if not os.path.exists(os.path.join(folder, HIERARCHY_FILE)):
-        logging.error(f"File {os.path.join(folder, HIERARCHY_FILE)} does not exist")
+    hierarchy_file = folder / HIERARCHY_FILE
+    if not os.path.exists(hierarchy_file):
```
jeremyshoemaker, line 77: I think this can be replaced with a Path method call.
```suggestion
    if not hierarchy_file.exists():
```

tools/github_readme_sync/hierarchy.py
```diff
@@ -110,9 +114,7 @@ def check_hierarchy_file(folder: str):
             parent_stack[-1]["children"].append(new_doc)
             parent_stack.append(new_doc)
 
-            full_path = (
-                os.path.join(folder, *(el["slug"] for el in parent_stack)) + ".md"
-            )
+            full_path = folder / ("/".join(el["slug"] for el in parent_stack) + ".md")
```
jeremyshoemaker: Hmm. I'd prefer to not use `/` in strings, since I've been commenting on it this whole time. I can't think of how I'd rewrite this though.

But looking at the `pathlib` docs, it seems like the library handles them in a safe way. Maybe that makes it moot?

tools/print_version/cli.py
```diff
@@ -10,15 +10,14 @@
...
 
 import semver
 
 
 def get_version():
-    version_module_path = os.path.join(
-        os.path.dirname(__file__), "../../src/tbp/monty/__init__.py"
-    )
+    project_root = Path(__file__).resolve().parent.parent.parent
+    version_module_path = project_root / "src/tbp/monty/__init__.py"
```
jeremyshoemaker: Same here with using a `/` in the string instead of the operator.

benchmarks/configs/infer_compositional_objects.py
carver: Yeah, I felt ambivalent about this choice. On one hand, there's something satisfying about breaking it apart and seeing the path separation that way. On the other hand, internal slashes make it much more compact, and there are many places both in this code and in general where people are expected to understand slash-separated paths.

I landed on slightly preferring internal slashes when it keeps the line count down, but it's definitely not something I think is worth quibbling over. So I'll make the changes everywhere that this diff touches.

carver:
> Looks good overall. This is one I've been wanting to do for a while now. Nice it see it getting done.

Glad to hear it's valuable!

> I am concerned about whether this breaks anything because I've run into a few things in my current work where something expects a string instead of a Path, so I'll kick off the benchmarks for this as well as the tests.

Yeah, that would be great. I did happen to do a speed run against `base_10multi_distinctobj_dist_agent` before pushing, but that is obviously a pretty small surface area of testing.

src/tbp/monty/frameworks/experiments/monty_experiment.py
carver: Yes, for sure. There were a lot of similar places that I was tempted to clean it up and extend to the other path-style syntax. Based on my last PR review, and because this was so big already, I was biased to delaying other changes to later. I am intending to get through all the `Path`-related ruff rules (there is one for `exists()`) and then move onto something new.

tools/github_readme_sync/hierarchy.py
carver: I didn't spot it the first time, but on a quick doc review, [`.joinpath()`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.joinpath) seems to handle this nicely. I'll make the change

jeremyshoemaker: Oh, that's perfect. Much better to just splat the list comprehension into the args.

src/tbp/monty/frameworks/environments/two_d_data.py
carver:
```py
            self.data_path / "images_background" / self.current_alphabet / "character"
            + str(self.character_id).zfill(2)
```

jeremyshoemaker:
Benchmark results look good. Nothing appears to be different or broken.

src/tbp/monty/frameworks/environments/two_d_data.py
jeremyshoemaker: Ah, yeah I like the f-string solution better. There was another place I saw us using `zfill` that I think you changed that might be better with the same f-string modifier approach.

src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -474,10 +474,10 @@ def load_new_scene_data(self):
         """
         # Set data paths
         current_depth_path = (
-            self.data_path + f"{self.current_scene}/depth_{self.scene_version}.data"
+            self.data_path / f"{self.current_scene}/depth_{self.scene_version}.data"
```
jeremyshoemaker: There's a `/` in this f-string, and the one below on line 480.

carver: Ooh, quick eyes, I just pushed it! Funny, I just spotted them at around the same time.

(I had used a `\w\/\w` search to find them in the diff, which obviously didn't work on this case with the `}/`. The other one I missed I have no idea why, just a whoopsie.)

jeremyshoemaker: I noticed you got it, somehow before I submitted my review, so they must have passed each other. 😀

PR #547, opened 2025-11-11
refactor!: remove **kwargs from ActionSampler subclasses

carver:
These seem unused, and are causing failures in the ruff rule ARG002.

Peeled off from #541 

@tristanls-tbp recommended breaking the PR into smaller chunks. This change in particular was mentioned to run tests against.


carver:
Nevermind, folded into #541 . I realized that the recommendation to split up was only about a subset of the review.

PR #548, opened 2025-11-11
refactor!: remove output_dir from update_episode_data()

carver:
The GraphMatchingLogger's do not use the output dir in any update_episode_data implementations, so remove them all.

Peeled off of #541 

I'm still getting a feel for how big of a PR you folks like, so let me know if I over-corrected.


carver:
I thought this might be bigger, at first, and after re-reading the code review, it seems like it was fine to fold this change in to #541 .

PR #549, opened 2025-11-11
refactor!: remove object_scale from graph matching

carver:
The value is ignored. This is in preparation for enabling ruff's arg002

Split from #541


carver:
Nevermind, folding into #541

PR #550, opened 2025-11-11
refactor!: remove unused _compute_goal_confidence()

carver:
in EvidenceGoalStateGenerator. This had a carryover effect of making the second returned value of _compute_graph_mismatch() unused (target_location_separation). So that was also removed.

There is ambition to do something smarter than use the parent LM's confidence, but until that's implemented, let's remove this unnecessary pass-through method.

Split from #541 

Since there were some notes sketched in here, about ideas for how to calculate the confidence, I copied the notes into: https://thousandbrains.discourse.group/t/calculating-the-confidence-of-a-goal-state/833


tristanls-tbp (approved):
Very nice, and much easier to review in separation, thank you.

FYI @vkakerbeck

PR #551, opened 2025-11-11
feat: log rotation error in degrees instead of radians

vkakerbeck:
As discussed, since we are reporting rotation error in degrees in our docs it would simplify this process if it would also be logged in degrees. This change makes monty log in degrees everywhere (csv, json, wandb). It also updates the benchmark results tables accordingly.


vkakerbeck:
Okay, I reran all the benchmarks and there were no rounding differences. I updated the benchmark table for the compositional experiments to use degrees as the others.

vkakerbeck:
Oh awesome, good idea, thanks for adding this! I did run them on wandb and tagged with PR#551 :)

tristanls-tbp:
Adding the WandB results here as a screenshot for my future reference
Screenshot 2025-11-12 at 17 48 13

PR #553, opened 2025-11-11
feat: update how features are selected for evidence calculation

hlee9212:
Updated feature selection logic. Instead of checking if there are at least two feature counts, checks for existence of a non-morphological feature.

The previous version incorrectly returned False when I specified only "hsv" and "pose_vectors" (excluding "pose_fully_defined") in tolerances. It now returns True since there is at least one non-morphological feature, "hsv". 

**Note**: There is a TODO comment in Line 217 of `learning_moidule.py`:
```python
# TODO: not ideal solution
self.graph_memory.features_to_use = self.tolerances
```
In the future, we may need an overall better method or larger refactor to encapsulate which features are used for matching step.


tristanls-tbp:
**question**: Does this happen to affect any benchmarks?

vkakerbeck:
It shouldn't but @hlee9212 make sure to run at least one 77 object one to double check (if it would affect benchmarks it would affect all of them)

hlee9212:
CleanShot 2025-11-14 at 17 21 21

@vkakerbeck and @tristanls-tbp confirmed that this PR did not change the benchmark results.

ramyamounir:
Just leaving a quick note here that this PR changed the logos benchmark because it allowed for object ids to be used for matching between two stacked columns. This is how it should be, so the benchmarks were updated in [this PR](https://github.com/thousandbrainsproject/tbp.monty/pull/569)

PR #557, opened 2025-11-14
chore: stable wandb import sorts after local run

carver:
Sometimes after running a benchmark, wandb creates a local folder in the project root.

This confuses isort/ruff into thinking that wandb is a local python module, and it tries to move the import line down into the first party modules section.

By forcing ruff to recognize wandb as a third party module, the import grouping is always consistent.

See https://github.com/astral-sh/ruff/issues/10519 for more discussion

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# wandb local configs
+wandb/
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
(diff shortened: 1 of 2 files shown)


tristanls-tbp (approved):
Oh, that's unfortunate. Thank you for the fix!

PR #558, opened 2025-11-14
feat: add speedrun option for benchmarks

carver:
When debugging a benchmark failure, it's very nice to have a quick response whether it succeeds or fails.

This PR adds a `--speedrun` (and `-s`) flag to the benchmark runner, which shrinks the configured number of objects, epochs, etc in a predifined configuration.

Timing on my development machine:
- `python benchmarks/run.py -e base_10multi_distinctobj_dist_agent` takes ~42 minutes
- `python benchmarks/run.py -e base_10multi_distinctobj_dist_agent --speedrun` takes ~6 seconds

This makes it feasible to include some benchmark speedruns in CI.

I only ran the speedrun against that one benchmark, it's probably slow against some/all of the others. But support could be added incrementally from here.

This came up while debugging #541

### DIFF
--- a/benchmarks/run.py
+++ b/benchmarks/run.py
@@ -22,6 +22,7 @@ sys.path.insert(
 
 from benchmarks.configs.load import load_configs
 from benchmarks.configs.names import NAMES
+from tbp.monty.frameworks.config_utils import shrink_config
 from tbp.monty.frameworks.config_utils.cmd_parser import create_cmd_parser
 from tbp.monty.frameworks.run_env import setup_env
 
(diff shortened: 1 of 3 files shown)


tristanls-tbp:
Thank you @carver, I appreciate the feature idea, however this code and the `benchmark` folder are being removed soon™️ in scope of #540.

Let's wait until after #540 is merged.

ramyamounir:
@carver, you might find [this option](https://github.com/thousandbrainsproject/tbp.monty/pull/523) useful for debugging a single episode, which is usually just a single object at a single rotation in the experiment.

carver:
> @carver, you might find [this option](https://github.com/thousandbrainsproject/tbp.monty/pull/523) useful for debugging a single episode, which is usually just a single object at a single rotation in the experiment.

Interesting, thanks! I didn't find it because I was not using `run.py`, which doesn't have that flag. I would open a PR to add it to run.py also, but it sounds like now isn't a good time to tinker with the benchmark code.

carver:
This approach is obsolete after the config refactor in #540 -- If I reimplement this idea then I would start from scratch, so I'll close this PR.

PR #562, opened 2025-11-16
feat: add sensor_module_id accessor for HabitatSalienceSM

scottcanoe:
`sensor_module_id` is Python-private in `HabitatSalienceSM` (i.e., it is `HabitatSalienceSM._sensor_module_id`). But Monty experiment classes need to be able to access every SM's `sensor_module_id`. This PR just adds the property needed to get it.


tristanls (approved):
Ignore my previous comment. Looks good.

PR #563, opened 2025-11-17
refactor!: add `Observations` classes, SensorID & Modality newtypes

AgentRev:
As part of issue #506 for the Simulator protocol refactor, this PR integrates the following changes:

* implement `Observations(Dict[AgentID, AgentObservations])`
* `Simulator.reset()` returns `Observations`
* `Simulator.apply_actions()` returns `Observations`
* remove extraneous `Simulator.observations()`

Essentially, pretty much everything between https://github.com/thousandbrainsproject/prototype.simulator/commit/3fafeb1666084b60b6f94ac1650bb24e1866e977 and https://github.com/thousandbrainsproject/prototype.simulator/commit/810830cd5470fb39fd73b11d722e720ecd0c4f30, inclusively.

I took the liberty of placing `SensorID` into its own `sensors.py` file, just like you did with `AgentID` in `agents.py`.

I've also included the environment changes alongside. Let me know if you'd prefer those in a separate PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -256,7 +256,6 @@ ignore = [
     "PT009", # PT009: Use a regular `assert` instead of unittest-style `assertEqual`
     "PT018", # PT018: Assertion should be broken down into multiple parts
     "PT027", # PT027: Use `pytest.raises` instead of unittest-style `assertRaises`
-    "PTH111", # PTH111: `os.path.expanduser()` should be replaced by `Path.expanduser()`
     "PTH119", # PTH119: `os.path.basename()` should be replaced by `Path.name`
     "PTH120", # PTH120: `os.path.dirname()` should be replaced by `Path.parent`
     "PTH123", # PTH123: `open()` should be replaced by `Path.open()`
--- a/src/tbp/monty/frameworks/models/motor_policies.py
+++ b/src/tbp/monty/frameworks/models/motor_policies.py
@@ -15,8 +15,8 @@ import copy
 import json
 import logging
 import math
-import os
 from dataclasses import dataclass, field
+from pathlib import Path
 from typing import Any, Literal, Mapping, cast
 
 import numpy as np
@@ -1588,7 +1588,7 @@ def read_action_file(file: str) -> list[Action]:
     Returns:
         List of actions
     """
-    file = os.path.expanduser(file)
+    file = Path(file).expanduser()
     with open(file) as f:
         file_read = f.read()
 


src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -118,10 +125,12 @@ def step(self, actions: Sequence[Action]) -> dict:
...
 
         Returns:
-            The observation.
+            The observations.
         """
+        obs = self._observation()
+
         if not actions:
-            return self._observation()
+            return obs
```
jeremyshoemaker: What's the thinking behind changing this? We're not using `obs` below except in the return statement. The `for` loop below is guaranteed to run because of this check here, so `obs` will be set before then. I think it's cleaner to keep this as it was.

src/tbp/monty/simulators/habitat/simulator.py
```diff
@@ -547,55 +548,38 @@ def apply_actions(self, actions: Sequence[Action]) -> dict[str, dict]:
...
-                        }
-                    }
+            All observations.
         """
         agent_indices = range(len(self._agents))
         obs = self._sim.get_sensor_observations(agent_ids=agent_indices)
         return self.process_observations(obs)
 
-    def process_observations(self, obs) -> dict:
+    def process_observations(self, habitat_obs: Any) -> Observations:
```
jeremyshoemaker: Did you check to see if `habitat_obs` has a more specific type we can use here? We'd prefer not to add any `Any`s that we don't have to.

src/tbp/monty/frameworks/environments/real_robots.py
```diff
@@ -33,7 +35,7 @@ def sample(self):
         return self.action_space.sample()
 
 
-class RealRobotsEnvironment(EmbodiedEnvironment):
+class RealRobotsEnvironment:
```
jeremyshoemaker: I don't follow why this isn't an **EmbodiedEnvironment** anymore. Maybe I'm missing something obvious. I haven't looked at the **RealRobotsEnvironment** much yet.

src/tbp/monty/frameworks/environments/real_robots.py
```diff
@@ -51,29 +53,18 @@ def __init__(self, id):  # noqa: A002
...
         observation = self._env.reset()
         return dict(**observation, reward=0, done=False, info=None)
 
     def close(self) -> None:
         if self._env is not None:
             self._env.close()
             self._env = None
+
+    def __del__(self):
+        self.close()
```
jeremyshoemaker: I've always been reluctant to rely on `__del__` for resource management, since it's not always obvious when it actually runs. This sort of thing is almost always better handled with a context manager.

src/tbp/monty/frameworks/models/abstract_monty_classes.py
```diff
@@ -256,7 +290,7 @@ def build_model(self, observations):
         pass
 
     @abc.abstractmethod
-    def update_model(self, obersevations):
+    def update_model(self, observations):
```
jeremyshoemaker, line 293: 👍

jeremyshoemaker (commented):
Overall this looks good to me. I'll need to run the benchmarks on it to make sure nothing has changed, though I don't expect it to. I have some questions though.

src/tbp/monty/frameworks/environments/real_robots.py
tristanls-tbp: **note**: If I recall correctly, we broke the hierarchy here because the `Observations` coming back from a `RealRobotsEnvironment` look nothing like the `Observations` in the `EmbodiedEnvironment`. Before, it used to be a generic `dict[str, Any]`, but with the changes in this PR, it is a more specific type.

src/tbp/monty/frameworks/environments/two_d_data.py
tristanls-tbp: **issue**: Also, the pattern below in `SaccadeOnImageEnvironment.step()` is different than here. It initializes `obs = {}` on a line before the `for action in actions:`. I prefer that pattern for consistency if initialization should be explicit.

src/tbp/monty/simulators/habitat/sensors.py
```diff
@@ -59,7 +61,7 @@ def process_observations(self, sensor_obs) -> dict:
         Returns:
             The processed observations grouped by agent_id
         """
-        return sensor_obs
+        return cast("SensorObservations", sensor_obs)
```
tristanls-tbp: **question**: Seeing this here in this context, is this true? Do the raw habitat-sim observations have the requisite types like `npt.NDArray[np.int_]` and `npt.NDArray[np.float64]` in the respective modalities?

src/tbp/monty/simulators/habitat/simulator.py
```diff
@@ -547,55 +548,38 @@ def apply_actions(self, actions: Sequence[Action]) -> dict[str, dict]:
 
             action.act(self)
 
-            observations = self.observations
-
-        return observations
+        return self.observations
```
tristanls-tbp, line 552: **praise**: Nice cleanup, thank you.

tests/unit/embodied_data_test.py
```diff
@@ -57,27 +64,39 @@ def __init__(self):
...
         self._current_state += 1
-        return {
-            f"{AGENT_ID}": {
-                f"{SENSOR_ID}": {"sensor": EXPECTED_STATES[self._current_state]}
+        return Observations(
+            {
+                AgentID(AGENT_ID): AgentObservations(
+                    {
+                        SensorID(SENSOR_ID): SensorObservations(
+                            {Modality("sensor"): EXPECTED_STATES[self._current_state]}
```
tristanls-tbp: **issue**: `sensor` is not a valid modality in `SensorObservations`.

**suggestion**:  Use `Modality("raw")` here. This might require an update to how `EXPECTED_STATES` is created.

tests/unit/habitat_data_test.py
```diff
@@ -36,8 +38,8 @@
 NUM_STEPS = 10
 DEFAULT_ACTUATION_AMOUNT = 0.25
 AGENT_ID = AgentID("camera")
-SENSOR_ID = "sensor_id_0"
-SENSORS = ["depth"]
+SENSOR_ID = SensorID("sensor_id_0")
+SENSORS = [Modality("depth")]
```
tristanls-tbp: **issue**: It appears that `SENSORS` should be renamed to `MODALITIES`.

tristanls-tbp: **suggestion**: On second thought, this is only accessed via `SENSORS[0]`, so it should not be a list, but a single constant called `MODALITY`.

tests/unit/habitat_sim_test.py
```diff
@@ -92,8 +94,8 @@ def test_create_environment(self):
         # Single agent with two 5x5 cameras
         agents = create_agents(num_agents=1, resolution=(5, 5))
         # Get first agent id
-        agent_id = agents[0].agent_id
-        sensor_id = agents[0].sensor_id
+        agent_id = AgentID(agents[0].agent_id)
+        sensor_id = SensorID(agents[0].sensor_id)
```
tristanls-tbp: **issue**: This should be fixed in `SingleSensorAgent` instead.

src/tbp/monty/simulators/habitat/simulator.py
AgentRev: The exact type seems to be `dict[int, ObservationDict]` along `from habitat_sim.simulator import ObservationDict`, I'll change it

src/tbp/monty/frameworks/environments/two_d_data.py
AgentRev: My IDE had a warning about using a potentially-undefined variable, you're right that it's not supposed to happen here, but I'd rather make the warning disappear 🙂

Since these return `self._observation()` by default, I think the cleanest way would be starting with `obs = self._observation()` and removing the `if not actions` (since it will skip the `for` if empty anyway)

tristanls-tbp: **praise**: That's a nice solve.

**suggestion (non-blocking)**: If it makes sense, could you also update the pattern I referenced so that they're consistent?

src/tbp/monty/frameworks/environments/real_robots.py
AgentRev: Yeah that's what I suspected, the [real_robots](https://github.com/aicrowd/real_robots) functions return different vars of different types that make it incompatible with `EmbodiedEnvironment`, hence why Tristan did that here: https://github.com/thousandbrainsproject/prototype.simulator/commit/dc55bd107773c6eed125c5f6541342c528fec7d1

tests/unit/habitat_sim_test.py
AgentRev: Yeah, the `AgentID` here doesn't make sense anymore since that type was integrated into `SingleSensorAgent` in the main repo.

However, `SingleSensorAgent` has `sensor_id: str`, should I change it to `sensor_id: SensorID | None` instead and remove the `SensorID(...)` above too?

src/tbp/monty/frameworks/environments/real_robots.py
AgentRev: I suppose this must be a defensive fallback that @tristanls-tbp added in https://github.com/thousandbrainsproject/prototype.simulator/commit/dc55bd107773c6eed125c5f6541342c528fec7d1.

The `real_robots.py` code seems a bit old, it doesn't appear to be used anywhere and I'm not sure if the [real_robots](https://pypi.org/project/real-robots/) package even works anymore. It hasn't been updated in 5 years and I can't install it due to errors in upstream dependencies. Maybe we could just... remove it altogether and let it rest in peace? 😉

jeremyshoemaker:
Benchmarks look fine as of yesterday. It should be good after finishing up any fixes that @tristanls-tbp pointed out and dealing with the merge conflicts.

src/tbp/monty/frameworks/environments/real_robots.py
tristanls-tbp: @vkakerbeck ☝️ ?

tests/unit/habitat_sim_test.py
tristanls-tbp: **suggestion**: Yes, but let's not accommodate the `| None`. It should be `sensor_id: SensorID`. The uuid ID generator can be removed. We currently rely too heavily on explicit sensor IDs to autogenerate them.

```python
        # can be removed
        if sensor_id is None:
            sensor_id = uuid.uuid4().hex
```

src/tbp/monty/frameworks/environments/real_robots.py
vkakerbeck: yes, we can remove all real_robots code and dependency :)

src/tbp/monty/simulators/habitat/sensors.py
AgentRev: I'm not even sure if the cast is needed here, the only place where this function is called is in `simulators/habitat/agents.py` and it's already a `SensorObservations` since the parent `obs_by_sensor` is an `AgentObservations`:

image

I could just change the def to this instead:

    def process_observations(self, sensor_obs: SensorObservations) -> SensorObservations:
        return sensor_obs

The very first version of that function was simply this:

    def process_observations(self, sensor_obs):
        return sensor_obs

tristanls-tbp: Oh nice, `def process_observations(self, sensor_obs: SensorObservations) -> SensorObservations:` makes sense to me 👍 .

tests/unit/embodied_data_test.py
```diff
@@ -111,7 +139,7 @@ def reset(self):
         self._current_state = 0
         return {
             f"{AGENT_ID}": {
-                f"{SENSOR_ID}": {"sensor": EXPECTED_STATES[self._current_state]}
+                f"{SENSOR_ID}": {"raw": EXPECTED_STATES[self._current_state]}
             }
         }
```
tristanls-tbp: **issue**: I think `reset()` should be returning `Observations` and not a `{}`.

tests/unit/embodied_data_test.py
```diff
@@ -152,18 +180,20 @@ def test_embodied_env_interface_dist(self):
             obs_dist, _ = env_interface_dist.step(motor_system_dist())
             print(obs_dist)
             self.assertTrue(
-                np.all(obs_dist[AGENT_ID][SENSOR_ID]["sensor"] == EXPECTED_STATES[i])
+                np.all(obs_dist[AGENT_ID][SENSOR_ID]["raw"] == EXPECTED_STATES[i])
```
tristanls-tbp: **issue**: `Modality("raw")`

tests/unit/embodied_data_test.py
```diff
@@ -186,18 +216,21 @@ def test_embodied_env_interface_abs(self):
         for i in range(1, NUM_STEPS):
             obs_abs, _ = env_interface_abs.step(motor_system_abs())
             self.assertTrue(
-                np.all(obs_abs[AGENT_ID][SENSOR_ID]["sensor"] == EXPECTED_STATES[i])
+                np.all(obs_abs[AGENT_ID][SENSOR_ID]["raw"] == EXPECTED_STATES[i])
```
tristanls-tbp: **issue**: `Modality("raw")`

tests/unit/embodied_data_test.py
```diff
@@ -217,7 +250,7 @@ def test_embodied_env_interface_dist_states(self):
 
         for i, item in enumerate(env_interface_dist):
             self.assertTrue(
-                np.all(item[AGENT_ID][SENSOR_ID]["sensor"] == EXPECTED_STATES[i])
+                np.all(item[AGENT_ID][SENSOR_ID]["raw"] == EXPECTED_STATES[i])
```
tristanls-tbp: **issue**: `Modality("raw")`

tests/unit/embodied_data_test.py
```diff
@@ -239,7 +272,7 @@ def test_embodied_env_interface_abs_states(self):
 
         for i, item in enumerate(env_interface_abs):
             self.assertTrue(
-                np.all(item[AGENT_ID][SENSOR_ID]["sensor"] == EXPECTED_STATES[i])
+                np.all(item[AGENT_ID][SENSOR_ID]["raw"] == EXPECTED_STATES[i])
```
tristanls-tbp: **issue**: `Modality("raw")`

tristanls-tbp (commented):
Thank you for the updates.

I added additional comments where more explicit typing is needed.

AgentRev:
@tristanls-tbp Good catch, I found a couple more as well, should be fine now.

jeremyshoemaker:
I can't believe I just did that. Updated this branch, set it to auto merge, and then went and merged a separate PR. 🤦

src/tbp/monty/frameworks/models/abstract_monty_classes.py
```diff
@@ -10,8 +10,42 @@
...
+    """Observations from a sensor."""
+
+    rgba: npt.NDArray[np.int_]  # TODO: Verify specific type
+    depth: npt.NDArray[np.float64]  # TODO: Verify specific type
+    semantic: npt.NDArray[np.int_]  # TODO: Verify specific type
+    semantic_3d: npt.NDArray[np.int_]  # TODO: Verify specific type
+    sensor_frame_data: npt.NDArray[np.int_]  # TODO: Verify specific type
+    world_camera: npt.NDArray[np.int_]  # TODO: Verify specific type
+    pixel_loc: npt.NDArray[np.float64]  # TODO: Verify specific type
+    raw: npt.NDArray[np.uint8]
```
goingmaymay, line 36: Hi @AgentRev, I came across this code and wanted to share a few thoughts.

`SensorObservations` appears to be doing two things at once:
1. It defines a nominal subtype by subclassing `Dict[Modality, npt.NDArray[Any]]`.
2. It also declares member annotations (`rgba`, `depth`, etc.) that look like structural typing for dictionary entries, as if this were a `TypedDict` or dataclass.

However, those annotations apply only to class attributes and do not affect the types of the dictionary values. As a result, the code suggests modality-specific typing (for example, an int array for `rgba` and a float array for `depth`), but mypy only sees a generic `Dict[Modality, npt.NDArray[Any]]`, so the modality-specific types are effectively ignored.

I am not sure what expectations we have for `SensorObservations` and the different modalities. In particular, it is not clear whether we want:
1. support for arbitrary modalities (that is, users adding their own modalities),
2. a fixed, strongly typed set of known modalities, or
3. support for both of the above.

### If we need to support arbitrary modalities

If the goal is to support arbitrary modalities, then we should treat this as a nominal mapping type and remove the structural-looking annotations:

```python
class SensorObservations(Dict[Modality, npt.NDArray[Any]]):
    """Observations from a sensor."""
    pass
```

This keeps the meaning clear and avoids implying that `rgba`, `depth`, etc. are type-checked dictionary members.

Note: An alternative to subclassing `Dict` would be to use `NewType`, which would prevent accidental attribute additions and make the intent a bit clearer. More broadly, I am not sure nominal subtyping provides much benefit here, and a plain type alias might be a simpler fit.

### If we only need to support a fixed set of modalities

If instead the set of modalities is intended to be fixed and known, then we should represent it as a structured type (for example, a dataclass or `TypedDict`) and drop `Modality`, since the fields themselves define the modalities:

```python
@dataclass
class SensorObservations:
    rgba: npt.NDArray[np.int_]
    depth: npt.NDArray[np.float64]
    semantic: npt.NDArray[np.int_]
    semantic_3d: npt.NDArray[np.int_]
    sensor_frame_data: npt.NDArray[np.int_]
    world_camera: npt.NDArray[np.int_]
    pixel_loc: npt.NDArray[np.float64]
    raw: npt.NDArray[np.uint8]
```

This makes the structure explicit and allows the value types to be meaningfully checked.

### If we want to support both

If we want to support both a fixed, strongly typed set of known modalities and the ability to introduce additional modalities over time, then we will need to look at other options. Supporting both would likely require a different design, such as representing each modality as an explicit `SensorObservation` object (with type parameters) rather than storing everything directly in a raw dictionary. That could allow canonical modalities to remain strongly typed, while still leaving room for new modalities when needed.

Even if extensibility is a long-term goal, we may still want to begin with a finite, strongly typed set of modalities for now.

---

Thanks for reading. I would be interested in your thoughts on which of these directions best matches our intended usage.

AgentRev: Good catch, I think the intent here was most likely a fixed set of modalities. The author's actually @tristanls-tbp, so I'll let him confirm: https://github.com/thousandbrainsproject/prototype.simulator/blame/89ff24322dbf36c112e30fe91d572ab37e5c662d/src/tbp/monty/frameworks/models/abstract_monty_classes.py#L20-L30

tristanls-tbp: @AgentRev got the correct intuition, and I agree with your recommendation @mf-maymay. We want a fixed set of modalities.

My intent here is to signal that there are specific modalities for which we have established handling mechanisms, while providing a future-proof escape valve in the form of `raw`, which anyone can use to add their own modalities on an ad hoc basis.

I am not attached to the Python version of the `Modality` enum I introduced. I was playing with an alternate data organization, doing some Rust code with `Vec`s, taking advantage of Rust's enums:
```rust
pub enum Modality {
  Rgb(Vec<u8>),
  Raw(Vec<u8>)
}
pub type Observation = Vec<Modality>;
pub type Observations = HashMap<SensorId, Observation>;
```
But that doesn't translate well to here, as you pointed out, and I didn't notice.

**note**: The `# TODO: Verify specific type` comments are important. I am _not sure_ if I picked the correct ones. I did not look through the code to verify.

goingmaymay: Thank you both. I will work to redefine `SensorObservations` as a dataclass, and I'll update the modality-specific types while I'm at it. I'll drop the `Modality` type and instead rely on the dataclass fields to specify supported modalities.

Minor note: `NewType` does not actually define an enum, since its use is not limited to a fixed set of values. If we wanted a type with a fixed set of values, we'd want to use the `enum` module or `Literal["rgba", "depth", ...]`. (`Literal` technically doesn't define an enum, but it is more ergonomic and invites less human error, in my experience.)

---

Here's the PR I was working on when I noticed this situation: https://github.com/thousandbrainsproject/tbp.monty/pull/662

tristanls-tbp: > Minor note: `NewType`does not actually define an enum, since its use is not limited to a fixed set of values. If we wanted a type with a fixed set of values, we'd want to use the `enum` module or `Literal["rgba", "depth", ...]`. (`Literal` technically doesn't define an enum, but it is more ergonomic and invites less human error, in my experience.)

Yeah, fair point 🙂.

I imagine you might find a few more "half-solutions" like that. I feel like there were a few places where I wanted to keep the shape of things mostly the same during refactoring, but provide some guidance/context/breadcrumbs for future development.

PR #564, opened 2025-11-18
fix: move rmtree code outside of context manager in run_parallel cleanup

ramyamounir:
When running pretraining experiments on AWS, they fail with this error below:

pic-selected-251117-1949-16

After investigating, it looks like it's refusing to delete a directory for configs that is still being accessed by the context manager. I moved this logic outside to fix the problem. Strangely, this error only shows on AWS not locally :man_shrugging:


tristanls-tbp:
~~**question**: I thought pre-training could not be run with `run_parallel` ? Am I misremembering? @vkakerbeck ?~~

I was misremembering.

vkakerbeck:
yes, pretraining episodes are actually very well suited for parallelization since every episode takes the same number of steps and they are not dependent on each other (if supervised). We actually ran the 77 object pretraining on 80 CPUs.

PR #567, opened 2025-11-19
refactor!: follow ruff rule PTH118 (take 2)

carver:
A reattempt at #545, which failed at merge. It was reverted by #566.

### DIFF
--- a/src/tbp/monty/frameworks/experiments/monty_experiment.py
+++ b/src/tbp/monty/frameworks/experiments/monty_experiment.py
@@ -89,8 +89,6 @@ class MontyExperiment:
         if self.show_sensor_output:
             self.live_plotter = LivePlotter()
 
-        logger.info(self.config)
-
     def setup_experiment(self, config: dict[str, Any]) -> None:
         """Set up the basic elements of a Monty experiment and initialize counters.
 
@@ -98,6 +96,7 @@ class MontyExperiment:
             config: config specifying variables of the experiment.
         """
         self.init_loggers(self.config["logging"])
+        logger.info(self.config)
         self.model = self.init_model(
             monty_config=config["monty_config"],
             model_path=self.model_path,
@@ -329,6 +328,9 @@ class MontyExperiment:
 
         if not os.path.exists(self.output_dir):
             os.makedirs(self.output_dir)
+
+        # Treat as root logger
+        logger.propagate = False
 
         # Clear any existing tpb.monty logger handlers
         for handler in logger.handlers:
--- a/tools/github_readme_sync/file.py
+++ b/tools/github_readme_sync/file.py
@@ -49,7 +49,7 @@ def find_markdown_files(
     ignore_files = DEFAULT_IGNORE_FILES if ignore_files is None else ignore_files
 
     md_files = []
-    for root, _, files in os.walk(folder):
+    for root, _, files in os.walk(Path(folder).resolve()):
         path_parts = Path(root).parts
         if any(part.startswith(".") for part in path_parts):
             continue
--- a/tools/github_readme_sync/index.py
+++ b/tools/github_readme_sync/index.py
@@ -129,7 +129,7 @@ def process_markdown_files(docs_dir: str) -> list[dict]:
     if _is_empty(docs_dir):
         raise ValueError("docs_dir cannot be empty")
 
-    docs_path = Path(docs_dir)
+    docs_path = Path(docs_dir).resolve()
     if not docs_path.exists():
         raise ValueError(f"Directory {docs_dir} does not exist")
 
--- a/tools/github_readme_sync/tests/file_test.py
+++ b/tools/github_readme_sync/tests/file_test.py
@@ -89,6 +89,30 @@ class TestFindMarkdownFiles(unittest.TestCase):
 
         self.assertEqual(len(result), 2)
 
+    def test_find_markdown_files_with_relative_path_containing_dotdot(self):
+        docs_dir = Path(self.temp_dir) / "docs"
+        docs_dir.mkdir()
+        (docs_dir / "readme.md").write_text("# Readme")
+        (docs_dir / "guide.md").write_text("# Guide")
+        (docs_dir / "subdir").mkdir()
+        (docs_dir / "subdir" / "nested.md").write_text("# Nested")
+
+        subdir = Path(self.temp_dir) / "subdir"
+        subdir.mkdir()
+
+        original_cwd = Path.cwd()
+        try:
+            os.chdir(subdir)
+            relative_path = "../docs"
+
+            result = find_markdown_files(relative_path)
+
+            result_basenames = sorted([Path(path).name for path in result])
+            expected_basenames = sorted(["readme.md", "guide.md", "nested.md"])
+            self.assertEqual(result_basenames, expected_basenames)
+        finally:
+            os.chdir(original_cwd)
+
 
 if __name__ == "__main__":
     unittest.main()


carver:
Hm, `load_new_character_data()` in `two_d_data.py` looks bugged. Let me see if I can run a local test that confirms that, before I fix it.

carver:
Oh, I can't find any predefined benchmark configs for the omniglot dataset or any data that seems to be 2D. Headed to the tutorials to figure out how it's supposed to be run...

carver:
Yup:
```
File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/environments/two_d_data.py", line 231, in load_new_character_data
    current_image = load_img(img_char_dir + char_dir + ".png")
TypeError: unsupported operand type(s) for +: 'PosixPath' and 'str'
```

carver:
The omniglot benchmark  runs cleanly for me now. I'm going to do another once-over and see if I spot anything else.

carver:
Ok, I did another look through the whole thing, especially looking for downstream usages of the path object for usages like `Path() + str()`. I didn't find any more after the omniglot one.

While I was reviewing it again, I did throw in some more type annotations and doc updates. I noticed several types that were now wrong (ie~ saying that an argument could only be a `str`, when it's possible that it could be a `Path`). I also updated the code examples in the docs to use `Path` slash separators.

I kept these ^ changes as separate commits in case you'd rather have me open a new PR for them.

Once CI turns green, it's ready for a review.

src/tbp/monty/frameworks/config_utils/make_env_interface_configs.py
```diff
@@ -58,7 +58,7 @@ class ExperimentArgs:
     # e.g. being off the object)
     n_train_epochs: int = 3
     n_eval_epochs: int = 3
-    model_name_or_path: str = ""
+    model_name_or_path: str | Path = ""
```
jeremyshoemaker: For flexibility and [per our typing guidelines](https://thousandbrainsproject.readme.io/docs/typing-guide#methods-and-functions-should-accept-the-broadest-possible-type), we should probably use `os.PathLike` here, instead of Path. Unfortunately, `str` isn't a `PathLike` so the union type is still needed.

src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -213,24 +213,25 @@ def reset(self):
...
+            / f"character{self.character_id:02d}"
         )
-        first_img_char_child = next(Path(img_char_dir).iterdir()).name
+        first_img_char_child = next(img_char_dir.iterdir()).name
         char_img_names = first_img_char_child.split("_")[0]
-        char_dir = "/" + char_img_names + "_" + str(self.character_version).zfill(2)
-        current_image = load_img(img_char_dir + char_dir + ".png")
-        move_path = load_motor(stroke_char_dir + char_dir + ".txt")
-        logger.info(f"Finished loading new image from {img_char_dir + char_dir}")
+        char_stem = f"{char_img_names}_{self.character_version:02d}"
```
jeremyshoemaker, line 202: Had I not just read the Python `pathlib` documentation, I would have asked why "stem", but that's the term they use. I don't think I've really seen it elsewhere, but it's nice and short.

src/tbp/monty/simulators/habitat/environment.py
```diff
@@ -93,7 +96,7 @@ def __init__(
         objects: list[dict | ObjectConfig] | None = None,
         scene_id: str | None = None,
         seed: int = 42,
-        data_path: str | None = None,
+        data_path: str | Path | None = None,
```
jeremyshoemaker, line 99: `Path` -> `PathLike`

src/tbp/monty/simulators/habitat/simulator.py
```diff
@@ -107,7 +107,7 @@ class HabitatSim(HabitatActuator):
     def __init__(
         self,
         agents: list[HabitatAgent],
-        data_path: str | None = None,
+        data_path: str | Path | None = None,
```
jeremyshoemaker, line 110: `Path` -> `PathLike`

src/tbp/monty/frameworks/config_utils/make_env_interface_configs.py
jeremyshoemaker: Actually, this seems like a weird edge case. I think since it's a field on a dataclass we want to be specific, but it's also a constructor argument, which we'd prefer to be more flexible with. 🤷

carver: If I read it correctly, `PathLike` is a *very* loose type, that only implements `.__fspath__()`. I imagine that the goal for the repo is to switch to using `Path` objects everywhere instead of `str`. Then you can use things like `.open()` without converting it first. But a `PathLike` argument would still require conversion to `Path` before using any methods like `.open()` or properties like `.parent`.

Do you read it the same way? And if so, do you still think I should set these to `PathLike`?

carver: My intuition says `PathLike` is not the ideal choice here, but in the interest of getting this merged, I switched it. It's clearly better than just a `str` type, which would now be an _incorrect_ type.

carver:
Hm, that's a new one:
```
Collecting package metadata (repodata.json): - done
/home/runner/work/_temp/8811a1c2-169d-46ba-909e-f621a74ac467.sh: line 3:  4250 Killed                  conda env create
Solving environment: / 
Error: The operation was canceled.
```

carver:
Maybe conda was just grumpy for a bit? I guess it's good to review again.

src/tbp/monty/frameworks/config_utils/make_env_interface_configs.py
jeremyshoemaker: Yeah, I think you're right. And this is a concrete field so a more concrete type makes sense. Sorry about the back and forth.

jeremyshoemaker:
> Hm, that's a new one:
> 
> ```
> Collecting package metadata (repodata.json): - done
> /home/runner/work/_temp/8811a1c2-169d-46ba-909e-f621a74ac467.sh: line 3:  4250 Killed                  conda env create
> Solving environment: / 
> Error: The operation was canceled.
> ```

GitHub seems to still be having some issues, with Actions in particular.

src/tbp/monty/frameworks/config_utils/make_env_interface_configs.py
carver: All good! It's reverted, and tests are passing.

jeremyshoemaker:
We probably should have merged this before we merged our big config refactor in #540 so you wouldn't have to deal with these conflicts, but here we are. 

It should be pretty straightforward though. A lot of it is files that have been deleted. Let me know if you run into anything complicated.

carver:
Doh, hah, ok. I'll take a look.

carver:
I want to do a local test of omniglot and a speedrun of one more benchmark before I call it ready to go. I can't do them quickly because my local branches to prepare them don't apply cleanly anymore. Not sure when I'll get a chance to fix those up.

ETA: Oh right, both of those changes are almost entirely configuration changes, which is now very different. Seems very likely I won't get to it during the weekend, then.

carver:
Omniglot is working, but still a path issue on `base_10multi_distinctobj_dist_agent`.

tristanls-tbp:
@carver, I think maybe your path issue might be fixed with #584, now merged?

carver:
> @carver, I think maybe your path issue might be fixed with #584, now merged?

Yup, everything looks good to me now. 

Separately, I was able to do the workaround for the install-monty CI bug. Luckily the fix is sticky: I can revert it and still pass CI again, because the installation gets cached.

carver:
This failure looks unrelated to me:
> Failed to get https://dash.readme.com/api/v1/docs/style-guide <!DOCTYPE html><html lang="en"><head><title>We're having issues!</title>
> ... ✂️  ...
>  Traceback (most recent call last):
>  File  "/home/runner/miniconda/envs/tbp.monty/lib/python3.8/runpy.py", line 194, in _run_module_as_main
>     return _run_code(code, main_globals, None,
>   File "/home/runner/miniconda/envs/tbp.monty/lib/python3.8/runpy.py", line 87, in _run_code
>     exec(code, run_globals)
>   File "/home/runner/work/tbp.monty/tbp.monty/tbp.monty/tools/github_readme_sync/cli.py", line 159, in <module>
>     main()
>   File "/home/runner/work/tbp.monty/tbp.monty/tbp.monty/tools/github_readme_sync/cli.py", line 115, in main
>     upload(hierarchy, args.folder, rdme=ReadMe(args.version))
>   File "/home/runner/work/tbp.monty/tbp.monty/tbp.monty/tools/github_readme_sync/upload.py", line 37, in upload
>     process_children(
>   File "/home/runner/work/tbp.monty/tbp.monty/tbp.monty/tools/github_readme_sync/upload.py", line 69, in process_children
>     doc_id, created = rdme.create_or_update_doc(
>   File "/home/runner/work/tbp.monty/tbp.monty/tbp.monty/tools/github_readme_sync/readme.py", line 301, in create_or_update_doc
>     raise ValueError(f"Failed to create doc {doc['title']}")
>  ValueError: Failed to create doc Style Guide

_Edit: Yup, a re-run with an empty commit succeeds the same job._

carver:
> It should be pretty straightforward though. A lot of it is files that have been deleted. Let me know if you run into anything complicated.

@jeremyshoemaker there was a bit of rework unfortunately, I think from some logic that got moved around. But it's all set now. I am hoping to get it merged before some other merge creates more conflicts. :)

PR #571, opened 2025-11-19
build: recipe/meta.yaml reflecting latest; added note to remember

tristanls-tbp:
`recipe/meta.yaml` should reflect what is in `environment.yml` and `environment_arm64.yml`.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -21,6 +21,7 @@ channels:
   - conda-forge
   - defaults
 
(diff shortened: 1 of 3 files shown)


environment.yml
```diff
@@ -21,6 +21,7 @@ channels:
   - conda-forge
   - defaults
 
+# Note: Remember to update recipe/meta.yaml when updating this file.
```
jeremyshoemaker, line 24: 👍

PR #572, opened 2025-11-19
chore: add opencv-python pip dependency to Monty

tristanls-tbp:


hlee9212:
Hi @tristanls-tbp (and @scottcanoe), I will need `opencv-python` as dependency for my upcoming implementation project for 2D SM. Can this PR be opened?

tristanls-tbp:
@hlee9212, if you need a dependency as part of your implementation, then please include the dependency as part of that PR.

tristanls-tbp:
@hlee9212, I was wondering why this is a separate pull request. I think, in the past, the reason behind adding a dependency separately was to create infrastructure images so we could run benchmarks.

I think that since then, we have been building infrastructure images from a specific pull request (which was not available when this PR was opened).

tristanls-tbp:
So.. the plot thickens... this draft pull request is here specifically so that we can build a benchmark image in our infrastructure 🙂 (I looked, and the image is called `MontyCPUWorkstationPR572...`).

@hlee9212, when you open your PR, we'll be able to build an image specifically for your PR.

PR #578, opened 2025-11-21
refactor!: remove policy_setup_utils.py

jeremyshoemaker:
Since we moved the configs into Hydra, we have a few places where fragments of configurations are used in tests. This removes the policy_setup_utils module as it is no longer needed once we start loading those fragments using Hydra.


tristanls-tbp (approved):
Nice, thank you!

PR #581, opened 2025-11-23
feat: print yaml-formatted configs in `run.py` and `run_parallel.py`

scottcanoe:
`run.py` and `run_parallel.py` print experiment configs, but they're not very readable. This PR suggests we use `omegaconf` to print prettier.

In pictures, we'll go from looking at
Screenshot 2025-11-23 at 11 39 05 AM
to
Screenshot 2025-11-23 at 11 38 50 AM

**Note:** I tried to perform node-interpolation only so that a field like
```yaml
   min_eval_steps: ${benchmarks.min_eval_steps}
```
would optionally print out as
```yaml
   min_eval_steps: 20
```
I couldn't figure it out though. Ideally, I could print out the full config with those values dropped in for future reference. But that's a side quest for another time.

### DIFF
--- a/.github/workflows/monty.yml
+++ b/.github/workflows/monty.yml
@@ -145,7 +145,7 @@ jobs:
         run: |
           export PATH="$HOME/miniconda/bin:$PATH"
           source activate tbp.monty
-          deptry benchmarks src tests
+          deptry src tests conf
 
   check_license_monty:
     name: check-license-monty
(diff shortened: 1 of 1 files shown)


tristanls-tbp (commented):
None of the notes below are intended to block this pull request. I think this pull request is fine. Having said that...

**note**: Seeing `print_config` duplicated, it should probably be extracted into `src/tbp/monty/hydra.py`

**note**: It really bugs me that we are putting human-readable logs where machine logs should be. We have a `print_config` flag that will print the configuration and exit. It really feels like we should have a `HumanReadableHandler` that just pretty prints all the things and use that instead of polluting machine logs.

tristanls-tbp:
**note**: You can find pretty printed YAML configurations in the `outputs` directory for your run. 
Screenshot 2025-12-05 at 16 17 45

PR #579, opened 2025-11-21
docs: running your first experiment tutorial update

tristanls-tbp:
This pull request updates the "Running Your First Experiment" tutorial to reference Hydra configuration.

PR #584, opened 2025-11-24
fix: once again, append run_name to output_dir to get output_dir/run_name

tristanls-tbp:
While reviewing #579 , [@scottcanoe identified that `run.py` no longer appends the `run_name` to `output_dir` as before](https://github.com/thousandbrainsproject/tbp.monty/pull/579#pullrequestreview-3497907903). This pull request corrects that bug.

### DIFF
--- a/src/tbp/monty/frameworks/run.py
+++ b/src/tbp/monty/frameworks/run.py
@@ -10,9 +10,9 @@ from __future__ import annotations
 
 import logging
 import os
-from pathlib import Path
 import pprint
 import time
+from pathlib import Path
 
 import hydra
 from omegaconf import DictConfig


scottcanoe (approved):
LGTM. Works as expected for `tutorial/first_experiment`over at [579](https://github.com/thousandbrainsproject/tbp.monty/pull/579).

**thought:** I'm not sure what your plans are for saving flat configs to experiment output directories, but it would be problematic if the flat config is saved _after_ the concatenation.

tristanls-tbp:
> **thought**: I'm not sure what your plans are for saving flat configs to experiment output directories, but it would be problematic if the flat config is saved after the concatenation.

@scottcanoe, yes, good point. I don't like that we modify the config. Initially, I implemented this change within `MontyExperiment`, but that required a significant amount of work in determining all the directory naming conventions in `run_parallel.py`. Therefore, I opted to follow the approach taken by the previous `run.py` for now.

PR #594, opened 2025-12-02
docs: update pretraining-a-model tutorial for Hydra configs

tristanls-tbp:


scottcanoe (approved):
LGTM. Text is good, and I confirmed that the experiment and analysis script run as expected.

PR #598, opened 2025-12-03
refactor: enable ruff PTH107 for Path.unlink()

carver:
Use `Path.unlink()` instead of `os.remove(path_str)`. When it doesn't affect functionality, use `.unlink(missing_ok=True)` to avoid an `os.path.exists(path_str)` check first.


jeremyshoemaker (approved):
Looks good to me.

PR #599, opened 2025-12-03
refactor: enable ruff rule PTH103, use Path.mkdir()

carver:
`os.makedirs()` creates all folders in the path, which is the equivalent of `Path.mkdir(parents=True)`. So `parents=True` is used in all replacements. `exist_ok=True` is used sometimes, instead of doing a file `exists()` check first.


jeremyshoemaker (approved):
Looks good to me. I kinda wish they had made `Path.mkdir` default to `parent=True`, but I suspect they're just following how the `mkdir` command does it.

PR #600, opened 2025-12-03
docs: multi-lm tutorial update for Hydra

tristanls-tbp:


scottcanoe (approved):
LGTM. Everything runs as expected.

**note:** ‏The screenshot of the eval stats table is outdated. I don't plan on replacing these regularly, but rotation error is reported in degrees now so that's a fairly significant change. I'll replace this image soon, hopefully after we stop reporting `stepwise_performance` for experiments that don't use the semantic sensor so we can get rid of that too.

tristanls-tbp:
@scottcanoe this is what I got on my machine
Screenshot 2025-12-04 at 13 01 46

PR #603, opened 2025-12-04
fix: crash in MontyRequestHandler.do_PUT

carver:
Fixes a bug introduced by #544 

While opening an output file, the server constructs the path using `data_path + "/" + new_filename`. But `data_path` was recently converted to a pathlib.Path object. So concatenating with a string will fail.

The code also includes a new test to catch regressions in the server. It's not 100% comprehensive, but covers a lot more than before.

---

_Stale commentary:_
 
I just spotted the bug by scanning the code. The only instructions for running the server involves running iOS, which I don't have. So I was not able to verify the original bug, or the fix, yet. I wanted to get the fix up for review before diving deeper into the reproduction and writing a test case.

It's not clear to me how severe it is to leave this bug active for a few days, so I wanted to start the conversation as soon as possible. If this gets merged as-is, I'll still do the follow-up work to build a reproduction test case.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -257,7 +257,6 @@ ignore = [
     "PT018", # PT018: Assertion should be broken down into multiple parts
     "PT027", # PT027: Use `pytest.raises` instead of unittest-style `assertRaises`
     "PTH119", # PTH119: `os.path.basename()` should be replaced by `Path.name`
-    "PTH120", # PTH120: `os.path.dirname()` should be replaced by `Path.parent`
     "PTH123", # PTH123: `open()` should be replaced by `Path.open()`
     "RET503", # RET503: Missing explicit `return` at the end of function able to return non-`None` value
     "RUF005", # RUF005: Consider {expression} instead of concatenation
--- a/src/tbp/monty/frameworks/run_parallel.py
+++ b/src/tbp/monty/frameworks/run_parallel.py
@@ -562,8 +562,7 @@ def post_parallel_train(experiments: list[Mapping], base_dir: str) -> None:
 
     with exp:
         exp.model.load_state_dict_from_parallel(parallel_dirs, True)
-        output_dir = os.path.dirname(experiments[0]["config"]["logging"]["output_dir"])
-        output_dir = Path(output_dir)
+        output_dir = Path(experiments[0]["config"]["logging"]["output_dir"]).parent
         if isinstance(exp, MontySupervisedObjectPretrainingExperiment):
             output_dir = output_dir / "pretrained"
         output_dir.mkdir(exist_ok=True, parents=True)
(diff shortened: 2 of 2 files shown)


vkakerbeck:
Thanks for spotting that @carver !

Nothing I can think of that you need to watch out for. There are some instructions for running monty meets world in this tutorial https://thousandbrainsproject.readme.io/docs/using-monty-for-robotics#example-1-monty-meets-world-ipad-app
Unfortunately it will be a bit hard to test that way, as like you said, that line just get's invoked when data comes in from the MontyMeetsWorld iPad app. Let me know if you want me to set it up again and test (May be good to do anyways after @tristanls-tbp updates the config to Hydra)

carver:
> Thanks for spotting that @carver !

Sure! I would love to celebrate more, but I also introduced the bug 🤦🏻‍♂️, so that kind of cancels it out 😆 

I'll take a look at writing a test so that manual validation isn't required.

carver:
I added a test that verifies server functionality. Having more test coverage is celebration-worthy. :)

As expected, when running the test without the bugfix, the test shows the failure now:
```
Exception occurred during processing of request from ('127.0.0.1', 47278)
  File "/home/carver/.local/share/uv/python/cpython-3.9.22-linux-x86_64-gnu/lib/python3.9/socketserver.py", line 316, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/home/carver/.local/share/uv/python/cpython-3.9.22-linux-x86_64-gnu/lib/python3.9/socketserver.py", line 347, in process_request
    self.finish_request(request, client_address)
  File "/home/carver/.local/share/uv/python/cpython-3.9.22-linux-x86_64-gnu/lib/python3.9/socketserver.py", line 360, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/home/carver/.local/share/uv/python/cpython-3.9.22-linux-x86_64-gnu/lib/python3.9/http/server.py", line 668, in __init__
  File "/home/carver/.local/share/uv/python/cpython-3.9.22-linux-x86_64-gnu/lib/python3.9/socketserver.py", line 747, in __init__
    self.handle()
  File "/home/carver/.local/share/uv/python/cpython-3.9.22-linux-x86_64-gnu/lib/python3.9/http/server.py", line 433, in handle
    self.handle_one_request()
  File "/home/carver/.local/share/uv/python/cpython-3.9.22-linux-x86_64-gnu/lib/python3.9/http/server.py", line 421, in handle_one_request
    method()
  File "/home/carver/code/tbp.monty/src/tbp/monty/frameworks/environment_utils/server.py", line 48, in do_PUT
    with open(data_path + "/" + new_filename, "wb") as output_file:
TypeError: unsupported operand type(s) for +: 'PosixPath' and 'str'
```

With the fix, the test passes. This hardly qualifies as a unit test, but I wasn't sure where else to put it.

I think this is ready for review/merge once CI turns green, if you're good with it.

tristanls-tbp:
> This hardly qualifies as a unit test, but I wasn't sure where else to put it.

**note**: Yeah, don't worry about it, `tests/unit/...` is fine until we reorganize the tests 🙂.

PR #604, opened 2025-12-04
docs: custom application tutorial update for Hydra

tristanls-tbp:
FYI @vkakerbeck


tristanls-tbp:
Thanks @scottcanoe, yeah it's doing its thing (running and evaluating).

Interestingly, the images are presented upside down.

Screenshot 2025-12-11 at 18 07 41

vkakerbeck:
yeah I think the upside down bit was something about how the iPad took and saved the images (we might have fixed it for later datasets) but it shouldn't impact Monty's ability to detect the object :)

PR #605, opened 2025-12-05
chore: UI improvements to the future work widget

codeallthethingz:
This PR makes some UI improvements to the future work widget as it was getting a bit cramped.  
- RFC column is hidden.
- Avatars are now clickable to go to the github profile page.
- Status is now clickable to filter by that status.
- New checkbox to hide completed items by default.

From this:
image

to this:
image


codeallthethingz:
> Nice updates, that looks better :) A couple of minor thoughts (that could be done in a different PR later on)
> 
> * I'm not sure how useful it is to link to the person's github profile. Intuitively when I click on someone's icon I would expect it to filter it by items this person is working on. Would that be hard to change to?
> * When I now click on the github icon, will it open in a new tab?
> * I worry that filtering by status will have effects we don't want. People will filter by "open" and then not see the "paused" or "evergreen" items anymore that also welcome contributors (or even "in-progress" items). I think the "hide completed" checkbox is all we need. But maybe I am missing a usecase where we would want to filter more specifically?

@vkakerbeck I made the github avatar filter instead of going to the profile page.  and the status is no longer clickable.

image

vkakerbeck:
great, thank you!

PR #614, opened 2025-12-05
test: configuration snapshot tests

tristanls-tbp:

As we still have numerous configuration improvements to make, I wanted to assert some control and stability over the configurations so that all our changes are deliberate and we identify mistakes quickly without having to run any benchmarks.

This is what a test failure looks like:

Screenshot 2025-12-09 at 09 27 46

From the `test/conf/README.md`:

# Snapshot testing

The purpose of snapshot tests is to ensure that any changes to configurations are deliberate.

## Why?

Because the configurations use inheritance and are assembled from many smaller configurations files, a small change in one of the configuration files can have cascading effects across all configurations. This may not be immediately apparent from the small diff in the one configuration file that changed.

By failing a snapshot test every time a configuration is changed, we ensure that the overall cascading effects are deliberate.

## Updating snapshots

Once you observe failing tests and decide that the final configuration changes are as intended, follow this two-step checklist:

1. Run `python conf/update_snapshots.py`. This automatically updates all of the snapshots to reflect the current configurations. After this, your tests will once again pass, since you are comparing generated configs to the ones you just generated.

2. Commit the `tests/conf/snapshots` changes to source control to confirm that all changes are intended.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -261,14 +261,12 @@ ignore = [
     "S101", # S101: Use of `assert` detected
     "S104", # S104: Possible binding to all interfaces
     "S301", # S301: `pickle` and modules that wrap it can be unsafe when used to deserialize untrusted data, possible security issue
-    "S311", # S311: Standard pseudo-random generators are not suitable for cryptographic purposes
     "S605", # S605: Starting a process with a shell, possible injection detected
     "SIM102", # SIM102: Use a single `if` statement instead of nested `if` statements
     "SIM108", # SIM108: Use ternary operator {contents} instead of if-else-block
     "SIM117", # SIM117: Use a single `with` statement with multiple contexts instead of nested `with` statements
     "SIM118", # SIM118: Use `key in dict` instead of `key in dict.keys()`
     "SLF001", # SLF001: Private member accessed: {access}
-    "SLOT001", # SLOT001: Subclasses of `tuple` should define `__slots__`
     "TC001", # TC001: Move application import {} into a type-checking block
     "TC002", # TC002: Move third-party import {} into a type-checking block
     "TD001", # TD001: Invalid TODO tag: `FIXME`
--- a/tools/future_work_widget/run_local.py
+++ b/tools/future_work_widget/run_local.py
@@ -36,7 +36,7 @@ def main():
     with tempfile.NamedTemporaryFile(mode="w", suffix=".json", delete=False) as tmp:
         index_file = Path(tmp.name)
 
-    generate_index(str(docs_dir), str(index_file))
+    generate_index(str(docs_dir), index_file)
     result = build(index_file, output_dir, docs_snippets_dir)
     index_file.unlink()
     logger.info(result.model_dump_json(exclude_none=True, indent=2))
--- a/tools/github_readme_sync/cli.py
+++ b/tools/github_readme_sync/cli.py
@@ -124,7 +124,11 @@ def main():
         rdme.delete_version()
 
     elif args.command == "generate-index":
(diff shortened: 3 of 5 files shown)


tests/conf/conf_test.py
```diff
@@ -0,0 +1,72 @@
...
+            )
+            # force resolving the config for any parsing errors
+            OmegaConf.to_object(config)
+            current_config_yaml = OmegaConf.to_yaml(config)
+            try:
+                snapshot_config_yaml = snapshot_path.read_text()
+            except FileNotFoundError:
+                snapshot_config_yaml = None
+            if snapshot_config_yaml is not None:
+                assert snapshot_config_yaml == current_config_yaml
```
vkakerbeck: It would be good to throw a useful error message here to point the user to the readme on how to update the snapshots

tests/conf/conf_test.py
```diff
@@ -0,0 +1,72 @@
...
```
vkakerbeck: Same here, an error message would be useful since it requires the used to take specific steps (check if changes are deliberate & update snapshots)

tests/conf/conf_test.py
jeremyshoemaker: The `assert` statement can take a second argument that will allow customizing the message it prints.

E.g.:
```suggestion
                assert snapshot_config_yaml == current_config_yaml, "<failure message to display>"
```

tristanls-tbp:
> We might want to include the test configs since some of those inherit from the main configs and would be affected by any changes there.

@jeremyshoemaker, after some thought, I don't think this is necessary. The tests test themselves. For instance, if a main configuration affects an existing test, the existing test will fail on its own, without requiring any additional configuration testing.

tristanls-tbp:
@vkakerbeck, @jeremyshoemaker, I updated the AssertionError to include a reference to the README while retaining the details of which specific lines changed.

Screenshot 2025-12-09 at 09 27 46

jeremyshoemaker:
> @jeremyshoemaker, after some thought, I don't think this is necessary. The tests test themselves. For instance, if a main configuration affects an existing test, the existing test will fail on its own, without requiring any additional configuration testing.

It's possible the configs could subtly change without breaking the tests, though in that case I'm not sure if it's actually a problem or not. I'd feel better if we checked them too, but if you don't think it's a problem, then I'm fine with that.

jeremyshoemaker (approved):

tristanls-tbp:
>It's possible the configs could subtly change without breaking the tests, though in that case I'm not sure if it's actually a problem or not.

In my mind, the primary use case for benchmark and tutorial configs is that we only run them occasionally, so the config tests that run in every PR provide a faster feedback loop. Since tests also run every PR, if they break, the feedback loop is already fast. However, I'm unsure how to conceptualize the scenario you highlighted, where config changes occur, but the test doesn't fail. What comes to mind is that it might be a flawed test, but testing for flawed tests feels like another layer of testing that, to me, seems excessive at this time.

jeremyshoemaker:
> However, I'm unsure how to conceptualize the scenario you highlighted, where config changes occur, but the test doesn't fail.

It's pretty straightforward. The config could be over-specifying things that the test doesn't care about.

tristanls-tbp:
>It's pretty straightforward. The config could be over-specifying things that the test doesn't care about.

In that case, I would say that making a test fail due to something the test doesn't care about would be unnecessary noise.

jeremyshoemaker:
> In that case, I would say that making a test fail due to something the test doesn't care about would be unnecessary noise.

Yeah, the more I think about it, the more I agree with you.

PR #615, opened 2025-12-06
ci: no more code in benchmarks, check conf now

tristanls-tbp:
This pull request updates our CI to check the new code in the `conf` folder. The old `benchmarks` folder has no code and is removed.

### DIFF
--- a/.rsyncignore
+++ b/.rsyncignore
@@ -8,5 +8,6 @@
 - .venv
 - .vscode
 - docs
+- outputs
 - rfcs
 - *.pyc


.github/workflows/monty.yml
```diff
@@ -145,7 +145,7 @@ jobs:
         run: |
           export PATH="$HOME/miniconda/bin:$PATH"
           source activate tbp.monty
-          deptry benchmarks src tests
+          deptry src tests conf
```
jeremyshoemaker, line 148: This isn't critical, but the original version was sorted alphabetically and this one isn't.

tristanls-tbp: The original reason was that `deptry` errored if the `benchmarks` folder wasn't first. I forgot the details of why that was the case.

PR #618, opened 2025-12-09
refactor: enable ruff rule PTH123, for Path.open()

carver:
Prefer `Path(pathstr).open()` over `open(pathstr)`

This wraps up the last `Path`-focused ruff rule!

### DIFF
--- a/tools/github_readme_sync/index.py
+++ b/tools/github_readme_sync/index.py
@@ -101,14 +101,16 @@ def generate_index(docs_dir: str, output_file_path: str) -> str:
 
     entries = process_markdown_files(docs_dir)
 
-    Path(output_file_path).parent.mkdir(exist_ok=True, parents=True)
-    with open(output_file_path, "w", encoding="utf-8") as f:
+    output_file_pathstr, output_file_path = output_file_path, Path(output_file_path)
+
+    output_file_path.parent.mkdir(exist_ok=True, parents=True)
+    with output_file_path.open("w", encoding="utf-8") as f:
         json.dump(entries, f, indent=2, ensure_ascii=False)
 
     logger.info(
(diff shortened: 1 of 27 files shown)


tools/github_readme_sync/index.py
```diff
@@ -101,14 +101,16 @@ def generate_index(docs_dir: str, output_file_path: str) -> str:
...
+
+    output_file_path.parent.mkdir(exist_ok=True, parents=True)
+    with output_file_path.open("w", encoding="utf-8") as f:
         json.dump(entries, f, indent=2, ensure_ascii=False)
 
     logger.info(
         f"{GREEN}Generated index with {len(entries)} entries: {output_file_path}{RESET}"
     )
-    return output_file_path
+    return output_file_pathstr
```
carver, line 113: This return value doesn't seem to be used anywhere but in testing, and is easy to replace there. It seems reasonable to me to drop the return altogether. I posted a `refactor!` PR #619 to do that, unless the reviewer prefers for me to include that change here.

tools/github_readme_sync/index.py
```diff
@@ -101,14 +101,16 @@ def generate_index(docs_dir: str, output_file_path: str) -> str:
 
     entries = process_markdown_files(docs_dir)
 
-    Path(output_file_path).parent.mkdir(exist_ok=True, parents=True)
-    with open(output_file_path, "w", encoding="utf-8") as f:
+    output_file_pathstr, output_file_path = output_file_path, Path(output_file_path)
```
tristanls-tbp, line 104: **note**: When I see lots of `Path`/`str` back and forth, it seems like things should be updated to use `Path`s. However, I understand the desire to keep the PR scope limited for easy review.

PR #619, opened 2025-12-09
chore!: drop redundant return in generate_index()

carver:
It's just returning the input argument, so it seems cleaner to reuse the input on the caller side, instead of threading it back out the return value.

### DIFF
--- a/tools/github_readme_sync/index.py
+++ b/tools/github_readme_sync/index.py
@@ -87,7 +87,7 @@ def generate_index(docs_dir: str, output_file_path: Path):
         output_file_path: Path where to write the output file.
 
     Raises:
-        ValueError: If docs_dir or output_file_path is empty.
+        ValueError: If docs_dir is empty.
     """
     if _is_empty(docs_dir):
         raise ValueError("docs_dir cannot be empty")


tools/github_readme_sync/tests/index_test.py
```diff
@@ -34,11 +34,10 @@ def _create_file_and_generate_index(
         with open(md_file_path, "w", encoding="utf-8") as f:
             f.write(content)
 
-        index_file_path = generate_index(
-            self.temp_dir, str(Path(self.temp_dir) / "index.json")
-        )
+        index_file_path = Path(self.temp_dir) / "index.json"
+        generate_index(self.temp_dir, str(index_file_path))
```
tristanls-tbp: **suggestion (non-blocking)**: Instead of Path/str/Path back and forth, it seems that making `generate_index` accept `Path`s as input would be more straightforward.

tristanls-tbp (commented):
Looks good, thank you.

Looks like a conflict due to #618 merge.

**note**: Withholding approval until seeing the final form.

tristanls-tbp:
**note**: I updated PR title to `chore` as [`refactor` is reserved for `src/tbp/monty` changes](https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0010_conventional_commits.md#types).

tools/github_readme_sync/tests/index_test.py
carver: Cool, I took a pass at converting the method argument to a `Path`

tools/github_readme_sync/index.py
```diff
@@ -79,38 +79,30 @@ def _check_and_sanitize(
...
 
     Args:
         docs_dir: The directory containing markdown files to scan.
         output_file_path: Path where to write the output file.
 
-    Returns:
-        Path to the generated output file.
-
     Raises:
         ValueError: If docs_dir or output_file_path is empty.
```
tristanls-tbp:
```suggestion
        ValueError: If docs_dir is empty.
```

tristanls-tbp (approved):
Thank you for the update, looks good. I can see how this can easily scope creep, so thank you for stopping 🙂.

PR #620, opened 2025-12-09
docs: Future work updates

vkakerbeck:
I decided to make these updates in stages instead of one giant update. Here is the first batch :) 

Some pages don't have a lot of detail on them yet, but I just wanted to make sure they are in there before we retire the spreadsheet (https://docs.google.com/spreadsheets/d/1HlKqhMzYWxpW6_D0LtDd5qEuWRYHl6xMfxbIZ76lzng/edit?gid=0#gid=0). 
@ramyamounir when you get a chance, could you have a look at and add a little more detail on 
- Dynamic adjustment of hypothesis space size
- Sample offspring hypotheses
- Less dependency on first observation
- Improve Bounded Evidence Performance

where you think necessary. Don't worry spending too much time on the items that are already prototyped but some basic info on the outline of these can be useful, + we can always transition it into the main docs once integrated into the main code.

### DIFF
--- a/docs/future-work/learning-module-improvements/test-grid-object-models-for-unsupervised-learning.md
+++ b/docs/future-work/learning-module-improvements/test-grid-object-models-for-unsupervised-learning.md
@@ -16,6 +16,8 @@ To test this, one could use our existing environments (e.g. YCB objects in Habit
 
 As a next step, shortcomings and issue can be identified and potential solutions tested (ideally starting with just tweaking the existing parameters).
 
-Additionally, one could test on different datasets, such as showing a collection of different types of mugs and trying to learn their commonalities in a model.
+The existing datasets aren't ideal for testing the learning of categories of objects and generalizing. One would at the least have to update the way performance is measured to quantify improvements in those abilities. One could also [develop different datasets aimed more an testing categories](../environment-improvements/create-dataset-and-metrics-to-evaluate-categories-and-generalization.md), such as showing a collection of different types of mugs and trying to learn their commonalities in a model.
+
+One may also explore the GridObjectModel in combination with a hierarchical Monty setup and test whether object parts naturally emerge (e.g. mug handle learned at lower level and full mug at higher level without explicit supervision). Testing this would likely deserve it's own future work item.
 
 It would also be good to test how useful the `GridObjectModel` is for "forgetting" statistical irregularities. For instance, how fast a noisy observation will be forgotten (or never even added) or how different sets of model parameters lead to learning "that specific mug with a chip in it" or "any mug in my cabinet".
--- a/conf/experiment/config/logging/defaults.yaml
+++ b/conf/experiment/config/logging/defaults.yaml
@@ -13,4 +13,3 @@ resume_wandb_run: false
 wandb_id:
   _target_: wandb.util.generate_id
 wandb_group: "debugging"
-log_parallel_wandb: false
--- a/conf/experiment/config/logging/parallel_evidence_lm.yaml
+++ b/conf/experiment/config/logging/parallel_evidence_lm.yaml
@@ -14,4 +14,3 @@ wandb_handlers:
 wandb_group: "parallel_eval_runs"
 monty_log_level: "BASIC"
 python_log_level: "WARNING"
-log_parallel_wandb: true
--- a/conf/experiment/tutorial/monty_meets_world_2dimage_inference.yaml
+++ b/conf/experiment/tutorial/monty_meets_world_2dimage_inference.yaml
@@ -19,6 +19,7 @@ _target_: tbp.monty.frameworks.experiments.object_recognition_experiments.MontyO
 config:
   model_name_or_path: ${path.expanduser:${benchmarks.pretrained_dir}/surf_agent_1lm_numenta_lab_obj/pretrained/}
   n_eval_epochs: 1
+  show_sensor_output: true
   logging:
     run_name: monty_meets_world_2dimage_inference
     wandb_group: benchmark_experiments
--- a/conf/update_snapshots.py
+++ b/conf/update_snapshots.py
@@ -1,0 +1,60 @@
+# Copyright 2025 Thousand Brains Project
+#
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+"""Updates test snapshots from the current config.
+
+Usage:
+    python update_snapshots.py
+"""
+
+from pathlib import Path
+
+import hydra
+from omegaconf import OmegaConf
+
+from tbp.monty.frameworks.run_env import setup_env
+from tbp.monty.hydra import register_resolvers
+
+
+def update_snapshots(
+    experiment_dir: Path = Path(__file__).parent / "experiment",
+    experiment_prefix: str = "",
+    snapshots_dir: Path = Path(__file__).parent.parent / "tests" / "conf" / "snapshots",
+):
+    """Update snapshots for all experiments in the experiment directory.
+
+    Args:
+        experiment_dir: The directory containing the experiments.
+        experiment_prefix: The prefix to add to the experiment name (e.g. "tutorial/")
+        snapshots_dir: The directory to write the snapshots to.
+    """
+    for file_path in experiment_dir.glob("*.yaml"):
+        print(f"Updating snapshot: {file_path}")
+        with hydra.initialize(version_base=None, config_path="."):
+            config = hydra.compose(
+                config_name="experiment",
+                overrides=[f"experiment={experiment_prefix}{file_path.stem}"],
+            )
+            OmegaConf.to_object(config)
+            current_config_yaml = OmegaConf.to_yaml(config)
+            snapshot_path = snapshots_dir / f"{file_path.stem}.yaml"
+            with snapshot_path.open("w") as f:
+                f.write(current_config_yaml)
+
+
+if __name__ == "__main__":
+    setup_env()
+    register_resolvers()
+    update_snapshots()
+    update_snapshots(
+        experiment_dir=Path(__file__).parent / "experiment" / "tutorial",
+        experiment_prefix="tutorial/",
+        snapshots_dir=(
+            Path(__file__).parent.parent / "tests" / "conf" / "snapshots" / "tutorial"
+        ),
+    )
--- a/docs/contributing/documentation.md
+++ b/docs/contributing/documentation.md
@@ -23,21 +23,36 @@ You can add new words to the dictionary by adding them to `.vale/styles/config/v
 
 ## How to Install and Use Vale
 
-1. **Install Vale**
-  Download Vale from its [installation page](https://vale.sh/docs/install).
+### 1. **Install Vale**
 
-2. **Run Vale**
-  Use the following command in your terminal to run Vale:
+Install Vale using the following commands (do not use Homebrew):
 
-    ```bash
-    vale .
-    ```
+```shell Intel
+curl -L -o /tmp/vale.tar.gz \
+  https://github.com/errata-ai/vale/releases/download/v3.9.3/vale_3.9.3_macOS_64-bit.tar.gz
+tar -xzf /tmp/vale.tar.gz -C /tmp
+sudo mv /tmp/vale /usr/local/bin/vale
+```
+```shell ARM64 (Apple Silicon)
+curl -L -o /tmp/vale.tar.gz \
+  https://github.com/errata-ai/vale/releases/download/v3.9.3/vale_3.9.3_macOS_arm64.tar.gz
+tar -xzf /tmp/vale.tar.gz -C /tmp
+sudo mv /tmp/vale /usr/local/bin/vale
+```
 
-    Example output:
-    ```
-    ➜  tbp.monty git:(main) vale .
-    ✔ 0 errors, 0 warnings and 0 suggestions in 141 files.
-    ```
+### 2. **Run Vale**
+
+Use the following command in your terminal to run Vale:
+
+```bash
+vale .
+```
+
+Example output:
+```
+➜  tbp.monty git:(main) vale .
+✔ 0 errors, 0 warnings and 0 suggestions in 141 files.
+```
 
 # Relative Links
 
--- a/docs/contributing/style-guide.md
+++ b/docs/contributing/style-guide.md
@@ -120,6 +120,9 @@ For example, the following markdown creates the image below:
 
 Videos can be embedded using markdown link syntax. The tool will automatically convert video links to embedded video players.
 
+> [!NOTE]
+> Video links must be on their own line to be automatically converted. Links that appear inline with other text will remain as regular markdown links.
+
 YouTube videos:
 
 ```markdown
--- a/docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
+++ b/docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
@@ -2,10 +2,6 @@
 title: Using Monty in a Custom Application
 ---
 
-> [!WARNING]
->
-> Apologies, the code for this tutorial is out of date due to the major change in how we configure Monty. We'll update it soon™️.
-
 > ⚠️ This is an Advanced Tutorial
 > If you've arrived at this page and you're relatively new to Monty, then we would recommend you start by reading some of our other documentation first. Once you're comfortable with the core concepts of Monty, then we think you'll enjoy learning about how to apply it to custom applications in the following tutorial!
 
@@ -19,9 +15,7 @@ Monty aims to implement a **general-purpose algorithm for understanding and inte
 In the current implementation, movement should happen in 3D (or less) space and be tracked using Euclidean location coordinates. Although we are convinced that the basic principles of Monty will also apply to abstract spaces ([potentially embedded in 3D space](https://thousandbrains.discourse.group/t/abstract-concept-in-monty/533/4)) and we know that the [brain uses different mechanisms to encode space](https://youtu.be/zRRazfFstvY), the current implementation relies on 3D Euclidean space.
 
 # Customizing Monty for Your Application
-The diagram below shows the base abstract classes in Monty. For general information on how to customize those classes, see our guide on [Customizing Monty](../customizing-monty.md).
-The Experiment class coordinates the experiment (learning and evaluation). It initializes and controls Monty and the environment and coordinates the interaction between them.
-The `EmbodiedEnvironment` class is wrapped in an `EnvironmentInterface` subclass, which exposes methods to interact with the environment. An experiment can have two environment interfaces associated with it: one for training and one for evaluation.
+The diagram below shows the base abstract classes in Monty. For general information on how to customize those classes, see our guide on [Customizing Monty](../customizing-monty.md). The Experiment class coordinates the experiment (learning and evaluation). It initializes and controls Monty and the environment and coordinates the interaction between them. The `EmbodiedEnvironment` class is wrapped in an `EnvironmentInterface` subclass, which exposes methods to interact with the environment. An experiment can have two environment interfaces associated with it: one for training and one for evaluation.
 
 ![Class structure in tbp.monty. Each class can be customized independently, allowing for easy modification and testing of individual components.](../../figures/how-to-use-monty/monty_class_diagram.png#width=500px)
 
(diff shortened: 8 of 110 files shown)


ramyamounir:
@vkakerbeck, I added more context to the 4 items you mentioned. Let me know what you think and if they need more detail.

vkakerbeck:
Looks great, thank you Ramy!

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
```diff
@@ -0,0 +1,40 @@
...
+---
+
+The current Monty system initializes a fixed hypothesis space at the start of each episode and maintains it until the episode completes.
+While this approach simplifies hypotheses updating, it introduces several constraints that limit adaptability and scalability.
+Most importantly, it ties inference to episode boundaries and prevents Monty from adapting its hypothesis space as new observations are collected.
+
+There are several important limitations in the current `DefaultHypothesesUpdater` design:
+
+- **Fixed hypothesis space for the full episode.**
+A fixed hypothesis space implies a supervised notion of episode start and end.
```
nielsleadholm:
```suggestion
A fixed hypothesis space implies a supervised notion of an episode's start and end.
```

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
```diff
@@ -0,0 +1,40 @@
...
+Some optimizations exist, such as updating only the top hypotheses based on confidence thresholds (i.e., `x_percent_threshold`), but the overall hypothesis space remains fixed and cannot shrink or grow.
+
+The goal of this future work item is to replace the fixed hypothesis space with a dynamic hypothesis space.
+The system should be able to identify and delete hypotheses that fail to accumulate evidence, while also sampling new hypotheses that are informed by more recent observations, when needed.
+This would enable a shift toward continuous, episode-free inference where the system can navigate and understand its environment in a more natural way.
+
+> [!NOTE]
+> See the [feat.dynamic_resizing](https://github.com/thousandbrainsproject/feat.dynamic_resizing) feature branch to see the current prototype. This prototype validated that dynamically deleting and sampling hypotheses based on their evidence slope helps with accuracy (both object and pose detection) which speeding up inference as most steps have a very small hypothesis space. This feature is planned to be integrated into Monty. However, further ideas for heuristics to delete or resample new hypotheses are appreciated.
```
nielsleadholm:
```suggestion
> See the [feat.dynamic_resizing](https://github.com/thousandbrainsproject/feat.dynamic_resizing) feature branch to see the current prototype. This prototype validated that dynamically deleting and sampling hypotheses based on their evidence slope helps with accuracy (both object and pose detection) while speeding up inference as most steps have a very small hypothesis space. This feature is planned to be integrated into Monty. However, further ideas for heuristics to delete or resample new hypotheses are appreciated.
```

docs/future-work/learning-module-improvements/improve-bounded-evidence-performance.md
```diff
@@ -9,3 +9,17 @@ skills: python, research, monty
...
+In the current model, evidence grows infinitely as Monty compares each hypothesis’s predictions with the observed features at every matching step.
+Hypotheses accumulate evidence proportional to their similarity to the observed features, which produces unbounded scores.
+This behavior is undesirable both for computational reasons and for biological plausibility, real neurons appear to operate within finite ranges.
+
+Bounding evidence through methods such as [exponential moving averages](https://youtu.be/A1cOwvZpgjU?si=1oyk6-BTij6TaRdG&t=1668), weighted averages, or normalization by hypothesis age can keep scores within a bounded range.
+However, these techniques compress historical information and cause memory to fade over time.
+Monty loses the influence of important past features, such as a mug handle, as a result, recognition can become less accurate in long sequences.
+
+A promising direction is to pair bounded evidence with saliency driven saccades or other attention mechanisms.
+If the system consistently revisits the key discriminative features that are important for recognition, then those features remain present in recent history.
```
nielsleadholm:
```suggestion
If the system consistently revisits the key discriminative features that are important for recognition, and does so efficiently, then those features remain present in recent history.
```

docs/future-work/learning-module-improvements/improve-bounded-evidence-performance.md
```diff
@@ -9,3 +9,17 @@ skills: python, research, monty
 contributor: ramyamounir, scottcanoe
 status: paused
 ---
+
+In the current model, evidence grows infinitely as Monty compares each hypothesis’s predictions with the observed features at every matching step.
+Hypotheses accumulate evidence proportional to their similarity to the observed features, which produces unbounded scores.
+This behavior is undesirable both for computational reasons and for biological plausibility, real neurons appear to operate within finite ranges.
```
nielsleadholm: @vkakerbeck @ramyamounir is it worth talking about the issue that you (Ramy) highlighted of new hypotheses being unable to out-compete old ones? Unlike biological plausibility, that is a real practical issue, which bounded evidence could help resolve.

One other small benefit to maybe mention - bounded evidence would be more interpretable, as evidence values would be fixed to a given range, rather than arbitrarily high.

In general we should probably avoid listing biological plausibility as a motivator, unless we follow it with something like - "we are suspicious that ignoring this *particular* biological constraint will cause us to implement misguided approaches elsewhere". For example in this case, relying on the memory of unbounded evidence could cause us to ignore the importance of more efficient policies. Without this context, it implies we are chasing biological realism for it's sake alone. If we find ourselves giving this context multiple times, we could have a short statement in our documentation about this (similar to the "no deep learning please"), and which we can link to when the question of biological realism comes up.

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
```diff
@@ -0,0 +1,40 @@
...
+Most importantly, it ties inference to episode boundaries and prevents Monty from adapting its hypothesis space as new observations are collected.
+
+There are several important limitations in the current `DefaultHypothesesUpdater` design:
+
+- **Fixed hypothesis space for the full episode.**
+A fixed hypothesis space implies a supervised notion of episode start and end.
+Monty initializes hypotheses at the beginning and deletes them at the end, which prevents true continuous inference.
+Ideally, Monty should operate without the concept of episodes, able to explore, move across objects, switch contexts, and update its hypotheses in an online unsupervised manner.
+
+- **Dependence on the first observation.**
```
nielsleadholm, line 24: It would be worth linking to the less-dependence-on-first-observation future work item here.

docs/future-work/learning-module-improvements/less-dependency-on-first-observation.md
```diff
@@ -9,3 +9,14 @@ skills: python, research, monty
 contributor: ramyamounir
 status: in-progress
 ---
+
+In the current implementation, Monty initializes its entire hypothesis space during the first step of an episode.
+This design makes the system highly dependent on the quality of that initial observation.
+Any sensor noise, occlusion, or partial view influence all hypotheses generated for the remainder of the episode.
```
nielsleadholm:
```suggestion
Any sensor noise, occlusion, or partial view influences all hypotheses generated for the remainder of the episode.
```

docs/future-work/learning-module-improvements/sample-offspring-hypotheses.md
```diff
@@ -0,0 +1,23 @@
...
+This produces hypothesis spaces that do not benefit from Monty’s own evolving beliefs.
+As a result, Monty cannot refine promising hypotheses through targeted exploration of nearby poses.
+A useful extension is to introduce offspring hypotheses.
+Offspring hypotheses are sampled around the current highest scoring hypotheses.
+This approach resembles refinement strategies used in particle filters, and in SLAM systems that rely on particle filtering, where the most promising pose estimates generate additional candidate samples.
+By sampling around these top contenders, Monty can progressively refine its estimate of the object’s pose and sensor position as more evidence accumulates.
+
+This mechanism is particularly effective when the object is stable or stationary, because the refinement process can converge toward a true stationary pose with increasing precision.
+However, offspring sampling alone is not enough for general adaptability.
+Monty must also be able to sample hypotheses informed directly by the latest observation, so that it can detect abrupt changes in object ID and avoid over committing to a single hypotheses cluster.
```
nielsleadholm, line 24: Maybe add something like, "One can imagine this sampling being controlled in a goal-directed manner by an LM, such that offspring hypotheses are leveraged when it is particularly important to have a highly accurate hypothesis."

docs/future-work/learning-module-improvements/test-grid-object-models-for-unsupervised-learning.md
```diff
@@ -0,0 +1,21 @@
...
+skills: research, monty
+contributor: 
+status: open
+---
+
+Since developing the [constrained object models](../../how-monty-works/learning-module/object-models.md#object-models) (`GridObjectModel` class), we have not had the chance to extensively test them. We have especially not tested them much in unsupervised learning scenarios and as a way to generalize common features of similar objects.
+
+To test this, one could use our existing environments (e.g. YCB objects in Habitat) and unsupervised learning configs (like outline in [this tutorial](../../how-to-use-monty/tutorials/unsupervised-continual-learning.md)) as a starting point. From there, one would need to update configs to use the `GridObjectModel`, run experiments with different model parameters, and visualize the results.
+
+As a next step, shortcomings and issue can be identified and potential solutions tested (ideally starting with just tweaking the existing parameters).
```
nielsleadholm, line 17: It might be worth highlighting here that the existing dataset is limited when exploring unsupervised categoriazation, and linking to the `create-dataset-and-metrics-to-evaluate-categories-and-generalization` item. For example, you cannot simply look at the output "accuracy" metric and chase improving that.

docs/future-work/learning-module-improvements/test-grid-object-models-for-unsupervised-learning.md
```diff
@@ -0,0 +1,21 @@
...
+
+Additionally, one could test on different datasets, such as showing a collection of different types of mugs and trying to learn their commonalities in a model.
```
nielsleadholm: Maybe add something like, "Try to explore what decomposition naturally emerges with constrained models - does a model emerge, for example, for the concept of `mug-handle` (as suggested by visualizing the learned model)?"

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
```diff
@@ -0,0 +1,40 @@
...
+
+- **Inefficiency from maintaining unlikely hypotheses.**
+Since all hypotheses remain stored throughout the episode, even unlikely hypotheses require some computation and memory.
+This overhead becomes significant as Monty learns about more objects.
+Some optimizations exist, such as updating only the top hypotheses based on confidence thresholds (i.e., `x_percent_threshold`), but the overall hypothesis space remains fixed and cannot shrink or grow.
+
+The goal of this future work item is to replace the fixed hypothesis space with a dynamic hypothesis space.
+The system should be able to identify and delete hypotheses that fail to accumulate evidence, while also sampling new hypotheses that are informed by more recent observations, when needed.
+This would enable a shift toward continuous, episode-free inference where the system can navigate and understand its environment in a more natural way.
+
```
nielsleadholm, line 39: On the topic of "further heuristics to delete or resample", the following relates to a discussion about sampling more hypotheses at umbilical points. Maybe we could add that in somewhere here as a sub-item? If it doesn't get covered here when this item is completed, it could become a separate Future Work item.

--> The amount of samples added could be proportional to the certainty about the sensory information/how noisy it is. An umbilical point is inherently more noisy/less well defined, so we want to sample more, but there might be other forms of noisy input signals that we could accommodate in this way.

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
ramyamounir: I thought about linking to other existing future work items but I wasn't sure what would happen if those become completed and their pages were deleted. Do we have some way of detecting these broken links to future works that are removed?

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
ramyamounir: Fixed in [dafe2c563ef3](https://github.com/ramyamounir/tbp.monty/commit/dafe2c563ef3ead0ee7e1f50ab4fd8d63286d258)

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
ramyamounir: Fixed in [dafe2c563ef3](https://github.com/ramyamounir/tbp.monty/commit/dafe2c563ef3ead0ee7e1f50ab4fd8d63286d258)

docs/future-work/learning-module-improvements/improve-bounded-evidence-performance.md
ramyamounir: Fixed in [dafe2c563ef3](https://github.com/ramyamounir/tbp.monty/commit/dafe2c563ef3ead0ee7e1f50ab4fd8d63286d258)

docs/future-work/learning-module-improvements/less-dependency-on-first-observation.md
ramyamounir: Fixed in [dafe2c563ef3](https://github.com/ramyamounir/tbp.monty/commit/dafe2c563ef3ead0ee7e1f50ab4fd8d63286d258)

docs/future-work/learning-module-improvements/sample-offspring-hypotheses.md
ramyamounir: Note added in [81467fbeccae](https://github.com/ramyamounir/tbp.monty/commit/81467fbeccae4ef365d0cb13bd7c83567fa10035)

docs/future-work/learning-module-improvements/improve-bounded-evidence-performance.md
ramyamounir: Nice, yeah I think the points you mentioned make a better motivation for bounded evidence. Also the point about avoiding biological plausibility as the only motivation makes sense.

I've updated the writing a bit in [a7cab4069bb3](https://github.com/ramyamounir/tbp.monty/commit/a7cab4069bb3d929bc71c511102751b90cd2a3af), what do you think?

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
nielsleadholm: My understanding is that future work items don't get deleted - they will be marked as completed instead. @vkakerbeck that's correct right?

nielsleadholm:
Nice, thanks for those changes @ramyamounir !

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
vkakerbeck: yes, the plan is to just mark them as completed but not delete them. But I think there is also automatic link checking in the docs that would pick up if a page was deleted/renamed (@codeallthethingz)

vkakerbeck:
Thanks for the review Niels! I think I addressed all the comments that applied to my part of this PR. Let me know if I missed something.

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
nielsleadholm: @ramyamounir just going to tag you as not sure you will have seen this comment (it didn't come through until later because it was "Pending" on my end).

ramyamounir: I agree that it's generally a good idea to add samples proportional to the certainty about the sensory information. With an umbilical point, we are missing a big chunk of the information (i.e., PC directions), so it makes sense to sample more points around the surface normal.

My only concern here is that sampling 8 orientations uniformly around the surface normal may be a bit too coarse for the current prototype to make use of them. With the `DefaultHypothesesUpdater`, we can sample a "close enough" hypothesis and it can become the `correct_mlh` because we never delete it. But in `ResamplingHypothesesUpdater`, we aggressively delete hypotheses that do not have high enough recent evidence growth. I'm just speculating that the extra hypotheses sampled uniformly around the surface normal have a low chance of surviving the deletion phase, and will possibly only add to the runtime. The increase in runtime can be significant and it scales with the number of objects because we sample from all objects in memory; those extra hypotheses have to stick around for at least 5 steps (`min_age`).

Another point is that we have less dependency on a single observation with the current burst sampling prototype. We sample over 5 steps, so even if one of them is umbilical, the rest could have PC directions (especially if using saliency to sample). This was critical with the `DefaultHypothesesUpdater` because we only had one chance to get good hypotheses, but it's not that critical anymore.

We can still try it out and determine empirically how it affects performance and runtime, this is just my intuition.

nielsleadholm: Yeah that's a good point. If we adopt that approach, then this has some similarity to waiting until you sense a distinct feature before sampling much, because until you do sense it, the hypothesis space is too wide open. This also feels like what we intuitively do, e.g. if you are blindfolded and start touching an object, you might start to try to recognize a handful of basic surfaces in one or two orientations. Only after you start sensing interesting things (e.g. a tail, a spout) might you start to form hypotheses about specific objects - and specific poses.

The other thing with umbilical points is they may be more easily handled once we have better handling of symmetry. In other words, once we store that certain poses are symmetric, then initializing a single pose hypothesis will often cover a large number of potential object poses without needing to store and update these as separate hypothesis points. 

Curious for your thoughts on this @vkakerbeck

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
ramyamounir: I added a couple of links in [87ae70a93e05](https://github.com/ramyamounir/tbp.monty/commit/87ae70a93e05812e110856bda48f8da867a87aa0)

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
vkakerbeck: yes I think it makes a lot of sense to just sample more concrete hypotheses once we actually sense interesting parts (i.e. parts where PCs are defined). I think the umbilical point oversampling was important up to now bc we only initialize once, but now it will become less relevant. Maybe a different thing to tweak may be to explicitly initialize some extra features when salient and well fedinde features are visited (i.e. don't just add hypotheses when prediction error is high). This may also help/replace reanchoring.

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
codeallthethingz: confirmed.  There is an internal link checker that makes sure all the docs in the `hierarchy.md` file exist that runs as a GitHub action but you can also run it locally like this:

```
> python -m tools.github_readme_sync.cli check docs
Hierarchy check complete
```

ramyamounir: Cool!

vkakerbeck: @codeallthethingz does this just check the hierarchy.md file links? We were talking about links between pages like the ones Ramy added here https://github.com/ramyamounir/tbp.monty/commit/87ae70a93e05812e110856bda48f8da867a87aa0 But I think that is checked too right? With the external link checking here https://github.com/thousandbrainsproject/tbp.monty/blob/27e98e6dc1b581401a99546e9c91457e58fc681b/tools/github_readme_sync/cli.py#L68 ?

codeallthethingz: Yes, all the internal links between pages are also validated. 
The external link checker checks links to places on the internet so we can remove dead links if they become stale.

nielsleadholm (approved):
Looks great!

docs/future-work/learning-module-improvements/dynamic-adjustment-of-hypothesis-space-size.md
vkakerbeck: Where in the docs do we describe how to run the link checking locally?
Also, while looking for the command I came across this https://thousandbrainsproject.readme.io/docs/documentation#relative-links In the commit Ramy linked he added links to the pages directly, instead of relative links. Should I update this? Is it an issue?

vkakerbeck: Ah I found the answer to my first question (https://thousandbrainsproject.readme.io/docs/documentation#checking-links) I might put all the local check commands in one place to make it a bit easier to find

PR #624, opened 2025-12-10
refactor!: use `ProprioceptiveState` throughout

AgentRev:
As part of issue https://github.com/thousandbrainsproject/tbp.monty/issues/506 for the Simulator protocol refactor, this PR integrates all changes from [`prototype.simulator`](https://github.com/thousandbrainsproject/prototype.simulator) between commits https://github.com/thousandbrainsproject/prototype.simulator/commit/d5aa84e03107dc83aa04f66b900e804bad51cc50 and https://github.com/thousandbrainsproject/prototype.simulator/commit/8a8592e5e23a4fe02590c771b0144212968192ca inclusively, which are mostly `ProprioceptiveState`-related.

From [Tristan's list](https://github.com/thousandbrainsproject/tbp.monty/issues/506#issuecomment-3417492100), this corresponds to the item "`Simulator.states()` returns `ProprioceptiveState`".

The PR is bigger than usual, but all these changes have to come thru together for all tests to pass. However, the review probably won't be that difficult, since a lot of the changes are similar to one another.

### DIFF
--- a/src/tbp/monty/frameworks/models/salience/sensor_module.py
+++ b/src/tbp/monty/frameworks/models/salience/sensor_module.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 from typing import Any
 
 import numpy as np
+import quaternion as qt
 
 from tbp.monty.frameworks.models.abstract_monty_classes import SensorModule
 from tbp.monty.frameworks.models.motor_system_state import AgentState, SensorState
@@ -67,7 +68,8 @@ class HabitatSalienceSM(SensorModule):
         """Update information about the sensors location and rotation."""
         sensor = agent.sensors[SensorID(self.sensor_module_id + ".rgba")]
         self.state = SensorState(
-            position=agent.position + sensor.position,
+            position=agent.position
+            + qt.rotate_vectors(agent.rotation, sensor.position),
             rotation=agent.rotation * sensor.rotation,
         )
         self.motor_only_step = agent.motor_only_step
--- a/src/tbp/monty/frameworks/models/sensor_modules.py
+++ b/src/tbp/monty/frameworks/models/sensor_modules.py
@@ -424,7 +424,8 @@ class Probe(SensorModule):
         """Update information about the sensors location and rotation."""
         sensor = agent.sensors[SensorID(self.sensor_module_id + ".rgba")]
         self.state = SensorState(
-            position=agent.position + sensor.position,
+            position=agent.position
+            + qt.rotate_vectors(agent.rotation, sensor.position),
             rotation=agent.rotation * sensor.rotation,
         )
         self.motor_only_step = agent.motor_only_step
@@ -640,7 +641,8 @@ class HabitatSM(SensorModule):
         """Update information about the sensors location and rotation."""
         sensor = agent.sensors[SensorID(self.sensor_module_id + ".rgba")]
         self.state = SensorState(
-            position=agent.position + sensor.position,
+            position=agent.position
+            + qt.rotate_vectors(agent.rotation, sensor.position),
             rotation=agent.rotation * sensor.rotation,
         )
         self.motor_only_step = agent.motor_only_step


src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -134,9 +135,11 @@ def reset(self):
 
         if self.transform is not None:
             observation = self.apply_transform(self.transform, observation, state)
-        return observation, ProprioceptiveState(state) if state else None
+        return observation, state if state else None
```
tristanls-tbp: **issue**: Originally, I had to do the `if else` thing because `ProprioceptiveState(None)` would throw. I think that now it only needs to be `return observation, state` here and elsewhere.

tristanls-tbp:
```suggestion
        return observation, state
```

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -149,7 +152,7 @@ def step(self, actions: Sequence[Action]):
         state = self.env.get_state()
         if self.transform is not None:
             observation = self.apply_transform(self.transform, observation, state)
-        return observation, ProprioceptiveState(state) if state else None
+        return observation, state if state else None
```
tristanls-tbp:
```suggestion
        return observation, state
```

src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -625,21 +629,20 @@ def pre_episode(self):
         self.visited_locs = []
         self.visited_normals = []
 
-    def update_state(self, state):
+    def update_state(self, agent: AgentState):
```
tristanls-tbp, line 640: **issue**: This is a `SensorModule` API change. It looks like a good change, but it needs to be done in `SensorModule` and all implementations.

tests/unit/frameworks/models/salience/sensor_module_test.py
```diff
@@ -53,10 +55,10 @@ def setUp(self) -> None:
             return_inhibitor_class=MagicMock,
             snapshot_telemetry_class=MagicMock,
         )
-        self.state = {
-            "rotation": "i'm a rotation",
-            "location": "i'm a position",
```
tristanls-tbp, line 58: 🙂

tests/unit/frameworks/models/motor_policies_test.py
```diff
@@ -48,14 +52,17 @@ def setUp(self):
...
+        expected_state = AgentState(
+            sensors=self.agent_sensors,
+            position=self.default_sensor_state.position,
+            rotation=self.default_sensor_state.rotation,
+        )
         state = MotorSystemState(
             {
-                AgentID(f"{self.agent_id}"): expected_state,
-                "different_agent_id": {},
+                AgentID(self.agent_id): expected_state,
```
tristanls-tbp: **issue (non-blocking)**: Unnecessary initialization, `self.agent_id` is already an `AgentID`, here and below.

tests/unit/policy_test.py
```diff
@@ -10,6 +10,12 @@
 
 import pytest
 
+from tbp.monty.frameworks.agents import AgentID
+from tbp.monty.frameworks.models.motor_system_state import (
+    AgentState,
+    ProprioceptiveState,
+)
+
```
tristanls-tbp: **issue**: We want to avoid imports before evaluating `pytest.importskip`. All of these should come after `pytest.importskip`.

tristanls-tbp (changes requested):
@AgentRev all benchmarks fail. Here's a sample failure:

```
Wandb setup took 28.16535520553589 seconds
Error executing job with overrides: ['num_parallel=16', 'experiment=base_config_10distinctobj_dist_agent']
multiprocessing.pool.RemoteTraceback:
"""
  File "/home/me/.conda/envs/tbp.monty/lib/python3.8/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/home/me/tbp/tbp.monty/src/tbp/monty/frameworks/run_parallel.py", line 395, in single_evaluate
    exp.evaluate()
  File "/home/me/tbp/tbp.monty/src/tbp/monty/frameworks/experiments/monty_experiment.py", line 577, in evaluate
    self.run_epoch()
  File "/home/me/tbp/tbp.monty/src/tbp/monty/frameworks/experiments/monty_experiment.py", line 533, in run_epoch
    self.run_episode()
  File "/home/me/tbp/tbp.monty/src/tbp/monty/frameworks/experiments/object_recognition_experiments.py", line 39, in run_episode
    last_step = self.run_episode_steps()
  File "/home/me/tbp/tbp.monty/src/tbp/monty/frameworks/experiments/object_recognition_experiments.py", line 80, in run_episode_steps
    for loader_step, observation in enumerate(self.env_interface):
  File "/home/me/tbp/tbp.monty/src/tbp/monty/frameworks/environments/embodied_data.py", line 436, in __next__
    return self.execute_jump_attempt()
  File "/home/me/tbp/tbp.monty/src/tbp/monty/frameworks/environments/embodied_data.py", line 635, in execute_jump_attempt
    for ii, current_sensor in enumerate(pre_jump_state["sensors"].keys()):
TypeError: 'AgentState' object is not subscriptable
"""
The above exception was the direct cause of the following exception:
  File "/home/me/tbp/tbp.monty/src/tbp/monty/frameworks/run_parallel.py", line 727, in main
    run_episodes_parallel(
  File "/home/me/tbp/tbp.monty/src/tbp/monty/frameworks/run_parallel.py", line 621, in run_episodes_parallel
    for result in p.imap(single_evaluate, experiments):
  File "/home/me/.conda/envs/tbp.monty/lib/python3.8/multiprocessing/pool.py", line 868, in next
    raise value
TypeError: 'AgentState' object is not subscriptable 
```

To reproduce, run:

```
python run_parallel.py experiment=base_config_10distinctobj_dist_agent
```

src/tbp/monty/frameworks/models/sensor_modules.py
AgentRev: There's a little issue with that; this line now passes a `SensorState` to `HabitatSalienceSM(SensorModule)` instead of the `"i'm a rotation"` dict:

https://github.com/thousandbrainsproject/tbp.monty/blob/5a2e07bff691d2c493e34c975da6941da30b79df/tests/unit/frameworks/models/salience/sensor_module_test.py#L68

while this line passes an `AgentState` to `HabitatSM(SensorModule)`:

https://github.com/thousandbrainsproject/tbp.monty/blob/5a2e07bff691d2c493e34c975da6941da30b79df/src/tbp/monty/frameworks/models/monty_base.py#L147

so I'm unsure of how to deal with that function in `HabitatSalienceSM`:

https://github.com/thousandbrainsproject/tbp.monty/blob/5a2e07bff691d2c493e34c975da6941da30b79df/src/tbp/monty/frameworks/models/salience/sensor_module.py#L64-L66

Clearly, `SensorModule.state` should always be a `SensorState`, and `HabitatSalienceSM` has no `AgentState` to pass to `update_state()`.

Should an `AgentState` be added somehow, or should there be 2 separate methods both types, or something else?

tristanls-tbp: Oh, that's a mistake on my part when I implemented `HabitatSalienceSM`. That should correspond to whatever `monty_base.py` is doing, i.e. `AgentState`. Let me see if I can fix that...

tristanls-tbp: ~~Oh, I see now. The `sensor_module_test.py` should use `AgentState`, not `SensorState`. In this case, `sensors` can be empty and unused. I'll also make this comment inline to be explicit.~~

tests/unit/frameworks/models/salience/sensor_module_test.py
```diff
@@ -53,10 +55,10 @@ def setUp(self) -> None:
...
-        }
+        self.state = SensorState(
```
tristanls-tbp: ~~**issue**: As I mentioned in the other comment, it looks like this should be `AgentState`.~~

src/tbp/monty/frameworks/models/sensor_modules.py
tristanls-tbp: What might need to be done ~~later (not this PR)~~ is to examine the `AgentState` and retrieve the `rotation` and `position` using `self.sensor_module_id`. Without explicit types, it was not as evident as it is now.

tristanls-tbp: And a more comprehensive update, ~~even later on (not this PR),~~ would be to implement what you stated, that it should always be `SensorState`, which sounds right to me.

tristanls-tbp:
Benchmark results look good, no change.

Screenshot 2025-12-11 at 11 29 43

src/tbp/monty/frameworks/models/sensor_modules.py
AgentRev: So, the func signature would now be `update_state(self, agent: AgentState)` everywhere.

In that case, should the definition of `HabitatSalienceSM.update_state` be this?
```
self.state = agent.sensors[SensorID(self.sensor_module_id)]
```

or this, like `Probe.update_state`?
```
self.state = SensorState(
    position=agent.position,
    rotation=agent.rotation,
)
```

tristanls-tbp: 🤔 I think `HabitatSalienceSM.update_state` should be updated to be like the full implementation of ~~`Probe.update_state`~~ `HabitatSM.update_state`. Specifically:
```python
def update_state(self, agent: AgentState) -> None:
        sensor_position = agent.sensors[
            SensorID(self.sensor_module_id + ".rgba")
        ].position
        self.motor_only_step = agent.motor_only_step

        sensor_rotation = agent.sensors[
            SensorID(self.sensor_module_id + ".rgba")
        ].rotation
        self.state = SensorState(
            position=agent.position + sensor_position,
            rotation=agent.rotation * sensor_rotation,
        )
```
I don't think that's gonna affect any `HabitatSalienceSM` functionality as `self.state` is only used for snapshot telemetry.

@scottcanoe, do you see any issues with the above implementation of `update_state` for `HabitatSalienceSM`?

tests/unit/frameworks/models/salience/sensor_module_test.py
tristanls-tbp: **note**: Based on [my comment in the other thread](https://github.com/thousandbrainsproject/tbp.monty/pull/624#discussion_r2625067620), this should now be fine if we always process `AgentState` into a `SensorState`.

src/tbp/monty/frameworks/models/sensor_modules.py
AgentRev: That code is actually `HabitatSM.update_state` rather than `Probe.update_state`. If I use that, I need to add ` + ".rgba"` to `HabitatSalienceSMTest`'s `SensorID`.

However, even if I do that, the `assert_called_once_with` call from `test_step_snapshots_raw_observation_as_needed` fails, because `assert_called_once_with` is unequipped to compare non-identical NumPy arrays:

```
        if self.should_snapshot:  # type: ignore[attr-defined]
>           self.sensor_module._snapshot_telemetry.raw_observation.assert_called_once_with(  # type: ignore[attr-defined]
                data, self.state.rotation, self.state.position
            )

tests/unit/frameworks/models/salience/sensor_module_test.py:82: 
../miniconda3/envs/tbp.monty/lib/python3.8/unittest/mock.py:925: in assert_called_once_with
    return self.assert_called_with(*args, **kwargs)
../miniconda3/envs/tbp.monty/lib/python3.8/unittest/mock.py:911: in assert_called_with
    if expected != actual:
```
[...]
```
E       ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

../miniconda3/envs/tbp.monty/lib/python3.8/unittest/mock.py:2462: ValueError
```
ChatGPT said:

> The problem is that, for example, array([0, 0, 0]) == array([0, 0, 0]) produces array([True, True, True]) — NumPy refuses to evaluate that as a single boolean, so it raises the ValueError.

and suggested a custom `np.array_equal` matcher:

```
class ArrayEqual:
    def __init__(self, arr):
        self.arr = arr
    def __eq__(self, other):
        return np.array_equal(self.arr, other)

self.sensor_module._snapshot_telemetry.raw_observation.assert_called_once_with(
    data, self.state.rotation, ArrayEqual(self.state.position)
)
```

which does indeed resolve the `ValueError`. If you find that fix appropriate, where should I put `class ArrayEqual`?

tristanls-tbp: As this is the only place the matcher is used, please place the matcher in the same test file: `tests/unit/frameworks/models/salience/sensor_module_test.py` (if we ever use it elsewhere, we'll refactor).

**suggestion**: I think the `arr` type is `numpy.typing.ArrayLike`. I think `other` type is `Any`.

src/tbp/monty/frameworks/environment_utils/transforms.py
```diff
@@ -27,10 +29,36 @@
...
+    "Transform",
+    "TransformContext",
 ]
 
 
-class MissingToMaxDepth:
+@dataclass
+class TransformContext:
+    rng: np.random.RandomState
+    state: ProprioceptiveState | None = None
```
AgentRev, line 37: @tristanls-tbp Is the change from `State` to `ProprioceptiveState` here okay?

tristanls-tbp: Oh wow, how did I miss that? That's a bug that I introduced. Yeah, it should be `ProprioceptiveState`. Thank you.

tristanls-tbp:
**note**: Please re-request the review when I should take a look again.

Screenshot 2025-12-21 at 08 21 21

AgentRev:
Yeah I was waiting for your reply to my `TransformContext` comment, in case there were further changes needed before final review. All good now. In the meanwhile, I've already prepared the next PR.

tristanls-tbp:
fyi @thousandbrainsproject/monty-maintainers 

**note**: I think the code for `update_state` is correct. However, the side effect is that `Probe` and `HabitatSalienceSM` sensor modules will only ever work with Habitat sensors or sensors that split into an `.rgba` and `.depth` sensor (see Undesirable Effect 118 - I have to split one sensor state into two that use .depth and .rgba suffixes.). Also note that this is already a constraint of `HabitatSM`.

This is because when we setup a HabitatAgent with sensors, we query `RGBDSensorConfig` for its specification via [`RGBDSensorConfig.get_spec`](https://github.com/thousandbrainsproject/tbp.monty/blob/a71ee31deb0a60323dff2e94806a6ea46601aec4/src/tbp/monty/simulators/habitat/sensors.py#L92-118) and this method _multiplies_ the number of sensors by suffixing `.rgba` and `.depth` to the provided sensor ID. This would not be a problem in and of itself if it was kept as an implementation detail of Habitat. However, this multiplication of sensor _leaks out_ in two places. First, [`HabitatAgent.process_observations`](https://github.com/thousandbrainsproject/tbp.monty/blob/a71ee31deb0a60323dff2e94806a6ea46601aec4/src/tbp/monty/simulators/habitat/agents.py#L105):
```python
        for sensor in self.sensors:
            sensor_id = SensorID(sensor.sensor_id)
            sensor_obs = obs_by_sensor.get(sensor_id)
            if sensor_obs is not None:
                obs_by_sensor[sensor_id] = sensor.process_observations(sensor_obs)
```
The above `for sensor in self.sensors` loop does not combine the `.rgba` and `.depth` sensors back into one, but instead exposes them to Monty.

Second, [`HabitatSim.states`](https://github.com/thousandbrainsproject/tbp.monty/blob/a71ee31deb0a60323dff2e94806a6ea46601aec4/src/tbp/monty/simulators/habitat/simulator.py#L589):
```python
            sensors = {}
            for sensor_id, sensor in agent_node.node_sensors.items():
                rotation = sim_utils.quat_from_magnum(sensor.node.rotation)
                sensors[sensor_id] = {
                    "position": sensor.node.translation,
                    "rotation": rotation,
                }
```
Similarly as before, the `for sensor_id, sensor in agent_node.node_sensors.items()` loop does not combine the `.rgba` and `.depth` sensor back into one.

This has consequences in multiple places, enumerated in the Current Reality Tree:

Screenshot 2025-12-22 at 11 16 33


This pull request's update to `update_state` is not the cause of this problem. It just dutifully deals with and propagates the consequences of the Habitat RGBDSensorConfig implementation.

After I merge this pull request, I'll need to update the Current Reality Tree with additional observed effects of: 
- HabitatSalienceSM assumes sensor state has a hardcoded "{sensor_module_id}.rgba" rgba sensor pose.
- Probe assumes sensor state has a hardcoded "{sensor_module_id}.rgba" rgba sensor pose.

Again, the fix for this is to address the Critical Causes(405 HabitatAgent.process_observations does not recombine .rgba and .depth sensors back into one.) and (408 HabitatSim.states does not recombine .rgba and .depth sensors back into one.), which are beyond the scope of this pull request as they will require dealing with a few cascading effects. Specifically, implementing (Injection # 73 Do not leak Habitat's RGBD sensor implementation (rgba + depth split) outside of Habitat.)
Screenshot 2025-12-22 at 11 02 27

src/tbp/monty/frameworks/models/salience/sensor_module.py
```diff
@@ -61,9 +63,14 @@ def sensor_module_id(self) -> str:
...
 
-    def update_state(self, state):
-        """Update the state of the sensor module."""
-        self.state = state
+    def update_state(self, agent: AgentState):
+        """Update information about the sensors location and rotation."""
+        sensor = agent.sensors[SensorID(self.sensor_module_id + ".rgba")]
+        self.state = SensorState(
+            position=agent.position + sensor.position,
+            rotation=agent.rotation * sensor.rotation,
```
tristanls-tbp: **issue**: It took me a while... but seeing this for the nth time, I finally noticed that this is not the correct way to compose 6D poses. The sensor position needs to be rotated into the agent's orientation before being added. I think this only worked before because most of the time the sensor rotation and position are identity (i.e., identity quaternion and (0,0,0))) and this is only used for logging.

**suggestion**: I believe the correct formula for composing 6D poses is:
```python
position=agent.position + qt.rotate_vectors(agent.rotation, sensor.position),
rotation=agent.rotation * sensor.rotation,
```

@scottcanoe am I right?

**note**: Not your fault @AgentRev, I know this copies what we have in `HabitatSM`.

src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -416,19 +420,19 @@ def __init__(
...
         """Update information about the sensors location and rotation."""
-        # TODO: This stores the entire AgentState. Extract sensor-specific state.
-        self.state = state
+        sensor = agent.sensors[SensorID(self.sensor_module_id + ".rgba")]
+        self.state = SensorState(
+            position=agent.position + sensor.position,
+            rotation=agent.rotation * sensor.rotation,
```
tristanls-tbp: **issue**: incorrect 6D pose composition (mentioned elsewhere)

**suggestion**: I believe the correct formula for composing 6D poses is:
```python
position=agent.position + qt.rotate_vectors(agent.rotation, sensor.position),
rotation=agent.rotation * sensor.rotation,
```

src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -625,21 +629,14 @@ def pre_episode(self):
...
-        agent_rotation = state["rotation"]
-        sensor_rotation = state["sensors"][self.sensor_module_id + ".rgba"]["rotation"]
-        self.state = {
-            "location": agent_position + sensor_position,
-            "rotation": agent_rotation * sensor_rotation,
-        }
+        sensor = agent.sensors[SensorID(self.sensor_module_id + ".rgba")]
+        self.state = SensorState(
+            position=agent.position + sensor.position,
+            rotation=agent.rotation * sensor.rotation,
```
tristanls-tbp: **issue**: incorrect 6D pose composition (mentioned elsewhere)

**suggestion**: I believe the correct formula for composing 6D poses is:
```python
position=agent.position + qt.rotate_vectors(agent.rotation, sensor.position),
rotation=agent.rotation * sensor.rotation,
```

tests/unit/frameworks/models/salience/sensor_module_test.py
```diff
@@ -13,17 +13,32 @@
 from unittest.mock import MagicMock, patch, sentinel
 
 import numpy as np
-import numpy.testing as npt
+import numpy.testing as npte
```
tristanls-tbp: **nitpick**: Can we call it `nptest`? I'm starting to lose tracking with the three of them: `np`, `npt`, `npte`.

AgentRev: Sure haha, Ruff was very adamant that `numpy.typing` be `npt`, and there was no precedent for having `testing` and `typing` in the same file. Should I change all the `numpy.testing as npt` to `nptest` everywhere else too for consistency?

src/tbp/monty/frameworks/models/salience/sensor_module.py
AgentRev: Those suggested formulas are correct if `AgentState.sensors[...]: SensorState` is in egocentric coordinates and `HabitatSalienceSM.state: SensorState` is in global coordinates. However, `SensorState`'s docstrings explicitly state "relative to the agent" (egocentric), so there's some ambiguity to clarify here.

scottcanoe: @tristanls-tbp I believe your formula is correct... it mirrors what we have in `DepthTo3DLocations` ~line no. 470. And that code works with non-trivial SM displacements and orientations. (That said, I'm not really familiar with `quaternion.rotate_vectors`, so I can't speak to its usage in particular.)

The egocentric vs global coordinates thing is confusing, as @AgentRev said. I have noticed that SM states are always _logged_ in global coordinates, not relative to the agent, and that is a gotcha. I wanted to get a response in regarding your formula, but I can take a closer look at the entire PR tomorrow to get a better sense of which is the correct choice.

tristanls-tbp:
Reverified benchmarks still show no change for the current code (as expected)

Screenshot 2025-12-23 at 07 24 38

tests/unit/frameworks/models/salience/sensor_module_test.py
tristanls-tbp: > Should I change all the numpy.testing as npt to nptest everywhere else too for consistency?

Yes please, that sounds reasonable, especially considering the linting demands.

src/tbp/monty/frameworks/models/salience/sensor_module.py
tristanls-tbp: Ah, right, good point regarding being explicit about what reference frame we want to be in @AgentRev 🙂.

[Documentation also states](https://github.com/thousandbrainsproject/tbp.monty/blob/a71ee31deb0a60323dff2e94806a6ea46601aec4/docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md?plain=1#L61):
> Related to defining how actions change observations, you will also need to define how actions change the state of the agent. This is what the `get_state()` method returns. The returned state needs to be a `ProprioceptiveState(Dict[str, AgentState])` with an entry per agent in the environment. The entry should contain the agent's position and orientation relative to some global reference point (`AgentState`). For each sensor associated with that agent, a `SensorState` should contain the sensor's position and orientation relative to the agent.

So, I would surmise that @AgentRev's assessment is correct and `AgentState` being passed into `update_state()` is in global coordinates, and the `SensorState` inside `AgentState.sensors[...]` is in egocentric coordinates, and `SensorModule.state` is in global coordinates (this is what is logged as @scottcanoe highlighted).

I think we do want `SensorModule.state` to be in global coordinates. I see no reason not to do that for our current use cases. **suggestion**: It would be helpful to make a note of this in `SensorModule.state` docstring.

_It continues to be unfortunate that we do not include the reference frame as part of the coordinates. We'll get to it eventually 😬._

Screenshot 2025-12-23 at 07 39 53

**suggestion**: I think the source of confusion is that the docstring on `SensorState` claims what reference frame it is in, where it cannot make that claim. I think the docstring about what reference frame `AgentState` and `SensorState` are in should be restricted to `ProprioceptiveState`, so that when we use the impoverished (without reference frame) types of `AgentState` and `SensorState`, we can use them in different reference frames.

I created #660 to address the docstrings.

@scottcanoe how does this sound?

scottcanoe: @tristanls-tbp I think that sounds good, assuming I'm understanding it correctly, but I'm a little behind on the `[Proprioceptive,Agent,Sensor]State` refactoring. My only real concern is that coordinate systems/reference frames aren't changed, at least at the level of the motor system. The `LookAt` policy expects sensor module states in global coordinates (I'm pretty sure... ), so performing any kind of transformation is going to break all that.

**Note:** It did feel a bit awkward to me that I'd have to convert a sensor module's state into local coordinates in order to generate the correct actions for the `LookAt` policy, which basically means inverting all the work that was done to get its state in global coordinates to begin with. At the time, it seemed like this was a good reason to provide `SensorModule.state` in local coordinates. All of that was weirdly non-trivial though, and I really don't want to put that particular thinking cap back on right now. So I'm just reiterating my preference for not changing any expected coordinate systems until we really really want to.

nielsleadholm: Just adding my two cents - when performing some kind of computation, like moving the sensor-agent pair, it makes most sense to store and generally handle the sensor using coordinates relative to the agent. At the end of the day, it is coupled to the agent, so things like the position and orientation of the sensor will be affected by changes to the agent. By storing and passing a world-coordinate location for the sensor, this dependency risks getting lost - it might feel like we can move it independently of the agent, which we cannot (or should not be able to).

On the other hand, it is more intuitive when outputing the sensor's state for logging/visualization purposes to give it's location in global coordinates. This prevents needing to reconstruct its location and orientation from details of the agent when you just want to visualize it for debugging purposes or scientific interest. 

"It continues to be unfortunate that we do not include the reference frame as part of the coordinates. We'll get to it eventually" - agreed! Could we take a page from this already, and update the variable names to nod to this for this particular state/location?

tristanls-tbp (commented):
Thank you @AgentRev, things look good to me.

I will run the benchmarks to verify again that the changes in pose composition formulas have no impact on anything. If those pass then I'll approve and merge.

Non-blocking suggestions below can be different PRs. They don't need to be included here and they don't need to be done by you 🙂.

**suggestion (non-blocking)** We repeat the pose composition formula in multiple places. It's not the easiest one to remember. Feels like a refactoring opportunity.

**suggestion (non-blocking)**: Per @nielsleadholm's [comment](https://github.com/thousandbrainsproject/tbp.monty/pull/624#discussion_r2643917369), I also think that it might be helpful to put the reference frame into the variable name (until we fix things more broadly). So, a rename from `SensorModule.state` to something like `SensorModule.state_in_agent_reference_frame` might be useful. I didn't suggest calling it `SensorModule.state_in_global_reference_frame` since locally, we don't know what reference frame `AgentState` is in. We only know that we are transforming into the agent frame.

PR #625, opened 2025-12-10
refactor: enable ruff rule S301, avoid pickle

carver:
The only active usage is in `TestLogger`. The `json` library is basically a drop-in replacement for this usage. The only change is to convert the [serialized file type from binary to text](https://docs.python.org/3/library/json.html#json.dump).

### DIFF
--- a/conf/update_snapshots.py
+++ b/conf/update_snapshots.py
@@ -1,0 +1,60 @@
+# Copyright 2025 Thousand Brains Project
+#
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+"""Updates test snapshots from the current config.
+
+Usage:
+    python update_snapshots.py
+"""
+
+from pathlib import Path
+
+import hydra
+from omegaconf import OmegaConf
+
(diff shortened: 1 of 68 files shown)


src/tbp/monty/frameworks/loggers/exp_logger.py
```diff
@@ -166,8 +166,8 @@ def post_eval(self, logger_args, output_dir, model):
     @override
     def close(self, logger_args, output_dir, model):
         outfile = Path(output_dir) / "fake_log.pkl"
```
tristanls-tbp:
```suggestion
        outfile = Path(output_dir) / "fake_log.json"
```

tests/unit/run_test.py
```diff
@@ -111,8 +111,8 @@ def test_main(self):
 
         output_dir = Path(self.cfg.experiment.config.logging.output_dir)
 
-        with (output_dir / "fake_log.pkl").open("rb") as f:
-            exp_log = pickle.load(f)
+        with (output_dir / "fake_log.pkl").open("r") as f:
```
tristanls-tbp:
```suggestion
        with (output_dir / "fake_log.json").open("r") as f:
```

carver:
Ah yes, good catch on the file extension 👍🏻

tristanls-tbp (approved):

PR #626, opened 2025-12-10
refactor: add small utility to make a data path

carver:
It defaults to being based in `env['MONTY_DATA']`, and always returns a Path object.

The quirkiest bit of this change is that it's less ergonomic to specify the subpath as a proper `Path()` object now, especially using the `/` separator. I just left it as the concatenated string, because it looks cleaner to me. Here's how it looks using the `/` separator:

```diff
diff --git a/src/tbp/monty/frameworks/environments/two_d_data.py b/src/tbp/monty/frameworks/environments/two_d_data.py
index e98ceec..a317cde 100644
--- a/src/tbp/monty/frameworks/environments/two_d_data.py
+++ b/src/tbp/monty/frameworks/environments/two_d_data.py
@@ -274,7 +274,7 @@ class SaccadeOnImageEnvironment(EmbodiedEnvironment):
         # the same. Since we don't use this, value doesn't matter much.
         self.rotation = qt.from_rotation_vector([np.pi / 2, 0.0, 0.0])
         self.state = 0
-        self.data_path = make_data_path(data_path, "worldimages/labeled_scenes")
+        self.data_path = make_data_path(data_path, Path("worldimages") / "labeled_scenes")
         self.scene_names = [a.name for a in sorted(self.data_path.glob("[!.]*"))]
         self.current_scene = self.scene_names[0]
         self.scene_version = 0
```

### DIFF
--- a/conf/update_snapshots.py
+++ b/conf/update_snapshots.py
@@ -1,0 +1,60 @@
+# Copyright 2025 Thousand Brains Project
+#
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+"""Updates test snapshots from the current config.
+
+Usage:
+    python update_snapshots.py
+"""
+
+from pathlib import Path
+
+import hydra
+from omegaconf import OmegaConf
+
+from tbp.monty.frameworks.run_env import setup_env
+from tbp.monty.hydra import register_resolvers
+
+
+def update_snapshots(
+    experiment_dir: Path = Path(__file__).parent / "experiment",
+    experiment_prefix: str = "",
+    snapshots_dir: Path = Path(__file__).parent.parent / "tests" / "conf" / "snapshots",
+):
+    """Update snapshots for all experiments in the experiment directory.
+
+    Args:
+        experiment_dir: The directory containing the experiments.
+        experiment_prefix: The prefix to add to the experiment name (e.g. "tutorial/")
+        snapshots_dir: The directory to write the snapshots to.
+    """
+    for file_path in experiment_dir.glob("*.yaml"):
+        print(f"Updating snapshot: {file_path}")
+        with hydra.initialize(version_base=None, config_path="."):
+            config = hydra.compose(
+                config_name="experiment",
+                overrides=[f"experiment={experiment_prefix}{file_path.stem}"],
+            )
+            OmegaConf.to_object(config)
+            current_config_yaml = OmegaConf.to_yaml(config)
+            snapshot_path = snapshots_dir / f"{file_path.stem}.yaml"
+            with snapshot_path.open("w") as f:
+                f.write(current_config_yaml)
+
+
+if __name__ == "__main__":
+    setup_env()
+    register_resolvers()
+    update_snapshots()
+    update_snapshots(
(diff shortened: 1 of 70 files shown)


src/tbp/monty/path.py
tristanls-tbp, line 1: **issue**: Please move to `src/tbp/monty/path.py` (edited).

**note**: I am trying to purge any "utils" from code. I'm in the theres-always-a-better-name-than-utils camp.

**note**: I thought that this might belong in `run_env.py`, but I'm uncertain about that.

src/tbp/monty/frameworks/utils/path.py
```diff
@@ -0,0 +1,27 @@
...
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+
+import os
+from pathlib import Path
+
+
+def make_data_path(custom_data_path: str, default_subpath: str) -> Path:
```
tristanls-tbp: **issue**: Please rename to `monty_data_path`. I feel that it more closely reflects what it is doing (creating and being coupled to the MONTY_DATA path) and is clearer in usage:

```python
 self.data_path = monty_data_path(data_path, "omniglot/python")
```

tests/unit/frameworks/utils/path_test.py
```diff
@@ -0,0 +1,38 @@
...
+    ("data_path", "subpath", "expected_result"),
+    [
+        (None, "default/subpath", Path(MOCK_MONTY_DATA) / "default" / "subpath"),
+        (None, "", Path(MOCK_MONTY_DATA)),
+        ("/custom/data/path", "default/subpath", Path("/custom/data/path")),
+        # A slightly odd scenario, but supported:
+        ("", "default/subpath", Path()),
+        # Not officially supported cases, but good to support anyway:
+        (Path("/custom/already/path"), "default/subpath", Path("/custom/already/path")),
+        (None, Path("sub") / "aspath", Path(MOCK_MONTY_DATA) / "sub" / "aspath"),
```
tristanls-tbp: **issue**: This suggests that function signature should be:
```python
def monty_data_path(custom_data_path: str|Path, default_subpath: str|Path) -> Path: ...
```

src/tbp/monty/path.py
tristanls-tbp: **issue**: Correction, please move to `src/tbp/monty/path.py`.

carver:
> **note**: The pattern of encapsulating an `if x is None:` check in a function feels weird to me. Maybe that's just my taste. I considered creating a property accessor and placing that logic within it, but it didn't feel any better.

Right, I also considered forcing the `data_path` argument to environments be a `Path`. But it looks like the arguments must be primitives, because they are coming direct from the yaml configuration files. If `data_path` was already a `Path`, then the helper could just build the default path. Something like:
```py
self.data_path = data_path if data_path else default_monty_data("omniglot/python")
```

If `data_path` stays a `str`, then the line expands to:

```py
self.data_path = Path(data_path) if data_path is not None else default_monty_data("omniglot/python")
```
(which is 100 characters, without indentation, and a smidge more work to read)

---

So I'll just make the requested changes. I also can't find something I like better.

src/tbp/monty/path.py
carver: Cool, I also moved the test from `tests/unit/frameworks/utils/path_test.py` to `tests/unit/path_test.py`. If there's a place that makes more sense to you, let me know.

PR #627, opened 2025-12-11
refactor!: enable ruff PLR1704, don't reuse argument name

carver:
Resolve [PLR1704](https://docs.astral.sh/ruff/rules/redefined-argument-from-local/). The trivial change in these two rule warnings would be to just change the variable name in the loop. The first commit uses this approach.

But I think they both point out awkwardness in the method API. Both rule failures are in methods that try to combine the call case of "get 1 thing by index" and "get all things". By splitting them out into two methods we get several benefits:
- simpler implementation in each
- simpler call convention (no special case argument value of `None` or `"all"`)
- simpler return value (one method always returns a single object, the other always returns a container of multiple)
- simpler docstring explanation

I feel most confident that the `get_current_displacement()` change is correct.

The `_calculate_most_likely_hypothesis()` is not so clear: 4c97a98b53778bb3cb16a33730ce1ef6f3fde808
I tried tracing the call sites to find out if a `graph_id` is ever passed as `None`. But I only got a few call sites deep before it branched out too far for me to manually verify. I included that change here just to share the concept, and out of curiosity if any test cases clearly break. I think it's a good direction. But since I couldn't verify correctness manually, and I don't trust the test coverage, I will revert it back out and just leave the trivial rule fix in place.

Also, I noticed that the conf tests break when habitat is not installed, so I added the skip mark.

### DIFF
--- a/src/tbp/monty/__init__.py
+++ b/src/tbp/monty/__init__.py
@@ -8,4 +8,4 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
-__version__ = "0.14.0"
+__version__ = "0.14.1"
--- a/src/tbp/monty/frameworks/environments/two_d_data.py
+++ b/src/tbp/monty/frameworks/environments/two_d_data.py
@@ -9,9 +9,7 @@
 # https://opensource.org/licenses/MIT.
 
 import logging
-import os
 import time
-from pathlib import Path
 from typing import Sequence
 
 import matplotlib.pyplot as plt
@@ -34,6 +32,7 @@ from tbp.monty.frameworks.models.abstract_monty_classes import (
     SensorID,
     SensorObservations,
 )
+from tbp.monty.path import monty_data_path
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


tests/conf/conf_test.py
```diff
@@ -7,6 +7,13 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
+import pytest
+
+pytest.importorskip(
+    "habitat_sim",
+    reason="Habitat Sim optional dependency not installed.",
+)
+
```
tristanls-tbp, line 16: **note**: Thank you. It's a bit disappointing, but forced due to our use of `OmegaConf.to_object(config)`, which will run the resolvers, which will look for Habitat Sim classes specified in `${monty.class:*}` resolvers, e.g., `${monty.class:tbp.monty.simulators.habitat.environment.HabitatEnvironment}`.

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -281,13 +281,19 @@ def get_current_displacement(self, input_channel):
...
+        """Get all current displacements.
+
+        Returns:
+            A dictionary mapping channels to all current displacements.
+        """
+        return {
+            channel: self.get_current_displacement(channel)
+            for channel in self.displacements.keys()
+        }
+
```
tristanls-tbp, line 296: **praise**: Thanks, this is much clearer 🙂.

tristanls-tbp:
No changes to benchmarks.

Screenshot 2025-12-11 at 17 40 20

PR #631, opened 2025-12-12
refactor: enable ruff rule S605, avoid os.system()

carver:
It's too easy to accidentally create shell injection attacks.


tristanls-tbp (approved):
Thank you, looks good.

**note**: I wonder if `"wb"` and `"rb"` would be safer.

carver:
Yeah, I suppose there's nothing to lose with binary encode/decode, and it's more robust in case someone uses a non-utf-8 encoding for some reason 👍🏻

PR #632, opened 2025-12-12
docs: Update Future Work description of off-object movements and out-of-model movements

nielsleadholm:
An effort to update and consolidate our current thinking on "off object" and "out of model" movements.

@hlee9212 I thought you would be interested to see these updated pages given your involvement with the earlier RFC drafts. Let me know any thoughts you have.

### DIFF
--- a/docs/future-work/learning-module-improvements/use-off-object-observations.md
+++ b/docs/future-work/learning-module-improvements/use-off-object-observations.md
@@ -3,8 +3,8 @@ title: Use Off-Object Observations
 description: Ensure that off-object observations are processed by LMs, resulting in evidence updates.
 rfc: optional
 estimated-scope: medium
-improved-metric: numsteps, accuracy
-output-type: experiments, analysis, PR
+improved-metric: numsteps
+output-type: analysis, PR
 skills: python, research, monty, refactoring
 contributor: hlee
 status: scoping
@@ -18,7 +18,7 @@ Currently we have methods to move the sensor back onto the object, however we do
 *Example of a sensor moving off of an object and observing nothing.*
 
 To address this, we need to update how these observations are processed such that:
-1. These observations are formulated as an appropriate "null"-type observation - there is no surface to observe, and so there should be no morphological features. However, there would still be a location (e.g. where the finger tip is hovering, or an eye looking into the far distance), and there can be non-morphological features like color.
+1. These observations are formulated as an appropriate "null"-type observation - there is no surface to observe, and so there should be no morphological features. However, there would still be a location (e.g. where the finger tip is hovering, or an eye looking into the far distance), and there can be non-morphological features like color. In the `State` class, this would likely involve setting the morphological features to a `None` type value.
 2. These observations are still passed to the learning module.
 3. If a hypothesis predicts an observation, then the learning module's evidence update appropriately results in negative evidence, as no object model should ever store a "null" feature that would match this observation.
 
@@ -45,7 +45,9 @@ From an initial look, this change might appear relatively straightforward, howev
 5. There is a [discontinued RFC on this topic](https://github.com/thousandbrainsproject/tbp.monty/pull/425/files) which has some useful discussion of the above points.
 
 ### Outcome Measures
-- We anticipate that the primary benefit of this change will be that Monty can use off-object observations to more quickly eliminate hypotheses; this will result in fewer steps before convergence.
-- However, it may also improve Monty's accuracy, at least on certain recognition tasks. In particular, consider the instance where Monty has two hypotheses, which are a number "1" and a number "7", or a letter "i" vs a short vertical bar "ı" without a dot. When the actual object observed is the one with fewer features / a smaller model (the "1" rather than the "7", or the i without a dot (ı) rather than "i"), then Monty currently cannot recognize the object. If it moves to the location where it might expect the upper bar of a seven, or the dot of an i, it will perceive nothing. Without processing this observation, Monty cannot distinguish the hypotheses. Thus, this future work item is necessary to resolve this issue. In addition however, resolving this setting will also require adjusting the [Use of Out of Model Movements](./use-out-of-model-movements.md) to ensure we briefly maintain the correct hypothesis (1 or ı, respectively). With these two changes in place, we expect the incorrect hypothesis for the larger model (7 or i) to receive negative evidence, while the correct hypothesis for the smaller model is maintained such that it can continue to accumulate evidence, and ultimately win out. This scenario would benefit from explicitly testing and examining once these changes have been made, which may require creating specific experimental setups.
+- We anticipate that the primary benefit of this change will be that Monty can use off-object observations to more quickly eliminate hypotheses; this will result in fewer steps before convergence. In particular, this should reduce the number of `monty_steps`, which include steps where we currently do not pass any information to LMs because Monty is "off object".
+- With follow-up work on [Use of Out of Model Movements](./use-out-of-model-movements.md), off-object observations can also help boost accuracy on certain challenging classification tasks. See this separate Future Work item for more details.
+- Until the [Use of Out of Model Movements](./use-out-of-model-movements.md) item has been completed, large movements off object may overly punish reasonable hypotheses before we can move back onto the object. Practically, this may reduce accuracy on some of the benchmarks. It is therefore suggested when implementing off-object observations that their use can be toggled via a hyperparameter, as we may find that we need to keep them unused in our benchmarks until the second stage (out-of-model-movement processing) is completed.
 
-NOTE: The instance where Monty moves onto *another* object (for example, a distant wall, or an occluding robotic hand) is already handled by the existing matching process in Monty. In particular, this should result in unexpected sensory input if an LM believes it is still on the original object, and therefore will (correctly) result in negative evidence. This instance is therefore not considered the "off object" setting described above that we wish to resolve.
+> [!NOTE]
+> The instance where Monty moves onto *another* object (for example, a distant wall, or an occluding robotic hand) is already handled by the existing matching process in Monty. In particular, this should result in unexpected sensory input if an LM believes it is still on the original object, and therefore will (correctly) result in negative evidence. This instance is therefore not considered the "off object" setting described above that we wish to resolve.
--- a/docs/future-work/learning-module-improvements/use-out-of-model-movements.md
+++ b/docs/future-work/learning-module-improvements/use-out-of-model-movements.md
@@ -10,7 +10,7 @@ contributor: hlee
 status: paused
 ---
 
-This task relates to [Use Off Object Observations](./use-off-object-observations.md), and so it is recommended that you read that description first.
+This task relates to [Use Off Object Observations](./use-off-object-observations.md), and so it is recommended that you read that description first. We also recommend that Use Off Object Observations is completed before beginning work on Use Out-of-Model Movements.
 
 Consider the situation described under using off-object observations. What if a hypothesis actually believes it moved out of the model's space? Or alternatively, what if Monty senses something (such as another object), but the hypothesis again believes it has moved out of the model's space? This setup is shown diagrammatically below:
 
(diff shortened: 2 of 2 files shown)


carver:
One of the big questions that has come up in my mind is: will the Cortical Message Protocol (specifically, the `State` class) need any change? The first idea that comes to mind is to add a new explicit way to communicate an empty sensory experience. 

Right now, a sensor sets `State().use_state=False` when off-object, but other things can also set that to `False`, like when the surface normal or principal curvature can't be calculated. In order for the LM to interpret that the observation is off-object, it could look for a new explicit attribute like `State().null_observation`.

Any alternative ideas?

---

I've been assuming that the other cases where `use_state=False` are *not* equivalent to off-object movements. Things like the invalid surface normal, principal curvature, or a non-change in the `FeatureChangeFilter`. But if `use_state=False` is *always* interpretable as being off-object, then I guess changing the `State` class isn't necessary.

docs/future-work/learning-module-improvements/use-off-object-observations.md
```diff
@@ -1,12 +1,51 @@
 ---
 title: Use Off-Object Observations
-rfc: https://github.com/thousandbrainsproject/tbp.monty/pull/425/files
+description: Ensure that off-object observations are processed by LMs, resulting in evidence updates.
+rfc: optional
+estimated-scope: medium
+improved-metric: speed, numsteps, accuracy
```
vkakerbeck: Maybe these tags are a bit unclear but `speed` usually refers to improved speed per step (i.e. if the experiment is just faster because we take less steps the this is `numsteps`) I'm not sure why this would improve here.

docs/future-work/learning-module-improvements/use-off-object-observations.md
```diff
@@ -1,12 +1,51 @@
...
+output-type: experiments, analysis, PR
+skills: python, research, monty
+contributor: hlee
+status: scoping
 ---
 
-There are a variety of instances where a Monty system has a hypothesis about the current object, and then moves off the hypothesis-space of that object, either sensing nothing/empty space, or another object. For example, this can occur due to a model-free driven action like a saccade moving off the object, or the surface agent leaving the surface of the object. Similarly, a model-based action like the hypothesis-testing "jump" can move an agent to a location where the object doesn't exist if the hypothesis it tested was false.
+There are a few instances where Monty can move off of an object, sensing nothing/empty space. For example, this can occur due to a saccade moving off the object and pointing at a void (such as the sky), or a touch-based agent leaving the surface of an object.
```
vkakerbeck:
```suggestion
There are a few instances where Monty can move off of an object, sensing nothing/empty space (or at least no morphological features). For example, this can occur due to a saccade moving off the object and pointing at a void (such as the sky), or a touch-based agent leaving the surface of an object.
```

docs/future-work/learning-module-improvements/use-off-object-observations.md
```diff
@@ -1,12 +1,51 @@
...
+Currently we have methods to move the sensor back onto the object, however we do not make use of the information that there was no object at the perceived location before doing so. This is valuable information, as the absence of the object at a location will be inconsistent with some object and pose hypotheses - hypotheses that should therefore have their associated evidence reduced. This setup is shown in the diagram below.
 
-For example, if the most-likely hypothesis is a coffee mug and the system performs a saccade that results in the nearest feature being very far away (such as the distant surface of a table), then any hypotheses about the pose of the mug that predicted there would be mug-parts should receive evidence against them. On the other hand, a hypothesis about the mug's pose that is *consistent* with the absence of the mug at that location should receive positive evidence. The same would apply if the surface agent leaves the surface of the mug, where the absence of mug is consistent with a subset of hypotheses.
+![Example of off object observations](../../figures/future-work/off_object_obs.png)
```
vkakerbeck, line 17: Not sure how hard this will be to update but it could make it a lot easier to understand these figures if the mug hypotheses are visualized differently than what is actually sensed. For example showing the learned point cloud instead of the mug image

docs/future-work/learning-module-improvements/use-off-object-observations.md
```diff
@@ -1,12 +1,51 @@
...
+See [Use Out of Model Movements](./use-out-of-model-movements.md)
+
+### What if the movement is just off the surface of the object?
+
+This situation is shown in the below diagram:
+
+![Example of off object observations near a surface](../../figures/future-work/off_object_near_surface.png)
+*Example of a sensor moving off of an object and observing nothing when it is close to the object's surface.*
+
+In this case, while some hypotheses on the surface of the object might be close (even within the neighborhood tolerance) to the sensors actual location, these will receive negative evidence, as they will predict being on the actual object. This is the desired behavior, as the LM should not predict that it is still on the object if it is no longer sensing it.
```
vkakerbeck:
```suggestion
In this case, while some hypotheses on the surface of the object might be close (even within the neighborhood tolerance) to the sensors actual location, these will receive negative evidence, as they will predict being on the actual object. This is the desired behavior, as the LM should not predict that it is still on the object if it is no longer sensing it. However, we should keep this effect in mind when testing this as it might cause "almost correct" hypotheses to get a lot of negative evidence and thereby lead to some performance drops.
```

docs/future-work/learning-module-improvements/use-out-of-model-movements.md
```diff
@@ -0,0 +1,36 @@
...
+
+This task relates to [Use Off Object Observations](./use-off-object-observations.md), and so it is recommended that you read that description first.
+
+Consider the situation described under using off-object observations. What if a hypothesis actually believes it moved out of the model's space? Or alternatively, what if Monty senses something (such as another object), but the hypothesis again believes it has moved out of the model's space? This setup is shown diagrammatically below:
+
+<img src="../../figures/future-work/out_of_model_movement.png" alt="Example of a sensor movement resulting in the hypothesis moving out of the learned model an object" width="500">
+
+*Example of a sensor movement resulting in the hypothesis moving out of the learned model of an object.*
+
+NOTE: In the past we have sometimes referred to this as an "out of reference frame" movement. However, it is most accurate to describe the reference frame as the unbounded representational space where a model is learned. As such, describing it as an out-of-model-movement (OOMM) is more accurate.
```
vkakerbeck:
```suggestion
> [!NOTE]
> In the past we have sometimes referred to this as an "out of reference frame" movement. However, it is most accurate to describe the reference frame as the unbounded representational space where a model is learned. As such, describing it as an out-of-model-movement (OOMM) is more accurate.
```

docs/future-work/learning-module-improvements/use-out-of-model-movements.md
```diff
@@ -0,0 +1,36 @@
...
+
+We believe the best approach is that the LM would retain the OOMM hypothesis for a short period of time (perhaps 2-3 steps in the current, discrete version of Monty). If it then moves back into the model space before this time has elapsed, then the hypothesis can continue to accumulate evidence.
+
+If a hypothesis stays outside of the model space for too long however, then the hypothesis should receive strong negative evidence. In the resampling version of Monty, this would effectively be handled by the evidence slope being sufficiently small that the hypothesis is deleted.
+
+This approach ensures that a Learning Module is not expected to maintain a long-term memory of objects it is not currently on. At the same time, it means that we can move out of a model, which can be a useful way to test other hypotheses (i.e. for objects where we do expect to find something there). By then moving back onto the object and continuing to accumulate evidence, certain hypotheses can continue to grow. See the description of the 7 vs 1 and i vs ı under [Use Off Object Observations](./use-off-object-observations.md) for a concrete example of where this is relevant.
+
+Other details
```
vkakerbeck:
```suggestion
## Other details
```

docs/future-work/learning-module-improvements/use-out-of-model-movements.md
```diff
@@ -0,0 +1,36 @@
...
+- If a most-likely-hypothesis moves out of a model and becomes "clamped" at that evidence value as a result, it should not continue to pass information up in a hierarchy of LMs. Intuitively, if we believe we are on a mug at location x, and then move to location y which is off of the mug, then we should not communicate to the next LM in the hierarchy that there is a mug at location y.
+- What about object permanence, for example when moving back to where the object was after a significant period of time has elapsed? This should be captured via hierarchy - a higher-level LM would have a scene-like representation, and could use top-down biasing to help the LM in question recall what object it had just been observing. The mechanism here for OOMM only maintains hypotheses for a short period of time as a brief trace to enable returning to an object if a sensor strays off of it.
+
+Potential "gotchas" to be aware of:
```
vkakerbeck:
```suggestion
## Potential "gotchas" to be aware of:
```

docs/future-work/learning-module-improvements/use-off-object-observations.md
```diff
@@ -1,12 +1,51 @@
 ---
 title: Use Off-Object Observations
-rfc: https://github.com/thousandbrainsproject/tbp.monty/pull/425/files
+description: Ensure that off-object observations are processed by LMs, resulting in evidence updates.
+rfc: optional
+estimated-scope: medium
+improved-metric: speed, numsteps, accuracy
+output-type: experiments, analysis, PR
+skills: python, research, monty
```
vkakerbeck: It sounds like most of the experimenting and analysis is in the "Use Out of Model Movements" part. This element here is mostly about updating code and making sure everything still works as expected. I've used `experiments` for cases where new experiment configs would be added. Do you think this would be the case here? Maybe we also add `refactoring` to the skills?

vkakerbeck:
@carver thats a good question. One option we've been thinking about is setting `morphological_features` (https://github.com/thousandbrainsproject/tbp.monty/blob/38e191fffc38420c3d66428c88b523c204d35176/src/tbp/monty/frameworks/models/states.py#L58) to `None` (+updating the asserts here https://github.com/thousandbrainsproject/tbp.monty/blob/38e191fffc38420c3d66428c88b523c204d35176/src/tbp/monty/frameworks/models/states.py#L181). Because although there is no morphological feature (i.e. an orientation of a surface), there may still be non-morphological features like the blue color of the sky or a vague location. Although I am not sure if it makes any difference whether we transmit this information or not, since we wouldn't be using it in the LM anyways.
You've definitely right that the `use_state` flag is used in different instances where we would not want to send observations to the LM.

carver:
> there may still be non-morphological features like the blue color of the sky or a vague location. Although I am not sure if it makes any difference whether we transmit this information or not, since we wouldn't be using it in the LM anyways.

Won't we use the location in the LM? I thought the location of the null of sensory experience would be used to give negative evidence. I imagined it going like this: we match the null feature at a particular location against an expected feature in a model at a particular location.

This works more cleanly for a surface sensor, and less cleanly for a distant one. The distance sensor is giving us evidence about the absence of things in a whole column of visual space (I guess a bit more like a cone, really).

So, say we have two pose hypotheses of a coffee mug:
1. with the handle pointed mostly away from the sensor, but still peaking out behind the right side
2. with the handle pointed mostly toward the sensor, but slightly to the right

Then say we move the distant sensor to the right, where the hypothesized handles are, but it gets a null feature. Both models have expected points that should receive negative evidence by that feature. But those hypothesized points are also pretty distant from each other in hypothesis space.

So I suppose we have to collect all hypothesis points in a cone away from the sensor, to apply negative evidence to them.

vkakerbeck:
oh yes, you are right! We would still want to use the location to update the hypothesized location. So I think the way I was suggesting (setting non_morphological_features to None but still transmitting other parts of the message) would be the best way to go.

docs/future-work/learning-module-improvements/use-off-object-observations.md
nielsleadholm: Good point, removed

docs/future-work/learning-module-improvements/use-off-object-observations.md
nielsleadholm: Good idea, I've updated all of these, let me know what you think.

docs/future-work/learning-module-improvements/use-off-object-observations.md
nielsleadholm: Hmm I feel like there is actually a change (maintain evidence at same values for x steps) that goes beyond a refactor, and the out of model movement is specifically necessary for testing the resolution of the 7 vs 1 problem, which might benefit from a custom experiment to evaluate. I've added some words making it clearer about how to evaluate, and the connection to the other item. How does that look?

nielsleadholm:
Good questions/points @carver , and agree with Viviane's suggestion about setting the morphological features to a None type value, but still passing it so that we can process this as a mismatch (and therefore negative evidence) as you say.

Re. collecting hypotheses in a cone, in Monty we don't need to work backward from a visual cone to work out which hypotheses to update. Instead, all the hypotheses that are for that model would undergo movement and receive sensory observations to either increase or lower their evidence, no matter where they are in hypothesis space. This might include e.g. a hypothesis for the bottom of the mug if that still has enough evidence to continue being updated. In practice, after several steps, only hypotheses that are consistent with the physical observations will survive, but the main point is we don't need to work backward from a visual cone/ray. Hope that makes sense.

nielsleadholm:
Thanks for the suggestions @vkakerbeck , I've had a go at addressing those.

docs/future-work/learning-module-improvements/use-off-object-observations.md
vkakerbeck: nice, looks great!

docs/future-work/learning-module-improvements/use-off-object-observations.md
vkakerbeck: I'm not sure your changes are all applied yet, I still see the `speed` metric and don't see the additional experiment clarification you mentioned in the other comment...

docs/future-work/learning-module-improvements/use-off-object-observations.md
vkakerbeck: I'm not sure I see the added wording...
This is the metadata for `Use Off-Object Observations`, not `Use Out of Model Movements`, why would this item here include maintaining a hypothesis?
The skills aren't meant to be exclusive. So just because something requires the ability to do refactoring doesn't mean it is "only" a refactoring task. Does that make sense?

docs/future-work/learning-module-improvements/use-off-object-observations.md
nielsleadholm: Oh sorry, not sure why the speed change hadn't saved.

Re. the additional experiment clarification, let me re-adjust it (see other thread).

docs/future-work/learning-module-improvements/use-off-object-observations.md
nielsleadholm: Ah ok I think I misread your comment the other way around. Let me read it again and think if there is anything else to change.

nielsleadholm: Yeah makes sense now why you mentioned refactor, for some reason I thought you were referring to the other future work item. 

There are now some wording adjustments in both tickets, and I have also added refactoring. Let me know what you think.

In off-object:
`... This scenario would benefit from explicitly testing and examining once these changes have been made, which may require creating specific experimental setups.`

In out-of-model:
```
## Evaluations
-  See the description of the 7 vs 1 and i vs ı under [Use Off Object Observations](./use-off-object-observations.md) for a concrete example of where the above described situation is relevant for classification.
- It is recommended to develop experiments that specifically evaluate the use of out-of-model movements to help resolve this classification issue in tandem with off-object observations. 
```

vkakerbeck: I am still not sure I understand how this item would be able to test this (and achieve the distinction) without also tackling the OOMMs item. Can't we keep the two separate items? One is about sending the off object obs and making sure all related code still works, and the other item is about correctly processing them? Or this item here just adds negative evidence using the existing evidence update code, if there are points nearby in the model (although this may still be a bit problematic on its own)

nielsleadholm: Yeah they are definitely intertwined, in that both need to be addressed to fully resolve this issue. That is what I was hoping to capture with the text, i.e. you can still work on these in isolation, but just be aware that the eventual aim is to resolve both, and then perform these evaluations. "once these changes have been made" refers to both changes. Another option is to combine them into a single Future Work item, but I think they can be worked on in parallel, so personally I like the separation. 

If it's easier we can discuss it in person in the technical meeting.

vkakerbeck: Yeah I think if they are inseperabel we should combine them into one item. But from how I understand they are more like a sequence: `Use Off-Object Observations` ->  `Use Out of Model Movements` -> `test on specific new experiment`. The last one could be part of the OOMM one. But do we have to include the experiments in the first one already if they aren't meaningful without the second one? Couldn't we just add a note to the OOMM item that says that the off object observation passing item needs to be completed first?

nielsleadholm: Thanks @vkakerbeck , wording has been updated to generally make the sequence and outcomes clearer. Let me know what you think.

carver:
> Re. collecting hypotheses in a cone, in Monty we don't need to work backward from a visual cone to work out which hypotheses to update. Instead, all the hypotheses that are for that model would undergo movement and receive sensory observations to either increase or lower their evidence, no matter where they are in hypothesis space. ... Hope that makes sense.

Maybe I get it. If the following interpretation isn't right, then let's talk about it on the Thursday call instead of going back and forth in text. I'm imagining that for each hypothesis, the LM predicts the feature, shape, _and location_ of the model, given the sensor's orientation. If we treat the null observation as equivalent to a distant wall, giving it a far away location, we can measure the gap in location of the predicted location, and the sensed location.

Although now that I think about it, predicting the location of a sensed point on a hypothesis morphology takes some work. The most obvious way to do that in my mind is to measure a ray leaving the sensor, and finding where it intersects the model, which is essentially the cone idea I was toying with earlier. So maybe I don't get it, and we can talk on Thursday 😆 .

nielsleadholm:
> > Re. collecting hypotheses in a cone, in Monty we don't need to work backward from a visual cone to work out which hypotheses to update. Instead, all the hypotheses that are for that model would undergo movement and receive sensory observations to either increase or lower their evidence, no matter where they are in hypothesis space. ... Hope that makes sense.
> 
> Maybe I get it. If the following interpretation isn't right, then let's talk about it on the Thursday call instead of going back and forth in text. I'm imagining that for each hypothesis, the LM predicts the feature, shape, _and location_ of the model, given the sensor's orientation. If we treat the null observation as equivalent to a distant wall, giving it a far away location, we can measure the gap in location of the predicted location, and the sensed location.
> 
> Although now that I think about it, predicting the location of a sensed point on a hypothesis morphology takes some work. The most obvious way to do that in my mind is to measure a ray leaving the sensor, and finding where it intersects the model, which is essentially the cone idea I was toying with earlier. So maybe I don't get it, and we can talk on Thursday 😆 .

It's very close, yeah let's discuss on Thursday! But basically an LMs main job is to predict the location _within_ a model. All of its hypotheses correspond to different possible locations in the model's reference frame, and these are what get updated by movement. [Watching this video on particle filters may help](https://www.youtube.com/watch?v=NrzmH_yerBU) - Monty is doing something similar, but where the environment is a learned object model (e.g. 3D mug model) rather than a 2D room. 

For what it's worth, if you then want the location of the object in the world, an LM can work this out based on where its sensor is sensing, and where it thinks it is in the object's reference frame (as well as the object's orientation).

PR #636, opened 2025-12-12
refactor!: object params are deterministic per seed, epoch, and episode

tristanls-tbp:
This pull request is part of the work to get rid of follow-up configs by deterministically resetting random sources before each episode.

The object initializers are updated to accept `seed`, `epoch`, and `episode` parameters so that they can deterministically generate the same object parameters conditional on experiment seed, epoch, and episode. These correspond to reproducibility-per-episode that we want without presenting the same object params for every episode that we would get by initializing only from `seed` for each episode.

I added an `episode_seed` function to deterministically hash seed, epoch, and episode into a seed that `np.random.RandomState` will accept. It is placed in `tbp.monty.frameworks.experiments` because it knows about the concept of epochs and episodes, which makes it part of the experiment framework.

The workaround/hack in `run_parallel.py` that predefines object params into the parallel configurations is updated so that the object params correspond to what an object initializer would output for that seed, epoch, and episode during a non-parallell `run.py` run.

## Verification

To verify that this works, use a `Default` or `RandomRotation` object initializer, and run an experiment using `run.py` and then `run_parallel.py`. The `primary_target_position` and `primary_target_rotation_euler` for both run types should be equal for this PR and not equal before this PR.

### Before PR `@02f76dd` randrot_10distinctobj_surf_agent (RandomRotation)

Note that the values in the `primary_target_rotation_euler` column do not match between the two run modes (don't worry about the column format, focus on the values).

`run.py` (see: https://wandb.ai/thousand-brains-project/Monty/runs/1p6jqe3s)

Screenshot 2025-12-12 at 17 34 06

`run_parallel.py` (see: https://wandb.ai/thousand-brains-project/Monty/runs/ehq79g90)

Screenshot 2025-12-12 at 17 42 40

### This PR randrot_10distinctobj_surf_agent (RandomRotation)

Note that the values in the `primary_target_rotation_euler` column match between the two run modes. The overall results are still different due to other sources of randomness, but the focus of this PR is on object param generation.

`run.py` (see: https://wandb.ai/thousand-brains-project/Monty/runs/zysjftrb)

Screenshot 2025-12-12 at 17 45 14

`run_parallel.py` (see: https://wandb.ai/thousand-brains-project/Monty/runs/i18rle99)

Screenshot 2025-12-12 at 17 54 56

## Benchmarks

Every benchmark that uses the `RandomRotation` object initializer has changed since we modified the source of randomness used to determine object parameters for each evaluation, e.g., `randrot_noise_10distinctobj_dist_agent`. The ones using `Predefined` object initializer are unaffected, e.g., `base_config_10distinctobj_dist_agent`.

Screenshot 2025-12-15 at 10 50 37

### DIFF
--- a/benchmarks/ycb_10objs.csv
+++ b/benchmarks/ycb_10objs.csv
@@ -1,13 +1,13 @@
 Experiment,Correct (%)|align right,Used MLH (%)|align right,Num Match Steps|align right,Rotation Error (degrees)|align right,Run Time (mins)|align right,Episode Run Time (s)|align right
 base_config_10distinctobj_dist_agent,99.29,0.71,34,10.69,3,9
 base_config_10distinctobj_surf_agent,99.29,0.00,27,11.96,3,11
-randrot_noise_10distinctobj_dist_agent,95.00,5.00,60,24.40,5,28
-randrot_noise_10distinctobj_dist_on_distm,96.00,4.00,44,13.73,4,18
-randrot_noise_10distinctobj_surf_agent,100.00,1.00,28,30.21,3,20
-randrot_10distinctobj_surf_agent,100.00,0.00,29,19.08,2,13
-randrot_noise_10distinctobj_5lms_dist_agent,100.00,0.00,53,44.71,5,40
+randrot_noise_10distinctobj_dist_agent,97.00,5.00,55,25.14,5,28
+randrot_noise_10distinctobj_dist_on_distm,98.00,3.00,38,15.11,4,18
+randrot_noise_10distinctobj_surf_agent,100.00,1.00,28,25.14,3,20
+randrot_10distinctobj_surf_agent,100.00,0.00,28,13.46,2,13
+randrot_noise_10distinctobj_5lms_dist_agent,100.00,0.00,52,49.54,5,40
 base_10simobj_surf_agent,92.14,10.71,83,11.00,7,34
-randrot_noise_10simobj_dist_agent,78.00,31.00,196,33.98,11,92
-randrot_noise_10simobj_surf_agent,87.00,35.00,174,38.66,17,140
-randomrot_rawnoise_10distinctobj_surf_agent,62.00,77.00,13,102.96,4,5
+randrot_noise_10simobj_dist_agent,78.00,36.00,201,21.46,11,92
+randrot_noise_10simobj_surf_agent,90.00,32.00,166,27.15,17,140
+randomrot_rawnoise_10distinctobj_surf_agent,65.00,83.00,13,101.60,4,5
 base_10multi_distinctobj_dist_agent,85.00,12.86,27,25.30,3,1
--- a/benchmarks/ycb_77objs.csv
+++ b/benchmarks/ycb_77objs.csv
@@ -1,6 +1,6 @@
 Experiment,Correct (%)|align right,Used MLH (%)|align right,Num Match Steps|align right,Rotation Error (degrees)|align right,Run Time (mins)|align right,Episode Run Time (s)|align right
 base_77obj_dist_agent,90.91,12.99,102,14.89,20,58
 base_77obj_surf_agent,98.70,5.19,53,6.70,14,35
-randrot_noise_77obj_dist_agent,91.34,31.60,190,35.54,41,137
-randrot_noise_77obj_surf_agent,97.40,22.94,109,41.63,36,115
-randrot_noise_77obj_5lms_dist_agent,94.81,1.30,77,54.50,13,119
+randrot_noise_77obj_dist_agent,92.21,26.84,174,34.01,41,137
+randrot_noise_77obj_surf_agent,94.37,21.65,111,33.90,36,115
+randrot_noise_77obj_5lms_dist_agent,93.51,0,72,56.34,13,119


src/tbp/monty/frameworks/experiments/seed.py
```diff
@@ -0,0 +1,32 @@
...
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+import hashlib
+
+
+def episode_seed(seed: int, epoch: int, episode: int) -> int:
+    """Generate a seed for an episode.
+
+    In some cases, we want a deterministic random for an episode,
```
jeremyshoemaker: The phrase "a deterministic random" sounds incomplete. Did you mean something like the Random class, or just a random number generator in general?

jeremyshoemaker (approved):

src/tbp/monty/frameworks/experiments/seed.py
tristanls-tbp: Ah, good call out.

🤔

In some cases, for each episode, we want to deterministically modify the experiment's random seed based on the epoch and episode. We don't want to start with the same experiment random seed for each episode. (...)

tristanls-tbp: I updated the docstring. Please let me know if it makes sense now.

tristanls-tbp:
> I'm sorry if the answer is obvious, but what's the reason that we now pass seed separately from rng (which was also initialized with the seed)? Couldn't we initialize the rng at the experiment level with a specific seed every episode and then only pass the rng to the other classes? Or does that not work bc. the rng is only passed at init?

Hi, sorry, I could have explained the reasoning better.

Object init samplers / object initializers are special.

As we know, once we initialize an RNG, it will output a deterministic sequence of randoms. For example:
```python
>>> rng = np.random.RandomState(42)
>>> rng.random()
0.3745401188473625
>>> rng.random()
0.9507143064099162
```
And when we reset it, it starts over
```python
>>> rng = np.random.RandomState(42)
>>> rng.random()
0.3745401188473625
```
The problem of our own creation is that it is difficult to guarantee that `run.py` and `run_parallel.py` will invoke the RNG in the same sequence. What we specifically want is for when the code reaches the object initializer, for the RNG to have been called with precisely the same sequence since it was reset at the beginning of the episode. This way, when the object initializer invokes the RNG, it will get the same value when started from `run.py` and from `run_parallel.py`.

On the one hand, in `run.py` we have the following code:
```python
with experiment as exp:
    ...
```
The `with experiment as exp` runs the following code:
```python
    def __enter__(self) -> Self:
        """Context manager entry method.

        Returns:
            MontyExperiment self to allow assignment in a with statement.
        """
        self.setup_experiment(self.config)
        return self
```
And then `self.setup_experiment(self.config)` does the following:
```python
    def setup_experiment(self, config: dict[str, Any]) -> None:
        """Set up the basic elements of a Monty experiment and initialize counters.

        Args:
        """
        self.init_loggers(self.config["logging"])
        logger.info(self.config)
        self.model = self.init_model(
            monty_config=config["monty_config"],
            model_path=self.model_path,
        )
        self.load_environment_interfaces(config)
        self.init_monty_data_loggers(self.config["logging"])
        self.init_counters()
```
Object initializers access RNG inside `self.load_environment_interfaces(config)`.

`self.init_model` occurs before, and it passes the RNG to learning modules, sensor modules, and the motor system policy. Then, it initializes the Monty model. Only after all that is done, and whatever calls all of that made to RNG, only then is `self.load_environment_interfaces` invoked.

Now, usually, this presents no problems, except for how we have `run_parallel.py` working.

Consider `run_parallel.generate_parallel_eval_configs`:
```python
def generate_parallel_eval_configs(
    experiment: DictConfig,
    name: str,
) -> list[Mapping]:
    """Generate configs for evaluation episodes in parallel.

    Create a config for each object and rotation in the experiment. Unlike with parallel
    training episodes, a config is created for each object + rotation separately.

    Args:
        experiment: Config for experiment to be broken into parallel configs.
        name: Name of experiment.

    Returns:
        List of configs for evaluation episodes.
    """
    sampler = hydra.utils.instantiate(
        experiment.config["eval_env_interface_args"]["object_init_sampler"]
    )
    sampler.rng = np.random.RandomState(experiment.config["seed"])
    object_names = experiment.config["eval_env_interface_args"]["object_names"]
    # sampler_params = sampler.all_combinations_of_params()

    new_experiments = []
    epoch_count = 0
    episode_count = 0
    n_epochs = experiment.config["n_eval_epochs"]

    params = sample_params_to_init_args(sampler())

    # ...
            new_experiment["config"]["eval_env_interface_args"].update(
                object_names=[obj],
                object_init_sampler=Predefined(**params),
            )
```
This means that we seed the RNG from the configuration seed and immediately invoke it. **This happens before we instantiate any experiments.**

Object initializers having this special relationship in `run_parallel.py` case, where they invoke RNG before the experiment even exists, is the reason why they always accept configuration seed instead of an already created RNG. I don't think there is a robust way of ensuring that RNG will be called precisely **zero** times before reaching an object initializer during `run.py`. 

Hence, the approach of always ensuring that we execute the `run_parallel.py` case, i.e., 1) instantiate from experiment configuration seed, 2) invoke right away. Now, as soon as I did this, I realized that every episode would always get the same object init parameters. This is what caused me to mix in the epoch and episode number before seeding the RNG. But the approach was first to ensure `run_parallel.py` use case is used in both `run_parallel.py` and `run.py` case, and then having to mix in epoch and episode numbers to ensure that we get different object params per episode.

> Also, if the benchmarks changed, shouldn't the respective .csv files also be updated?

My mistake, I hit the "Ready for review" button too soon. I'll add these.

PR #638, opened 2025-12-15
refactor: enable ruff perf402, list comprehension to copy

carver:
I'm not *wild* about double inline `for` loops. The convention always seemed backwards to me. But I don't see anything better.


src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -229,10 +229,7 @@ def get_unique_euler_poses(poses):
     Returns:
         unique_poses: array of unique poses
     """
-    all_poses = []
-    for path_poses in poses:
-        for pose in path_poses:
-            all_poses.append(pose)
+    all_poses = [pose for path_poses in poses for pose in path_poses]
```
jeremyshoemaker, line 232: I can't remember off the top of my head, but does this work with the `for`s swapped? It reads backwards to me at the moment, but I could probably count on one hand the number of times I've written a comprehension with double loops.

Like this:
```suggestion
    all_poses = [pose for pose in path_poses for path_poses in poses]
```

carver: Nope, the outer for loop has to go on the left. As I said above, I agree with you that Python's convention here feels backwards!

```python
In [68]: poses = [[1,], [2, 3]]

In [69]: [pose for pose in path_poses for path_poses in poses]
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Cell In[69], line 1
----> 1 [pose for pose in path_poses for path_poses in poses]

NameError: name 'path_poses' is not defined

In [70]: [pose for path_poses in poses for pose in path_poses]
Out[70]: [1, 2, 3]
```

PR #645, opened 2025-12-18
refactor!: formalize Transform protocol and expect TransformContext

tristanls-tbp:
The purpose of this pull request is to reduce the number of places where we would need to reset RNG before each episode. (Resetting RNG before each episode is follow-on work to replace the follow-up configs functionality)

Currently, when transforms are handed to an `EnvironmentInterface` on `EnvironmentInterface.__init__`, if they need RNG, the `EnvironmentInterface` sets the RNG on the transform. If we now want to reset RNG before every episode, then we need to reach into each transform and update the RNG. 

This pull request avoids having to reach into transforms on each episode by passing RNG on each transform invocation.

The `Transform` protocol is reified.

The `Transform.__call__` method is generic across all transforms.

Each transform then implements its own `call` method and uses `__call__` to adapt the generic interface to its own particular needs.

The reason for wrapping `call`s in `__call__` is because in some cases, like `src/tbp/monty/frameworks/environments/two_d_data.py`, we know the specific transform we created and we don't want to create unnecessary generic context to invoke that transform.

### DIFF
--- a/docs/future-work/learning-module-improvements/use-off-object-observations.md
+++ b/docs/future-work/learning-module-improvements/use-off-object-observations.md
@@ -6,7 +6,7 @@ estimated-scope: medium
 improved-metric: numsteps
 output-type: analysis, PR
 skills: python, research, monty, refactoring
-contributor: hlee
+contributor: hlee9212
 status: scoping
 ---
 
--- a/docs/future-work/learning-module-improvements/use-out-of-model-movements.md
+++ b/docs/future-work/learning-module-improvements/use-out-of-model-movements.md
@@ -6,7 +6,7 @@ estimated-scope: medium
 improved-metric: accuracy
 output-type: experiments, analysis, PR
 skills: python, research, monty
-contributor: hlee
+contributor: hlee9212
 status: paused
 ---
 
--- a/docs/future-work/sensor-module-improvements/extract-better-features.md
+++ b/docs/future-work/sensor-module-improvements/extract-better-features.md
@@ -6,7 +6,7 @@ estimated-scope: medium
 improved-metric: noise, accuracy, numsteps
 output-type: monty-feature, PR
 skills: python, research, computer-vision
-contributor: 
+contributor: tcdubs, ryan9186, roccolep
 status: evergreen
 ---
 
--- a/docs/theory/theory-overview.md
+++ b/docs/theory/theory-overview.md
@@ -1,5 +1,5 @@
 ---
-title: Neuroscience Theory - Overview
+title: Theory Overview
 description: Intro page to the underlying theory of this project.
 ---
 
--- a/src/tbp/monty/__init__.py
+++ b/src/tbp/monty/__init__.py
@@ -8,4 +8,4 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
-__version__ = "0.14.1"
+__version__ = "0.15.0"


src/tbp/monty/frameworks/environment_utils/transforms.py
```diff
@@ -26,10 +28,36 @@
...
+    """A transform that can be applied to observations."""
+
+    def __call__(
+        self, observations: Observations, ctx: TransformContext
+    ) -> Observations:
+        """Apply the transform to the observations.
+
+        Args:
+            observations: Observations to modify in place.
+            ctx: The transform context.
```
vkakerbeck, line 52: nit: is this a commonly known way to abbreviate context? Why not just call it `context`?

src/tbp/monty/frameworks/environment_utils/transforms.py
```diff
@@ -49,26 +77,29 @@ def __init__(self, agent_id: AgentID, max_depth, threshold=0):
...
         self.threshold = threshold
-        self.needs_rng = False
 
-    def __call__(self, observation, _state=None):
+    def __call__(
+        self, observations: Observations, _ctx: TransformContext
+    ) -> Observations:
+        return self.call(observations)
+
+    def call(self, observations: Observations) -> Observations:
```
vkakerbeck, line 86: curiosity: what's the reason we have a separate `call` function here instead of just executing this code in `__call__` like we do in other transforms? Is this bc this call doesn't need context? But aren't we passing context anyways to all transforms now bc of the protocol?

src/tbp/monty/frameworks/environment_utils/transforms.py
```diff
@@ -81,38 +112,39 @@ def __init__(self, agent_id: AgentID, sigma):
         """
         self.agent_id = agent_id
         self.sigma = sigma
-        self.needs_rng = True
 
-    def __call__(self, observation, _state=None):
+    def __call__(
```
vkakerbeck, line 116: curiosity: why aren't we wrapping a `call` function into this one? That way we could just pass `ctx.rgn` as the rng parameter (`ctx.rng.normal` looks a bit confusing)

tristanls-tbp:
> I feel like this needs to be documented somewhere, but I'm not sure where. The call method isn't part of the Transform protocol, so it probably shouldn't go there. I think we'll forget this detail if it isn't kept somewhere in the code comments.

It's not a rule to follow, so I don't think it should be documented. For example, `AddNoiseToRawDepthImage` does not have a `call` because it is never called by itself in a random place, and it doesn't need to alter `__call__` signature. What the non-protocol shape of the transform is depends on the use case.

src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -562,7 +562,7 @@ def get_3d_scene_point_cloud(self):
         # Apply gaussian smoothing transform to depth image
         # Uncomment line below and add import, if needed
         # transform = GaussianSmoothing(agent_id=agent_id, sigma=2, kernel_width=3)
-        # obs = transform(obs, state=state)
+        # obs = transform.call(obs)
```
vkakerbeck, line 565: I know this is just a comment, but wouldn't we have to pass context here too?

vkakerbeck (approved):
Looks good, just a few small questions but nothing blocking

src/tbp/monty/frameworks/environment_utils/transforms.py
tristanls-tbp: Yes, it is a common way to abbreviate it. e.g., [Go code](https://pkg.go.dev/context), [pytorch](https://github.com/search?q=repo%3Apytorch%2Fpytorch%20ctx&type=code), [flask](https://tedboy.github.io/flask/_modules/flask/ctx.html), etc.

src/tbp/monty/frameworks/environments/two_d_data.py
tristanls-tbp: The `GaussianSmoothing.call` signature is:
```python
class GaussianSmoothing(Transform):
    def call(self, observations: Observations) -> Observations:
```

src/tbp/monty/frameworks/environment_utils/transforms.py
tristanls-tbp: > Is this bc this call doesn't need context?

This is the reason.

> But aren't we passing context anyways to all transforms now bc of the protocol?

Yes, we are passing context to all the transforms when we invoke them in `embodied_data.py`, e.g.,:
```python
    def apply_transform(self, transform, observation, state):
        ctx = TransformContext(rng=self.rng, state=state)
        if isinstance(transform, Iterable):
            for t in transform:
                observation = t(observation, ctx)
        else:
            observation = transform(observation, ctx)
        return observation
```
However, in this case, all we know is that we have a list of `Transform`s, which require `TransformContext` to be passed.

However, we have use cases like in `two_d_data.py`, where know which specific transform we have, so we can only provide enough parameters to the transform we use, and this is what `call` is for, so that we don't have to artificially create a `TransformContext` only to throw it away.

```python
        transform = DepthTo3DLocations(
            agent_id=agent_id,
            # ...
            clip_value=1.1,
        )
        obs_3d = transform.call(obs, state=state)
```

src/tbp/monty/frameworks/environment_utils/transforms.py
tristanls-tbp: We could. But as I mentioned to @jeremyshoemaker, it's not really a necessary pattern. I'm familiar with accessing things off of `ctx`, so it seemed fine to me at the time.

But, since both of you mentioned it, I made the update for this transform to also delegate to `call`.

PR #647, opened 2025-12-18
refactor: Address typing issues stemming from `np.quaternion` and `numbers.Number`

mf-maymay:
## Overview

This PR reduces existing mypy errors and removes a few typing pain points, with the longer-term goal of making it feasible to run `mypy src` reliably during development and in CI.

The changes introduced in this PR bring our mypy complaints (in `src`) down from `128 errors in 30 files` to `121 errors in 28 files`.

### `np.quaternion`

The `quaternion-numpy` package [monkey patches numpy on import](https://github.com/moble/quaternion/blob/8658799289278aa4f7e2bdcf65660790300d0234/src/quaternion/__init__.py#L49), adding `quaternion` to the `numpy`/`np` namespace. Because this happens at import time, mypy is not aware of the change and does not accept the reference. This PR updates these references to point directly to `quaternion.quaternion`. (`quaternion.quaternion` does not have useful type information, unfortunately, but it is at least a valid reference.)

> ℹ️ Note: Some of the existing references to `np.quaternion` did not result in mypy complaints because the functions containing them were not annotated. Mypy only inspects functions with annotations.

Also, I noticed that we have a convention of calling the `quaternion` package `qt` ([it's even listed as an auto-import common statement](https://github.com/thousandbrainsproject/tbp.monty/blob/0506f367c20f6ede583c40a3d04e4ce9c4494c57/pyproject.toml#L151)), so I updated imports and references to standardize on that convention. (`qt` typically refers to a totally different package in Python, but it should be clear within context that we are not using it for graphics.)

## `numbers.Number`

Believe it or not, `numbers.Number` does not play well with ints or floats in static type checks. Here is a mypy comment explaining that the mypy maintainers do not intend to add better support for this.

As such, this PR removes references to `numbers.Number`, replacing the `Number` hints with `float`. (`float` is considered a supertype of `int` [as far as type checkers are concerned](https://typing.python.org/en/latest/spec/special-types.html#special-cases-for-float-and-complex).)

### DIFF
--- a/src/tbp/monty/__init__.py
+++ b/src/tbp/monty/__init__.py
@@ -8,4 +8,4 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
-__version__ = "0.15.0"
+__version__ = "0.16.0"
--- a/src/tbp/monty/frameworks/models/motor_system_state.py
+++ b/src/tbp/monty/frameworks/models/motor_system_state.py
@@ -22,9 +22,9 @@ class SensorState(TypedDict):
     """
 
     position: Any  # TODO: Stop using magnum.Vector3 and decide on Monty standard
-    """The sensor's position relative to the agent."""
+    """The sensor's position."""
     rotation: Any  # TODO: Stop using quaternion.quaternion and decide on Monty standard
-    """The sensor's rotation relative to the agent."""
+    """The sensor's rotation."""
 
 
 class AgentState(TypedDict):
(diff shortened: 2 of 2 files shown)


goingmaymay:
I've been a bit quieter recently due to managing multiple releases for work and a cross-state move, but I'm excited to be back and contributing more regularly again. 💪

goingmaymay:
@jeremyshoemaker @scottcanoe Thank you! I have addressed the merge conflict.

I also prefer `quaternion` to `qt`, actually, but I saw strong evidence that `qt` was the prevailing convention here:
- It's used in documentation (in an RFC):
  https://github.com/thousandbrainsproject/tbp.monty/blob/e7006fab6d05b8424b530eeefc8e4132bdddb8ab/rfcs/0004_action_object.md?plain=1#L104-L109
- There's an "autoimport" statement targeting the alias in the TOML:
  https://github.com/thousandbrainsproject/tbp.monty/blob/e7006fab6d05b8424b530eeefc8e4132bdddb8ab/pyproject.toml#L151

Using `quaternion` instead of `qt` is certainly still on the table. (There's no lint rule forbidding it.) If we ever standardize on `quaternion`, I'd suggest that we update the places listed above.

Thanks! Happy holidays! 🎄🎄🎄

PR #651, opened 2025-12-18
refactor: update variable invalid_signal->valid_signal

carver:
`invalid_signal` sets you up for awkward double-negative expressions, like:

```py
if not invalid_signal:
  use_good_signal()
```
So flipping the meaning (and name) to `valid_signal` is an improvement.

This came up during a call about off-object observations.

## Implementation notes

I `grep`ed for `_extract_and_add_features()`, and it's only used in this one place.

This does change the function signature (the meaning of the returned value), but in a "private" method, so I didn't consider it a breaking change. But if you consider these `_`-prefixed methods as part of the public API, then it should be a `refactor!:`.


carver:
A quick cleanup that came up during our call today cc @vkakerbeck @nielsleadholm

carver:
Do you want me to run one and report back? Any preference of which one?

I am reasonably confident that I can prove that this change is logically consistent. But of course I would never advocate for skipping testing

carver:
This is the result after running on the branch with: `./run.py experiment=randrot_noise_77obj_surf_agent`
[eval_stats.csv](https://github.com/user-attachments/files/24562185/eval_stats.csv)

It takes a good amount of time to run. I don't really know how to eyeball to decide if it looks right. I'm re-running against the parent commit now, and I'll look for differences. Unfortunately, my out-of-memory killer shut off the experiment during a couple of previous attempts, and the full run takes many hours, so I'm not sure how long until I have the results.

I can't tell yet whether Monty is itself responsible for a spike in memory usage mid-run. Strangely, when I inspect memory during a run manually, it only seems to use 3.5GB. So I guess I'll assume not for now.

carver:
> I can't tell yet whether Monty is itself responsible for a spike in memory usage mid-run. Strangely, when I inspect memory during a run manually, it only seems to use 3.5GB. So I guess I'll assume not for now.

Ah, looks like it **was** a mid-run spike. After 8.5 hours, in the span of 90 seconds, the experiment process went from using 3.5GB to 10.3GB of RAM. At that point, my out-of-memory killer terminated the process. Not sure how high it would have maxed out at. I'll try running it again later when there is more free memory to spare.

vkakerbeck:
Ok so the last remaining thing to change is to update the copyright notice on the first line of this file to
`# Copyright 2025-2026 Thousand Brains Project` (i.e. adding `-2026`).

carver:
Great! Looks all set now. :)

PR #652, opened 2025-12-18
refactor!: pass RNG to DefaultMessageNoise during call, not init

tristanls-tbp:
This pull request continues the effort to remove RNG from the internal state of objects that do not participate in the `pre_episode` or `reset` protocol/functionality. I am working toward a point where I can use `pre_episode` and/or `reset` to reset the RNG with the configured seed before every episode, thereby eliminating the need for the follow-on configs feature.

Function `no_message_noise` is turned into an equivalent callable class, to facilitate the next step (different PR) of injecting an instance of MessageNoise when constructing a Sensor Module, instead of passing in `noise_params` and the sensor module having to construct the MessageNoise itself.

**note**: You'll start noticing the pattern here, that I am passing RNG in method calls instead of constructors. For this go at it, I will stop refactoring this way once I reach something that has `pre_episode` and/or `reset`. However, if you think about the future when we extract the experimental framing from the platform, we'll get rid of these `pre_episode` methods, and then this pattern of passing in RNG during a call will likely migrate into the `step()` calls, which will then require RNG. I think this makes sense because the future Experiment will be the only thing that knows when an episode starts, so it will reset its RNG appropriately and then call the Platform `step(rng)`, which won't know anything about episodes.

### DIFF
--- a/conf/experiment/config/logging/eval.yaml
+++ b/conf/experiment/config/logging/eval.yaml
@@ -4,10 +4,8 @@ defaults:
 output_dir: ${path.expanduser:${oc.env:MONTY_LOGS}/projects/feature_eval_runs/logs}
 monty_handlers:
   - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.BasicCSVStatsHandler}
-  # - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.DetailedJSONHandler}
   - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.ReproduceEpisodeHandler}
 wandb_handlers:
   - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbTableStatsHandler}
   - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbChartStatsHandler}
-  # - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.DetailedWandbMarkedObsHandler}
 wandb_group: "gm_eval_runs"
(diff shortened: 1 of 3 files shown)


src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -440,24 +440,24 @@ def pre_episode(self):
...
 
 class MessageNoise(Protocol):
-    def __call__(self, state: State) -> State: ...
+    def __call__(self, state: State, rng: np.random.RandomState) -> State: ...
 
 
-def no_message_noise(state: State) -> State:
-    """No noise function.
+class NoMessageNoise(MessageNoise):
+    def __call__(self, state: State, rng: np.random.RandomState) -> State:  # noqa: ARG002
```
tristanls-tbp, line 447: **note**: If I were to rename `rng` to `_rng` (instead of `# noqa: ARG002`), then the following call would fail:
```python
observed_state = self._message_noise(observed_state, rng=self._rng)
```
because `NoMessageNoise` would not have an `rng` argument (as it would have `_rng` argument).

jeremyshoemaker: I'm generally not a fan of changing arguments to be prefixed with an underscore in Python when not used for this exact reason.

jeremyshoemaker (approved):

src/tbp/monty/frameworks/models/sensor_modules.py
tristanls-tbp: Yup, I will stop doing it :)

jeremyshoemaker: We might want to consider purposefully disabling ARG002. It leads to these all the time, and I don't completely agree with it as a lint. Just because it exists doesn't mean we have to use it.

tristanls-tbp: One thing I like about having to put in `# noqa: ARG002` is that it allows me to say, "I'm doing this on purpose." However, if I do it by accident and don't notice, the linter won't notify me if we disable the rule.

vkakerbeck (approved):
LGTM

PR #653, opened 2025-12-18
refactor!: pass RNG to ActionSampler during call, not init.

tristanls-tbp:
This pull request is the last in the series of pull requests (for now) that change the code to pass RNG during a call instead of `__init__` (also see #652 and #645).

The reason we are doing this is to be able to reset the RNG before every episode. To perform that reset, we will call `pre_episode(rng)` on a few things to reset the RNG accordingly. By passing RNG into ActionSampler calls, we avoid having to write `pre_episode` code for every component that uses RNG. As we want to eventually remove `pre_episode` code from the Platform, not adding any `pre_episode` code is desired.

### DIFF
--- a/conf/experiment/config/logging/eval.yaml
+++ b/conf/experiment/config/logging/eval.yaml
@@ -4,10 +4,8 @@ defaults:
 output_dir: ${path.expanduser:${oc.env:MONTY_LOGS}/projects/feature_eval_runs/logs}
 monty_handlers:
   - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.BasicCSVStatsHandler}
-  # - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.DetailedJSONHandler}
   - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.ReproduceEpisodeHandler}
 wandb_handlers:
   - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbTableStatsHandler}
   - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbChartStatsHandler}
-  # - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.DetailedWandbMarkedObsHandler}
 wandb_group: "gm_eval_runs"
--- a/conf/experiment/config/logging/eval_evidence_lm.yaml
+++ b/conf/experiment/config/logging/eval_evidence_lm.yaml
@@ -8,6 +8,5 @@ monty_handlers:
 wandb_handlers:
   - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbTableStatsHandler}
   - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbChartStatsHandler}
-  # - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.DetailedWandbMarkedObsHandler}
 wandb_group: "evidence_eval_runs"
 monty_log_level: "BASIC"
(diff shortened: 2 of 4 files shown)


src/tbp/monty/frameworks/actions/action_samplers.py
```diff
@@ -94,77 +92,92 @@ def __init__(
...
+        super().__init__(actions=actions)
         self.absolute_degrees = absolute_degrees
         self.direction = direction if direction is not None else (0.0, 0.0, 0.0)
         self.location = location if location is not None else (0.0, 0.0, 0.0)
         self.rotation_degrees = rotation_degrees
         self.rotation_quat = rotation_quat if rotation_quat is not None else qt.one
         self.translation_distance = translation_distance
 
-    def sample_look_down(self, agent_id: AgentID) -> LookDown:
+    def sample_look_down(self, agent_id: AgentID, _rng: RandomState) -> LookDown:
```
jeremyshoemaker: Technically this is changing the signature of this method. If I loaded this up in PyCharm it will complain to me about it. It's specifically because code expecting the arguments from the protocol calling this with keyword arguments will fail because the name changed.

tests/unit/frameworks/actions/action_samplers_test.py
```diff
@@ -23,11 +23,15 @@
 from tbp.monty.frameworks.actions.actions import Action, QuaternionWXYZ, VectorXYZ
 from tbp.monty.frameworks.agents import AgentID
 
+AGENT_ID_1 = AgentID("agent1")
+AGENT_ID_2 = AgentID("agent2")
 RNG_SEED = 1337
```
jeremyshoemaker, line 28: 🙄

tests/unit/frameworks/actions/action_samplers_test.py
```diff
@@ -23,11 +23,15 @@
 from tbp.monty.frameworks.actions.actions import Action, QuaternionWXYZ, VectorXYZ
 from tbp.monty.frameworks.agents import AgentID
 
+AGENT_ID_1 = AgentID("agent1")
+AGENT_ID_2 = AgentID("agent2")
```
jeremyshoemaker, line 27: I appreciate these being factored out while you're here.

tests/unit/frameworks/actions/actions_test.py
```diff
@@ -51,12 +55,13 @@ def test_agent_id_returns_configured_agent_id(self) -> None:
...
     def setUp(self) -> None:
         self.action = LookDown(agent_id=AgentID("test"), rotation_degrees=47)
+        self.rng = RandomState(RNG_SEED)
 
     def test_delegates_to_sample_look_down(self) -> None:
         sampler = Mock()
         sampler.sample_look_down = Mock()
-        LookDown.sample(agent_id=AgentID("test"), sampler=sampler)
-        sampler.sample_look_down.assert_called_once_with("test")
+        LookDown.sample(agent_id=AgentID("test"), sampler=sampler, rng=self.rng)
```
jeremyshoemaker: You _could_ factor out the `AgentID("test")` while you're here, since you did something similar in the other test.

jeremyshoemaker (approved):
Looks fine to me.

src/tbp/monty/frameworks/actions/action_samplers.py
tristanls-tbp: I'll update this to `# noqa: ARG002`. I prefer not to prefix `_` now that I have found where it hurts.

tests/unit/frameworks/actions/action_samplers_test.py
tristanls-tbp: 😛

tests/unit/frameworks/actions/actions_test.py
tristanls-tbp: Yeah, I noticed that too. I'll update.

tests/unit/frameworks/actions/action_samplers_test.py
vkakerbeck: What happened to 42? :D

tristanls-tbp: 1337 is more eleet 😬

PR #654, opened 2025-12-19
feat!: MontyExperiment.run()

tristanls-tbp:

This pull request adds `MontyExperiment.run()`. Experiments are fully capable of knowing when to run their own `self.train()` and `self.evaluate()`. This is because every experiment receives a configuration that has `do_eval` and `do_train` fields. Experiments should handle any complexity related to mode selection and tracking.

Implementation is straightforward:
```python
    def run(self):
        """Run the experiment."""
        if self.do_train:
            self.train()

        if self.do_eval:
            self.evaluate()
```

This simplifies the external API to the usual case being:
```python
    with experiment:
        experiment.run()
```
The current detailed access to `.train()` and `.evaluate()` are replaced via configuration. That is, instead of doing training-only like this:
```python
exp = hydra.utils.instantiate(experiment)
with exp:
    exp.train()
```
Make sure the config is a training config:
```yaml
# ...
config:
  do_eval: false
  do_train: true
  # ...
```
And do:
```python
exp = hydra.utils.instantiate(experiment)
with exp:
    exp.run()
```
Similarly, for evaluation-only.

Overall, the idea is to configure the experiment like we mean it so that `Experiment.run()` does the correct thing. We should not be cherry-picking `exp.train()` or `exp.evaluate()` because we don't want to change the config.

## Why do this change now?

I am working on resetting RNG before every episode with episode-specific RNG seeds. As I was looking at the distinction between training and evaluation epochs and episodes, I noticed an inversion of control, which complicates my work a bit. By inversion of control, I mean that data that `MontyExperiment` should be in charge of and control is instead carried and looked up on the `Monty` model. Specifically:
```python
# MontyExperiment

    def train(self):
        """Run n_train_epochs."""
        self.logger_handler.pre_train(self.logger_args)
        self.model.set_experiment_mode("train")
        for _ in range(self.n_train_epochs):
            self.run_epoch()
        self.logger_handler.post_train(self.logger_args)

    def evaluate(self):
        """Run n_eval_epochs."""
        # TODO: check that number of eval epochs is at least as many as length
        # of environment interface number of rotations
        self.logger_handler.pre_eval(self.logger_args)
        self.model.set_experiment_mode("eval")
        for _ in range(self.n_eval_epochs):
            self.run_epoch()
        self.logger_handler.post_eval(self.logger_args)
```
Notice that `MontyExperiment` tells `Monty` model what experiment mode it ought to be in. But then, in other code, `MontyExperiment` uses `Monty` model to check what experiment mode it is in, e.g.,:
```python
        if self.model.experiment_mode != "train":
            self.max_steps = self.max_eval_steps

# ...

        mode = self.model.experiment_mode

        if mode == "train":
            epoch = self.train_epochs
            episode = self.train_episodes

# ...

        if self.model.experiment_mode == "train":
            self.train_episodes += 1
            self.total_train_steps += steps

# ...

        if self.model.experiment_mode != "train":
            self.env_interface = self.eval_env_interface

# ...

        if self.model.experiment_mode == "train":
            self.train_epochs += 1
            self.train_env_interface.post_epoch()
```
All of these should instead be `if self.experiment_mode` lookups on `MontyExperiment`. And once we have `self.experiment_mode`, then it is clear that `MontyExperiment` can handle `run()` by itself and do appropriate train vs evaluation switching based on its configuration.

Benchmarks are unchanged

Screenshot 2025-12-19 at 14 25 31

### DIFF
--- a/conf/experiment/config/logging/eval.yaml
+++ b/conf/experiment/config/logging/eval.yaml
@@ -4,10 +4,8 @@ defaults:
 output_dir: ${path.expanduser:${oc.env:MONTY_LOGS}/projects/feature_eval_runs/logs}
 monty_handlers:
   - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.BasicCSVStatsHandler}
-  # - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.DetailedJSONHandler}
   - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.ReproduceEpisodeHandler}
 wandb_handlers:
   - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbTableStatsHandler}
   - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbChartStatsHandler}
-  # - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.DetailedWandbMarkedObsHandler}
 wandb_group: "gm_eval_runs"
--- a/conf/experiment/config/logging/eval_evidence_lm.yaml
+++ b/conf/experiment/config/logging/eval_evidence_lm.yaml
@@ -8,6 +8,5 @@ monty_handlers:
 wandb_handlers:
   - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbTableStatsHandler}
   - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbChartStatsHandler}
-  # - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.DetailedWandbMarkedObsHandler}
 wandb_group: "evidence_eval_runs"
 monty_log_level: "BASIC"
--- a/src/tbp/monty/frameworks/actions/action_samplers.py
+++ b/src/tbp/monty/frameworks/actions/action_samplers.py
@@ -13,7 +13,7 @@ from typing import Callable
 
 import quaternion as qt
 from numpy import cos, pi, sin, sqrt
-from numpy.random import Generator, default_rng
+from numpy.random import RandomState
 
 from tbp.monty.frameworks.actions.actions import (
     Action,
@@ -48,10 +48,8 @@ class ActionSampler:
 
     def __init__(
         self,
-        rng: Generator | None = None,
         actions: list[type[Action]] | None = None,
     ):
-        self.rng: Generator = rng if rng is not None else default_rng()
         self._actions: list[type[Action]] = actions if actions is not None else []
         self._action_names: list[str] = [
             action.action_name() for action in self._actions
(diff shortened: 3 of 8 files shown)


src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -441,16 +452,14 @@ def init_monty_data_loggers(self, logging_config: dict[str, Any]) -> None:
         )
 
     def get_epoch_state(self):
-        mode = self.model.experiment_mode
-
-        if mode == "train":
+        if self.experiment_mode == ExperimentMode.TRAIN:
```
jeremyshoemaker: It isn't critical to change this since it works with `==`, but Enum values should be compared by identity using `is`. The check will be faster since `is` only looks at whether the two variables reference the same thing. Using `==` I think someone could still use a raw string when setting it and it would still work. See https://docs.python.org/3/howto/enum.html#comparisons
```suggestion
        if self.experiment_mode is ExperimentMode.TRAIN:
```

tristanls-tbp: Thank you, will update.

jeremyshoemaker: I could have sworn there were more to change than that, but I must have been remembering all the assignments.

tristanls-tbp: Yeah, there's a bunch of assignments, in tests especially.

src/tbp/monty/frameworks/run.py
```diff
@@ -48,15 +48,7 @@ def main(cfg: DictConfig):
     output_dir.mkdir(exist_ok=True, parents=True)
     experiment = hydra.utils.instantiate(cfg.experiment)
     start_time = time.time()
-    with experiment as exp:
-        # TODO: Later will want to evaluate every x episodes or epochs
-        # this could probably be solved with just setting the logging freqency
-        # Since each trainng loop already does everything that eval does.
-        if exp.do_train:
-            print("---------training---------")
```
vkakerbeck, line 56: Could we add these prints back into the respective train and eval calls (maybe using the logger instead)? It seems like useful info to let the user know what just started

tristanls-tbp: I thought that, in yesterday's meeting, we discussed that the case where someone is doing training followed by evaluation in a single experiment almost never happens. Am I misremembering?

With experiments being only training or only evaluation (we also discussed how neither term is a great description for what Monty is doing), will the user be confused if these are omitted?

tristanls-tbp: Would a generic episode start marker be more appropriate?

vkakerbeck: Yes, I was mostly thinking of a way to tell the user that the experiment started so I'd be happy with just "Experiment running" or similar. I agree they basically never run one after the other (except in some unit tests) but it may still clarify things for the user to be more specific without having to look into the config. But its more of a bonus info so I am also fine if we don't add that.

tristanls-tbp: @vkakerbeck how's [this](https://github.com/thousandbrainsproject/tbp.monty/pull/654/changes/58d6fa32ddfdaf57f48975a64fcbe91becbe2ccc)?

vkakerbeck: That's great, thank you!

PR #656, opened 2025-12-20
refactor: remove `run_name` placeholder from logging configs

ramyamounir:
I have been trying to find an easy way to make any experiment use the detailed logger by overriding it in CLI. This is supposed to be straightforward with Hydra and I came up with this command below to do it:
```zsh
python run_parallel.py experiment=[exp_name] +experiment/config/logging=defaults
```
One problem is that the configs under `conf/experiment/config/logging/defaults.yaml` define an empty `run_name=""` field as a placeholder. So when I try to merge defaults back as an override, it overwrites the `run_name` field that was set by the experiment to an empty string and dumps the results in the parent logs directory. I don't see a reason for why we would need this placeholder in defaults. Every experiment should merge in its `run_name` at the experiment level. Am I missing something here?

I also realized that small changes like these will make the snapshot tests fail. But it seems that the tests are failing because the order of the fields have changed (e.g., `run_name` moved to the end). I think the snapshot testing is doing a string comparison. I was thinking it would be much more useful if we can make it smart enough to be invariant to the order of the fields inside the config groups when making the comparisons? This way it fails when there is a real mismatch, otherwise it's just too noisy to be useful.

I have not yet run the new snapshots, will do before merging if approved.

### DIFF
--- a/tests/conf/snapshots/base_10multi_distinctobj_dist_agent.yaml
+++ b/tests/conf/snapshots/base_10multi_distinctobj_dist_agent.yaml
@@ -96,11 +96,11 @@ experiment:
       python_log_to_file: true
       python_log_to_stderr: true
       output_dir: ${path.expanduser:${oc.env:MONTY_LOGS}/projects/monty_runs}
-      run_name: base_10multi_distinctobj_dist_agent
       resume_wandb_run: false
       wandb_id:
         _target_: wandb.util.generate_id
       wandb_group: benchmark_experiments
+      run_name: base_10multi_distinctobj_dist_agent
     monty_config:
       motor_system_config:
         motor_system_args:
--- a/tests/conf/snapshots/base_10simobj_surf_agent.yaml
+++ b/tests/conf/snapshots/base_10simobj_surf_agent.yaml
@@ -96,11 +96,11 @@ experiment:
       python_log_to_file: true
       python_log_to_stderr: true
       output_dir: ${path.expanduser:${oc.env:MONTY_LOGS}/projects/monty_runs}
-      run_name: base_10simobj_surf_agent
       resume_wandb_run: false
       wandb_id:
         _target_: wandb.util.generate_id
       wandb_group: benchmark_experiments
+      run_name: base_10simobj_surf_agent
     monty_config:
       motor_system_config:
         motor_system_args:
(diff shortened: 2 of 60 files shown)


ramyamounir:
Thanks for your review @tristanls-tbp.

> Is an error raised if nothing provides the `run_name`?

Yes, an experiment has to provide the run_name to the logger, otherwise an error is raised. The empty string `run_name` doesn't solve the problem, it will put the results in the parent directory. I think this error raised is ideal to remind the user of the missing `run_name`.

```zsh
$ python run_parallel.py experiment=randrot_noise_10distinctobj_dist_agent "~experiment.config.logging.run_name"

omegaconf.errors.ConfigAttributeError: Key 'run_name' is not in struct
    full_key: experiment.config.logging.run_name
    object_type=dict
```

> Please ensure that the tests pass. Generally, we don't review PRs that don't pass tests. Changing the snapshot is part of the PR, so this PR is incomplete until all the changes are committed and snapshot tests are passing.

:+1:  Yes, makes sense. I will update the snapshots.

> "Small" here is relative.

Sorry, I wasn't very clear. I didn't mean small as in the change I made was fewer lines. I meant the functional change it had on the snapshots is small. All of them changed but the actual effect is small, functionally they are all still the same. The error message saying that the snapshot changed is not very useful because the configs will work exactly the same way if only the order of some lines change. 

> note (non-blocking): I think it is confusing that to turn on a detailed logger the override is called "defaults". I would think there'd be a config called "detailed" or something along those lines. But that's not blocking this PR.

I agree. I was working on simplifying the logging configs by refactoring basic and detailed handlers into different files, then importing them in the defaults of these loggers. Then I was going to create a detailed logger config. It just became tedious to manually compare the snapshots to the new configs, with every change, to make sure that the error is just due to the order of lines changing. I decided to first ask if it's a good idea to try and do the comparisons differently.

> Have you explored adding the detailed logger directly to the list? Something like +config.logging.monty_handlers=${monty.class:tbp.monty.frameworks.loggers.monty_handlers.DetailedJSONHandler} ?

Yes but I also had to change the `monty_log_level` from BASIC to DETAILED. It just felt like this would be better done by either creating a detailed logging override file, or using the existing `default.yaml` one temporarily.

ramyamounir:
Snapshots updated. All changes are intended.

ramyamounir:
Removing the `run_name` doesn't affect the experiments but it does affect the unit tests. Many unit tests such as [this one](https://github.com/thousandbrainsproject/tbp.monty/blob/07c89bc7492334f3b248bebe70dbbec04c0fdba5/tests/unit/base_config_test.py?plain=1#L176) don't define a `run_name`, so they fail. I will close this PR and create a new one with a detailed logger config file that doesn't have `run_name`.

PR #657, opened 2025-12-22
refactor: episode_spec is always a str; never None

ramyamounir:
With the Hydra change, it seems like it parses the argument `episodes="0"` and casts the numeric values to `int`. This causes the error below because the function expects a `str | None`:
```zsh
Error executing job with overrides: ['experiment=randrot_noise_10distinctobj_dist_agent', 'episodes=0']
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/run_parallel.py", line 722, in main
    eval_configs = filter_episode_configs(eval_configs, cfg.episodes)
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/run_parallel.py", line 246, in filter_episode_configs
    idxs = parse_episode_spec(episode_spec, len(configs))
  File "/home/ramy/tbp/tbp.monty/src/tbp/monty/frameworks/run_parallel.py", line 193, in parse_episode_spec
    s = episode_spec.strip().lower()
AttributeError: 'int' object has no attribute 'strip'
```
This PR casts the `episode_spec` value to always become a `str`, and removes the handling of the value `None`. It should never happen, since the default is `"all"` and overriding it will now becomes just a different `str`. Unit tests are also updated.

### DIFF
--- a/src/tbp/monty/frameworks/run_parallel.py
+++ b/src/tbp/monty/frameworks/run_parallel.py
@@ -156,7 +156,7 @@ def print_config(config: DictConfig) -> None:
     print("-" * 100)
 
 
-def parse_episode_spec(episode_spec: str | None, total: int) -> list[int]:
+def parse_episode_spec(episode_spec: str, total: int) -> list[int]:
     """Parses a zero-based episode selection string into episode indices.
 
     Converts a human-friendly selection string into a sorted list of unique,
@@ -188,8 +188,6 @@ def parse_episode_spec(episode_spec: str | None, total: int) -> list[int]:
     Raises:
         ValueError: If the selection contains any invalid index or range.
     """
-    if episode_spec is None:
-        return list(range(total))
     s = episode_spec.strip().lower()
     if s in ("", "all", ":"):
         return list(range(total))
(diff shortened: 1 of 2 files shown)


ramyamounir:
Closed because it seems better to just quote the values in cli. For example: `episodes=\'0\'` or `episodes=\'0,1\'`

The below was suggested by Hydra:
```zsh
Ambiguous value for argument 'episodes=0,1'
1. To use it as a list, use key=[value1,value2]
2. To use it as string, quote the value: key=\'value1,value2\'
3. To sweep over it, add --multirun to your command line
```

tristanls-tbp:
FWIW, I think `episodes=[0,1]` would be a better UI. This way, we pass in a list of ints, as opposed to having to escape a string, only to have it parsed into a list of ints.

ramyamounir:
Yes, a list would be less confusing, I think the slicing syntax would be slightly more useful in cases where we may want to select many episodes, for example `episodes=\':15\'` or `episodes=\':30,80:\'`

tristanls-tbp:
[I think you can reduce amount of escaping by quoting the whole parameter](https://hydra.cc/docs/advanced/override_grammar/basic/#quoted-values). e.g.,
```
$ python my_app.py '+foo="{a: 10}"'
foo: '{a: 10}'
```

PR #658, opened 2025-12-22
chore: add detailed logging configs

ramyamounir:
I was looking for a quick way to override the logging configs to become detailed. There exists a `conf/experiment/config/logging/defaults.yaml` file that contains detailed configs but also contains a placeholder `run_name=""`. Removing this placeholder causes issues with unit tests, see [here](https://github.com/thousandbrainsproject/tbp.monty/pull/656#issuecomment-3678244593) for more context.

This PR adds a detailed logging config that we can use with Hydra CLI overrides. I couldn't figure out an easier way than to just add a new file to override the existing logging config. Here is the command I would use to run a quick detailed experiment with a few select episodes:

```zsh
python run_parallel.py experiment=[experiment_name] episodes=\"0,1\" +experiment/config/logging=detailed_evidence_lm
```


ramyamounir:
Thanks @tristanls-tbp, I've added a unit test so that this file is always being tested and is not dead code. Can you take another look?

tests/unit/base_config_test.py
```diff
@@ -195,5 +195,50 @@ def test_logging_info_level(self) -> None:
...
+class DetailedLoggingConfigTest(unittest.TestCase):
+    def setUp(self):
+        self.output_dir = tempfile.mkdtemp()
+
+        with hydra.initialize(version_base=None, config_path="../../conf"):
+            self.cfg = hydra.compose(
+                config_name="test",
+                overrides=[
+                    "test=evidence_lm/base",
+                    "+test/config/logging=detailed_evidence_lm",
```
tristanls-tbp: **issue**: I don't think this works. The `+` syntax is merge syntax. I don't think it actually finds "detailed_evidence_lm". But, merging an empty logging config with a working logging config would result in a working logging config.

**note**: I think the test passes because you are using `evidence_lm/base`, which includes
```yaml
  - /experiment/config/logging/defaults@config.logging
```
which in turn already has the detailed logger
```yaml
monty_log_level: DETAILED
monty_handlers:
  - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.BasicCSVStatsHandler}
  - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.DetailedJSONHandler}
  - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.ReproduceEpisodeHandler}
wandb_handlers: []
python_log_level: INFO
python_log_to_file: true
python_log_to_stderr: true
output_dir: ${path.expanduser:${oc.env:MONTY_LOGS}/projects/monty_runs/}
run_name: ""
resume_wandb_run: false
wandb_id:
  _target_: wandb.util.generate_id
wandb_group: "debugging"
```

One way to verify is to see if the test passes if you remove `detailed_evidence_lm.yaml`, which I think the test will pass.

ramyamounir:
Yes, you are right, thanks for catching this sneaky bug. I have fixed it, can you take another look? :crossed_fingers:

tests/unit/base_config_test.py
```diff
@@ -195,5 +195,50 @@ def test_logging_info_level(self) -> None:
             self.assertTrue(warning_message in log)
 
 
+class DetailedLoggingConfigTest(unittest.TestCase):
```
tristanls-tbp: **suggestion (non-blocking)**: `DetailedEvidenceLMLoggingConfigTest` as this tests `detailed_evidence_lm.yaml` specifically.

PR #660, opened 2025-12-23
docs: clarify ProprioceptiveState docstrings regarding reference frames

tristanls-tbp:
In #624 we realized that the docstrings attached to `SensorState` and `AgentState` claim too much. They cannot guarantee what reference frame the `position` or `rotation` are in. 

This pull request clarifies that the specific reference frame claims are restricted only to when those poses (they should probably be encapsulated as "poses" in the future) are embedded in a specific way, e.g., `SensorState` in `AgentState.sensors[key]` or `AgentState` in `ProprioceptiveState[key]`.

FYI @AgentRev


scottcanoe (approved):
Fair point, LGTM.

AgentRev:
LGTM as well.

PR #666, opened 2025-12-30
refactor!: deterministically reset rng before every episode

tristanls-tbp:

This pull request implements per-episode determinism across `run.py`, `run_parallel.py`, and `run_parallel.py 'episodes="..."'` ways of starting an experiment. It makes follow-up configs obsolete, hence, they are removed.

The below screenshot captures `eval_stats.csv` when ran in three different ways:

top:
 ```
 python run_parallel.py experiment=randrot_noise_10distinctobj_surf_agent num_parallel=2 'episodes="33,34,35"'
 ```
middle:
```
python run_parallel.py experiment=randrot_noise_10distinctobj_surf_agent num_parallel=2
```
bottom:
```
python run.py experiment=randrot_noise_10distinctobj_surf_agent
```

The per-episode outputs (aside from timings) are all identical.

Screenshot 2025-12-31 at 13 32 07

A new group of tests, `test-integration-monty`, is added, and the reproducibility tests (both for training and evaluation) are placed there. This group runs after `test-monty` completes successfully and runs in parallel with `test-tutorials-monty`.

## Removals

- Obsolete `ReproduceEpisodeHandler` is removed.
- Obsolete `follow_up_configs` utility is removed.
- Obsolete reproducibility tests are removed from GraphLearningTest. (Note that `run_parallel_test` continues to test reproducibility. There are also the new reproducibility integration tests in `tests/integration`).
- ~~`hierarchy_test` and associated configurations are removed as being too brittle to work with other RNG seeds. Note that invariant-focused hierarchy tests are being reintroduced in the scope of #678.~~
- `get_episode_id` and `episode_id_parallel` are removed. They only worked for experiments that do `eval` and did not correctly determine an episode ID for logger use cases when an experiment ran `train` followed by `eval`.

## Benchmarks

All changed. This is because of the extensive changes that reset RNG before every episode.

For latest benchmark differences table, see https://github.com/thousandbrainsproject/tbp.monty/pull/666#issuecomment-3775387487

The screenshot below includes a bonus sample of four benchmarks run as `run.py`, showing that they result in exactly the same overall statistics (because there's no longer a difference between how episodes are run).

Screenshot 2025-12-31 at 18 44 34

### How do I reproduce an episode now?

Use the `episodes=` argument with `run_parallel.py`. For example:
```
python run_parallel.py experiment=<your_experiment> 'episodes="1:4"'
```
```
    Supported forms:
      - `"all"`, `":"`, or empty string: select all valid indices `[0, total)`
      - Comma-separated integers and ranges, for example `"0,3,5:8"`
      - Open-ended ranges (end-exclusive):
          - `":N"` selects `[0, N)` (i.e., indices `0` through `N-1`)
          - `"N:"` selects `[N, total)`
```

### Why is episode seed not dependent on epoch?

Episode count ignores epochs, so it is never reset, therefore the combination of `seed`, `mode`, and `episode` is guaranteed to be unique, no matter what epoch the episode is in. E.g.,
| epoch | episode |
|---|---|
|0|0
|0|1
|1|2
|1|3
|2|4
|...|...

### Why is experiment mode included in episode seed?

Because we track `train_episodes` and `eval_episodes` separately, starting from `0`, we want to be able to generate different seeds for the `0th` train episode vs the `0th` eval episode. Hence, the inclusion of experiment mode.

### Are training episodes reproducible?

No. Training episodes continue not to be reproducible between `run.py` and `run_parallel.py`. This is OK. The follow-up config functionality was not used for training episodes, so the episode reproducibility not working for training does not remove any existing capabilities.

### DIFF
--- a/conf/experiment/test/reproducibility_eval_episodes.yaml
+++ b/conf/experiment/test/reproducibility_eval_episodes.yaml
@@ -6,7 +6,7 @@ defaults:
   - /experiment/config/monty/clear_monty_config@config
   - /experiment/config/monty/patch_and_view@config.monty_config
   - /experiment/config/monty/motor_system/clear_motor_system_config@config.monty_config
-  - /experiment/test/config/monty/motor_system/fixed@config.monty_config.motor_system_config
+  - /experiment/test/config/monty/motor_system/uniformly_distributed@config.monty_config.motor_system_config
   - /experiment/config/monty/args/clear_monty_args@config.monty_config
   - /experiment/config/monty/args/defaults@config.monty_config.monty_args
   - /experiment/config/environment_interface/clear_eval_env_interface_args@config
--- a/conf/experiment/test/config/monty/motor_system/uniformly_distributed.yaml
+++ b/conf/experiment/test/config/monty/motor_system/uniformly_distributed.yaml
@@ -1,0 +1,9 @@
+defaults:
+  - /experiment/config/monty/motor_system/policy/informed@motor_system_args.policy_args
+  - /experiment/config/monty/motor_system/action_space/distant_agent_no_translation@motor_system_args.policy_args.action_sampler_args
+
+motor_system_class: ${monty.class:tbp.monty.frameworks.models.motor_system.MotorSystem}
+motor_system_args:
+  policy_class: ${monty.class:tbp.monty.frameworks.models.motor_policies.InformedPolicy}
+  policy_args:
+    action_sampler_class: ${monty.class:tbp.monty.frameworks.actions.action_samplers.UniformlyDistributedSampler}


src/tbp/monty/frameworks/experiments/__init__.py
```diff
@@ -7,21 +7,3 @@
...
-)
-from .profile import ProfileExperimentMixin
-
-__all__ = [
-    "MontyExperiment",
-    "MontyGeneralizationExperiment",
-    "MontyObjectRecognitionExperiment",
-    "MontySupervisedObjectPretrainingExperiment",
-    "ProfileExperimentMixin",
-]
```
tristanls-tbp, line 27: **note**: These are removed due to circular import errors.

src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -75,6 +67,10 @@ def __init__(self, config: DictConfig) -> None:
         """
         self.config = config
 
+        # TODO: Is it necessary to set the RNG here?
```
nielsleadholm: Is this TODO still outstanding?

nielsleadholm (commented):
Amazing, thanks for putting this together! 

Just had a brief look to begin with, overall looks great. Two high-level questions/comments:
- @ramyamounir created a script to quickly give an overview of the performance difference between two different runs of the benchmarks. Would it be possible to generate this visualization to show the differences from our change in seed? It would help to sanity check the impact on the results (I am expecting the differences to be, quite literally, due to noise, but e.g., a large change might reflect an underlying bug or unexpected behavior that changing the seed manifests, and it would be good to catch these). (I tagged you Ramy as I don't recall where this script is actually kept, in case you can share it/point to it)
- Re. not factoring in epoch - the reasoning makes sense. One thought is that, following a refactor in epochs, our experiment setup, or somewhere else (e.g. adding a concept of meta-epochs), it isn't hard to imagine a scenario where episode count *is* reset, and we start using the same seeds within an experiment. This would  fail silently, in that we may not become aware of the fact that a seed is being reused. What do you think of adding a lightweight register + check at the level of `MontyExperiment` to check whether a seed ever reoccurs in the scope of a single experiment? This isn't foolproof against undesired seed re-use, but could help catch the most frequent instances. Another, complementary step could be to add a column to CSV stats that lists the seed integer for each episode. That way we have an explicit log of which seeds are being used where.

src/tbp/monty/frameworks/experiments/monty_experiment.py
tristanls-tbp: Thank you. It is resolved, and yes, it is necessary so that all other things get initialized with an RNG as well.

tristanls-tbp:
@nielsleadholm I added the following for `MontyExperiment` to warn on episode seed reuse: https://github.com/thousandbrainsproject/tbp.monty/pull/666/changes/8c152385d0ddf8ba716cf32d8f4a6f69349cdc5b

tristanls-tbp:
@nielsleadholm and here's the `episode_seed` column addition to `eval_stats.csv`: https://github.com/thousandbrainsproject/tbp.monty/pull/666/commits/bd0a7d118bc6f628d21dcbe51545a8f53e0a5681

ramyamounir:
@nielsleadholm, I ended up maintaining this script myself on my personal github account. It is available [here](https://github.com/ramyamounir/tbp.benchmarks) and now includes [installation/usage](https://github.com/ramyamounir/tbp.benchmarks?tab=readme-ov-file#usage) instructions as well as [custom configs](https://github.com/ramyamounir/tbp.benchmarks/blob/main/configs/README.md) instructions.

@tristanls-tbp, to use this script on a full benchmark,

1) export wandb results csv of the experiments with the following columns included:
```txt
overall/percent_correct
overall/percent_correct_child_or_parent
overall/avg_rotation_error
overall/avg_num_monty_matching_steps
overall/parallel_run_time
overall/percent_used_mlh_after_timeout
overall/run_time
overall/avg_episode_run_time
overall/avg_prediction_error
```

2) Clone and run the comparison script:
```zsh
uv run compare \
  --baseline ~/tbp/tbp.monty \
  --proposed /path/to/wandb_export.csv \
  [--out summary.md]
```

3) Optionally, you can write back the results automatically to `tbp.monty` (to avoid human errors of manually copying the results):
```zsh
uv run compare \
  --baseline ~/tbp/tbp.monty \
  --proposed /path/to/wandb_export.csv \
  --modify-in-place \
  [--dry-run]
```

**Note that you can compare repositories against each other, or compare wandb export csv files against each other; there are configs for all combinations. Both `--baseline` and `--proposed` can accept a repo or a wandb export file.**

tristanls-tbp:
Thank you for the instructions @ramyamounir, it was very easy to set up, use, and customize 👍. It looks like there's a discrepancy where [`base_10multi_distinctojb_dist_agent_r0po8d40` took 45 minutes to run](https://wandb.ai/thousand-brains-project/Monty/runs/r0po8d40) but the script reports it as 5 minutes.

@nielsleadholm 

| Experiment | Metric | Baseline | Proposed | Change |
| base_config_10distinctobj_dist_agent | percent_correct (%) | 99.29 | 99.29 | 0 |
|  | used_mlh (%) | 0.71 | 0.71 | 0 |
|  | match_steps | 34 | 36 | 2 |
|  | rotation_error (deg) | 10.69 | 15.19 | 4.5 |
|  | runtime (min) | 3 | 2 | -1 |
|  | episode_runtime (sec) | 9 | 8 | -1 |
| base_config_10distinctobj_surf_agent | percent_correct (%) | 99.29 | 100 | 0.71 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 27 | 27 | 0 |
|  | rotation_error (deg) | 11.96 | 15.57 | 3.61 |
|  | runtime (min) | 3 | 3 | 0 |
|  | episode_runtime (sec) | 11 | 12 | 1 |
| randrot_noise_10distinctobj_dist_agent | percent_correct (%) | 97 | 96 | -1 |
|  | used_mlh (%) | 5 | 2 | -3 |
|  | match_steps | 55 | 44 | -11 |
|  | rotation_error (deg) | 25.14 | 21.6 | -3.54 |
|  | runtime (min) | 5 | 4 | -1 |
|  | episode_runtime (sec) | 28 | 22 | -6 |
| randrot_noise_10distinctobj_dist_on_distm | percent_correct (%) | 98 | 99 | 1 |
|  | used_mlh (%) | 3 | 0 | -3 |
|  | match_steps | 38 | 33 | -5 |
|  | rotation_error (deg) | 15.11 | 14.05 | -1.06 |
|  | runtime (min) | 4 | 2 | -2 |
|  | episode_runtime (sec) | 18 | 15 | -3 |
| randrot_noise_10distinctobj_surf_agent | percent_correct (%) | 100 | 99 | -1 |
|  | used_mlh (%) | 1 | 0 | -1 |
|  | match_steps | 28 | 27 | -1 |
|  | rotation_error (deg) | 25.14 | 22.25 | -2.89 |
|  | runtime (min) | 3 | 3 | 0 |
|  | episode_runtime (sec) | 20 | 21 | 1 |
| randrot_10distinctobj_surf_agent | percent_correct (%) | 100 | 100 | 0 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 28 | 28 | 0 |
|  | rotation_error (deg) | 13.46 | 15.06 | 1.6 |
|  | runtime (min) | 2 | 2 | 0 |
|  | episode_runtime (sec) | 13 | 12 | -1 |
| randrot_noise_10distinctobj_5lms_dist_agent | percent_correct (%) | 100 | 100 | 0 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 52 | 51 | -1 |
|  | rotation_error (deg) | 49.54 | 58.54 | 9 |
|  | runtime (min) | 5 | 6 | 1 |
|  | episode_runtime (sec) | 40 | 47 | 7 |
| base_10simobj_surf_agent | percent_correct (%) | 92.14 | 93.57 | 1.43 |
|  | used_mlh (%) | 10.71 | 10.71 | 0 |
|  | match_steps | 83 | 82 | -1 |
|  | rotation_error (deg) | 11 | 13.93 | 2.93 |
|  | runtime (min) | 7 | 7 | 0 |
|  | episode_runtime (sec) | 34 | 35 | 1 |
| randrot_noise_10simobj_dist_agent | percent_correct (%) | 78 | 80 | 2 |
|  | used_mlh (%) | 36 | 36 | 0 |
|  | match_steps | 201 | 221 | 20 |
|  | rotation_error (deg) | 21.46 | 33.13 | 11.67 |
|  | runtime (min) | 11 | 11 | 0 |
|  | episode_runtime (sec) | 92 | 92 | 0 |
| randrot_noise_10simobj_surf_agent | percent_correct (%) | 90 | 86 | -4 |
|  | used_mlh (%) | 32 | 36 | 4 |
|  | match_steps | 166 | 185 | 19 |
|  | rotation_error (deg) | 27.15 | 34.13 | 6.98 |
|  | runtime (min) | 17 | 20 | 3 |
|  | episode_runtime (sec) | 140 | 168 | 28 |
| randomrot_rawnoise_10distinctobj_surf_agent | percent_correct (%) | 65 | 72 | 7 |
|  | used_mlh (%) | 83 | 72 | -11 |
|  | match_steps | 13 | 15 | 2 |
|  | rotation_error (deg) | 101.6 | 99.45 | -2.15 |
|  | runtime (min) | 4 | 4 | 0 |
|  | episode_runtime (sec) | 5 | 7 | 2 |
| base_10multi_distinctobj_dist_agent | percent_correct (%) | 85 | 77.14 | -7.86 |
|  | used_mlh (%) | 12.86 | 12.86 | 0 |
|  | match_steps | 27 | 33 | 6 |
|  | rotation_error (deg) | 25.3 | 22.63 | -2.67 |
|  | runtime (min) | 3 | 5 | 2 |
|  | episode_runtime (sec) | 1 | 2 | 1 |
| base_77obj_dist_agent | percent_correct (%) | 90.91 | 95.24 | 4.33 |
|  | used_mlh (%) | 12.99 | 10.82 | -2.17 |
|  | match_steps | 102 | 93 | -9 |
|  | rotation_error (deg) | 14.89 | 14.66 | -0.23 |
|  | runtime (min) | 20 | 19 | -1 |
|  | episode_runtime (sec) | 58 | 54 | -4 |
| base_77obj_surf_agent | percent_correct (%) | 98.7 | 99.57 | 0.87 |
|  | used_mlh (%) | 5.19 | 7.36 | 2.17 |
|  | match_steps | 53 | 57 | 4 |
|  | rotation_error (deg) | 6.7 | 8.75 | 2.05 |
|  | runtime (min) | 14 | 15 | 1 |
|  | episode_runtime (sec) | 35 | 38 | 3 |
| randrot_noise_77obj_dist_agent | percent_correct (%) | 92.21 | 87.88 | -4.33 |
|  | used_mlh (%) | 26.84 | 25.54 | -1.3 |
|  | match_steps | 174 | 169 | -5 |
|  | rotation_error (deg) | 34.01 | 32.06 | -1.95 |
|  | runtime (min) | 41 | 38 | -3 |
|  | episode_runtime (sec) | 137 | 130 | -7 |
| randrot_noise_77obj_surf_agent | percent_correct (%) | 94.37 | 92.64 | -1.73 |
|  | used_mlh (%) | 21.65 | 25.11 | 3.46 |
|  | match_steps | 111 | 115 | 4 |
|  | rotation_error (deg) | 33.9 | 31.52 | -2.38 |
|  | runtime (min) | 36 | 40 | 4 |
|  | episode_runtime (sec) | 115 | 132 | 17 |
| randrot_noise_77obj_5lms_dist_agent | percent_correct (%) | 93.51 | 93.51 | 0 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 72 | 72 | 0 |
|  | rotation_error (deg) | 56.34 | 47.7 | -8.64 |
|  | runtime (min) | 13 | 15 | 2 |
|  | episode_runtime (sec) | 119 | 128 | 9 |
| unsupervised_inference_distinctobj_dist_agent | percent_correct (%) | — | — | — |
|  | match_steps | — | — | — |
|  | runtime (min) | — | — | — |
|  | episode_runtime (sec) | — | — | — |
| unsupervised_inference_distinctobj_surf_agent | percent_correct (%) | — | — | — |
|  | match_steps | — | — | — |
|  | runtime (min) | — | — | — |
|  | episode_runtime (sec) | — | — | — |
| infer_comp_lvl1_with_monolithic_models | correct_child_or_parent (%) | 59.52 | 59.52 | 0 |
|  | used_mlh (%) | 55.95 | 51.19 | -4.76 |
|  | match_steps | 340 | 327 | -13 |
|  | rotation_error (deg) | 66.18 | 91.19 | 25.01 |
|  | avg_prediction_error | 0.37 | 0.38 | 0.01 |
|  | runtime (min) | 30 | 31 | 1 |
| infer_comp_lvl1_with_comp_models | correct_child_or_parent (%) | 80.95 | 89.29 | 8.34 |
|  | used_mlh (%) | 53.57 | 38.1 | -15.47 |
|  | match_steps | 73 | 45 | -28 |
|  | rotation_error (deg) | 54.53 | 47.77 | -6.76 |
|  | avg_prediction_error | 0.34 | 0.33 | -0.01 |
|  | runtime (min) | 6 | 4 | -2 |
| infer_comp_lvl2_with_comp_models | correct_child_or_parent (%) | 76.19 | 80.48 | 4.29 |
|  | used_mlh (%) | 49.05 | 41.9 | -7.15 |
|  | match_steps | 43 | 36 | -7 |
|  | rotation_error (deg) | 43.37 | 43.53 | 0.16 |
|  | avg_prediction_error | 0.33 | 0.34 | 0.01 |
|  | runtime (min) | 12 | 11 | -1 |
| infer_comp_lvl3_with_comp_models | correct_child_or_parent (%) | 53.71 | 64.86 | 11.15 |
|  | used_mlh (%) | 64.57 | 52.29 | -12.28 |
|  | match_steps | 46 | 34 | -12 |
|  | rotation_error (deg) | 45.8 | 45.32 | -0.48 |
|  | avg_prediction_error | 0.32 | 0.33 | 0.01 |
|  | runtime (min) | 22 | 18 | -4 |
| infer_comp_lvl4_with_comp_models | correct_child_or_parent (%) | 52.2 | 58.52 | 6.32 |
|  | used_mlh (%) | 65.38 | 58.79 | -6.59 |
|  | match_steps | 45 | 36 | -9 |
|  | rotation_error (deg) | 47.51 | 46.39 | -1.12 |
|  | avg_prediction_error | 0.32 | 0.33 | 0.01 |
|  | runtime (min) | 23 | 22 | -1 |

ramyamounir:
Cool, nice customization! 

WandB reports its (the experiment's) runtime as `273.97304` seconds, which the script correctly converts to 5 minutes. This seems to be a logging issue or something I don't understand about this experiment. I also don't really understand why this experiment needs to run with `run.py` and not `run_parallel.py`, the only difference is adding distractor objects to each episode. But maybe I'm missing something.

pic-selected-260102-1800-56

nielsleadholm:
Thanks @tristanls-tbp for sharing that table of results, that's really helpful.

Yeah worth understanding the 5 vs 45 minute discrepancy for base_10multi_distinctobj_dist_agent

There are also a few experiments where the matching steps or run-time are different by 1 between the proposed CSV results, and the "Proposed" in the comparison column, i.e. results that should be the same. It might just be worth checking for rounding/transcription errors. These are:

randrot_noise_10simobj_dist_agent (matching steps 220 in CSV vs 221 in proposed column)
infer_comp_lvl2_with_comp_models (runtime 12 in CSV vs 11 in proposed column)
infer_comp_lvl3_with_comp_models (runtime 19 in CSV vs 18 in proposed column)

src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -75,6 +67,8 @@ def __init__(self, config: DictConfig) -> None:
         """
         self.config = config
 
+        self.rng = np.random.RandomState(config["seed"])
```
nielsleadholm, line 69: I find it a bit confusing that this initial self.rng, which is used for model setup etc, is set differently from the self.rng that is used in all the episodes (i.e. the one that makes use of `reset_episode_rng`). In other words, they are both set differently, and also used differently. What do you think of calling this one `self.meta_rng` or `self.setup_rng` or the like, to make it clear that this is essentially a different class of rng?

src/tbp/monty/frameworks/environments/object_init_samplers.py
```diff
@@ -10,13 +10,14 @@
...
+from tbp.monty.frameworks.experiments.mode import ExperimentMode
 from tbp.monty.frameworks.experiments.seed import episode_seed
 from tbp.monty.frameworks.utils.transform_utils import scipy_to_numpy_quat
 
 
 class Default:
-    def __call__(self, seed: int, epoch: int, episode: int):
-        seed = episode_seed(seed, epoch, episode)
+    def __call__(self, seed: int, mode: ExperimentMode, epoch: int, episode: int):  # noqa: ARG002
+        seed = episode_seed(seed, mode, episode)
```
nielsleadholm, line 20: As I understand it, an rng is available in the `EnvironmentInterface` classes, which call the object init samplers. Is there a reason we don't pass the rng, and instead pass a seed and create a new rng (other than it being how it was done before)? Practically speaking, I'm wondering whether there's a risk that if we call object-init-sampler with a different tempo in the future (e.g. multiple times in an episode because of a task setting where multiple objects spawn), then there is a risk of seed re-use. 

Maybe it's obvious but just want to check my understanding, as this seems to differ from the pattern used elsewhere.

tests/unit/evidence_lm_test.py
```diff
@@ -177,6 +176,7 @@ def get_elm_with_fake_object(
         gsg_args=None,
     ):
         graph_lm = EvidenceGraphLM(
+            rng=np.random.RandomState(),
```
nielsleadholm, line 179: Is there a reason a seed isn't being passed when these rngs are initialized? Doesn't that make the unit tests non-deterministic?

tests/unit/evidence_sdr_lm_test.py
```diff
@@ -455,6 +455,7 @@ def get_eslm(self) -> EvidenceSDRGraphLM:
             Evidence SDR Graph Learning Module.
         """
         return EvidenceSDRGraphLM(
+            rng=np.random.RandomState(),
```
nielsleadholm, line 458: Again just checking about non-deterministic seeds in the unit tests.

nielsleadholm (commented):
I've now gone through the full PR, just added a couple minor questions/comments. 

Also just highlighting the updating-the-hierarchy-test point here so we don't lose track of that.

src/tbp/monty/frameworks/experiments/monty_experiment.py
jeremyshoemaker: We could have Hydra create this object for us, and then just change this to something like `self.rng = config["rng"]`.

src/tbp/monty/frameworks/environments/object_init_samplers.py
tristanls-tbp: Short explanation: object initializers are special.

[Long explanation is available in this other pull request](https://github.com/thousandbrainsproject/tbp.monty/pull/636#issuecomment-3662745403).

tests/unit/evidence_lm_test.py
tristanls-tbp: The `rng` is unused in these tests, so they don't matter.

tests/unit/evidence_sdr_lm_test.py
tristanls-tbp: Similarly to before, `rng` is unused, so it doesn't matter.

src/tbp/monty/frameworks/experiments/monty_experiment.py
tristanls-tbp: I recommend not framing this as different RNGs.

There's only one RNG used by the `MontyExperiment`. The RNG is reset with different seeds depending on the circumstances.

On initialization of `MontyExperiment`, there is nothing special about this RNG. The constraint is that there are no episodes or experiment mode, so the only seed that we can use is the `seed`.

Later on, once the experiment starts, and we want to change the RNG seed, we want to set the seed according to different criteria, depending on experiment mode and episode.

conf/test/graph_building/load_habitat.yaml
```diff
@@ -6,7 +6,7 @@ defaults:
   - /experiment/config/environment_interface/per_object_train@config.train_env_interface_args
   - /experiment/config/environment_interface/per_object_eval@config.eval_env_interface_args
 
-_target_: tbp.monty.frameworks.experiments.MontyObjectRecognitionExperiment
+_target_: tbp.monty.frameworks.experiments.object_recognition_experiments.MontyObjectRecognitionExperiment
```
jeremyshoemaker, line 9: Do we need to have `experiments` in this module name, since it's already in the `experiments` module?

The file name already doesn't match the class name (and this isn't Java anyway). The same question applies to the `pretraining_experiments` module as well.

tristanls-tbp:
> Yeah worth understanding the 5 vs 45 minute discrepancy for base_10multi_distinctobj_dist_agent

@nielsleadholm I don't think this should block this pull request. The value is 45 minutes.

If there is a problem, it is with why `overall/run_time` reports 5 minutes when the experiment runs for 45 minutes, but this is not caused by this pull request.
Screenshot 2026-01-05 at 13 31 39

conf/test/graph_building/load_habitat.yaml
tristanls-tbp: We do not 🙂. PR welcome 🙂.

tristanls-tbp:
> randrot_noise_10simobj_dist_agent (matching steps 220 in CSV vs 221 in proposed column)
infer_comp_lvl2_with_comp_models (runtime 12 in CSV vs 11 in proposed column)
infer_comp_lvl3_with_comp_models (runtime 19 in CSV vs 18 in proposed column)

@nielsleadholm infer_comp_lvl2_with_comp_models and infer_comp_lvl3_with_comp_models are correct. 11m37s -> 12m. 18m35s -> 19m.

I updated randrot_noise_10simobj_dist_agent entry.

ramyamounir:
@tristanls-tbp

> infer_comp_lvl2_with_comp_models and infer_comp_lvl3_with_comp_models are correct. 

Do you know where the data for this `RUNTIME` column is coming from? Is it reported by wandb itself or one of our experiment loggers? The script looks at the `overall/parallel_run_time` for `run_parallel.py` experiments and `overall/run_time` for `run.py` experiments. The values reported for these columns are coming from our loggers and match the results from the script tables.
- `lvl2`: 677.35612 seconds -> 11.289 mins -> 11 mins
- `lvl3`: 1088.11376 seconds -> 18.135 mins -> 18 mins

pic-selected-260105-1640-54

tristanls-tbp:
@ramyamounir 

WandB itself reports the RUNTIME column.

> **Runtime**: Measures the total time from the start to the end of the run. It’s the wall-clock time for the run, including any time where the run is paused or waiting for resources. This metric provides the complete elapsed time for your run.
> -- https://docs.wandb.ai/models/runs#overview-tab

Considering that `overall/run_time` is incorrect, I trust WandB's reporting. The difference between the rounded values of `overall/parallel_run_time` and `RUNTIME` seems insignificant.

ramyamounir:
@tristanls-tbp 

**Nitpick**: Yeah, I see your point about the incorrect `overall/run_time`, but this is a bug not a case of slightly inaccurate reporting of time. We should figure it out and fix it. That doesn't mean `overall/parallel_run_time` is inaccurate. Based on [the code](https://github.com/thousandbrainsproject/tbp.monty/blob/49dd20a8a1a6e5fde4c174a546a846aab8c3ee31/src/tbp/monty/frameworks/run_parallel.py?plain=1#L614-L634), it looks like `overall/parallel_run_time` reports the time starting after the wandb setup, and ending before the `post_parallel` functions. I think it's coded like that on purpose because we don't want to include this overhead as part of the experiment run time. I would prefer that we report the logged run time value that doesn't include the extra overhead of setting up and clean up function (which sometimes hangs). In this current case, the difference of 1 min is insignificant, but I don't know if it can become significant under different circumstances (e.g., slower internet/wandb servers, or high I/O workload).

tristanls-tbp:
@ramyamounir why do you insist on using `overall/parallel_run_time`? It really does not matter what runtime we use. Here is an example of `no_change` runs of `base_77obj_dist_agent`. The difference between "identical" runs in `overall/parallel_run_time` is _multiples_ of the difference between `overall/parallel_run_time` and `RUNTIME`. It doesn't matter which one we use.

Screenshot 2026-01-05 at 20 13 57

> _multiples_

`1328s` = `22.1m`
`1172s` = `19.5m`
`1291s` = `21.5m`

The differences between "identical" runs are `2-3` minutes. The differences between `RUNTIME` and `overall/parallel_run_time` is ~`1` minute.

The run time is the least relevant part of the benchmark and is only there for order-of-magnitude sanity checks, like, for example, 5 min vs 45 min.

tristanls-tbp:
@nielsleadholm I opened #678 to reintroduce the hierarchy test that this PR deletes for expediency.

ramyamounir:
@tristanls-tbp 

> why do you insist on using `overall/parallel_run_time`?

I don't. I used the keyword **Nitpick** to indicate that I don't. But if it makes any difference, here is my reasoning:
- The run time metric is a lower-is-better metric. I see no reason to include overhead time and report a worse value for the metric than the actual runtime of the experiment.
- There is a difference between the variance in runtime between identical runs (+/- 2 mins) and the added overhead in the `RUNTIME` metric (+ 1 min). In any run, the `RUNTIME` metric will **always** be higher than `overall/parallel_run_time`. If we decide to report the mean of 10 runs to provide a more accurate estimate, the mean of `RUNTIME` will still be higher than the mean of `overall/parallel_run_time`.
- This is PR 666, so I wanted to give you some trouble to make it special :smiling_imp: 

It really doesn't matter, but just let me know if you decide to use `RUNTIME` for the documentation so that I can update my benchmark script to do the same. I just prefer that we all report the same thing.

src/tbp/monty/frameworks/environments/object_init_samplers.py
nielsleadholm: Nice, that's helpful thanks.

nielsleadholm (approved):
Thanks Tristan, LGTM!

tristanls-tbp:
@ramyamounir, for now, personally, I'm using `RUNTIME`. I took a look, and `overall/run_time` looks non-trivially incorrect for a cascade of reasons that I don't want to spend time debugging/fixing right now.

tests/unit/graph_learning_test.py
```diff
@@ -336,262 +328,6 @@ def test_fixed_actions_feat(self):
         eval_stats = pd.read_csv(output_dir / "eval_stats.csv")
         self.check_eval_results(eval_stats)
 
-    def test_reproduce_single_episode(self):
```
vkakerbeck, line 339: shouldn't there be a new equivalent test for this which, instead of using the follow-up configs feature, reruns some episodes of a full experiment with the `episodes=` feature and checks the results are the same?

tests/unit/hierarchy_test.py
vkakerbeck, line 1: I'm not sure I understand why we need to remove these tests here? I agree they are not ideal since they rely on a specific action sequence to produce the desired result, but they are still the only end-to-end test of hierarchy we have at the moment. Why not just pass a fixed action sequence instead to fix it? I am worried if we just remove it we will break hierarchy features without noticing.

tests/unit/graph_learning_test.py
tristanls-tbp: Sorry, I thought I mentioned it but I did not. I had the same thought, but then noticed `tests/unit/run_parallel_test.py::RunParallelTest::test_run_parallel_equals_serial_for_various_n_eval_epochs`. I believe it performs exactly the check you mention here.
```python
    def test_run_parallel_equals_serial_for_various_n_eval_epochs(self):
        # serial run
        exp = hydra.utils.instantiate(self.supervised_pre_training_cfg.experiment)
        with exp:
            exp.run()

        # parallel run
        OmegaConf.clear_resolvers()  # main will re-register resolvers
        main(self.supervised_pre_training_cfg)

        ###
        # Compare results
        ###
        parallel_model = torch.load(
            self.output_dir
            / self.supervised_pre_training_cfg.experiment.config.logging.run_name
            / "pretrained"
            / "model.pt"
        )
        serial_model = torch.load(self.output_dir / "pretrained" / "model.pt")

        # Same objects
        self.assertEqual(
            parallel_model["lm_dict"][0].keys(), serial_model["lm_dict"][0].keys()
        )

        # Same number of features
        self.assertEqual(
            parallel_model["lm_dict"][0]["graph_memory"]["capsule3DSolid"][
                "patch"
            ].x.size(1),
            serial_model["lm_dict"][0]["graph_memory"]["capsule3DSolid"][
                "patch"
            ].x.size(1),
        )

        # Same number of data points
        self.assertEqual(
            parallel_model["lm_dict"][0]["graph_memory"]["capsule3DSolid"][
                "patch"
            ].num_nodes,
            serial_model["lm_dict"][0]["graph_memory"]["capsule3DSolid"][
                "patch"
            ].num_nodes,
        )

        ###
        # n_eval_epochs == len(eval_rotations)
        ###
        # serial run
        exp = hydra.utils.instantiate(self.eval_cfg.experiment)
        with exp:
            exp.run()

        # parallel run
        OmegaConf.clear_resolvers()  # main will re-register resolvers
        main(self.eval_cfg)

        ###
        # Compare results
        ###
        eval_dir = self.output_dir / "eval"
        parallel_eval_dir = eval_dir / "test_eval"

        # Check that csv files are the same
        # Note that you can't easily do this if they actually run in parallel because
        # you don't know the execution order so it will have the same data, just
        # different order

        scsv = pd.read_csv(eval_dir / "eval_stats.csv")
        pcsv = pd.read_csv(parallel_eval_dir / "eval_stats.csv")

        # We have to drop these columns because they are not the same in the parallel
        # and serial runs. In particular, 'stepwise_performance' and
        # 'stepwise_target_object' are derived from the mapping between semantic IDs to
        # names which depend on the number of objects in the environment, and
        # environments only have one object in parallel experiments.
        drop_cols = ["time", "stepwise_performance", "stepwise_target_object"]
        scsv = scsv.drop(columns=drop_cols)
        pcsv = pcsv.drop(columns=drop_cols)

        self.assertTrue(pcsv.equals(scsv))

        ###
        # n_eval_epochs < len(eval_rotations)
        ###
        # serial run
        exp = hydra.utils.instantiate(self.eval_lt_cfg.experiment)
        with exp:
            exp.run()

        # parallel run
        OmegaConf.clear_resolvers()  # main will re-register resolvers
        main(self.eval_lt_cfg)

        eval_dir_lt = self.output_dir / "lt"
        parallel_eval_dir_lt = eval_dir_lt / "test_eval_lt"

        scsv_lt = pd.read_csv(eval_dir_lt / "eval_stats.csv")
        pcsv_lt = pd.read_csv(parallel_eval_dir_lt / "eval_stats.csv")

        # Remove columns that are not the same in the parallel and serial runs.
        scsv_lt = scsv_lt.drop(columns=drop_cols)
        pcsv_lt = pcsv_lt.drop(columns=drop_cols)

        self.assertTrue(pcsv_lt.equals(scsv_lt))

        ###
        # n_eval_epochs > len(eval_rotations)
        ###
        # serial run
        exp = hydra.utils.instantiate(self.eval_gt_cfg.experiment)
        with exp:
            exp.run()

        # parallel run
        OmegaConf.clear_resolvers()  # main will re-register resolvers
        main(self.eval_gt_cfg)

        eval_dir_gt = self.output_dir / "gt"
        parallel_eval_dir_gt = eval_dir_gt / "test_eval_gt"

        scsv_gt = pd.read_csv(eval_dir_gt / "eval_stats.csv")
        pcsv_gt = pd.read_csv(parallel_eval_dir_gt / "eval_stats.csv")

        scsv_gt = scsv_gt.drop(columns=drop_cols)
        pcsv_gt = pcsv_gt.drop(columns=drop_cols)

        self.assertTrue(pcsv_gt.equals(scsv_gt))

        shutil.rmtree(self.output_dir)
```

The way I'm reading these deleted tests in `graph_learning_test.py` is that they are specifically testing the follow up utility functions, which are being deleted. 
| test | method tested|
|---|---
|`test_reproduce_single_episode`|`create_eval_episode_hydra_cfg()`|
|`test_reproduce_multiple_episodes`|`create_eval_multiple_episodes_hydra_cfg()`|
|`test_reproduce_single_episode_with_multiple_episode_function`|`create_eval_multiple_episodes_hydra_cfg()` differently|

vkakerbeck: Ah okay thanks, I wasn't aware of this test. I am surprised that it passed before this change. Why were we able to get the same results with run and run parallel before this change? 
Also, does it cover the case where we just rerun select episodes from an experiment?

tristanls-tbp: After inspection, it appears that the test settings eliminate the non-determinism caused by RNG using two mechanisms. First, all of the object init samplers are `Predefined`. Second, all actions come from `resources/fixed_test_actions.jsonl`.

I'm going to see if I can put together an integration test to test episode reproducibility.

tristanls-tbp:
Ok. I added integration tests to verify supervised training reproducibility and eval per-episode reproducibility. 

These tests are in `tests/integration` and run in the `test-integration-monty` group (after `test-monty` and in parallel with `test-tutorials-monty`). These are not included by default in `pytest`, and to run them, one must run `pytest tests/integration`. Once this pull request is merged, I will make the `test-integration-monty` job required.

I believe all that remains is to update the benchmark numbers since the baseline shifted due to other commits.

tristanls-tbp:
@jeremyshoemaker I re-requested a review from you. I could use your set of eyes on `tests/integration`.

src/tbp/monty/frameworks/loggers/graph_matching_loggers.py
```diff
@@ -289,9 +290,10 @@ def update_overall_stats(self, mode, episode, episode_steps, monty_matching_step
                 # but still have an overall performance of correct (or other).
                 episode_performance = p
 
-        for p in self.performance_options:
-            stats[f"episode_{p}"] = int(p == episode_performance)
-            stats[f"num_{p}"] += int(p == episode_performance)
+        if episode_performance is not None:
```
jeremyshoemaker: This has been on my list for a while to fix, glad to see it being fixed. 

Just one suggestion. The only way this change could be a problem is if `p` could be an empty string (which is falsey), but I don't think that's going to happen.
```suggestion
        if episode_performance:
```

tests/integration/reproducibility/config.py
jeremyshoemaker, line 1: This looks like it was factored out of one of the tests where I inlined it, or something similar. I thought about doing this then, but decided to wait to see if we needed it in more places.

Do we want to go ahead and update those other copies to use this one instead, or save that for a future PR? (I think you'll probably say the latter, so this is more a comment for myself to follow up on).

jeremyshoemaker (approved):
The updates since my last review look fine to me.

tests/integration/reproducibility/config.py
tristanls-tbp: Initially, I had one reproducibility test that was basically a copy of `run_parallel_test.py`. I only refactored `hydra_config` out because I split my reproducibility test into two files.

The other four places that `def hydra_config` all define it with a different signature, so I don't see an obvious refactor yet.

tests/unit/graph_learning_test.py
tristanls-tbp: The integration tests are now in `tests/integration` and part of the `test-integration-monty` job.

tests/integration/reproducibility/config.py
jeremyshoemaker: That's my memory of it as well, that they were dissimilar enough to not warrant a refactor. I won't worry about it then.

tests/unit/hierarchy_test.py
tristanls-tbp: I undeleted the test files after #701 and #702 were merged.

Observing new test failure as a result.

```
pytest tests/unit/hierarchy_test.py::HierarchyTest::test_semisupervised_stacked_lms_experiment
```
Screenshot 2026-01-16 at 16 27 38

tristanls-tbp: Interesting excerpt from the log:
```
INFO:tbp.monty:run_epoch:580:Running a simulation to model object: capsule3DSolid
INFO:tbp.monty.frameworks.experiments.object_recognition_experiments:pre_episode:58:running eval epoch 0 eval episode 0
INFO:tbp.monty:reset_episode_rng:119:resetting RNG to seed 1170100843
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:240:Sending input from ['patch_0'] to learning_module_0
INFO:tbp.monty.frameworks.models.evidence_matching.learning_module:_calculate_most_likely_hypothesis:1226:current most likely hypothesis: capsule3DSolid with evidence 1.0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:240:Sending input from ['patch_0'] to learning_module_0
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 98 out of 98 hypotheses for capsule3DSolid (evidence > 0.5890196078431373)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 106 out of 106 hypotheses for cubeSolid (evidence > 0.5490196078431373)
INFO:tbp.monty.frameworks.models.evidence_matching.learning_module:_calculate_most_likely_hypothesis:1226:current most likely hypothesis: capsule3DSolid with evidence 3.0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:240:Sending input from ['patch_0'] to learning_module_0
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 98 out of 98 hypotheses for capsule3DSolid (evidence > -0.41098039215686266)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 106 out of 106 hypotheses for cubeSolid (evidence > -0.4509803921568627)
INFO:tbp.monty.frameworks.models.evidence_matching.learning_module:_calculate_most_likely_hypothesis:1226:current most likely hypothesis: capsule3DSolid with evidence 4.5
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:240:Sending input from ['patch_0'] to learning_module_0
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 98 out of 98 hypotheses for capsule3DSolid (evidence > -1.4109803921568627)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 106 out of 106 hypotheses for cubeSolid (evidence > -1.4509803921568627)
INFO:tbp.monty.frameworks.models.evidence_matching.learning_module:_calculate_most_likely_hypothesis:1226:current most likely hypothesis: capsule3DSolid with evidence 6.0
...
 INFO:tbp.monty.frameworks.models.graph_matching:update_terminal_condition:746:Pose for learning_module_0 not narrowed down yet
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:240:Sending input from ['patch_0'] to learning_module_0
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 98 out of 98 hypotheses for capsule3DSolid (evidence > -2.410980392156863)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 106 out of 106 hypotheses for cubeSolid (evidence > -2.450980392156863)
INFO:tbp.monty.frameworks.models.evidence_matching.learning_module:_calculate_most_likely_hypothesis:1226:current most likely hypothesis: capsule3DSolid with evidence 7.5
...
INFO:tbp.monty.frameworks.models.graph_matching:update_terminal_condition:746:Pose for learning_module_0 not narrowed down yet
...
 INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:240:Sending input from ['patch_0'] to learning_module_0
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 98 out of 98 hypotheses for capsule3DSolid (evidence > -3.410980392156863)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 106 out of 106 hypotheses for cubeSolid (evidence > -3.450980392156863)
INFO:tbp.monty.frameworks.models.evidence_matching.learning_module:_calculate_most_likely_hypothesis:1226:current most likely hypothesis: capsule3DSolid with evidence 9.0
...
INFO:tbp.monty.frameworks.models.graph_matching:update_terminal_condition:746:Pose for learning_module_0 not narrowed down yet
...
 INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:240:Sending input from ['patch_0'] to learning_module_0
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 98 out of 98 hypotheses for capsule3DSolid (evidence > -4.410980392156863)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 106 out of 106 hypotheses for cubeSolid (evidence > -4.450980392156863)
INFO:tbp.monty.frameworks.models.evidence_matching.learning_module:_calculate_most_likely_hypothesis:1226:current most likely hypothesis: capsule3DSolid with evidence 10.5
...
 INFO:tbp.monty.frameworks.models.graph_matching:update_terminal_condition:746:Pose for learning_module_0 not narrowed down yet
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:254:Skipping step on learning module 0
...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:240:Sending input from ['patch_0'] to learning_module_0
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 98 out of 98 hypotheses for capsule3DSolid (evidence > -5.410980392156863)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 106 out of 106 hypotheses for cubeSolid (evidence > -5.450980392156863)
INFO:tbp.monty.frameworks.models.evidence_matching.learning_module:_calculate_most_likely_hypothesis:1226:current most likely hypothesis: capsule3DSolid with evidence 9.5
# now the evidence starts decreasing by -1 every time input from patch_0 goes to learning_module_0... 
# eventually it reaches...
INFO:tbp.monty.frameworks.models.graph_matching:_step_learning_modules:240:Sending input from ['patch_0'] to learning_module_0
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 98 out of 98 hypotheses for capsule3DSolid (evidence > -15.410980392156862)
INFO:tbp.monty.frameworks.models.evidence_matching.hypotheses_displacer:displace_hypotheses_and_compute_evidence:156:Testing 106 out of 106 hypotheses for cubeSolid (evidence > -15.450980392156863)
INFO:tbp.monty.frameworks.models.evidence_matching.learning_module:_calculate_most_likely_hypothesis:1226:current most likely hypothesis: capsule3DSolid with evidence -0.5
INFO:tbp.monty.frameworks.models.graph_matching:set_individual_ts:901:Setting terminal state of learning_module_0 to no_match
```

vkakerbeck: PR 701 and 701 fixed `test_two_lm_heterarchy_experiment`. They updated the `two_lms_heterarchy.yaml` file to use the spiral policy for the deterministic action sequence. I would propose to try the same fix for the `test_semisupervised_stacked_lms_experiment`. Its hard to interpret these logs without more context of where this excerpt was taken from and what the conditions were. If you like we can do a pair programming session to fix this.

vkakerbeck: Okay so making the same changes to the `two_lms_constrained.yaml` config gets us to the desired behavior. Here is the config I used:
```
defaults:
  - /test/hierarchy/base@
  - /experiment/config/supervised_pretraining@config
  - /experiment/config/logging/clear_logging@config
  - /experiment/config/logging/pretrain@config.logging
  - /experiment/config/monty/clear_monty_config@config
  - /experiment/config/monty/two_lm_stacked@config.monty_config
  - /experiment/config/monty/motor_system/clear_motor_system_config@config.monty_config
  - /experiment/config/monty/motor_system/naive_scan_spiral@config.monty_config.motor_system_config
  - /experiment/config/monty/motor_system/policy/naive_scan@config.monty_config.motor_system_config.motor_system_args.policy_args
  - /experiment/config/monty/learning_modules/clear_learning_module_configs@config.monty_config
  - /experiment/config/environment/two_lm_stacked_distant_mount_habitat@config.env_interface_config

_target_: tbp.monty.frameworks.experiments.pretraining_experiments.MontySupervisedObjectPretrainingExperiment
config:
  do_eval: false
  max_train_steps: 30
  max_eval_steps: 30
  max_total_steps: 60
  logging:
    python_log_level: INFO
  monty_config:
    monty_args:
      num_exploratory_steps: 30
    learning_module_configs:
      learning_module_0:
        learning_module_class: ${monty.class:tbp.monty.frameworks.models.evidence_matching.learning_module.EvidenceGraphLM}
        learning_module_args:
          max_match_distance: 0.001
          tolerances:
            patch_0:
              hsv: [0.1, 1, 1]
              principal_curvatures_log: [1, 1]
          feature_weights: {}
          max_graph_size: 0.3
          num_model_voxels_per_dim: 200
          max_nodes_per_graph: 2000
          object_evidence_threshold: 20
      learning_module_1:
        learning_module_class: ${monty.class:tbp.monty.frameworks.models.evidence_matching.learning_module.EvidenceGraphLM}
        learning_module_args:
          max_match_distance: 0.001
          tolerances:
            patch_1:
              hsv: [0.1, 1, 1]
              principal_curvatures_log: [1, 1]
            # Object ID currently is an int representation of the strings
            # in the object label so we keep this tolerance high. This is
            # just until we have added a way to encode object ID with some
            # real similarity measure.
            learning_module_0:
              object_id: 1
          feature_weights:
            learning_module_0:
              object_id: 1
          max_graph_size: 0.4
          num_model_voxels_per_dim: 200
          max_nodes_per_graph: 2000
    sensor_module_configs:
      sensor_module_1:
        sensor_module_args:
          delta_thresholds:
            distance: 0.02
    motor_system_config:
      motor_system_args:
        policy_args:
          fixed_amount: 10
  env_interface_config:
    env_init_args:
      data_path: null
```

However, the assert here seems wrong: https://github.com/thousandbrainsproject/tbp.monty/blob/aacd23d23dcf7599c5632e42e63c47645dd58fad/tests/unit/hierarchy_test.py#L283 As far as I can see, the object was shown in both of the rotations that it is tested on during training so it should also recognize both. I will double check this again tomorrow morning but here is my suggested change for lines 275-289:
```
for episode in range(2):
                for lm_id in range(num_lms):
                    self.assertIn(
                        eval_stats["primary_performance"][episode * 2 + lm_id],
                        ["correct", "correct_mlh"],
                        f"LM {lm_id} did not recognize the object on first episode.",
                    )
```

I can make a separate PR with the fix tomorrow.

vkakerbeck: Ok, PR with proposed fix is here: https://github.com/thousandbrainsproject/tbp.monty/pull/703

tristanls-tbp:
Latest benchmarks:

| Experiment | Metric | Baseline | Proposed | Change |
| base_config_10distinctobj_dist_agent | percent_correct (%) | 99.29 | 99.29 | 0 |
|  | used_mlh (%) | 0.71 | 0.71 | 0 |
|  | match_steps | 35 | 36 | 1 |
|  | rotation_error (deg) | 10.69 | 15.19 | 4.5 |
|  | runtime (min) | 2 | 2 | 0 |
|  | episode_runtime (sec) | 9 | 9 | 0 |
| base_config_10distinctobj_surf_agent | percent_correct (%) | 99.29 | 100 | 0.71 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 27 | 27 | 0 |
|  | rotation_error (deg) | 11.96 | 15.57 | 3.61 |
|  | runtime (min) | 2 | 2 | 0 |
|  | episode_runtime (sec) | 10 | 11 | 1 |
| randrot_noise_10distinctobj_dist_agent | percent_correct (%) | 97 | 96 | -1 |
|  | used_mlh (%) | 5 | 2 | -3 |
|  | match_steps | 55 | 44 | -11 |
|  | rotation_error (deg) | 25.14 | 21.6 | -3.54 |
|  | runtime (min) | 5 | 4 | -1 |
|  | episode_runtime (sec) | 31 | 20 | -11 |
| randrot_noise_10distinctobj_dist_on_distm | percent_correct (%) | 98 | 99 | 1 |
|  | used_mlh (%) | 3 | 0 | -3 |
|  | match_steps | 38 | 33 | -5 |
|  | rotation_error (deg) | 15.11 | 14.05 | -1.06 |
|  | runtime (min) | 4 | 3 | -1 |
|  | episode_runtime (sec) | 24 | 17 | -7 |
| randrot_noise_10distinctobj_surf_agent | percent_correct (%) | 100 | 99 | -1 |
|  | used_mlh (%) | 1 | 0 | -1 |
|  | match_steps | 28 | 27 | -1 |
|  | rotation_error (deg) | 25.14 | 22.25 | -2.89 |
|  | runtime (min) | 3 | 3 | 0 |
|  | episode_runtime (sec) | 21 | 20 | -1 |
| randrot_10distinctobj_surf_agent | percent_correct (%) | 100 | 100 | 0 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 28 | 28 | 0 |
|  | rotation_error (deg) | 13.46 | 15.06 | 1.6 |
|  | runtime (min) | 2 | 2 | 0 |
|  | episode_runtime (sec) | 12 | 13 | 1 |
| randrot_noise_10distinctobj_5lms_dist_agent | percent_correct (%) | 100 | 100 | 0 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 52 | 51 | -1 |
|  | rotation_error (deg) | 49.54 | 58.54 | 9 |
|  | runtime (min) | 5 | 6 | 1 |
|  | episode_runtime (sec) | 42 | 44 | 2 |
| base_10simobj_surf_agent | percent_correct (%) | 92.14 | 93.57 | 1.43 |
|  | used_mlh (%) | 10.71 | 10.71 | 0 |
|  | match_steps | 83 | 82 | -1 |
|  | rotation_error (deg) | 11 | 13.93 | 2.93 |
|  | runtime (min) | 6 | 6 | 0 |
|  | episode_runtime (sec) | 31 | 29 | -2 |
| randrot_noise_10simobj_dist_agent | percent_correct (%) | 78 | 80 | 2 |
|  | used_mlh (%) | 36 | 36 | 0 |
|  | match_steps | 201 | 221 | 20 |
|  | rotation_error (deg) | 21.46 | 33.13 | 11.67 |
|  | runtime (min) | 10 | 11 | 1 |
|  | episode_runtime (sec) | 85 | 90 | 5 |
| randrot_noise_10simobj_surf_agent | percent_correct (%) | 90 | 86 | -4 |
|  | used_mlh (%) | 32 | 36 | 4 |
|  | match_steps | 166 | 185 | 19 |
|  | rotation_error (deg) | 27.15 | 34.13 | 6.98 |
|  | runtime (min) | 15 | 19 | 4 |
|  | episode_runtime (sec) | 127 | 162 | 35 |
| randomrot_rawnoise_10distinctobj_surf_agent | percent_correct (%) | 65 | 72 | 7 |
|  | used_mlh (%) | 83 | 72 | -11 |
|  | match_steps | 13 | 15 | 2 |
|  | rotation_error (deg) | 101.6 | 99.45 | -2.15 |
|  | runtime (min) | 4 | 4 | 0 |
|  | episode_runtime (sec) | 6 | 7 | 1 |
| base_10multi_distinctobj_dist_agent | percent_correct (%) | 85 | 77.14 | -7.86 |
|  | used_mlh (%) | 12.86 | 12.86 | 0 |
|  | match_steps | 27 | 33 | 6 |
|  | rotation_error (deg) | 25.3 | 22.63 | -2.67 |
|  | runtime (min) | 3 | 47 | baseline is an error |
|  | episode_runtime (sec) | 1 | 2 | 1 |
| base_77obj_dist_agent | percent_correct (%) | 90.91 | 95.24 | 4.33 |
|  | used_mlh (%) | 12.99 | 10.82 | -2.17 |
|  | match_steps | 102 | 93 | -9 |
|  | rotation_error (deg) | 14.89 | 14.66 | -0.23 |
|  | runtime (min) | 21 | 21 | 0 |
|  | episode_runtime (sec) | 61 | 60 | -1 |
| base_77obj_surf_agent | percent_correct (%) | 98.7 | 99.57 | 0.87 |
|  | used_mlh (%) | 5.19 | 7.36 | 2.17 |
|  | match_steps | 53 | 57 | 4 |
|  | rotation_error (deg) | 6.7 | 8.75 | 2.05 |
|  | runtime (min) | 15 | 13 | -2 |
|  | episode_runtime (sec) | 37 | 35 | -2 |
| randrot_noise_77obj_dist_agent | percent_correct (%) | 92.21 | 87.88 | -4.33 |
|  | used_mlh (%) | 26.84 | 25.54 | -1.3 |
|  | match_steps | 174 | 169 | -5 |
|  | rotation_error (deg) | 34.01 | 32.06 | -1.95 |
|  | runtime (min) | 41 | 40 | -1 |
|  | episode_runtime (sec) | 137 | 136 | -1 |
| randrot_noise_77obj_surf_agent | percent_correct (%) | 94.37 | 92.64 | -1.73 |
|  | used_mlh (%) | 21.65 | 25.11 | 3.46 |
|  | match_steps | 111 | 115 | 4 |
|  | rotation_error (deg) | 33.9 | 31.52 | -2.38 |
|  | runtime (min) | 34 | 37 | 3 |
|  | episode_runtime (sec) | 114 | 120 | 6 |
| randrot_noise_77obj_5lms_dist_agent | percent_correct (%) | 93.51 | 93.51 | 0 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 72 | 72 | 0 |
|  | rotation_error (deg) | 56.34 | 47.7 | -8.64 |
|  | runtime (min) | 13 | 13 | 0 |
|  | episode_runtime (sec) | 126 | 113 | -13 |
| unsupervised_inference_distinctobj_dist_agent | percent_correct (%) | 97 | 97 | 0 |
|  | match_steps | 99 | 100 | 1 |
|  | runtime (min) | 14 | 14 | 0 |
|  | episode_runtime (sec) | 8 | 9 | 1 |
| unsupervised_inference_distinctobj_surf_agent | percent_correct (%) | 97 | 93 | -4 |
|  | match_steps | 99 | 97 | -2 |
|  | runtime (min) | 25 | 21 | -4 |
|  | episode_runtime (sec) | 15 | 13 | -2 |
| infer_comp_lvl1_with_monolithic_models | correct_child_or_parent (%) | 59.52 | 59.52 | 0 |
|  | used_mlh (%) | 55.95 | 51.19 | -4.76 |
|  | match_steps | 340 | 327 | -13 |
|  | rotation_error (deg) | 66.18 | 91.19 | 25.01 |
|  | avg_prediction_error | 0.37 | 0.38 | 0.01 |
|  | runtime (min) | 27 | 31 | 4 |
| infer_comp_lvl1_with_comp_models | correct_child_or_parent (%) | 80.95 | 89.29 | 8.34 |
|  | used_mlh (%) | 53.57 | 38.1 | -15.47 |
|  | match_steps | 73 | 45 | -28 |
|  | rotation_error (deg) | 54.53 | 47.77 | -6.76 |
|  | avg_prediction_error | 0.34 | 0.33 | -0.01 |
|  | runtime (min) | 6 | 3 | -3 |
| infer_comp_lvl2_with_comp_models | correct_child_or_parent (%) | 76.19 | 80.48 | 4.29 |
|  | used_mlh (%) | 49.05 | 41.9 | -7.15 |
|  | match_steps | 43 | 36 | -7 |
|  | rotation_error (deg) | 43.37 | 43.53 | 0.16 |
|  | avg_prediction_error | 0.33 | 0.34 | 0.01 |
|  | runtime (min) | 12 | 10 | -2 |
| infer_comp_lvl3_with_comp_models | correct_child_or_parent (%) | 53.71 | 64.86 | 11.15 |
|  | used_mlh (%) | 64.57 | 52.29 | -12.28 |
|  | match_steps | 46 | 34 | -12 |
|  | rotation_error (deg) | 45.81 | 45.32 | -0.49 |
|  | avg_prediction_error | 0.32 | 0.33 | 0.01 |
|  | runtime (min) | 21 | 16 | -5 |
| infer_comp_lvl4_with_comp_models | correct_child_or_parent (%) | 52.2 | 58.52 | 6.32 |
|  | used_mlh (%) | 65.38 | 58.79 | -6.59 |
|  | match_steps | 45 | 36 | -9 |
|  | rotation_error (deg) | 47.51 | 46.39 | -1.12 |
|  | avg_prediction_error | 0.32 | 0.33 | 0.01 |
|  | runtime (min) | 22 | 19 | -3 |

tristanls-tbp:
@vkakerbeck, ~~the UI is not letting me re-request a review from you, but~~ this pull request is ready for another review now that #703 was merged and `test-integration-monty` tests were added. All the benchmarks changed as [shown in the table in the comment above](https://github.com/thousandbrainsproject/tbp.monty/pull/666#issuecomment-3775387487). The tests now pass.

conf/experiment/test/reproducibility_eval_episodes.yaml
```diff
@@ -0,0 +1,36 @@
+defaults:
+  - /experiment/test/supervised_pre_training@
+  - /experiment/config/defaults@config
+  - /experiment/config/logging/clear_logging@config
+  - /experiment/config/logging/defaults@config.logging
+  - /experiment/config/monty/clear_monty_config@config
+  - /experiment/config/monty/patch_and_view@config.monty_config
+  - /experiment/config/monty/motor_system/clear_motor_system_config@config.monty_config
+  - /experiment/test/config/monty/motor_system/fixed@config.monty_config.motor_system_config
```
vkakerbeck: If you are evaluating with a fixed action sequence, what is the potential randomness that would make the reproducibility test fail if things weren't working right? I would have expected using a random policy and then testing whether we get the same results with run and run parallel. If there is no randomness in the experiment, I wouldn't expect there to be any difference between the two, even without the new seed setting.

tristanls-tbp:
@vkakerbeck, regarding the plan for `tests/integration`, yes to what you said.

conf/experiment/test/reproducibility_eval_episodes.yaml
tristanls-tbp: Thank you, good catch, I did not notice that line before.

There is a source of randomness in `object_init_sampler` being `RandomRotation`, so the test is not entirely useless, as that randomness is reproducible. However, I'll update the action sampler to be `UniformlyDistributedSampler` so that we also have randomness there.

tristanls-tbp: Updated in https://github.com/thousandbrainsproject/tbp.monty/pull/666/changes/6412c7ed47b0661bac1f873fb6c3ae69fa254096

PR #667, opened 2025-12-31
docs: update future work item on voting

nielsleadholm:
This updates the description of the Future Work item on voting with more detail re. concrete next steps. This information is informed by the scoping and discussion in this PR: https://github.com/thousandbrainsproject/tbp.monty/pull/358


docs/future-work/voting-improvements/generalize-voting-to-associative-connections.md
```diff
@@ -19,4 +19,21 @@ However, in the setting of unsupervised learning, the object ID that is associat
...
+- Furthermore, before evaluating on our standard voting benchmarks, we learn with only a single object present. In that case, it is trivial to develop the necessary associative connections for voting. This is a good thing, because it means we can implement this change to the code almost as a refactor - even though learning of associative voting connections will technically be unsupervised, we should get the exact same performance when evaluating inference on the voting benchmarks (`randrot_noise_10distinctobj_5lms_dist_agent` and `randrot_noise_77obj_5lms_dist_agent`) as we currently do with supervised voting. This also means this we don't need to create any new learning or inference experiment configs - instead we can just replace the existing, supervised version of voting with the new approach as the default, including making sure this utilizes randomly initialized object IDs, rather than exactly matching strings.
```
vkakerbeck:
```suggestion
- Furthermore, before evaluating on our standard voting benchmarks, we learn with only a single object present. In that case, it is trivial to develop the necessary associative connections for voting. This is a good thing, because it means we can implement this change to the code almost as a refactor - even though learning of associative voting connections will technically be unsupervised, we should get the exact same performance when evaluating inference on the voting benchmarks (`randrot_noise_10distinctobj_5lms_dist_agent` and `randrot_noise_77obj_5lms_dist_agent`) as we currently do with supervised voting. This also means that we don't need to create any new learning or inference experiment configs - instead we can just replace the existing, supervised version of voting with the new approach as the default, including making sure this utilizes randomly initialized object IDs, rather than exactly matching strings.
```

docs/future-work/voting-improvements/generalize-voting-to-associative-connections.md
```diff
@@ -19,4 +19,21 @@ However, in the setting of unsupervised learning, the object ID that is associat
...
+- Currently, votes are compiled based on `graph_ids` in `send_out_vote`, and also filtered with `graph_ids` when being received and passed to `_update_evidence_with_vote`. 
+- Instead, when we compile and receive votes, these should be filtered such that if a receiving model ID has a strong association with another model ID, then the former is the key that unlocks access to the votes from the latter. In this sense, a receiving model in a receiving column will "listen" for votes coming from other models in other columns that pass a minimum association-strength threshold.
+- We would need to ensure that when receiving votes, these can come from multiple models, if multiple models surpass the threshold of associative strength.
+- It shouldn't be necessary to modify the CMP signals. Associative connections can be stored at the level of Monty, similar to the existing connectivity matrices.
```
vkakerbeck: Why would we store this at the Monty level? I would have thought of this being learned in the receiving LM. I.e. the receiving LM gets a bunch of inputs and learns how to interpret them based on their coocurrences with their own representations. So initially LM1 may get votes for object IDs that it has no idea what they mean (they could all be for the same object but since they come from different LMs, they will not be represented the same). It then associates those inputs with it's own current hypothesis (or starts strengthening the association). The next time it sees these votes, they will add a bit more evidence for the hypothesis that was active last time the inputs were received. I don't see a reason that Monty would need to get involved in this but maybe I am missing something?

vkakerbeck (commented):
Generally looks good but I am not sure it adds a lot more concrete guidance beyond what was already described on this page. What is the reason to add these points instead of working them into a more concrete RFC? What is the refinement beyond "we want to learn associations based on co-ocurrance"? Is it about describing a first step for implementing it (i.e. the one object setup)? In that case, it may be worth also outlining the next steps more, since the title of this doc refers to the general associative connection approach. I agree we could start with the one object one, but it may miss a lot of issues that start coming up once there are multiple objects present in a scene.

docs/future-work/voting-improvements/generalize-voting-to-associative-connections.md
nielsleadholm: Ok yeah, I like that approach, as it might scale better to more asynchronous/distributed versions of Monty (i.e. each LM just stores information about it's "in-edges", rather than Monty needing to track this). My suggestion emerged because the initial PR had proposed changing the CMP to store and send this information, which I don't think is appropriate. I'll update the wording.

nielsleadholm:
From my perspective there wasn't much detail in this Future Work item before, which is why I think there was some confusion about how to approach this problem in the existing public PR. I think there is a fair amount of hidden knowledge/assumptions that we have access to, but someone coming across this Future Work item wouldn't have, so being more explicit about this is helpful. It's not clear to me that most of what I wrote is superfluous. 

I had initially included the more complex multi-object setting, but when we discussed this in person, I thought you were against this, because it means creating new measurements, such as looking at the imperfect associations that emerge in the context of distractors, and creating a new "voting-with-multiple-objects" experiment config. My reading of your comment here is that you _would_ like more detail on this? In which case I'm happy to include it.

vkakerbeck:
Sorry for the confusion. I think when we talked about this, I understood it as "what's the best first experiment to test an implementation of this?" instead of "what are all the experiments that would need to be run to test this?". I agree that the 1 object scenario would be the best first step to test a prototype of this and implement the basic infrastructure for this approach. However, I think it would be worth clarifying that this isn't the only experiment. I'm not sure I would integrate a prototype of this into Monty that doesn't also demonstrate it working in multi-object scenarios, where the neighboring objects vary between episodes. Does that make sense?

nielsleadholm:
Sorry for letting this one go stale, I've updated the final description to make it clearer that replicating performance on the existing benchmark would only be a first step. 

I've also updated the description to clarify that the existing RFC would need re-writing.

vkakerbeck (approved):
Looks good, thanks for the update!

PR #668, opened 2025-12-31
ci: copyright 2026 check

tristanls-tbp:
This pull request checks any changed Python files for 2026 copyright notice.

The change to `check_changed_files.sh` is that with the new Hydra configuration, we do not need to run Monty actions when only the `benchmarks` folder changes, as it only contains CSVs used in generating documentation.

[Example check failure](https://github.com/thousandbrainsproject/tbp.monty/actions/runs/20622295047/job/59226509073):
Screenshot 2025-12-31 at 09 54 20
... given the following diff:
Screenshot 2025-12-31 at 09 55 10

### DIFF
--- a/.github/workflows/monty.yml
+++ b/.github/workflows/monty.yml
@@ -209,13 +209,11 @@ jobs:
           ruff format --check src tests conf run.py run_parallel.py
       - name: Changed files for a pull request
         if: ${{ needs.should_run_monty.outputs.should_run_monty == 'true' && github.event_name == 'pull_request' }}
-        id: changed_files_for_pull_request
         working-directory: tbp.monty
         run: |
           git diff --name-only origin/${{ github.base_ref }} > changed_files.txt
       - name: Changed files for a push
         if: ${{ needs.should_run_monty.outputs.should_run_monty == 'true' && github.event_name == 'push' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/monty.yml
```diff
@@ -205,8 +205,26 @@ jobs:
...
+        working-directory: tbp.monty
+        run: |
+          git diff --name-only ${{ github.sha }}^1 > changed_files.txt
+      - name: Check Copyright 2026
+        if: ${{ needs.should_run_monty.outputs.should_run_monty == 'true' }}
+        working-directory: tbp.monty
+        run: |
+          export PATH="$HOME/miniconda/bin:$PATH"
+          source activate tbp.monty
+          cat changed_files.txt | grep -E '^src/|^tests/|^conf/|^run.py$|^run_parallel.py$' | grep '\.py$' > changed_files_filtered.txt || exit 0
```
tristanls-tbp, line 226: **note**: If `grep` fails to find any matches, that means we don't have any files to check. When `grep` fails to find matches, it exits with `1`. We don't want that, so we `|| exit 0` in that case and do not do a ruff check on an empty list.

scottcanoe (approved):
LGTM

PR #670, opened 2026-01-01
refactor!: use possible_hyps bool array to track possible poses ids

ramyamounir:
This is part of the hypotheses resampling implementation project, which integrates `feat.dynamic_resizing` into `tbp.monty`.

@nielsleadholm @vkakerbeck This is more of an engineering solution that replaces the other symmetry detection fix, so I thought it would be better to open this PR directly into `tbp.monty` to get the IP started. But it would be great if you can take a look and let me know what you think.

Unlike [feat.dynamic_resizing:PR#3](https://github.com/thousandbrainsproject/feat.dynamic_resizing/pull/3), this PR proposes a simple approach to keeping track of possible hypotheses ids between steps. This fix is needed mainly for resampling where the position of a hypothesis in the hypotheses list is not consistent over time (i.e., hypotheses can be deleted, causing shifts in the positions of subsequent hypotheses).

Instead of manually remapping the hypotheses ids from the previous step to the current step, we track the possible hypotheses with a boolean array. We then apply the same transformations (add/remove) to this boolean array that we normally apply to the evidence/location/poses arrays during a hypotheses updater step. The boolean array is updated [here](https://github.com/thousandbrainsproject/tbp.monty/blob/dd3ee88413799836b3aba0dd9b84fea5eb183a40/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L894-L898)

## Benchmark Results

The runs are on WandB tagged under `PR#670`

**NOTE: the baseline match steps appear to be an incorrect rounding. I see the [latest run](https://wandb.ai/thousand-brains-project/Monty/runs/eq6mlbh0) 2 days ago by @tristanls-tbp reported `34.72857`, which when rounded up should become `35`. @tristanls-tbp could you confirm that it should be 35?**

| percent_correct (%) | 99.29 | 99.29 | 0 | ⬜ |
| used_mlh (%) | 0.71 | 0.71 | 0 | ⬜ |
| match_steps | 34 | 35 | 1 | ❌ |
| rotation_error (deg) | 10.69 | 10.69 | 0 | ⬜ |
| runtime (min) | 3 | 2 | -1 | ✅ |
| episode_runtime (sec) | 9 | 9 | 0 | ⬜ |

| percent_correct (%) | 99.29 | 99.29 | 0 | ⬜ |
| match_steps | 27 | 27 | 0 | ⬜ |
| rotation_error (deg) | 11.96 | 11.96 | 0 | ⬜ |
| runtime (min) | 3 | 2 | -1 | ✅ |
| episode_runtime (sec) | 11 | 10 | -1 | ✅ |

| percent_correct (%) | 97 | 97 | 0 | ⬜ |
| used_mlh (%) | 5 | 5 | 0 | ⬜ |
| match_steps | 55 | 55 | 0 | ⬜ |
| rotation_error (deg) | 25.14 | 25.14 | 0 | ⬜ |
| runtime (min) | 5 | 5 | 0 | ⬜ |
| episode_runtime (sec) | 28 | 31 | 3 | ❌ |

| percent_correct (%) | 98 | 98 | 0 | ⬜ |
| used_mlh (%) | 3 | 3 | 0 | ⬜ |
| match_steps | 38 | 38 | 0 | ⬜ |
| rotation_error (deg) | 15.11 | 15.11 | 0 | ⬜ |
| runtime (min) | 4 | 4 | 0 | ⬜ |
| episode_runtime (sec) | 18 | 24 | 6 | ❌ |

| used_mlh (%) | 1 | 1 | 0 | ⬜ |
| rotation_error (deg) | 25.14 | 25.14 | 0 | ⬜ |
| runtime (min) | 3 | 3 | 0 | ⬜ |
| episode_runtime (sec) | 20 | 21 | 1 | ❌ |

| rotation_error (deg) | 13.46 | 13.46 | 0 | ⬜ |
| runtime (min) | 2 | 2 | 0 | ⬜ |
| episode_runtime (sec) | 13 | 12 | -1 | ✅ |

| match_steps | 52 | 52 | 0 | ⬜ |
| rotation_error (deg) | 49.54 | 49.54 | 0 | ⬜ |
| runtime (min) | 5 | 5 | 0 | ⬜ |
| episode_runtime (sec) | 40 | 42 | 2 | ❌ |

### base_10simobj_surf_agent

| percent_correct (%) | 92.14 | 92.14 | 0 | ⬜ |
| used_mlh (%) | 10.71 | 10.71 | 0 | ⬜ |
| match_steps | 83 | 83 | 0 | ⬜ |
| rotation_error (deg) | 11 | 11 | 0 | ⬜ |
| runtime (min) | 7 | 6 | -1 | ✅ |
| episode_runtime (sec) | 34 | 31 | -3 | ✅ |

| percent_correct (%) | 78 | 78 | 0 | ⬜ |
| used_mlh (%) | 36 | 36 | 0 | ⬜ |
| match_steps | 201 | 201 | 0 | ⬜ |
| rotation_error (deg) | 21.46 | 21.46 | 0 | ⬜ |
| runtime (min) | 11 | 10 | -1 | ✅ |
| episode_runtime (sec) | 92 | 85 | -7 | ✅ |

| percent_correct (%) | 90 | 90 | 0 | ⬜ |
| used_mlh (%) | 32 | 32 | 0 | ⬜ |
| match_steps | 166 | 166 | 0 | ⬜ |
| rotation_error (deg) | 27.15 | 27.15 | 0 | ⬜ |
| runtime (min) | 17 | 15 | -2 | ✅ |
| episode_runtime (sec) | 140 | 127 | -13 | ✅ |

| percent_correct (%) | 65 | 65 | 0 | ⬜ |
| used_mlh (%) | 83 | 83 | 0 | ⬜ |
| match_steps | 13 | 13 | 0 | ⬜ |
| rotation_error (deg) | 101.6 | 101.6 | 0 | ⬜ |
| runtime (min) | 4 | 4 | 0 | ⬜ |
| episode_runtime (sec) | 5 | 6 | 1 | ❌ |

| percent_correct (%) | 85 | 85 | 0 | ⬜ |
| used_mlh (%) | 12.86 | 12.86 | 0 | ⬜ |
| match_steps | 27 | 27 | 0 | ⬜ |
| rotation_error (deg) | 25.3 | 25.3 | 0 | ⬜ |
| runtime (min) | 3 | 3 | 0 | ⬜ |
| episode_runtime (sec) | 1 | 1 | 0 | ⬜ |

### base_77obj_dist_agent

| percent_correct (%) | 90.91 | 90.91 | 0 | ⬜ |
| used_mlh (%) | 12.99 | 12.99 | 0 | ⬜ |
| match_steps | 102 | 102 | 0 | ⬜ |
| rotation_error (deg) | 14.89 | 14.89 | 0 | ⬜ |
| runtime (min) | 20 | 21 | 1 | ❌ |
| episode_runtime (sec) | 58 | 61 | 3 | ❌ |

### base_77obj_surf_agent

| percent_correct (%) | 98.7 | 98.7 | 0 | ⬜ |
| used_mlh (%) | 5.19 | 5.19 | 0 | ⬜ |
| match_steps | 53 | 53 | 0 | ⬜ |
| rotation_error (deg) | 6.7 | 6.7 | 0 | ⬜ |
| runtime (min) | 14 | 15 | 1 | ❌ |
| episode_runtime (sec) | 35 | 37 | 2 | ❌ |

| percent_correct (%) | 92.21 | 92.21 | 0 | ⬜ |
| used_mlh (%) | 26.84 | 26.84 | 0 | ⬜ |
| match_steps | 174 | 174 | 0 | ⬜ |
| rotation_error (deg) | 34.01 | 34.01 | 0 | ⬜ |
| runtime (min) | 41 | 41 | 0 | ⬜ |
| episode_runtime (sec) | 137 | 137 | 0 | ⬜ |

| percent_correct (%) | 94.37 | 94.37 | 0 | ⬜ |
| used_mlh (%) | 21.65 | 21.65 | 0 | ⬜ |
| match_steps | 111 | 111 | 0 | ⬜ |
| rotation_error (deg) | 33.9 | 33.9 | 0 | ⬜ |
| runtime (min) | 36 | 34 | -2 | ✅ |
| episode_runtime (sec) | 115 | 114 | -1 | ✅ |

| percent_correct (%) | 93.51 | 93.51 | 0 | ⬜ |
| match_steps | 72 | 72 | 0 | ⬜ |
| rotation_error (deg) | 56.34 | 56.34 | 0 | ⬜ |
| runtime (min) | 13 | 13 | 0 | ⬜ |
| episode_runtime (sec) | 119 | 126 | 7 | ❌ |

| percent_correct (%) | 99 | 97 | -2 | ❌ |
| match_steps | 100 | 99 | -1 | ✅ |
| runtime (min) | 26 | 14 | -12 | ✅ |
| episode_runtime (sec) | 15 | 8 | -7 | ✅ |

| percent_correct (%) | 96 | 97 | 1 | ✅ |
| match_steps | 100 | 99 | -1 | ✅ |
| runtime (min) | 21 | 25 | 4 | ❌ |
| episode_runtime (sec) | 12 | 15 | 3 | ❌ |

### infer_comp_lvl1_with_monolithic_models

| correct_child_or_parent (%) | 59.52 | 59.52 | 0 | ⬜ |
| used_mlh (%) | 55.95 | 55.95 | 0 | ⬜ |
| match_steps | 340 | 340 | 0 | ⬜ |
| rotation_error (deg) | 66.18 | 66.18 | 0 | ⬜ |
| avg_prediction_error | 0.37 | 0.37 | 0 | ⬜ |
| runtime (min) | 30 | 27 | -3 | ✅ |

### infer_comp_lvl1_with_comp_models

| correct_child_or_parent (%) | 80.95 | 80.95 | 0 | ⬜ |
| used_mlh (%) | 53.57 | 53.57 | 0 | ⬜ |
| match_steps | 73 | 73 | 0 | ⬜ |
| rotation_error (deg) | 54.53 | 54.53 | 0 | ⬜ |
| avg_prediction_error | 0.34 | 0.34 | 0 | ⬜ |
| runtime (min) | 6 | 6 | 0 | ⬜ |

### infer_comp_lvl2_with_comp_models

| correct_child_or_parent (%) | 76.19 | 76.19 | 0 | ⬜ |
| used_mlh (%) | 49.05 | 49.05 | 0 | ⬜ |
| match_steps | 43 | 43 | 0 | ⬜ |
| rotation_error (deg) | 43.37 | 43.37 | 0 | ⬜ |
| avg_prediction_error | 0.33 | 0.33 | 0 | ⬜ |
| runtime (min) | 12 | 12 | 0 | ⬜ |

### infer_comp_lvl3_with_comp_models

| correct_child_or_parent (%) | 53.71 | 53.71 | 0 | ⬜ |
| used_mlh (%) | 64.57 | 64.57 | 0 | ⬜ |
| match_steps | 46 | 46 | 0 | ⬜ |
| rotation_error (deg) | 45.8 | 45.81 | 0.01 | ❌ |
| avg_prediction_error | 0.32 | 0.32 | 0 | ⬜ |
| runtime (min) | 22 | 21 | -1 | ✅ |

### infer_comp_lvl4_with_comp_models

| correct_child_or_parent (%) | 52.2 | 52.2 | 0 | ⬜ |
| used_mlh (%) | 65.38 | 65.38 | 0 | ⬜ |
| match_steps | 45 | 45 | 0 | ⬜ |
| rotation_error (deg) | 47.51 | 47.51 | 0 | ⬜ |
| avg_prediction_error | 0.32 | 0.32 | 0 | ⬜ |
| runtime (min) | 23 | 22 | -1 | ✅ |

### DIFF
--- a/.github/workflows/monty.yml
+++ b/.github/workflows/monty.yml
@@ -412,6 +412,8 @@ jobs:
         continue-on-error: true
         run: |
           uv venv --seed
+          # Need to install torch separately because torch-scatter doesn't declare it properly
+          uv pip install torch==1.13.1
           uv sync --locked --extra dev --extra simulator_mujoco
       - name: Run tests
         if: ${{ needs.should_run_monty.outputs.should_run_monty == 'true' }}
--- a/UV_PROTOTYPE.md
+++ b/UV_PROTOTYPE.md
@@ -10,5 +10,6 @@ Some notes on how to set up this environment.
 ```sh
 # The --seed is needed so we can build the torch packages
 uv venv -p 3.9.22 --seed
+uv pip install torch==1.13.1 # Use the version from pyproject.toml
 uv sync --extra dev --extra simulator_mujoco
 ```
--- a/src/tbp/monty/frameworks/config_utils/make_env_interface_configs.py
+++ b/src/tbp/monty/frameworks/config_utils/make_env_interface_configs.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
 # Copyright 2022-2024 Numenta Inc.
 #
 # Copyright may exist in Contributors' modifications
@@ -9,13 +9,15 @@
 # https://opensource.org/licenses/MIT.
 from __future__ import annotations
 
+from typing import Literal
+
 import numpy as np
 
 
 def make_sensor_positions_on_grid(
     n_sensors: int,
     delta: float = 0.01,
-    order_by: str = "distance",
+    order_by: Literal["distance", "spiral"] = "distance",
     add_view_finder: bool = True,
 ) -> np.ndarray:
     """Generate sensor positions on a 2D grid.
@@ -113,13 +115,12 @@ def make_sensor_positions_on_grid(
                 if current_dir % 2 == 0:
                     steps += 1
 
-    indices = np.array(indices)[:n_sensors]
+    indices = indices[:n_sensors]
 
     # Convert indices to locations in agent space.
-    positions = []
-    for idx in indices:
-        positions.append((x[idx[0], idx[1]] * delta, y[idx[0], idx[1]] * delta))
-    positions = np.array(positions)
+    positions = np.array(
+        [[x[idx[0], idx[1]] * delta, y[idx[0], idx[1]] * delta] for idx in indices]
+    )
 
     # Add z-positions.
     positions = np.hstack((positions, np.zeros([positions.shape[0], 1])))
--- a/src/tbp/monty/frameworks/loggers/monty_handlers.py
+++ b/src/tbp/monty/frameworks/loggers/monty_handlers.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
 # Copyright 2022-2024 Numenta Inc.
 #
 # Copyright may exist in Contributors' modifications
@@ -44,8 +44,9 @@ class MontyHandler(metaclass=abc.ABCMeta):
     def close(self):
         pass
 
-    @abc.abstractclassmethod
-    def log_level(self):
+    @classmethod
+    @abc.abstractmethod
+    def log_level(cls):
         """Handlers filter information from the data they receive.
 
         This class method specifies the level they filter at.
--- a/src/tbp/monty/frameworks/models/buffer.py
+++ b/src/tbp/monty/frameworks/models/buffer.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
 # Copyright 2022-2024 Numenta Inc.
 #
 # Copyright may exist in Contributors' modifications
@@ -575,7 +575,7 @@ class FeatureAtLocationBuffer:
 class BufferEncoder(json.JSONEncoder):
     """Encoder to turn the buffer into a JSON compliant format."""
 
-    _encoders: ClassVar[dict[type, Callable | json.JSONEncoder]] = {}
+    _encoders: ClassVar[dict[type, Callable]] = {}
 
     @classmethod
     def register(
@@ -659,7 +659,7 @@ class BufferEncoder(json.JSONEncoder):
         encoder = self._find(obj)
         if encoder is not None:
             return encoder(obj)
-        return json.JSONEncoder.default(self, obj)
+        return super().default(obj)
 
 
 BufferEncoder.register(np.generic, lambda obj: obj.item())
--- a/src/tbp/monty/frameworks/run_parallel.py
+++ b/src/tbp/monty/frameworks/run_parallel.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
 # Copyright 2022-2024 Numenta Inc.
 #
 # Copyright may exist in Contributors' modifications
@@ -612,7 +612,7 @@ def run_episodes_parallel(
         )
     print(f"Wandb setup took {time.time() - start_time} seconds")
     start_time = time.time()
-    with mp.Pool(num_parallel) as p:
+    with mp.Pool(num_parallel, maxtasksperchild=1) as p:
         if train:
             # NOTE: since we don't use wandb logging for training right now
             # it is also not covered here. Might want to add that in the future.
--- a/src/tbp/monty/frameworks/utils/evidence_matching.py
+++ b/src/tbp/monty/frameworks/utils/evidence_matching.py
@@ -9,7 +9,6 @@
 from __future__ import annotations
 
 from collections import OrderedDict
-from typing import OrderedDict as OrderedDictType
 
 import numpy as np
 import numpy.typing as npt
@@ -39,7 +38,7 @@ class ChannelMapper:
         Args:
             channel_sizes: Dictionary of {channel_name: size}.
         """
-        self.channel_sizes: OrderedDictType[str, int] = (
+        self.channel_sizes: OrderedDict[str, int] = (
             OrderedDict(channel_sizes) if channel_sizes else OrderedDict()
         )
 
@@ -87,14 +86,13 @@ class ChannelMapper:
         Raises:
             ValueError: If the channel is not found.
         """
-        if channel_name not in self.channel_sizes:
-            raise ValueError(f"Channel '{channel_name}' not found.")
-
         start = 0
         for name, size in self.channel_sizes.items():
             if name == channel_name:
                 return (start, start + size)
             start += size
+
+        raise ValueError(f"Channel '{channel_name}' not found.")
 
     def resize_channel_by(self, channel_name: str, value: int) -> None:
         """Increases or decreases the channel by a specific amount.
@@ -499,7 +497,7 @@ def evidence_update_threshold(
     if evidence_all_channels.size == 0:
         return 0.0
 
-    if type(evidence_threshold_config) in [int, float]:
+    if isinstance(evidence_threshold_config, (int, float)):
         return evidence_threshold_config
 
     if evidence_threshold_config == "mean":
@@ -520,7 +518,7 @@ def evidence_update_threshold(
         return max_global_evidence - x_percent_of_max
 
     if evidence_threshold_config == "x_percent_threshold":
-        x_percent_of_max = max_global_evidence / 100 * x_percent_threshold
+        x_percent_of_max = max_global_evidence / 100 * float(x_percent_threshold)
         return max_global_evidence - x_percent_of_max
 
     if evidence_threshold_config == "all":
--- a/src/tbp/monty/frameworks/utils/graph_matching_utils.py
+++ b/src/tbp/monty/frameworks/utils/graph_matching_utils.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
 # Copyright 2023-2024 Numenta Inc.
 #
 # Copyright may exist in Contributors' modifications
@@ -469,7 +469,7 @@ def find_step_on_new_object(
 
 def possible_sensed_directions(
     sensed_directions: np.ndarray, num_hyps_per_node: int
-) -> np.ndarray:
+) -> list[np.ndarray]:
     """Returns the possible sensed directions for all nodes.
 
     This function determines the possible sensed directions for a given set of sensed
--- a/src/tbp/monty/frameworks/utils/logging_utils.py
+++ b/src/tbp/monty/frameworks/utils/logging_utils.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
 # Copyright 2023-2024 Numenta Inc.
 #
 # Copyright may exist in Contributors' modifications
@@ -383,9 +383,10 @@ def get_time_stats(all_ds, all_conditions) -> pd.DataFrame:
                         ]
                     )
                 )
-    time_stats = pd.concat(time_stats, axis=1).T
-    time_stats.columns = ["model_type", "time", "step"]
-    return time_stats
+
+    df = pd.concat(time_stats, axis=1).T
+    df.columns = ["model_type", "time", "step"]
+    return df
 
 
 def compute_pose_errors(
@@ -440,10 +441,11 @@ def compute_pose_error(
     Returns:
         The minimum angular error in radians (or degrees if `return_degrees` is True).
     """
-    rotation_error = np.min(compute_pose_errors(predicted_rotation, target_rotation))
-    if return_degrees:
-        return rotation_error * 180 / np.pi
-    return rotation_error
+    rotation_error: float = np.min(  # type: ignore[assignment]
+        compute_pose_errors(predicted_rotation, target_rotation)
+    )
+
+    return rotation_error * 180 / np.pi if return_degrees else rotation_error
 
 
 def print_overall_stats(stats):
--- a/uv.lock
+++ b/uv.lock
@@ -86,20 +86,6 @@ wheels = [
 ]
 
 [[package]]
-name = "aicrowd-api"
-version = "0.1.25"
-source = { registry = "https://pypi.org/simple" }
-dependencies = [
-    { name = "redis", version = "6.1.1", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.9'" },
-    { name = "redis", version = "6.2.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.9'" },
-    { name = "requests" },
-]
-sdist = { url = "https://files.pythonhosted.org/packages/bd/02/893a50e8888a305eacb96636512936153415d4b0bc24b32a684d37bde95a/aicrowd_api-0.1.25.tar.gz", hash = "sha256:5c57c6785ccacce335b627e9534a6c8d0bc19a475032dbfeed4516598797aa94", size = 11317, upload-time = "2024-02-07T21:34:39.415Z" }
-wheels = [
-    { url = "https://files.pythonhosted.org/packages/2a/09/b66f1caaddae133e9dc09240c1208a93d1380f29e526280dfb2d9d2de33d/aicrowd_api-0.1.25-py2.py3-none-any.whl", hash = "sha256:ff9057e57e1ffc6bc65b8ab8ba37b12e5dba3b7e0fe9c7265fc5f90376949fe6", size = 10117, upload-time = "2024-02-07T21:34:36.768Z" },
-]
-
-[[package]]
 name = "alabaster"
 version = "0.7.13"
 source = { registry = "https://pypi.org/simple" }
@@ -156,6 +142,12 @@ wheels = [
 ]
 
 [[package]]
+name = "antlr4-python3-runtime"
+version = "4.9.3"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/3e/38/7859ff46355f76f8d19459005ca000b6e7012f2f1ca597746cbcd1fbfe5e/antlr4-python3-runtime-4.9.3.tar.gz", hash = "sha256:f224469b4168294902bb1efa80a8bf7855f24c99aef99cbefc1bcd3cce77881b", size = 117034, upload-time = "2021-11-06T17:52:23.524Z" }
+
+[[package]]
 name = "anyio"
 version = "4.9.0"
 source = { registry = "https://pypi.org/simple" }
@@ -186,15 +178,6 @@ source = { registry = "https://pypi.org/simple" }
 sdist = { url = "https://files.pythonhosted.org/packages/4a/e7/82da0a03e7ba5141f05cce0d302e6eed121ae055e0456ca228bf693984bc/asttokens-3.0.0.tar.gz", hash = "sha256:0dcd8baa8d62b0c1d118b399b2ddba3c4aff271d0d7a9e0d4c1681c79035bbc7", size = 61978, upload-time = "2024-11-30T04:30:14.439Z" }
 wheels = [
     { url = "https://files.pythonhosted.org/packages/25/8a/c46dcc25341b5bce5472c718902eb3d38600a903b14fa6aeecef3f21a46f/asttokens-3.0.0-py3-none-any.whl", hash = "sha256:e3078351a059199dd5138cb1c706e6430c05eff2ff136af5eb4790f9d28932e2", size = 26918, upload-time = "2024-11-30T04:30:10.946Z" },
-]
-
-[[package]]
-name = "async-timeout"
-version = "5.0.1"
-source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/a5/ae/136395dfbfe00dfc94da3f3e136d0b13f394cba8f4841120e34226265780/async_timeout-5.0.1.tar.gz", hash = "sha256:d9321a7a3d5a6a5e187e824d2fa0793ce379a202935782d555d6e9d2735677d3", size = 9274, upload-time = "2024-11-06T16:41:39.6Z" }
-wheels = [
-    { url = "https://files.pythonhosted.org/packages/fe/ba/e2081de779ca30d473f21f5b30e0e737c438205440784c7dfc81efc2b029/async_timeout-5.0.1-py3-none-any.whl", hash = "sha256:39e3809566ff85354557ec2398b55e096c8364bacac9405a7a1fa429e77fe76c", size = 6233, upload-time = "2024-11-06T16:41:37.9Z" },
 ]
 
 [[package]]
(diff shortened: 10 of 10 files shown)


src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -519,20 +524,24 @@ def get_unique_pose_if_available(self, object_id):
         """
         possible_object_hypotheses_ids = self.get_possible_hypothesis_ids(object_id)
         # Only try to determine object pose if the evidence for it is high enough.
-        if possible_object_hypotheses_ids is not None:
+        if len(possible_object_hypotheses_ids):
```
nielsleadholm: This is personal preference, but I prefer a bit more verbosity in these kinds of cases, e.g. `if len(possible_object_hypotheses_ids) > 0:`, even if they are functionally the same. I feel it makes the code more readable and reduces uncertainty about the intended/desired behavior.

nielsleadholm (commented):
Nice, thanks for putting this together. Just a few high level comments to begin with:

1. Do you know why the unsupervised inference results have changed?
2. Is there any coverage of this change in the unit tests? I'm sure it's right, but I'm struggling to 100% visualize how this is working. In particular, how does `self.possible_hyps[graph_id] = mapper.update` not break the correspondence between the boolean arrays on two subsequent steps? Can you remind me please what the values of these arrays looks like, e.g. if a hypothesis's _index_ in the array shifts, can its _ID_ remain the same?

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
tristanls-tbp: **suggestion**: [I'm surprised that `ruff` didn't flag it](https://docs.astral.sh/ruff/rules/len-test/), but I think this should be `if possible_object_hypotheses_ids:`

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -222,11 +223,16 @@ def __init__(
...
         # length as self.evidence and store the corresponding hypotheses.
-        self.evidence: dict[str, np.ndarray] = {}
-        self.possible_locations: dict[str, np.ndarray] = {}
-        self.possible_poses: dict[str, np.ndarray] = {}
+        self.possible_locations: dict[str, npt.NDArray[np.float64]] = {}
+        self.possible_poses: dict[str, npt.NDArray[np.float64]] = {}
+
+        # self.possible_hyps is a boolean array with the same structure as self.evidence
+        # and stores which hypotheses are "possible" for symmetry checks.
+        self.possible_hyps: dict[str, npt.NDArray[np.bool_]] = {}
```
tristanls-tbp, line 235: **suggestion**: Document in a `docstring`.
```suggestion
        self.possible_hyps: dict[str, npt.NDArray[np.bool_]] = {}
        """A boolean array with the same structure as self.evidence.
        
        Stores which hypotheses are "possible" for symmetery checks."""
```

tristanls-tbp:
@ramyamounir, yes, it should be `35`. It looks like I misread the [`base_config_10distinctobj_dist_agent` benchmark data](https://wandb.ai/thousand-brains-project/Monty/runs/b3c21nki/overview) when I updated.

ramyamounir:
@nielsleadholm 

1. Yes, the current symmetry detection implementation for unsupervised inference in `tbp.monty` is not correct. The fix was proposed in `feat.dynamic_resizing` but never made it to `tbp.monty`, it's part of the IP. So, this PR fixes the symmetry detection for unsupervised inference and leaves the other benchmarks unchanged (as shown in the benchmark runs).
2. I can do a better job explaining this. Instead of storing a list of ids (e.g., [2, 3, 5]), we store a boolean array (e.g., `[0, 0, 1, 1, 0, 1]`), mimicking the structure of `self.evidence`. This allows us to delete the stored value along with the hypotheses other values when pruning. The array `self.possible_hyps[graph_id]` shadows `self.evidence[graph_id]` in all resampling (add/delete) operations, so that we don't have to do it manually later on. When using `ResamplingHypothesesUpdater`:
    - The boolean array entries are created along with other values (e.g., evidence, location, rotation, age, etc.). This happens [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e8a6414036fafdc7b6302ea6e7a3b371af4f8020/src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py?plain=1#L649-L650)
    - If a hypotheses is deleted, the entry is also deleted along with the other values (e.g., evidence, location, rotation, etc.). This happens [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e8a6414036fafdc7b6302ea6e7a3b371af4f8020/src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py?plain=1#L519)
    - When the updater returns the new array, the channel mapper updates values for `possible_hyps` similar to how it updates the `evidence` values. This happens [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e8a6414036fafdc7b6302ea6e7a3b371af4f8020/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L894-L898)
    - When checking for symmetry, I convert from the boolean array to ids [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e8a6414036fafdc7b6302ea6e7a3b371af4f8020/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L534-L536), then compare against the [current ids](https://github.com/thousandbrainsproject/tbp.monty/blob/e8a6414036fafdc7b6302ea6e7a3b371af4f8020/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L1047-L1050). The current ids are converted back to boolean array and stored in `self.possible_hyps` [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e8a6414036fafdc7b6302ea6e7a3b371af4f8020/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L543-L544)
    - The [symmetry test](https://github.com/thousandbrainsproject/tbp.monty/blob/e8a6414036fafdc7b6302ea6e7a3b371af4f8020/tests/unit/evidence_lm_test.py?plain=1#L539-L589) for evidence lm should still cover this, also the benchmarks reported the same value showing that symmetry detection still works the same when we use this boolean mask approach to store the ids.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
ramyamounir: I was following the style that already exists in [here](https://github.com/thousandbrainsproject/tbp.monty/blob/e8a6414036fafdc7b6302ea6e7a3b371af4f8020/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py?plain=1#L216-L238). Should I change all the other comments as well (e.g., `self.evidence`, `self.possible_locations`, `self.possible_poses`)?

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
ramyamounir: I wasn't aware that this is bad. Ruff says:

> You can either remove the call to len or compare the length against a scalar.

Since @nielsleadholm prefers explicit comparison and either is accepted by ruff, I will change it to `if len(possible_object_hypotheses_ids) > 0` unless you prefer the other way @tristanls-tbp. Either way is fine by me.

ramyamounir: Changed in [b9821fd8d498](https://github.com/ramyamounir/tbp.monty/commit/b9821fd8d498fc242f66012158c0321811dde5ad)

tristanls-tbp: The Python idiomatic way is `if possible_object_hypotheses_ids`

tristanls-tbp: The `== 0` idiomatic check in Python is `if not possible_object_hypotheses_ids`

nielsleadholm (approved):
Thanks Ramy for clarifying those points, really helpful. Overall looks great.

The only other comment I have, which relates to something your PR did not introduce, and so which could be a follow-up item: I've always found the use of the term "possible" here a bit confusing. A better word would probably be "probable" - by definition, all of the hypotheses are "possible" - but some are more likely, which is what we mean in this context. Curious if you and @vkakerbeck also feel that updating the wording would be clearer for newcomers.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
jeremyshoemaker: It also unlocks an optimization in the CPython interpreter if it knows that it's a list. I've been suggesting it everywhere I see it, but there are a few places where we're doing this on NumPy arrays, which **do not** like to be checked for length this way.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses.py
```diff
@@ -19,9 +20,10 @@ class Hypotheses:
     hypothesis.
     """
 
-    evidence: np.ndarray
-    locations: np.ndarray
-    poses: np.ndarray
+    evidence: npt.NDArray[np.float64]
+    locations: npt.NDArray[np.float64]
+    poses: npt.NDArray[np.float64]
+    possible: npt.NDArray[np.bool_]
```
jeremyshoemaker, line 26: Aside from whether you do the name change that @nielsleadholm suggested, I've found that booleans, whether they be single boolean variables, or arrays like here, are more obvious if they're named something like `is_*`, e.g. `is_possible`, or `has_*` if that makes more sense (which doesn't really here). 

The word "possible" by itself seems like there's something missing in the name. I find myself going, "possible what? oh, possible hypothesis". Swapping the word order around you get "hypothesis is possible" -> `is_possible`.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
jeremyshoemaker: A docstring doesn't make sense in the middle of a constructor or any other method. It would just be a bare string that doesn't get assigned to a variable.

src/tbp/monty/frameworks/models/graph_matching.py
```diff
@@ -709,6 +709,15 @@ def propose_goal_states(self) -> list[GoalState]:
 
         return []
 
+    def _clear_possible_hyps(self):
+        """Clears the possible hypotheses by setting all hypotheses values to False."""
+        possible_hyps = getattr(self, "possible_hyps", None)
```
jeremyshoemaker: Why not set `self.possible_hyps` to `None` in the constructor and use it directly? Then you don't need to use `getattr()`.

I tend to view `getattr` as a code smell, because the string argument obscures the connection between the definition and use of the thing being looked up. It makes it harder for IDEs and other tools to follow the connection.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
jeremyshoemaker: Having reviewed the code, I think all the places this _could_ be done here are all NumPy arrays which don't work with this sort of idiomatic test for length.

ramyamounir:
@nielsleadholm 

> I've always found the use of the term "possible" here a bit confusing.

Yes, I agree. There are `possible_{locations, poses}` and there are `possible_hyps` now. The word `possible` refers to two completely different things in the two cases. `probable` is a good idea, but I was also thinking we could drop the `possible` in `self.possible_{locations, poses}`, and instead maybe call them `self.hypotheses_{locations, poses, evidence}` and `self.is_possible_hypotheses` for the symmetry boolean array, to incorporate @jeremyshoemaker's suggestion. All of the existing hypotheses are technically possible (we do not have impossible hypotheses), so we don't really need to label them as possible. But yeah as you mentioned, this could be a follow-up refactor.

src/tbp/monty/frameworks/models/graph_matching.py
ramyamounir: The parent class (`GraphLM`) doesn't really have a notion of evidence (locations, poses, evidence, or `possible_hyps`). The evidence and their values were really introduced in the child class `EvidenceGraphLM`. So, it felt wrong specifying an evidence related state in the parent class. The `getattr` is really checking if this is an instance of `EvidenceGraphLM` or `GraphLM`. I think the ideal solution here would be to work on extracting the symmetry detection logic into `GraphLMSymmetryDetector` and `EvidenceLMSymmetryDetector`? Or perhaps overriding the `update_terminal_condition` in the child class?

Note that before this PR, we were just [setting](https://github.com/thousandbrainsproject/tbp.monty/blob/37613dbdd7e21fcc1ae43ceec8163c24da424405/src/tbp/monty/frameworks/models/graph_matching.py?plain=1#L721) `self.last_possible_hypotheses=None` in `update_terminal_condition`, which was correctly resetting the `self.last_possible_hypotheses` in the child class `EvidenceGraphLM`, but creating new state in the parent class `GraphLM`. I doubt that we use `GraphLM` by itself in any experiment but there are a bunch of unit tests that do, so I had to do this.

jeremyshoemaker: If this method is accessing things that don't exist in this class, then it doesn't belong here. I think your suggestion of overriding `update_terminal_condition` in a child class is the correct approach. If the order matters, have it call `self._clear_possible_hyps` before calling `super` in the child version.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
tristanls-tbp: Well, since NumPy is gonna NumPy, shouldn't the check be:
```python
if possible_object_hypotheses_ids.size > 0:
```
Because
```python
>>> arr = np.array([[], [], []])
>>> len(arr)
3
>>> arr.size
0
```

tristanls-tbp: But... `len(possible_object_hypotheses_ids)` will return the length of the 1st dimension... so it should work in all these cases. So I guess `len(possible_object_hypotheses_ids) > 0` it is.

src/tbp/monty/frameworks/models/graph_matching.py
ramyamounir: Updated in [eb6ced3e5ee6](https://github.com/ramyamounir/tbp.monty/commit/eb6ced3e5ee63763faabcc342e9d212ebe3bf1ab)

I also ran 2 experiments just to double check that the changes didn't affect the results. 

| percent_correct (%) | 92.21 | 92.21 | 0 | ⬜ |
| used_mlh (%) | 26.84 | 26.84 | 0 | ⬜ |
| match_steps | 174 | 174 | 0 | ⬜ |
| rotation_error (deg) | 34.01 | 34.01 | 0 | ⬜ |
| runtime (min) | 41 | 41 | 0 | ⬜ |
| episode_runtime (sec) | 137 | 137 | 0 | ⬜ |

| percent_correct (%) | 94.37 | 94.37 | 0 | ⬜ |
| used_mlh (%) | 21.65 | 21.65 | 0 | ⬜ |
| match_steps | 111 | 111 | 0 | ⬜ |
| rotation_error (deg) | 33.9 | 33.9 | 0 | ⬜ |
| runtime (min) | 34 | 37 | 3 | ❌ |
| episode_runtime (sec) | 114 | 121 | 7 | ❌ |

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
tristanls-tbp: >  It would just be a bare string that doesn't get assigned to a variable.

False.

Image

Also, appears in Google docstring examples:
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

Image

Also, our API docs will include it

Image

ramyamounir: Thanks @tristanls-tbp, I didn't know sphinx parses the docstring like that. Do we need to include the type in the docstring as they show? Or just a description is fine?
```python
        self.attr5 = None
        """str: Docstring *after* attribute, with type specified."""
```
I will update all the other attributes for this class in this commit as shown below:

```python
        self.evidence: dict[str, npt.NDArray[np.float64]] = {}
        """Dictionary with graph_ids as keys.

        For each graph we initialize a set of hypotheses at the first step of an
        episode. Each hypothesis has an evidence count associated with it which is
        stored here."""

        self.possible_locations: dict[str, npt.NDArray[np.float64]] = {}
        """Float arrays with the same dictionary structure as self.evidence.

        Stores the current locations of hypotheses."""

        self.possible_poses: dict[str, npt.NDArray[np.float64]] = {}
        """Float arrays with the same dictionary structure as self.evidence.

        Stores the object pose for each hypothesis."""

        self.possible_hyps: dict[str, npt.NDArray[np.bool_]] = {}
        """Boolean arrays with the same dictionary structure as self.evidence.

        Stores which hypotheses are "possible" for symmetry checks."""
```

jeremyshoemaker: From what I can find, that seems to be something specific to doc generation tools, and not something built into Python. That doesn't mean we shouldn't use them, just that they don't work the same way as function and class docstrings (which actually get stored as an attribute on the thing they document).

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -1045,6 +1073,24 @@ def _enough_symmetry_evidence_accumulated(self):
...
+        Returns:
+            Terminal state of the LM.
+        """
+        if len(self.get_possible_matches()) != 1:
+            self._clear_possible_hyps()
+
+        super().update_terminal_condition()
+
+        return self.terminal_state
+
```
jeremyshoemaker: This is better, though I have a few further suggestions.

Since we're only using `_clear_possible_hyps` in one place, and it's so short, we should probably just inline it to aid readability. Having it separated like this means you have to look in multiple places to completely understand what the code in `update_terminal_condition` is doing.

Also, `super().update_terminal_condition()` returns `self.terminal_state` so you can just return that directly instead of having a separate return statement that might get out of sync with the base version.

```suggestion
    def update_terminal_condition(self):
        """Check if we have reached a terminal condition for this episode.

        Returns:
            Terminal state of the LM.
        """
        if len(self.get_possible_matches()) != 1:
            # Clears the possible hypotheses by setting all hypotheses values to False.
            for mask in self.possible_hyps.values():
                mask[:] = False
        return super().update_terminal_condition()

```

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
jeremyshoemaker: I think it would make more sense to use the `#:` comments preceding the variable instead of multiline strings aftewards. The main reason being that a lot of editors and syntax highlighters won't treat multiline strings outside of function and class docstrings as comments, but as strings making the highlighting mismatch the rest of the comments and docstrings.

tristanls-tbp: > Do we need to include the type in the docstring as they show? Or just a description is fine?

The description is fine. A while back @jeremyshoemaker added a plugin that pulls types from the code into the documentation.

If you want the type in this type of docstring, it appears we need to annotate the instance attribute:
```python
self.agent_id: AgentID = agent_id
"""I get assigned to self.agent_id"""
```
which will result in:
Screenshot 2026-01-06 at 10 01 02

jeremyshoemaker: Here's an example of what I mean from the latest version of PyCharm:
image

vs:

image

vkakerbeck:
> The only other comment I have, which relates to something your PR did not introduce, and so which could be a follow-up item: I've always found the use of the term "possible" here a bit confusing. A better word would probably be "probable" - by definition, all of the hypotheses are "possible" - but some are more likely, which is what we mean in this context. Curious if you and @vkakerbeck also feel that updating the wording would be clearer for newcomers.

In my mind, possible hypotheses made most sense up to now. Basically we have a set of hypotheses, a bunch of them may have negative or very low evidence so we don't update or consider them anymore, but they are still there and could be returned to. The possible hypotheses are only the ones that are above the x_percent_threshold. I wouldn't consider a hypothesis possible if it has negative evidence or evidence that is significantly below the top hypothesis... This changes with the new mechanisms that Ramy is implementing, where hypotheses are just being deleted. So in that case we could consider changing the terminology.

nielsleadholm:
@vkakerbeck yeah makes sense. Maybe worth waiting then until burst resampling has replaced the default evidence-update approach in all of our LMs.

vkakerbeck:
Are we planning to replace it in all LMs (I don't think this would be possible for displacement and feature LMs) or just write a new hypothesis updater that can be swapped out in the evidence LM? I would tend to just rename it in that one

nielsleadholm:
Yeah fair point, we could leave it alone int he base ones (although I'm still in favor of eventually condensing + simplifying the number of LM varieties to streamline the code-base).

ramyamounir:
Thanks for making these changes @jeremyshoemaker, looks great. I will work on the next PR.

PR #671, opened 2026-01-01
docs: future work automate benchmarks

codeallthethingz:
This PR adds details to the future work item "Automate Benchmark Experiments and Analysis".

Unusually, this future work item comprises of two connected RFCs.
- Converting from Bash to Python
- Automating the Benchmarks.  

Open questions

- [ ] Because this work was originally started in March of 2025, is it an out of date assumption that we even want to reimplement in Python?
- [ ] In the future work metadata I chose to link to the second RFC which is the final outcome - should we link from the second RFC to the first in the tbp.cli repo? 
- [ ] Should I split out another future work item with just the first RFC?  I think not but I'd like to hear thoughts.

### DIFF
--- a/docs/future-work/framework-improvements/automate-benchmark-experiments-and-analysis.md
+++ b/docs/future-work/framework-improvements/automate-benchmark-experiments-and-analysis.md
@@ -28,7 +28,7 @@ The tool would be better suited to Python. Transitioning from a Bash implementat
 
 Running the Monty benchmarks is time-consuming for researchers. Automating this process with the cli tool would be beneficial. However, implementing this feature in Bash proved challenging due to the numerous moving parts, and would be better suited to a full-fledged programming language.
 
-This RFC proposes rewriting the tool in Python to make maintainance and future enhancements easier. Using Python also ensures that both the research and engineering teams can more easily contribute to and maintain the tool moving forward.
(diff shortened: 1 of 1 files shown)


docs/future-work/framework-improvements/automate-benchmark-experiments-and-analysis.md
```diff
@@ -0,0 +1,67 @@
...
+## Summary
+
+This RFC proposes a new feature that allows the research team to run benchmark experiments on our AWS infrastructure, and have the results from those benchmarks update the CSV files in your local tbp.monty repository in `/benchmarks/results`.
+
+## Motivation
+
+Currently, updating the benchmarks is a manual process that takes up valuable researcher time. Automating this process would be beneficial for the following reasons
+- Manual updating is error prone.
+- Manual updating is time consuming.
+- Automation nudges towards more frequent updates without additional overhead.
```
nielsleadholm, line 67: As part of the "analysis" part of this, it could be worth integrating with the script that Ramy developed ([see discussion here](https://github.com/thousandbrainsproject/tbp.monty/pull/666#issuecomment-3706134128)). In particular, this creates a nice visual overview of results before and after a change, rather than trying to rely on git's diff view, which is a pain to visually parse.

PR #664, opened 2025-12-29
refactor!: finalize Simulator protocol cleanup

AgentRev:
As part of issue https://github.com/thousandbrainsproject/tbp.monty/issues/506 for the Simulator protocol refactor, this PR integrates all remaining changes from [prototype.simulator](https://github.com/thousandbrainsproject/prototype.simulator)'s main branch:

 - rename `Simulator.apply_actions` to `Simulator.step`;
 - `Simulator.reset()` returns `tuple[Observations, ProprioceptiveState]`;
 - `Simulator.step()` returns `tuple[Observations, ProprioceptiveState]`;
 - remove `Simulator.initialize_agent()`;
 - remove `Simulator.num_objects()`;

in addition to

 - remove `Simulator.states()` and `Environment.get_state()`.

PR #672, opened 2026-01-02
refactor!: split `EmbodiedEnvironment` into protocols

AgentRev:
As part of issue #506 for the Simulator protocol refactor, following a [suggestion](https://github.com/thousandbrainsproject/tbp.monty/pull/664#issuecomment-3702748809) by @tristanls-tbp, this PR splits ~the `Simulator` in two; `Simulator` to interact with objects, and `ExperimentSimulator` to setup/reset objects~ the `EmbodiedEnvironment` abstract class into `Environment`, `ObjectEnvironment`, and `ResettableEnvironment` protocols, along with some minor bugfixes and comments.

### DIFF
--- a/docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
+++ b/docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
@@ -15,7 +15,7 @@ Monty aims to implement a **general-purpose algorithm for understanding and inte
 In the current implementation, movement should happen in 3D (or less) space and be tracked using Euclidean location coordinates. Although we are convinced that the basic principles of Monty will also apply to abstract spaces ([potentially embedded in 3D space](https://thousandbrains.discourse.group/t/abstract-concept-in-monty/533/4)) and we know that the [brain uses different mechanisms to encode space](https://youtu.be/zRRazfFstvY), the current implementation relies on 3D Euclidean space.
 
 # Customizing Monty for Your Application
-The diagram below shows the base abstract classes in Monty. For general information on how to customize those classes, see our guide on [Customizing Monty](../customizing-monty.md). The Experiment class coordinates the experiment (learning and evaluation). It initializes and controls Monty and the environment and coordinates the interaction between them. The `SimulatedObjectEnvironment` protocol is wrapped in an `EnvironmentInterface` subclass, which exposes methods to interact with the environment. An experiment can have two environment interfaces associated with it: one for training and one for evaluation.
+The diagram below shows the base abstract classes in Monty. For general information on how to customize those classes, see our guide on [Customizing Monty](../customizing-monty.md). The Experiment class coordinates the experiment (learning and evaluation). It initializes and controls Monty and the environment and coordinates the interaction between them. The environment that implements the `Environment` protocol is wrapped in an `EnvironmentInterface` subclass, which exposes methods to interact with the environment. An experiment can have two environment interfaces associated with it: one for training and one for evaluation.
 
 ![Class structure in tbp.monty. Each class can be customized independently, allowing for easy modification and testing of individual components.](../../figures/how-to-use-monty/monty_class_diagram.png#width=500px)
 
@@ -25,13 +25,13 @@ Information flow in Monty implements a sensorimotor loop. Observations from the 
 
 Additionally, the `EnvironmentInterface` and Environment can implement specific functions to be executed at different points in the experiment, such as resetting the agent position and showing a new object or scene at the beginning of a new episode.
 
-To use Monty in a custom environment, you usually need to customize the `EnvironmentInterface` and implement the `SimulatedObjectEnvironment` protocol in custom classes. For example, if you look back at the previous tutorials, you will see that for those Habitat experiments, we've been using the `EnvironmentInterfacePerObject` and the `HabitatEnvironment`. The diagram below shows some key elements that need to be defined for these two classes. It's best to start thinking about the environment setup first, as this will force you to think through how to structure your application correctly for Monty to tackle.
+To use Monty in a custom environment, you usually need to customize the `EnvironmentInterface` class and write a custom implementation of the `SimulatedObjectEnvironment` protocol. For example, if you look back at the previous tutorials, you will see that for those Habitat experiments, we've been using the `EnvironmentInterfacePerObject` and the `HabitatEnvironment`. The diagram below shows some key elements that need to be defined for these two classes. It's best to start thinking about the environment setup first, as this will force you to think through how to structure your application correctly for Monty to tackle.
 ![Key elements to define for a custom environment interface](../../figures/how-to-use-monty/defining_env_and_env_interface.png)
 
 ### Environment
 The first thing to figure out is how movement should be defined in your environment. What actions are possible, and how do these actions change the agent's state and observations?
 
-If you are working with an existing environment, such as one used for reinforcement learning (for example, the Habitat environment we are using), you might just need to wrap this into the `.step()` function of your custom `Environment` class such that when `env.step(actions)` is called, an observation is returned. If you work with an application that isn't already set up like that, defining how actions lead to the next observation may be more involved. You can look at the `OmniglotEnvironmentInterface` or `SaccadeOnImageEnvironmentInterface` as examples (more details below).
+If you are working with an existing environment, such as one used for reinforcement learning (for example, the Habitat environment we are using), you might just need to wrap this into the `.step()` function of your custom `Environment` class such that when `env.step(actions)` is called, observations and proprioceptive state are returned. If you work with an application that isn't already set up like that, defining how actions lead to the next observation may be more involved. You can look at the `OmniglotEnvironmentInterface` or `SaccadeOnImageEnvironmentInterface` as examples (more details below).
 
 The observations should be returned as `Observations` with one entry per agent in the environment. Each agent should have `SensorObservations` for each of its sensors. For example, if there is one agent with two sensors that each sense two types of modalities, it would look like this:
 
@@ -58,9 +58,9 @@ obs = Observations(
 )
 ```
 
-Related to defining how actions change observations, you will also need to define how actions change the state of the agent. This is what the `get_state()` method returns. The returned state needs to be a `ProprioceptiveState(Dict[str, AgentState])` with an entry per agent in the environment. The entry should contain the agent's position and orientation relative to some global reference point (`AgentState`). For each sensor associated with that agent, a `SensorState` should contain the sensor's position and orientation relative to the agent.
+Related to defining how actions change observations, you will also need to define how actions change the state of the agent. The returned proprioceptive state needs to be a `ProprioceptiveState(Dict[str, AgentState])` with an entry per agent in the environment. The entry should contain the agent's position and orientation relative to some global reference point (`AgentState`). For each sensor associated with that agent, a `SensorState` should contain the sensor's position and orientation relative to the agent.
 
-For example, if you have one agent with two sensors, the state dictionary could look like this:
+For example, if you have one agent with two sensors, the proprioceptive state could look like this:
 ```python
 state = ProprioceptiveState(
   {
@@ -89,7 +89,7 @@ state = ProprioceptiveState(
 )
 ```
 
-Lastly, you need to define what happens when the environment is initialized (`__init__()`), when it is reset (`reset()`, usually at the end of an episode), and when it is closed (`close()`, at the end of an experiment). Resetting could include loading a new scene, resetting the agent position, or changing the arrangement of objects in the environment. It might also reset some of the environment's internal variables, such as step counters. Note that, as customary for RL environments, the `reset()` method is also expected to return `Observations`.
+Lastly, you need to define what happens when the environment is initialized (`__init__()`), when it is reset (`reset()`, usually at the end of an episode), and when it is closed (`close()`, at the end of an experiment). Resetting could include loading a new scene, resetting the agent position, or changing the arrangement of objects in the environment. It might also reset some of the environment's internal variables, such as step counters. Note that, as customary for RL environments, the `reset()` method is also expected to return `Observations` and `ProprioceptiveState`.
 
 ### Environment Interface
 The `EnvironmentInterface` manages retrieving observations from the `Environment`-given actions. The `Environment`, in turn, applies basic transforms to the raw observations from the environment.
@@ -265,7 +265,7 @@ For inference, we use the RGBD images taken with the iPad camera. Movement is de
 This can be implemented using two custom classes the [SaccadeOnImageEnvironment](https://github.com/thousandbrainsproject/tbp.monty/blob/4bc857580ae6ac015586af1a61b3e292a7827b6f/src/tbp/monty/frameworks/environments/two_d_data.py#L258) and [SaccadeOnImageEnvironmentInterface](https://github.com/thousandbrainsproject/tbp.monty/blob/db6b0404e3c3fa9c95688db5db33fa58053fcb8d/src/tbp/monty/frameworks/environments/embodied_data.py#L898):
 1. `SaccadeOnImageEnvironment`:
    - Defines initialization of all basic variables in the `__init__(patch_size, data_path)` method.
-   - Defines the `step(actions)` method, which uses the sensor's current location, the given actions, and their amounts to determine the new location on the image and extract a patch. It updates `self.current_loc` and returns the sensor patch observations as a dictionary.
+   - Defines the `step(actions)` method, which uses the sensor's current location, the given actions, and their amounts to determine the new location on the image and extract a patch. It updates `self.current_loc` and returns the sensor patch observations and proprioceptive state.
    - Defines `get_state()`, which returns the current state as a dictionary. The dictionary mostly contains `self.current_loc` and placeholders for the orientation, as the sensor and agent orientation never change.
    - Helper methods such as
      - `switch_to_object(scene_id, scene_version_id)` to load a new image
@@ -339,7 +339,7 @@ If your application uses sensors different from our commonly used cameras and de
 
 If your application requires a specific policy to move through the environment or you have a complex actuator to control, you might want to implement a custom `MotorSystem` or `MotorPolicy` class. For more details on our existing motor system and policies, see our [documentation on Monty's policies](https://thousandbrainsproject.readme.io/docs/policy).
 
-Writing those custom classes works the same way as it does for the `EnvironmentInterface` and `Environment` protocols. For general information, see our documentation on [customizing Monty](https://thousandbrainsproject.readme.io/docs/customizing-monty).
+Writing those custom classes works the same way as it does for the `EnvironmentInterface` class. For general information, see our documentation on [customizing Monty](https://thousandbrainsproject.readme.io/docs/customizing-monty).
 
 # Conclusion
-This tutorial was a bit more text than practical code. This is because every application is different, and we try to convey the general principles here. The first step for any application is to think about if and how the task can be phrased as a sensorimotor environment. What is Monty's action space? How is movement defined? How does it change observations? How do movement and sensation determine the sensor's location and orientation in space? This will then help you figure out how to define a custom `Environment` and `EnvironmentInterface`, and their associated `__init__`, `step`, `get_state`,`reset`, `pre_episode`, and `post_episode` methods. If you run into issues customizing Monty to your application, please come over to our [Discourse Forum](https://thousandbrains.discourse.group/) and ask for help!
+This tutorial was a bit more text than practical code. This is because every application is different, and we try to convey the general principles here. The first step for any application is to think about if and how the task can be phrased as a sensorimotor environment. What is Monty's action space? How is movement defined? How does it change observations? How do movement and sensation determine the sensor's location and orientation in space? This will then help you figure out how to define a custom `Environment` and `EnvironmentInterface`, and their associated `__init__`, `step`,`reset`, `pre_episode`, and `post_episode` methods. If you run into issues customizing Monty to your application, please come over to our [Discourse Forum](https://thousandbrains.discourse.group/) and ask for help!


AgentRev:
@tristanls-tbp What do you suggest we do with `self._env`'s type below? It would now require functions from both protocols. https://github.com/thousandbrainsproject/tbp.monty/blob/63d740496ad50a40556c8b279d4887be8ec983f3/src/tbp/monty/simulators/habitat/environment.py#L118

src/tbp/monty/simulators/simulator.py
```diff
@@ -30,6 +30,34 @@ class Simulator(Protocol):
...
+            If the actions are an empty sequence, the current observations are returned.
+        """
+        ...
+
+    def close(self) -> None:
+        """Close any resources used by the simulator."""
+        ...
+
+
+class ExperimentSimulator(Protocol):
```
tristanls-tbp: **issue**: This should go to ~~`tbp.monty.experiments`~~ (edit: `tbp.monty.frameworks.experiments`) 🙂, specifically ~~`tbp.monty.experiments.simulator.py`~~ (edit: `tbp.monty.frameworks.experiments.simulator.py`). At some point in the future, we're going to extract ~~`tbp.monty.experiments`~~ (edit: `tbp.monty.frameworks.experiments`) into its own thing.

Image

**note**: I imagine that eventually, `tbp.monty.simulators` will end up in `tbp.monty.environments.simulators` since we want Platform, Experiment, and Environment to be decoupled and simulators will end up in the environment portion of things. We'll likely _ship_ a simulator with `tbp.monty`, but it's technically part of an environment in the `Platform <-> Environment` interaction (sometimes controlled and intermediated by `Experiment`).

tristanls-tbp:
> @tristanls-tbp What do you suggest we do with `self._env`'s type below? It would now require functions from both protocols. 

@AgentRev , its type is `HabitatSim`. I think that's reasonable considering we are in a `HabitatEnvironment` wrapper. What do you think?

tristanls-tbp:
@AgentRev , one more bit on `HabitatEnvironment` vs `HabitatSim`. It's come up between @jeremyshoemaker and I that we're not sure that they are distinguishable. I think the distinction at this point might be historical and no longer material.

AgentRev:
All makes sense.

tristanls-tbp:
Sorry @AgentRev, sometimes I forget what is the "future" and what is now. My comment should have said `tbp.monty.frameworks.experiments`. We'll get rid of the extra `.frameworks.` in the future, but let's keep it for now to reduce dissonance.

AgentRev:
No worries.

tristanls-tbp:
@AgentRev, thank you for putting this together.

Upon seeing the result of my suggestion, I'm now questioning my suggestion.

In essence, should we be splitting `Simulator`, or should we be splitting `EmbodiedEnvironment`? Is it a `Simulator` protocol, or is it an `Environment` protocol?

When seeing the split, the following mapping comes to mind:
`tbp.monty.simulators.simulator.Simulator(Protocol)` -> `tbp.monty.frameworks.environments.environment.Environment(Protocol)`

`tbp.monty.frameworks.experiments.simulator.ExperimentSimulator(Protocol)` -> `tbp.monty.frameworks.experiments.environment.ExperimentEnvironment(Environment)`

This would result in
```python
class Environment(Protocol):
  def step(...) -> ... : ...
  def close(...) -> ... : ...
```
and
```python
class ExperimentEnvironment(Environment):
  def remove_all_objects(...) -> ... : ...
  def add_object(...) -> ... : ...
  def reset(...) -> ... : ...
```

This comes from the following mental model (drawn up when I was looking at actions and the Monty main loop, so it lacks the return values of observations and proprioceptive state):
Screenshot 2026-01-02 at 13 31 31

With some additional detail:
Screenshot 2026-01-02 at 13 32 14

We do need some local code accessible to Monty to interact with the environment. However, the code that does the actual interaction could be remote. 

Screenshot 2026-01-02 at 16 10 18


This is why it feels to me like the `step(actions) -> obs, state` is really an environment protocol, and now I'm thinking that the split should happen between `Environment(Protocol)` and `ExperimentEnvironment(Environment)` and that maybe the `Simulator(Protocol)` is fine as it is (I'm having a hard time imagining a non-experiment simulator).

**issue**: I think the `ExperimentSimulator(Simulator)` (or `ExperimentEnvironment(Environment)`) should _inherit_ from the other protocol. One of the _rare_ cases where I think the "is a" relationship holds.

AgentRev:
Yeah I think that's logical, I found it a little odd that `ExperimentSimulator` wouldn't inherit from `Simulator`, and like you said, a non-experiment simulator makes little sense. Personally, I'd go for splitting `EmbodiedEnvironment`.

Right now, `EmbodiedEnvironment` is an abstract base class and not a protocol, so how would you prefer integrating `ExperimentEnvironment`? Would `EmbodiedEnvironment` implement the `ExperimentEnvironment` protocol directly, or would it become the protocol itself rather than being an ABC, or something else?

tristanls-tbp:
@AgentRev so... take a look at https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/environments/two_d_data.py ...

That file suggests that there are _three_ protocols 🫤:
```python
class Environment(Protocol):
  def step(...) -> ... : ...
  def close(...) -> ... : ...
```
```python
# note the lack of inheritance
class ObjectEnvironment(Protocol):
  def remove_all_objects(...) -> ... : ...
  def add_object(...) -> ... : ...
```
```python
# note the lack of inheritance
class ResettableEnvironment(Protocol):
  def reset(...) -> ... : ...
```

With the resulting future implementation being:
```python
class OmniglotEnvironment(Environment, ResettableEnvironment):
    ...

class SaccadeOnImageEnvironment(Environment, ResettableEnvironment):
    ...
```
and Habitat future implementation elsewhere:
```python
class HabitatEnvironment(Environment, ObjectEnvironment, ResettableEnvironment):
    ...
```

Feels very fine-grained, but we literally have the use cases 🤔.

FYI @jeremyshoemaker 

As to your question
> Would `EmbodiedEnvironment` implement the `ExperimentEnvironment `protocol directly, or would it become the protocol itself rather than being an ABC, or something else?

It would become the protocol, the ABC would go away in favor of protocols.

> _Why?_ We want to move to [Protocols](https://typing.python.org/en/latest/spec/protocol.html#protocols) eventually. (...) 
>
> _Why do we want to move to Protocols eventually?_ We want to catch errors as early as possible, and using Protocols allows us to do this at type check time. Using abstract classes delays this until class instantiation, once the program runs.
>
> -- https://thousandbrainsproject.readme.io/docs/code-style-guide#abstract-classes-may-be-used-to-specify-interfaces-without-implementations

AgentRev:
Um, for simplicity purposes, would it make sense to define this alias along the aforementioned changes? Or would that be confusing?

    class EmbodiedEnvironment(Environment, ObjectEnvironment, ResettableEnvironment):
        pass

Just asking, cause the name `EmbodiedEnvironment` is kind of scattered everywhere across the codebase. With this, we could keep existing `EmbodiedEnvironment` references wherever appropriate (like for `HabitatEnvironment`), or otherwise switch to one of the base three.

tristanls-tbp:
I'm only seeing 36 instances. (Not sure what "only" says about my previous refactoring exposure though, maybe I'm desensitized to reference counts by now 🫠). 

Screenshot 2026-01-07 at 15 49 18

I don't think `EmbodiedEnvironment` is a good summary name for the three, so I don't see a reason to keep it around for convenience reasons alone.

AgentRev:
All good, I just wanted to check before doing a strafing run on those files

tristanls-tbp:
🤔 I wonder if there are figures that have "EmbodiedEnvironment" depicted in them.

tristanls-tbp:
**note**: Looks like the following figures will need to be updated. Making a note here as you don't have the access to the original drawings to edit.

```
how-monty-works/monty_class_diagram.png
how-to-use-monty/defining_env_and_env_interface.png
how-to-use-monty/monty_class_diagram.png
```

AgentRev:
@tristanls-tbp I did some tentative changes, went with protocols `SteppableEnvironment` and `ObjectEnvironment` instead of 3. `Steppable` in the same vein as `Resettable`, because ~I felt that just `Environment` was too broad of a name~ the name `ObjectEnvironment` would imply inheritance from `Environment`, which would not be the case here.

I moved `ResettableEnvironment`'s lone function into `SteppableEnvironment` directly, because all environments implement `reset()` already, and because the concept of a "non-resettable environment" felt odd. (well, apart from the real world, haha)

I also created a `SteppableObjectEnvironment` alias, because `EnvironmentInterface` calls functions from both protocols, and it needs a discrete type hint for its `self.env`.

Let me know if that makes sense or not, I'll readjust anything as needed. I'll take care of the docs and docstrings after your feedback.

AgentRev:
Could also do something like:

```
class Environment(Protocol):
  def reset(...) -> ... : ...
  def close(...) -> ... : ...

class SteppableEnvironment(Environment, Protocol):
  def step(...) -> ... : ...

class ObjectEnvironment(Environment, Protocol):
  def remove_all_objects(...) -> ... : ...
  def add_object(...) -> ... : ...

class SteppableObjectEnvironment(SteppableEnvironment, ObjectEnvironment, Protocol):
    pass
```

(note to self: clean the `NotImplementedError` stuff in the environments)

tristanls-tbp:
So, the options so far as I understand them:

@AgentRev option:
```python
@runtime_checkable
class SteppableEnvironment(Protocol):
  def step(...) -> ... : ...
  def reset(...) -> ... : ...
  def close(...) -> None : ...

class ObjectEnvironment(Protocol):
  def add_object(...) -> ... : ...
  def remove_all_objects(...) -> None : ...

# intersection type
class SteppableObjectEnvironment(SteppableEnvironment, ObjectEnvironment, Protocol:
  pass
```

@tristanls-tbp option:
```python
class Environment(Protocol):
  def step(...) -> ... : ...
  def close(...) -> ... : ...

class ObjectEnvironment(Protocol):
  def remove_all_objects(...) -> ... : ...
  def add_object(...) -> ... : ...

class ResettableEnvironment(Protocol):
  def reset(...) -> ... : ...

# intersection type
class SimulatedObjectEnvironment(Environment, ObjectEnvironment, ResettableEnvironment, Protocol):
  pass

# intersection type
class SimulatedEnvironment(Environment, ResettableEnvironment, Protocol):
  pass
```

Regarding `@runtime_checkable`... if `init_env` is the only place where we `isinstance(self.env, SteppableEnvironment)`, then let's get rid of that check from `init_env`.

My reason for separating `ResettableEnvironment` from `Environment` is that `reset()` is an Experiment concern, not a Runtime concern. `reset()` is only invoked in constructors and in experiment hooks like `pre_episode()`. I think `ResettableEnvironment` will be part of the experiment context, like`ObjectEnvironment` 🤔. I think calling `reset()` in a constructor is probably extraneous and could be removed. With the new protocol, any initial observations and proprioception can be gathered by calling `.step([])` with an empty action sequence.

So, the rationale goes: 1) separate runtime from experiment, 2) separate experiment use cases.

According to that rationale. When we separate runtime from experiment, we end up with `Environment.step+close` and `ExperimentEnvironment.add_object+remove_all_objects+reset`. Then, when we separate the experiment use cases ([two_d_data.py](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/environments/two_d_data.py)), we end up with `ObjectEnvironment.add_object+remove_all_objects` and `ResettableEnvironment.reset`.

```
# for illustration/framing purposes only, not a refactoring suggestion
tbp.monty.environments.environment.Environment
tbp.monty.experiments.environment.ObjectEnvironment
tbp.monty.experiments.environment.ResettableEnvironment
tbp.monty.experiments.environment.SimulatedObjectEnvironment
tbp.monty.experiments.environment.SimulatedEnvironment
```
The point of the above is that any runtime machinery (in the future) should never call `.reset()`, as it's not a runtime thing to do. Only the experiment should be aware of a `ResettableEnvironment`. Any runtime machinery should only do `.step()` or (maybe?) `.close()`. Otherwise, it's a signal/smell that an experiment crept into runtime and should be extracted.

The above runtime vs experiment organization is also why I call it `Environment` instead of `SteppableEnvironment`. There isn't another one to distinguish against within the runtime context of `tbp.monty.environments`.

AgentRev:
Ok yeah, with those precisions it does make a lot of sense. I'll make the necessary changes.

AgentRev:
> I think calling `reset()` in a constructor is probably extraneous and could be removed. With the new protocol, any initial observations and proprioception can be gathered by calling `.step([])` with an empty action sequence.

When I change this line in `EnvironmentInterface.__init__()` to `... = self.step([])`, I get some `HabitatDataTest` failures

https://github.com/thousandbrainsproject/tbp.monty/blob/0979fa9ed2ed2a59fc23929ff85f92259bf10846/src/tbp/monty/frameworks/environments/embodied_data.py#L112

image<br>

`self.env` is `HabitatEnvironment` in this context, for which the initial `self.step([])` call seems to behave differently from `self.reset()`.

tristanls-tbp:
Oh, I meant sometime in the future 🙂. I gotta be careful about what I say to you, or you'll finish the whole roadmap 😆.

But, to address the specific finding...

You'll notice that `HabitatEnvironment` is mostly a wrapper around `HabitatSim`, where the difference between `.step([])` and `.reset()` is that reset is resetting the actual agents in the sim:

https://github.com/thousandbrainsproject/tbp.monty/blob/84e26538dd92f4ae4e610b3566258d29c850accd/src/tbp/monty/simulators/habitat/simulator.py#L626-L627

I assume you swapped `.reset()` for `.step([])` in `EnvironmentInterface` constructor per your link in the comment.

I _think_ the below snippet can probably be removed from the constructor entirely:
https://github.com/thousandbrainsproject/tbp.monty/blob/84e26538dd92f4ae4e610b3566258d29c850accd/src/tbp/monty/frameworks/environments/embodied_data.py#L108-L111

In that case, I _think_ [`__next__(self)`](https://github.com/thousandbrainsproject/tbp.monty/blob/84e26538dd92f4ae4e610b3566258d29c850accd/src/tbp/monty/frameworks/environments/embodied_data.py#L122) would need to be updated to:
```python
    def __next__(self):
        actions = self.motor_system() if self._counter > 0 else []
        self._observation, proprioceptive_state = self.step(actions)
        self.motor_system._state = (
            MotorSystemState(proprioceptive_state) if proprioceptive_state else None
        )
        self._counter += 1
        return self._observation
```
This way, the constructor is not calling `.step(...)` before `.reset()`, which may need to be called first?

AgentRev:
> I assume you swapped `.reset()` for `.step([])` in `EnvironmentInterface` constructor per your link in the comment.

Yup. I tried all permutations of my code and your suggested code for `EnvironmentInterface`, but I always run into this error if  `self.reset()` is never called:
```
>           self.assertTrue(np.all(camera_obs_dist[MODALITY] == EXPECTED_STATES[i]))
E           AssertionError: False is not true
```

So, I set that aside for now, and pushed the other changes. Let me know if I missed anything within the scope of this PR (besides docs and docstrings).

> I gotta be careful about what I say to you, or you'll finish the whole roadmap 😆

As long as it's pre-approved and peer-reviewed, ain't nothin' wrong that 🤠

src/tbp/monty/frameworks/environments/environment.py
```diff
@@ -101,19 +110,24 @@ def remove_all_objects(self) -> None:
...
         pass
 
-    @abc.abstractmethod
-    def close(self) -> None:
-        """Close the environmnt releasing all resources.
 
-        Any call to any other environment method may raise an exception
-        """
-        pass
+# Intersection type
```
tristanls-tbp: **issue**: I don't think we need these comments. I intended them to provide context for our discussion, but it should be clearer what these are here.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -69,7 +74,7 @@ class EnvironmentInterface:
     returned by `motor_system` is applied.
 
     Attributes:
-        env: An instance of a class that implements :class:`EmbodiedEnvironment`.
+        env: An instance of a class that implements :class:`Environment`.
```
tristanls-tbp: **issue**: The type in `__init__` requires a `SimulatedObjectEnvironment`.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -502,6 +507,7 @@ def __next__(self):
 
     def pre_episode(self):
         super().pre_episode()
+        # TODO: self.env._agents is not part of SimulatedObjectEnvironment
```
tristanls-tbp, line 510: **note**: _nods_

src/tbp/monty/frameworks/environments/two_d_data.py
tristanls-tbp, line 1: **note**: Great to see a change like this 🙂.

src/tbp/monty/frameworks/experiments/simulator.py
tristanls-tbp, line 1: **issue**: I think we decided we're not splitting the `Simulator` and decided to split `EmbodiedEnvironment` instead, yes?

src/tbp/monty/simulators/habitat/simulator.py
```diff
@@ -524,9 +526,6 @@ def step(
             TypeError: If the action type is invalid
             ValueError: If the action name is invalid
         """
-        if not actions:
-            return self.observations
-
```
tristanls-tbp, line 529: **praise**: Oh wow, nice catch.

tests/unit/embodied_data_test.py
```diff
@@ -62,16 +61,14 @@
...
 
 
-class FakeEnvironmentRel(EmbodiedEnvironment):
+class FakeEnvironmentRel(SimulatedObjectEnvironment):
     def __init__(self):
         self._current_state = 0
 
-    @override
-    def add_object(self, *args, **kwargs) -> ObjectID:
+    def add_object(self, *_, **__) -> ObjectID:
```
tristanls-tbp: **issue**: We decided to use `# noqa: ARG002`, if needed, instead of using underscores. Otherwise, we start running into typing and IDE issues. You'll see some underscores (probably many of them from me), but that's because we haven't gotten around to undoing them.

**question**: What's the reason to remove `@override`?

tests/unit/embodied_data_test.py
```diff
@@ -62,16 +61,14 @@
...
         return ObjectID(-1)
 
-    @override
-    def step(self, actions) -> tuple[Observations, ProprioceptiveState]:
+    def step(self, _) -> tuple[Observations, ProprioceptiveState]:
```
tristanls-tbp: **issue**: use `# noqa: ARG002` instead, if needed.

... and below.

tests/unit/embodied_data_test.py
```diff
@@ -330,7 +326,7 @@ def test_omniglot_data_loader(self):
 
         env = FakeOmniglotEnvironment()
         omniglot_data_loader_abs = OmniglotEnvironmentInterface(
-            env=env,
+            env=env,  # TODO: FakeOmniglotEnvironment is not an OmniglotEnvironment
```
tristanls-tbp, line 328: **note**: Good point.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -839,7 +845,7 @@ def __init__(
         self.rng = rng
         self.motor_system = motor_system
         self.transform = transform
-        self._observation, proprioceptive_state = self.reset()
+        self._observation, proprioceptive_state = self.step([])
```
tristanls-tbp: **issue**: Might as well leave `self.reset()` in place since we're using `OmniglotEnvironment` which we know has reset.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -940,7 +947,7 @@ def __init__(
...
```
tristanls-tbp: **issue**: Might as well leave as `.reset()` since we're using `SaccadeOnImageEnvironment` which has reset. We can defer this when we do the step vs reset refactor globally.

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -1037,7 +1045,7 @@ def __init__(
...
```
tristanls-tbp: **issue**: Ditto on reset vs step comment. Let's leave `.reset()` for now.

tristanls-tbp (commented):
Thank you. Looks good overall.

Few high-level things.

**issue** I believe we ended up not splitting `Simulator`, so `ExperimentSimulator` is extraneous.

**issue**: All the changes from `.reset()` to `.step([])` are done on concrete classes that have `.reset()` methods. Let's leave the resets in place and reserve that change for when we figure out how to not do reset in any constructor.

src/tbp/monty/frameworks/environments/embodied_data.py
AgentRev: Good catch, I changed the type hint a couple times but the docstring didn't follow 😅

tests/unit/embodied_data_test.py
AgentRev: I assume  `@override` was put here to mute the unused arg errors for `add_object` and `step`, because the other methods below don't have `@override` despite inheriting from the same parents, so the decorator wasn't needed anymore with the underscores.

I'll undo the underscores and put `# noqa: ARG002`, because using `@override` for this purpose is a bit confusing.

src/tbp/monty/frameworks/experiments/simulator.py
AgentRev: Indeed, I forgot to delete it!

AgentRev:
(tried renaming the branch, forgot GitHub doesn't like that)

docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
```diff
@@ -339,7 +339,7 @@ If your application uses sensors different from our commonly used cameras and de
 
 If your application requires a specific policy to move through the environment or you have a complex actuator to control, you might want to implement a custom `MotorSystem` or `MotorPolicy` class. For more details on our existing motor system and policies, see our [documentation on Monty's policies](https://thousandbrainsproject.readme.io/docs/policy).
 
-Writing those custom classes works the same way as it does for the `EnvironmentInterface` and `EmbodiedEnvironment` classes. For general information, see our documentation on [customizing Monty](https://thousandbrainsproject.readme.io/docs/customizing-monty).
+Writing those custom classes works the same way as it does for the `EnvironmentInterface` and `Environment` protocols. For general information, see our documentation on [customizing Monty](https://thousandbrainsproject.readme.io/docs/customizing-monty).
```
tristanls-tbp: **suggestion**: 
```suggestion
Writing those custom classes works the same way as it does for the `EnvironmentInterface` class. For general information, see our documentation on [customizing Monty](https://thousandbrainsproject.readme.io/docs/customizing-monty).
```

docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
```diff
@@ -339,7 +339,7 @@ If your application uses sensors different from our commonly used cameras and de
...
+This tutorial was a bit more text than practical code. This is because every application is different, and we try to convey the general principles here. The first step for any application is to think about if and how the task can be phrased as a sensorimotor environment. What is Monty's action space? How is movement defined? How does it change observations? How do movement and sensation determine the sensor's location and orientation in space? This will then help you figure out how to define a custom `Environment` and `EnvironmentInterface`, and their associated `__init__`, `step`, `get_state`,`reset`, `pre_episode`, and `post_episode` methods. If you run into issues customizing Monty to your application, please come over to our [Discourse Forum](https://thousandbrains.discourse.group/) and ask for help!
```
tristanls-tbp: **suggestion (non-blocking)**: `get_state` is no longer part of the framework, although it is an internal convenience method in some classes.
```suggestion
This tutorial was a bit more text than practical code. This is because every application is different, and we try to convey the general principles here. The first step for any application is to think about if and how the task can be phrased as a sensorimotor environment. What is Monty's action space? How is movement defined? How does it change observations? How do movement and sensation determine the sensor's location and orientation in space? This will then help you figure out how to define a custom `Environment` and `EnvironmentInterface`, and their associated `__init__`, `step`,`reset`, `pre_episode`, and `post_episode` methods. If you run into issues customizing Monty to your application, please come over to our [Discourse Forum](https://thousandbrains.discourse.group/) and ask for help!
```

docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
```diff
@@ -15,7 +15,7 @@ Monty aims to implement a **general-purpose algorithm for understanding and inte
...
+The diagram below shows the base abstract classes in Monty. For general information on how to customize those classes, see our guide on [Customizing Monty](../customizing-monty.md). The Experiment class coordinates the experiment (learning and evaluation). It initializes and controls Monty and the environment and coordinates the interaction between them. The `SimulatedObjectEnvironment` protocol is wrapped in an `EnvironmentInterface` subclass, which exposes methods to interact with the environment. An experiment can have two environment interfaces associated with it: one for training and one for evaluation.
```
tristanls-tbp: **suggestion**:
```suggestion
The diagram below shows the base abstract classes in Monty. For general information on how to customize those classes, see our guide on [Customizing Monty](../customizing-monty.md). The Experiment class coordinates the experiment (learning and evaluation). It initializes and controls Monty and the environment and coordinates the interaction between them. The environment that implements the `Environment` protocol is wrapped in an `EnvironmentInterface` subclass, which exposes methods to interact with the environment. An experiment can have two environment interfaces associated with it: one for training and one for evaluation.
```

docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
```diff
@@ -25,13 +25,13 @@ Information flow in Monty implements a sensorimotor loop. Observations from the
...
 
-To use Monty in a custom environment, you usually need to customize the `EnvironmentInterface` and `EmbodiedEnvironment` classes. For example, if you look back at the previous tutorials, you will see that for those Habitat experiments, we've been using the `EnvironmentInterfacePerObject` and the `HabitatEnvironment`. The diagram below shows some key elements that need to be defined for these two classes. It's best to start thinking about the environment setup first, as this will force you to think through how to structure your application correctly for Monty to tackle.
+To use Monty in a custom environment, you usually need to customize the `EnvironmentInterface` and implement the `SimulatedObjectEnvironment` protocol in custom classes. For example, if you look back at the previous tutorials, you will see that for those Habitat experiments, we've been using the `EnvironmentInterfacePerObject` and the `HabitatEnvironment`. The diagram below shows some key elements that need to be defined for these two classes. It's best to start thinking about the environment setup first, as this will force you to think through how to structure your application correctly for Monty to tackle.
```
tristanls-tbp: **suggestion**:

```suggestion
To use Monty in a custom environment, you usually need to customize the `EnvironmentInterface` class and write a custom implementation of the `SimulatedObjectEnvironment` protocol. For example, if you look back at the previous tutorials, you will see that for those Habitat experiments, we've been using the `EnvironmentInterfacePerObject` and the `HabitatEnvironment`. The diagram below shows some key elements that need to be defined for these two classes. It's best to start thinking about the environment setup first, as this will force you to think through how to structure your application correctly for Monty to tackle.
```

docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
tristanls-tbp: **suggestion**: (removes `get_state()` sentence)

```suggestion
Related to defining how actions change observations, you will also need to define how actions change the state of the agent. The returned proprioceptive state needs to be a `ProprioceptiveState(Dict[str, AgentState])` with an entry per agent in the environment. The entry should contain the agent's position and orientation relative to some global reference point (`AgentState`). For each sensor associated with that agent, a `SensorState` should contain the sensor's position and orientation relative to the agent.
```

docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
```diff
@@ -25,13 +25,13 @@ Information flow in Monty implements a sensorimotor loop. Observations from the
...
 
-If you are working with an existing environment, such as one used for reinforcement learning (for example, the Habitat environment we are using), you might just need to wrap this into the `.step()` function of your custom `EmbodiedEnvironment` class such that when `env.step(actions)` is called, an observation is returned. If you work with an application that isn't already set up like that, defining how actions lead to the next observation may be more involved. You can look at the `OmniglotEnvironmentInterface` or `SaccadeOnImageEnvironmentInterface` as examples (more details below).
+If you are working with an existing environment, such as one used for reinforcement learning (for example, the Habitat environment we are using), you might just need to wrap this into the `.step()` function of your custom `Environment` class such that when `env.step(actions)` is called, an observation is returned. If you work with an application that isn't already set up like that, defining how actions lead to the next observation may be more involved. You can look at the `OmniglotEnvironmentInterface` or `SaccadeOnImageEnvironmentInterface` as examples (more details below).
```
tristanls-tbp: **suggestion**:
```suggestion
If you are working with an existing environment, such as one used for reinforcement learning (for example, the Habitat environment we are using), you might just need to wrap this into the `.step()` function of your custom `Environment` class such that when `env.step(actions)` is called, observations and proprioceptive state are returned. If you work with an application that isn't already set up like that, defining how actions lead to the next observation may be more involved. You can look at the `OmniglotEnvironmentInterface` or `SaccadeOnImageEnvironmentInterface` as examples (more details below).
```

docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
tristanls-tbp: **suggestion**:
```suggestion
For example, if you have one agent with two sensors, the proprioceptive state could look like this:
```

docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
```diff
@@ -92,7 +92,7 @@ state = ProprioceptiveState(
 Lastly, you need to define what happens when the environment is initialized (`__init__()`), when it is reset (`reset()`, usually at the end of an episode), and when it is closed (`close()`, at the end of an experiment). Resetting could include loading a new scene, resetting the agent position, or changing the arrangement of objects in the environment. It might also reset some of the environment's internal variables, such as step counters. Note that, as customary for RL environments, the `reset()` method is also expected to return `Observations`.
```
tristanls-tbp: **suggestion**:
```suggestion
Lastly, you need to define what happens when the environment is initialized (`__init__()`), when it is reset (`reset()`, usually at the end of an episode), and when it is closed (`close()`, at the end of an experiment). Resetting could include loading a new scene, resetting the agent position, or changing the arrangement of objects in the environment. It might also reset some of the environment's internal variables, such as step counters. Note that, as customary for RL environments, the `reset()` method is also expected to return `Observations` and `ProprioceptiveState`.
```

docs/how-to-use-monty/tutorials/using-monty-in-a-custom-application.md
tristanls-tbp: **suggestion**:
```suggestion
   - Defines the `step(actions)` method, which uses the sensor's current location, the given actions, and their amounts to determine the new location on the image and extract a patch. It updates `self.current_loc` and returns the sensor patch observations and proprioceptive state.
```

src/tbp/monty/frameworks/environments/environment.py
```diff
@@ -44,8 +47,36 @@ class ObjectInfo:
...
+        Args:
+            actions: The actions to apply to the environment.
+
+        Returns:
+            The current observations and proprioceptive state.
+
+        Note:
+            If the actions are an empty sequence, the current observations are returned.
+        """
+        pass
```
tristanls-tbp: **issue**: I think this should be `...` in a protocol definitions. I believe `...` communicates that this method needs to be implemented. Whereas `pass` can be (incorrectly) interpreted as defining an empty implementation. This difference becomes more pronounced when protocols _do_ have default implementations.

src/tbp/monty/frameworks/environments/environment.py
```diff
@@ -101,19 +114,28 @@ def remove_all_objects(self) -> None:
...
-    @abc.abstractmethod
-    def close(self) -> None:
-        """Close the environmnt releasing all resources.
 
-        Any call to any other environment method may raise an exception
-        """
-        pass
+class SimulatedEnvironment(Environment, ResettableEnvironment, Protocol):
+    """Protocol for steppable and resettable simulated environments."""
+
```
tristanls-tbp: **issue**: Please remove the doctring. The `class` definition itself already expresses this.

src/tbp/monty/frameworks/environments/environment.py
```diff
@@ -101,19 +114,28 @@ def remove_all_objects(self) -> None:
...
+    pass
+
+
+class SimulatedObjectEnvironment(
+    Environment, ObjectEnvironment, ResettableEnvironment, Protocol
+):
+    """Protocol for steppable and resettable simulated object environments."""
+
```
tristanls-tbp: **issue**: Please remove the doctring. The `class` definition itself already expresses this.

tests/unit/embodied_data_test.py
tristanls-tbp: **suggestion (non-blocking)**: This can now be removed.

tristanls-tbp: **suggestion (non-blocking)**: This can now be removed.

tristanls-tbp (commented):
Thank you for going through the docs.

I added a few comments.

I'll get started on updating the figures.

tristanls-tbp:
**todo**: Please update the figures below as well. I attached the new versions.

**note**: I assume it's easier for you to do it than for both of us to commit to your branch, but I can also do this myself at the end once you're done with all the other changes.

```
how-monty-works/monty_class_diagram.png
how-to-use-monty/defining_env_and_env_interface.png
how-to-use-monty/monty_class_diagram.png
```
Screenshot 2026-01-20 at 18 16 20

monty_class_diagram

PR #675, opened 2026-01-03
test: run all tutorial experiments

tristanls-tbp:
This pull request adds a test that runs all tutorial experiments, just as we instruct people to do in the documentation. We don't look at the results. We only ensure that tutorials can run for a few steps of one episode. We run full training where eval follows full training.

**note**: Discovered that `monty_meets_world_2dimage_inference` fails due to `SaccadeOnImageEnvironment` not including proprioceptive state for the `view_finder`. Fix is included in this pull request.

The test does not run by default when `pytest` is invoked. It requires `pytest tests/tutorials` to execute.

New job `Monty/test-tutorials` runs the test. It completes under 15 minutes.

After merging, I intend to make `test-tutorials` required, ensuring that no PR is merged that crashes a tutorial.

### DIFF
--- a/src/tbp/monty/frameworks/run.py
+++ b/src/tbp/monty/frameworks/run.py
@@ -30,7 +30,7 @@ def print_config(config: DictConfig) -> None:
     print("-" * 100)
 
 
-def run_name_output_dir(config: DictConfig) -> Path:
+def output_dir_from_run_name(config: DictConfig) -> Path:
     """Configure the output directory unique to the run name.
 
     The output directory is created if it does not exist.
@@ -58,7 +58,7 @@ def main(cfg: DictConfig):
     print_config(cfg)
     register_resolvers()
 
-    cfg.experiment.config.logging.output_dir = str(run_name_output_dir(cfg))
+    cfg.experiment.config.logging.output_dir = str(output_dir_from_run_name(cfg))
 
     experiment = hydra.utils.instantiate(cfg.experiment)
     start_time = time.time()
--- a/tests/tutorials/tutorials_test.py
+++ b/tests/tutorials/tutorials_test.py
@@ -19,7 +19,7 @@ from unittest import TestCase
 
 import hydra
 
-from tbp.monty.frameworks.run import run_name_output_dir
+from tbp.monty.frameworks.run import output_dir_from_run_name
 
 
 class TutorialsTest(TestCase):
@@ -40,7 +40,7 @@ class TutorialsTest(TestCase):
                 overrides=["experiment=tutorial/surf_agent_2obj_train"],
             )
             config.experiment.config.logging.output_dir = str(
-                run_name_output_dir(config)
+                output_dir_from_run_name(config)
             )
             experiment = hydra.utils.instantiate(config.experiment)
             with experiment:
@@ -83,7 +83,7 @@ class TutorialsTest(TestCase):
                 overrides=["experiment=tutorial/dist_agent_5lm_2obj_train"],
             )
             config.experiment.config.logging.output_dir = str(
-                run_name_output_dir(config)
+                output_dir_from_run_name(config)
             )
             experiment = hydra.utils.instantiate(config.experiment)
             with experiment:
@@ -109,7 +109,7 @@ class TutorialsTest(TestCase):
                 config_name="experiment",
                 overrides=["experiment=tutorial/omniglot_training"],
             )
-            inference_output_dir = str(run_name_output_dir(config))
+            inference_output_dir = str(output_dir_from_run_name(config))
             config.experiment.config.logging.output_dir = inference_output_dir
             experiment = hydra.utils.instantiate(config.experiment)
             with experiment:


tristanls-tbp:
Converting back to Draft since none of the data is available to instantiate the tutorials in the test environment.

tristanls-tbp:
@jeremyshoemaker I think that downloading the data sets on every test run would not be a great solution. Do you have any ideas for how to make the requisite datasets available to the tests? Our Conda installation invalidates our entire GitHub cache, so the dataset wouldn't be cached for long. But not sure how else to get these tests running. I'm thinking of sucking it up and downloading the datasets each time until we get rid of Conda, which should give us some GitHub cache.

jeremyshoemaker:
> @jeremyshoemaker I think that downloading the data sets on every test run would not be a great solution. Do you have any ideas for how to make the requisite datasets available to the tests? Our Conda installation invalidates our entire GitHub cache, so the dataset wouldn't be cached for long. But not sure how else to get these tests running. I'm thinking of sucking it up and downloading the datasets each time until we get rid of Conda, which should give us some GitHub cache.

I haven't looked at our workflow to see how Conda is clearing the entire cache, but my first thought was a cache that used a different key that doesn't change for each PR. I'm not sure there's a good option here if that's not possible. Something else I thought of was a custom test running image that has the dataset loaded already, but that might be too much work for something like this.

We're already downloading Conda and our dependencies each time. How large are the datasets compared to that? It might just be best to download it each time.

tristanls-tbp:
> I haven't looked at our workflow to see how Conda is clearing the entire cache

[Conda is so large that four installations fill the cache, which starts to evict everything.](https://github.com/thousandbrainsproject/tbp.monty/actions/caches)

jeremyshoemaker:
> > I haven't looked at our workflow to see how Conda is clearing the entire cache
> 
> [Conda is so large that four installations fill the cache, which starts to evict everything.](https://github.com/thousandbrainsproject/tbp.monty/actions/caches)

Of course it does. 🤦‍♂️

tristanls-tbp:
@vkakerbeck, @jeremyshoemaker So, the problem I am running into here is that our tutorials build upon each other, so some of them require that we run the previous tutorial. Therefore, it doesn't seem there is a good way to test these, except by running the entire sequence of tutorials. That starts to smell more like benchmarks territory to me. That's probably a good thing one day, but I don't think today is that day 🙂.

vkakerbeck:
Why are the pretrained models and datasets required to instantiate the configs? If we are not actually running the experiment, couldn't we just create temp files at those paths or some other hack around that so it doesn't throw a file-not-found error?

tristanls-tbp:
> Why are the pretrained models and datasets required to instantiate the configs?

We already have tests for instantiating the configs (snapshot testing), which don't require any data. This test instantiates the experiment itself (I'll update the PR name), so some initializers expect the data to be present so that they can present initial observations to Monty.

Good call on fake data, I'll see if I can get that to work 👍.

tristanls-tbp:
Ok, things are running... now just need to vastly reduce the duration...

src/tbp/monty/frameworks/run.py
```diff
@@ -30,6 +30,25 @@ def print_config(config: DictConfig) -> None:
     print("-" * 100)
 
 
+def run_name_output_dir(config: DictConfig) -> Path:
```
tristanls-tbp: **note**: I'm not thrilled about this name, but nothing else comes to mind.

tristanls-tbp:
> Good call on fake data, I'll see if I can get that to work 👍. 

@vkakerbeck I ended up downloading the real data.

.github/workflows/monty.yml
```diff
@@ -373,6 +373,68 @@ jobs:
...
+  test_tutorials:
+    name: test-tutorials
+    runs-on:
+      group: tbp.monty
+      labels: tbp-linux-x64-ubuntu-nvidia-4core-t4gpu
+    timeout-minutes: 120
+    needs:
+      - install_monty
+      - should_run_monty
+      - test_monty
```
jeremyshoemaker, line 393: Depending on how fast these tutorial tests run, we might want to have them run in parallel with the regular test suite instead of having to wait for them to finish first.

src/tbp/monty/frameworks/run.py
jeremyshoemaker: Maybe `output_dir_from_run_name`? Longer, but less clunky? 🤷 
Or maybe `unique_logging_output_dir` since that's what it's used for below?

.github/workflows/monty.yml
```diff
@@ -373,6 +373,68 @@ jobs:
...
+        with:
+          working_directory: tbp.monty
+      - name: Download data for tutorials
+        if: ${{ needs.should_run_monty.outputs.should_run_monty == 'true' }}
+        working-directory: tbp.monty
+        run: |
+          export PATH="$HOME/miniconda/bin:$PATH"
+          source activate tbp.monty
+          set -e
+          python -m habitat_sim.utils.datasets_download --uids ycb --data-path ~/tbp/data/habitat
```
vkakerbeck, line 428: oh man I didn't know this was an option! That changes a lot of what I was thinking about automating benchmarks. I now think this would be quite doable but will have to think through it a bit more. Anyways, not related to this PR, but I am glad I saw this.

tristanls-tbp:
Thank y'all. I renamed `run_name_output_dir` to `output_dir_from_run_name` per @jeremyshoemaker's suggestion. Otherwise, merging when ready.

PR #676, opened 2026-01-04
refactor!: Use TypedDict for `SensorObservations`

mf-maymay:
This PR makes `SensorObservations` a `TypedDict` instead of a `Dict` subclass.
I believe that this matches the original intent behind the current implementation.

These changes were discussed in [this thread](https://github.com/thousandbrainsproject/tbp.monty/pull/563#discussion_r2649486078).

In that thread, I proposed using a `dataclass`, not a `TypedDict`.
I originally set out to use a dataclass here, but I found that doing so increased the scope of the changes without commensurate benefit.

This PR also removes the `Modality` type.
The set of "modalities" is the set of keys of `SensorObservations`, so a new type is not needed.
This changes the API, so I've marked this as a breaking change.

### DIFF
--- a/src/tbp/monty/frameworks/environment_utils/transforms.py
+++ b/src/tbp/monty/frameworks/environment_utils/transforms.py
@@ -132,8 +132,8 @@ class AddNoiseToRawDepthImage(Transform):
             NoDepthSensorPresent: if no depth sensor is present.
         """
         # loop over sensor modules
-        for sm in observations[self.agent_id].keys():
-            if "depth" in observations[self.agent_id][sm].keys():
+        for sm in observations[self.agent_id]:
+            if "depth" in observations[self.agent_id][sm]:
                 noise = rng.normal(
                     0,
                     self.sigma,
@@ -441,7 +441,7 @@ class DepthTo3DLocations(Transform):
             # We need a semantic map that masks off-object pixels. We can use the
             # ground-truth semantic map if it's available. Otherwise, we generate one
             # from the depth map and (temporarily) add it to the observation dict.
-            if "semantic" in agent_obs.keys():
+            if "semantic" in agent_obs:
                 semantic_patch = agent_obs["semantic"]
             else:
                 # The generated map uses depth observations to determine whether


src/tbp/monty/frameworks/environment_utils/transforms.py
```diff
@@ -135,13 +133,13 @@ def call(
         """
         # loop over sensor modules
         for sm in observations[self.agent_id].keys():
-            if Modality("depth") in observations[self.agent_id][sm].keys():
+            if "depth" in observations[self.agent_id][sm].keys():
```
tristanls-tbp: **issue (non-blocking)**: Extraneous `.keys()`.

src/tbp/monty/frameworks/environment_utils/transforms.py
```diff
@@ -438,13 +436,13 @@ def call(
...
             agent_obs = observations[self.agent_id][sensor_id]
-            depth_patch = agent_obs[Modality("depth")]
+            depth_patch = agent_obs["depth"]
 
             # We need a semantic map that masks off-object pixels. We can use the
             # ground-truth semantic map if it's available. Otherwise, we generate one
             # from the depth map and (temporarily) add it to the observation dict.
-            if Modality("semantic") in agent_obs.keys():
-                semantic_patch = agent_obs[Modality("semantic")]
+            if "semantic" in agent_obs.keys():
```
tristanls-tbp: **issue (non-blocking)**: Extraneous `.keys()`.

src/tbp/monty/frameworks/models/abstract_monty_classes.py
```diff
@@ -20,19 +20,16 @@
...
+class SensorObservations(TypedDict, total=False):
     """Observations from a sensor."""
 
     rgba: npt.NDArray[np.int_]  # TODO: Verify specific type
     depth: npt.NDArray[np.float64]  # TODO: Verify specific type
     semantic: npt.NDArray[np.int_]  # TODO: Verify specific type
     semantic_3d: npt.NDArray[np.int_]  # TODO: Verify specific type
     sensor_frame_data: npt.NDArray[np.int_]  # TODO: Verify specific type
-    world_camera: npt.NDArray[np.int_]  # TODO: Verify specific type
+    world_camera: npt.NDArray[np.float64]  # TODO: Verify specific type
```
tristanls-tbp, line 32: **question**: Is the TODO still needed since you verified it?

tristanls-tbp (approved):
Thank you.

I assume the `@dataclass` complication was the maybe None issue, which sort of makes a mess of everything?

I don't like the loss of "modality" as a named concept, but I don't have a better suggestion.

Maybe @jeremyshoemaker has ideas? I asked him to review as well. 

In the meantime, this looks good to me.

src/tbp/monty/frameworks/models/abstract_monty_classes.py
goingmaymay: That's a valid question.

I verified it within the existing codebase (at least where there was a type error), but since this code is meant to be framework-agnostic, I think the types should be more carefully considered to make sure that they truly define the contract we want to enforce across all frameworks.

tristanls-tbp: OK. In that case, let's leave the TODO In place 🙂

goingmaymay:
Thanks, @tristanls-tbp.

Yes, the `None` handling was part of it, but the bigger issue was that `SensorObservations`-like dicts are used/expected in more places than you’d expect from the `Modality` references and mypy usage alone. The tests showed that the change would impact a large part of the codebase/API and likely cause friction for users.

I don’t have the WIP branch anymore, but here a few of my findings from working on it:
- Some call sites use `Mapping` instead of `SensorObservations` ([example](https://github.com/thousandbrainsproject/tbp.monty/blob/e00656010a3438cc4be77feec943b3fe78bf1fa7/src/tbp/monty/frameworks/models/motor_policies.py#L610)). I plan to clean this up as part of a broader effort to improve typing.
- There’s also a `HabitatObservations` TypedDict ([here](https://github.com/thousandbrainsproject/tbp.monty/blob/e00656010a3438cc4be77feec943b3fe78bf1fa7/src/tbp/monty/frameworks/models/sensor_modules.py#L84)) that looks related. I’d want to understand how these two are meant to relate before merging or separating them.
- Unless I’m mistaken, I saw that transforms were called during config instantiation, which I didn’t expect.

From here, my plan is to keep reducing mypy errors and tightening up the types so that changes like this are easier.

goingmaymay:
@tristanls-tbp, If we want to keep `Modality` as a named concept (and even a type), then my suggestion would be to use a `Literal` to define the set of allowed modalities and to add an assertion to ensure that the set of `Modality` options and `SensorObservation` keys match:

```python
from __future__ import annotations

from typing import Literal, TypedDict, get_args

import numpy as np
import numpy.typing as npt

Modality = Literal[
    "rgba",
    "depth",
    "semantic",
    "semantic_3d",
    "sensor_frame_data",
    "world_camera",
    "pixel_loc",
    "raw",
]


class SensorObservations(TypedDict, total=False):
    """Observations from a sensor."""

    rgba: npt.NDArray[np.int_]  # TODO: Verify specific type
    depth: npt.NDArray[np.float64]  # TODO: Verify specific type
    semantic: npt.NDArray[np.int_]  # TODO: Verify specific type
    semantic_3d: npt.NDArray[np.int_]  # TODO: Verify specific type
    sensor_frame_data: npt.NDArray[np.int_]  # TODO: Verify specific type
    world_camera: npt.NDArray[np.float64]  # TODO: Verify specific type
    pixel_loc: npt.NDArray[np.float64]  # TODO: Verify specific type
    raw: npt.NDArray[np.uint8]


assert set(get_args(Modality)) == set(SensorObservations.__annotations__), (
    "SensorObservations keys must match Modality options"
)
```

tristanls-tbp:
Thanks @mf-maymay. That looks somewhat brittle for low payoff. I don't feel it justifies my discomfort.

On some reflection, I think updating `SensorObservations` docstring to something like
```python
"""Observations from a sensor.

Each key corresponds to an explicitly supported modality."""
```
would be enough. But, I'm fine with the PR as is.

PR #678, opened 2026-01-06
test: remove seed-dependent brittle hierarchy tests

tristanls-tbp:
While working on #666, I discovered that some of the existing hierarchy tests are highly brittle and dependent on specific seeds.

This pull request removes the portion of `test_two_lm_heterarchy_experiment` that, in essence, hardcodes the seed-dependent execution history. It turned out that using different seeds (as surfaced by #666), there was no straightforward way to adapt tests to pass, as, depending on the seed, almost any combination of outcomes between the two LMs was possible during each episode. To reproduce this, just iterate `conf/test/hierarchy/two_lms_heterarchy.yaml config.seed` to `43`, `44`, random number, etc.

The assertion that models in LM1 should not store input from LM0 in episode 0 is removed. This can happen in episode 0 because of the uniqueness of what happens in episode 0. We immediately reach `no_match` because there are no objects, we set detected object to `new_object0` and then this can become input to LM1.


nielsleadholm:
Thanks Tristan, I'm copying in our 1:1 correspondence on this for additional context: @vkakerbeck I'm curious if you can think of a reason why the final point might be happening (LM0 sending an ID up the hierarchy on episode 1)?

Screenshot 2026-01-06 at 11 05 08 AM

Screenshot 2026-01-06 at 11 05 24 AM

Screenshot 2026-01-06 at 11 05 44 AM

tristanls-tbp:
@vkakerbeck and I discussed this PR a bit. @vkakerbeck would like to not remove without replacement. We discussed that the replacement could be a fake Environment, specifically, have `Environment.step()` return data. We then noticed that the data would have to be `tuple[Observations, ProprioceptiveState]`, but that's removed from the learning modules. A better option here might be a fake SensorModule that hardcodes what `State` it passes to the learning modules, so that what the learning modules observe is independent of the seed. This pattern of fake `State` is used in `evidence_sdr_lm_test.py` and `policy_test.py`. Furthermore, `unit_test_utils.py` has some utilities that return lists of `State` as well for similar purpose.

tests/unit/resources/unit_test_utils.py
```diff
@@ -385,84 +385,6 @@ def check_multilm_eval_results(self, eval_stats, num_lms, min_done, num_episodes
...
-            self.assertEqual(
-                eval_stats["primary_performance"][episode * 2],
-                "correct",
-                f"LM0 should detect the correct object in episode {episode}",
-            )
-            # NOTE: LM1 gets no match (due to incomplete models, especially of LM
-            # input channel). Will not test this here since maybe in the future this
-            # will be better and it is not a feature of the system.
-
-    def check_hierarchical_models(self, models):
```
tristanls-tbp, line 425: **note**: Moved to `hierarchy_test.py`

tests/unit/resources/unit_test_utils.py
```diff
@@ -385,84 +385,6 @@ def check_multilm_eval_results(self, eval_stats, num_lms, min_done, num_episodes
...
-                50,
-                "LM0 should have <= max_nodes_per_graph nodes in"
-                f" its graph for {model} but has {num_nodes}",
-            )
-        # Check LM1 does not store LM0 input in first epoch yet.
-        self.assertNotIn(
-            "learning_module_0",
-            models["0"]["LM_1"]["new_object0"].keys(),
-            "models in LM1 should not store input from LM0 in episode 0 yet.",
-        )
```
tristanls-tbp, line 456: **note**: While the entire method was moved to `hierarchy_test.py`, this assertion is not included in the move. It is removed since we learned that it does not generally hold (see PR description).

tests/unit/resources/unit_test_utils.py
```diff
@@ -385,84 +385,6 @@ def check_multilm_eval_results(self, eval_stats, num_lms, min_done, num_episodes
...
-                f"LM0 should detect the correct object in episode {episode}",
-            )
-            # NOTE: LM1 gets no match (due to incomplete models, especially of LM
-            # input channel). Will not test this here since maybe in the future this
-            # will be better and it is not a feature of the system.
-
-    def check_hierarchical_models(self, models):
-        for model in ["new_object0", "new_object1"]:
-            # Check that graph was extended when recognizing object.
-            self.assertLess(
```
tristanls-tbp, line 428: **note**: This becomes `self.assertLessEqual` in `hierarchy_test.py` to make it more generic.

tristanls-tbp:
I proposed that the removed `check_hierarchical_lm_train_results` does not test anything about successful training, but that those are inference assertions. Training assertions appear to be contained in `check_hierarchical_models`, which is retained. Since we have inference benchmarks, I proposed that `check_hierarchical_lm_train_results` be removed without replacement.

Awaiting feedback.

tristanls-tbp:
Superseded by #701.

PR #690, opened 2026-01-12
refactor!: Policies produce action sequences

scottcanoe:
This PR follows up on [PR 490](https://github.com/thousandbrainsproject/tbp.monty/pull/490) ("framework handles sequence of actions instead of single action") by having `MotorPolicy.dynamic_call` and `MotorPolicy.predefined_call` return `Sequence[Action]`.

This turned out to be more complicated than I expected, but I think I found a pretty minimal set of changes that'll do the trick, though it may not look like it. Having  `MotorPolicy.dynamic_call` return `Sequence[Action]` meant `BasePolicy.action` needed to become a `Sequence[Action]` too. This had a bunch of downstream effects.

I also ran a couple of benchmark experiments to verify nothing changed.

### DIFF
--- a/src/tbp/monty/frameworks/models/motor_policies.py
+++ b/src/tbp/monty/frameworks/models/motor_policies.py
@@ -128,17 +128,10 @@ class MotorPolicy(abc.ABC):
         if self.is_predefined:
             action: Sequence[Action] = self.predefined_call()
         else:
-            action: Sequence[Action] = self.dynamic_call(state)
-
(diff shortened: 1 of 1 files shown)


tristanls-tbp (changes requested):
@scottcanoe let's talk about this PR when you get back, it'll be easier during a live session.

There are conflicting ideas here that are being forced together: things that work for a single action, vs the desire to handle multiple actions.

PR #693, opened 2026-01-12
fix!: locations are now padded and filtered during memory update

ramyamounir:
I have been doing some debugging to figure out why running the burst sampling errors out with a [size mismatch](https://github.com/thousandbrainsproject/tbp.monty/blob/554b1dfcee3def3598a9f24bd5cebec9eed9c610/src/tbp/monty/frameworks/models/graph_matching.py?plain=1#L1496) in `_extract_entries_with_content` during memory update of stacked LMs experiments and I identified a couple of possible issues with the buffer.

## Issue 1
Throughout the buffer, when a channel doesn't send data at certain steps, the `_fill_old_values_with_nans` [function](https://github.com/thousandbrainsproject/tbp.monty/blob/e00656010a3438cc4be77feec943b3fe78bf1fa7/src/tbp/monty/frameworks/models/buffer.py?plain=1#L550-L572) is used to pad arrays with nan values to maintain alignment across timesteps ([location padding example](https://github.com/thousandbrainsproject/tbp.monty/blob/e00656010a3438cc4be77feec943b3fe78bf1fa7/src/tbp/monty/frameworks/models/buffer.py?plain=1#L543-L548), [feature padding example](https://github.com/thousandbrainsproject/tbp.monty/blob/e00656010a3438cc4be77feec943b3fe78bf1fa7/src/tbp/monty/frameworks/models/buffer.py?plain=1#L504-L508)). This is intentional because downstream in the `_extract_entries_with_content` [function](https://github.com/thousandbrainsproject/tbp.monty/blob/554b1dfcee3def3598a9f24bd5cebec9eed9c610/src/tbp/monty/frameworks/models/graph_matching.py?plain=1#L1484-L1500), missing entries are detected using `np.isnan` and then filtered out before building the graph. However, in `get_all_features_on_object`, when a channel's feature array is shorter than the buffer length in the last few steps, the code uses `ndarray.resize()` to [pad it](https://github.com/thousandbrainsproject/tbp.monty/blob/e00656010a3438cc4be77feec943b3fe78bf1fa7/src/tbp/monty/frameworks/models/buffer.py?plain=1#L361-L367). Unlike `_fill_old_values_with_nans`, `resize()` pads with zeros, not nans. This means these padded entries won't be identified as missing by the `np.isnan()` check, and invalid zero values could be incorporated into the graph instead of being filtered out. The fix is to replace the `resize()` call with explicit nan-padding, consistent with how gaps are handled elsewhere in the buffer. This was fixed in [0d0d9a55c13f](https://github.com/ramyamounir/tbp.monty/commit/0d0d9a55c13f693e8f2054f0a4ceb33523bccbad).

## Issue 2
During supervised learning of objects, Monty updates memory at the end of each episode via `GraphLM.post_episode()` -> `GraphLM._update_memory()` -> `GraphMemory.update_memory()`. The `_update_memory` method calls `buffer.get_infos_for_graph_update()` to [retrieve the locations and features](https://github.com/thousandbrainsproject/tbp.monty/blob/e00656010a3438cc4be77feec943b3fe78bf1fa7/src/tbp/monty/frameworks/models/buffer.py?plain=1#L432-L437) collected during the episode, which are then used to build or extend the object graph. When `get_all_locations_on_object()` is called without an input_channel argument (as done by `get_infos_for_graph_update`), it returns the `self.locations` dictionary [without any padding or filtering](https://github.com/thousandbrainsproject/tbp.monty/blob/e00656010a3438cc4be77feec943b3fe78bf1fa7/src/tbp/monty/frameworks/models/buffer.py?plain=1#L241-L242). In contrast, `get_all_features_on_object()` pads each channel's features to `len(self)` and then filters by `global_on_object_ids` (indices where at least one channel was on object). In stacked LM configurations, some channels (like LM inputs) may have shorter location arrays than others because they don't send data at every step. This creates a shape mismatch: features have length `len(global_on_object_ids)` while locations may be shorter. When `_extract_entries_with_content` tries to apply the boolean mask `locations[~missing_features]`, it sometimes fails with an `IndexError` because the mask length doesn't match the array length. The fix is to make `get_all_locations_on_object()` mirror the behavior of `get_all_features_on_object()` by padding each channel's locations to the full buffer length using nans, then filter by `global_on_object_ids`. This was fixed in [f9df6c7031d8](https://github.com/ramyamounir/tbp.monty/commit/f9df6c7031d818f0b71189815eb8d3060f1dccf4)

## Possible Simplification (implemented in [7067e121455a](https://github.com/ramyamounir/tbp.monty/commit/7067e121455a02abe509a147b14557b2e1d4956e))
We can reuse the function `_fill_old_values_with_nans` instead of reimplementing the padding logic but we would have to modify the function a bit because it automatically pads to `len(self) + 1`, but we want to pad to `len(self)`. We can refactor that as a `target_len` parameter passed to the function. I'm not sure which way is better.

### DIFF
--- a/src/tbp/monty/frameworks/models/buffer.py
+++ b/src/tbp/monty/frameworks/models/buffer.py
@@ -16,6 +16,7 @@ import time
 from typing import Any, Callable, ClassVar
 
 import numpy as np
+import numpy.typing as npt
 import quaternion as qt
 import torch
 from omegaconf import DictConfig, ListConfig, OmegaConf
@@ -475,7 +476,7 @@ class FeatureAtLocationBuffer:
             # If no symmetry was detected, this will be None.
             self.stats["symmetric_rotations_ts"] = self.stats["symmetric_rotations"]
 
-    def _global_on_object_ids(self) -> np.ndarray:
+    def _global_on_object_ids(self) -> npt.NDArray[np.bool_]:
         """Get indices of steps where at least one input was on the object.
 
         Returns:


tristanls-tbp:
**issue**: Missing a unit test that fails before this fix and passes after this fix.

ramyamounir:
Thanks @tristanls-tbp, I added buffer padding tests for both issues described in the PR description. They fail without the fix, can you take a look?

Note that I couldn't find any existing buffer tests. What I added is a good starting point for the parts with bugs, but it doesn't cover the whole buffer.

tests/unit/frameworks/models/feature_at_location_buffer_test.py
```diff
@@ -0,0 +1,178 @@
...
+        # Step 2: Only SM sends data
+        state_sm_2 = create_mock_state(
+            sender_id="SM_0",
+            sender_type="SM",
+            location=np.array([4.0, 5.0, 6.0]),
+            on_object=True,
+        )
+        self.buffer.append([state_sm_2])
+
+        # Get all features on object
```
tristanls-tbp: **nitpick**: Extraneous comment due to `.get_all_features_on_object()` sharing all the words 🙂.

tests/unit/frameworks/models/feature_at_location_buffer_test.py
```diff
@@ -0,0 +1,178 @@
...
+        features = self.buffer.get_all_features_on_object()
+
+        # LM_0 features should be padded with nans for step 2
```
tristanls-tbp: **issue**: This seems like an extraneous comment. The below two lines do not test this nor works with NaNs in any way.

tests/unit/frameworks/models/buffer_test.py
tristanls-tbp, line 1: **issue**: Please rename the file to `buffer_test.py`. This is so it corresponds directly to `src/tbp/monty/frameworks/models/buffer.py`. The test class `FeatureAtLocationBufferPaddingTest` is sufficient to partition the `buffer_test.py` internally from any other `buffer.py` internals we may want to test in the future.

tests/unit/frameworks/models/buffer_test.py
```diff
@@ -0,0 +1,178 @@
...
+        features = self.buffer.get_all_features_on_object()
+
+        # Both should have the same channels
+        self.assertEqual(set(locations.keys()), set(features.keys()))
+
+        # For each channel, locations should have same number of rows as features
+        for channel in locations.keys():
+            loc_rows = locations[channel].shape[0]
+            # Use any feature to compare (e.g., pose_vectors)
+            feat_rows = features[channel]["pose_vectors"].shape[0]
```
tristanls-tbp, line 195: **suggestion (non-blocking)**: You could loop through all of them to be sure.

tristanls-tbp (changes requested):
The fix looks good to me 👍, thank you for adding a unit test for it.

I have administrative comments, one rename and some code comment comments, but functionality and test-wise, the code looks good.

tristanls-tbp:
@ramyamounir I double-checked what test files exist. I see `tests/unit/frameworks/models/buffer_encoder.test`, which also tests a `buffer.py` thing. I think these should be combined into one `buffer_test.py` file for the same reason as I mentioned before, to maintain file naming correspondence with `src`.

tests/unit/frameworks/models/buffer_test.py
ramyamounir: This feels like it would be a different buffer test (outside the scope of padding tests) that makes sure `buffer.append()` modifies all the features in the same way. I can come back to adding some more tests to the buffer class later, but here I just wanted to test locations vs. features, assuming that features have the correct structure.

src/tbp/monty/frameworks/models/buffer.py
vkakerbeck, line 1: To avoid future discrepancies in what happens to locations and features, could we write one or two helper functions that are called in both cases instead of using essentially the same code? Like one for getting `global_on_object_ids` and one for everything in the `if len(channel_locs) < len(self):` branch?

tests/unit/frameworks/models/buffer_test.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+class FeatureAtLocationBufferPaddingTest(unittest.TestCase):
+    """Tests for FeatureAtLocationBuffer focusing on padding and filtering behavior."""
+
+    def setUp(self):
+        """Create a fresh buffer for each test."""
+        self.buffer = FeatureAtLocationBuffer()
+
+    def test_get_all_features_on_object_pads_with_nans_not_zeros(self):
```
vkakerbeck: could we remove the `not_zeros` part from this test name? It seems like the important part that is tested here is that it pads with nans, the not zeros part seems specific to the issue we saw before but isn't specifically tested

src/tbp/monty/frameworks/models/buffer.py
ramyamounir: Yes, I've simplified the code a bit.
* global_on_object_ids is refactored in [e983ed7152b0](https://github.com/ramyamounir/tbp.monty/commit/e983ed7152b07695ec42e6b639791fa61646ff5b)
* The padding function is also refactored and simplified in [7067e121455a](https://github.com/ramyamounir/tbp.monty/commit/7067e121455a02abe509a147b14557b2e1d4956e)
* Added a few more unit tests for the new padding function in [660cbbac3520](https://github.com/ramyamounir/tbp.monty/commit/660cbbac35204310b4f9c0c2b9ea0e4ad46ebdd6)

This addresses the simplification I proposed in the PR description.

tests/unit/frameworks/models/buffer_test.py
ramyamounir: Yes, good catch. Removed in [ff990d275109](https://github.com/ramyamounir/tbp.monty/commit/ff990d275109ac1e889020e6bf4d1def832fadde)

PR #694, opened 2026-01-13
chore: fix uv tests again (third time)

jeremyshoemaker:
Dependencies like `torch-scatter` don't correctly declare their build dependencies, and the error message suggests an experimental UV feature that doesn't seem to solve the problem.

Installing `torch` first and then installing `torch-scatter` with build isolation disabled (already configured in `pyproject.toml`), the environment gets set up correctly.

I'm not comfortable with this solution since it means specifying the `torch` version outside of the `pyproject.toml` file, but I don't have a better solution at the moment.

### DIFF
--- a/UV_PROTOTYPE.md
+++ b/UV_PROTOTYPE.md
@@ -10,5 +10,6 @@ Some notes on how to set up this environment.
 ```sh
 # The --seed is needed so we can build the torch packages
 uv venv -p 3.9.22 --seed
+uv pip install torch==1.13.1 # Use the version from pyproject.toml
 uv sync --extra dev --extra simulator_mujoco
 ```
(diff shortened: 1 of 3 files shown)


UV_PROTOTYPE.md
```diff
@@ -10,5 +10,6 @@ Some notes on how to set up this environment.
 ```sh
 # The --seed is needed so we can build the torch packages
 uv venv -p 3.9.22 --seed
+uv pip install torch==1.13.1 # Use the version from pyproject.toml
```
tristanls-tbp, line 13: **question**: Would it use `pyproject.toml` version if the version is omitted here?

jeremyshoemaker: No, it doesn't. I wish it did, but it ends up installed version 2.8.0 instead.

tristanls-tbp (approved):

PR #695, opened 2026-01-14
chore: fix the seed for EXPECTED_STATES in EmbodiedDataTest

ramyamounir:
This is an attempt to address the brittle unit test [here](https://github.com/thousandbrainsproject/tbp.monty/actions/runs/20973253798/job/60281989301). By fixing the EXPECTED_STATES to a fixed array, we eliminate the chance of state collision.

### DIFF
--- a/.github/workflows/monty.yml
+++ b/.github/workflows/monty.yml
@@ -359,7 +359,7 @@ jobs:
         uses: ./tbp.monty/.github/actions/install_monty
         with:
           working_directory: tbp.monty
-      - name: Run python tests
+      - name: Run configuration and unit tests
         if: ${{ needs.should_run_monty.outputs.should_run_monty == 'true' }}
         working-directory: tbp.monty
         run: |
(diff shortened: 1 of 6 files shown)


jeremyshoemaker:
Normally I'd ask for a description explaining why this change is needed, but it seems self-explanatory.

It continues to bother me that we use random number generators, but then fix the seed everywhere. I think ideally, we should run things with a random seed, and output the seed on execution so if we want to fix it for reproducing an issue and debugging we can, but we aren't potentially missing broken situations because we're always testing with the same seed every time. If the system is supposed to have stochastic elements, we should test it stochastically in general, and only fix the seed when we absolutely have to.

ramyamounir:
> It continues to bother me that we use random number generators, but then fix the seed everywhere.

I agree, but I'm not very familiar with these tests to do something about it :sweat_smile: I added a bit of context in the description.

tests/unit/embodied_data_test.py
```diff
@@ -57,7 +57,7 @@
     f"{AGENT_ID}.turn_right",
 ]
 POSSIBLE_ACTIONS_ABS = [f"{AGENT_ID}.set_yaw", f"{AGENT_ID}.set_sensor_pitch"]
-EXPECTED_STATES: npt.NDArray[np.uint8] = np.random.randint(
+EXPECTED_STATES: npt.NDArray[np.uint8] = np.random.RandomState(42).randint(
     0, 256, size=NUM_STEPS, dtype=np.uint8
 )
```
tristanls-tbp: Looks like this used to be
```python
EXPECTED_STATES = np.random.rand(NUM_STEPS)
```
We then elaborated it to
```python
EXPECTED_STATES: npt.NDArray[np.uint8] = np.random.randint(
    0, 256, size=NUM_STEPS, dtype=np.uint8
)
```
But, the overall intent seems to be that we want a `NUM_STEPS` long numpy array. By making it deterministic, we ensure that no steps are equal to each other.

```suggestion
EXPECTED_STATES: npt.NDArray[np.uint8] = np.arange(0, NUM_STEPS, dtype=np.uint8)
```

PR #696, opened 2026-01-14
docs: update WSL tutorial

AgentRev:
This PR updates the WSL tutorial with the following:
- Recommended a fresh Linux install;
- Added note about Git for Windows usage;
- Added warning to avoid manually installing Habitat-Sim;
- Removed optional `GALLIUM_DRIVER` variable due to [issues](https://thousandbrains.discourse.group/t/tutorial-monty-habitat-sim-on-windows-via-wsl-vs-code/749/8) with Intel GPUs;
- Added Omniglot download instructions;
- Added link to Ruff extension for VS Code;
- Explained what the `code .` command does.


docs/how-to-use-monty/getting-started/getting-started-on-windows-via-wsl.md
```diff
@@ -15,7 +15,15 @@ Follow the Microsoft guide to install Windows Subsystem for Linux: https://learn
 
 It is important that you complete every step of their guide, otherwise you may encounter issues during the next steps below.
 
-Regarding the Linux distribution, [Ubuntu 24.04 LTS](https://apps.microsoft.com/detail/9nz3klhxdjp5) is recommended. Once installed, the Linux filesystem can directly be accessed via the "Linux" section in the Windows File Explorer, which should normally point to this path: `\\wsl.localhost\Ubuntu-24.04\`
+A fresh Linux installation is recommended for Monty. If desired, you may attempt to re-use an existing Linux installation, but be aware you might run into GPU-related errors if default libraries and settings were significantly altered.
+
+Regarding the Linux distribution, [Ubuntu 24.04 LTS](https://apps.microsoft.com/detail/9nz3klhxdjp5) is recommended. Once installed, its filesystem can directly be accessed via the "Linux" section in the Windows File Explorer, which should normally point to `\\wsl.localhost\Ubuntu-24.04\` or `\\wsl$\Ubuntu-24.04\` 
+
+> [!NOTE]
+> If you plan on using Git for Windows on the WSL filesystem, enter this command on Windows in PowerShell or CMD or Git Bash:
```
jeremyshoemaker: I think we should mention what problem this is avoiding, maybe mention the error message someone might hit if they don't use this that they can look for when they come back to the documentation. It would help people resolve the problem quicker.

docs/how-to-use-monty/getting-started/getting-started-on-windows-via-wsl.md
```diff
@@ -177,6 +193,8 @@ Still from Ubuntu, enter this to launch VS Code with Monty:
 cd ~/tbp && code .
 ```
 
+The `code .` command launches both VS Code for Windows and the Linux VS Code Server directly into the active WSL folder.
```
jeremyshoemaker: Since it's inside a Linux environment, I might phrase this as "in the current directory" rather than "into the active WSL folder", but maybe this is the more common way to talk about it in the WSL world. It's confusing because in the old DOS days, they were directories, hence the command `dir`.

docs/how-to-use-monty/getting-started/getting-started-on-windows-via-wsl.md
```diff
@@ -78,18 +89,12 @@ This might take a few minutes or more to run, depending on your download speed.
...
+Then, install libopengl0 to allow running Habitat-Sim:
 
 ```shell
 sudo apt -y install libopengl0
 ```
 
-Then, configure Linux to directly use Windows GPU drivers:
-
-```shell
-echo "export GALLIUM_DRIVER=d3d12" >> ~/.bashrc && exec $SHELL
```
jeremyshoemaker, line 90: Are there any situations where someone might need this? Should we mention it in an aside?

docs/how-to-use-monty/getting-started/getting-started-on-windows-via-wsl.md
AgentRev: Indeed. I reclarified it.

docs/how-to-use-monty/getting-started/getting-started-on-windows-via-wsl.md
AgentRev: Yup that's I meant, I got mixed up with conda's "active environment". Fixed it.

docs/how-to-use-monty/getting-started/getting-started-on-windows-via-wsl.md
AgentRev: Nah, I initially added it because I thought it would help with Habitat's `Failed to get EGL display` error, but further testing revealed it has zero effect on that error, which is actually caused by a wrong Habitat version.

In practice, all `GALLIUM_DRIVER=d3d12` seems to do is allow native GPU rendering within WSL, such as with the `habitat-viewer` program for example, but it provides no benefit for Monty.

PR #700, opened 2026-01-15
feat!: Burst sampling in ResamplingHypothesesUpdater

ramyamounir:
This is the second PR in the implementation project of the burst sampling hypothesis updater. The PR mainly introduces two main features:
- Decoupling the resampling parameters: In the past, we kept the parameters coupled such that the hypothesis space size stays the same, i.e., we resample the same number of deleted hypotheses every step. Here we decouple this behavior such that the hypothesis space size can change over time. Additionally, a `HypothesesSelection` container class in introduced to refer to a subset of hypotheses between different parts of the code.
- Burst sampling: The decoupling of parameters allowed for redesigning the heuristics for deciding on which hypotheses are to be deleted, resampled, or maintained. We delete hypotheses if they don't accumulate enough evidence for a few steps (as measured by their smoothed evidence slope metric). The sampling happens in bursts with a set step duration. If the best hypothesis is not able to accumulate enough evidence, we trigger a burst of hypotheses sampling for all objects. This mechanism can cause some unlikely object hypothesis spaces to have zero hypotheses, some modifications to the code (specifically, LM, GSG, and the channel mapper) were applied to handle this scenario.

Additional, more detailed, context can be found in this [PR description](https://github.com/thousandbrainsproject/feat.dynamic_resizing/pull/4)

## Benchmarks

We will not be updating any benchmarks (other than unsupervised inference) because this PR doesn't set `ResamplingHypothesesUpdater`. That said, I still ran the benchmarks to make sure we are getting the same benefits we saw in the feature repo. Results of the benchmark runs are tagged with `PR#700`

| percent_correct (%) | 99.29 | 100 | 0.71 | ✅ |
| used_mlh (%) | 0.71 | 0 | -0.71 | ✅ |
| match_steps | 35 | 31 | -4 | ✅ |
| rotation_error (deg) | 10.69 | 8.67 | -2.02 | ✅ |
| runtime (min) | 2 | 2 | 0 | ⬜ |
| episode_runtime (sec) | 9 | 7 | -2 | ✅ |

| percent_correct (%) | 99.29 | 100 | 0.71 | ✅ |
| match_steps | 27 | 28 | 1 | ❌ |
| rotation_error (deg) | 11.96 | 3.65 | -8.31 | ✅ |
| runtime (min) | 2 | 2 | 0 | ⬜ |
| episode_runtime (sec) | 10 | 12 | 2 | ❌ |

| percent_correct (%) | 97 | 99 | 2 | ✅ |
| used_mlh (%) | 5 | 1 | -4 | ✅ |
| match_steps | 55 | 37 | -18 | ✅ |
| rotation_error (deg) | 25.14 | 12.83 | -12.31 | ✅ |
| runtime (min) | 5 | 3 | -2 | ✅ |
| episode_runtime (sec) | 31 | 22 | -9 | ✅ |

| percent_correct (%) | 98 | 100 | 2 | ✅ |
| used_mlh (%) | 3 | 1 | -2 | ✅ |
| match_steps | 38 | 35 | -3 | ✅ |
| rotation_error (deg) | 15.11 | 17.65 | 2.54 | ❌ |
| runtime (min) | 4 | 3 | -1 | ✅ |
| episode_runtime (sec) | 24 | 19 | -5 | ✅ |

| used_mlh (%) | 1 | 1 | 0 | ⬜ |
| match_steps | 28 | 29 | 1 | ❌ |
| rotation_error (deg) | 25.14 | 12.35 | -12.79 | ✅ |
| runtime (min) | 3 | 4 | 1 | ❌ |
| episode_runtime (sec) | 21 | 24 | 3 | ❌ |

| rotation_error (deg) | 13.46 | 9.27 | -4.19 | ✅ |
| runtime (min) | 2 | 2 | 0 | ⬜ |
| episode_runtime (sec) | 12 | 13 | 1 | ❌ |

| match_steps | 52 | 55 | 3 | ❌ |
| rotation_error (deg) | 49.54 | 44.85 | -4.69 | ✅ |
| runtime (min) | 5 | 5 | 0 | ⬜ |
| episode_runtime (sec) | 42 | 41 | -1 | ✅ |

### base_10simobj_surf_agent

| percent_correct (%) | 92.14 | 98.57 | 6.43 | ✅ |
| used_mlh (%) | 10.71 | 3.57 | -7.14 | ✅ |
| match_steps | 83 | 56 | -27 | ✅ |
| rotation_error (deg) | 11 | 3.96 | -7.04 | ✅ |
| runtime (min) | 6 | 5 | -1 | ✅ |
| episode_runtime (sec) | 31 | 24 | -7 | ✅ |

| percent_correct (%) | 78 | 85 | 7 | ✅ |
| used_mlh (%) | 36 | 24 | -12 | ✅ |
| match_steps | 201 | 176 | -25 | ✅ |
| rotation_error (deg) | 21.46 | 30.88 | 9.42 | ❌ |
| runtime (min) | 10 | 10 | 0 | ⬜ |
| episode_runtime (sec) | 85 | 83 | -2 | ✅ |

| percent_correct (%) | 90 | 97 | 7 | ✅ |
| used_mlh (%) | 32 | 32 | 0 | ⬜ |
| match_steps | 166 | 165 | -1 | ✅ |
| rotation_error (deg) | 27.15 | 19.94 | -7.21 | ✅ |
| runtime (min) | 15 | 19 | 4 | ❌ |
| episode_runtime (sec) | 127 | 149 | 22 | ❌ |

| percent_correct (%) | 65 | 70 | 5 | ✅ |
| used_mlh (%) | 83 | 75 | -8 | ✅ |
| match_steps | 13 | 13 | 0 | ⬜ |
| rotation_error (deg) | 101.6 | 83.67 | -17.93 | ✅ |
| runtime (min) | 4 | 4 | 0 | ⬜ |
| episode_runtime (sec) | 6 | 7 | 1 | ❌ |

| percent_correct (%) | 85 | 71.43 | -13.57 | ❌ |
| used_mlh (%) | 12.86 | 21.43 | 8.57 | ❌ |
| match_steps | 27 | 28 | 1 | ❌ |
| rotation_error (deg) | 25.3 | 12.13 | -13.17 | ✅ |
| runtime (min) | 3 | 4 | 1 | ❌ |
| episode_runtime (sec) | 1 | 2 | 1 | ❌ |

### base_77obj_dist_agent

| percent_correct (%) | 90.91 | 96.54 | 5.63 | ✅ |
| used_mlh (%) | 12.99 | 5.63 | -7.36 | ✅ |
| match_steps | 102 | 73 | -29 | ✅ |
| rotation_error (deg) | 14.89 | 10.24 | -4.65 | ✅ |
| runtime (min) | 21 | 15 | -6 | ✅ |
| episode_runtime (sec) | 61 | 41 | -20 | ✅ |

### base_77obj_surf_agent

| percent_correct (%) | 98.7 | 99.57 | 0.87 | ✅ |
| used_mlh (%) | 5.19 | 4.33 | -0.86 | ✅ |
| match_steps | 53 | 41 | -12 | ✅ |
| rotation_error (deg) | 6.7 | 1.56 | -5.14 | ✅ |
| runtime (min) | 15 | 12 | -3 | ✅ |
| episode_runtime (sec) | 37 | 29 | -8 | ✅ |

| percent_correct (%) | 92.21 | 93.51 | 1.3 | ✅ |
| used_mlh (%) | 26.84 | 17.75 | -9.09 | ✅ |
| match_steps | 174 | 140 | -34 | ✅ |
| rotation_error (deg) | 34.01 | 30.58 | -3.43 | ✅ |
| runtime (min) | 41 | 34 | -7 | ✅ |
| episode_runtime (sec) | 137 | 112 | -25 | ✅ |

| percent_correct (%) | 94.37 | 96.54 | 2.17 | ✅ |
| used_mlh (%) | 21.65 | 21.21 | -0.44 | ✅ |
| match_steps | 111 | 111 | 0 | ⬜ |
| rotation_error (deg) | 33.9 | 23.37 | -10.53 | ✅ |
| runtime (min) | 34 | 33 | -1 | ✅ |
| episode_runtime (sec) | 114 | 110 | -4 | ✅ |

| percent_correct (%) | 93.51 | 96.1 | 2.59 | ✅ |
| match_steps | 72 | 68 | -4 | ✅ |
| rotation_error (deg) | 56.34 | 58.43 | 2.09 | ❌ |
| runtime (min) | 13 | 15 | 2 | ❌ |
| episode_runtime (sec) | 126 | 150 | 24 | ❌ |

| percent_correct (%) | 97 | 99 | 2 | ✅ |
| match_steps | 99 | 99 | 0 | ⬜ |
| runtime (min) | 14 | 11 | -3 | ✅ |
| episode_runtime (sec) | 8 | 6 | -2 | ✅ |

| percent_correct (%) | 97 | 100 | 3 | ✅ |
| match_steps | 99 | 99 | 0 | ⬜ |
| runtime (min) | 25 | 19 | -6 | ✅ |
| episode_runtime (sec) | 15 | 11 | -4 | ✅ |

### DIFF
--- a/benchmarks/ycb_unsupervised_inference.csv
+++ b/benchmarks/ycb_unsupervised_inference.csv
@@ -1,3 +1,3 @@
 Experiment,Correct (%)|align right,Num Match Steps|align right,Run Time (mins)|align right,Episode Run Time (s)|align right
-unsupervised_inference_distinctobj_dist_agent,97.00,100,15,9
-unsupervised_inference_distinctobj_surf_agent,95.00,100,24,14
+unsupervised_inference_distinctobj_dist_agent,100.00,100,10,6
+unsupervised_inference_distinctobj_surf_agent,100.00,99,19,12


src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -56,6 +56,18 @@ class ChannelHypothesesUpdateTelemetry:
 
 
 class HypothesesUpdater(Protocol):
+    def pre_step(self) -> None:
+        """Runs once per step before updating the hypotheses."""
+        ...
```
jeremyshoemaker: This is really more of a style thing, so it's not critical. The Python docs aren't helpful on this point either, because they show this style inconsistently.

The usual way to have a method or function that does nothing is to use `pass` and not `...`. That said, a method or function that has a docstring doesn't need a body at all. You only need `pass` when there's no docstring, because there has to be at least one statement inside a block. So these `...` could just be removed completely.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -695,15 +704,24 @@ def get_evidence_for_each_graph(self):
...
+        available_graph_ids = []
+        available_graph_evidences = []
         for graph_id in graph_ids:
-            graph_evidences.append(np.max(self.evidence[graph_id]))
-        return graph_ids, np.array(graph_evidences)
+            if len(self.hyp_evidences_for_object(graph_id)):
+                available_graph_ids.append(graph_id)
+                available_graph_evidences.append(np.max(self.evidence[graph_id]))
+
+        return available_graph_ids, np.array(available_graph_evidences)
```
jeremyshoemaker, line 720: I noticed this inconsistency when I loaded up this branch in PyCharm. The type of the second value in this tuple is inconsistent in this function. Up on line 706, we return a `list`, but here we return an `np.array`. I think the one on 706 should be changed to match, since `list`s and `np.array`s behave slightly differently.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -695,15 +704,24 @@ def get_evidence_for_each_graph(self):
...
 
     def get_all_evidences(self):
         """Return evidence for each pose on each graph (pointer)."""
         return self.evidence
 
+    def hyp_evidences_for_object(self, object_id):
```
jeremyshoemaker: I don't think this is a good method to add, since it actually makes all the places it's used longer than they would be without really adding anything. It also adds in an additional function call which isn't free.

```python
self.evidence[object_id]
# vs
self.hyp_evidences_for_object(object_id)
```
If this method name is more descriptive, then it _might_ make sense to change the name of `self.evidence`, but I don't know that it adds much.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -750,6 +770,8 @@ def _update_possible_matches(self, query):
         self.previous_mlh = self.current_mlh
         self.current_mlh = self._calculate_most_likely_hypothesis()
 
+        self.hypotheses_updater.post_step()
```
jeremyshoemaker: I know we use this `pre_` and `post_` pattern in a lot of places already, but this really would be better as a context manager using `__enter__` and `__exit__`. The `__exit__` method also has the advantage of being called when an exception is raised allowing action to be taken if needed.

Then the body of this method would be something like:
```python
with self.hypotheses_updater:
    thread_list = []
    # The rest of the body
    ...
```

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -310,7 +357,7 @@ def update_hypotheses(
 
             # We only displace existing hypotheses since the newly resampled hypotheses
             # should not be affected by the displacement from the last sensory input.
-            if existing_count > 0:
+            if len(hypotheses_selection.maintain_ids) > 0:
```
jeremyshoemaker: You did this elsewhere, so let's be consistent.
```suggestion
            if len(hypotheses_selection.maintain_ids):
```

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -487,38 +530,32 @@ def _sample_existing(
...
+
         # Return empty arrays for no hypotheses to sample
-        if existing_count == 0:
+        if len(maintain_ids) == 0:
             # Clear all channel hypotheses from the tracker
-            remove_ids = np.arange(tracker.total_size(input_channel))
             tracker.clear_hyp(input_channel)
 
-            channel_hypotheses = ChannelHypotheses(
+            return ChannelHypotheses(
```
jeremyshoemaker, line 550: The function definition says it returns a `tuple`, but this return statement is only returning a `ChannelHypotheses`.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -487,38 +530,32 @@ def _sample_existing(
...
-            channel=input_channel,
-        )
 
         # Update tracker by removing the remove_ids
-        tracker.remove_hyp(remove_ids, input_channel)
+        tracker.remove_hyp(hypotheses_selection.remove_ids, input_channel)
 
         channel_hypotheses = mapper.extract_hypotheses(hypotheses, input_channel)
-        maintained_channel_hypotheses = ChannelHypotheses(
+        return ChannelHypotheses(
```
jeremyshoemaker, line 562: Same here regarding the return type.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -659,3 +696,26 @@ def _sample_informed(
             evidence=selected_feature_evidence,
             possible=possible_hyp,
         )
+
+    def _max_global_slope(self) -> float:
+        """Compute the maximum slope over all objects and channels.
+
+        Returns:
+            The maximum global slope if finitie, otherwise float("nan")
```
jeremyshoemaker:
```suggestion
            The maximum global slope if finite, otherwise float("nan")
```

src/tbp/monty/frameworks/models/goal_state_generation.py
```diff
@@ -909,7 +909,16 @@ def _check_conditions_for_hypothesis_test(self):
...
 
-        if top_id == second_id:
+        # This happens when all hypothesis spaces are empty
+        if top_id is None and second_id is None:
+            return False
+
+        if second_id is None:
+            # If we only have one object with a single hypothesis, we should not
+            # attempt to generate a goal state.
+            if len(self.parent_lm.hyp_evidences_for_object(top_id)) == 1:
```
jeremyshoemaker: Is this why that method exists? Would it make more sense to have a method on the LM class that returns all the information this method needs at once? Something like `get_top_two_mlh` that returns two dataclasses that contain both the IDs and the evidences (and whatever else we need)?

Currently we're fetching the IDs, then turning around and using those to fetch evidence, and maybe later doing more things. It seems like these pieces of information could be connected and fetched in one go, even in the cases where one of them isn't set.

It seems like this method knows too much about the details of `parent_lm`. We're even calling "private" methods on `parent_lm`, like `_threshold_possible_matches`, which is a red flag that this method knows too much about `parent_lm`.

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -143,6 +143,20 @@ def _channel_telemetry(
         assert isinstance(self, EvidenceGraphLM)
 
         mapper = self.channel_hypothesis_mapping[graph_id]
+
+        if input_channel not in mapper.channels:
+            channel_evidence = np.empty(shape=(0,), dtype=np.float64)
+            channel_rotations_inv = np.empty(shape=(0, 3, 3), dtype=np.float64)
+            channel_locations = np.empty(shape=(0, 3), dtype=np.float64)
```
jeremyshoemaker: PyCharm seems to think these are unused, and I also don't see any references to them. They get overwritten right after this `if` statement and the return below doesn't use them either.

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -184,9 +198,12 @@ def _theoretical_limit_target_object_pose_error(self) -> float:
         """
         assert isinstance(self, EvidenceGraphLM)
 
-        hyp_rotations = Rotation.from_matrix(
-            self.possible_poses[self.primary_target]
-        ).inv()
+        object_possible_poses = self.possible_poses[self.primary_target]
+        if not len(object_possible_poses):
+            return -1
```
jeremyshoemaker, line 199: Is this a sentinel value we're checking for elsewhere? Shouldn't the minimum possible pose error be 0?

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -415,50 +416,160 @@ def clear_hyp(self, channel: str) -> None:
...
+        return np.flatnonzero(self._maintain_mask).astype(int)
+
+    @property
+    def remove_ids(self) -> npt.NDArray[np.int_]:
+        """Returns the indices of removed hypotheses."""
+        return np.flatnonzero(~self._maintain_mask).astype(int)
+
+    def __len__(self) -> int:
+        """Returns the total number of hypotheses in the selection."""
+        return int(self._maintain_mask.size)
```
jeremyshoemaker: Doesn't `ndarray.size` already return an `int`? 

> _a.size_ returns a standard arbitrary precision Python integer.
> ~ [Docs for ndarray.size](https://numpy.org/doc/1.23/reference/generated/numpy.ndarray.size.html)

src/tbp/monty/frameworks/utils/graph_matching_utils.py
```diff
@@ -239,32 +239,36 @@ def get_scaled_evidences(evidences, per_object=False):
     scaled_evidences = {}
     if per_object:
         for graph_id in evidences.keys():
-            scaled_evidences[graph_id] = (
-                evidences[graph_id] - np.min(evidences[graph_id])
-            ) / (np.max(evidences[graph_id]) - np.min(evidences[graph_id]))
-            # put in range(-1, 1)
-            scaled_evidences[graph_id] = (scaled_evidences[graph_id] - 0.5) * 2
+            if len(evidences[graph_id]):
+                graph_evidences = evidences[graph_id]
```
jeremyshoemaker: Why not change this `for` loop to iterate over `evidences.items()`? That way we don't have to look up the value multiple times, but we still have both the key and the value.

src/tbp/monty/frameworks/utils/graph_matching_utils.py
```diff
@@ -239,32 +239,36 @@ def get_scaled_evidences(evidences, per_object=False):
...
-            maxev = np.max(evidences[graph_id])
-            if maxev > max_evidence:
-                max_evidence = maxev
+            graph_evidences = evidences[graph_id]
+            if len(graph_evidences):
+                min_evidence = min(min_evidence, np.min(graph_evidences))
+                max_evidence = max(max_evidence, np.max(graph_evidences))
+
         for graph_id in evidences.keys():
+            graph_evidences = evidences[graph_id]
```
jeremyshoemaker: Same thing. If we iterate over `evidence.items()` we get the `graph_id` and the `graph_evidences` at the same time.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -254,23 +162,15 @@ def test_sampling_count(self):
...
-        self._count_multiplier(rlm)
-        self._count_multiplier_maximum(rlm, pose_defined=True)
-        self._count_multiplier_maximum(rlm, pose_defined=False)
-
-        # test existing to informed ratio
-        self._count_ratio(rlm, pose_defined=True)
-        self._count_ratio(rlm, pose_defined=False)
+        self._resampling_multiplier(rlm)
+        self._resampling_multiplier_maximum(rlm, pose_defined=True)
+        self._resampling_multiplier_maximum(rlm, pose_defined=False)
```
jeremyshoemaker: You mentioned that the only reason we're doing all of these in a single test is because of concerns about setup costs for the experiment. 

I think we should refactor this class to make these individual methods into `test_` methods, and move the contents of `setUp` and `get_pretrained_resampling_lm` to the `setUpClass` classmethod. That way, the experiment gets set up only once for this whole class, and then each individual test can test what they need to.

This will shake out any unintended dependencies between these tests, and make it easier to run them in isolation if we're ever debugging a specific issue.

jeremyshoemaker (commented):
I know this is currently a Draft, but I started reading through it and had some comments that I didn't want to lose, so I just reviewed it anyway.

We can go through some of these when we pair later.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
jeremyshoemaker: See my notes below in the `goal_state_generation` changes, since that makes things more complicated.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
jeremyshoemaker: See https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass

ramyamounir:
Thanks @jeremyshoemaker, I addressed the comments we discussed, except for the additional tests. I'll be working on those next.

ramyamounir:
@jeremyshoemaker, I've added a few unit tests that specifically cover the burst sampling behavior. We can discuss if we need to add more when we pair next week.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -56,17 +57,18 @@ class ChannelHypothesesUpdateTelemetry:
 
 
 class HypothesesUpdater(Protocol):
-    def pre_step(self) -> None:
-        """Runs once per step before updating the hypotheses."""
-        ...
+    def __enter__(self) -> Self:
```
jeremyshoemaker: I don't think these context manager methods should be part of the `HypothesesUpdater` protocol, at least not directly. I'm not sure the right way to do it though. We might be able to have it extend the `ContextManager` from `typing`, but I'm not sure if that would cause problems with the Protocol.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -266,8 +270,10 @@ def pre_step(self) -> None:
         ):
             self.sampling_burst_steps = self.sampling_burst_duration
 
-    def post_step(self) -> None:
-        """Runs once per step after updating the hypotheses.
+        return self
```
jeremyshoemaker, line 273: Good! I was worried this detail of the `__enter__` method's contract might be missed.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -266,8 +270,10 @@ def pre_step(self) -> None:
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
```
jeremyshoemaker, line 275: It probably doesn't matter since the body is so simple, but do we need to do anything when an exception in the `with` block is raised? If not, we can just add a check at the beginning of the method:

```python
if exc_type:
    return False
```
Or:
```python
if not exc_type:
    # current body of method
```

Also, from the documentation, I think the return type should be a `bool`. Though, I don't think we're ever going to want to return True (which would suppress any thrown exceptions). We _could_ just return `None` since it is falsey, which is what the second version is going to do.

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -145,14 +145,10 @@ def _channel_telemetry(
...
         if input_channel not in mapper.channels:
-            channel_evidence = np.empty(shape=(0,), dtype=np.float64)
-            channel_rotations_inv = np.empty(shape=(0, 3, 3), dtype=np.float64)
-            channel_locations = np.empty(shape=(0, 3), dtype=np.float64)
-
             return HypothesesUpdaterChannelTelemetry(
                 hypotheses_updater=channel_telemetry.copy(),
                 evidence=np.empty(shape=(0,), dtype=np.float64),
-                rotations=np.empty(shape=(0, 3, 3), dtype=np.float64),
+                rotations=np.empty(shape=(0, 3), dtype=np.float64),
```
jeremyshoemaker, line 151: Whoa, I didn't catch that in my previous review. Did it come up in testing, or was it just working with the extra dimension?

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -38,22 +40,23 @@
 
 
 class ResamplingHypothesesUpdaterTest(TestCase):
-    def setUp(self) -> None:
-        super().setUp()
+    @classmethod
+    def setUpClass(cls) -> None:
+        """Runs once for all tests in this class."""
```
jeremyshoemaker: Tristan mentioned this in another PR recently. We don't need to document methods that are part of `TestCase`.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -38,22 +40,23 @@
...
+        super().setUpClass()
 
-        self.output_dir = tempfile.mkdtemp()
+        cls.output_dir = tempfile.mkdtemp()
```
jeremyshoemaker: We don't appear to be using this anywhere else outside of line 55 below, so it can be a local variable instead of being added to the class.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -157,20 +130,45 @@ def _resampling_multiplier_maximum(self, rlm, pose_defined):
...
+                graph_id=graph_id,
+            )
+            self.assertEqual(graph_num_nodes * resampling_multiplier, informed_count)
+
+        # Reset mapper
+        self.rlm.channel_hypothesis_mapping[graph_id] = ChannelMapper()
+
+    def test_resampling_multiplier_maximum_pose_defined(self):
+        """Test that resampling multiplier respects maximum scaling (pose defined)."""
+        self._resampling_multiplier_maximum(self.rlm, pose_defined=True)
```
jeremyshoemaker: What are we testing in these two tests? I assume it's what's being asserted on line 160, but that isn't obvious to me. I just want to make sure we're not missing any additional assertions in these tests.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -252,3 +252,243 @@ def test_update_hypotheses_ids_map_correctly(self) -> None:
...
+        self.updater.burst_trigger_slope = 1.0
+        self.updater.sampling_burst_duration = 5
+        self.updater.sampling_burst_steps = 0
+
+        # Set a low-slope tracker to trigger a burst.
+        # max_slope (0.5) <= burst_trigger_slope (1.0)
+        tracker = EvidenceSlopeTracker(min_age=0)
+        tracker.add_hyp(3, "patch")
+        tracker.update(np.array([0.0, 0.2, 0.1]), "patch")
+        tracker.update(np.array([0.25, 0.5, -0.1]), "patch")
```
jeremyshoemaker, line 139: Why are these "patch" and not "patch1"? Is that a default value somewhere, and if so, maybe we should change the `test_update_hypotheses_ids_map_correctly` test to use that instead of setting "patch1" explicitly. 

Anything that's not relevant to what's being tested shouldn't be changed because it makes things noisier than they need to be. If it is relevant, then there should be an explicit unit test for it.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -252,3 +252,243 @@ def test_update_hypotheses_ids_map_correctly(self) -> None:
...
+        self.assertEqual(informed_count % num_hyps_per_node, 0)
+
+        # Cannot exceed the available max number of hypotheses
+        self.assertLessEqual(informed_count, graph_num_nodes * num_hyps_per_node)
+
+    def test_sample_count_returns_zero_informed_count_when_not_in_burst(self) -> None:
+        """Test that _sample_count returns informed_count == 0 when not in burst.
+
+        When sampling_burst_steps == 0, _sample_count should return
+        informed_count == 0 regardless of other parameters.
```
jeremyshoemaker: > regardless of other parameters.

Would it make sense to do some property testing on this? Which parameters can vary without changing `informed_count`?

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -252,3 +252,243 @@ def test_update_hypotheses_ids_map_correctly(self) -> None:
...
+        tracker.update(np.array([0.0, 0.0, 0.0]), "patch")
+        tracker.update(np.array([0.5, 0.5, 0.5]), "patch")
+        self.updater.evidence_slope_trackers = {"object1": tracker}
+
+        burst_steps_history = []
+        for _ in range(5):
+            with self.updater:
+                burst_steps_history.append(self.updater.sampling_burst_steps)
+
+        self.assertEqual(burst_steps_history, [5, 4, 3, 2, 1])
```
jeremyshoemaker, line 341: Am I missing something or is this just asserting that `for` loops and `list.append` work as expected? Couldn't we rewrite the core part of this test to just the loop and remove this unneeded assert?

```suggestion
        for _ in range(5):
            with self.updater:
                pass
```

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -252,3 +252,243 @@ def test_update_hypotheses_ids_map_correctly(self) -> None:
...
+        At the start of an episode (no trackers), max_slope is -inf which is
+        below any threshold, so a burst should be triggered.
+        """
+        self.updater.burst_trigger_slope = 1.0
+        self.updater.sampling_burst_duration = 5
+        self.updater.sampling_burst_steps = 0
+        self.updater.evidence_slope_trackers = {}
+
+        with self.updater:
+            self.assertEqual(self.updater.sampling_burst_steps, 5)
```
jeremyshoemaker: I didn't think about this until now, so this might be relevant to tests above that also do this.

Writing the assert this way, inside the context manager, means that it checks after `__enter__` but before `__exit__`? Is that necessary for these tests or are we trying to test what the final state is?

If we're testing the final state it might make more sense to replace the `with` block body with a `pass` and move this assert to after it.

jeremyshoemaker (commented):
These changes look good overall, especially the new unit tests.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
ramyamounir: Yeah, I think this might work fine, how does [this](https://github.com/ramyamounir/tbp.monty/commit/e43b84626cf2c007ef70926bdd2073143af22866) look?

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
ramyamounir: Updated in [7ccd2f208d75](https://github.com/ramyamounir/tbp.monty/commit/7ccd2f208d757e88826c924c61b5e4e909d1a9b8)

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
ramyamounir: Removed in [be0869613d6f](https://github.com/ramyamounir/tbp.monty/commit/be0869613d6f3a203d93cc66c70980997769f4fe)

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
ramyamounir: Fixed in [be0869613d6f](https://github.com/ramyamounir/tbp.monty/commit/be0869613d6f3a203d93cc66c70980997769f4fe)

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
ramyamounir: That's a good point. Looking more closely at this whole class now (`ResamplingHypothesesUpdaterTest`), it seems like what we're trying to test with these integration tests is now covered in the new unit tests. They mainly test that we calculate the correct number of hypotheses with sample_count, but we now cover that in the new function `test_sample_count_returns_*`. I think we can remove this integration test class, what do you think?

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
ramyamounir: Yes, I think patch is the [default value](https://github.com/thousandbrainsproject/tbp.monty/blob/be0869613d6f3a203d93cc66c70980997769f4fe/conf/experiment/config/monty/patch_and_view_sota.yaml?plain=1#L12). But we use [patch_0](https://github.com/thousandbrainsproject/tbp.monty/blob/be0869613d6f3a203d93cc66c70980997769f4fe/conf/experiment/config/monty/two_lm_stacked.yaml?plain=1#L22) and [patch_1](https://github.com/thousandbrainsproject/tbp.monty/blob/be0869613d6f3a203d93cc66c70980997769f4fe/conf/experiment/config/monty/two_lm_stacked.yaml?plain=1#L45) in stacked LM experiments when we have more than 1 sensor module. I'll change all of them to "patch" since there is no reason to expect a hierarchical experiment here, but it really doesn't matter. Updated in [56500dff3d41](https://github.com/ramyamounir/tbp.monty/commit/56500dff3d413cb607f529a3e248c99b8a41c147)

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
ramyamounir: I was just testing that every `__exit__` decrements by exactly one and that it starts from the `sampling_burst_duration` we set at the beginning of the function (i.e., `5`). If, e.g., the `__exit__` function sets `sampling_burst_steps=0` without properly decrementing by one every step, then what you suggested would still pass, even though it doesn't properly decrement.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
ramyamounir: Yes, I did take that into consideration. I think it's confusing to assert against the decremented final state after we exit the context manager. I want to use the state during the `update_hypotheses` function instead. That said, I still test the `__exit__` decrement logic in other unit tests. I have changed this test a bit in [df588e6ad791](https://github.com/ramyamounir/tbp.monty/commit/df588e6ad791ebecfdb12e075716835e8aa066d9), by making sure the burst sets the number of steps to `sampling_burst_duration`. If I were to use the final state, we would assert that it equals `max(0, sampling_burst_duration -1)`, which can be confusing.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
ramyamounir: Yes, I've added a couple more variables [here](https://github.com/ramyamounir/tbp.monty/commit/5c8898b8f401125fca680f9a584b4b074716522f) that would make sense.

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
ramyamounir: It was working, the linter won't catch it because it's just expecting an empty float array, so any shape would work. We're also not doing anything with these values except writing them in the logs when the hypothesis space is empty. I just noticed it when fixing the error you pointed out earlier.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
ramyamounir: I removed the integration tests, but let me know if I'm missing something about why we should keep those.

jeremyshoemaker: It might make sense to compare the coverage report for running these tests vs the old report to make sure we haven't lost any coverage of branches that we were covering before. Like, run the tests for just this file, before and after. We're getting coverage from other tests, so we don't want that to conflate what these specific tests are testing or not testing.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
jeremyshoemaker: Oh 🤦‍♂️ , for some reason I wasn't seeing that we're pulling the current value out of the updater.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
jeremyshoemaker: Just some more notes, since I'm out today and don't want to leave you hanging. Specifically, I'd run `pytest` with the `--cov` and `--cov-report html` flags. I'm not sure if the second flag implies the first, haven't tried it yet, so I'm just copying what our CI tests do.

Then open up the `htmlcov/index.html` in a browser and find the coverage report for the updater. It will give you some nice details on what code paths were covered, which ones weren't and sometimes why (like a branch always being true).

ramyamounir: Cool, I ran the coverage report and looks like it's exactly the same with and without these integration tests.
pic-selected-260122-1545-51
pic-selected-260122-1541-53

tristanls-tbp: @ramyamounir, in the browser, you should be able to click on those files and see line-by-line visuals. Here's an example of what I see for a different test I ran. The red lines show me that the test did not exercise those. For example, my tests did not cover the case where `action_params` is `None`:

Screenshot 2026-01-22 at 15 23 20

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -71,32 +73,44 @@ class ChannelHypothesesResamplingTelemetry(ChannelHypothesesUpdateTelemetry):
...
+    objects hypothesis spaces. If the hypothesis with the highest slope is unable to
+    accumulate evidence at a high enough slope, i.e., none of the current hypotheses
+    match the incoming observations well, a sampling burst is triggered. A sampling
+    burst adds new hypotheses over a specified `sampling_burst_duration` number of
+    consecutive steps to all hypothesis spaces. This burst duration reduces the effect
+    of sensor noise. Hypotheses are deleted when their smoothed evidence slope is below
+    `deletion_trigger_slope`.
+
+    The resampling process is governed by four main parameters:
+      - `resampling_multiplier`: Determines the number of the hypotheses to resample
```
nielsleadholm:
```suggestion
      - `resampling_multiplier`: Determines the number of hypotheses to resample
```

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -71,32 +73,44 @@ class ChannelHypothesesResamplingTelemetry(ChannelHypothesesUpdateTelemetry):
...
-        It would be better to decouple the amount of hypotheses added from the amount
-        deleted in each step. At the moment, this is decided by the
-        `hypotheses_count_multiplier`. For example, when the multiplier is set to 1.0,
-        the hypotheses sampled is equal to the hypotheses removed. We ideally can
-        decouple theses then use a slope threshold to decide on which hypotheses to
-        remove, and use prediction error or other heuristics to decide to how many
-        hypotheses to resample.
+    hypotheses only at the beginning of the episode, you can set:
+        - `resampling_multiplier=2` (or `umbilical_num_poses` if PC undefined)
+        - `deletion_trigger_slope=-1.0` (no deletion is allowed)
```
nielsleadholm: Didn't you find that this could still result in deletions due to numerical imprecision? Wasn't it more consistent if this was set to `-np.inf`?

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -71,32 +73,44 @@ class ChannelHypothesesResamplingTelemetry(ChannelHypothesesUpdateTelemetry):
...
+        - `sampling_burst_duration=1` (sample the full burst over a single step)
+        - `burst_trigger_slope=-1.0` (never trigger additional bursts)
```
nielsleadholm: As above

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -71,32 +73,44 @@ class ChannelHypothesesResamplingTelemetry(ChannelHypothesesUpdateTelemetry):
...
+
+    These parameters will trigger a single-step burst at the first step of the episode.
+    Note that if the PC of the first observation is undetermined,
+    `resampling_multiplier` should be set to `umbilical_num_poses` to reproduce the
```
nielsleadholm: It might be worth clarifying that you need to predict this number, i.e. you can't literally set `resampling_multiplier="umbilical_num_poses"` as a string or variable, and then expect it to handle this. Your description kind of alludes to this, but I think it might not be clear to someone returning to this.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -148,11 +164,17 @@ def __init__(
             features_for_matching_selector: Class to
                 select if features should be used for matching. Defaults to the default
                 selector.
-            hypotheses_count_multiplier: Scales the total number of hypotheses
-                every step. Defaults to 1.0.
-            hypotheses_existing_to_new_ratio: Controls the proportion of the
-                existing vs. newly sampled hypotheses during resampling. Defaults to
-                0.0.
+            resampling_multiplier: Determines the number of the hypotheses to resample
```
nielsleadholm: Separately you use the language sampling over resampling (e.g. sampling_burst_duration). I think this is a bit clearer, e.g. we sample on the first step, and we don't necessarily "resample" in the way that is often meant in probabilistic settings (that would be more akin to the "offspring hypotheses" we've discussed, i.e. resampling based on an updated distribution). As such, I think just using "sampling" would be clearer.

src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```diff
@@ -406,76 +470,62 @@ def _sample_count(
...
         self,
-        existing_count: int,
+        hypotheses_selection: HypothesesSelection,
         hypotheses: Hypotheses,
         input_channel: str,
         mapper: ChannelMapper,
         tracker: EvidenceSlopeTracker,
-    ) -> tuple[ChannelHypotheses, npt.NDArray[np.int_]]:
+    ) -> ChannelHypotheses:
         """Samples the specified number of existing hypotheses to retain.
```
nielsleadholm, line 524: Re-reading this, it might be worth clarifying that we are not sampling these in a probabilistic sense (e.g. randomly choosing, or generating hypotheses randomly using existing hypotheses as a kind of seed). Instead, we will deterministically determine which hypotheses to maintain.

src/tbp/monty/frameworks/models/goal_state_generation.py
```diff
@@ -911,7 +911,16 @@ def _check_conditions_for_hypothesis_test(self):
...
+        # This happens when all hypothesis spaces are empty
+        if top_id is None and second_id is None:
+            return False
+
+        if second_id is None:
+            # If we only have one object with a single hypothesis, we should not
+            # attempt to generate a goal state.
+            if len(self.parent_lm.evidence[top_id]) == 1:
+                return False
+
```
nielsleadholm, line 923: Given the updated logic for when only one object remains in the hypothesis space, I would suggest updating the below comment:

```
            # If we only know (i.e. have learned) about one object, we can focus on pose
            # In this case, get_top_two_mlh_ids returns the same IDs for top_id and
            # second_id
```

src/tbp/monty/frameworks/utils/evidence_matching.py
```diff
@@ -86,48 +86,49 @@ def channel_range(self, channel_name: str) -> tuple[int, int]:
...
         if channel_name not in self.channel_sizes:
             raise ValueError(f"Channel '{channel_name}' not found.")
-        if self.channel_sizes[channel_name] + value <= 0:
-            raise ValueError(
-                f"Channel '{channel_name}' size cannot be negative or zero."
-            )
-        self.channel_sizes[channel_name] += value
+        if new_size < 0:
+            raise ValueError(f"Channel '{channel_name}' size must be positive.")
+        if new_size == 0:
```
nielsleadholm, line 114: This is a basic question, but I'm struggling to remember - do we store hypotheses entirely independently for different channels? If a high level LM gets input from an SM, as well as a low-level LM, would the initialized hypotheses given an observation of a child object be entirely distinct from the hypotheses initialized by the SM input? Intuitively I would have thought that we would want a single hypothesis to be informed by both (e.g. centre of mug gains evidence both for the sensed surface, as well as the presence of the logo). Otherwise the hypothesis spaces are essentially totally separate, but I'm just realizing now that we might not currently capture this.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
ramyamounir: Nice, thanks @tristanls-tbp. I did what you suggested and started covering the branches that were not covered by the existing tests. I'm now at 98% :slightly_smiling_face: 

@jeremyshoemaker, I ended up adding tests for the _sample_informed function, which is not really part of the burst sampling PR scope. Let me know if you want those in a different PR.

pic-selected-260123-0832-58

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -335,28 +100,683 @@ def test_update_hypotheses_ids_map_correctly(self) -> None:
...
+        """
+        self.updater.burst_trigger_slope = 1.0
+        self.updater.sampling_burst_duration = 5
+        self.updater.sampling_burst_steps = 0
+
+        # Low max_slope hypotheses to trigger a burst in the first iteration.
+        tracker = EvidenceSlopeTracker(min_age=0)
+        tracker.add_hyp(3, "patch")
+        tracker.update(np.array([0.0, 0.0, 0.0]), "patch")
+        tracker.update(np.array([0.5, 0.5, 0.5]), "patch")
```
nielsleadholm, line 333: To check my understanding, is this line meant to be another potential trigger for bursting (but blocked due to the `sampling_burst_steps > 0` condition)?

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -335,28 +100,683 @@ def test_update_hypotheses_ids_map_correctly(self) -> None:
...
+        self.assertEqual(max_slope, float("-inf"))
+
+    @given(
+        pose_fully_defined=st.booleans(),
+        num_euler_angles=st.integers(min_value=1, max_value=10),
+    )
+    def test_num_hyps_per_node_with_initial_possible_poses(
+        self, pose_fully_defined, num_euler_angles
+    ) -> None:
+        """Test _num_hyps_per_node returns len(initial_possible_poses) when set.
```
nielsleadholm:
```suggestion
        """Test _num_hyps_per_node returns len(initial_possible_poses) when latter is set.
```

src/tbp/monty/frameworks/utils/evidence_matching.py
ramyamounir: Yes, we store hypotheses for different channels completely independently. In the `ChannelMapper` and the `EvidenceSlopeTracker`, they are separated by different dictionary keys. In `EvidenceGraphLM`, the channels are stored in one array back-to-back and we use the ChannelMapper to index and perform operations on the channels we want. In the stacked LM pretraining experiment, the HLLM creates independent graphs; one for each channel (patch_1 and learning_module_0).

> Intuitively I would have thought that we would want a single hypothesis to be informed by both

Yeah, good point. Since stacked LMs have co-located receptive field, I think we can merge the hypotheses from SM1 and LM0 into a single channel. Maybe we can discuss this in the next meeting, I'm sure I'm probably missing something about why we separate hypotheses into channels and maintain that separation (e.g., voting, flexible non co-located LM-to-LM wiring). But I agree that it would also be more efficient to merge these hypotheses from different channels when possible to reduce the size of the hypothesis space during inference.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
jeremyshoemaker: I think it's related enough that it falls under the "leave things better than you found it" idea. We're modifying this class, we saw that its tests are lacking coverage and we improved that while we're here.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
ramyamounir: Yes, exactly. This line consistently forces the max slope to be `0.5`, which when compared against the `burst_trigger_slope=1.0` will always try to trigger a burst. The burst will be triggered only at the first step then it will be blocked for subsequent steps for the duration of the burst, as you mentioned.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -39,138 +38,6 @@
...
-    def test_resampling_multiplier_maximum_pose_defined(self):
-        """Test that resampling multiplier respects maximum scaling (pose defined)."""
-        self._resampling_multiplier_maximum(self.rlm, pose_defined=True)
-
-    def test_resampling_multiplier_maximum_pose_undefined(self):
-        """Test that resampling multiplier respects maximum scaling (pose undefined)."""
-        self._resampling_multiplier_maximum(self.rlm, pose_defined=False)
-
-
 class ResamplingHypothesesUpdaterUnitTestCase(TestCase):
```
jeremyshoemaker: Now that you're removing the other one, we can simplify the name of this one.
```suggestion
class ResamplingHypothesesUpdaterTest(TestCase):
```

src/tbp/monty/frameworks/models/goal_state_generation.py
ramyamounir: Nice catch, thanks.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -479,16 +354,429 @@ def test_max_global_slope_returns_inf_when_no_trackers(self) -> None:
...
+        )
+        self.mock_graph_memory.get_locations_in_graph = Mock(
+            return_value=np.zeros((channel_size, 3))
+        )
+
+        mapper = ChannelMapper(channel_sizes={"patch": channel_size})
+        self.updater.evidence_slope_trackers = {}
+
+        # Create a pre-initialized tracker with the "patch" channel
+        # This will patch the call to `EvidenceSlopeTracker`
```
jeremyshoemaker: Did you mean something like this? The current version seems ambiguous since this section of code doesn't do the patching.
```suggestion
        # This will be used in the mocked call to `EvidenceSlopeTracker`
```
I'm suggesting not using the "patch" term here since we have "patch" as part of our domain logic, to prevent confusion.

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -479,16 +354,429 @@ def test_max_global_slope_returns_inf_when_no_trackers(self) -> None:
...
+        )
+        self.mock_graph_memory.get_locations_in_graph = Mock(
+            return_value=np.random.rand(num_nodes, 3)
+        )
+
+        # Set up updater with feature matching enabled
+        self.updater.use_features_for_matching = {"patch": True}
+        self.updater.feature_weights = {"patch": {"feature1": 1.0}}
+        self.updater.tolerances = {"patch": {"feature1": 0.1}}
+        self.updater.feature_evidence_increment = 1
```
jeremyshoemaker: We seem to be doing this in several of these tests, but this is the first time I noticed it.

Are these fields that would normally be set from the constructor? It might make sense to change this code to create a new updater for just this test, using those constructor arguments. In this case they're probably just being set like this, but that could change and the way this is currently written, the test knows too much about the internals of the class (which is a fuzzy line, I know). We're setting things internally elsewhere because the constructor doesn't give us a way to change them otherwise.

We want to minimize how much a test depends on the internals of the thing being tested, that way we can make changes to them without having to change the tests to match. As much as possible, we want to be able to change code OR tests in between test runs, and not be forced to change both at the same time.

If the updater is easy to create (which I think it is) we should just create a new one if we need to. The one from `setUp` is just a handy shortcut for the tests that don't need to manipulate as much of the initial state.

ramyamounir: I see, how does [02a0b714332d](https://github.com/ramyamounir/tbp.monty/commit/02a0b714332d05592ca0fa4b05660bec5bb61f01) look?

src/tbp/monty/frameworks/utils/evidence_matching.py
nielsleadholm: Ok that's really helpful thanks, yeah let's discuss it in today's meeting. (also for the avoidance of doubt, this was just a separate discussion point that came up, not an issue/blocker with this PR)

src/tbp/monty/frameworks/models/goal_state_generation.py
```diff
@@ -922,8 +922,6 @@ def _check_conditions_for_hypothesis_test(self):
                 return False
 
             # If we only know (i.e. have learned) about one object, we can focus on pose
```
nielsleadholm: Suggestion (since it's not only a situation that occurs due to only knowing one object)

```python
# If the LM's hypothesis space only contains one object, focus on pose
```

tests/unit/frameworks/models/evidence_matching/resampling_hypotheses_updater_test.py
```diff
@@ -36,7 +38,7 @@
 )
 
 
-class ResamplingHypothesesUpdaterUnitTestCase(TestCase):
+class ResamplingHypothesesUpdaterTest(TestCase):
```
nielsleadholm, line 41: Suggestion: `ResamplingHypothesesUpdaterTest` --> `SamplingHypothesesUpdaterTest`

ramyamounir: The test class name here matches the updater class name (`ResamplingHypothesesUpdater`). It still makes sense to me that we would call it resampling. In my mind, both updaters do sample, the `DefaultHypothesesUpdater` samples one time at the beginning of every episodes, whereas the `ResamplingHypothesesUpdater` continues to re-sample during the episode as it needs to.

For more clarity maybe we can call it `BurstSamplingHypothesesUpdater`?

nielsleadholm: Makes sense, yeah I like `BurstSamplingHypothesesUpdater`

ramyamounir: Ok cool, I started renaming resampling to burst sampling but then I realized that this is a larger refactor as it includes changes in many different places of the code base, including configurations. I think it would be better to isolate this change to a separate PR for easier reviewing. I started a [draft PR](https://github.com/thousandbrainsproject/tbp.monty/pull/737) as a follow-up to this one ([Renaming commit](https://github.com/thousandbrainsproject/tbp.monty/pull/737/commits/c9ed001beaffbdd4d79c5fd7bd21c20eb32b26cf)).

@jeremyshoemaker, FYI

ramyamounir:
Thanks for the reviews @jeremyshoemaker and @nielsleadholm . I will be merging the PR shortly, but I will first rerun the unsupervised inference experiments and update the benchmark if needed.

ramyamounir:
### Updated Benchmarks (tagged `PR#700a`)

| percent_correct (%) | 97 | 100 | 3 | ✅ |
| match_steps | 100 | 100 | 0 | ⬜ |
| runtime (min) | 15 | 10 | -5 | ✅ |
| episode_runtime (sec) | 9 | 6 | -3 | ✅ |

| percent_correct (%) | 95 | 100 | 5 | ✅ |
| match_steps | 100 | 99 | -1 | ✅ |
| runtime (min) | 24 | 19 | -5 | ✅ |
| episode_runtime (sec) | 14 | 12 | -2 | ✅ |

PR #701, opened 2026-01-16
test: Make hierarchy test less brittle

vkakerbeck:
Okay, I did a bit of debugging on the issue described in this PR: https://github.com/thousandbrainsproject/tbp.monty/pull/678

This draft PR is mostly to highlight some of the issues and potential fixes. I think the exploratory steps multiplier fix (described below) will need to be its own PR with a rerun of pretraining and benchmarks. Here is a description of the proposed fixes to remove brittleness and pass the tests again:

1. Fixed action sequence using the spiral policy
The brittleness of this test largely comes from the randomness in the policy. Since we take random actions, those will be different with different seeds and not all action sequences will make Monty recognize the objects as expected in this test. This test builds on the assumption that Monty receives unique features and locations of the two objects so that it properly detects them as new objects in the first episodes and recognizes them when presented again. If it doesn't receive unique features (which can for example happen when only exploring a local area on the object, since they are quite similar locally), we can't recognize the correct objects and test graph building and extension in a hierarchy. 
**The fix**: Update the config to use NaiveSpiralPolicy with stepsize 10 (`fixed_amount: 10`) to make the action sequence deterministic and make sure a large enough area of the objects is covered to distinguish and recognize them.

2. Lowering the number of exploratory steps
After fix 1, the asserts in `check_hierarchical_lm_train_results` and `check_hierarchical_lm_eval_results` pass (i.e. we correctly detect no_match when expected and the objects later on) but this check https://github.com/thousandbrainsproject/tbp.monty/blob/12fa632288b4562317940e8c1f81ea58769da8a4/tests/unit/resources/unit_test_utils.py#L428 still fails (i.e. the model does not contain more points after the second epoch, when the objects were recognized). This is because after epoch 0, the model already contains all the observations that can land on the object with the spiral policy (i.e. we took exploratory steps until the sensor was off the object). So even though Monty tries to extend the graph after recognizing the object in epoch 2, there are no novel observations that aren't already stored in the graph.
**The fix**: Lower number of exploratory steps by setting `num_exploratory_steps: 30` in the config. (There are 73 steps before the spiral policy calls StopIterator and ends the experiment https://github.com/thousandbrainsproject/tbp.monty/blob/12fa632288b4562317940e8c1f81ea58769da8a4/src/tbp/monty/frameworks/experiments/object_recognition_experiments.py#L121 )

3. Fix a bug in Monty that made num_exploratory_steps grow with each no_match episode
This line https://github.com/thousandbrainsproject/tbp.monty/blob/12fa632288b4562317940e8c1f81ea58769da8a4/src/tbp/monty/frameworks/models/graph_matching.py#L149 multiplies the number of exploratory steps we want to take by 10 if no_match is detected. The idea was that if we build a new model we want to take more exploratory steps than when we add some knowledge to an existing model. However, the way we apply the multiplier is by changing a property of the LM class. This property does not get reset at the end of an episode, which means with every no_match episode, the number of exploratory steps increases by a factor of 10. 
**The fix**: One solution would be to reset this property at the end of an episode. However, I think it can cause some confusing behaviors if we change variables set in the configs in other places in the code and don't think using more exploratory steps in the no_match case is crucial. I would propose just removing this line. @nielsleadholm let me know if you disagree.
This is a bug that definitely needs to be fixed. I think it will need to be in a separate PR and we will likely need to retrain our models and update benchmark results after the fix. If may not have a large effect for the models learned with distant agent (or any effect at all) bc the spiral policy terminates once it reaches a certain eccentricity so after num_exploratory_steps goes above that number it makes no difference anymore what its value is. The effect may be more for the surface agent models (ideally making pretraining run much faster), although also there, it may not be as extreme if pretraining is run in parallel.

4. Update parameters so that LM1 converges fast enough to no_match
After making all the above changes, the first episode on the second object was still covering all of the object, such that on second presentation no new points would be added. This was because the matching phase took too many steps to converge onto no_match. Then matching_steps + exploratory_steps was > 72 (when spiral went off the object). The large number of matching steps were because LM1 would only get input on step 0 and step 12, so it needed at 13 steps to notice the observed object doesn't match with the model of object 0. After fixing graph_delta_thresholds, it would get input from LM0 too early, LM0 would then converge on no_match and not send any further input, and it wouldn't be able to decrement that evidence again (I'll still need to look into this general problem but it is unrelated to this unit test).
**The fix**: Set `distance: 0.02` in the `delta_thresholds` of SM1 (to make LM1 receive input on more steps) and `object_evidence_threshold: 20` for LM0 (to prevent LM0 to send object IDs to LM1 too early). 

With the fixes above, we should deterministically get the behavior we want and be able to test whether Monty correctly builds and extends compositional models. I agree with @tristanls comment here https://github.com/thousandbrainsproject/tbp.monty/pull/678#issuecomment-3729824878 that the asserts on the .csv files aren't strictly necessary, as those things are already tested in other places and the important part of the test here is `check_hierarchical_models`. However, it might help with debugging future errors to leave them in, since knowing that objects were not correctly recognized would help narrow down why the models don't look as expected.

### DIFF
--- a/src/tbp/monty/frameworks/models/graph_matching.py
+++ b/src/tbp/monty/frameworks/models/graph_matching.py
@@ -145,8 +145,6 @@ class MontyForGraphMatching(MontyBase):
                 all_lms_no_match = False
 
         if all_lms_no_match:
-            # Take more exploratory steps if we are building a new graph
-            self.num_exploratory_steps = self.num_exploratory_steps * 10
             # No need to check any other conditions if all LMs have no_match
             return True
 


tristanls:
@tristanls-tbp

vkakerbeck:
Ah sorry, just blindly took githubs first suggestion!

vkakerbeck:
For the record, this is how the training stats look like:
Screenshot 2026-01-16 at 3 04 51 PM
As described in the unit test, on the first 4 episodes, we want to recognize distinct objects (episode 2 and 3 show the same objects from angles that look uniquely different) and learn a new model for each of them. On episode 5 and 6 we want to recognize object 0 and 1 respectively and extend their models.

tests/unit/hierarchy_test.py
```diff
@@ -66,6 +66,7 @@ def setUp(self):
             )
 
     def tearDown(self):
+        """Code that gets executed after every test."""
```
tristanls-tbp: **issue**: Please don't reintroduce this docstring. There is no other possible interpretation for `tearDown` in the context of `setUp`, `tearDown` methods in a test suite. The docstring is extraneous.

vkakerbeck: Good point

vkakerbeck:
Moving out of draft mode for review. However, I opened this PR https://github.com/thousandbrainsproject/tbp.monty/pull/702 to make the num_exploratory_steps change separately as this one might affect benchmarks.

tristanls-tbp (approved):
Looks good to me.

src/tbp/monty/frameworks/models/graph_matching.py
```diff
@@ -145,8 +145,6 @@ def check_terminal_conditions(self):
                 all_lms_no_match = False
 
         if all_lms_no_match:
-            # Take more exploratory steps if we are building a new graph
-            self.num_exploratory_steps = self.num_exploratory_steps * 10
```
tristanls-tbp: **question**: Do you want to remove this change since #702 will test it?

tristanls-tbp:
@vkakerbeck one more note on the PR/commit title. Currently, this PR includes the #702 change, and that makes this PR a `fix!: Make hierarchy test less brittle`, because it changes the source code while `test:` is reserved for test-only changes. However, if you remove that code change, then the `test: Make hierarchy test less brittle` will be fine.

src/tbp/monty/frameworks/models/graph_matching.py
vkakerbeck: yes I'll remove it but that means the test will not pass anymore on this branch until the fix is merged

vkakerbeck:
I removed it. I originally had it in here to show that the test passes. I'm planning to merge the fix PR before this one here anyways.

conf/test/hierarchy/two_lms_heterarchy.yaml
```diff
@@ -3,13 +3,17 @@ defaults:
   - /experiment/config/monty/clear_monty_config@config
   - /experiment/config/monty/two_lm_stacked@config.monty_config
   - /experiment/config/monty/learning_modules/clear_learning_module_configs@config.monty_config
+  - /experiment/config/monty/motor_system/clear_motor_system_config@config.monty_config
+  - /experiment/config/monty/motor_system/naive_scan_spiral@config.monty_config.motor_system_config
+  - /experiment/config/monty/motor_system/policy/naive_scan@config.monty_config.motor_system_config.motor_system_args.policy_args
   - /experiment/config/environment/two_lm_stacked_distant_mount_habitat@config.env_interface_config
 
 config:
+  seed: 44
```
tristanls-tbp, line 12: **note**: Probably no longer necessary. This was a demo of a breaking path, but the changes made RNG a no-factor.

nielsleadholm:
Nice debugging! I agree about removing that line that modifies the number of exploratory steps. Even if there is a minor performance hit (which doesn't seem likely), it's confusing to alter user-set hyperparameters like that.

PR #702, opened 2026-01-16
fix!: remove exploratory steps multiplier

vkakerbeck:
As described under 3 here https://github.com/thousandbrainsproject/tbp.monty/pull/701 we can some unintended behavior with the num_exploratory_steps multiplier. This PR removes those lines. 
I am still running pretraining and benchmarks to see whether this change affects them.

### DIFF
--- a/src/tbp/monty/frameworks/models/graph_matching.py
+++ b/src/tbp/monty/frameworks/models/graph_matching.py
@@ -145,8 +145,6 @@ class MontyForGraphMatching(MontyBase):
                 all_lms_no_match = False
 
         if all_lms_no_match:
-            # Take more exploratory steps if we are building a new graph
-            self.num_exploratory_steps = self.num_exploratory_steps * 10
             # No need to check any other conditions if all LMs have no_match
             return True
 


vkakerbeck:
Ok, after running some test experiments (getting identical results) and thinking about it a bit more, I don’t think it would have any effect on supervised pretraining experiments. The condition only applies in no_match cases, but during supervised pretraining we always provide a label, even on the first episode. Therefor this multiplier is never applied and the fix here has no effect on pretrained models or benchmark inference experiments that use the models.
I do think it would have an effect on the unsupervised learning experiments https://thousandbrainsproject.readme.io/docs/benchmark-experiments#unsupervised-learning However, those seem to be broken at the moment so I would fix them in a separate PR and simply adjust the number of exploratory steps in the config then if necessary.
Moving this PR out of draft mode now.

jeremyshoemaker:
@tristanls-tbp This looks like a straightforward change, but you have more context from the other PR than I have currently.

tristanls-tbp (approved):
Feels right 🙂.

PR #703, opened 2026-01-20
test: fix brittle semisupervised heterarchy test

vkakerbeck:
Applying same fix as for https://github.com/thousandbrainsproject/tbp.monty/pull/701 (making action sequence deterministic by using the spiral policy. Also fixed assert that assumed the second evaluated rotation wasn't shown during training.

(also snuck in addressing this comment https://github.com/thousandbrainsproject/tbp.monty/pull/701#discussion_r2700032324 )

### DIFF
--- a/tests/unit/hierarchy_test.py
+++ b/tests/unit/hierarchy_test.py
@@ -276,7 +276,7 @@ class HierarchyTest(unittest.TestCase):
                     self.assertIn(
                         eval_stats["primary_performance"][episode * 2 + lm_id],
                         ["correct", "correct_mlh"],
-                        f"LM {lm_id} did not recognize the object on first episode.",
+                        f"LM {lm_id} did not recognize the object.",
                     )
             # check that prediction errors are logged
             self.assertIn(


tests/unit/hierarchy_test.py
```diff
@@ -270,22 +270,14 @@ def test_semisupervised_stacked_lms_experiment(self):
...
-                    eval_stats["primary_performance"][episode * 2 + lm_id],
-                    "LMs should not recognize object on second episode as it is a "
-                    "previously unseen view.",
-                )
+            for episode in range(2):
+                for lm_id in range(num_lms):
+                    self.assertIn(
+                        eval_stats["primary_performance"][episode * 2 + lm_id],
+                        ["correct", "correct_mlh"],
+                        f"LM {lm_id} did not recognize the object on first episode.",
```
tristanls-tbp: **issue (non-blocking)**: This is now technically the first two episodes, yes?

vkakerbeck: yes, which is the entire experiment

vkakerbeck: Oh I see, the error message need to be updated

PR #711, opened 2026-01-23
feat: enable basic and detailed logging for supervised pretraining experiments

ramyamounir:
While working on compositional benchmark, I ended up making some changes that would allow us to use our loggers with the pretraining experiments. At the moment, the loggers don't work for a few reasons:
1. In a stacked LM experiment, if the LM is not supervised (e.g., LL-LM), we [wipe the hypotheses](https://github.com/thousandbrainsproject/tbp.monty/blob/4348e42d445fd6b81ffd21a80776933ba6591ad9/src/tbp/monty/frameworks/experiments/pretraining_experiments.py?plain=1#L127-L133) to prevent updating the models. However, this is not enough because the logger [expects](https://github.com/thousandbrainsproject/tbp.monty/blob/4348e42d445fd6b81ffd21a80776933ba6591ad9/src/tbp/monty/frameworks/utils/logging_utils.py?plain=1#L588-L589) a `detected_object` to exist if the `primary_performance=="match"`. To solve this we can also wipe the `lm.possible_matches` which would [set](https://github.com/thousandbrainsproject/tbp.monty/blob/4348e42d445fd6b81ffd21a80776933ba6591ad9/src/tbp/monty/frameworks/utils/logging_utils.py?plain=1#L581-L582) the `primary_performance` to "no_match" in the logger.
2. The detected_pose is wiped in the experiment, but the logger expects a value for it. So, I handle it by checking if it's set to `None` [here](https://github.com/thousandbrainsproject/tbp.monty/blob/5c49bf17192f0f6c6bbe94dce4696e01bc502c65/src/tbp/monty/frameworks/utils/logging_utils.py?plain=1#L666-L672)
3. The run_parallel script does not consolidate logging handler files in the cleanup, I added the required functions to do so [here](https://github.com/thousandbrainsproject/tbp.monty/blob/5c49bf17192f0f6c6bbe94dce4696e01bc502c65/src/tbp/monty/frameworks/run_parallel.py?plain=1#L542-L563)

After merging in these changes, we should be able to run an experiment like this:
```zsh
python run_parallel.py experiment=supervised_pre_training_objects_with_logos_lvl1_comp_models \
episodes=\'0\' \
+experiment/config/logging=detailed_evidence_lm
```

and get detailed_run_stats.json with all 14 episodes for the 14 views we use to learn the object, as shown below:

pic-selected-260123-1720-41

**I'm tagging the compositional benchmark team because this LM wiping code was added during the compositional IP**


src/tbp/monty/frameworks/run_parallel.py
```diff
@@ -535,6 +535,33 @@ def post_parallel_train(experiments: list[Mapping], base_dir: Path) -> None:
...
+            continue
+
+        if issubclass(handler, BasicCSVStatsHandler):
+            filename = "train_stats.csv"
+            filenames = [pdir / filename for pdir in parallel_dirs]
+            outfile = base_dir / filename
+            maybe_rename_existing_file(outfile)
+            post_parallel_log_cleanup(filenames, outfile, cat_fn=cat_csv)
+            continue
+
```
tristanls-tbp, line 566: **issue**: First, this looks right to me. My issue is with code architecture and not the implementation. This code doesn't seem to belong here. I know it follows a pattern, but we can improve on the pattern. I'm thinking that each handler that writes files should know how to clean up after itself. So, I feel this should be something more like (
```python
for handler in logging_config["monty_handlers"]:
  handler.clean_up()
```
Although, having written the above, it should probably be more like
```python
exp.clean_up()
```
and that implementation invokes the handler clean up loop.

Notice that, for example, `DetailedJSONHandler` is fully aware of where it is writing everything...
```python
    def _save_per_episode(self, output_dir: str, global_episode_id: int, stats: dict):
        """Save detailed stats for a single episode.

        Args:
            output_dir: Directory where results are written.
            global_episode_id: Combined train+eval episode id used for DETAILED logs.
            stats: Dictionary containing episode stats keyed by global episode id.
        """
        episodes_dir = Path(output_dir) / "detailed_run_stats"
        episodes_dir.mkdir(exist_ok=True, parents=True)

        episode_file = episodes_dir / f"episode_{global_episode_id:06d}.json"
        maybe_rename_existing_file(episode_file)

        with episode_file.open("w") as f:
            json.dump(
                {global_episode_id: stats[global_episode_id]},
                f,
                cls=BufferEncoder,
            )

        logger.debug(
            "Saved detailed JSON for episode %s to %s",
            global_episode_id,
            str(episode_file),
        )
```
So, since logging handlers have all the information to write the files, they have all the information to clean up the files. Therefore, they should also have a method that cleans up and `run_parallel.py` should invoke it.

**note**: I am not intimately familiar with the nuance, so maybe there is a gotcha that I'm not thinking of that prevents the approach I outlined above. I am mostly starting from the point that this code seems to attempt to undo details it shouldn't know about (inside logging handlers).

ramyamounir: Maybe I'm missing something here, but I think the cleanup operation is at the same level as the multiprocess spawning operations. The word cleanup is misleading because it sounds like we would be just deleting files, but we're actually consolidating log files from all experiments logs. This cleanup operates on the log directories of all spawned experiments at the same time, not a single experiment. For example, it will [find the logging directories for all experiments](https://github.com/thousandbrainsproject/tbp.monty/blob/5c49bf17192f0f6c6bbe94dce4696e01bc502c65/src/tbp/monty/frameworks/run_parallel.py?plain=1#L529-L531) and concatenate all the `detailed_run_stats.json` files [from each experiment into one global file](https://github.com/thousandbrainsproject/tbp.monty/blob/5c49bf17192f0f6c6bbe94dce4696e01bc502c65/src/tbp/monty/frameworks/run_parallel.py?plain=1#L554). The only thing we might be able to do from within the experiment itself is to [delete the log file](https://github.com/thousandbrainsproject/tbp.monty/blob/5c49bf17192f0f6c6bbe94dce4696e01bc502c65/src/tbp/monty/frameworks/run_parallel.py?plain=1#L106-L107) at the end of the cleanup. None of the handlers know where the other handlers have logged their data to carry out this cleanup, otherwise there would be a leak of abstraction.

src/tbp/monty/frameworks/utils/logging_utils.py
```diff
@@ -663,9 +663,13 @@ def get_graph_lm_episode_stats(lm):
...
-        "detected_location": lm.detected_pose[:3],
-        "detected_rotation": lm.detected_pose[3:6],
-        "detected_scale": lm.detected_pose[6],
+        "detected_location": lm.detected_pose[:3]
+        if lm.detected_pose is not None
+        else None,
+        "detected_rotation": lm.detected_pose[3:6]
+        if lm.detected_pose is not None
+        else None,
+        "detected_scale": lm.detected_pose[6] if lm.detected_pose is not None else None,
```
nielsleadholm: Doesn't the end of this line (`if lm.detected_pose is not None else None,`) duplicate the code above?

ramyamounir: Yes, we're splitting `lm.detected_pose` in to `detected_location`, `detection_rotation` and `detected_scale` based on the slice where this data exists. This only works if `lm.detected_pose != None`, so I have to do this check for each of the data we are trying to extract. Maybe I missed what you mean by "duplicate the code above", let me know if this doesn't answer your question.

nielsleadholm: Ah sorry I was reading the ordering backwards, got confused by the multi-line syntax. Never mind.

src/tbp/monty/frameworks/run_parallel.py
jeremyshoemaker: I think you might be talking past each other. The problem that I think @tristanls-tbp is trying to point out is that the code as written knows far too much about the individual handlers.

Anytime I see things like `isinstance` or `issubclass`, that tells me that something should be using polymorphism and isn't. It usually indicates that a class is doing work that the class it's talking to should be doing instead. As the caller, it shouldn't care what type it has, as long as it matches some protocol. This is what @tristanls-tbp was pointing to when he said this should be something like `handler.clean_up()`. Whether that's the correct name, or what the details of the arguments it needs are to be determined.

@ramyamounir it sounds to me like you're saying that this code needs some context that the handlers individually don't have, which to me means that that additional context needs to be provided as arguments to the `handler.clean_up()` method (whatever we actually want to call it), which might not have been obvious in @tristanls-tbp's example, since he didn't write something like `handler.clean_up(...)`.

This `for` loop over the handlers is also missing a case for when we don't have either a `DetailedJSONHandler` or a `BasicCSVStatsHandler`. Maybe that's intentional, maybe a case where it's neither of these would indicate a programming error.

src/tbp/monty/frameworks/utils/logging_utils.py
jeremyshoemaker: This code is confusing because we're putting conditionals insides a dict literal. I think this would read much better if we pull out these conditionals, and use local variables to pass the result of the conditionals into the dict literal.

src/tbp/monty/frameworks/run_parallel.py
ramyamounir: Thanks @jeremyshoemaker. I now understand better what you and @tristanls-tbp are pointing out. The point I'm trying to make is not that we need to provide the logger with more context to perform the cleanup, but that this additional context doesn't belong in the logger class. Should the logger know that it is a part of a multiprocessing script? Should we provide it with information about the other processes running in parallel? My intuition is that the `run_parallel.py` script abstracts that multiprocessing information away and runs each experiment as a separate process. Every logger thinks that it is the only process running and writes to its own logging directory as if it is in a `run.py` experiment. Only the `run_parallel.py` script knows how split up the experiments into multiple configs with separate logging directories, and it knows how to handle the multiprocessing clean up. Providing multiprocessing context to the logger class seems like it would leak information about the multiprocessing abstraction that only `run_parallel.py` knows about right now.

That said, I only just learned about these concepts recently from @tristanls-tbp, so I'm probably missing something here. I'm happy to implement what you and @tristanls-tbp suggested. Although I don't know if this refactoring falls under the scope of this PR. Are you okay with doing this refactor in a follow-up PR?

src/tbp/monty/frameworks/utils/logging_utils.py
ramyamounir: Yes, good idea. I updated it in [a1eb454198b3](https://github.com/ramyamounir/tbp.monty/commit/a1eb454198b3a0213297534080b1ce2dc46b630f)

src/tbp/monty/frameworks/run_parallel.py
jeremyshoemaker: I've been meaning to take some time to look at this and see how complicated a refactor it would be. Sometimes it's easier to show a solution than try to describe it. But I haven't had the chance to yet.

As for whether it's in scope, from what I'm reading, you're adding this block of code that wasn't here before. The approach is less than ideal, and while a better solution might be more of a refactor than I realize without looking into, I do think it's in scope.

ramyamounir: Sounds good. I was just thinking it might be useful to separate the `feat` from `refactor` because I know that the `feat` currently works and it would be easier to trace a bug introduced during the `refactor`. But doing the `refactor` here is fine too, let me know if you'd like to pair up sometime to work through it together. I'll change it to a draft for now.

jeremyshoemaker: Having taken a closer look at it, I think there's a detail @tristanls-tbp and I missed. 

We don't actually have instances of the handlers here, but rather their classes from the configuration. We _could_ put the cleanup functionality into class methods on those classes, but I'm starting to wonder if that makes sense, since they aren't constructed instances, and don't have the details that an instance would have.

I think this code is going to need a refactor, but as you said, it seems like it would be larger and outside the scope of this PR.

ramyamounir: Ok thanks for the explanation.

PR #713, opened 2026-01-26
Fix spelling-, grammar- and lexical errors

TimothyAlexisVass:
As I was watching the videos on YouTube, I saw this and decided to just fix that.
Screenshot 2026-01-26 at 13 30 44

I then found some more things to fix. 😅

PR #714, opened 2026-01-26
Fix: Todo usage

TimothyAlexisVass:
⚠️ This branch is based on:
- #713

**TODOs should exist in comments, not in docstrings.**

Reasons:

1. **Docstrings are part of the public API**
   Docstrings are user-facing documentation (help(), IDE tooltips, docs generators). TODOs leak internal, temporary state.

2. **Docstrings describe *what is*, not *what is missing***
   They define behavior, contracts, and guarantees. TODOs describe incomplete work, which contradicts that purpose.

3. **Tooling expectations**
   TODO scanners, linters, and CI checks typically scan comments, not docstrings. Putting TODOs in docstrings makes them:

   * Harder to detect
   * Inconsistently enforced

4. **Documentation generators**
   Sphinx, pydoc, MkDocs will surface TODOs to end users unless explicitly filtered.

**Rule of thumb:**

* Use `# TODO:` in comments near the relevant code.
* Avoid putting TODOs in documentation.

### DIFF
--- a/src/tbp/monty/frameworks/environments/embodied_data.py
+++ b/src/tbp/monty/frameworks/environments/embodied_data.py
@@ -422,8 +422,8 @@ class InformedEnvironmentInterface(EnvironmentInterfacePerObject):
     extension of the BasePolicy). This policy can then make use of the observation
     to decide on the next action.
 
-    Also has the following, additional functionality; TODO refactor/separate these
-    out as appropriate
+    Also has the following, additional functionality that should be refactored or
+    separated as appropriate.
 
     i) this environment interface allows for early stopping by adding the set_done
     method which can for example be called when the object is recognized.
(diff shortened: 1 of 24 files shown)


jeremyshoemaker:
I'm going to close this for now, as there are several issues with it that need to be addressed before we can consider it .This PR contains a commit from your other unmerged PR (see the quote from our docs below). I also don't think we should be moving around TODO comments as it makes it more difficult to track who left them if we need to get more context. Ideally, they shouldn't be in docstrings, but we've already crossed that bridge.

> A valid Pull Request is on-topic, well-formatted, contains expected information, does not depend on an unmerged Pull Request, and does not violate the code of conduct.
> ~ from [our triage requirements](https://thousandbrainsproject.readme.io/docs/triage#2-triage)

PR #725, opened 2026-01-27
Fix spelling in monty

TimothyAlexisVass:
Split out from
- #713 


jeremyshoemaker:
This was actually correct before. The only real change might be to remove the space from "new type" since we're referring to a programming language concept here, see [the Python NewType docs](https://docs.python.org/3/library/typing.html#newtype), though even in the docs they only mash it together in the camel-cased version.

I don't think there's really a change to make here, so I'm going to close this without merging.

PR #726, opened 2026-01-27
docs: Fix spelling in models/salience

TimothyAlexisVass:
Split out from
- #713 


src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
```
jeremyshoemaker, line 1: I think this same file slipped into several of the PRs without any actual changes, and so we don't need to update the date yet.
```suggestion
# Copyright 2025 Thousand Brains Project
```

TimothyAlexisVass: You are free to remove it in the branches where you can confirm that it does not crash the actions workflow.
If you decide to do that, then I would appreciate if that could be done without me receiving another 30 emails about workflows crashing because this file was missing copyright.

👉 If the [check-style-monty](https://github.com/thousandbrainsproject/tbp.monty/actions/runs/21389947846/job/61649974213#logs) run should _not_ crash on unchanged files, then I suggest someone have a look at that.

PR #728, opened 2026-01-27
docs: Fix spelling in models/evidence_matching

TimothyAlexisVass:
Split out from
- #713 


src/tbp/monty/frameworks/models/evidence_matching/graph_memory.py
```diff
@@ -119,15 +119,15 @@ def _initialize_model_with_graph(self, graph_id, graph):
...
 
         This initializes a new GridObjectModel and calls model.build_graph.
 
         Args:
-            locations: List of x,y,z locations.
+            locations: List of x, y, z locations.
             features: List of features.
-            graph_id: name of new graph.
-            input_channel: ?
+            graph_id: Name of the new graph.
```
jeremyshoemaker: I'm not sure why the original said "name" but this should probably be "ID" similar to the comment above on line 73.
```suggestion
            graph_id: ID of the new graph.
```

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -68,7 +68,7 @@ def update_hypotheses(
         """Update hypotheses based on sensor displacement and sensed features.
 
         Args:
-            hypotheses: Hypotheses for all input channels for the graph_id
+            hypotheses: Hypotheses for all input channels in the graph
             features: Input features
             displacements: Given displacements
             graph_id: Identifier of the graph being updated
```
jeremyshoemaker: We seem to be inconsistent in our nomenclature. Maybe the previous changed comments should use "Identifier" instead of "ID". Alternatively, we could change this one to match those.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -68,7 +68,7 @@ def update_hypotheses(
         """Update hypotheses based on sensor displacement and sensed features.
 
         Args:
-            hypotheses: Hypotheses for all input channels for the graph_id
+            hypotheses: Hypotheses for all input channels in the graph
```
jeremyshoemaker, line 75: @ramyamounir Since we've been looking at this code recently, can you check that this isn't changing the meaning here? I thought it was, but after thinking about it some more, I don't think that it is, but I want to confirm.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
```diff
@@ -77,7 +77,7 @@ def update_hypotheses(
             evidence_update_threshold: Evidence update threshold
 
         Returns:
-            The list of channel hypotheses updates to be applied.
+            The list of hypothesis updates to be applied.
```
jeremyshoemaker: I'm not sure this is right, but I'm also not sure how to rephrase it to be clear.

We're updating multiple hypotheses, rather than making multiple updates to a single hypothesis.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -85,29 +85,29 @@ class EvidenceGraphLM(GraphLM):
...
             from one object to another and to generally make setting thresholds etc.
             more intuitive.
-        vote_weight: Vote evidence (between -1 and 1) in multiplied by this  value
-            when being added to the overall evidence of a hypothesis. If past and
-            current_weight add up to 1, it is use as weight in np.average to keep
-            the evidence in a fixed range.
+        vote_weight: Vote evidence (between -1 and 1) is multiplied by this value
+            when being added to the overall evidence of a hypothesis. If past_weight
+            and vote_weight add up to 1, it is used as a weight in np.average to
+            keep the evidence in a fixed range.
```
jeremyshoemaker, line 91: @ramyamounir I think this change is correct, since there isn't a `current_weight` in the constructor, but I want to be sure. This might not be related to changes we made recently, so maybe you don't have the context for it.

src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
```
jeremyshoemaker, line 1: This looks like an updated copyright notice that doesn't need to happen in this PR.
```suggestion
# Copyright 2025 Thousand Brains Project
```

TimothyAlexisVass: It might be a symptom that there is something wrong with the actions workflow.

jeremyshoemaker: I think it might be an issue with how GitHub creates a merge ref to try to run tests against what the branch would look like when it's merged. It might be throwing off our check for which files have changed. Let me try updating this branch with `main` and seeing if that corrects the issue, because this file has already been updated there, so this change _should_ just disappear from the diff.

jeremyshoemaker: Actually, I can't because there are unresolved conflicts. Let me see if there's another PR I can try it on.

jeremyshoemaker: Doing that on a different branch seemed to fix the issue, as expected, so I'll start going through these again.

TimothyAlexisVass: 🥳 Thank you

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
jeremyshoemaker: On second thought, we use ID everywhere, so let's just stick with that. We can change this comment here later if you don't feel like doing it now.

src/tbp/monty/frameworks/models/evidence_matching/graph_memory.py
```diff
@@ -119,15 +119,15 @@ def _initialize_model_with_graph(self, graph_id, graph):
...
         This initializes a new GridObjectModel and calls model.build_graph.
 
         Args:
-            locations: List of x,y,z locations.
+            locations: List of x, y, z locations.
             features: List of features.
-            graph_id: name of new graph.
-            input_channel: ?
+            graph_id: ID of the new graph.
+            input_channel: Identifier of the input channel.
```
ramyamounir, line 130: I'm not sure why `input_channel` didn't have a description here, but good catch!

src/tbp/monty/frameworks/models/evidence_matching/graph_memory.py
```diff
@@ -165,13 +165,14 @@ def _extend_graph(
         """Add new observations into an existing graph.
 
         Args:
-            locations: List of x,y,z locations.
-            features: Features observed at the locations.
+            locations: List of x, y, z locations.
+            features: Features observed at those locations.
```
ramyamounir:
```suggestion
            features: Features observed at the provided locations.
```

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
ramyamounir: I don't think it changes the meaning. `graph_id` makes a more explicit reference to the other argument `graph_id` passed to this function, but this looks fine to me too.

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
ramyamounir: I think it is correct as it was originally phrased, `ChannelHypotheses` are a group of hypotheses for a specific input channel. We are returning a list of those so we can update each input channel hypotheses independently. I'm also not sure how to rephrase it to be clearer. How about

```suggestion
            The list of `ChannelHypotheses` updates to be applied.
```

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
ramyamounir: Yeah, it should be `present_weight` instead of `vote_weight`.

src/tbp/monty/frameworks/models/evidence_matching/model.py
```diff
@@ -93,15 +93,15 @@ def _combine_votes(self, votes_per_lm):
...
+                            # apply to the receiving LM's sensor. Basically, if my
+                            # sensor is here in this pose, then your sensor should be
                             # there in that pose.
                             # NOTE: rotation votes are not being used right now.
                             transformed_lm_states_for_object = []
                             for s in lm_states_for_object:
                                 # need to make a copy because the same vote state may be
                                 # transformed in different ways depending on the
-                                # receiving LMs' pose
+                                # receiving LM's pose
```
ramyamounir:
```suggestion
                                # receiving LMs' poses
```

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
TimothyAlexisVass:
```suggestion
        vote_weight: Vote evidence (between -1 and 1) is multiplied by this value
            when being added to the overall evidence of a hypothesis. If past_weight
            and present_weight add up to 1, it is used as a weight in np.average to
            keep the evidence in a fixed range.
```

src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
TimothyAlexisVass:
```suggestion
            graph_id: ID of the graph being updated
```

ramyamounir (approved):
Thanks for the changes, LGTM.

PR #733, opened 2026-01-27
docs: Fix spelling in frameworks/environments

TimothyAlexisVass:
Split out from
- #713 


src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -226,8 +226,8 @@ def __init__(
                 and scale of objects when re-initializing.
             parent_to_child_mapping: dictionary mapping parent objects to their child
                 objects. Used for logging.
-            *args: ?
-            **kwargs: ?
+            *args: Unused?
+            **kwargs: Unused?
```
jeremyshoemaker: These aren't unused so much as they're just here to pass additional arguments to the constructor of the parent class. I'm not sure what the intent was behind the `?` here. Maybe we could change these to say something about them being passed to the `super` call?
```suggestion
            *args: passed to `super()` call
            **kwargs: passed to `super()` call
```

src/tbp/monty/frameworks/environments/embodied_data.py
```diff
@@ -744,7 +744,8 @@ def handle_successful_jump(self):
                 direction=(0, 0, 0),
             )
 
-            # TODO cleanup where this is performed, and make variable names more general
+            # TODO clean up where this is performed, and make variable names more
+            # general
```
jeremyshoemaker: Some tooling expects continuing lines of TODO comments to be indented, otherwise it considers it just a regular comment after the TODO comment. An example would be the syntax highlighting for TODO comments in PyCharm. I think it also pulls the comment text out when listing all the TODO comments in a codebase as well, but I haven't used that recently.

We haven't been consistent with it, but since you're changing this, might as well.
```suggestion
            #   general
```

src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, patch_size=10, data_path=None):
 
         Args:
             patch_size: height and width of patch in pixels, defaults to 10
-            data_path: path to the omniglot dataset. If None its set to
+            data_path: path to the omniglot dataset. If None it's set to
```
jeremyshoemaker: This might be a better way to phrase this.
```suggestion
            data_path: path to the omniglot dataset. If None, defaults to
```

src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -266,7 +266,7 @@ def __init__(self, patch_size=64, data_path=None):
 
         Args:
             patch_size: height and width of patch in pixels, defaults to 64
-            data_path: path to the image dataset. If None its set to
+            data_path: path to the image dataset. If None it's set to
```
jeremyshoemaker: Same here.
```suggestion
            data_path: path to the image dataset. If None, defaults to
```

src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -418,8 +418,8 @@ def switch_to_object(self, scene_id, scene_version_id):
     def reset(self) -> tuple[Observations, ProprioceptiveState]:
         """Reset environment and extract image patch.
 
-        TODO: clean up. Do we need this? No reset required in this env interface, maybe
-        indicate this better here.
+        TODO: clean up. Do we need this? No reset is required in this environment
+        interface, so this should be indicated more clearly.
```
jeremyshoemaker: Here's another case of indenting a TODO comment continuation line. It's the difference between this:
image

and this:
image

It works with a single space, but I usually use two.
```suggestion
          interface, so this should be indicated more clearly.
```

src/tbp/monty/frameworks/environments/two_d_data.py
```diff
@@ -690,7 +690,7 @@ def __init__(self, patch_size=64, data_path=None):
 
         Args:
             patch_size: height and width of patch in pixels, defaults to 64
-            data_path: path to the image dataset. If None its set to
+            data_path: path to the image dataset. If None it's set to
```
jeremyshoemaker:
```suggestion
            data_path: path to the image dataset. If None, defaults to
```

PR #735, opened 2026-01-27
docs: Fix spelling in frameworks/actions

TimothyAlexisVass:
Split out from
- #713 


src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
```
jeremyshoemaker, line 1:
```suggestion
# Copyright 2025 Thousand Brains Project
```

TimothyAlexisVass: Since this change [seemed to be required on almost every branch](https://github.com/thousandbrainsproject/tbp.monty/actions/runs/21389947846/job/61649974213), I would prefer if we could let it be.

PR #736, opened 2026-01-27
docs: Fix spelling in evidence_matching/feature_evidence

TimothyAlexisVass:
Split out from
- #713 


src/tbp/monty/frameworks/models/mixins/no_reset_evidence.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
```
jeremyshoemaker, line 1:
```suggestion
# Copyright 2025 Thousand Brains Project
```

TimothyAlexisVass: Yes, it is changed in every branch because my inbox got [flooded with failed action runs](https://github.com/thousandbrainsproject/tbp.monty/actions/runs/21389947846/job/61649974213) all failing on the same thing:
image
I was equally surprised, since the file was only changed in one of the branches, but just went ahead and changed it in each of the branches since the majority of them failed on this specific thing.
I suggest that we let it be on the minority of the branches where it is not necessary.

jeremyshoemaker: I just merged `main` into this branch, and it looks like this change has disappeared from the diff.

jeremyshoemaker (approved):

PR #741, opened 2026-01-27
ci: add help message to copyright date check

jeremyshoemaker:
The current output of the copyright date check isn't very helpful, as Ruff just says the copyright notice is missing. The title of the workflow step gives some indication of the problem, but not the solution. This adds code to print out a message linking to the documentation on failure to help guide people to the right solution.


jeremyshoemaker:
Here's how it looks when it fails.
image

PR #743, opened 2026-01-28
ci: Only check changes from PR branch

jeremyshoemaker:
Recently we had our copyright date check fail on files in PRs that shouldn't have needed updating. From what I can tell, this happened because the files were changed in `main`, which included an update to the copyright date, and in the PR branch they weren't changed, and so when we did the diff, it showed up as changed, and then the copyright check failed because the version in the branch didn't have the current year in it.

I think what we should do is change how we're generating the list of changes files to **only** include files that were changed in the branch, but not files that **only** changed in the `main` branch. This is similar to how GitHub and other Git forges render their diffs.

This can be done using `git merge-base` or in the case of recent versions of Git, using the `git diff target...source` syntax. In the code change below, the `source` is omitted because it defaults to HEAD.

See [the `git diff` documentation on this syntax](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-gitdiffoptionscommitcommit--path-1) for details, though the terminology can make it difficult to understand.

### DIFF
--- a/src/tbp/monty/frameworks/actions/action_samplers.py
+++ b/src/tbp/monty/frameworks/actions/action_samplers.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
 # Copyright 2024 Numenta Inc.
 #
 # Copyright may exist in Contributors' modifications
@@ -81,7 +81,7 @@ class ConstantSampler(ActionSampler):
     that take a `rotation_degrees` parameter will have it set to `5.0`.
 
     When sampling an Action, only applicable parameters are used. For example,
-    when sampling a `MoveForward` action, only the ConstantCreator's
+    when sampling a `MoveForward` action, only the ConstantSampler's
     `translation_distance` parameter is used to determine the action's `distance`
     parameter.
     """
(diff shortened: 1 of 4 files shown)


jeremyshoemaker:
Here's an example of the difference using #728 as an example. It might be difficult to reproduce locally, because I have my local copy set up to pull GitHub's `pull` refs. Also, it looks like the author merged `main` into it, and I'm working on the version from before when he did that.

**Note:** I'm using the `pull/728/merge` ref because that's what the `checkout` action uses.

```sh
 $ git checkout upstream/pull/728/merge
HEAD is now at 1091c916 Merge 71aefb35291721dbfbd7861c4d4b8ae341a895a3 into 452131867f91e81f18f9d63ce03a4c16ee06d38c
 $ git diff upstream/main --name-only | wc -l
106
 $ git diff upstream/main... --name-only | wc -l
6
 $ git diff upstream/main... --name-only
src/tbp/monty/frameworks/models/evidence_matching/graph_memory.py
src/tbp/monty/frameworks/models/evidence_matching/hypotheses_displacer.py
src/tbp/monty/frameworks/models/evidence_matching/hypotheses_updater.py
src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
src/tbp/monty/frameworks/models/evidence_matching/model.py
src/tbp/monty/frameworks/models/evidence_matching/resampling_hypotheses_updater.py
```

With the way we're currently doing it, we get 106 differences, and with this change, we only see 6 files changed, the ones that are visible in the diff in GitHub.

PR #745, opened 2026-01-29
0000_RL_models

h555yand:
RFC for initial improvement related to the problem explained [here](https://thousandbrainsproject.readme.io/docs/learn-policy-using-rl)


nielsleadholm:
Hi @h555yand Aleksei, thanks very much for opening this RFC. Just to keep you updated, I'll be the primary point of contact for this RFC and associated work. My aim is to look at the RFC and provide some feedback in the coming days.

h555yand:
Hi @nielsleadholm Niels, nice to meet you. Looking forward to communicate)

nielsleadholm:
Thank you again for sharing this Aleksei. Generally speaking there is some interesting scope, but the RFC could be clearer. A couple high level comments to start things off:

1. It's not entirely clear to me what the RFC is proposing. It begins with an abstract description of RL, with some occasional references to Monty. Is the proposal to create a new RL module? If so, it would be helpful to make this more explicit, and elaborate on this point. Some related points that would be worthwhile exploring would include:

	a. How does this tie in with known neuroanatomy? The basal ganglia appear important for reward learning in the brain; is the proposal to implement an analogous structure?
	
	b. What are the consequences of having the reward module external from learning modules? How can it make use of the models they learn if it is a separate structure?
	
	c. What is a concrete task that you could design to test this setup, given Monty's current capabilities? Bear in mind that Monty's actions are currently limited to improving inference by moving sensors; it does not currently have a mechanism to change the state of the world by e.g. pushing or lifting objects.
	
2. There are several points where the language or details differ significantly from what we use in Monty. For example, you've described "Models are stored as graphs, where the vertices are states and the edges are transition actions.". However, there are generally speaking no edges in the form of transition actions within Monty. Similarly, the definition of State ("images, sounds, smells, kinesthetics, etc. The scene code is formed as a concatenation of the codes of the sensory components included in the scene.") is very different from how the `State` class is used in Monty, where it forms the packets of the CMP. It would be helpful to read up on how these terms are defined and used in Monty, and adjust the language in the RFC appropriately. 

If you haven't already seen them, I would recommend the [How Monty Works documentation](https://thousandbrainsproject.readme.io/docs/implementation-overview), followed by the [Methods section](https://arxiv.org/abs/2507.04494) in one of our recent papers, as useful resources.

Thanks again for starting this RFC; I know it's a lot of information to digest before contributing, so we're also here to help (either in this thread, or on [Discourse](https://thousandbrains.discourse.group/latest)) if you have any questions that I can clarify.

h555yand:
Hi @nielsleadholm Niels, 
many thanks for your feedback, detail comments, documentation. It’s very useful.
I agree that the current proposal doesn't look like a typical pull request with specific proposed changes. It's more of a description of the approach and a reason to start a discussion about how and where it can best be applied in Monty.
I've answered the questions, as well as additional comments and potential further developments.
I propose an iterative discussion and refinement of the possible detailed tasks.


I'm not an expert in neuroanatomy, so I'll try to answer based on my limited knowledge :)
Yes, this is a proposal to create a similar to brain structure for action selection and planning.
    • The prefrontal cortex (PFC) generates multiple candidates.
    • The hippocampus "plays out" outcomes.
    • The orbitofrontal cortex (OFC) and ventromedial PFC (vmPFC) transform information about action options into "subjective value".
    • The basal ganglia (BG) decide which plan to proceed with, when to slow down/stop, and how to adapt future choices based on rewards and errors.
    • Dopamine (VTA) changes the weights in corticostriatal synapses so that the desired response is strengthened in the appropriate context.
I've marked which parts are executed in the pseudocode.


As I understand, the current LM focuses on object recognition and learns an internal model of the world (locations / poses, properties).
Current model-free policy for following areas of prominent curvature looks like a predefined heuristic function.
Model-based hypothesis-testing policy is a fixed action selection procedure that uses this model on the fly.
There is no policy learning in terms of RL.

If the goal is to effectively "cover the surface" and the heuristics already do this well, the RL value function may be excessive.
For a model-free system, RL may be useful if:
⦁ Adaptation: for different objects / sensors / friction, the value will adjust priorities better than rigid rules. Currently, finger movements are not performed as an real action with motor skills.
⦁ Active sensing: choose steps that maximize information gain / uncertainty reduction or increase classifier confidence.
⦁ Limited resources: optimize "information per second/steps."
⦁ Safety / reliability: penalize the risk of contact failure, excessive pressure, and jamming.
Possible model-free extension can be:
⦁ Keep the current heuristic as one of the primitives. Add 1–2 alternative movement primitives. Possible actions: "follow the principal curvature k1+ / k1−", "spiral / along the generatrix", "move to the adjacent patch", "return to the ROI". 
⦁ Choose reward as:
- Δ recognition confidence or Δ difference in the target class logits,
- Δ reduction in variance / entropy in the local surface model (ensemble, dropout uncertainty),
- Δ coverage of the unique area,
- cost: time / energy / dropout risk penalty.

⦁ Train policy model Q(s, a) to select a primitive based on curvature, normal, texture with a reward = Δ object model confidence − cost.

Possible model- based extension can be:
⦁ State: belief - discrete hypotheses about {class/object ID, pose} with weights
⦁ Actions: A discrete set of heuristics as example.

- Heuristic 1: DisambiguateTop2 — a point from the Top1 model that is as far away as possible from its nearest neighbor in the Top2 model
- Heuristic 2: LookAtCentroidTop1 — look at the center of mass of Top1
- Heuristic 3: RandomProbeAroundTop1 — a random point near the surface of Top1

⦁ Updating beliefs in a step with Bayesian filter
⦁ Reward: r = H(b_t) − E[H(b_{t+1})] − λ*cost,
It estimates through sensor simulation for each hypothesis.
Cost of the step: time, energy, collision risk, etc.
⦁ Train policy model Q(s, a) and then during inference choose actions based on past experience.

This can be done by extending the current LM module and using graphs with edges as actions, and its own state. Monty's state can be part of the state. CMP can be also extended.
I think this is a natural extension, since the brain uses different parts and connections for different tasks.

The further extantion could be to develop into real-world actions and interactions with the outside world. For example, I saw a task about setting up a table, where spoons and forks need to be moved.

Current approaches lack specificity. I suggest deciding whether this makes sense and is valuable for the Monty. If so, I can develop the approach in more detail, taking into account the specifics of the Monty. Perhaps within the framework of a separate RPC.

What do you think?

h555yand:
Hi @nielsleadholm Niels,
I hope you're doing well.
Any updates on my proposals?

Best regards, Aleksei

nielsleadholm:
Hi @h555yand , apologies for the delay in getting back to you.

The update you sent was a really nice overview, and I think does a good job framing the scope of what you could work on. If I was to suggest a first approach, it would consist of the following:

### Develop a model-free RL system to enact the goal state sent from the LMs and received by the motor system.
- Given a target goal state and the current proprioceptive state of the agent, the RL system could propose incremental actions to get it into the target state.
- This would be model-free (i.e. no learned explicit model), but would still make use of the current sensory observations. Sensory observations would consist of the sensory patch input, as well as proprioceptive information.
- These goal-states come from the hypothesis-testing policy, which is implemented in the Evidence LM's Goal-State Generator. Currently, these goals are enacted by the JumpToGoalState mixin, but your model-free RL agent would be a much better solution.
- As a first version, you could keep any collision in Habitat disabled, such that the agent can theoretically move to the target location by clipping through the object, if necessary. This would be a simple "sanity check" that the model-free agent learns to move towards the target state based on its current proprioceptive data, even if it's not a realistic/long-term solution.
- When implementing the above, you could formulate a dense reward signal by making the reward inversely proportional to the distance between the agent's current pose, and the target pose.
- As a more advanced, follow-up implementation, you could activate collision detection. In that case, the agent actually needs to learn how to move around the object if it is in the way.
- Implementation-wise, this means the new RL system would be entirely external to the LMs (e.g., it could be a component of the motor system). This has the advantage that we can leave LMs in their current format alone. You also wouldn't need to modify the CMP, because the reward could be calculated locally in the motor system's RL system.

Overall this would be a well-scoped problem that brings a tangible benefit to Monty. Anyone in the future who wants something like a robotic arm that moves to a target goal-state could use something built on your implementation. How does that sound?

h555yand:
Hi @nielsleadholm Niels, many thanks for your feedback. 
I like the approach you proposed.
What are the next steps should be to move forward?
As I understand It needs to prepare more concrete detailed RFC.
Should I create a separate document or fill details within this one?
Any guidelines from your side would be appreciated.

Best regards, Aleksei

nielsleadholm:
Ok great, yes the next best step would be to lay out a concrete RFC proposal. Feel free to update the one you have started here. You may also wish to update the title to make it a bit more specific.

h555yand (commented):
Hi Niels @nielsleadholm, I hope you're doing well.

I've prepared updated RFC based on our discussion. Looking forward to your feedback and happy to answer on questions.

Best regards, Aleksei

nielsleadholm:
Exciting, thanks for the update Aleksei. Just a heads up - we have some paper submission deadlines this week, and an in-person meet-up next week, so I probably won't get round to reviewing the update until the week of April 20th. However, I'm looking forward to diving into the details. All the best until then.

nielsleadholm:
Hi Aleksei,
Thank you for the update. In general this looks like a step in the right direction, although I found it challenging to understand some of the proposals and how they relate to the existing implementation of Monty.

Can I ask if this RFC was written with the use of LLMs? [One thing to highlight is our policy on the use of LLMs in RFC and code contributions](https://thousandbrainsproject.readme.io/docs/large-language-model-use). If you would like to use LLMs in your work, please note that we reserve the right to do more limited reviews of submissions. For example, if you want to take this RFC as is and attempt to implement it with e.g. Claude Code, then it will be harder for me to help you with getting it working if you run into issues. I would still be able to review the submission if you manage to get it working, but until you reach that stage, my input will be more limited.

On the other hand, if you implement it yourself, including writing the RFC and code in your own words, then I would be happy to spend more time working with you on this project, including meeting on some video calls to talk through the details. I hope it makes sense why we need to take this approach.

With that context provided, below are some specific comments:

- In terms of the algorithmic approach taken, it would actually be fine to use deep-learning for this task. As this concerns the "sub-cortical" structures of the brain, one can imagine that a fair amount of this learning has occurred on evolutionary time scales. Deep learning is a reasonable model-free method for approximating this. It would also increase the chance that your work could be reapplied in a continuous-control settings such as a robotic limb.
- Please note that the motor system is in the final stages of a significant refactor. You can read about some of these changes here: https://thousandbrains.discourse.group/t/new-motor-policy-selection-framework/ but the main take-away is to please pull from upstream frequently.
- The action spaces for the surface agent and distant agent are different (MoveForward, MoveTangentially, OrientHorizontal, OrientVertical vs. LookUp, LookDown, TurnLeft, TurnRight). To simplify things, I would suggest beginning by focusing on the surface agent, and not considering the distant agent at all.
- The proposed auxiliary rewards look good, thanks for adding these.
- It's not clear to me that we should blend "heuristics" with random actions as a way of boosting learning. The surface policy heuristic will not generally move towards the goal, while it should be quite straightforward for the policy to start learning reasonable proposals given the dense learning signal.

h555yand:
Hi Niels, many thanks for your feedback and comments. 
I agree that it’s not easy to understand t when there are many different proposals without enough clarifications. Let me try to provide some answers. <br>
Ideas, thesises, architecture, approaches are mine. I used LLM for following tasks:
- Clarify some concrete questions
-  Summarize and translate some my sentences from my native language to English
- Help me to analize Monty source code and answer on my questions
- Help me to write code to fast proof concepts and ideas, as example  trimesh environment to test Q learning progress with HNSW <br>
I’m not a fan of vibecoding and don’t use it in my work. I prefer questions-answers and concrete tasks. It improves my performance without loosing quality.
I believe we can find approach to collaborate without risks of copyright restrictions. It’s not a problem for me to describe details using my “simple” English and not to use LLM generated code in public.  <br>
I will answer on your specific comments later as I’m busy coming next days. I have some points regarding to deep learning, surface agent and distant agent.
Best regards, Aleksei

nielsleadholm:
Ok great, thanks for clarifying Aleksei. Overall that sounds really positive - obviously we're not totally against using LLM agents. They are clearly a powerful tool, but it can quickly result in a quagmire of verbose output rather than focused clarity. 

Per my previous comments, I think a deep reinforcement learning approach could actually be appropriate here. If you still want to go for the HNSW approach, I would appreciate if you can elaborate on this, as it's not an algorithm I'm familiar with. Feel free to link to some external resources that you've found useful, if there are any. 

Looking forward to hearing more soon.

h555yand:
Hi Niels, I hope you are doing well.
I've started to answer on your questions and realized that it would be better to update RFC itself.
I've concreted Q-learning (HNSW + kNN + Gaussian Kernel Interpolation) and deep learning (SAC) collaboration, distant and surface agents action spaces, heuristics.
In addition I've removed some verbose details that it's early to discuss on current stage or to not bring much value.
Hope this helps to understand main idea. Looking forward to your feedback and happy to provide details.
Best regards, Aleksei

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+This aligns with Monty's broader goal of biologically plausible computation.
+
+### Theoretical publications
+- Ormoneit & Sen's (2002) 'Kernel-Based Reinforcement Learning' proposed and analyzed a variant of Q-learning for large state spaces, in which the Q-function is approximated by kernel regression on the data rather than by a neural network or table. A key contribution is the conditions under which such a 'kernel Q-learning' scheme converges.
+- Blundell et al. (2016) 'Model-Free Episodic Control'. The agent remembers its past successful actions and returns and, when faced with a similar state, simply reproduces the best of what has already worked.
+
+### HNSW + kNN + Gaussian Kernel Interpolation
+
+Hierarchical Navigable Small World (HNSW) is an approximate nearest neighbor search algorithm based on a layered graph data structure. It belongs to the family of proximity graphs, where nodes (vertices) are connected based on their proximity, typically measured by the Euclidean distance.  
+HNSW is currently actively used in embedding databases for searching similar text by vectors. So, I decided to use it to store and find states.
```
vkakerbeck: What are you referring to as "states" here? Are those the target poses provided by the LM? And so if a similar target is provided, the system looks up what action sequence was performed before for this target?

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+- Blundell et al. (2016) 'Model-Free Episodic Control'. The agent remembers its past successful actions and returns and, when faced with a similar state, simply reproduces the best of what has already worked.
+
+### HNSW + kNN + Gaussian Kernel Interpolation
+
+Hierarchical Navigable Small World (HNSW) is an approximate nearest neighbor search algorithm based on a layered graph data structure. It belongs to the family of proximity graphs, where nodes (vertices) are connected based on their proximity, typically measured by the Euclidean distance.  
+HNSW is currently actively used in embedding databases for searching similar text by vectors. So, I decided to use it to store and find states.  
+During the learning process, the agent stores experience in a graph and then uses the weighted past experience in a similar situation. Thehnically it looks like: store point in HNSW graph, then find the K closest ones and mix them with Gaussian kernel weights.
```
vkakerbeck, line 56: How would the system start to get good data points into its memory? If a target pose was not observed before, it sounds like it could not retrieve an action sequence and so would have to act randomly. This random action sequence will most likely not end up at the target pose so it wouldn't be useful to retrieve in the future. Or would it be more represented as an after-the-fact-type memory (i.e. "If I had wanted to end up here, this would be the action sequence to take (even though this was not my goal right now)")?

rfcs/0000_RL_model_free_for_motor_system.md
vkakerbeck: Also, would the states (if they are target poses) be relative to the starting state? So for instance, the agent might need to navigate from any where in the world to anywhere else in the world. Going from (0,0) to (5,5) requires the same action sequence as going from (5,5) to (10, 10) so to make this generalize to any possible displacement, it seems to make sense to represent the target pose as being relative to the current pose instead of an absolute location in the world. Is that what you were thinking?

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Why not only Deep Learing
+I'm not opposed to deep learning. I agree that it's well suited for approximation, embedding, and many other tasks.  
+I'm not suggesting replacing neural networks, I'm suggesting supplementing it and improving the learning process.
```
vkakerbeck, line 60: I like this approach. Another way that deep learning could be supplemented in this case could be to use an inverse kinematics approach or imitation learning. Inverse kinematics would be more specific to the specific robot body used and imitation learning would require some supervised examples. But in case the memory retrieval doesn't work well enough in the very beginning, they could be interesting options to consider.

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+**RLGoalApproachController** computes **State Vector** using **goal_pose** as well as sensory patch input and proprioceptive information.   
+   ↓  
+**At the begining for new states** it needs to learn before inference  
+   ↓  
+**RL Q-leraning with HNSW + kNN + Gaussian Kernel Interpolation** working with discrete actions. Action Selection with **Heuristic-Guided Exploration**. Objective: To obtain smart behavior and training data for SAC. 
+HNSW graph points collection (gathering experience). Copying of **successful traces into replay buffer**.
+**HNSWStateStore** stores **State Vector**, actions, Q-values.
+Upon reaching a certain threshold of successful validation operations moves to next step of learning.    
+   ↓  
+**Behavioral cloning (BC)** — method of imitation learning in which an agent learns to imitate the behavior of an expert by directly copying his actions based on data.
```
vkakerbeck: Would this be part of the HNSW approach or another supplement before SAC?

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+   ↓  
+**At the begining for new states** it needs to learn before inference  
+   ↓  
+**RL Q-leraning with HNSW + kNN + Gaussian Kernel Interpolation** working with discrete actions. Action Selection with **Heuristic-Guided Exploration**. Objective: To obtain smart behavior and training data for SAC. 
+HNSW graph points collection (gathering experience). Copying of **successful traces into replay buffer**.
+**HNSWStateStore** stores **State Vector**, actions, Q-values.
+Upon reaching a certain threshold of successful validation operations moves to next step of learning.    
+   ↓  
+**Behavioral cloning (BC)** — method of imitation learning in which an agent learns to imitate the behavior of an expert by directly copying his actions based on data. 
+Translation of discrete actions from replay buffer into continuous ones.
```
vkakerbeck: What would this be for? At what point would this happen?

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+## Advantages of this scheme:
+- Quick start with Q-learning and discrete actions – no need to wait for the SAC to learn from scratch.
+- Stability – Heuristic-Guided Exploration learns in new areas.
+- Precision – then SAC makes movements smooth and optimal as skills.
+- Biologically plausible – like human learning: first we copy, then we hone.
+
+
+**Below is explanation of the main components**:
+
+## State Vector (13D)
```
vkakerbeck, line 100: This is very interesting. I originally thought of just proving a 6D vector (the displacement between the current sensor location and orientation and the target one). Are you thinking of including all the other information so that the policy can learn some heuristics for how to navigate around an object? This is a nice idea I think, especially since we can provide updated sensory information at every step on the way to the target location so the policy can always know how close/far it is currently from the object surface and how the surface in front of it is shaped.

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+Update state → normalize → KNN search
+→ if near existing point: update it
+→ else: insert new point with interpolated init
+
+Get state → normalize → KNN search → kernel interpolation → Q-values
+
+
+## MontyActionSpace (18D)
+
+To properly train hand, finger movement, it is suggested to use the combined agent mode (distant + surface) from the beginning.
```
vkakerbeck: Why would you combine the two? I think it would be better to provide an action space to the learning algorithm, depending on what agent type is available. So, the algorithm itself should be agnostic to what action space is available. When we are working with the surface agent, we would only provide it's action space to the policy learning algorithm. When working with the distant agent we would provide a different action space. And similarly we could provide other action spaces for other agents, such as a robotic arm that has specific joints that can be controlled.

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+#### Problem with Standard ε-Greedy
+
+Standard ε-greedy exploration selects random actions with probability ε. In a 18-action space, a random action has only small chance of being useful (moving toward the goal). This means the most of exploration steps are wasted, resulting in slow learning and poor initial behavior.
+
+#### My Approach: Blending Q-Values with Heuristics
+
+Instead of random exploration, blend learned Q-values with heuristic bias derived geometric reasoning:
+
+```python
+combined = (1 - ε) × Q_normalized + ε × heuristic_normalized
```
vkakerbeck, line 193: what does `heuristic_normalized` correspond to here?

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+| Near goal on surface | +0.5 | No | `distance < 3 × surface_step` AND `on_object = true` |
+| Oscillation | -0.5 | No | Current action is opposite of previous action |
+| Timeout | -10.0 | Yes | `steps >= max_steps_per_goal` |
+
+
+### Action Selection with Heuristic-Guided Exploration
+
+#### Problem with Standard ε-Greedy
+
+Standard ε-greedy exploration selects random actions with probability ε. In a 18-action space, a random action has only small chance of being useful (moving toward the goal). This means the most of exploration steps are wasted, resulting in slow learning and poor initial behavior.
```
vkakerbeck, line 186: I think with the dense rewards you describe above, the issue wouldn't be so bad (i.e. every step has an informative reward that tells the system whether it moved closer or further away from the goal or violated some condition like crossing the object surface). Maybe something to test before going with a more involved solution.

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+
+## RLMotorPolicy
+Integration point with Monty. Extends existing motor policy, activates RL when LM sends goal, falls back to standard Monty behavior otherwise.
+This means:
+All existing Monty behavior is preserved: When no goal is active, the parent class handles exploration (surface crawl, curvature-informed steps, orient to surface, etc.)
+No modifications to LMs: Goal states are read from LM attributes that already exist for JumpToGoalState
+No modifications to CMP: Reward is computed locally from proprioceptive and sensor data
+Graceful degradation: If RL fails (timeout/collision), control returns to standard Monty exploration
+
+**Inheritance**: RLMotorPolicy → SurfacePolicyCurvatureInformed → SurfacePolicy → InformedPolicy → BasePolicy → MotorPolicy
```
vkakerbeck: Just as a heads up, since recent refactors, there isn't such a deep hierarchical nesting anymore. Instead we have independent policies and a policy selector that can switch between them. We posted a bit about this here: https://thousandbrains.discourse.group/t/new-motor-policy-selection-framework/1048 But plan to document more details in the next weeks.

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+> This section is intended to encourage you as an author to think about the lessons from other frameworks and provide readers of your RFC with a fuller picture.
+> If there is no prior art, that is fine. Your ideas are interesting to us, whether they are brand new or adaptations from other places.
+>
+> Note that while precedent set by other frameworks is some motivation, it does not on its own motivate an RFC.
+> Please consider that Monty sometimes intentionally diverges from common approaches.
+
+# Unresolved questions
+
+## Open Questions
+### Q1: Optimal State Dimensionality
```
vkakerbeck, line 290: This is a good question and probably best determined empirically. I'm curious on what will work best :)

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -0,0 +1,289 @@
...
+   reducing dimensionality.
+
+4. **Use Embeddings**: It needs a separate embedding module but we can increase dimensions without constaints risk   
+
+### Q2: Hyperparameter and Config Sensitivity
+How sensitive is the system to sigma, k_neighbors, and learning rate?
+What are optimal parameters for reward weights, action steps, etc?
+
+Current position: To be determined empirically during many tests.
+Mitigation: It makes sense to develop Lightweight Enviroment with trimesh for mesh-based primitives simulation to standalone RL training without Habitat.
```
vkakerbeck, line 313: I agree, that sounds like a great approach

.gitignore
vkakerbeck, line 1: This file shouldn't be part of the RFC PR

vkakerbeck (commented):
Hi @h555yand 

I am taking over from Niels as he is currently on leave (and I also find this topic super interesting!)

Thank you for the nice updates to the RFC. It sounds like a very interesting proposal and contains many good ideas. I left several questions and suggestions in there. On a high-level my main comment would be to try and make the solution one that is action space independent. So something where the user can provide an arbitrary action space and the system learns to achieve the goals given that action space. Let me know if that doesn't make sense.

I think a lot of the remaining questions are questions that will need to be tested empirically. Do you have an idea for a good, simple testbed for this (as you allude in Q2)? 

In terms of next steps, we can discuss more on this RFC here until you have a clear idea of how you would implement and test these approaches. Once you feel like you are ready to dive into testing these ideas, I would recommend starting a prototype repository using a fork of tbp.monty. We have a bit of guidance on this here: https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/0018_code_guidance_for_researchers_and_community.md#workflow-for-community-members-conducting-research. This will be a good way to test out all the ideas you have and the different possible options. Once an optimal solution is found, we can look into how to best integrate it into the main tbp.monty repository.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: As I understand these questions were clarified below in the State article. I've added the link in RFC.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: Heuristic-Guided Exploration approach can help here.  For cold start nearly pure heuristic actions with some random ones are used for new states. Fixed amount of steps per episode (config parameter) is used to end up. Action sequence is saved in HNSW with their Q values.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: Agree

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: BC (Behavior Cloning) is the simplest form of imitation learning: we teach the robot's policy as a supervised learning task—to predict the action of an expert based on observation.
It can be used as independent approach but usually used as supplement before SAC or other off-policy RL algorithm.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: I've added details in RFC.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: Agree it's a good idea. What parameters and optimal dimension we will clarify during testing.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: It seems that we are talking about the same but in different words:)
I've updated RFC section with details.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: Agree, It needs to test. Anyway I think that mix of heuristic and random actions can improve learning.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: It needs to properly compare and choose action. Q-value and heuristic biases have their own ranges and we transfer them to [-1, 1] range.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: Thanks, I will keep it in mind.

h555yand (commented):
Hi @vkakerbeck Viviane,
nice to meet you and many thanks for the detailed feedback:)
I've provided information in direct answers or in updated RFC sections. 
Happy to answer on additional questions if any.

I agree with you that the best next step could be to test out all the ideas.
I've added in RFC the special section for Lightweight environment. So technically I'm ready to start prototyping in a separate fork.
In parallel we can continue to discuss open questions.
Looking forward to you comments.
Best regards, Aleksei

rfcs/0000_RL_model_free_for_motor_system.md
vkakerbeck: If we have good heuristics (like inverse kinematics) to generate those initial action sequences, what is the additional need for storing and retrieving them? Is it because we can retrieve similar ones? It sounds like it would be harder to make the same action sequence work in similar situations than to just use heuristics in all cases at the beginning. But you probably have more experience with this.

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -119,10 +120,31 @@ Update state → normalize → KNN search
...
 
 
-## MontyActionSpace (18D)
+## MontyActionSpace
+
+To properly train hand, finger movement, it is suggested to use the combined action spaces (distant + surface) from the beginning.  When the finger is on the surface it plays like surface agent, when is on the air like distant. For me current split for distant and surface for finger looks a little bit artificial in RL.
+Proposed action space is universal high level primitives for moving finger to the goal.
+These primitives can be used for any other similar agents and tasks and they are not dependent from specific low level robotic actions (joints, speed, strength etc). 
+Transfer primitives to low level robotic is a separate and well known task.
+In the industry standard is a mix: the Neural Network chooses 'What to do' (high level primitives), and the classical algorithm (Inverse kinematics & Impedance) decides 'How exactly to move the motors'.
```
vkakerbeck, line 129: In this case, I am not sure what the purpose of using RL here is. If the specific robots actions are assumed to be calculated using inverse kinematics and the LM already provides the target location, why do we need the intermediate step of medium-abstract actions and RL?

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -119,10 +120,31 @@ Update state → normalize → KNN search
 Get state → normalize → KNN search → kernel interpolation → Q-values
 
 
-## MontyActionSpace (18D)
+## MontyActionSpace
+
+To properly train hand, finger movement, it is suggested to use the combined action spaces (distant + surface) from the beginning.  When the finger is on the surface it plays like surface agent, when is on the air like distant. For me current split for distant and surface for finger looks a little bit artificial in RL.
```
vkakerbeck, line 125: The idea behind the surface and distance agent is that the distant agent is more like a camera that is mounted in one location (i.e. it can't move freely, it can only tilt up, down, left and right). We write a bit more about this here: https://thousandbrainsproject.readme.io/docs/policy#action-spaces
You could consider an agent that can do both, although it doesn't seem nescessary to have the tilting actions of the distant agent if you already have the relatively free movement of the surface agent. I guess my main point is that there are different agents that have different capabilities and constrains and so the solution should be general enough to cover all those cases.

rfcs/0000_RL_model_free_for_motor_system.md
vkakerbeck: I understand the normalization part, I was just wondering what the heuristic referred to here is.

rfcs/0000_RL_model_free_for_motor_system.md
```diff
@@ -209,14 +221,37 @@ No modifications to LMs: Goal states are read from LM attributes that already ex
 No modifications to CMP: Reward is computed locally from proprioceptive and sensor data
 Graceful degradation: If RL fails (timeout/collision), control returns to standard Monty exploration
 
-**Inheritance**: RLMotorPolicy → SurfacePolicyCurvatureInformed → SurfacePolicy → InformedPolicy → BasePolicy → MotorPolicy
+## Lightweight Enviroment
+To fast test hypophesys and ideas we need to create relevant approximation of Habitat, especially for training policies based on haptics/active perception.
+It should not simulate graphics, but it should accurately reproduces the key physics that are important for training: contact, normals, ray casting, movement on surfaces.
```
vkakerbeck, line 226: Why wouldn't you just use habitat and the observations returned from it? It sounds much more complicated to implement something similar that is still guaranteed to return the correct physical properties than to just keep using this library.

vkakerbeck (commented):
Hi Aleksei,

thanks for the answers and updates to the RFC. I left a few follow-up questions and comments. Mainly, I am still unsure why an 18D action space is necessary (as opposed to allowing any action space, testing both surface and distant agent action space, and the algorithm learning in either case). I am also unsure why these tests couldn't just be run in Habitat, as it sounds unnescessarily complicated to replicate the physics, observations, and actions outside of a simulator. But I am sure you have good reasons for it, I am just curious.

I think it is a good idea to start testing this on a prototype fork, which can also make the discussions here more concrete and will maybe resolve some of my open questions already :)

Best wishes,
Viviane

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: I suggest to use heuristics at the very beginning with Heuristic-Guided Exploration.
Yes, storing and retrieving of action sequences is part of Episodic memory approach.
I will answer about using of inverse kinematics in the below article.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: Let's synchronize our understanding of inverse kinematics (IK). It helps to clarify this question and other ones.
My perspective:
Inverse kinematics is calculating the joint angles required to reach a specific position (and orientation) of a target point (e.g., a hand) in space.
Classical inverse kinematics itself doesn't calculate motion (trajectory). It solves a static problem: "Here's point A. Here's point B. Give me the joint configuration to get to point B."
A standard IK solver is "blind"—it knows nothing about the environment. It sees only the target and finds the shortest path through the space of joint configurations, which may lead to a collision.
To account for obstacles, we need to go beyond pure IK and use more sophisticated approaches. RL is one the approaches to build trajectories without collisions.
There are two options of RL training:
- Use low level robot action space (specifying angles for each of its joints and other parameters)
- Use high level intuitive action space like in the current proposal  

As a result: The RL agent doesn't need to learn complex movement mechanics. It learns the strategy ("first raise the hand, then move it forward"), and IK takes care of the tactics (how exactly to bend the joints).
It allows us to collect experience as intuitive actions in Episodic memory.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: Perhaps I'm misunderstanding how actions work.
As I understand it, the surface agent's MoveTangentially, OrientHorizontal, and OrientVertical actions are bound to the surface. Free MoveForward actions move in the direction the agent is facing and have a movement distance parameter. To move in a specific direction, we need to rotate the agent first. If this is not the case, then we can simplify 18D action space.
Anyway, we should fix the action space before starting to learn the concrete agent.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: Habitat is demanding of hardware resources and it doesn't work on my laptop)
It's easy to start with simple primitives (cube, sphere) and then complicate learning with Habitat.
It's not hard task to implement light simulator for 13D state space.

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: At the beginning we choose between stored state actions and heuristic ones depending on their weights. If no any similar experience we prefer heuristic.

h555yand (commented):
Hi Viviane,
thanks a lot for the feedback.
I've answered the questions, hope it helps to clarify them)
Looking forward to your comments.

As agreed, I will start prototyping based on the guide that you sent before.

Best regards,
Aleksei

rfcs/0000_RL_model_free_for_motor_system.md
vkakerbeck: Ah okay, thanks, that makes sense. The RL part focuses on the collision avoidance using sensory data.

rfcs/0000_RL_model_free_for_motor_system.md
vkakerbeck: The idea behind the surface agent was that once it is on the surface (taken care by a `PositioningProcedure` before an experiment starts), the agent should stay on the surface. The `SurfacePolicy` always goes through a sequence of 4 actions (see doc string in `get_next_action` function of the `SurfacePolicy` class). We could make a new action space where the agent can move freely in space, but also along a surface. You could also add other actions besides the ones we currently have in Monty by default. I would leave it up to you to decide what action space would be a good one to start with. I would just advocate for making sure that the solution is general enough that it works with different action spaces. Does that make sense?

rfcs/0000_RL_model_free_for_motor_system.md
vkakerbeck: Ah okay, that makes sense. Habitat is a bit problematic for many reasons and we are actually trying to move away from it ourselves (switching to MuJoCo). A light simulator sounds like the best solution in your case then.

vkakerbeck:
Nice! I am looking forward to hearing how the prototyping goes and feel free to reach out if you run into any questions (or just to give an update). You can also post updates on our Discourse forum (https://thousandbrains.discourse.group/latest) and/or we can also set up a call some time if that would be quicker to discuss details. 

Just as a quick sumamry of how you could prototype. There are two main options:
1. Make a fork of tbp.monty and edit the code on your fork directly. 
2. Start a new repository and import tbp.monty as a dependency. You can use our template repository for this: https://github.com/thousandbrainsproject/tbp.monty_project_template

If you plan to touch a lot of places across the code base, option 1 is probably best. If you would mostly add or customize specific classes (like in your case, the Motor System), option 2 might be easier. Here is an example repository where we went with option 2: https://github.com/thousandbrainsproject/tbp.ultrasound_perception (and a more recent one here: https://github.com/thousandbrainsproject/feat.evidence_sdr_matching) and one where we went with option 1: https://github.com/thousandbrainsproject/feat.compositional_testbed

Let me know if anything is unclear.

Best wishes,
Viviane

rfcs/0000_RL_model_free_for_motor_system.md
h555yand: Agree, Yes, action space is an input parameter and RL learns based on available actions.

h555yand (commented):
Hi Viviane,
many thanks for the clear guideline for next steps)

I agree, that the option 1 is more suitable for my case.
I've created the project and will push updates there
https://github.com/h555yand/hybrid_RL_prototype

At the very beginning I would prefer to share / discuss updates / questions with you and Niels. After having some source-code basement and initial ideas proofs I think we can share it with others on Discourse forum.
What do you think?
Anyway I prefer iteration approach and will keep you posted regularly)

Best regards,
Aleksei

vkakerbeck:
Sounds great! We are looking forward to your next updates :)

h555yand:
Hi Viviane,
I hope you are doing well)
Let me share updates on hybrid RL approach.
I've made Episodic memory part:
- Lightweight environment with trimesh
- HNSW-based State Store for Q-Learning
- RL Goal Approach Controller

Now I've been testing config parameters and bugfixing.
Source code is a work-in-progress, so it does not make sense to review it.
I hope I can present more concrete results in coming weeks)
Best regards, Aleksei

vkakerbeck:
Cool! Thank you for the update :) I am looking forward to hearing more

h555yand:
Hi Viviane,

after a lot of developing and testing iterations let me share  some updates on the hybrid RL prototype)

##  Main Proof of Concept Results

The prototype has been implemented and tested on the Lightweight Environment (Trimesh-based). Below are the key results that I hope validate the ideas from the RFC.

### Training Pipeline Validated

The full pipeline works end-to-end: **Q-learning → Behavioral Cloning → SAC → Arbitrage**

| Stage | Method | Success Rate | Object |
|-------|--------|-------------|--------|
| Q-learning (episodic memory) | HNSW + kNN + Heuristic-Guided Exploration | 65–70% | Mug |
| SAC (skill, softmax policy) | BC warm-start → SAC | 71–77% | Mug |
| Arbitrage (memory + skill) | Q-store + SAC + Heuristic | **78%** | Mug |
| SAC zero-shot transfer | No retraining | **73%** | Cup (new object) |

### Key Findings

**1. Episodic Memory (HNSW State Store) works as proposed.** One-shot/few-shot learning through Q-update is effective. Update hit rate 0.84–0.95 confirms that kNN + Gaussian Kernel interpolation correctly finds and updates similar states. The store grows organically — dense where the agent visits often, sparse where it rarely goes.

**2. Heuristic-Guided Exploration significantly outperforms ε-greedy.** This confirms the hypothesis from the RFC. Pure geometric heuristics (move toward goal, detach when stuck, steer in air) provide reasonable behavior from step 1, and Q-values gradually take over as experience accumulates.

**4. Arbitrage between episodic memory and skills works.** On familiar objects, Q-store handles 59% of decisions (precise, local knowledge), SAC handles 41% (general strategy). Agreement rate 31% shows they provide complementary information.

**5. On new objects transfer limitation discovered.** Q-store from one object can hurt performance on a new object (arbitrage zero-shot: 52% vs pure SAC: 73%).  This is expected — episodic memory is local by nature.  SAC generalizes better on new objects. This validates the skill-based approach — the neural network learns general navigation patterns that transfer across similar geometries. One more SAC advantage is continues parameters.
The solution can be:
- Start with pure SAC on new objects and let Q-store accumulate fresh experience.
- Use Online / Offline learning approach during working with new objects. If success rate below online threshold during 100 episodes to start online learning in parallel, if below offline threshold to start offline and after finish to continue working with new object.
---

## Answers to Open Questions

### Q1: Optimal State Dimensionality

I started with 13D and expanded to **15D** during testing. The additional features (mean curvature, Gaussian curvature) improved surface navigation.
The state vector is action-space independent — it describes the agent's situation relative to the goal, not the specific actions available.

### Q2: Hyperparameter Sensitivity

Extensive empirical testing was performed. Key findings:
- **Split the HNSW Q store** into two different graphs:
`q_store_free` — airborne state (on_object = 0). Navigation: turns, forward, landing
` q_store_surface` — surface state (on_object = 1). Crawling, detaching, orientation
The same position in space requires **opposite strategies** depending on whether the agent is touching the surface
- **Reward weights:** Increase penalty for collision from -5 to -15 
- **Action steps:** `surface_step=3mm`, `free_step=5mm`, `rotation_step=5°` — smaller steps reduce collisions but increase episode length.
- **SAC alpha bounds:** Critical for stability. `alpha_type ≥ 0.135` prevents policy collapse, `alpha_param ≤ 1.0` prevents parameter noise explosion.
- **Replay buffer:** BC-protected reservoir (15%) prevents catastrophic forgetting of expert demonstrations.
- **At the beginning I used 18 actions and then 2 macro actions were added**
| Index | Action | Category | Description |
|-------|--------|----------|-------------|
| 18 | Detach | macro | Detach from surface along normal, then fly toward goal (blending goal direction with surface normal to avoid collision). Multi-step action: lift off along normal → turn toward goal → series of forward steps with depth checking and collision avoidance → land on contact with surface. Each sub-step incurs a step penalty. Collisions during flight are penalized but do not terminate the episode — the agent is repositioned to pre-collision location and the macro action terminates early |
| 19 | DetachEdge | macro | Detach from surface along normal, then fly upward along the object's longest bounding-box axis until reaching the edge, turn toward goal and fly over to the other side. Used when the goal is through a thin wall (agent and goal normals are opposite). Same penalty and collision handling as Detach |

The Lightweight Environment (Trimesh) proved essential for rapid iteration — each experiment takes ~ 30-60 minutes on my laptop)

---

### On action space independence

> "Try and make the solution one that is action space independent"

**Status: Partially addressed.**
- `ActionSpace` is a configurable input parameter — adding or removing actions does not require architectural changes to Q-learning or SAC
- HNSW and SAC work with any number of discrete actions
- The system was tested with 18 actions (without detach) and 20 actions (with detach), confirming that the core learning pipeline adapts

**Limitation:** Heuristic biases currently reference specific action indices (e.g., `IDX_DETACH`, `IDX_FREE_FORWARD`). For a different action space, heuristics would need to be adapted. This is by design — heuristics encode domain-specific geometric reasoning that depends on what actions are available. A fully action-space independent heuristic system would require a mapping layer between geometric intentions (e.g., "move toward goal") and available actions.

### On the need for HNSW if good heuristics exist

> "If we have good heuristics to generate initial action sequences, what is the additional need for storing and retrieving them?"

**Answer from experiments:** Heuristics alone achieve ~48% success rate. 
Of course they can be improved, but it's continues effort to tune them for new cases.. HNSW stores the **learned corrections** — situations where heuristics were wrong and Q-learning motivated to discovery better actions. After training, Q-store + heuristics achieve 65–70%, and with SAC added, 78%.

The key insight: heuristics provide the **starting point**, episodic memory stores **exceptions and refinements**, and SAC learns **general patterns**. Each layer adds value.


Of course it's not enough to understand all things, I will prepare and share details later together with the source code. Now it works but it needs to review, comment, check one more time etc.
After that we can discuss details and next steps.

Best regards,
Aleksei

vkakerbeck:
Hi @h555yand 

Thank you for the update! Wow, you did a lot of work! Many results and details here and in the repo :) I have a lot of thoughts and questions. Would it be easiest to have a meeting sometime and talk through all of your results?

I saw you have a visualize_env.py script. Is it possible to share some visualizations of the learned policies in action? And a list of what your current main issues and open questions are?

Best wishes,
Viviane

h555yand:
Hi Viviane,

Aha, now It looks like a teaser)
Yes, sure. I plan to translate in English details that I already have and prepare some kind of visual demo for success and failed trails with descriptions, what iterations I did, what questions I have.
It's only first results and I also very interested to clarify my open questions and discuss them together.
I will let you know when I structure results and we set up a call for demo.

Best regards,
Aleksei

vkakerbeck:
Sounds great, looking forward to it!

h555yand:
Hi Viviane,

I've structured source code and updated Readme with training / validation strategy with links on detailed log results. Key findings / updates are there as well.
Now source code and Readme are consistent. 
Last thing that I plan to finish before set up a call is to log visualization details.
I can see it in debug mode but have some tech issues when saving it as a file with action description,  It's not a big deal to fix.

I will keep you posted)

Best regards,
Aleksei

h555yand:
Hi Viviane,

I've added visualization details and I think we can discuss results and next steps on call.
There are a lot of materials) What are your expectations from the call?

PS
I'm in GMT + 3 timezone.

Best regards,
Aleksei

vkakerbeck:
Great! Could you write me at vclay@thousandbrains.org for scheduling?

Where can I find the visualizations you mentioned? What would be the best resources for me to look through before our call? The reason I suggested a call is that it seems like a quicker way to ask questions and for me to get an understanding of everything, as well as give any feedback/thoughts on questions you might have (+it would be nice to meet you after all this technical discussion here!)

nielsleadholm:
Hi Aleskei, thanks again for meeting with us just now and all the great work you've started. In terms of the main next step, as discussed, it would be great to get your approach working with Mujoco in tbp.monty (continuing to work in your existing branch). Then it should be possible to evaluate the performance with the YCB objects. The important thing before we look into integrating this approach into `tbp.monty` would be to demonstrate the success rate on our usual environment (YCB objects) and its effect on object recognition.

Some other suggestions and comments are below.

**Re. visualizations + results**
- Could you turn some of the sets of images you've captured into videos? It would be quite interesting to see the policy in action. 
- Do you have any instances of the crawl alone being successful, particularly with curriculum 0 and 1? I.e. are there instances of success without needing to detach and fly? These would be interesting to see in action.
- It would be useful to further breakdown how performance improves as a function of learning. Some approaches to help with this would include:
	- Adding a metric for mean steps until goal reached (in addition to success rate). 
	- When you show the arbitrage result, also have a condition where we only use Q-store weight, a condition where we only use SAC weight, and then a one where you arbitrage between the two (as you currently do). That way, it will be clearer how each of these is changing as a function of learning (100, 1k, 1.5k, 2k, steps etc).
	- In the above, include a trend line / results for curriculum level 0, and curriculum level 1. For example, you showed a subtle improvement (51% --> 55%) on curriculum 2. Does that hold for easier problems (curriculum 0 and 1)? Is there even more improvement there?

**Re. improving performance:**
- As mentioned, you might try replacing mean and Gaussian curvature with the separate curvature components (kappa_1 and kappa_2), along with the vector directions of kappa_1 and kappa_2 (in the agent's reference frame). The system might be able to use this information to help crawl along curved surfaces towards a goal.
- At the moment, I don't believe the reward accounts for the orientation of the agent relative to the target surface normal. Instead, it's just based on the Euclidean distance. However, you could add in something similar to alignment, such that the agent's pose more closely orients towards pointing down at the target surface normal. This will prevent the policy from being close to the target object, without actually orienting towards it.
- It would be really cool if the policies you learn on basic geometric objects are sufficient to perform well on the full set of YCB objects. To enable this, it could be worth adding a few more basic geometric shapes, such as i) thin cylinder/pencil (which will manifest as locally more extreme principal curvature), ii) a flat square object like a "sheet of paper", iii) a sphere, iv) cone or pyramid. Even though YCB objects are quite complex, their local geometry can probably be covered by this set up objects, and so the learned policies might be sufficient.

h555yand:
Hi Niels,
Thank you for the detailed feedback and suggestions — very helpful for prioritizing next steps.

I agree that the main priority is getting the approach working with MuJoCo and evaluating on YCB objects. I'll continue working in my existing branch and focus on this integration.
I will ask specific questions if any when I start MuJoCo integration.

Before moving to MuJoCo, I'm currently running a new training iterations with several improvements based on issues I talked during our demo:
- Fixed and simplified the detach macro action.
- Improved BC data balancing — the previous BC was strongly biased by MoveTangentially, which caused SAC to mostly crawl. Now balanced across action types.
- SAC hyperparameter adjustments.

Once training iterations will be completed, I'll have a stronger baseline before moving to MuJoCo.

On your specific suggestions:
Re. visualizations:
- Videos: Great idea, I'll modify the function to convert the image sequences into videos.
- Crawl-only successes: Yes, it's possible, the function have filters. I'll prepare these episodes.
- Additional metrics: I'll add mean steps to goal, and run separate  evaluations for Q-store only, SAC only, and arbitrage. Breaking down  by curriculum level is straightforward — I'll include this in the next  results.

Re. improving performance:
- Curvature directions (kappa_1, kappa_2 + vectors): Agreed this would help with curved surface navigation. I'll add this after establishing  the current baseline, so we can measure the improvement.
- Orientation reward: Good point — currently reward is purely distance-based. 
  Adding an orientation component when close to the goal would help the agent arrive with the correct pose. I'll implement this in the next iteration.
- Additional geometric shapes: This is a great idea for generalization. 
  I'll add thin cylinder, flat square, sphere, and cone/pyramid to the training set. As you note, YCB objects' local geometry should be coverable by this expanded set.

Best regards,
Aleksei

nielsleadholm:
Perfect, thanks Aleksei. That all sounds great.

h555yand:
Hi Niels,
nice to communicate with you again)

Now we have solid baseline version on trimesh environment  with average success rate ~90%.
You can find details in readme:
https://github.com/h555yand/hybrid_RL_prototype/blob/adapt-baseline/README.md
Results with videos:
https://github.com/h555yand/hybrid_RL_prototype/tree/adapt-baseline/results_publish/adaptive_logs_cup
I'm happy to answer on questions.

There are many possible improvements and I think it makes sense to make them already during testing on MuJoCo YCB objects.
I will study integration with Monty MuJoCo and ask questions if any.

Best regards,
Aleksei

nielsleadholm:
That's great, thanks very much Aleksei. Looking forward to taking a closer look. Are you able to please summarize in your own words the results as they relate to the questions and suggestions I [posted earlier](https://github.com/thousandbrainsproject/tbp.monty/pull/745#issuecomment-5091422746)? The Readme is getting rather long, so this would be helpful. 

Note that, to encourage everyone in the TBP community to interact authentically, you are very welcome to respond in your native language if you prefer - I can then translate this on my end.

h555yand:
Hi Niels,
thanks for the feedback.

It's not a problem to translate my local minds in English)
Just there are many improvements to note so I suggest more efficient way to have a call as we did before.
# Agenda
## Present current status
 - Q, SAC, adaptive mode results on trimesh environment
 - YCB results on MuJoCo environment
## Key findings
- Q-learning, SAC, and adaptive arbitrage generalize to unseen objects (cup: 70-80% adaptive, never seen during training)
- Sim-to-real transfer works — policies trained on trimesh primitives navigate YCB objects in MuJoCo without retraining (banana 99%, can 81%, box 83%)
- The solution is ready for integration testing with Monty's Learning Module and Sensor Module
- This is a prototype. The goal is to demonstrate that the approach works, while being transparent about current limitations.
## Known Limitations
- Surface movement mechanics - edge traversal is unreliable on complex geometry.
- Navigation strategy for hollow objects - the agent doesn't always understand it needs to crawl to the rim, not toward the goal.
- Air navigation instability - flying through air is less reliable than surface crawling.
- Online SAC learning shows limited improvement - after 20 online SAC updates during 2000 adaptive episodes, SAC success rate did not meaningfully increase

I suggest to set up a call next week (from 14 September). I'm in GMT+3, if you are OK with it could you please let me know  your available timeslots?

Best regards,
Aleksei

PS
Current repo is:  https://github.com/h555yand/hybrid_RL_prototype/tree/mujoco

nielsleadholm:
Great, many thanks Aleksei,and yes sounds good. I'm away all of next week, but I could meet the week after. I'm in GMT, so maybe 10AM my time (1pm for you) on e.g. Wednesday, September 23rd could work well?

h555yand:
Hi Niels,
Yes, 10AM your time (1pm my time) on Wednesday, September 23rd looks good
for me.
Could you send a meeting request ?

Aleksei

вт, 8 сент. 2026 г. в 13:37, Niels Leadholm ***@***.***>:

nielsleadholm:
Hi @h555yand , it was nice chatting just now and seeing the latest results. I think your general proposal of integrating with a new fork of `tbp.monty` makes sense. That way you can explore the actual impact on inference performance. 

A few other suggestions and comments based on our discussion are below. Some of these include earlier comments I made and which I believe are still outstanding. 

## Re. visualizations + results

#### For Learning
- It would be useful to further breakdown how performance improves as a function of learning. Some approaches to help with this would include:
	- Adding a metric for mean steps until goal reached (in addition to success rate).
	- When you show the results across learning, have separate conditions where we only use Q-store weight, a condition where we only use SAC weight, one for heuristics only, and then one where you arbitrage between them. That way, it will be clearer how each of these is changing as a function of learning (e.g. at 100, 1k, 1.5k, 2k, steps), and you can actually plot how the above different mechanisms contribute differently as a function of training. For example, we would expect the heuristic to be flat (training has no effect), while Q-store and especially SAC improve with more training. 

#### For Inference
- At inference time, it would also be helpful to better understand the contribution of the different approaches. To do this, you can plot what proportion of steps use the result from Q-store, the result from SAC, from heuristics, or the blend. For example, it might be that, with arbitrage, only 2% of episode steps actually use SAC. We don't know this at the moment however. 
- Similarly, it would be helpful to add a table like the below, but which considers the arbitrage condition. That way it would be clearer how it compares to the standalone approaches.

Screenshot 2026-09-23 at 11 28 30 AM


## Re. improving performance:

- As mentioned, you might try adding the vector directions of kappa_1 and kappa_2 (in the agent's reference frame). The system might be able to use this information to help crawl along curved surfaces towards a goal. Note that the directions can be flipped arbitrarily 180 degrees - for them to be most useful, you might therefore pick the vector directions that point most directly towards the goal.
- It would be useful to debug why the transfer from sim (Trimesh) to "real" (YCB) performed so poorly on the cups/hollow objects. While I appreciate these are challenging, it is notable that you were getting pretty good performance on the Trimesh. It may be that there is a subtle bug here that is preventing a more graceful degradation in performance. 
- For goals on opposite ends of a surface (e.g. hollow objects / thin square), a certain amount of model-based influence is likely to be necessary to help guide any intelligent system around the object. Once you've integrated into your fork of Monty, it could be possible to update the hypothesis-testing policy to output a series of intermediate but reachable (with model-free policy) goals that ultimately arrive at the final goal which is e.g. on the other side of the object. That way the model-free policy doesn't have to achieve the (likely unreasonable) expectation of being able to navigate around such complex barriers.

h555yand:
Hi @nielsleadholm , 
many thanks for the quick feedback.
My comments below
1) Re. visualizations + results
Yes, ti makes sense. I have all information in detailed log files, it's just a question of presetting results.  I suggest to update tables already for MuJoCo adaptive mode validation for the next meeting.
2) Re. improving performance:
- kappa_1 and kappa_2 - I will try
-  transfer from sim (Trimesh) to "real" (YCB) 
Main reason - Edge traversal is unreliable on complex geometry in MuJoCo environment. The _move_tangentially + snap-to-surface mechanism struggles at sharp edges — mug, bowl rims. The agent attempts a tangential step, the snap algorithm fails to find the surface on the other side of the edge, and the agent either rolls back (stuck) or detaches into air (unintended). The issue is worse in MuJoCo because ray-cast-based snapping is less forgiving than trimesh's nearest-point projection.
This is an environment-level problem, not an RL problem. Even a perfect policy cannot crawl over an edge if the physics engine cannot execute the move. 
I have ideas how to improve it and will check.
- goals on opposite ends of a surface (e.g. hollow objects / thin square)
Current model-free RL is able to achieve goals  on opposite ends of a surface in Trimesh (70-80% success rate) and MuJoCo (even with edge traversal issue). When edge traversal issue is fixed in MuJoCo it's no need to split into sub goals specially. Anyway we will test it in real cases.

Best regards,
Aleksei

PR #750, opened 2026-01-31
refactor: Fix SIM118 violations: replace 'in dict.keys()' with 'in dict'

TimothyAlexisVass:
I picked a straightforward fix:
- Replace all membership checks using 'in dict.keys()' with 'in dict'
- Replace all 'for x in dict.keys():' with 'for x in dict:'
- Remove SIM118 from ruff ignore list in pyproject.toml

[SIM118](https://docs.astral.sh/ruff/rules/in-dict-keys/) recommends using 'key in dict' instead of 'key in dict.keys()' for membership checks and 'for key in dict' instead of 'for key in dict.keys()' for iteration, which is more Pythonic and avoids creating an intermediate view object.

### DIFF
--- a/src/tbp/monty/frameworks/models/object_model.py
+++ b/src/tbp/monty/frameworks/models/object_model.py
@@ -813,13 +813,12 @@ class GridObjectModel(GraphObjectModel):
                         avg_feat = previous_average
                     elif use_cds_to_update is False:
                         avg_feat[3:] = previous_average[3:]
-                elif feature == "object_id":
+                elif feature == "object_id" and avg_feat != previous_average:
                     # TODO: Figure out a more nuanced way to take into account past obs
-                    if avg_feat != previous_average:
-                        if num_old_obs > num_new_obs:
-                            avg_feat = previous_average
-                        else:
-                            previous_average = avg_feat
+                    if num_old_obs > num_new_obs:
+                        avg_feat = previous_average
+                    else:
+                        previous_average = avg_feat
                 # NOTE: could weight these
                 avg_feat = (avg_feat + previous_average) / 2
             target_ids = target_fm[feature]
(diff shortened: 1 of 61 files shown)


src/tbp/monty/frameworks/models/object_model.py
```diff
@@ -133,12 +133,12 @@ def feature_mapping(self):
 
     @property
     def edge_index(self):
-        if (self._graph is not None) and ("edge_index" in self._graph.keys):
+        if self._graph is not None and "edge_index" in self._graph.keys:
```
tristanls-tbp: **issue**: This change appears to be unrelated to SIM118.

src/tbp/monty/frameworks/models/object_model.py
```diff
@@ -133,12 +133,12 @@ def feature_mapping(self):
...
             return self._graph.edge_index
 
     @property
     def edge_attr(self):
-        if (self._graph is not None) and ("edge_attr" in self._graph.keys):
+        if self._graph is not None and "edge_attr" in self._graph.keys:
```
tristanls-tbp: **issue**: This change appears to be unrelated to SIM118.

src/tbp/monty/frameworks/models/object_model.py
```diff
@@ -773,9 +773,7 @@ def _get_new_voxel_features(
             New average features for a voxel.
         """
         new_feature_avg = np.zeros(target_feat_dim)
-        if ("pose_vectors" in obs_fm.keys()) and (
-            "pose_fully_defined" in obs_fm.keys()
-        ):
+        if "pose_vectors" in obs_fm and "pose_fully_defined" in obs_fm:
```
tristanls-tbp: **issue**: Parens drop appears unrelated to SIM118.

tristanls-tbp:

PR #751, opened 2026-01-31
refactor: Fix SIM102 violations: combine nested if statements

TimothyAlexisVass:
* Fix SIM102 violations by combining nested if statements into single conditions
* Remove SIM118 from ruff ignore list in pyproject.toml

[SIM102](https://docs.astral.sh/ruff/rules/merge-else-if/) recommends using a single `if` statement with combined conditions instead of nested `if` statements when the inner `if` has no `else` clause. This is more Pythonic, improves readability, and avoids unnecessary nesting by combining conditions with `and`.

### DIFF
--- a/docs/how-to-use-monty/tutorials/multiple-learning-modules.md
+++ b/docs/how-to-use-monty/tutorials/multiple-learning-modules.md
@@ -168,8 +168,8 @@ learning_module_args:
   # Use this to update all hypotheses > 80% of the max hypothesis evidence
   evidence_threshold_config: 80%
   x_percent_threshold: 20
-  gsg_class: ${monty.class:tbp.monty.frameworks.models.goal_state_generation.EvidenceGoalStateGenerator}
-  gsg_args:
+  gsg:
+    _target_: tbp.monty.frameworks.models.goal_state_generation.EvidenceGoalStateGenerator
     # Tolerance(s) when determining goal-state success
     goal_tolerances:
       location: 0.015 # distance in meters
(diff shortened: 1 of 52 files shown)


src/tbp/monty/frameworks/models/object_model.py
```diff
@@ -816,11 +816,10 @@ def _get_new_voxel_features(
...
                     # TODO: Figure out a more nuanced way to take into account past obs
-                    if avg_feat != previous_average:
-                        if num_old_obs > num_new_obs:
-                            avg_feat = previous_average
-                        else:
-                            previous_average = avg_feat
+                    if avg_feat != previous_average and num_old_obs > num_new_obs:
+                        avg_feat = previous_average
+                    else:
+                        previous_average = avg_feat
```
tristanls-tbp: **issue**: This is not equivalent to the previously existing logic.

tristanls-tbp (changes requested):
Thank you. I think there's one logic error. Otherwise, looks great!

src/tbp/monty/frameworks/models/object_model.py
TimothyAlexisVass: You are correct, even if the logical end result is the same, there is a redundant reassignment of the same value.
Which would break functionality if assignment has side-effects.
If you think this is correct, please feel free to commit it.

```suggestion
                    if avg_feat != previous_average and num_old_obs > num_new_obs:
                        avg_feat = previous_average
                    elif avg_feat != previous_average:
                        previous_average = avg_feat
```

tristanls-tbp: **question**: Should the change be at one level up in the control flow?
```python
elif feature == "object_id" and avg_feat != previous_average:
    if num_old_obs > num_new_obs:
        avg_feat = previous_average
    else:
        previous_average = avg_feat
```

TimothyAlexisVass: Yes! I think that is cleaner and more direct.

TimothyAlexisVass:
@tristanls-tbp I felt like your final suggestion was the optimal solution, so I committed that change.

tristanls-tbp:

PR #752, opened 2026-01-31
[DO NOT MERGE] Reference for Improving Hydra Configs

hlee9212:
Hi @tristanls-tbp  and @jeremyshoemaker, here's my draft PR in case it will be helpful during your Feb 2026 focus week. Instead of making a draft PR with 100+ files, I decided to just include a few to illustrate the points. Feel free to incorporate as much or as little suggestions from below, leave questions, and close it when it is no longer needed. 

# Changes I made to Hydra Configs

1. Use of `@package` syntax and `override`. 
- Using these will eliminate the ugly and long injection syntax under `defaults`, and all files with `clear_*.yaml` pattern. 
- Adding `@package` to first line of files should be fairly straightforward. 
- Using `override` requires you to think about how the code is (or should be) organized by modules, which is much harder. See [conf/experiment.yaml](https://github.com/thousandbrainsproject/feat.2d_sensor/blob/1e8f6c519dbae42be3532154383eccb43839c583/conf/experiment.yaml) and [conf/monty/benchmark_evidence_sota.yaml](https://github.com/thousandbrainsproject/feat.2d_sensor/blob/1e8f6c519dbae42be3532154383eccb43839c583/conf/monty/benchmark_evidence_sota.yaml), which shows how I tried to organize it. 

**Update on the last point:** Actually, I think the current organization under `conf/experiment/config` is fine (i.e. `environment`, `environment_interface`, `logging`, `monty`) and these could be the initial config groups. 

2. Reorganization
- Small QoL: I organized experiments into separate folders, as shown below:
```
conf/experiment
├── 2d_sensor_module
│   ├── inference
│   └── learning
├── benchmarks
│   ├── compositional
│   │   ├── inference
│   │   └── training
│   ├── monty_meets_world
│   ├── unsupervised
│   │   ├── inference
│   │   └── learning
│   └── ycb
│       ├── 10obj
│       ├── 77obj
│       └── pretraining
└── tutorials
```
- I moved most of the folders from `conf/experiment/configs` to `conf/`, something like shown below. These folders are called [config groups in Hydra](https://hydra.cc/docs/tutorials/basic/your_first_app/config_groups/) and also reflected in the organization of [conf/experiment.yaml](https://github.com/thousandbrainsproject/feat.2d_sensor/blob/1e8f6c519dbae42be3532154383eccb43839c583/conf/experiment.yaml).
```
conf
├── _experiment # just renamed this so it's not mixed in with below
├── agent
│   └── transform
├── environment
├── hydra
├── logging
├── monty
│   ├── _preset
│   ├── connectivity
│   ├── learning_module
│   │   ├── feature_weights
│   │   ├── gsg
│   │   ├── hypotheses_updater
│   │   └── tolerances
│   ├── monty
│   ├── motor_system
│   │   ├── action_space
│   │   └── policy
│   └── sensor_module
│       ├── patch
│       └── view_finder
├── mode
└── vars # I renamed "benchmarks" to vars but basically contains bunch of variables like rotations 
```
**Update:** Again, I think the current organization is fine and can be moved outside from `conf/experiment/config` to `conf/` directly. 

## Guiding Principles
1. Identify "axes of variation" - these should be folders under `conf/`.
- These are like "decision points", e.g. an experiment consists of environment, logging, Monty (which consists of SM, LM, Motor System), etc. I mostly followed this:
image

- In a perfectly ideal world there should be about 1:1 correspondence between the Python class and config file. 
- **Update:** Again, I'm fine with current organization already.

2. Identify presets or "bundles" that are actually commonly used and those that are too specific (e.g. I think there used to be something like `benchmark_evidence_sota_noisy.yaml` that was a Monty preset used in one benchmark experiment). 
- **Update:** Seems like a lot of these have been identified and removed already. :) 
3. The top-level experiment should mostly a be thin selector.

## Other Tips
1. Keep packaging syntax at the leaf configs, and let the preset/bundle just select leaves.  
2. Rule of thumb for using `_self_`:
	1. Put `_self_` last in a "thin selector" file so local values in that file override imported defaults
	2. Put `_self_` early in "base defaults" if you want imported configs to override base values 
	5. Use `python run.py experiment=NAME --info defaults-tree` to verify the order.

I realize this is a monumental task (thanks!). Maybe it'll be possible to try to refactor out one config group (or "axis of variation") at a time? The simplest one to refactor out is dataset, as shown below. This means under "experiment.yaml" we can add in a config group, like "dataset: ycb" and any custom experiments can override with their own dataset. This one may sound trivial but it may actually be very helpful to users with their own datasets, and for me where I'm making a bunch of synthetic datasets to test my RP. I'm also happy to help "refactor" out sensor module related configs when I get to my IP project. 

```
conf/object_set
├── compositional
│   ├── curved_objects.yaml
│   ├── flat_objects.yaml
│   ├── logos.yaml
│   ├── lvl1.yaml
│   ├── lvl2.yaml
│   ├── lvl3.yaml
│   ├── lvl4.yaml
│   └── parent_to_child_mapping.yaml
├── numenta_lab.yaml
├── omniglot.yaml
├── tutorial.yaml
└── ycb
    ├── all_77.yaml
    ├── distinct_10.yaml
    └── similar_10.yaml
```

Random point: The purpose of [conf/hydra/defaults.yaml](https://github.com/thousandbrainsproject/feat.2d_sensor/blob/1e8f6c519dbae42be3532154383eccb43839c583/conf/hydra/defaults.yaml) was a QoL for me to save the hydra config with the experiment output folder instead of under `outputs/` in our root folder. Yes, you can configure Hydra via Hydra. 😆

### DIFF
--- a/conf/experiment.yaml
+++ b/conf/experiment.yaml
@@ -1,0 +1,19 @@
+# @package _global_
+
+defaults:
+  - _self_
+  - vars: defaults 
+  - logging: defaults
+  - monty: patch_and_view_inference
+  - agent: patch_view_habitat
+  - environment: ycb_distinct
+  - mode: inference
+  # Experiment config (required) - use override to change defaults above
+  - experiment: MISSING
+
+# Global settings
+config:
+  episodes: all
+  num_parallel: 16
+  print_cfg: false
+  quiet_habitat_logs: true
--- a/conf/experiment/benchmarks/ycb/inference/base_config_10distinctobj_dist_agent.yaml
+++ b/conf/experiment/benchmarks/ycb/inference/base_config_10distinctobj_dist_agent.yaml
@@ -1,0 +1,29 @@
+# @package _global_
+
+defaults:
+  - override /logging: defaults
+  - override /monty: benchmark_evidence_sota
+  - override /agent: patch_view_habitat
+  - override /environment: ycb_distinct
+
+_target_: tbp.monty.frameworks.experiments.object_recognition_experiments.MontyObjectRecognitionExperiment
+
+config:
+  model_name_or_path: ${vars.pretrained_dir}/surf_agent_1lm_10distinctobj/pretrained/
+  n_eval_epochs: ${vars.rotations_all_count}
+  logging:
+    python_log_level: WARNING
+    monty_log_level: BASIC
+    monty_handlers:
+      - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.BasicCSVStatsHandler}
+      - ${monty.class:tbp.monty.frameworks.loggers.monty_handlers.ReproduceEpisodeHandler}
+    wandb_handlers:
+      - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbTableStatsHandler}
+      - ${monty.class:tbp.monty.frameworks.loggers.wandb_handlers.BasicWandbChartStatsHandler}
+    wandb_group: benchmark_experiments
+    output_dir: ${path.expanduser:${oc.env:MONTY_LOGS}/projects/monty_runs}
+    run_name: base_config_10distinctobj_dist_agent
+  monty_config:
+    monty_args:
+      min_eval_steps: ${vars.min_eval_steps}
+  eval_env_interface_class: ${monty.class:tbp.monty.frameworks.environments.embodied_data.InformedEnvironmentInterface}
--- a/conf/mode/_defaults.yaml
+++ b/conf/mode/_defaults.yaml
@@ -1,0 +1,16 @@
+# @package _global_
+# Base experiment configuration - all modes extend this
+
+config:
+  do_train: true
+  do_eval: true
+  show_sensor_output: false
+  max_train_steps: 1000
+  max_eval_steps: 500
+  max_total_steps: 6000
+  n_train_epochs: 3
+  n_eval_epochs: 3
+  model_name_or_path: ""
+  min_lms_match: 1
+  seed: 42
+  supervised_lm_ids: []
--- a/conf/agent/patch_view_habitat.yaml
+++ b/conf/agent/patch_view_habitat.yaml
@@ -1,0 +1,30 @@
+# @package config.env_interface_config
+
+defaults:
+  - transform/patch_view_distant@_here_
+
+env_init_func: ${monty.class:tbp.monty.simulators.habitat.environment.HabitatEnvironment}
+env_init_args:
+  objects:
+    - name: coneSolid
+      position: [0.0, 1.5, -0.1]
+  scene_id: null
+  seed: 42
+  data_path: ${path.expanduser:${oc.env:MONTY_DATA}/habitat/objects/ycb}
+  agents:
+    agent_type: ${monty.class:tbp.monty.simulators.habitat.MultiSensorAgent}
+    agent_args:
+      agent_id: ${monty.agent_id:agent_id_0}
+      sensor_ids:
+        - patch
+        - view_finder
+      height: 0.0
+      position: [0.0, 1.5, 0.2]
+      resolutions: [[64, 64], [64, 64]]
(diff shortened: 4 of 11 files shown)


conf/mode/_defaults.yaml
tristanls-tbp, line 1: **question**: Is this prefixed with `_` so that it doesn't conflict with the `defaults` keyword?

conf/experiment/benchmarks/ycb/inference/base_config_10distinctobj_dist_agent.yaml
tristanls-tbp, line 1: **question**: How is this called from the command line? `python run.py experiment=?`.

conf/experiment.yaml
```diff
@@ -0,0 +1,19 @@
+# @package _global_
```
tristanls-tbp, line 1: **question**: Using `_global_` here is redundant, yes? Since we're in `conf/experiment.yaml`.

conf/experiment/benchmarks/ycb/inference/base_config_10distinctobj_dist_agent.yaml
```diff
@@ -0,0 +1,29 @@
+# @package _global_
```
tristanls-tbp, line 1: **question**: Shouldn't this be `# @package experiment`?

conf/mode/_defaults.yaml
hlee9212: No, I just added that so it's sorted nicely when I use `tree`. Actually I'd probably go with `default` (singular). Example use [here](https://github.com/ashleve/lightning-hydra-template/tree/main/configs/callbacks).

conf/experiment/benchmarks/ycb/inference/base_config_10distinctobj_dist_agent.yaml
hlee9212: `python run.py experiment=benchmarks/ycb/inference/base_config_10distinctobj_dist_agent`

conf/experiment.yaml
hlee9212: Yes, it is redundant.

conf/experiment/benchmarks/ycb/inference/base_config_10distinctobj_dist_agent.yaml
hlee9212: Yes, you are right. I used `# @package _global_` because I also updated my `run.py` to the following:

```python
@hydra.main(config_path="../../../conf", config_name="experiment", version_base=None)
def main(cfg: DictConfig):
    if cfg.config.quiet_habitat_logs:
        os.environ["MAGNUM_LOG"] = "quiet"
        os.environ["HABITAT_SIM_LOG"] = "quiet"
    print(OmegaConf.to_yaml(cfg))

    output_dir = Path(cfg.config.logging.output_dir) / cfg.config.logging.run_name
    cfg.config.logging.output_dir = str(output_dir)

    output_dir.mkdir(exist_ok=True, parents=True)
    experiment = hydra.utils.instantiate(cfg) # <-- instantiating from cfg, not cfg.experiment!!
    start_time = time.time()
    with experiment:
        experiment.run()

    logger.info(f"Done running {experiment} in {time.time() - start_time} seconds")
```

(Note that I'm instantiating from `cfg` directly, not `cfg.experiment`)

I wasn't sure of the benefit of nesting it under `experiment` (maybe there is a good reason? the only reason I could think of is that maybe we want to specify some variables outside of `experiment` instead of passing them as arguments?) and wanted a flatter structure. I'm down to instantiate from `cfg.experiment` - I was just playing around, using structure I am more used to (example [here](https://github.com/ashleve/lightning-hydra-template/blob/main/configs/experiment/example.yaml) for config and [here](https://github.com/ashleve/lightning-hydra-template/blob/bddbc24b82ab6ccfa6243e815a49dc5bfe8d4144/src/train.py#L57-L70) for "run.py"), and forgot to explain/note/include in this PR that I also updated the `run.py`. 🤪

PR #753, opened 2026-02-01
refactor: Fix RUF015 violations: use next(iter()) instead of single element slice

TimothyAlexisVass:
* Fix RUF015 violations by replacing `list(dict.keys())[0]` with `next(iter(dict.keys()))`
* Remove RUF015 from ruff ignore list in pyproject.toml

[RUF015](https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/) recommends using `next(iter(iterable))` instead of `list(iterable)[0]` when accessing the first element of an iterable. This is more efficient as it avoids creating an intermediate list, uses less memory, and is more Pythonic for single-element access from iterables like dictionaries.


tristanls-tbp (approved):
Thank you, looks good.

tristanls-tbp:

PR #755, opened 2026-02-02
refactor!: HabitatSalienceSM accepts instances of collaborators

tristanls-tbp:
This pull request removes `*_class`+`*_args` instantiation pattern in `HabitatSalienceSM` and replaces it with passing a created instance into the constructor, facilitating dependency injection.

It doesn't change behavior, but since it changes the API, I marked it a breaking change.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
 # Copyright 2024 Numenta Inc.
 #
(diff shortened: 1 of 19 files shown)


jeremyshoemaker (approved):
I guess we aren't overriding any of this in the configs?

tristanls-tbp:
> I guess we aren't overriding any of this in the configs?

Right, I was surprised too. These turned out to be injected dependencies so that we could mock them in tests. But, in production use, there are only singular implementations of these, so they don't require configuration.

Issue #52, opened 2024-11-21
Windows OS not supported?


internetscooter:
### Describe the bug

Hi TBP,

I think Windows OS is not supported due to https://github.com/facebookresearch/habitat-sim/issues/1570 - if this is the case there should be a note on this in the code/docs?

Below is the complete message. If the above is the case I will see what workaround I can come up with.

`PS C:\code\tbp.monty> conda env create        
C:\Users\paulm\anaconda3\Lib\argparse.py:2006: FutureWarning: `remote_definition` is deprecated and will be removed in 25.9. Use `conda env create --file=URL` instead.
  action(self, namespace, argument_values, option_string)
Channels:
 - aihabitat
 - pytorch
 - pyg
 - defaults
Platform: win-64
Collecting package metadata (repodata.json): done
Solving environment: failed

PackagesNotFoundError: The following packages are not available from current channels:

  - aihabitat::habitat-sim==0.2.2
  - wget

Current channels:

  - https://conda.anaconda.org/aihabitat
  - https://conda.anaconda.org/pytorch
  - https://conda.anaconda.org/pyg
  - https://repo.anaconda.com/pkgs/main
  - https://repo.anaconda.com/pkgs/r
  - https://repo.anaconda.com/pkgs/msys2
  - https://conda.anaconda.org/conda-forge

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.`

internetscooter:
Would yo be happy including SciPy?

[scipy.spatial.transform.Rotation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html#scipy.spatial.transform.Rotation )

tristanls:
@vkakerbeck, @nielsleadholm any thoughts on habitat_sim.utils replacements and @internetscooter 's suggestion above?

vkakerbeck:
I agree, scipy's Rotation function seems the most straight forward function to replace the use of quat_rotate_vector (especially since we are using this function in all the other places). It should already be included in the installed dependencies as well (https://github.com/thousandbrainsproject/tbp.monty/blob/70d610af588c71f2d239d8fb0f1146fac44649a1/pyproject.toml#L26) 
[...]

PR #756, opened 2026-02-03
refactor!: move `Monty.pre_episode`'s `rng` to `Monty.step` via `RuntimeContext`

AgentRev:
As suggested by [this comment](https://github.com/thousandbrainsproject/tbp.monty/issues/506#issuecomment-3808232223), the following PR moves the RNG from `Monty.pre_episode(...)` into a new `RuntimeContext` type, which is now passed via `Monty.step(...)`. This only covers the `Monty` object for now, i.e. "Phase 1" as suggested by the comment. I'm unsure if I instantiate the `ctx` and `rng` objects at the desired locations, and if my `set_context` methods are what you wanted or not, so don't hesitate to suggest changes.

### DIFF
--- a/benchmarks/ycb_unsupervised_inference.csv
+++ b/benchmarks/ycb_unsupervised_inference.csv
@@ -1,3 +1,3 @@
 Experiment,Correct (%)|align right,Num Match Steps|align right,Run Time (mins)|align right,Episode Run Time (s)|align right
-unsupervised_inference_distinctobj_dist_agent,99.00,96,9,5
-unsupervised_inference_distinctobj_surf_agent,100.00,99,18,11
+unsupervised_inference_distinctobj_dist_agent,98.00,98,11,5
+unsupervised_inference_distinctobj_surf_agent,100.00,100,21,12


src/tbp/monty/frameworks/models/abstract_monty_classes.py
```diff
@@ -58,15 +60,21 @@ class Observations(Dict[AgentID, AgentObservations]):
     pass
 
 
+@dataclass
+class RuntimeContext:
```
tristanls-tbp: **issue**: Please move to `src/tbp/monty/context.py`

src/tbp/monty/frameworks/models/abstract_monty_classes.py
```diff
@@ -58,15 +60,21 @@ class Observations(Dict[AgentID, AgentObservations]):
...
     ###
     # Methods that specify the algorithm
     ###
-    def _matching_step(self, observation):
+    def _matching_step(self, ctx: RuntimeContext, observation):
         """Step format for matching observations to graph.
 
         Used during training or evaluation.
         """
+        self.set_context(ctx)
```
tristanls-tbp: **issue**: Please don't set a context anywhere and pass it through the methods where it is needed instead.

src/tbp/monty/frameworks/models/monty_base.py
```diff
@@ -134,13 +132,13 @@ def __init__(
     # Basic methods that specify the algorithm
     ###
 
-    def step(self, observation):
+    def step(self, ctx: RuntimeContext, observation):
         # For the base class, just use matching step. Note that matching_step and
         # exploratory_step are fully implemented by the abstract class.
         if self.step_type == "matching_step":
-            self._matching_step(observation)
+            self._matching_step(ctx, observation)
```
tristanls-tbp, line 142: **praise**: Yes, this is the pattern. And it needs to be continued _all the way down_ to the last use of `ctx.rng`.

tristanls-tbp (commented):
Thank you for getting started on this.

Apologies, I don't have the time to do the full review, but I do want to comment sooner rather than later on the general pattern.

`set_context` should not be used. `ctx`, should be passed _all the way down_ through all the methods to the last place it `ctx.rng` is used.

src/tbp/monty/frameworks/models/abstract_monty_classes.py
jeremyshoemaker: This dataclass needs a docstring, what it is, why it exists.

AgentRev:
Thank you both, that's exactly the kind of feedback I was hoping for.

tristanls-tbp:
@AgentRev, in case you'll get to this PR this weekend. I want to call your attention to #762 and #763. They thread the `RuntimeContext` on the environment side of things: motor system, motor policies, positioning procedures, and environment interfaces. Once #763 is merged, you won't even need to create the runtime context, since all experiments will already create it.

AgentRev:
Nice, that was actually on my todo list, this will make things easier.

AgentRev:
Made some tentative changes in a separate branch based off the `tristanls-tbp:env_ctx` branch from PR 763: https://github.com/AgentRev/tbp.monty/commit/b97929509a30bb97e7dd0b38d8573a9ff281b4da

tristanls-tbp:
@AgentRev , #763 got merged, so I'm sure that'll be exciting for this code branch 😬.

src/tbp/monty/frameworks/experiments/data_collection_experiments.py
```diff
@@ -101,14 +101,17 @@ def pre_episode(self):
...
         self.env_interface.pre_episode(self.rng)
         self.max_steps = self.max_train_steps
         self.logger_handler.pre_episode(self.logger_args)
         if self.show_sensor_output:
             self.live_plotter.initialize_online_plotting()
 
-    def post_episode(self):
+    def post_episode(
+        self,
+        steps,  # noqa: ARG002
```
tristanls-tbp, line 113: **issue**: This adds an unused argument.

src/tbp/monty/frameworks/models/salience/sensor_module.py
```diff
@@ -11,7 +11,10 @@
 import numpy as np
 import quaternion as qt
 
-from tbp.monty.frameworks.models.abstract_monty_classes import SensorModule
+from tbp.monty.context import RuntimeContext
+from tbp.monty.frameworks.models.abstract_monty_classes import (
+    SensorModule,
+)
```
tristanls-tbp: **question**: Why the extra parens? It clearly fits on one line.

src/tbp/monty/frameworks/models/salience/sensor_module.py
tristanls-tbp: **question**: Any reason for not removing the unused `self._rng`?

src/tbp/monty/frameworks/models/abstract_monty_classes.py
```diff
@@ -343,7 +337,7 @@ def output_goal_states(self) -> list[GoalState]:
         pass
 
     @abc.abstractmethod
-    def step(self):
+    def step(self, ctx: RuntimeContext, observations: Observations):
```
tristanls-tbp, line 340: **question**: Why add `ctx` if it's going to be unused?

src/tbp/monty/frameworks/models/graph_matching.py
```diff
@@ -24,7 +25,10 @@
     DetailedGraphMatchingLogger,
     SelectiveEvidenceLogger,
 )
-from tbp.monty.frameworks.models.abstract_monty_classes import LearningModule, LMMemory
+from tbp.monty.frameworks.models.abstract_monty_classes import (
+    LearningModule,
+    LMMemory,
+)
```
tristanls-tbp, line 31: **question**: Why extra parens?

src/tbp/monty/frameworks/models/graph_matching.py
```diff
@@ -635,12 +635,16 @@ def matching_step(self, observations):
             self.set_individual_ts(terminal_state="no_match")
 
         if self.gsg is not None:
-            self.gsg.step(observations)
+            self.gsg.step(ctx, observations)
```
tristanls-tbp, line 639: **question**: I asked this elsewhere, but why pass `ctx` to GSG step?

src/tbp/monty/frameworks/models/graph_matching.py
tristanls-tbp: **question**: Any reason to keep the unused `self._rng`?

src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -18,7 +18,11 @@
 from scipy.spatial.transform import Rotation
 from skimage.color import rgb2hsv
 
-from tbp.monty.frameworks.models.abstract_monty_classes import SensorID, SensorModule
+from tbp.monty.context import RuntimeContext
+from tbp.monty.frameworks.models.abstract_monty_classes import (
+    SensorID,
+    SensorModule,
+)
```
tristanls-tbp: **question**: Parens again, but why?

src/tbp/monty/frameworks/models/sensor_modules.py
```diff
@@ -408,6 +409,7 @@ def __init__(
         super().__init__()
 
         self.is_exploring = False
+        self._rng = rng
```
tristanls-tbp: **issue**: Please, definitely don't _add_ an RNG anywhere. The point of RuntimeContext is to _remove_ RNG from everything downstream of Monty.step().

tests/unit/frameworks/models/salience/sensor_module_test.py
```diff
@@ -82,13 +83,17 @@ def setUp(self) -> None:
             rotation=self.default_sensor_state.rotation,
         )
 
+        # Generic RNG for tests that don't contain experiments
```
tristanls-tbp: **issue**: I don't understand this comment. Is it necessary?

tests/unit/frameworks/models/salience/sensor_module_test.py
```diff
@@ -82,13 +83,17 @@ def setUp(self) -> None:
...
+        seed = 42
+        self.ctx = RuntimeContext(rng=np.random.RandomState(seed))
```
tristanls-tbp: **question**: Is seeding this RNG necessary? Will the tests pass with any seed (i.e. `np.random.RandomState()`)?

tests/unit/frameworks/models/salience/sensor_module_test.py
```diff
@@ -170,6 +175,10 @@ def setUp(self) -> None:
             snapshot_telemetry=MagicMock(),
         )
 
+        # Generic RNG for tests that don't contain experiments
+        seed = 42
+        self.ctx = RuntimeContext(rng=np.random.RandomState(seed))
```
tristanls-tbp: **issue**: Same comments as for the previous class.

tests/unit/graph_learning_test.py
```diff
@@ -611,16 +611,14 @@ def gm_learn_object(
...
-        )
+        graph_lm.pre_episode(primary_target=self.placeholder_target)
 
         offset_obs = []
         for observation in observations:
             obs_to_learn = copy.deepcopy(observation)
             obs_to_learn.location += offset
             offset_obs.append(obs_to_learn)
-            graph_lm.exploratory_step([obs_to_learn])
+            graph_lm.exploratory_step(self.ctx, [obs_to_learn])
```
tristanls-tbp, line 621: **issue**: Here and in the rest of the file, is `self.ctx` defined? If it is, I'm not seeing it.

AgentRev: It originates from `BaseGraphTest.setUp()`, should I maybe rename it to `self.testCtx` for clarity?

tests/unit/frameworks/models/salience/sensor_module_test.py
AgentRev: Well, this RNG is for unit tests. Usually, you'd want tests to be deterministic, ensuring it doesn't affect the outcomes of test runs, e.g. a test failing one run then passing the next without changes. That kinda stuff is _not_ fun to debug, especially when the test author, in all their wisdom, did not log the seed... [_Vietnam War flashbacks_]

tests/unit/frameworks/models/salience/sensor_module_test.py
AgentRev: Multiple non-experiment tests call functions that require a `RuntimeContext` (which normally originates from `MontyExperiment`), so I created these "generic" `RuntimeContext` variables in the `setUp()` of most top-level test classes, for usage by those function calls. Maybe the comment is not worded properly? Should I instead write like "Fallback RuntimeContext for non-experiment tests", or something else?

src/tbp/monty/frameworks/models/sensor_modules.py
AgentRev: Probably a leftover that I forgot to undo from the first commit.

src/tbp/monty/frameworks/models/salience/sensor_module.py
AgentRev: Heh, that's just Ruff being ruff... My `RuntimeContext` was initially in `abstract_monty_classes` before you introduced yours in `tbp.monty.context`, and `ruff format` automatically added a trailing comma to the multi-line import. Later, I deleted mine from `abstract_monty_classes` and imported yours, then `ruff check --fix` removed my now-unused import, but didn't collapse it back into a single line, because the trailing comma previously added by `format` explicitly prevents `check --fix` from collapsing it. 😵

Fixing that by hand is a bit involved, basically gotta double-check the imports for each modified file before every commit. I did that in my previous PRs, but it was 2 AM when I updated this PR, so this likely slipped out of my depleted neocortex. There is apparently no formatter option to remove trailing commas from imports. There is however an option to ignore trailing commas, but it applies to arguments too and not just imports, so I dunno if it would be acceptable or not: https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma

src/tbp/monty/frameworks/models/salience/sensor_module.py
AgentRev: Oops, missed it

src/tbp/monty/frameworks/models/graph_matching.py
AgentRev: Oops, missed it

src/tbp/monty/frameworks/models/abstract_monty_classes.py
AgentRev: I thought you wanted `ctx` added to all `step` functions downstream of the `Monty.step()` callstack, did I misinterpret? (I assumed this unused arg would be for optional later usage)

src/tbp/monty/frameworks/experiments/data_collection_experiments.py
AgentRev: Solves the warning `Signature of method 'DataCollectionExperiment.post_episode()' does not match signature of the base method in class 'MontyExperiment'`

For a non-Monty project, I would normally solve it with a good ol' `(self, _)` but I think you previously mentioned preferring `# noqa: ARG002` for this, right?

tristanls-tbp: Ah, ok, thank you for pointing that out. 👍.

tristanls-tbp: The implementation and `noqa` is good, I was just confused why `steps` was added.

src/tbp/monty/frameworks/models/salience/sensor_module.py
tristanls-tbp: Wow. (Ruff got pretty complicated there 🙂)

Well, I'm not asking to fix all the imports in all the files. It feels like removing the trailing comma manually should restore the previous imports, no?

https://github.com/user-attachments/assets/a44af075-fcbe-41fb-a61d-d3b89d5ff4af

src/tbp/monty/frameworks/models/abstract_monty_classes.py
tristanls-tbp: Oh, I see. Sorry for not being clearer. The caveat is that it should be passed only to where it is used. For example, in the below diff, you didn't include `ctx` in all the other methods. Same reasoning here, if it's unused, it doesn't need the `ctx` argument.

Screenshot 2026-02-11 at 17 57 40

tests/unit/frameworks/models/salience/sensor_module_test.py
tristanls-tbp: I agree with the goal of deterministic tests. However, the `HabitatSalienceSM` does need to perform with seeds other than `42`. Another way to write deterministic tests is to test invariants that should always hold when randomness is involved, rather than testing for specific (and very brittle) values. At some level, picking the right seed so that the tests pass is equivalent to hardcoding a test to `assert True`.

Having said that, I see that the test already hardcodes the 42 seed, so my complaint is outside the scope of this PR, as this simply formats that seed in a different way.

For what it's worth, I looked over the tests, and they're testing collaborator calls and never use `ctx.rng`.

tests/unit/frameworks/models/salience/sensor_module_test.py
tristanls-tbp: Typically, in a test `setUp` method, there is no other use for things attached to `self` but to be some sort of generic thing used by the test methods. I don't think any comment is necessary.

tests/unit/graph_learning_test.py
tristanls-tbp: And this is why I'm not a fan of inheritance 🙂 ... Ok, thank you. I'm ok with it to keep the scope of this PR reasonable and avoid getting into removing inheritance.

tests/unit/policy_test.py
```diff
@@ -155,6 +155,10 @@ def hydra_config(test_name: str) -> DictConfig:
             State(**fo_2_corrupt_z),
         ]
 
+        # Generic RNG for tests that don't contain experiments
+        seed = 42
+        self.ctx = RuntimeContext(rng=np.random.RandomState(seed))
+
```
tristanls-tbp: **issue**: This is here because of a test near line 925. Notice that in that test, there is no seed provided, so the test is already resilient to different seeds. Since it is only used in the one test, please move this back down to the one test this is used in and remove the seed to correspond to the removed `np.random.RandomState()`

tests/unit/evidence_sdr_lm_test.py
```diff
@@ -340,7 +345,11 @@ def setUp(self):
         self.output_dir = tempfile.mkdtemp()
 
         # set seed for reproducibility
-        set_seed(42)
+        seed = 42
+        set_seed(seed)
+
+        # Generic RNG for tests that don't contain experiments
+        self.ctx = RuntimeContext(rng=np.random.RandomState(seed))
```
tristanls-tbp: Notice that the test is already resilient to unseeded random `np.random.RandomState()`
```suggestion
        self.ctx = RuntimeContext(rng=np.random.RandomState())
```

tests/unit/evidence_sdr_lm_test.py
```diff
@@ -36,7 +37,11 @@ class EvidenceSDRUnitTest(unittest.TestCase):
     def setUp(self):
         """Setup function at the beginning of each experiment."""
         # set seed for reproducibility
-        set_seed(42)
+        seed = 42
+        set_seed(seed)
+
+        # Generic RNG for tests that don't contain experiments
+        self.ctx = RuntimeContext(rng=np.random.RandomState(seed))
```
tristanls-tbp: **issue**: Please remove. This is unused.

tristanls-tbp:
One more thing you'll run into once you remove `rng` from goal state generator, learning modules, and sensor modules, you won't need to pass `rng` to the constructors anymore:

https://github.com/thousandbrainsproject/tbp.monty/blob/1513f8f68d94f8d4b464f51d4077e1a8873ff6a7/src/tbp/monty/frameworks/experiments/monty_experiment.py#L156

and 

https://github.com/thousandbrainsproject/tbp.monty/blob/1513f8f68d94f8d4b464f51d4077e1a8873ff6a7/src/tbp/monty/frameworks/experiments/monty_experiment.py#L166

src/tbp/monty/frameworks/models/salience/sensor_module.py
AgentRev: Yeah that's what I plan to do for now, I'm just trying to figure out if there's a way to prevent this in the future instead of manually fixing it every time, because it's not the first nor the last time, but so far I don't see any.

tests/unit/frameworks/models/salience/sensor_module_test.py
AgentRev: Yeah I wasn't thinking about `HabitatSalienceSM` in particular when I put that there, I just used the same snippet for all the test classes. A test that would need a "right seed" to pass would be a flaky, incorrect one. I usually always use fixed RNG seeds in unit tests (unless required otherwise, e.g. for testing some sort of statistical function), because I assume the worst, i.e. someone will inadvertently write a potentially-flaky test with it in the future; the fixed seed simply avoids flakiness.  However, I actually prefer moving forward over having the last word, so I'll just remove it 🙂

tests/unit/frameworks/models/salience/sensor_module_test.py
AgentRev: It's less about the generic part and more about its use-case, just to preemptively answer the classic "why does this exist again?" if someone asks in the future. But no biggie, I'll remove it.

tests/unit/frameworks/models/salience/sensor_module_test.py
tristanls-tbp: I thought I was letting you have the last word by saying that my complaint is outside the scope of this PR (since RNG is being seeded with 42 before this PR). I'll try to be clearer next time, apologies.

src/tbp/monty/frameworks/experiments/data_collection_experiments.py
```diff
@@ -101,14 +101,14 @@ def pre_episode(self):
...
-        self.model.pre_episode(self.rng)
+        self.model.pre_episode()
         self.env_interface.pre_episode(self.rng)
         self.max_steps = self.max_train_steps
         self.logger_handler.pre_episode(self.logger_args)
         if self.show_sensor_output:
             self.live_plotter.initialize_online_plotting()
 
-    def post_episode(self):
+    def post_episode(self, _):
```
tristanls-tbp: **issue**: I think I am miscommunicating. The `step, # noqa: ARG002` thing was good. I was confused by why parameters were added, but after you pointed out that `MontyExperiment` requires them, I was good with it and the `step, # noqa: ARG002`. As you mentioned, we want `# noqa: ARG002` over `_`.

src/tbp/monty/frameworks/models/graph_matching.py
```diff
@@ -628,19 +629,23 @@ def matching_step(self, observations):
...
 
         stats = self.collect_stats_to_save()
         self.buffer.update_stats(stats, append=self.has_detailed_logger)
 
-    def exploratory_step(self, observations):
+    def exploratory_step(
+        self,
+        ctx: RuntimeContext,  # noqa: ARG002
```
tristanls-tbp, line 646: **issue**: I don't think any implementation of `exploratory_step` currently uses `ctx`.

tests/unit/evidence_sdr_lm_test.py
```diff
@@ -36,7 +37,8 @@ class EvidenceSDRUnitTest(unittest.TestCase):
     def setUp(self):
         """Setup function at the beginning of each experiment."""
         # set seed for reproducibility
-        set_seed(42)
+        seed = 42
+        set_seed(seed)
```
tristanls-tbp: **issue (non-blocking)**: Variable assignment is now extraneous.

tristanls-tbp (commented):
Looking great, thank you for slogging through all these changes.

Just two more comments.

tests/unit/frameworks/models/salience/sensor_module_test.py
AgentRev: I wasn't sure if your statement meant "keep it as-is" or not, so I assumed the opposite to save time, since this particular point isn't critical anyway.

src/tbp/monty/frameworks/models/graph_matching.py
tristanls-tbp: **note**: Nevermind this.. I forgot about 
```
        for i in range(len(self.learning_modules)):
            sensory_inputs = self._collect_inputs_to_lm(i)
            getattr(self.learning_modules[i], self.step_type)(sensory_inputs)
```

src/tbp/monty/frameworks/experiments/data_collection_experiments.py
AgentRev: Oops that wasn't supposed to make it into the commit, I probably wrote that to poke the syntax highlighter and accidently saved it by mistake when I restarted the IDE. My WSL environment broke down yesterday because I ran out of RAM and it corrupted the git repo. I moved the code to a new clone outside WSL and pushed what I had so far, and I didn't have time to review everything before logging off for the night. I'll have to figure out a better WSL workflow that's out-of-RAM resilient...

src/tbp/monty/frameworks/models/abstract_monty_classes.py
AgentRev: Note that I ended up having to keep it as part of the `_rng` removals, which requires `GraphGoalStateGenerator.step()` to pass `ctx` down to `GraphGoalStateGenerator._check_need_new_output_goal()`.

tristanls-tbp: Yes, I saw that 👍.

tristanls-tbp:
Thank you, everything looks great. I'm gonna run benchmarks for sanity checks. I expect no benchmark changes, and will approve and merge once I verify that's the case.

tristanls-tbp:
I'm seeing only one benchmark change in `unsupervised_inference_distinctobj_dist_agent`. I'm gonna rerun to verify.

| Experiment | Metric | Baseline | Proposed | Change |
| base_config_10distinctobj_dist_agent | percent_correct (%) | 100 | 100 | 0 |
|  | used_mlh (%) | 1.43 | 1.43 | 0 |
|  | match_steps | 39 | 39 | 0 |
|  | rotation_error (deg) | 11.92 | 11.92 | 0 |
|  | runtime (min) | 2 | 3 | 1 |
|  | episode_runtime (sec) | 9 | 10 | 1 |
| base_config_10distinctobj_surf_agent | percent_correct (%) | 100 | 100 | 0 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 27 | 27 | 0 |
|  | rotation_error (deg) | 15.57 | 15.57 | 0 |
|  | runtime (min) | 3 | 2 | -1 |
|  | episode_runtime (sec) | 12 | 11 | -1 |
| randrot_noise_10distinctobj_dist_agent | percent_correct (%) | 99 | 99 | 0 |
|  | used_mlh (%) | 4 | 4 | 0 |
|  | match_steps | 44 | 44 | 0 |
|  | rotation_error (deg) | 21.57 | 21.57 | 0 |
|  | runtime (min) | 3 | 3 | 0 |
|  | episode_runtime (sec) | 16 | 16 | 0 |
| randrot_noise_10distinctobj_dist_on_distm | percent_correct (%) | 98 | 98 | 0 |
|  | used_mlh (%) | 5 | 5 | 0 |
|  | match_steps | 30 | 30 | 0 |
|  | rotation_error (deg) | 13.53 | 13.53 | 0 |
|  | runtime (min) | 3 | 2 | -1 |
|  | episode_runtime (sec) | 13 | 11 | -2 |
| randrot_noise_10distinctobj_surf_agent | percent_correct (%) | 98 | 98 | 0 |
|  | used_mlh (%) | 1 | 1 | 0 |
|  | match_steps | 28 | 28 | 0 |
|  | rotation_error (deg) | 19.99 | 19.99 | 0 |
|  | runtime (min) | 4 | 4 | 0 |
|  | episode_runtime (sec) | 25 | 22 | -3 |
| randrot_10distinctobj_surf_agent | percent_correct (%) | 100 | 100 | 0 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 28 | 28 | 0 |
|  | rotation_error (deg) | 15.06 | 15.06 | 0 |
|  | runtime (min) | 2 | 2 | 0 |
|  | episode_runtime (sec) | 11 | 11 | 0 |
| randrot_noise_10distinctobj_5lms_dist_agent | percent_correct (%) | 99 | 99 | 0 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 52 | 52 | 0 |
|  | rotation_error (deg) | 54.13 | 54.13 | 0 |
|  | runtime (min) | 5 | 5 | 0 |
|  | episode_runtime (sec) | 39 | 36 | -3 |
| base_10simobj_surf_agent | percent_correct (%) | 93.57 | 93.57 | 0 |
|  | used_mlh (%) | 10.71 | 10.71 | 0 |
|  | match_steps | 82 | 82 | 0 |
|  | rotation_error (deg) | 13.93 | 13.93 | 0 |
|  | runtime (min) | 6 | 6 | 0 |
|  | episode_runtime (sec) | 32 | 30 | -2 |
| randrot_noise_10simobj_dist_agent | percent_correct (%) | 81 | 81 | 0 |
|  | used_mlh (%) | 38 | 38 | 0 |
|  | match_steps | 211 | 211 | 0 |
|  | rotation_error (deg) | 33.62 | 33.62 | 0 |
|  | runtime (min) | 9 | 9 | 0 |
|  | episode_runtime (sec) | 76 | 71 | -5 |
| randrot_noise_10simobj_surf_agent | percent_correct (%) | 83 | 83 | 0 |
|  | used_mlh (%) | 33 | 33 | 0 |
|  | match_steps | 166 | 166 | 0 |
|  | rotation_error (deg) | 31.98 | 31.98 | 0 |
|  | runtime (min) | 18 | 17 | -1 |
|  | episode_runtime (sec) | 144 | 135 | -9 |
| randomrot_rawnoise_10distinctobj_surf_agent | percent_correct (%) | 71 | 71 | 0 |
|  | used_mlh (%) | 74 | 74 | 0 |
|  | match_steps | 14 | 14 | 0 |
|  | rotation_error (deg) | 96.27 | 96.27 | 0 |
|  | runtime (min) | 4 | 4 | 0 |
|  | episode_runtime (sec) | 6 | 6 | 0 |
| base_10multi_distinctobj_dist_agent | percent_correct (%) | 62.14 | 62.14 | 0 |
|  | used_mlh (%) | 4.29 | 4.29 | 0 |
|  | match_steps | 36 | 36 | 0 |
|  | rotation_error (deg) | 33.02 | 33.02 | 0 |
|  | runtime (min) | 35 | 4 | -31 |
|  | episode_runtime (sec) | 2 | 2 | 0 |
| base_77obj_dist_agent | percent_correct (%) | 93.07 | 93.07 | 0 |
|  | used_mlh (%) | 15.58 | 15.58 | 0 |
|  | match_steps | 109 | 109 | 0 |
|  | rotation_error (deg) | 16.93 | 16.93 | 0 |
|  | runtime (min) | 20 | 21 | 1 |
|  | episode_runtime (sec) | 61 | 63 | 2 |
| base_77obj_surf_agent | percent_correct (%) | 99.57 | 99.57 | 0 |
|  | used_mlh (%) | 7.36 | 7.36 | 0 |
|  | match_steps | 57 | 57 | 0 |
|  | rotation_error (deg) | 8.75 | 8.75 | 0 |
|  | runtime (min) | 14 | 15 | 1 |
|  | episode_runtime (sec) | 37 | 40 | 3 |
| randrot_noise_77obj_dist_agent | percent_correct (%) | 87.01 | 87.01 | 0 |
|  | used_mlh (%) | 28.14 | 28.14 | 0 |
|  | match_steps | 158 | 158 | 0 |
|  | rotation_error (deg) | 32.66 | 32.66 | 0 |
|  | runtime (min) | 35 | 35 | 0 |
|  | episode_runtime (sec) | 109 | 108 | -1 |
| randrot_noise_77obj_surf_agent | percent_correct (%) | 90.48 | 90.48 | 0 |
|  | used_mlh (%) | 23.81 | 23.81 | 0 |
|  | match_steps | 111 | 111 | 0 |
|  | rotation_error (deg) | 31.05 | 31.05 | 0 |
|  | runtime (min) | 38 | 36 | -2 |
|  | episode_runtime (sec) | 123 | 114 | -9 |
| randrot_noise_77obj_5lms_dist_agent | percent_correct (%) | 92.21 | 92.21 | 0 |
|  | used_mlh (%) | 1.3 | 1.3 | 0 |
|  | match_steps | 66 | 66 | 0 |
|  | rotation_error (deg) | 57.63 | 57.63 | 0 |
|  | runtime (min) | 12 | 12 | 0 |
|  | episode_runtime (sec) | 110 | 115 | 5 |
| unsupervised_inference_distinctobj_dist_agent | percent_correct (%) | 99 | 98 | -1 |
|  | match_steps | 96 | 98 | 2 |
|  | runtime (min) | 9 | 8 | -1 |
|  | episode_runtime (sec) | 5 | 5 | 0 |
| unsupervised_inference_distinctobj_surf_agent | percent_correct (%) | 100 | 100 | 0 |
|  | match_steps | 99 | 100 | 1 |
|  | runtime (min) | 17 | 17 | 0 |
|  | episode_runtime (sec) | 10 | 10 | 0 |
| infer_comp_lvl1_with_monolithic_models | correct_child_or_parent (%) | 60.71 | 60.71 | 0 |
|  | used_mlh (%) | 53.57 | 53.57 | 0 |
|  | match_steps | 321 | 321 | 0 |
|  | rotation_error (deg) | 80.71 | 80.71 | 0 |
|  | avg_prediction_error | 0.38 | 0.38 | 0 |
|  | runtime (min) | 29 | 27 | -2 |
| infer_comp_lvl1_with_comp_models | correct_child_or_parent (%) | 84.52 | 84.52 | 0 |
|  | used_mlh (%) | 41.67 | 41.67 | 0 |
|  | match_steps | 44 | 44 | 0 |
|  | rotation_error (deg) | 40.76 | 40.76 | 0 |
|  | avg_prediction_error | 0.33 | 0.33 | 0 |
|  | runtime (min) | 3 | 3 | 0 |
| infer_comp_lvl2_with_comp_models | correct_child_or_parent (%) | 79.05 | 79.05 | 0 |
|  | used_mlh (%) | 40.48 | 40.48 | 0 |
|  | match_steps | 37 | 37 | 0 |
|  | rotation_error (deg) | 44.91 | 44.91 | 0 |
|  | avg_prediction_error | 0.34 | 0.34 | 0 |
|  | runtime (min) | 10 | 10 | 0 |
| infer_comp_lvl3_with_comp_models | correct_child_or_parent (%) | 56.57 | 56.57 | 0 |
|  | used_mlh (%) | 56.57 | 56.57 | 0 |
|  | match_steps | 40 | 40 | 0 |
|  | rotation_error (deg) | 41.5 | 41.5 | 0 |
|  | avg_prediction_error | 0.34 | 0.34 | 0 |
|  | runtime (min) | 17 | 19 | 2 |
| infer_comp_lvl4_with_comp_models | correct_child_or_parent (%) | 58.24 | 58.24 | 0 |
|  | used_mlh (%) | 56.59 | 56.59 | 0 |
|  | match_steps | 33 | 33 | 0 |
|  | rotation_error (deg) | 42.52 | 42.52 | 0 |
|  | avg_prediction_error | 0.33 | 0.33 | 0 |
|  | runtime (min) | 17 | 18 | 1 |

tristanls-tbp:
@AgentRev, I duplicated the result where `unsupervised_inference_distinctobj_dist_agent` changed. 

**suggestion**: While you don't have access to our benchmark runs, you should see a difference between `unsupervised_inference_distinctobj_dist_agent` when you run it from `main` vs when you run it from this branch. Any help figuring out why they're different is appreciated.

AgentRev:
Narrowed it down to commit https://github.com/thousandbrainsproject/tbp.monty/commit/a26661c6ae2c9d347afde1312eb6155efe335da0, will keep investigating

AgentRev:
Just to make sure we're on the same page, once I pinpoint the cause, is the goal re-adjusting the code to make the benchmark achieve the same result as before? Or, are minor variations in post-refactoring benchmarks sometimes expected and we just want to ensure nothing critical was broken?

tristanls-tbp:
> Just to make sure we're on the same page, once I pinpoint the cause, is the goal re-adjusting the code to make the benchmark achieve the same result as before? Or, are minor variations in post-refactoring benchmarks sometimes expected and we just want to ensure nothing critical was broken?

@AgentRev, well, it's hard to answer that before we know the cause 🙂. If it's a bug, we shouldn't add bugs. If it's something reasonable, then OK, benchmarks are different now. To me, the goal is to first understand what caused the change.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -69,12 +69,10 @@ def __init__(self, *args, **kwargs):
...
         self.init_pre_episode = False
 
-    def pre_episode(
-        self, rng: np.random.RandomState, primary_target, semantic_id_to_label=None
-    ) -> None:
+    def pre_episode(self, primary_target, semantic_id_to_label=None) -> None:
         if not self.init_pre_episode:
             self.init_pre_episode = True
-            return super().pre_episode(rng, primary_target, semantic_id_to_label)
+            return super().pre_episode(primary_target, semantic_id_to_label)
```
AgentRev, line 75: After 90 benchmarks of `unsupervised_inference_distinctobj_dist_agent`, I think I found the damn bug. Or rather, its corpse.

For all episodes beyond the first, the `if not self.init_pre_episode` logic here prevented `reset_episode_rng()`'s new RNGs from propagating via `pre_episode()` to modules belonging to `MontyForNoResetEvidenceGraphMatching`. Therefore, modules were stuck with the same RNG for all episodes.

So, the bug is actually on the `main` branch, and the `ctx` changes in this PR happened to fix it. The correct RNGs are now passed to modules, which slightly altered unsupervised benchmarks, since they are the only ones using `MontyForNoResetEvidenceGraphMatching`.

tristanls-tbp: I concur with your assessment. Also, I just noticed that `unsupervised_inference_distinctobj_surf_agent` is also affected; its `match_steps` count went from 99 to 100.

So, this was a third option to your question. You fixed a bug :)

tristanls-tbp:
@AgentRev, thank you for figuring out what changed 🙌.

When you get a chance, please resolve the conflicts, and I'll make a benchmark update commit.

tristanls-tbp (approved):
Thanks again for this large effort @AgentRev. It's a big step in updating Monty's architecture.

AgentRev:
So, what's next?

tristanls-tbp:
@AgentRev, we noticed that experiment mode is unused in MotorSystem and MotorPolicies. Please remove. Injections `#52` and `#53`. Benchmarks should be unaffected.

Screenshot 2026-02-17 at 13 49 41

tristanls-tbp:
@AgentRev, alternatively, there's also the whole thing about SensorIDs not being treated as opaque. #791 partially addressed this, but there's still more work as indicated on the [Future Reality Tree](https://link.excalidraw.com/readonly/1zkoPDpkfKlbyGfCaLsY) (search for "300"). 


Screenshot 2026-02-17 at 17 19 43

There are also other near-adjacent updates, but it might be difficult to work on, as we (TBP team) are going to change policy APIs and hierarchy soon.

Of course, these are all suggestions. What's interesting to you? We can figure out how to coordinate around that work, or I can perhaps provide more context in a specific area that sounds more compelling. Either way, we always appreciate the help.

AgentRev:
My goal is to grasp the codebase at a wider and deeper level, to allow my effectiveness to progressively scale up. Therefore, anything goes; I don't really have preferences at the moment, other than tasks that benefit your timelines directly. So, alright, I'll deal with 52 and 53.

PR #757, opened 2026-02-03
test: extract HYDRA_ROOT from test setup

jeremyshoemaker:
In test for a separate change, I ran into a situation where I want to take the path where the Hydra configs are stored and append more path segments to dig down to a part of the configuration. This lead to a long string starting with `../../../../` because of where the test was located.

This change sets HYDRA_ROOT in the top level `test/__init__.py` so that the tests can all use that instead. It also uses a Path object instead of a raw string (though Hydra expects a string, so we have to convert to that in the initialize calls).

We're also changing the `hydra.initialize` call to `hydra.initialize_config_dir` because the former only takes relative paths, and we want to pass an absolute path.

### DIFF
--- a/src/tbp/monty/frameworks/environment_utils/graph_utils.py
+++ b/src/tbp/monty/frameworks/environment_utils/graph_utils.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Thousand Brains Project
+# Copyright 2025-2026 Thousand Brains Project
 # Copyright 2022-2024 Numenta Inc.
 #
 # Copyright may exist in Contributors' modifications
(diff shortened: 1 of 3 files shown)


jeremyshoemaker:
The `GetGoodViewTest::test_multi_object_target_not_visible` test appears to be flaky. I just sat and ran it about a dozen times locally, and it failed twice with this error message.

```
>               assert perc_on_target_obj >= target_perc_on_target_obj, (
                    f"Initial view is not good enough, {perc_on_target_obj} "
                    f"vs target of {target_perc_on_target_obj}"
                )
E               AssertionError: Initial view is not good enough, 0.0 vs target of 0.5
E               assert 0.0 >= 0.5
```

It does it in `main` so it's not my changes causing problems.

jeremyshoemaker:
> **suggestion (non-blocking)**: Maybe `CONF_ROOT` instead? After seeing it over and over again, it occurred to me that "hydra" doesn't really mean anything.

It's only ever used in the context of Hydra configs is why I chose HYDRA_ROOT. CONF_ROOT would also work, but it's not that big a difference.

PR #758, opened 2026-02-03
refactor!: make MotorPolicy take ActionSampler 

jeremyshoemaker:
Currently, we use the `*_class`, `*_args` pattern in the BasePolicy to create the ActionSampler in the constructor. This changes BasePolicy to take an instance of ActionSampler, and updates the configurations to have Hydra instantiate the ActionSampler for us and inject it into the constructor arguments for the policy.

The vast majority of this PR is just updating the existing tests and snapshots. That's in a separate commit to allow easier review of the actual changes.

### DIFF
--- a/src/tbp/monty/frameworks/models/salience/return_inhibitor.py
+++ b/src/tbp/monty/frameworks/models/salience/return_inhibitor.py
@@ -8,8 +8,6 @@
 # https://opensource.org/licenses/MIT.
 from __future__ import annotations
 
-from typing import Any
-
 import numpy as np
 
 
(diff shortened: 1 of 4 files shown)


jeremyshoemaker:
Actually, the split isn't as clean as I thought it was across those two commits. The second one contains all the changes to the configs that aren't in the first one. The first one was very focused on the code changes, and I forgot to include **all** the config changes there.

jeremyshoemaker:
I'm also running the benchmarks to confirm nothing changed unexpectedly.

jeremyshoemaker:
> I'm also running the benchmarks to confirm nothing changed unexpectedly.

No changes to benchmark results.

tests/unit/frameworks/models/motor_system/motor_system_config_test.py
tristanls-tbp, line 1: **issue**: Do we need to commit these? To me, it feels these tests served their purpose. I don't think we need to commit code that tracks that, at some point in the past, we were using `_class` + `_args` pattern.

tristanls-tbp (commented):
🙌 .

I do wonder though if we need to commit the test...

tests/unit/frameworks/models/motor_system/motor_system_config_test.py
jeremyshoemaker: I was wondering that too. I expect to need to update these when I move up the tree, so they might have some purpose still, but ultimately, we'd just be testing that Hydra works in the end. At that point, it makes sense to remove them.

PR #761, opened 2026-02-04
refactor!: make EvidenceGraphLM take GSG instance

jeremyshoemaker:
Part of our efforts to refactor from using `*_class` and `*_args` to using dependency injection and relying on Hydra to instantiate our objects. This change alters how GoalStateGenerators are created and passed to their parent LMs.

In a separate commit, I'm removing the `clear_gsg_args.yaml` file (I had first renamed it to `clear_gsg.yaml`) since it looks like based on the snapshot tests, we don't actually need it.

### DIFF
--- a/src/tbp/monty/frameworks/models/goal_state_generation.py
+++ b/src/tbp/monty/frameworks/models/goal_state_generation.py
@@ -10,7 +10,7 @@
 from __future__ import annotations
 
 import logging
-import typing
+from typing import TYPE_CHECKING
 
 import numpy as np
 
@@ -20,7 +20,7 @@ from tbp.monty.frameworks.models.abstract_monty_classes import (
 from tbp.monty.frameworks.models.states import GoalState
 from tbp.monty.frameworks.utils.communication_utils import get_state_from_channel
 
-if typing.TYPE_CHECKING:
+if TYPE_CHECKING:
     from tbp.monty.frameworks.models.graph_matching import GraphLM
 
 __all__ = [


src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
```diff
@@ -173,8 +172,7 @@ def __init__(
         max_nodes_per_graph=2000,
         num_model_voxels_per_dim=50,  # -> voxel size = 6mm3 (0.006)
         use_multithreading=True,
-        gsg_class=EvidenceGoalStateGenerator,
-        gsg_args=None,
+        gsg: EvidenceGoalStateGenerator | None = None,
```
jeremyshoemaker, line 175: This type might be too narrow. I don't think this ever expects something other than an `EvidenceGoalStateGenerator`, though.

tests/unit/evidence_lm_test.py
```diff
@@ -869,8 +867,7 @@ def test_hypothesis_testing_proposal_for_id(self):
             self.fake_obs_house,
             initial_possible_poses=[[0, 0, 0]],  # Note we isolate the influence of
             # ambiguous pose on the hypothesis testing
-            gsg_class=EvidenceGoalStateGenerator,
-            gsg_args=self.default_gsg_config,
+            gsg=EvidenceGoalStateGenerator(**self.default_gsg_config),
```
jeremyshoemaker, line 870: I thought about just creating one of these up in the test `setUp` method, but using it modifies it, so I don't think that would be safe.

src/tbp/monty/frameworks/models/evidence_matching/learning_module.py
tristanls-tbp: Sounds right for an `EvidenceLM`.

tristanls-tbp (approved):
Looks good.

**question**: Did you run a sample of benchmarks to ensure they work? I know we now test tutorials and have more confidence, but a sanity check against a benchmark would help. I don't think there's anything wrong, but that's the reason for a sanity check 🙂.

jeremyshoemaker:
> **question**: Did you run a sample of benchmarks to ensure they work? I know we now test tutorials and have more confidence, but a sanity check against a benchmark would help. I don't think there's anything wrong, but that's the reason for a sanity check 🙂.

I was debating whether to or not, since the snapshot configs didn't change in unexpected ways, but since you mentioned it I'll go ahead and do it.

tristanls-tbp:
@jeremyshoemaker I'm thinking more of the `parent_lm` and `gsg` setting and `reset()` sequencing. It looks right, but maybe we missed some interaction.

jeremyshoemaker:
> I was debating whether to or not, since the snapshot configs didn't change in unexpected ways, but since you mentioned it I'll go ahead and do it.

Benchmarks match expected values.

PR #744, opened 2026-01-28
refactor!: PredefinedPolicy just like any other

tristanls-tbp:
This pull request removes the `predefined_call` from the `MotorPolicy` API. Instead, a `PredefinedPolicy` is added that replaces the functionality.

A large number of file changes is due to not needing to specify the default `file_name` and related parameters for every policy, as these were inherited from the base predefined implementation.

PR #762, opened 2026-02-04
refactor!: MotorSystem and MotorPolicy access RNG through RuntimeContext

tristanls-tbp:
This pull request updates `MotorSystem` and `MotorPolicy` to access the random number generator through `RuntimeContext`. Currently, the context is created in environment `__next__` methods, but follow-on work will eliminate `__next__` methods and pass the runtime context from the experiment directly.

Because RNG is passed via `RuntimeContext`, `MotorSystem` and `MotorPolicy` no longer maintain `self.rng` and do not need RNG on initialization. As a consequence, their `pre_episode` methods no longer accept RNG since `self.rng` is no longer being maintained at this level in the framework.

`self.action` in `BasePolicy` is removed and initialized only in `InformedPolicy`. I did this to avoid needing RNG in the `BasePolicy` constructor via `get_random_action(ctx)` and validated with @scottcanoe that it is _unused_ by `BasePolicy` in our work on extracting `PredefinedPolicy` in #744.

## Benchmarks

No changes.

### DIFF
--- a/benchmarks/logos_on_objects.csv
+++ b/benchmarks/logos_on_objects.csv
@@ -1,6 +1,6 @@
 Experiment,Correct Child or Parent (%)|align right,Used MLH (%)|align right,Num Match Steps|align right,Rotation Error (degrees)|align right,Avg Prediction Error|align right,Run Time (mins)|align right
-infer_comp_lvl1_with_monolithic_models,60.71,53.57,321,80.71,0.38,28
+infer_comp_lvl1_with_monolithic_models,60.71,53.57,321,80.71,0.38,29
 infer_comp_lvl1_with_comp_models,84.52,41.67,44,40.76,0.33,3
-infer_comp_lvl2_with_comp_models,79.05,40.48,37,44.91,0.34,9
-infer_comp_lvl3_with_comp_models,56.57,56.57,40,41.50,0.34,19
-infer_comp_lvl4_with_comp_models,58.24,56.59,33,42.52,0.33,18
+infer_comp_lvl2_with_comp_models,79.05,40.48,37,44.91,0.34,10
+infer_comp_lvl3_with_comp_models,56.57,56.57,40,41.50,0.34,17
+infer_comp_lvl4_with_comp_models,58.24,56.59,33,42.52,0.33,17
--- a/benchmarks/ycb_10objs.csv
+++ b/benchmarks/ycb_10objs.csv
@@ -1,13 +1,13 @@
 Experiment,Correct (%)|align right,Used MLH (%)|align right,Num Match Steps|align right,Rotation Error (degrees)|align right,Run Time (mins)|align right,Episode Run Time (s)|align right
-base_config_10distinctobj_dist_agent,100.00,1.43,39,11.92,2,10
+base_config_10distinctobj_dist_agent,100.00,1.43,39,11.92,2,9
 base_config_10distinctobj_surf_agent,100.00,0.00,27,15.57,3,12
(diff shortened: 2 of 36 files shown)


jeremyshoemaker:
Haven't gotten to reviewing yet, but just as a comparison point, the benchmarks I ran for #761 last night matched the expected values on that benchmark. It's possible I missed a difference somewhere else since it's a manual spot check, but for that one, it matches.

jeremyshoemaker:
> Haven't gotten to reviewing yet, but just as a comparison point, the benchmarks I ran for https://github.com/thousandbrainsproject/tbp.monty/pull/761 last night matched the expected values on that benchmark. It's possible I missed a difference somewhere else since it's a manual spot check, but for that one, it matches.

Having reviewed the code, I'm not sure what could be causing the difference.

tristanls-tbp:
@jeremyshoemaker the only thing I can think of is the removal of call to `self.get_random_action()` in `BasePolicy` constructor: 

Screenshot 2026-02-05 at 10 30 09

Which did invoke RNG

Screenshot 2026-02-05 at 10 30 42

And my guess is that this somehow matters only on that one benchmark? 🤔

This is the motor config of `base_10multi_distinctobj_dist_agent`:
```yaml
      motor_system_config:
        motor_system_args:
          policy_args:
            file_name: null
            good_view_percentage: 0.5
            desired_object_distance: 0.03
            use_goal_state_driven_actions: true
            min_perc_on_obj: 0.25
            agent_id: ${monty.agent_id:agent_id_0}
            action_sampler:
              actions:
              - ${monty.class:tbp.monty.frameworks.actions.actions.LookUp}
              - ${monty.class:tbp.monty.frameworks.actions.actions.LookDown}
              - ${monty.class:tbp.monty.frameworks.actions.actions.TurnLeft}
              - ${monty.class:tbp.monty.frameworks.actions.actions.TurnRight}
              _target_: tbp.monty.frameworks.actions.action_samplers.ConstantSampler
              rotation_degrees: 10.0
          policy_class: ${monty.class:tbp.monty.frameworks.models.motor_policies.InformedPolicy}
        motor_system_class: ${monty.class:tbp.monty.frameworks.models.motor_system.MotorSystem}
```
Which is pretty much the standard distant-agent setup. The only difference is that in `base_10multi_distinctobj_dist_agent`, the `rotation_degrees: 10.0`, whereas other configurations are `rotation_degrees: 5.0`.

But this doesn't explain why only one benchmark is affected.

tristanls-tbp:
I created #764 to isolate the RNG sequencing change and confirm whether that's the cause of the discrepancy.

jeremyshoemaker:
> @jeremyshoemaker the only thing I can think of is the removal of call to `self.get_random_action()` in `BasePolicy` constructor:
> Which did invoke RNG

Oh, yeah that would definitely changes things. Even if we weren't using the value, we're pulling it out of the sequence, so anything afterward would be different.

> But this doesn't explain why only one benchmark is affected.

Well, didn't it affect the runtime of a few of the others, and I thought I saw one other stat different?

tristanls-tbp:
> Well, didn't it affect the runtime of a few of the others, and I thought I saw one other stat different?

I pretty much ignore runtime changes. They change pretty much on every single run. But yeah, there was one other benchmark change.

Screenshot 2026-02-05 at 12 17 14

tristanls-tbp:
I confirmed in #764 that the benchmark change was due to RNG sequencing. I'll rerun the benchmarks after that PR is merged, and we should see no changes then.

PR #893, opened 2026-04-03
test: Use property-based testing for GaussianBlurRGB

hlee9212:
1. Modified `test_blur_solid_image_returns_identical` to test for any solid image.
2. Divided up `test_blur_modifies_rgb_preserves_alpha` into two tests:
    - `test_blur_preserves_alpha` to make sure the alpha channel doesn't change with any examples drawn from `rgba_and_blur_params()`
    - `test_blur_reduces_total_variation` to make sure that [Total Variation](https://en.wikipedia.org/wiki/Total_variation) after blurring is <= before blurring.


tests/unit/frameworks/environment_utils/transforms_test.py
```diff
@@ -28,6 +31,29 @@
 AGENT_ID = AgentID("0")
 SENSOR_ID = SensorID("0")
 
+BLUR_KERNEL_SIZES = [0, 1, 3, 5, 7, 9, 11, 13, 15]
```
jeremyshoemaker: What's defining these values? If it's just odd values up to 15, this could be written with a list comprehension and I think that would be a little clearer (the `range` makes that a little harder not being inclusive, hence the extra constant).

```suggestion
MAX_KERNEL = 15
BLUR_KERNEL_SIZES = [n for n in range(MAX_KERNEL + 1) if n == 0 or n % 2 == 1]
```

tests/unit/frameworks/environment_utils/transforms_test.py
```diff
@@ -160,48 +186,69 @@ def test_rgba_not_in_sensor_observations(self):
...
-        self.assertEqual(smoothed_img.shape, rgba_img.shape)
-        np.testing.assert_allclose(smoothed_img[:, :, :3], expected_rgb, atol=1e-5)
-        np.testing.assert_array_equal(smoothed_img[:, :, 3], 255.0)
+        obs[AGENT_ID][SENSOR_ID] = SensorObservation({"rgba": rgba.copy()})
+        gaussian_smoother = GaussianBlurRGB(
+            agent_id=AGENT_ID, sigma=sigma, kernel_size=kernel_size
+        )
+        result_rgb = gaussian_smoother(obs, _ctx=None)[AGENT_ID][SENSOR_ID]["rgba"][
+            :, :, :3
+        ]
```
jeremyshoemaker: I'm not a bit fan of splitting indexing operations across lines like this. It might be clearer written this way.

```suggestion
        result_rgba = gaussian_smoother(obs, _ctx=None)[AGENT_ID][SENSOR_ID]["rgba"] 
        result_rgb = result_rgba[:, :, :3]
```

tests/unit/frameworks/environment_utils/transforms_test.py
jeremyshoemaker, line 256: I missed this in the original PR, but we don't need this since we're using `pytest`.
```suggestion
```

jeremyshoemaker (changes requested):
This is much better than the original. I like how you separated the alpha channel and the RGB properties into two tests.

I have some suggestions, they aren't blockers, but I want your thoughts on them before approving.

tests/unit/frameworks/environment_utils/transforms_test.py
```diff
@@ -160,48 +186,69 @@ def test_rgba_not_in_sensor_observations(self):
...
         obs = Observations()
         obs[AGENT_ID] = AgentObservations()
         obs[AGENT_ID][SENSOR_ID] = SensorObservation({"rgba": rgba_img.copy()})
-        gaussian_smoother = GaussianBlurRGB(agent_id=AGENT_ID, sigma=2, kernel_size=5)
-        result = gaussian_smoother(obs, _ctx=None)
-        result_img = result[AGENT_ID][SENSOR_ID]["rgba"]
+        gaussian_smoother = GaussianBlurRGB(
+            agent_id=AGENT_ID, sigma=sigma, kernel_size=kernel_size
+        )
+        result_img = gaussian_smoother(obs, _ctx=None)[AGENT_ID][SENSOR_ID]["rgba"]
```
jeremyshoemaker: I almost forgot. @tristanls-tbp pointed this out after we merged this in the first time.

For the `GaussianBlurRGB.__call__` method, we can't name the second argument `_ctx` because it's called `ctx` in the protocol, and that name change breaks compatibility. What we should do instead is use the `# noqa: ARG002` comment on the argument in the `__call__` method to silence the lint about it not being used.

Then here, where we use it, since the type of `ctx` is not optional, we can't use `None` here, but instead we should use a mock, e.g. `Mock()`.

tests/unit/frameworks/environment_utils/transforms_test.py
jeremyshoemaker: We can leave this for now, since we're having a discussion about it elsewhere.

jeremyshoemaker (approved):
Looks good now.

PR #898, opened 2026-04-07
feat: add LookAtGoal policy

scottcanoe:
This PR adds the `LookAtGoal` motor policy. It's purpose is to saccade to a goal location. This is in contrast to `InformedPolicy` which executes "jumps" (i.e., full agent/sensor repositioning) to achieve perceptual goals.

This policy is usable as-is, but it is has not yet been folded into any benchmark configs. For context, it was designed to specifically to execute goals generated by `SalienceSM`, and it will raise a `RuntimeError` if it does receive a goal. Therefore, it is currently not viable as the sole policy if it does not receive goals at every step (which only the `SalienceSM` is capable of generating).


src/tbp/monty/frameworks/models/salience/motor_policy.py
```diff
@@ -0,0 +1,143 @@
...
+        sensor_pitch = target_pitch_rel_agent - sensor_pitch_rel_agent
+
+        # Create actions to return to the the motor system.
+        actions = [
+            TurnLeft(agent_id=self._agent_id, rotation_degrees=np.degrees(agent_yaw)),
+            LookUp(agent_id=self._agent_id, rotation_degrees=np.degrees(sensor_pitch)),
+        ]
+
+        # For logging purposes only.
+        goal.info["attempted"] = True
```
tristanls-tbp: **question**: What sort of logging is this for? Any reason not to log here `logger.info("attempted...")`?

**issue**: This causes a side effect on an input that's not very obvious.

tristanls-tbp: Also, `goal.info` could be `None`.

tristanls-tbp (approved):
Let there be things to look at!

PR #899, opened 2026-04-07
refactor: do not step salience or return inhibitor for motor-only

scottcanoe:
This PR prevents the `SalienceSM` from performing salience computations or stepping its return inhibitor during `motor_only` steps. While motor-only steps are mostly generated for surface-agent configurations which do not use the `SalienceSM`, I wanted to guard against unexpectedly (and erroneously) modifying the return inhibitor's decay field during motor-only steps.

I also updated `SalienceSM.step`'s return signature. It only ever returns `None`, so I updated it and removed explicit return of `None` in accordance with ruff's rule on the matter.

Note that motor-only steps will still result in telemetry being generated for `SalienceSM` since that's how `CameraSM` works.

### DIFF
--- a/src/tbp/monty/frameworks/models/salience/sensor_module.py
+++ b/src/tbp/monty/frameworks/models/salience/sensor_module.py
@@ -11,7 +11,7 @@ from __future__ import annotations
 import numpy as np
 import quaternion as qt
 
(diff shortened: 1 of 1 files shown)


tristanls-tbp (approved):
👍

PR #909, opened 2026-04-10
fix: live plotter search for agent_id in observations

tristanls-tbp:
This makes the live plotter display again when `experiment.config.show_sensor_output: true`.

### DIFF
--- a/src/tbp/monty/frameworks/utils/live_plotter.py
+++ b/src/tbp/monty/frameworks/utils/live_plotter.py
@@ -6,10 +6,14 @@
 # Use of this source code is governed by the MIT
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
+from __future__ import annotations
 
 import matplotlib.pyplot as plt
 import numpy as np
 
+from tbp.monty.frameworks.agents import AgentID
+from tbp.monty.frameworks.models.abstract_monty_classes import Monty, Observations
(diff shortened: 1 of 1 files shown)


src/tbp/monty/frameworks/utils/live_plotter.py
```diff
@@ -41,7 +45,7 @@ def initialize_online_plotting(self):
         self.setup_sensor_ax()
         self.setup_mlh_ax()
 
-    def hardcoded_assumptions(self, observation, model):
+    def hardcoded_assumptions(self, observation: Observations, model: Monty):
```
jeremyshoemaker, line 48: Return type?

jeremyshoemaker (approved):
Looks good. I like that it's looking up the AgentID rather than just assuming it. One less assumption.

src/tbp/monty/frameworks/utils/live_plotter.py
tristanls-tbp: This will require me to type `current_mlh`, which in turn will require me to update `EvidenceGraphLM`. 

Since you approved the PR, I will interpret this as a non-blocking suggestion and defer it for now.

jeremyshoemaker: It's a function called `hardcoded_assumptions`, I kinda hope it goes away some day.

PR #910, opened 2026-04-10
test: fix failing test_blur_reduces_tv

hlee9212:
This fixes the failing test that @ramyamounir has noticed. 

Reason for failing: 
1. Our input was `np.float64` while `cv2` computes convolution (the Gaussian Blur) in `float32`. It's not really appropriate to upscale the input precision.
2. (Speculative) The `cv2` binaries between macOS and Ubuntu are likely different, hence why it failed on Github Actions and not locally. This may affect floating point error accumulation. 

How I fixed it:
1. The failed cases were all when the input image were solid (i.e. same values). Since this case is already covered by `test_blur_solid_image_returns_identical`, I used `assume` to filter these out.

### DIFF
--- a/tests/unit/frameworks/environment_utils/transforms_test.py
+++ b/tests/unit/frameworks/environment_utils/transforms_test.py
@@ -239,7 +239,7 @@ class GaussianBlurRGBTest(unittest.TestCase):
             )
 
         input_tv = total_variation(rgba[:, :, :3])
-        assume(input_tv > 0.0) # Exclude solid images
+        assume(input_tv > 0.0)  # Exclude solid images
 
         obs = Observations()
         obs[AGENT_ID] = AgentObservations()
--- a/tests/unit/frameworks/environment_utils/transforms_test.py
+++ b/tests/unit/frameworks/environment_utils/transforms_test.py
@@ -13,7 +13,7 @@ from unittest.mock import Mock
 
 import numpy as np
 import pytest
-from hypothesis import given
+from hypothesis import assume, given
 from hypothesis import strategies as st
 from hypothesis.extra.numpy import arrays
 
@@ -233,10 +233,13 @@ class GaussianBlurRGBTest(unittest.TestCase):
         rgba, sigma, kernel_size = params
 
         def total_variation(img):
-            img = img.astype(np.float64)
+            img = img.astype(np.float32)
             return np.sum(np.abs(np.diff(img, axis=0))) + np.sum(
                 np.abs(np.diff(img, axis=1))
             )
+
+        input_tv = total_variation(rgba[:, :, :3])
+        assume(input_tv > 0.0) # Exclude solid images
 
         obs = Observations()
         obs[AGENT_ID] = AgentObservations()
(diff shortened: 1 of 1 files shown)


jeremyshoemaker (commented):
Is this going to be a problem in actual use? Do we use float32 everywhere?

hlee9212:
@jeremyshoemaker A quick text search for `float64` reveals we are using this in several places. ~~However, I'd have to ask the team if this much precision is required or if everyone is okay using the slightly imprecise `float32`.~~

My general rule of thumb for when to use `float64` vs. `float32`. Writing this out because my background in ML/DL may bias me towards `float32`, so trying to be objective:

1. In context of Monty, `float32` would be fine for things like sensor data as sensor datas are generally noisy and we can't actually get very high input precision. 
2. `float64` earns its place when we are:
- Summing over large arrays
- Computing small differences in large numbers
- Iterative algorithms (like the various `step()` methods we have)
- I just learned yesterday that `scipy.transform.Rotation.magnitude()` returns `float64`, and that's not something we can change. 

I saw something like `self.evidence[graph_id] = np.ma.sum([...])` where evidence is summed across every step of an episode which may warrant `float64`. However, if we are choosing the most likely hypothesis based on the thinnest margin, maybe that's not right either. 🤣

TL;DR: 
1. Things related to Sensor Module: `float32` okay because world/data is noisy. 
2. Things related to LM/Motor: `float64` for now based on the math we are doing, but I think `float32` is okay in the future on the reason that our brain isn't that precise and our motor commands are not that precise. If it were I think I'd have better hand-eye coordination.

jeremyshoemaker:
> @jeremyshoemaker A quick text search for `float64` reveals we are using this in several places. However, I'd have to ask the team if this much precision is required or if everyone is okay using the slightly imprecise `float32`.

Maybe that's a discussion that can wait until we start using this. I don't want to hold up fixing this flaky test.

jeremyshoemaker (approved):

hlee9212:
@jeremyshoemaker Yes. (Also I have updated my comment above on when `float32` and `float64` are used.

Since this test is using `float32` for images coming from sensor, it is completely fine.

tristanls-tbp:
**question**: Where is this test failing? It didn't fail in our GitHub Actions.

hlee9212:
@tristanls-tbp - Ramy posted on [Slack](https://thousandbrainsproject.slack.com/archives/C0842EPEMHV/p1775836301899449?thread_ts=1775510020.525919&cid=C0842EPEMHV) that it was failing for his PR. This test seemed to occasionally fail previously. 

Tagging @jeremyshoemaker in case he wants to add anything.

jeremyshoemaker:
One of the downsides of `hypothesis` is that because it generates new values each time you run it unless you explicitly add `@example`'s, it can lead to flaky tests if you missed something. On the plus side, it gives you an example to recreate it. Unfortunately in this case, we've been getting different results on `aarch64` vs `x86_64`. Not sure if it's an OS thing, or something else, but it's related to floating-point operations.

tristanls-tbp:
Perfect, thanks, found it through the links 👍.

PR #911, opened 2026-04-13
refactor!: transforms for environment obs as middleware pattern

tcdubs:
Implementation of the environment_utils.transforms classes as a middleware pattern-style pipeline encapsulated inside of the sensor module class.


tristanls-tbp (commented):
Thank you for providing a mechansim to use transforms in sensor modules instead of environment. Very exciting to see these working.

**issue (non-blocking)**: Since the work here is intended to replace the current use of transforms, then they should replace the current use of transforms. However, since these are sensor module transforms, technically, they can live side-by-side with the existing environment transforms and a full migration can be done in the scope of a different PR.

**question**: Have you swapped to use sensor module transforms instead of environment transforms in an existing benchmark and verified that you get the same results?

**issue**: As these are transforms intended to be used in the sensor modules, they should not be in the environment part of the source tree. **suggestion**: `transform_handlers.py` -> `src/tbp/monty/sensors/transforms.py` (you'll need to create the `sensors` directory).

**question**: What are these test files with `_refactor_` in their name for?

**todo**: Copyrights should reflect the file creation time TBP 2026. https://thousandbrainsproject.readme.io/docs/code-style-guide#source-code-copyright-and-license-header

PR #912, opened 2026-04-13
test!: improved NormalizeTest and ProjectOntoTangentPlane test

hlee9212:
After digging into floating point precisions, `hypothesis`, `numpy.testing`, and rtol/atol -- ended up refactoring NormalizeTest and ProjectOntoTangentPlane test. Wasn't intentional but because these two functions are quite load bearing. I was actually trying to add tests for other classes/functions, but fiddling around with tolerances would break these two. 

Attaching the write-up about floating point precision et al. here for others as well. 

Note to self:
1. Small reminder to explain why `project_onto_tangent_plane` is actually quite load bearing (relates to [Gram-Schmidt process](https://en.wikipedia.org/wiki/Gram%E2%80%93Schmidt_process) or making vectors orthogonal). 
2. Strategies for generating orthogonal vectors (taking into account of numerical stability):
- 2D: Use analytical construction
- 3D: cross product
- 4D+: gram-schmidt (via `np.linalg.qr`)

### DIFF
--- a/src/tbp/monty/frameworks/utils/spatial_arithmetics.py
+++ b/src/tbp/monty/frameworks/utils/spatial_arithmetics.py
@@ -21,21 +21,24 @@ from scipy.spatial.transform import Rotation
 logger = logging.getLogger(__name__)
 
 
-def normalize(v: ArrayLike, epsilon: float = 1e-12) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


hlee9212:
Closing as it will be superseded by #918

PR #903, opened 2026-04-09
feat: Add TangentFrame class 

hlee9212:
Just putting the `__init__` class until I can draw a diagram for `transport()`. 

CleanShot 2026-04-08 at 18 47 50

Update: Drew diagram for `transport()`:

CleanShot 2026-04-08 at 18 58 08

PR #914, opened 2026-04-14
[DO NOT MERGE] Reference for spatial_arithmetics_test.py and TangentFrame

hlee9212:
This PR has the updated tests from #912 and implementation for TangentFrame class (#903) + their tests. It should be used as reference during pair programming for us to write from scratch (because this PR contains too many changes).


hlee9212:
Implemented in #921

PR #918, opened 2026-04-14
refactor!: change epsilon default and refactor NormalizeTest

hlee9212:
Introduces DEFAULT_TOLERANCE value for use in normalize() and its tests.

### DIFF
--- a/tests/unit/frameworks/models/salience/motor_policy_test.py
+++ b/tests/unit/frameworks/models/salience/motor_policy_test.py
@@ -15,7 +15,6 @@ from unittest.mock import Mock
 import numpy as np
 import quaternion as qt
 from hypothesis import given
-from hypothesis import strategies as st
 from scipy.spatial.transform import Rotation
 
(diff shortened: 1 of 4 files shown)


jeremyshoemaker (approved):

hlee9212:
@jeremyshoemaker Just requested re-review since I made the fix: aacd89f5a9b6a8f63ddfe422a344ab742f519fb7 (merge if it looks good)

jeremyshoemaker (approved):
Accidentally commented on the commit instead of here.

I think the change to add `()` to the call sites was sufficient to fix the broken tests, but I'm also fine with converting the function into a composite strategy.

tests/unit/frameworks/models/salience/motor_policy_test.py
tristanls-tbp, line 179: **issue**: Since `goal_xyz` no longer includes zero vectors, this exit for zero vectors is no longer needed.

PR #919, opened 2026-04-15
feat: Add compute_arc_from_tangent_projection

hlee9212:
To be worked on after directional_curvature() in #892 is merged.

CleanShot 2026-04-15 at 12 10 08
CleanShot 2026-04-15 at 12 10 36


src/tbp/monty/frameworks/utils/sensor_processing.py
```diff
@@ -23,6 +25,81 @@
...
+    """Correct displacement to true arc length on a curved surface.
+
+    When a sensor moves along a curved surface, the straight-line displacement
+    measured in the tangent plane underestimates the true distance traveled
+    along the curve. This function corrects that by converting the displacement
+    projection back to the actual arc length.
+
+    The correction assumes that the surface is locally a circle. This approximation
+    holds well when curvature is approximately constant over the displacement,
+    but is inaccurate for surfaces with rapidly varying curvature.
```
jeremyshoemaker, line 45: Is there a way to detect when this assumption is false and provide an appropriate response, e.g. raising an exception?

tests/unit/frameworks/utils/sensor_processing_test.py
```diff
@@ -27,11 +29,39 @@
...
+    kp = draw(st.floats(min_value=min_kp, max_value=max_kp))
+    min_k = max(kp / MAX_PROJ, 0.01)
+    k = draw(st.floats(min_value=min_k, max_value=MAX_K))
+    p = kp / k
+    sign = draw(st.sampled_from([-1, 1]))
+    return sign * p, k
+
+
+flat_params = regime_params(min_kp=0, max_kp=FLAT_THRESHOLD - DEFAULT_TOLERANCE)
+guard_params = regime_params(min_kp=1.0 + DEFAULT_TOLERANCE, max_kp=2.0)
```
jeremyshoemaker: Why is this called `guard_params`? What are we guarding against?

tests/unit/frameworks/utils/sensor_processing_test.py
```diff
@@ -47,6 +77,56 @@ def curvature_values(draw):
...
+
+    @given(tangent_projection=projections, curvature=curvatures)
+    def test_sign_preservation(self, tangent_projection, curvature):
+        result = compute_arc_from_tangent_projection(tangent_projection, curvature)
+        if tangent_projection > 0.0:
+            assert result > 0.0
+        elif tangent_projection < 0:
+            assert result < 0.0
+        else:
+            assert result == 0.0
```
jeremyshoemaker, line 104: We might want to use an `@example` to make sure we trigger this case.

jeremyshoemaker (commented):
So, overall I think this is okay. I have some questions though.

One of them is about what's going on with the commit history. This PR has 43 commits listed, and only 158 lines of change in 2 files. Some of these commits look like they belong to other branches we've worked on, that have since been merged into `main`.

I haven't followed your process closely enough to see what's happening. It looks like you're creating branches off of other branches to continue working while waiting for reviews, and that might be what's causing it. 

One way to prevent the PR commit history from being noisy would be to hold off on opening it until the other branch merges, then rebase the new branch onto `main`, _then_ open the PR. That might make pairing complicated though.

A part of the solution might be trying to adhere to stricter work in progress limits. I usually prefer not to get ahead of unmerged work by more than a single branch, and I usually don't open PRs _until_ I can base that PR off of `main`. One of the reasons is that I don't know what changes will be needed in the branch I'm working off of that's still in review, and those changes could cause lots of conflicts and rework in the new branch.

There's a tension we have to navigate here between not wanting to be blocked waiting on PR review, and not wanting to get so far ahead that we create more work for ourselves in dealing with conflics, or creating PRs that are more involved to review. Not to mention, GitHub lately seems to struggle with PRs that have a lot of content.

I hope that makes some sense.

jeremyshoemaker:
> One way to prevent the PR commit history from being noisy would be to hold off on opening it until the other branch merges, then rebase the new branch onto main, then open the PR. That might make pairing complicated though.

Actually, I just realized after going through this process myself that this doesn't work. You can't rebase on to `main` to clear things up because we're using squash merges to `main`, which means the shared commits are no longer shared.

hlee9212:
@jeremyshoemaker Re Git Commit History:

(Ugh, I accidentally closed the tab so writing this again)

1. Yes I've been experimenting with "creating branches off of other branches to continue working while waiting for reviews", and indeed that is the cause.

> One of the reasons is that I don't know what changes will be needed in the branch I'm working off of that's still in review, and those changes could cause lots of conflicts and rework in the new branch.

2. I agree and I've been dealing with merge conflicts (somewhat intentionally so I can get practice). I do like PyCharm's UI for merge resolution, at least.

3. I'm not good at Git and part of this is Hojae just doing random experiment to fail and learn. This week I learned that I don't like making branches off branches. What I really wanted to do was update the crappy unit tests from `feat.2d_sensor` and update to use property-based testing. What I did realize (albeit after a week) is that I should have just done this on the `feat.2d_sensor` repo where I've been doing the RP.

src/tbp/monty/frameworks/utils/sensor_processing.py
hlee9212: No, there is no easy/clean way to check this inside this function. 

Reason: It only sees curvature at a single point, and detecting non-constant curvature requires obtaining at least another sample along the displacement path. 

The documentation for TwoDSensorModule makes it clear that one **must** take small steps for this assumption to hold. The user can ignore that, but they shouldn't expect a good learned model then.

tests/unit/frameworks/utils/sensor_processing_test.py
hlee9212: "Guard" refers to the arcsin domain guard in the function as `arcsin(x)` is only defined for x in range [-1, 1]. 

Other possible candidates: `arcsin_domain_guard_params`, `beyond_radius_params`. Happy to change during pairing.

jeremyshoemaker:
> I'm not good at Git and part of this is Hojae just doing random experiment to fail and learn. 

No one starts out knowing this stuff, and that's a good way to learn. 😄 

> What I did realize (albeit after a week) is that I should have just done this on the `feat.2d_sensor` repo where I've been doing the RP.

That brings up another thing I noticed. This PR is coming from `feat.2d_sensor`. I think one of the underlying assumptions of our implementation project process is that we won't be pulling commits straight from a research prototype fork, but rather re-implementing them in branches on a fork of Monty. That might also help cut down on the commit history noise.

An example of this would be the MuJoCo work that I've been running somewhat like an implementation project based on the prototype we built in London. If you compare the NoopAgent that we just merged into `main` with the version in the prototype, they're similar, but also very different. None of the commits from the prototype were in that branch. I pretty much opened the prototype PR in one window, my IDE on a new branch in another, and manually copied (sometimes just by typing it out again) the things that made sense to copy, and rewrote the things that needed changing, plus adding all the tests.

tests/unit/frameworks/utils/sensor_processing_test.py
jeremyshoemaker: So, we're generating values outside that range in order to trigger behaviour in `compute_arc_from_tangent_projection` that returns the original value? Is that behaviour that we want?

I'm probably missing something in the math here that would help it make sense. I'm also trying to suggest a better name for this, since I don't think `guard_params` quite captures what we're going for.

src/tbp/monty/frameworks/utils/sensor_processing.py
```diff
@@ -23,6 +25,81 @@
 
 logger = logging.getLogger(__name__)
 
+FLAT_THRESHOLD = 0.001
+
+
+def compute_arc_from_tangent_projection(
```
jeremyshoemaker: I noticed this after spelling it out in a comment below. What do you think about renaming this to `arc_from_tangent_projection` or `arc_from_tan_projection`? Normally I'd not suggest abbreviating "tangent" but "tan" is a typical name for the function at least, so I think it would work here. The original version is a bit long, and I don't think we need the verb "compute" in the name.

tests/unit/frameworks/utils/sensor_processing_test.py
jeremyshoemaker: I think I'm confused by what `kp` is here, relative to your diagrams.

> Is that behaviour that we want?

Rereading the docstring on the function, the answer is yes, so answered my own question.

jeremyshoemaker (approved):

PR #925, opened 2026-04-16
ci: don't allow `test!` labels on PRs

jeremyshoemaker:
It doesn't make sense to allow `test!` labels on PRs, as changes that are limited to `tests` are by definition non-breaking changes.

This adds a check to our `triage` title check workflow and provides instructions for how to correct the problem.

### DIFF
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -13,9 +13,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check PR title format
+        env:
+          # Store in an environment variable to prevent trying to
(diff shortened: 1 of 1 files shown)


tristanls-tbp (commented):
Looks good, thank you. I'll approve once #924 is merged for sequencing.

tristanls-tbp:
**note**: Or I won't 🙂, I see you're still testing it.

jeremyshoemaker:
I actually found a bug in our existing script, so I'm having to fix that.

tristanls-tbp (approved):

PR #927, opened 2026-04-16
feat: BufferEncoder can encode dataclasses by default

scottcanoe:
This PR enables `BufferEncoder` to encode dataclass instances by default. Dataclasses are encoded as dictionaries.

Since there is no common type to dataclasses, support for encoding is implemented specially within `BufferEncoder` and not via the regular type/encoder registration.

I've added a couple of tests and removed some now unnecessary code. I also ran some experiments with detailed logging to verify everything works as expected.

Note: This will allow us to log telemetry without registering all telemetry classes. However, I'm going to propose we start using [orjson](https://github.com/ijl/orjson) once we're at Python >= 3.10, which is really fast and supports dataclass serialization by default (and also numpy data).


src/tbp/monty/frameworks/models/motor_system_state.py
```diff
@@ -63,32 +63,3 @@ class MotorSystemState(Dict[AgentID, AgentState]):
...
-
-    def convert_motor_state(self) -> dict[AgentID, Any]:
-        """Convert the motor state into something that can be pickled/saved to JSON.
-
-        i.e. substitute vector and quaternion objects; note e.g. copy.deepcopy does not
-        work.
-
-        Returns:
-            Copy of the motor state.
-        """
```
scottcanoe, line 75: Deleted since this was only used to support detailed logging and is now unnecessary.

tests/unit/frameworks/models/buffer_test.py
```diff
@@ -448,3 +450,25 @@ def test_buffer_encoder_encodes_actions_by_default(self):
...
+            json.loads(json.dumps(dataclass, cls=BufferEncoder)),
+            dataclass.__dict__,
+        )
+
+    def test_default_dataclass_encoding_overridden_by_registered_dataclass_encoder(
+        self,
+    ):
+        dataclass_1 = FakeDataclass1(data=[1, 2, 3])
+        dataclass_2 = FakeDataclass2(data=[1, 2, 3])
+        BufferEncoder.register(FakeDataclass2, lambda _: "CUSTOM_ENCODING")
```
scottcanoe, line 466: I'm sure there's a better way to write this test, but it works. I don't think we can use mocks since they'd need to be json-loadable.

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -732,6 +733,8 @@ def default(self, obj: Any) -> Any:
         encoder = self._find(obj)
         if encoder is not None:
             return encoder(obj)
+        if is_dataclass_instance(obj):
+            return obj.__dict__
```
tristanls-tbp: **suggestion**: Use `asdict(obj)` here using `from dataclasses import asdict`.

tristanls-tbp (commented):
Can we use `asdict` instead of reaching into "special" `__dict__`?

src/tbp/monty/frameworks/models/buffer.py
scottcanoe: `asdict` does recursive dict-ing all the way down, so any inner dataclasses that have custom encoders would miss the chance to have their encoder invoked.

But now that I'm vocalizing it, I'm realizing that I didn't write any tests to demonstrate this aspect.

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -732,6 +734,8 @@ def default(self, obj: Any) -> Any:
         encoder = self._find(obj)
         if encoder is not None:
             return encoder(obj)
+        if is_dataclass_instance(obj):
+            return {f.name: getattr(obj, f.name) for f in dataclasses.fields(obj)}
```
scottcanoe, line 738: This addresses the issue your raised about not wanting to touch `obj.__dict__` but in a non-recursive way.

tests/unit/frameworks/models/buffer_test.py
```diff
@@ -448,3 +452,42 @@ def test_buffer_encoder_encodes_actions_by_default(self):
...
+        dataclass_2_sentinel = "CUSTOM_ENCODING"
+        dataclass_2_encoder = Mock(return_value=dataclass_2_sentinel)
+        BufferEncoder.register(FakeDataclass2, dataclass_2_encoder)
+        dataclass_3 = FakeDataclass3(dataclass_1=dataclass_1, dataclass_2=dataclass_2)
+
+        dct = json.loads(json.dumps(dataclass_3, cls=BufferEncoder))
+
+        self.assertEqual(dct["dataclass_1"], dataclasses.asdict(dataclass_1))
+        dataclass_2_encoder.assert_called_once_with(dataclass_2)
+        self.assertEqual(dct["dataclass_2"], dataclass_2_sentinel)
```
scottcanoe, line 493: For the nested dataclass issue.

scottcanoe (commented):
@tristanls-tbp I think these edits address your suggestion and validate nested dataclass serialization.

tristanls-tbp (approved):
Thank you, really well done, much better than my suggestion 👍.

PR #928, opened 2026-04-16
chore: autoimport hints for our numpy import convention

tristanls-tbp:
Numpy import convention hints
```python
import numpy as np
import numpy.typing as npt
```
For `numpy.testing`, do not use `npt`.


jeremyshoemaker (approved):

tristanls-tbp:
@jeremyshoemaker requested a re-review since I removed the `nptest` thing... although now that I'm writing this, probably doesn't require re-review to do less than what I wanted, sorry.

jeremyshoemaker:
Coming from other Python projects I've worked with in the past, I don't think we ever did one of these qualified imports. We'd just import the things we need from whatever package, and if there ended up being a conflict, we'd address it then.

I feel like `np.testing.*` is just noise, when importing things like `assert_allclose` and using that directly would be better. It's not like our editors can't tell us where that comes from.

jeremyshoemaker (approved):

tristanls-tbp:
> I feel like `np.testing.*` is just noise, when importing things like `assert_allclose` and using that directly would be better. It's not like our editors can't tell us where that comes from.

I'm happy with that too. I'll update the description once again...

This PR will be about typing `npt` to typing and that's it.

PR #933, opened 2026-04-20
test: exclude MuJoCo configs from update_snapshots.py

jeremyshoemaker:
We want to temporarily exclude the MuJoCo configs from the update_snapshots script because it requires `mujoco` to be installed to load the config with Hydra.

### DIFF
--- a/src/tbp/monty/conf/update_snapshots.py
+++ b/src/tbp/monty/conf/update_snapshots.py
@@ -46,6 +46,10 @@ def update_snapshots(
         existing_snapshot.unlink()
 
     for file_path in config_dir.glob("*.yaml"):
+        # Exclude MuJoCo experiments
+        # TODO: Revert once we convert to MuJoCo
+        if file_path.stem.endswith("mujoco"):
+            continue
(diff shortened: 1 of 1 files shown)


src/tbp/monty/conf/update_snapshots.py
```diff
@@ -46,6 +46,10 @@ def update_snapshots(
         existing_snapshot.unlink()
```
scottcanoe, line 46: Probably don't want to delete the mujoco snapshots.

scottcanoe (changes requested):

src/tbp/monty/conf/update_snapshots.py
jeremyshoemaker: Repeating what I said in Slack, we don't have any yet. I'm holding off on generating them until I start migrating experiments over.

scottcanoe: Oh ok, I just assumed they were automatically generated. Approved.

scottcanoe (approved):
Looks good, thanks for the quick turnaround time.

PR #934, opened 2026-04-20
feat: compute_edge_features

hlee9212:
To be worked on during Pair Programming. 

1. We're going to try out integrating a larger chunk of the code in this PR. Still conceptually one thing, but has many helper functions (rather than ~1 function/PR in the previous week). 
2. I have done some refactoring so the screenshots of code you may have seen in my presentation from Nov 2025 will not be present (but the algorithm is the same, and, should be easier to understand with the refactoring). 
3. I'm open to encapsulating `compute_edge_features` and its helper functions + data into a Python class. I generally write write code more procedurally (functions first) for RP, and allow structure to emerge naturally. After refactoring, I think shoving all these into a class may make sense. However, I'm also open to punting this to [Post IP phase](https://app.shortcut.com/thousandbrainsproject/epic/4772?group_by=none&hide_subtasks=false&ct_workflow=all&cf_workflow=500000005) or as a follow-up as this is a conceptually dense PR to review. 

---
Helpful Images showing what we are testing:
edge_detection


jeremyshoemaker:
> However, I'm also open to punting this to Post IP phase or as a follow-up as this is a conceptually dense PR to review.

There is no "Post IP Phase". These sorts of refactors and clean-up are _meant_ to happen during the implementation project. The ideal goal state for any IP should be that whatever we've added is polished up and ready to be used by other researchers, outside contributors, future people using the platform to make things.

If we end up feeling like there's additional refactoring needed immediately after the fact, then we've all failed to scope the project accurately.

> I'm open to encapsulating compute_edge_features and its helper functions + data into a Python class.

I've been wondering this myself and haven't had a moment to sit down and look at what we've added and how we can refactor it into something like a coherent class. It might not make sense to do so. Sometimes functionality is separate. The thing that makes me think there _might_ be a better way to organize this is the fact that all of this code is under `utils`.

---

Since we're still working on this branch, let's make it a draft for now. It's currently showing up on our triage boards and it will prevent that since I don't think it's currently ready for review.

src/tbp/monty/frameworks/utils/edge_detection.py
```diff
@@ -0,0 +1,320 @@
...
+        self._coherence_threshold = coherence_threshold
+        self._radius = radius
+        self._sigma_r = sigma_r
+        self._depth_edge_threshold = depth_edge_threshold
+        self._max_center_offset = max_center_offset
+
+        if self._kernel_size % 2 != 1:
+            raise ValueError("Kernel size must be odd.")
+
+    def detect_feature(
```
jeremyshoemaker: I think this method could be renamed to `__call__` so that we can just use `detector(patch)` where it's used. We've already used this pattern of a class that does a single thing using `__call__` elsewhere, so continuing it here makes sense. Otherwise I fear we'll fall into the pattern of `Verber.verb`, which is redundant.

jeremyshoemaker (commented):
I have a small refactor to suggest here before we merge.

jeremyshoemaker (approved):

PR #935, opened 2026-04-20
feat: introduce Pose utilities

dschumacher-tbp:
These classes implement the Pose, Location, and Orientation concepts as defined in the [TBP Glossary](https://thousandbrainsproject.readme.io/docs/glossary#pose) and following documented [conventions](https://thousandbrainsproject.readme.io/docs/conventions) for axes, coordinates, and angles.  In additions, quaternions used to define a rotation are represented scalar-first (e.g.: [_w_, _x_, _y_, _z_]). Each object is aware of its Reference Frame, and the frame is taken into account during computations where appropriate. The overall goal is to make [reference frame transformations](https://thousandbrainsproject.readme.io/docs/reference-frame-transformations) easy to understand and use, as described in the [Future Reality Tree](https://link.excalidraw.com/readonly/1zkoPDpkfKlbyGfCaLsY).

image

### Reference Frames

A Pose is defined by a Location and Orientation relative to Reference Frame.

image

In the current design, the Reference Frame is also a Pose.

### Current Effort

The goal of this PR is to introduce these helper classes, in order to begin integrating them. As the integration proceeds, we should learn common usage patterns that will inform future expansion of their utillity. The API is expected to adapt.

### Future Work

It is likely that additional concepts such as Displacement, Rotation, Transformation, and Frame will also be helpful. However, that is not within the scope of this PR.

Bulk operations on NumPy `NDArray` object will also likely be important for achieving desired performance levels.


jeremyshoemaker:
@tristanls-tbp Are we not applying labels as [documented](https://thousandbrainsproject.readme.io/docs/triage#labels-1)? I know we've been inconsistent about it, but @dschumacher-tbp asked, and I referenced that page.

jeremyshoemaker:
Actually, now that I double-checked, one of those labels is for "Issues" not "Pull Requests". That's not the first time I've looked at the wrong part of that page.

dschumacher-tbp:
Taking a step back and considering what a more complete set of Classes might look like...
```mermaid
classDiagram
    class Transform {
        + displacement : Displacement
        + rotation : Rotation
    }

    class Displacement {
        + xyz : array_3 | array_3xN
        + copy() Displacement
        + move_by(xyz) Self
        + inverse() Displacement
    }

    class Rotation {
        + wxyz : array_4 | array_4xN
        + copy() Rotation
        + yaw(theta: Radians) Self
        + pitch(phi: Radians) Self
        + roll(psi: Radians) Self
        + compose(r: Rotation) Rotation
        + inverse() Rotation
    }

    class Framed {
        <<protocol>>
        + frame : Frame | None*
    }

    class Frame {
        + frame : Frame | None
        + label : str
        + transform : Transform
        + location : Location*
        + orientation : Orientation*
    }

    class Pose {
        + frame : Frame | None
        + location : Location
        + orientation : Orientation
    }

    class Location {
        + frame : Frame | None
        + displacement : Displacement
    }

    class Orientation {
        + frame : Frame | None
        + rotation : Rotation
    }

    Transform *--> Displacement
    Transform *--> Rotation
    Framed o--> Frame
    Framed <|.. Frame
    Framed <|.. Pose
    Framed <|.. Location
    Framed <|.. Orientation
    Frame *--> Transform
    Pose *--> Location
    Pose *--> Orientation
    Location *--> Displacement
    Orientation *--> Rotation
```

PR #945, opened 2026-04-22
feat: add TwoDSensorModule class

hlee9212:
Introduce a new class called `TwoDSensorModule` along with its tests. The PR also updates the structure of EdgeDetection class to separate out converting of edge angle to appropriate reference frame to make EdgeDetection agnostic to reference frame.


src/tbp/monty/frameworks/models/two_d_sensor_module.py
```diff
@@ -0,0 +1,368 @@
...
+    The 2D position is initialized to the world x,y coordinates. A 2D model is
+    built by accumulating tangent plane displacements.
+
+    Note:
+        This implementation represents the 2D model in the world xy-plane
+        by setting z to zero and initializing 2D position from world x/y.
+        This is appropriate when the relevant camera or object-facing projection
+        is aligned with world xy. It will be incorrect for views whose image
+        plane is not aligned with world xy, for example a camera looking along
+        the x-axis, where the visible plane is closer to yz.
```
jeremyshoemaker, line 75: I didn't see this when we were working on it, so I'll ask here.

What aspect of the code is leading to this limitation? Would it be difficult to make this work in more cases? If not, is there a way we can check for this case and log a warning or return an error?

jeremyshoemaker (approved):
I don't want to hold things up on my last comment, since I'm out tomorrow.

src/tbp/monty/frameworks/models/two_d_sensor_module.py
hlee9212: Hi @jeremyshoemaker, this was related to a TODO for me to add comment on (it was something like` # setting z = 0 assumes...TODO` before). This comment has been updated to ` # Setting z = 0 assumes that camera is aligned with world xy.` I figured I would add a longer explanation in the `Note` section.

I think future updates to TwoDSensorModule (or at least, what I'm writing in my documentation) is that it will be flexible enough to set any of x, y, or z to 0, or even just make sure that whatever the points we learn (even if x, y, and z are all non-zero) all lie on a 2D plane in 3D space.

PR #937, opened 2026-04-21
feat!: wxyz-friendly Rotation object

scottcanoe:
This PR addresses the lack of consistency in quaternion formats in `tbp.monty` by replacing the usage of `scipy.spatial.transform.Rotation` with `tbp.monty.geometry.Rotation` and eliminates all instances of quaternion-format handling throughout the code base.

**Changes**
 1. [non-breaking] I've added the file `src/tbp/monty/geometry.py` which contains the `Rotation` wrapper.
 2. Everywhere `scipy.spatial.transform.Rotation` was imported has been replaced with an import of `tbp.monty.geometry.Rotation` EXCEPT the new mujoco code.
 
Note that I left mujoco code alone for now. SInce that's being actively developed, I wanted to coordinate updating those with @jeremyshoemaker. Once those are updated, we can remove `transform_utils.py` entirely.

## Benchmarks (no changes)

❌: Changed
n.s.: Not Significant (runtimes only)

### base_config_10distinctobj_dist_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100 | 100 | 0 |   |
| used_mlh (%) | 2.86 | 2.86 | 0 |   |
| match_steps | 34 | 34 | 0 |   |
| rotation_error (deg) | 8.64 | 8.64 | 0 |   |
| runtime (min) | 3 | 3 | 0 |   |
| episode_runtime (sec) | 7 | 8 | 1 |  n.s. |
| num_episodes | 140 | 140 | 0 |   |

### base_config_10distinctobj_surf_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100 | 100 | 0 |   |
| used_mlh (%) | 0 | 0 | 0 |   |
| match_steps | 28 | 28 | 0 |   |
| rotation_error (deg) | 3.87 | 3.87 | 0 |   |
| runtime (min) | 3 | 3 | 0 |   |
| episode_runtime (sec) | 11 | 11 | 0 |   |
| num_episodes | 140 | 140 | 0 |   |

### randrot_noise_10distinctobj_dist_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100 | 100 | 0 |   |
| used_mlh (%) | 2 | 2 | 0 |   |
| match_steps | 35 | 35 | 0 |   |
[...]

PR #951, opened 2026-04-24
feat: implement MuJoCo DistantAgent

jeremyshoemaker:
Adds a DistantAgent to the MuJoCo simulator agents.

This change also includes some updates to the existing MuJoCo tests to handle edge cases that were missed previously.


src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -937,9 +594,9 @@ def _assert_undo_jump_was_successful(self, state: MotorSystemState) -> None:
         )
 
         for current_sensor in state[self.agent_id].sensors:
-            assert np.all(
-                state[self.agent_id].sensors[current_sensor].rotation
-                == self._pre_jump_state.sensors[current_sensor].rotation
+            assert np.allclose(
+                state[self.agent_id].sensors[current_sensor].rotation,
+                self._pre_jump_state.sensors[current_sensor].rotation,
```
jeremyshoemaker, line 942: This change _might_ be a breaking change. I'm not sure though. Depends on how much we want to abide by [Hyrum's Law](https://www.hyrumslaw.com/).

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -332,19 +376,9 @@ def _add_primitive_object(
...
         """
         world_body: MjsBody = self.spec.worldbody
-
-        try:
-            geom_type = PRIMITIVE_OBJECTS[object_type]
-        except KeyError:
-            raise UnknownObjectType(
-                f"Unknown MuJoCo primitive: {object_type}"
-            ) from None
-
```
jeremyshoemaker, line 347: This ended up being dead code due to the similar check in `add_object` before we call this method.

src/tbp/monty/frameworks/models/motor_policies.py
scottcanoe: `np.allclose` is better, but this whole policy will be gone very soon.

scottcanoe: For context, this class is only being used by surface agents (as of yesterday). With a small refactor, surface agents will use `JumpToGoal` lilke distant agents do. We also decided that this kind of checking is better off being unit/integration tested anyways, so we dropped the check altogether.

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -192,6 +218,8 @@ def state(self) -> AgentState:
 
         sensor_states = {}
         for sensor_id, sensor_cfg in self._sensor_configs.items():
+            # This code assumes that the position of the sensor body, or sensors
+            # relative to the agent CANNOT change.
```
scottcanoe:
```suggestion
            # This code assumes that the position of the sensor body, or sensors
            # relative to the agent CANNOT change. This is because we we use the
            # configured position of the sensor (rel. agent) to compute positions. This
            # constraint can be removed by computing the sensor's position relative
            # agent from their world coordinates.
```
It took me a minute to work out why this true.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -122,41 +123,79 @@ def __init__(
...
+        i.e. a directional light from the front of the object, with ambient lighting
+        on the back sides.
+
+        Using a fixed ambient light on the back side doesn't provide good results, but
+        putting it on the headlight does.
+
+        Note: this makes a lot of assumptions about the layout of our scene and
+          the objects in the scene.
+        """
+        # TODO: Consider making these configurable.
```
scottcanoe, line 172: I think we're switching to `"TODO(jeremyshoemaker): ..."` format.

tests/unit/simulators/mujoco/distant_agent_test.py
```diff
@@ -0,0 +1,209 @@
...
+
+@st.composite
+def unit_quaternion(draw) -> QuaternionWXYZ:
+    x_rad = draw(st.floats(min_value=0.0, max_value=np.pi * 2))
+    y_rad = draw(st.floats(min_value=0.0, max_value=np.pi * 2))
+    z_rad = draw(st.floats(min_value=0.0, max_value=np.pi * 2))
+    # run through a SciPy rotation object to get quaternion values
+    rotation = Rotation.from_euler("xyz", [x_rad, y_rad, z_rad], degrees=False)
+    normalized_rotation = rotation_as_quat(rotation)
+    return cast("QuaternionWXYZ", tuple(normalized_rotation))
```
scottcanoe, line 68: I've been meaning to bring up the idea of having a common module for these functions. I made a pretty similar one, but its ranges were [-pi, pi] instead of [0, 2 * pi]. It _shouldn't_ matter, but Euler angles ¯\\_(ツ)_/¯

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -40,6 +51,15 @@
 DEFAULT_CAMERA_FOVY: float = 45.0
 
 
+class Axis(Enum):
```
scottcanoe: **suggestion**: Use `IntEnum`, which makes the members behave like integers. Meaning we could use it directly for indexing. So like 245 would go from
```python
axis_vector = rotation_matrix[:, axis.value] * distance
```
to 
```python
axis_vector = rotation_matrix[:, axis] * distance
```

scottcanoe (commented):
Looking good. It's hard to verify the math just from reviewing, but as you mentioned, we can work out issues during testing (esp. integration tests). Let me know if I can help with any of that.

As for the rotation type, it sounds like @dschumacher-tbp and I are going to bring the [wxyz-friendly Rotation object](https://github.com/thousandbrainsproject/tbp.monty/pull/937) into main next week. If we can agree to use it as our standard, then this code will get simpler.

Last thoughts:
 - Having the `Agent` and `Embodiment` define `max_sensor_resolution` properties feels funny to me. Proprioceptive state seems like a real property of an embodiment, but the maximum sensor resolution doesn't. As far as I can tell, the `max_sensor_resolution` properties are just there to propagate a value up to the simulator, but I wonder if there's a way to just grab that value from the config during initialization or something.
- I forgot my other thought.

tests/unit/simulators/mujoco/distant_agent_test.py
jeremyshoemaker: I've had the same thought, but was waiting to see how things evolved to figure out what might work best. The last thing I want to do is create a `test_utils.py` or something, but there's definitely shared testing helper code between this test and the NoopAgent tests, and probably others.

Another note here, that maybe should be a code comment or docstring. The reason I'm using Euler angles here is so that I can have the Rotation object generate a unit quaternion, rather than try to construct one from raw coefficients. It also probably makes the rotations be equally distributed throughout 3D space, rather than the 4D space of quaternions.

src/tbp/monty/simulators/mujoco/agents.py
jeremyshoemaker: For some reason I thought that was a 3.11 feature.

jeremyshoemaker:
> Having the Agent and Embodiment define max_sensor_resolution properties feels funny to me. Proprioceptive state seems like a real property of an embodiment, but the maximum sensor resolution doesn't. As far as I can tell, the max_sensor_resolution properties are just there to propagate a value up to the simulator, but I wonder if there's a way to just grab that value from the config during initialization or something.

This had been bothering me too, but your comment nudged me to come up with a better solution. Turns out it's really straightforward to introspect a `partial` to figure out what the fixed arguments are.

jeremyshoemaker:
> It's hard to verify the math just from reviewing, but as you mentioned, we can work out issues during testing (esp. integration tests). Let me know if I can help with any of that.

I definitely discovered some issues with property testing, and I'm fairly certain I sorted out all the problems.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I don't think that's necessary. 99% of the time, the `git blame` for the line in question will have the author of the TODO comment, and if not, and if you _really_ need to find who wrote it, there are ways to do that easily.

I feel like this is in the same category as having commented out code because we _might_ need it in the future. Use Git features instead.

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -208,6 +220,49 @@ def state(self) -> AgentState:
...
+        qpos_addr = self.sim.model.jnt_qposadr[self.pitch_joint.id]
+        self.sim.data.qpos[qpos_addr] += delta_phi
+
+    def set_pose(self, location: VectorXYZ, rotation: QuaternionWXYZ) -> None:
+        """Set the location and rotation of the embodiment to the provided values."""
+        # TODO: replace with a property setter using a Pose object
+        self.position = location
+        self.rotation = np.array(rotation)
+
+    def set_sensor_rotation(self, rotation_quat: QuaternionWXYZ) -> None:
```
tristanls-tbp, line 274: **issue**: If I understand correctly, each embodiment can have multiple sensors. This code appears to assume there is only a single sensor.

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -208,6 +220,49 @@ def state(self) -> AgentState:
...
+        self.position = new_xyz
+
+    def yaw(self, delta_theta: float) -> None:
+        """Yaw the embodiment by delta_theta degrees."""
+        delta_theta_rot = Rotation.from_euler("xyz", (0, delta_theta, 0), degrees=True)
+        rotation = rotation_from_quat(self.rotation)
+        new_rotation = rotation * delta_theta_rot
+        self.rotation = rotation_as_quat(new_rotation)
+
+    def pitch(self, delta_phi: float) -> None:
```
tristanls-tbp: **issue**: If I understand correctly, each embodiment can have multiple sensors. This code appears to assume there is only a single sensor.

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -208,6 +220,49 @@ def state(self) -> AgentState:
     def reset(self) -> None:
         self.position = self._initial_position
         self.rotation = self._initial_rotation
+        self.set_sensor_rotation(IDENTITY_QUATERNION)
```
tristanls-tbp, line 234: **issue**: If I understand correctly, each embodiment can have multiple sensors. This code appears to assume there is only a single sensor.

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -251,3 +333,39 @@ def reset(self) -> None:
...
+    def actuate_set_sensor_rotation(self, action: SetSensorRotation) -> None:
+        rotation = action.rotation_quat
+        if isinstance(rotation, qt.quaternion):
+            # TODO: Fix all the places SetSensorRotation is created
+            logger.warning(
+                "SetSensorRotation rotation is a qt.quaternion and "
+                "not a QuaternionWXYZ."
+            )
+            rotation = cast("QuaternionWXYZ", tuple(qt.as_float_array(rotation)))
+        self._embodiment.set_sensor_rotation(rotation)
```
tristanls-tbp, line 375: **issue**: If I understand correctly, each embodiment can have multiple sensors. This code appears to assume there is only a single sensor.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -122,65 +125,111 @@ def __init__(
...
         for agent_partial in self._agent_partials:
             agent = agent_partial(self)
             self._agents[agent.id] = agent
 
     def _max_sensor_resolution(self) -> Resolution2D:
-        """Determine the maximum resolution of all the sensors.
+        """Returns the maximum width and heights of the sensors.
+
+        Used by the simulator to determine the size of the off-screen rendering
+        surface to insure it is always large enough for any sensor images we
```
tristanls-tbp:
```suggestion
        surface to ensure it is always large enough for any sensor images we
```

tests/unit/simulators/mujoco/distant_agent_test.py
tristanls-tbp, line 68: **note**: Since `Rotation` normalizes, this rotation creation would also work:
```python
rotation = Rotation.from_quat([
    draw(st.floats(min_value=-1.0, max_value=1.0)),
    draw(st.floats(min_value=-1.0, max_value=1.0)),
    draw(st.floats(min_value=-1.0, max_value=1.0)),
    draw(st.floats(min_value=-1.0, max_value=1.0))
])
```
Proof that it normalizes:
```
>>> r = Rotation.from_quat([0.0, 5.0, 110.0, 134.0])
>>> r.as_quat()
array([0.        , 0.02882862, 0.63422966, 0.77260704])
```

tests/unit/simulators/mujoco/distant_agent_test.py
```diff
@@ -0,0 +1,209 @@
...
+    def test_set_sensor_rotation(self, sensor_rotation):
+        action = SetSensorRotation(
+            agent_id=TEST_AGENT_ID, rotation_quat=sensor_rotation
+        )
+        self.sim.step([action])
+
+        sensor_state = self.sim.states[TEST_AGENT_ID].sensors[TEST_SENSOR_ID]
+        assert sensor_state.position == (0.0, 0.0, 0.0)
+        # Due to the transformations that happen to sensor.rotation, some of the
+        # quaternions come back as their equivalent negative. Changing to
```
tristanls-tbp: **nitpick**: I keep wondering where they are coming back from.

```suggestion
        # quaternions are rotated to their negative, but represent the same rotation. Changing to
```

tests/unit/simulators/mujoco/distant_agent_test.py
```diff
@@ -0,0 +1,209 @@
...
+        self.sim.step([action])
+
+        agent_state = self.sim.states[TEST_AGENT_ID]
+        assert agent_state.position == final_position
+        assert agent_state.rotation == qt.quaternion(*final_rotation)
+
+    @given(
+        sensor_rotation=x_rotation_quaterion(),
+    )
+    # Example of a quaternion that ends up coming back as its negative.
```
tristanls-tbp: **nitpick**: I keep wondering where they are coming back from.

```suggestion
    # Example of a quaternion that is rotated to its negative, which is an equivalent rotation.
```

tests/unit/simulators/mujoco/distant_agent_test.py
```diff
@@ -0,0 +1,209 @@
...
+    @given(distance=st.floats(min_value=0.0, max_value=10.0))
+    @settings(deadline=None)
+    def test_move_forward(self, distance):
+        action = MoveForward(agent_id=TEST_AGENT_ID, distance=distance)
+        with sim_resetter(self.sim):
+            self.sim.step([action])
+
+            agent_state = self.sim.states[TEST_AGENT_ID]
+            # Moving forward moves the agent in the negative Z direction, so the
+            # returned distance will be the negative of the requested distance.
```
tristanls-tbp:
```suggestion
            # moved distance will be the negative of the requested distance.
```

tests/unit/simulators/mujoco/distant_agent_test.py
```diff
@@ -0,0 +1,209 @@
...
+            agent_rot = rotation_from_quat(qt.as_float_array(agent_state.rotation))
+            expected_rot = Rotation.from_euler(
+                "xyz", [0.0, delta_theta, 0.0], degrees=True
+            )
+
+            assert agent_rot.approx_equal(expected_rot)
+
+    @given(delta_phi=st.floats(min_value=-180.0, max_value=180.0))
+    @settings(deadline=None)
+    def test_look_up(self, delta_phi):
```
tristanls-tbp: **issue**: `LookUp` and `LookDown` have built-in `constraint_degrees` that is `90.0` by default. This test passing tells me the constraint is being ignored.

src/tbp/monty/simulators/mujoco/agents.py
jeremyshoemaker: No, it doesn't currently work that way. While an agent can have multiple sensors (see the 5 sensor setups we have), they don't move independently. They essentially rotate as a group around a shared center. We're carrying that over here by having them all be inside of a "body" in the MuJoCo simulator, and pitching that one up and down.

This will need to change if we ever want to do it different, but for now this is following the existing limits of the system we have built.

tests/unit/simulators/mujoco/distant_agent_test.py
jeremyshoemaker: I ran into issues with that approach generating zero-quaternions, which aren't valid. By specifying angles instead, we easily ensure we always get a non-zero quaternion, and the Rotation takes care of making sure it's a unit quaternion.

Like, if you try to construct a Rotation object with all zeros, it will fail.

tests/unit/simulators/mujoco/distant_agent_test.py
jeremyshoemaker: That's a fair suggestion. I could have also said that they come back in the proprioceptive state as their negative equivalents. But this is shorter.

tests/unit/simulators/mujoco/distant_agent_test.py
jeremyshoemaker: Again, I'm thinking of the proprioceptive state that we get back from the agent. That is returned from the agent.

tests/unit/simulators/mujoco/distant_agent_test.py
jeremyshoemaker: Good catch. I don't think we caught that in our prototype either. Maybe @scottcanoe had a reason we were ignoring it. Sounds familiar. Either way, the code should make sure to respect the limit.

jeremyshoemaker: Updated both the tests to test for this, and the code to comply with the constraints.

tests/unit/simulators/mujoco/distant_agent_test.py
dschumacher-tbp: Is it possible to use `Rotation.random()` for generating test-cases?
https://docs.scipy.org/doc/scipy-1.10.1/reference/generated/scipy.spatial.transform.Rotation.random.html

jeremyshoemaker: The problem with using `Rotation.random` would be that it's outside of Hypothesis' infrastructure. The `draw` function isn't generally random. It's what allows Hypothesis to generate specific edge cases, and also to narrow down minimal reproduction cases.

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -251,3 +333,41 @@ def reset(self) -> None:
...
+        self._embodiment.move_along_local_axis(distance=-action.distance, axis=Axis.Z)
+
+    def actuate_turn_right(self, action: TurnRight) -> None:
+        self._embodiment.yaw(-action.rotation_degrees)
+
+    def actuate_turn_left(self, action: TurnLeft) -> None:
+        self._embodiment.yaw(action.rotation_degrees)
+
+    def actuate_look_up(self, action: LookUp) -> None:
+        rotation_degrees = min(action.rotation_degrees, action.constraint_degrees)
```
tristanls-tbp: **issue**: I think the constraint will not be applied if `rotation_degrees` is negative.

tests/unit/simulators/mujoco/distant_agent_test.py
tristanls-tbp: **note**: Right right... I forgot the nonzero requirement... 
```python
rotation = Rotation.from_quat([
    draw(st.one_of(st.floats(min_value=-1.0, max_value=-1e-6), st.floats(min_value=1e-6, max_value=1.0),
    ...
])
```

tests/unit/simulators/mujoco/distant_agent_test.py
```diff
@@ -0,0 +1,231 @@
...
+            agent_state = self.sim.states[TEST_AGENT_ID]
+            agent_rot = rotation_from_quat(qt.as_float_array(agent_state.rotation))
+            expected_rot = Rotation.from_euler(
+                "xyz", [0.0, delta_theta, 0.0], degrees=True
+            )
+
+            assert agent_rot.approx_equal(expected_rot)
+
+    @given(
+        delta_phi=st.floats(min_value=0.0, max_value=180.0),
```
tristanls-tbp: **issue**: Degrees can be negative

jeremyshoemaker: 🤦‍♂️ Of course they can. Because why use a LookDown when you can send LookUp with a negative value. We really should stop doing this.

dschumacher-tbp: Maybe the Actions should be something like Yaw, Pitch, and Roll?

jeremyshoemaker: Either that or we should be more strict about only accepting positive values and change the code to create the correct actions. The code that violates this that I'm aware of is the get good view stuff. It always gives LookDown and TurnLeft even when going up or right.

Another aspect of this that makes me a little hesitant with Yaw, Pitch, and Roll is that Pitch, i.e. LookUp and LookDown, are relative to the sensor body, not the agent body, while the other two are, so we'd be conflating two different reference frames.

dschumacher-tbp: `PitchSensor` and `YawSensorBody`? How much of the kinematic-chain are we encoding?

jeremyshoemaker: I just pushed up a change, but I don't think this math is correct either. Working on another fix.

tests/unit/simulators/mujoco/distant_agent_test.py
```diff
@@ -189,43 +189,82 @@ def test_turn_left(self, delta_theta):
...
+        # This is technically the same code that Embodiment uses, but I can't
+        # come up with an alternative way to calculate these values.
+        new_angle = current_angle + delta_phi
+
+        if new_angle > constraint_angle:
+            delta_phi = constraint_angle - current_angle
+        elif new_angle < -constraint_angle:
+            delta_phi = -constraint_angle - current_angle
+
+        return current_angle + delta_phi
```
tristanls-tbp: **suggestion**: I'm thinking that the check should be that no matter how many lookups or lookdowns you do, your orientation is always <= than the bound. This would avoid having to calculate any expected value and just assert that the constraint is being met. Something (loosely handwaving) like:
```python
roll, pitch, yaw = sensor_state.rotation.as_euler("xyz", degrees=True)
assert abs(pitch) <= constraint_angle
```

**suggestion**: Two tests... one where you expect to be within bounds, where you validate against the expected value. Second, where you expect the constraint to be met, no matter how many actions are actuated.

tristanls-tbp (approved):
Looks great!

PR #952, opened 2026-04-24
feat: log bursting frequency

ramyamounir:
We are interested in logging the number of bursts triggered every episode. The existing telemetry approach only works as part of the detailed logger. Here we want this metric to be reported in the basic logger in the CSV file as well as on WandB.
This metric can be used to optimize the burst sampling parameters. No benchmark runs are needed here.

An example of what this looks like in WandB is below:
image

Also this is what it looks like in the CSV:
image


nielsleadholm (approved):
Thanks for adding this, LGTM!

ramyamounir:
@tristanls-tbp I've followed the existing pattern of adding the prediction error (introduced during the first London retreat). I know this is not the best approach but I thought maybe this is okay for now as a temporary solution until the telemetry is rewritten?

tristanls-tbp (approved):
Without future telemetry implemented, I don't have a better suggestion.

PR #851, opened 2026-03-08
refactor!: remove channel tracking from hypothesis spaces

ramyamounir:
The `EvidenceGraphLM` uses the `channel_hypothesis_mapping` to track input channels in every graph hypotheses. This uses the `ChannelMapper` which stores the boundary indices between sets of hypotheses that represents input channels. The goal of this PR is to completely remove the channel separation and instead use a unified hypothesis space.

Specifically, a hypothesis should only be defined by three attributes: location, orientation, and evidence. We do not want to separate the hypotheses by input channels when updating their evidence. This allows a single hypothesis to collect evidence based on feature matching of all channels. We sum up the evidence from all channels' feature matching and apply it to the hypothesis space **only** based on distance (not distance and input channel). We need this for compositional models, at least.

This is a large scope refactor. I couldn't split it into multiple PR without breaking Monty. I'm happy to take suggestions to help me split it up into more atomic PRs to make the review process easier. I will request reviews in batches, starting with the research team first, then engineering.

The PR changes many classes:
- `EvidenceGraphLM`: Removed the `_set_hypotheses_in_hpspace`. We now get a single hypotheses update which we can directly assign to the hypotheses space attributes. Also removed `channel_hypothesis_mapping`, we are no longer tracking input channels.
- `DefaultHypothesesUpdater`: Modified the `update_hypotheses` to handle initializing new channels in the first step and in later steps. When a new channel is initialized later in the episode, it gets the mean of hypotheses evidence added to the initial evidence score. This is now handled in the hypothesis updater instead of in the LM.
- `BurstSamplingHypothesesUpdater`: The sampling multiplier is now in the range [0, 1], where 0 means no sampling and 1 means the total number of available informed hypothesis (graph nodes * hyps per node). Sampling is now computed proportionally per-channel based on the graph size and PC-defined.
- `DefaultHypothesesDisplacer`: It now takes multiple channels for features and displacements together. It displaces all hypotheses in a single step (possibly saving computations), then performs feature matching for every channel separately and sums up the evidence.
- `ChannelMapper`: Removed.
- `EvidenceSlopeTracker`: Simplified by removing the channel parameter from all functions. The tracker now operates on the full hypothesis space.
- Telemetry classes were updated and flattened. We do not log channel names in hypothesis spaces anymore. Will need to update visualizations.
- Tests were modified and some new tests for the displacer were added.

Notes:
- I ran some preliminary benchmark tests as sanity checks, they ran fine. I'm planning to run the full benchmarks.
[...]

PR #955, opened 2026-04-28
refactor: rename informed hypotheses to new hypotheses in burst sampling updater

ramyamounir:
This PR renames the informed hypotheses to avoid confusion. This is following the discussion thread [here](https://github.com/thousandbrainsproject/tbp.monty/pull/851#discussion_r3060301758). The change is not breaking since the informed naming is private within the `BurstSamplingHypothesesUpdater` and is different from informed in [initial possible poses](https://github.com/thousandbrainsproject/tbp.monty/blob/dcd09bad0ed281b0d5f49bb8b10360b5959cf4b8/src/tbp/monty/frameworks/utils/graph_matching_utils.py?plain=1#L145-L171), and informed policies.


src/tbp/monty/frameworks/models/evidence_matching/burst_sampling.py
```diff
@@ -324,22 +324,22 @@ def update_hypotheses(
...
-        hypotheses_selection, informed_samples_per_channel = self._sample_count(
+        hypotheses_selection, new_hypotheses_per_channel = self._sample_count(
             features=features,
             graph_id=graph_id,
             input_channels=input_channels_to_use,
             tracker=tracker,
         )
 
-        existing_hypotheses = self._sample_existing(
+        existing_hypotheses = self._sample_existing_hypotheses_based_on_evidence_slopes(
```
jeremyshoemaker: I think this function name is too verbose. Something like `_sample_existing_hypotheses` is probably sufficient. Having the _how_ in the name forces it to be renamed if how we do it ever changes.

If I want to know how it does it, the docstring should tell me this additional information, if not that then the code itself.

src/tbp/monty/frameworks/models/evidence_matching/burst_sampling.py
```diff
@@ -324,22 +324,22 @@ def update_hypotheses(
...
             hypotheses_selection=hypotheses_selection,
             hypotheses=hypotheses,
             tracker=tracker,
         )
-        informed_hypotheses = self._sample_informed(
+        new_hypotheses = self._sample_new_hypotheses_based_on_current_observation(
```
jeremyshoemaker: I can make a similar argument here. Something like `_sample_new_hypotheses` is probably sufficient.

ramyamounir:
Thanks @jeremyshoemaker , I've updated the names to what you suggested.

jeremyshoemaker (approved):

PR #963, opened 2026-05-01
docs: all code links must be absolute to work

tristanls-tbp:
I noticed that our relative source code links don't work. I updated the ones I found to be absolute.


scottcanoe (approved):
That's kind of annoying. Wonder what happened.

Anyways, LGTM.

tristanls-tbp:
@scottcanoe, I looked at our tooling. I don't think these ever worked.

PR #965, opened 2026-05-05
refactor!: rename world_camera to cam_to_world

hlee9212:
Renames the variable name `world_camera` to `cam_to_world` (short for camera-to-world). 

This rename was motivated by several of us asking what `world_camera` variable was during [IP] 2D Sensor Module. This is defined in `DepthTo3DLocations` transforms. Previously, I had also mistakenly described it as world-to-camera transformation (which is the [Camera Extrinsic Matrix](https://en.wikipedia.org/wiki/Camera_resectioning)), but upon closer inspection it is actually its transpose, or camera-to-world transformation. 

Niels has suggested to update the name to make it clearer during 1:1, this PR is the result of that. 

Edit:

## The Breaking Change 

Marking as breaking as it changes the API, especially around SensorObservation.

### DIFF
--- a/src/tbp/monty/frameworks/environment_utils/transforms.py
+++ b/src/tbp/monty/frameworks/environment_utils/transforms.py
@@ -540,7 +540,7 @@ class DepthTo3DLocations(Transform):
                     reference frame and are in the sensor's reference frame otherwise.
                     It is structured as a 2D array with shape (n_pixels, 4) with
                     columns containing x-, y-, z-coordinates, and a semantic ID.
-                - "world_camera": Sensor-to-world coordinate frame transform. Included
+                - "cam_to_world": Sensor-to-world coordinate frame transform. Included
                     only when `self.world_coord` is `True` (default).
                 - "sensor_frame_data": 3D coordinates for each pixel relative to the
                     sensor. Has the same structure as "semantic_3d". Included only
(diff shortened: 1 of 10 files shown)


src/tbp/monty/frameworks/environment_utils/transforms.py
```diff
@@ -627,14 +627,14 @@ def call(
...
                 # Apply the rotation and translation to get the world coordinates
                 rotation_matrix = qt.as_rotation_matrix(sensor_rotation_rel_world)
-                world_camera = np.eye(4)
-                world_camera[0:3, 0:3] = rotation_matrix
-                world_camera[0:3, 3] = sensor_translation_rel_world
-                xyz = np.matmul(world_camera, xyz)
+                cam_to_world = np.eye(4)
+                cam_to_world[0:3, 0:3] = rotation_matrix
+                cam_to_world[0:3, 3] = sensor_translation_rel_world
+                xyz = np.matmul(cam_to_world, xyz)
```
jeremyshoemaker: This is what sparked the discussion earlier. If we decide we want to go with `@`, it might make sense to change this one while we're here.

jeremyshoemaker: I also might prefer using `identity` instead of `eye` above, but that's a separate discussion.

tests/unit/frameworks/environment_utils/habitat_transform_test.py
```diff
@@ -234,11 +234,11 @@ def compute_expected_semantic_3d(
 
         translation = agent_position + agent_rotation_matrix @ sensor_position
 
-        world_camera = np.eye(4)
-        world_camera[0:3, 0:3] = rotation_matrix
-        world_camera[0:3, 3] = translation
+        cam_to_world = np.eye(4)
```
jeremyshoemaker: If you change `eye` to `identity` above, here's another one. Looks like we're already using the `@` below.

jeremyshoemaker (approved):
Looks good. I have some optional changes below.

src/tbp/monty/frameworks/environment_utils/transforms.py
hlee9212: Added a commit to use `@` and also `np.identity`

PR #973, opened 2026-05-07
refactor: use monty Rotation instead of scipy's Rotation in mujoco simulator

scottcanoe:
This PR swaps out usage of the scipy Rotation object and associated utilities for `tbp.monty.geometry.Rotation` which was introduced in [feat!: wxyz-friendly Rotation object #937](https://github.com/thousandbrainsproject/tbp.monty/pull/937).

There are no breaking changes.

### DIFF
--- a/src/tbp/monty/simulators/mujoco/agents.py
+++ b/src/tbp/monty/simulators/mujoco/agents.py
@@ -16,7 +16,6 @@ from typing import TYPE_CHECKING, Protocol, cast
 import numpy as np
 import quaternion as qt
 from mujoco import MjsBody, mjtJoint
-from scipy.spatial.transform import Rotation
 
 from tbp.monty.frameworks.actions.actions import (
     LookDown,
--- a/src/tbp/monty/simulators/mujoco/agents.py
+++ b/src/tbp/monty/simulators/mujoco/agents.py
@@ -34,10 +34,7 @@ from tbp.monty.frameworks.models.abstract_monty_classes import (
 )
 from tbp.monty.frameworks.models.motor_system_state import AgentState, SensorState
 from tbp.monty.frameworks.sensors import SensorConfig, SensorID
-from tbp.monty.frameworks.utils.transform_utils import (
-    rotation_as_quat,
-    rotation_from_quat,
-)
+from tbp.monty.geometry import Rotation
 from tbp.monty.math import IDENTITY_QUATERNION, ZERO_VECTOR, QuaternionWXYZ, VectorXYZ
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -244,9 +238,9 @@ def move_along_local_axis(self, distance: float, axis: Axis) -> None:
     def yaw(self, delta_theta: float) -> None:
         """Yaw the embodiment by delta_theta degrees."""
         delta_theta_rot = Rotation.from_euler("xyz", (0, delta_theta, 0), degrees=True)
-        rotation = rotation_from_quat(self.rotation)
+        rotation = Rotation.from_quat(self.rotation)
         new_rotation = rotation * delta_theta_rot
-        self.rotation = rotation_as_quat(new_rotation)
+        self.rotation = new_rotation.as_quat()
```
scottcanoe, line 243: Should I be casting here to `QuaternionWXYZ`?

dschumacher-tbp: I thought the idea was to use plain tuples, or aliases such as `QuaternionWXYZ`, for interfaces initialized by Hydra, then convert them to richer nominal types like `Rotation` for internal data structures and local variables. This would create a virtuous wave of refactoring that avoids converting back and forth between raw data and rich domain objects.

jeremyshoemaker: It looks like `as_quat` is returning an `nd.array` and the `position` property is expecting a tuple, so there's a type mismatch here. I was just thinking through what I want to do about this in general.

For now though, this _at least_ needs to be turned into a tuple, and then the types won't match without a `cast` because `QuaternionWXYZ` is a `tuple[float, float, float, float]` and turning an `nd.array` into a tuple returns a `tuple[Any]`.

jeremyshoemaker: I think this code was incorrect originally, and I just didn't catch it. Like it works because both `nd.array` and `tuple` are indexable.

jeremyshoemaker (approved):
For now, I don't think I'd worry too much about the type mismatch in `yaw`. I'll sort it out when I sort out the rest of it.

src/tbp/monty/simulators/mujoco/agents.py
scottcanoe: Alright, sounds good. I'll leave you to any casting/cleanup.

To @dschumacher-tbp's point, I'm assuming we'll do that at some point. Maybe in conjunction with eliminating `quaternion.quaternion` from motor system / proprioceptive state? I haven't thought about that much or talked to y'all about your thoughts and plans on that matter.

tristanls-tbp:
**note**: I thought transform utils were supposed to be deleted as part of this.

PR #979, opened 2026-05-11
refactor!: sensor_module.states is unused and removed

tristanls-tbp:
When considering how to refactor sensor modules during the engineering meeting today, we noticed that the `.states` attribute of sensor modules is unused.

This pull request removes the unused attribute.


vkakerbeck (approved):
LGTM

PR #980, opened 2026-05-12
docs: consistent use of "orientation"

dschumacher-tbp:
In the glossary entry for Reference Frame, use the terms "location" and "orientation" like they are used to define Pose.


scottcanoe (approved):
LGTM, appreciate the update.

PR #982, opened 2026-05-13
fix: principal_curvatures is from non_morphological_features 

hlee9212:
I noticed a small issue in `TwoDSensorModule` were we were accessing `principal_curvatures` from `Message.morphological_features` instead of `Message.non_morphological_features`, which led to this value being always False. Updating so we correctly access this feature value.


tests/unit/frameworks/models/two_d_sensor_module_test.py
```diff
@@ -56,7 +56,7 @@ def make_message(
     if pose_vectors is None:
         pose_vectors = np.identity(3)
     if principal_curvatures is None:
-        principal_curvatures = np.identity(3)
+        principal_curvatures = np.zeros(2)
```
jeremyshoemaker, line 59: I'm curious why this change was needed, especially given that the size has changed. I'm assuming it was just wrong before, but I'm not clear on why it was passing before. I tried playing around with the values to figure it out but it doesn't seem obvious.

jeremyshoemaker (approved):
I do have a question, but it's non-blocking.

src/tbp/monty/frameworks/models/two_d_sensor_module.py
```diff
@@ -338,7 +338,7 @@ def _update_2d_position_and_displacement(
         du = np.dot(d_tan, self._tangent_frame.basis_u)
         dv = np.dot(d_tan, self._tangent_frame.basis_v)
 
-        principal_curvatures = observed_state.morphological_features.get(
+        principal_curvatures = observed_state.non_morphological_features.get(
```
tristanls-tbp: **issue**: Please use `observed_state.get_feature_by_name`.

tristanls-tbp (changes requested):

src/tbp/monty/frameworks/models/two_d_sensor_module.py
hlee9212: Fixed in ce5f859a0a9a87b270392ad122df56f91c4018e8.

tests/unit/frameworks/models/two_d_sensor_module_test.py
hlee9212: Hi @jeremyshoemaker, I think you may have caught this on during stand-up today, but for posteriority:

1. `principal_curvatures` refer to the scalar values (k1, k2), not the directions (which are `pc1_dir` and `pc2_dir`), hence `np.zeros(2)` is the correctly shaped default representing k1 = 0 and k2 = 0. 
2. Because `principal_curvatures` was None before (in line 341 of `two_d_sensor_module.py`, we were never calling the following if block:
```python
if principal_curvatures is not None and pose_3d is not None:
    du, dv = arc_length_corrected_displacement(
        du,
        dv,
        self._tangent_frame.basis_u,
        self._tangent_frame.basis_v,
        principal_curvatures,
        pose_3d,
    )
```
Which is why the tests passed before.

hlee9212:
**Note**: I'm going to hold off on merging this PR until I resolve the RuntimeError (`ValueError: The pc1_dir and pc2_dir must be orthogonal.`) when running an experiment that uses `TwoDSensorModule`. 

I will create a separate PR for the above issue as it touches `principal_curvatures()` function in `sensor_processing.py`, which is outside the scope of 2D SM work and was a known issue based on:
```python
# Line 553 of sensor_processing.py
# TODO: sometimes dir1 and dir2 are not orthogonal, why?
```

tristanls-tbp (approved):

hlee9212:
Merging this since #1008 will prevent RuntimeError.

PR #984, opened 2026-05-13
refactor!: extract evidence sdr similarity measure code

tristanls-tbp:
This pull request proposes extracting the evidence SDR similarity measure and its associated code. The new home for the code would be at https://github.com/thousandbrainsproject/feat.evidence_sdr_matching (to be made public if this extraction is accepted).

[The code is listed as a "research feature" in the documentation](https://thousandbrainsproject.readme.io/docs/evidence-as-a-similarity-measure) and is unused in any of the benchmarks.

Because this code is unused by any benchmarks, it is unclear whether any refactors we make will improve or degrade performance. As we refactor configurations and the platform, maintaining this code without any performance feedback loops feels toilsome and lacks a clear payoff.

I propose that this code be extracted to https://github.com/thousandbrainsproject/feat.evidence_sdr_matching. If and when we return to SDR matching, that repository can serve as a prototype starting point, and once we decide to reintegrate it, we will then undertake an implementation project to do so.

## Breaking Changes
- Evidence SDR similarity measure and associated code are deleted.

## Benchmarks

Benchmarks are pending completion, but none of them crashed, indicating that none of them are configured with deleted code.


ramyamounir (approved):
Looks great, thanks @tristanls-tbp.

Suggestion: It might be more fitting to call it `feat.evidence_sdr_matching` because I think it's more of a research prototype (even if it's not an actual fork of `tbp.monty`).

tristanls-tbp:
Thanks @ramyamounir, I updated to https://github.com/thousandbrainsproject/feat.evidence_sdr_matching per your recommendation.

vkakerbeck (approved):
This makes sense, thanks for making the change.

tristanls-tbp:
**note**: FYI @ramyamounir @vkakerbeck , https://github.com/thousandbrainsproject/feat.evidence_sdr_matching repository has been updated to `tbp.monty==0.35.0` and now unit tests (`pytest`) and integration tests (`pytest tests/integration`) pass there 🙌.

PR #987, opened 2026-05-14
feat: add dict config support to logging

jeremyshoemaker:
[Python logging can usually be configured using config files or dictionary configurations](https://docs.python.org/3/library/logging.config.html), e.g. how the Django framework configures logging.

One of the things this unlocks is things like turning on DEBUG level logging for a single module, rather than the entire application. I realized we can just add the configuration to our Hydra configs and load it from there, so this is the initial first step of that.

An example of using this was I wanted to turn on DEBUG logging for just the MuJoCo simulator, so to do that, I'd just need to add this to my experiment configuration:

```yaml
config:
  logging:
    dict_config:
      loggers:
        tbp.monty.simulators.mujoco:
          level: DEBUG
```

Ideally, we should probably move all of our logging configuration logic to use this more standard approach, but that seems like a side-quest I shouldn't embark on just yet.


jeremyshoemaker:
I'm not sure if we actually should merge this, but I think it could be useful enough as it is to suggest it.

tristanls-tbp:
**praise**: Great. I thought we might need a separate file (`fileConfig`), but this looks promising.

**question**: I'm looking at what properties we use inside `logging` and none of them seem to conflict with what `dictConfig` would use. Have you tried just dumping the params into `logging` and see if that works? This way, we wouldn't need a separate "dict_config" namespace and there'd be less config refactoring in the future. As you mentioned, a programmatic `version: 1` may be needed here for the existing configs to work.

tristanls-tbp:
**note**: FYI @AgentRev... probably a good way to set up for telemetry configuration as well, by putting it in `config.telemetry` and using the `dictConfig` mechanism.

jeremyshoemaker:
> **praise**: Great. I thought we might need a separate file (`fileConfig`), but this looks promising.

I thought it would be more complicated too, but I noticed the `dictConfig` function and realized I could just pass it something from Hydra.

> **question**: I'm looking at what properties we use inside `logging` and none of them seem to conflict with what `dictConfig` would use. Have you tried just dumping the params into `logging` and see if that works? This way, we wouldn't need a separate "dict_config" namespace and there'd be less config refactoring in the future. As you mentioned, a programmatic `version: 1` may be needed here for the existing configs to work.

I hadn't tried that, mostly because I was trying to keep it simple and focused for now, but I can take some time today to see how far I can get on unifying everything.

jeremyshoemaker:
I don't think moving it over is going to be easy. There are things like `output_dir`, `run_name`, and `wand_id` that don't easily map to typical logging configuration.

It's really all the wandb logging configuration that's a special snowflake, see [`init_monty_data_loggers`](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/experiments/monty_experiment.py#L343) where we seem to be working against the grain. The `monty_log_level` is its own custom snowflake that does things not typical of logging configuration.

I could probably easily convert the code in [`init_loggers`](https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/experiments/monty_experiment.py#L299) over but that mostly seems to be hard coded.

In the end, I think we'll still need to have some sort of `config` or `dict_config` key in the `logging` to do this, even if we moved everything over, again because of things like `output_dir` and `run_name`, that get pulled in programmatically to set things up custom, don't fit the [configuration schema](https://docs.python.org/3/library/logging.config.html#configuration-dictionary-schema).

jeremyshoemaker:
With [e6df35e](https://github.com/thousandbrainsproject/tbp.monty/pull/987/commits/e6df35e248b734c52db95939dbe60c5de9c3513e), the configuration example I gave simplifies to:

```yaml
config:
  logging:
    dict_config:
      loggers:
        tbp.monty.simulators.mujoco:
          level: DEBUG
```

tristanls-tbp:
This works on my machine... does it not work for you?

```diff
diff --git a/src/tbp/monty/frameworks/experiments/monty_experiment.py b/src/tbp/monty/frameworks/experiments/monty_experiment.py
index d55e7157..762214e6 100644
--- a/src/tbp/monty/frameworks/experiments/monty_experiment.py
+++ b/src/tbp/monty/frameworks/experiments/monty_experiment.py
@@ -338,23 +338,18 @@ class MontyExperiment:
         for handler in python_logging_handlers:
             logger.addHandler(handler)
 
+        final_config = {
+            "version": 1,
+            # TODO: This can be removed if we configure all the loggers via
+            #   the dictConfig call below.
+            "incremental": True,
+        }
+        final_config.update(logging_config)
+        logging.config.dictConfig(final_config)
+
         logger.info("logger initialized")
         logger.debug(pprint.pformat(self.config))
 
-        # TODO: Migrate the rest of the configuration to use this
-        if "dict_config" in logging_config:
-            # Specify some defaults so we don't have to remember to configure
-            # them in Hydra.
-            final_config = {
-                "version": 1,
-                # TODO: This can be removed if we configure all the loggers via
-                #   the dictConfig call below.
-                "incremental": True,
-            }
-            dict_config = logging_config["dict_config"]
-            final_config.update(dict_config)
-            logging.config.dictConfig(final_config)
-
     def init_monty_data_loggers(self, logging_config: dict[str, Any]) -> None:
         """Initialize Monty data loggers.
```

jeremyshoemaker:
> This works on my machine... does it not work for you?

For certain values of "works". I wasn't expecting `dictConfig()` to just ignore unknown keys, but it seems to.

I'm a little concerned about mixing up values in that dictionary though. There's nothing to really save us from a conflict in keys. I guess it would just break, hopefully in an obvious way, but I don't like the idea.

tristanls-tbp:
It just bugs me that with `dict_config` field we basically have:

```yaml
config:
  logging:
    config: #really dict_config
      logging: #really loggers
```

I'd like to live in the future with a configuraiton that looks like:
```yaml
config:
  logging:
    formatters:
    filters:
    loggers:
    handlers:
    root:
    monty_log_level: # replaced by telemetry
    monty_loggers: # replaced by telemetry
    wandb_loggers: # replaced by telemetry
    log_to_file: # replaced by a handler config
    log_to_stderr: # replaced by a handler config
    run_name: # I think this will remain
    output_dir: # subsumed by handlers
```

After you introduce this, there will be no reason to add new fields to `config.logging`. So if our existing fields don't conflict, I don't see us adding any new ones. I'm hoping we can just strangle pattern the legacy `config.logging` params out of existence. For example, `python_log_level` immediately becomes `config.logging.root.level`.

tristanls-tbp (approved):
I'm going to approve this so that we get the benefits, but I hope you consider putting everything together into the `config.logging` property in the configuration.

jeremyshoemaker:
> I'm going to approve this so that we get the benefits, but I hope you consider putting everything together into the `config.logging` property in the configuration.

I'll go ahead and move it up, because I also don't like adding in the `dict_config`. I think you're right about where things will end up.

PR #989, opened 2026-05-15
refactor!: no need for input_channel in FeatureEvidenceCalculator

tristanls-tbp:
With #984 merged, there is no longer a need for `input_channel` in the `FeatureEvidenceCalculator` protocol. This pull request removes it.


ramyamounir (approved):
👍

PR #993, opened 2026-05-19
test: account for threshold ambiguity near rotation tolerance threshold

tristanls-tbp:
Explicitly account for numerical ambiguity as we get infinitessimally closer to rotation tolerance threshold.

Resolves
```

=================================== FAILURES ===================================
_ ScipyRotationsApproxEqualTest.test_against_alternate_implementation_near_boundary _
[gw3] linux -- Python 3.8.18 /home/runner/miniconda/envs/tbp.monty/bin/python

self = <tests.unit.geometry_test.ScipyRotationsApproxEqualTest testMethod=test_against_alternate_implementation_near_boundary>

    @given(
>       a=scipy_rotations(),
        axis=rotation_axes(),
        angle=st.floats(
            min_value=0.9 * ROTATION_TOLERANCE_RADIANS,
            max_value=1.1 * ROTATION_TOLERANCE_RADIANS,
        ),
    )

tests/unit/geometry_test.py:238: 
tests/unit/geometry_test.py:255: in test_against_alternate_implementation_near_boundary
    self.assertEqual(actual, expected)
E   AssertionError: True != False
E   Falsifying example: test_against_alternate_implementation_near_boundary(
E       self=<tests.unit.geometry_test.ScipyRotationsApproxEqualTest testMethod=test_against_alternate_implementation_near_boundary>,
E       a=(lambda angles: <unknown>)(array([1.5, 1.5, 1.5])),
E       axis=array([1., 0., 0.]),
E       angle=1.0000000000001e-06,
E   )
E   
E   You can reproduce this example by temporarily adding @reproduce_failure('6.113.0', b'AXicY2BgYGhgAANGBlRw0Cq/6kLcNwMAIssEug==') as a decorator on your test case
```


scottcanoe (approved):
Approving, but with a thought/suggestion.

We've been adjusting `ROTATION_TOLERANCE_THRESHOLD` to accommodate a numerical quirk that exists only in these approx_equal tests. That now seems off to me, since I don't think the way we construct examples for this test should have implications for anything outside of that context.

As I understand it, this PR places part of the test-specific tolerance where it belongs but still leaves the rest in `ROTATION_TOLERANCE_THRESHOLD`. Can/should we just put all of the "extra" tolerance in this assumption and return `ROTATION_TOLERANCE_THRESHOLD` to 1e-6?

tristanls-tbp:
Thanks @scottcanoe, good point. I returned the rotation tolerance back to the original `1e-6`.

PR #992, opened 2026-05-18
refactor!: remove `pre_episode` from LMs

dschumacher-tbp:
Split `LearningModule.pre_episode` into `reset_stm` and `fixme_reset_ground_truth`,
and remove `reset()` from `LearningModule` public interface.

## Breaking Changes

- Callers must replace calls to `pre_episode` with calls to `reset_stm` and `fixme_reset_ground_truth`.
- `reset` is no longer part of the `LearningModule` public interface, but is often inherited from `GraphLM`.

### base_config_10distinctobj_dist_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100 | 100 | 0 | ⬜ |
| used_mlh (%) | 2.86 | 2.86 | 0 | ⬜ |
| match_steps | 34 | 34 | 0 | ⬜ |
| rotation_error (deg) | 8.64 | 8.64 | 0 | ⬜ |
| runtime (min) | 3 | 3 | 0 | ⬜ |
| episode_runtime (sec) | 7 | 8 | 1 | ❌ |
| num_episodes | 140 | 140 | 0 | ⬜ |

### base_config_10distinctobj_surf_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100 | 100 | 0 | ⬜ |
| used_mlh (%) | 0 | 0 | 0 | ⬜ |
| match_steps | 28 | 28 | 0 | ⬜ |
| rotation_error (deg) | 3.87 | 3.87 | 0 | ⬜ |
| runtime (min) | 3 | 3 | 0 | ⬜ |
| episode_runtime (sec) | 11 | 12 | 1 | ❌ |
| num_episodes | 140 | 140 | 0 | ⬜ |

### randrot_noise_10distinctobj_dist_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100 | 100 | 0 | ⬜ |
| used_mlh (%) | 2 | 2 | 0 | ⬜ |
| match_steps | 35 | 35 | 0 | ⬜ |
| rotation_error (deg) | 13.23 | 13.23 | 0 | ⬜ |
| runtime (min) | 3 | 3 | 0 | ⬜ |
| episode_runtime (sec) | 14 | 14 | 0 | ⬜ |
| num_episodes | 100 | 100 | 0 | ⬜ |

[...]

PR #990, opened 2026-05-15
refactor!: FeatureEvidenceScorer encapsulation toward better DI

tristanls-tbp:
This work is partial progress toward removing `_class`/`_args` convention from configurations. The `_class`/`_args` convention is being replaced with dependency injection, but to do that, our code must be able to accept injected dependencies. Therefore...

This pull request moves the code architecture closer to where dependencies can be injected at construction time. The name `FeatureEvidenceScorer` is just a name for the newly encapsulated concept for checking if the feature should be used for matching, then calculating evidence for a given channel, and finally multiplying it by the configured evidence increment.

The best way to see the improvement is to notice that we removed three repetitions of essentially the same code:
```python
if self.use_features_for_matching[input_channel]:
    node_feature_evidence = self.feature_evidence_calculator.calculate(
        channel_feature_array=self.graph_memory.get_feature_array(graph_id)[
            input_channel
        ],
        channel_feature_order=self.graph_memory.get_feature_order(graph_id)[
            input_channel
        ],
        channel_feature_weights=self.feature_weights[input_channel],
        channel_query_features=features,
        channel_tolerances=self.tolerances[input_channel],
    )
    # ... few custom things
else:
    # ... zero case
```
and replaced that with 
```python
node_feature_evidence = self._feature_evidence_scorer(
    graph_id=graph_id,
    input_channel=input_channel,
    query_features=features,
)
```

The other thing to note is that various parameters are encapsulated in the scorer and no longer need to be maintained by the updater and displacer.

More refactors will follow.

## Breaking Changes

- Signature of the `DefaultHypothesesDisplacer` constructor is changed to require `FeatureEvidenceScorer` and no longer requires `tolerances`, `use_features_for_matching`, `feature_evidence_calculator`, nor `feature_evidence_increment`.

## Benchmarks

No changes

| Experiment | Metric | Baseline | Proposed | Change |
| --- | --- | --- | --- | --- |
| base_config_10distinctobj_dist_agent | percent_correct (%) | 100 | 100 | 0 |
|  | used_mlh (%) | 2.86 | 2.86 | 0 |
|  | match_steps | 34 | 34 | 0 |
|  | rotation_error (deg) | 8.64 | 8.64 | 0 |
|  | runtime (min) | 3 | 3 | 0 |
|  | episode_runtime (sec) | 7 | 8 | 1 |
|  | num_episodes | 140 | 140 | 0 |
| base_config_10distinctobj_surf_agent | percent_correct (%) | 100 | 100 | 0 |
[...]

PR #995, opened 2026-05-21
chore: version 0.36.0

jeremyshoemaker:
Minor release version due to 0 Major version and breaking changes since 0.35.0.

image

## refactor!: no need for input_channel in FeatureEvidenceCalculator (#989)
Removes `input_channel` from the `FeatureEvidenceCalculator` protocol.

## refactor!: FeatureEvidenceScorer encapsulation toward better DI (#990)
Signature of the `DefaultHypothesesDisplacer` constructor is changed to require `FeatureEvidenceScorer` and no longer requires `tolerances`, `use_features_for_matching`, `feature_evidence_calculator`, nor `feature_evidence_increment`.

## refactor!: remove pre_episode from LMs (#992)
Callers must replace calls to `pre_episode` with calls to `reset_stm` and `fixme_reset_ground_truth`.
The `reset` method is no longer part of the `LearningModule` public interface, but is often inherited from `GraphLM`.

### DIFF
--- a/src/tbp/monty/__init__.py
+++ b/src/tbp/monty/__init__.py
@@ -8,7 +8,7 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
-__version__ = "0.35.0"
+__version__ = "0.36.0"
 
 import sys
 


jeremyshoemaker:
@tristanls-tbp, It's not clear to me why #990 is marked as a breaking change. Do you have a succinct description of what it's breaking? I didn't review it when it went in, so I don't have a lot of the context. Looking at the change now, it seems like an internal refactor, but maybe I'm missing something obvious.

tristanls-tbp (approved):
Apologies @jeremyshoemaker. I updated #990 with a `Breaking Changes` section and updated this PR's description with the breaking change.

PR #997, opened 2026-05-26
fix: always emit edge_strength and coherence from 2D SM

ramyamounir:
When an edge is not detected, we don't attach `edge_strength` or `coherence` to the message sent to the LM. This causes issues with the buffer length of features and padding when using a `GridObjectModel` to store the models. Normally, the buffer will pad certain channels based on `on_objects` to make them equal in length, but it will not pad different feature lists within a channel. If we define `edge_strength` and `coherence` in the feature list of the SM, I think we should send them in every message.

This PR refactors the code a bit to send 0.0 for strength and coherence if we don't detect an edge. Please let me know if you see a problem with this approach.


hlee9212 (approved):
Thanks Ramy! Sounds good to me and the values (0.0) makes sense.

PR #1001, opened 2026-05-28
fix: better detect off-center geometric edges

hlee9212:
Hi @ramyamounir , I wanted to share an update related to `is_geometric_edge()` to better remove edge between object and the black void. 

The effect is as follows:

Previously, we had some extra edges detected (see blue rectangle):
CleanShot 2026-05-28 at 10 25 54

This PR improves the detection of `is_geometric_edge()` so we get rid of those:
021_logo_tbp

I don't think it would affect the compositional experiments you are running too much in terms of performance.


src/tbp/monty/frameworks/utils/edge_detection.py
```diff
@@ -418,7 +434,19 @@ def _is_geometric_edge(
         nx = np.cos(edge_normal_angle)
         ny = np.sin(edge_normal_angle)
 
-        cy, cx = depth_patch.shape[0] // 2, depth_patch.shape[1] // 2
-        depth_gradient_perp = abs(nx * depth_dx[cy, cx] + ny * depth_dy[cy, cx])
+        h, w = depth_patch.shape
+        cy, cx = h // 2, w // 2
+        rows, cols = np.meshgrid(np.arange(h), np.arange(w), indexing="ij")
+        dist_normal = nx * (cols - cx) + ny * (rows - cy)
+        depth_edge_band = np.abs(dist_normal - edge_center_offset) <= 1.5
```
ramyamounir, line 441: Do you think this `1.5` and the `95.0` percentile values need to be constructor parameters or constants?

ramyamounir (approved):
It all looks good to me given my limited context of the inner workings of the 2D SM edge detection. Thanks for making the changes.

src/tbp/monty/frameworks/utils/edge_detection.py
hlee9212: Hi Ramy, I originally put them as constructor parameters, then changed to global variables, then to the above, mostly for simplicity's sake. I can't think of a case where we need to change the two values - but of course someone in the future may refactor them out.

PR #1005, opened 2026-05-29
rfc: Sensor Modules as Transformation Pipeline

tristanls-tbp:
Implementing a new Sensor Module should require only adding necessary and sufficient code corresponding to the new functionality. To that end, every Sensor Module can be implemented as a pipeline of reusable transforms.


rfcs/0000_sensor_modules_as_transformation_pipeline.md
```diff
@@ -0,0 +1,943 @@
...
+
+```python
+class RuntimeTransform(Protocol):
+    def __call__(
+        self: Self,
+        ctx: TransformContext,
+        observation: SensorObservation,
+        percept: Message | None,
+        goals: list[Goal],
+    ) -> tuple[SensorObservation, Message | None, list[Goal]]:
```
jeremyshoemaker: What do you think about making this take a data structure, most likely a dataclass, that combines these three elements? That way the third argument and the return type can be the same type.

rfcs/0000_sensor_modules_as_transformation_pipeline.md
```diff
@@ -0,0 +1,943 @@
...
+
+Transferring more of the sensor module "business logic" to configuration results in system behavior being defined in the configuration. This puts it outside the domain of unit testing. The individual transforms can be unit tested, improving their individual reliability. However, verifying how the configured ensemble of transforms behaves now falls onto integration testing and benchmarks.
+
+## Verbosity
+
+While it is almost pleasant to read a single sensor module configuration laid out as a pipeline of transforms, this will get old when we have many sensor modules, especially if the configuration is repetitive. Hydra is uniquely terrible at allowing us to reuse portions of lists in a configuration. The general rule is that, when Hydra combines configuration fragments, objects are composed/combined, whereas lists are replaced in their entirety. We will likely have to handle this by creating some bespoke resolves that reduce the toil of repeating a configuration over and over again. Note that this issue is not unique due to the changes in this RFC, but also exists (or will exist after some changes) for specifying multiple learning modules.
```
jeremyshoemaker, line 791: There is something to be said for being explicit about what all is in the pipeline rather than having to look in several places and reconstruct it in your head. We discussed this the other day, and with further thought I think I'm actually okay with this.

jeremyshoemaker (commented):
Sounds reasonable to me. Just a couple comments I have.

rfcs/0000_sensor_modules_as_transformation_pipeline.md
tristanls-tbp: That makes a lot of sense. 

```python
@dataclass
class Payload:
    observation: SensorObservation
    percept: Message | None
    goals: list[Goal]

class RuntimeTransform(Protocol):
    def __call__(
        self: Self,
        ctx: TransformContext,
        payload: Payload,
    ) -> Payload:
```

jeremyshoemaker: Payload is a rather generic name, but I'm having trouble coming up with something better.

tristanls-tbp: Screenshot 2026-06-12 at 13 41 44

tristanls-tbp: I'll add that this is gonna be in the vicinity of `tbp.monty.sensor_module.Payload`

jeremyshoemaker: Yeah, I was thinking that too. I'm fine with Payload.

tristanls-tbp: I updated the RFC with the `Payload` dataclass.

rfcs/0000_sensor_modules_as_transformation_pipeline.md
```diff
@@ -0,0 +1,901 @@
...
+
+    def __call__(self: Self, ctx: TransformContext, payload: Payload) -> Payload:
+        # ...
+        return self._next_transform(ctx, payload)
+
+    def reset(self: Self) -> None:
+        # override the default no-op reset if needed ...
+```
+
+Note that it is possible to "early exit" out of the chain of transforms by not calling `self._next_transform`. For example:
```
vkakerbeck: This sounds like it might be a bit error-prone. We assume the user reads the documentation/code well enough to know to return next_transform() instead of just returning the result of the transform (which I think is what is the more intuitive thing someone would default to). It might cause unintended early exits from the transform pipeline. I am not sure there is a good way to avoid this but just highlighting my concern here. Maybe, if there is no specific use case for this feature, there could be a warning or something (at least until there is a usecase)

rfcs/0000_sensor_modules_as_transformation_pipeline.md
```diff
@@ -0,0 +1,901 @@
...
+### Filters
+
+Filter transforms' main purpose is to filter. As an example, implementation of the `FeatureChangeFilter` as a `Transform` would look as follows. Note that we provide a non-default `reset()` method that resets the filter:
+
+```python
+class FeatureChangeFilter(Transform):
+
+    _next_transform: Transform
+    _delta_thresholds: dict[str, Any]
+    _last_percent: Message
```
vkakerbeck:
```suggestion
    _last_percept: Message
```

rfcs/0000_sensor_modules_as_transformation_pipeline.md
```diff
@@ -0,0 +1,901 @@
...
+
+# Rationale and alternatives
+
+The main rationale behind this design is that it is a well exercised industry method of handling configurable and flexible data-plane-like computation. In essence, this RFC implements the [Chain of Responsibility behavioral pattern](https://refactoring.guru/design-patterns/chain-of-responsibility), also known as middleware. The configurable and flexible aspects of the solution are appealing from research perspective. Reusing existing functionality and only authoring truly new functionality should speed up the research feedback cycle. Similarly, reusing existing functionality and only authoring truly new functionality improves engineering testing and maintenance. Reusing already tested components is simpler than creating new ones. There is also less code to maintain.
```
vkakerbeck:
```suggestion
The main rationale behind this design is that it is a well exercised industry method of handling configurable and flexible data-plane-like computation. In essence, this RFC implements the [Chain of Responsibility behavioral pattern](https://refactoring.guru/design-patterns/chain-of-responsibility), also known as middleware. The configurable and flexible aspects of the solution are appealing from a research perspective. Reusing existing functionality and only authoring truly new functionality should speed up the research feedback cycle. Similarly, reusing existing functionality and only authoring truly new functionality improves engineering testing and maintenance. Reusing already tested components is simpler than creating new ones. There is also less code to maintain.
```

rfcs/0000_sensor_modules_as_transformation_pipeline.md
```diff
@@ -0,0 +1,901 @@
...
+
+It is unclear how toilsome it will be to configure many sensor modules and how much sensor-modules-via-configuration approach will contribute to the toil.
+
+# Future possibilities
+
+## Experiment could reconstruct Monty between each episode
+
+The generic Sensor Module approach described in this RFC works within the contemporary experimental framework limitations. That is, the contemporary experimental framework relies on much Monty internals having experiment-specific methods to facilitate resets and related experimental activity.
+
+One pervasive use case for experiment-specific methods throughout Monty is to reset all of Monty between each experimental episode and epoch, except for the models learned inside the learning modules. An alternative approach would be to reconstruct all of Monty between each episode and epoch. The only desired persistence between the episodes are the models learned inside the learning modules. If Monty could be instantiated from a snapshot containing only learning module memory, this would allow for removal of experimental code from within Monty.
```
vkakerbeck, line 823: side note: As far as I remember, the unsupervised inference (https://docs.thousandbrains.org/docs/benchmark-experiments#unsupervised-inference) experiments don't reset LMs between episodes. But it sounds like it probably wouldn't be too complicated to just not reinstantiate Monty between episodes but still reset the environment in this future possible setup.

vkakerbeck (commented):
Really nice RFC :) I like the proposal.

rfcs/0000_sensor_modules_as_transformation_pipeline.md
jeremyshoemaker: This is pretty similar to [how Django's middleware works](https://docs.djangoproject.com/en/6.0/topics/http/middleware/), though in their case they have to work with an HTTP call and response cycle, so each one seems to call the next one (the mechanics of how it works are different than a direct call) but gets the response and returns it on the way back out. That's mainly so they can inject code on the way in and on the way back out, which we don't have here.

An alternative here might be to have each transform return its result instead of calling the next transform and have the machinery of running the transforms handle whether to call the next one with that result or return it early. That's actually kinda similar to how Django does it (I keep using them as an example, because it's something a lot of Python programmers would probably be familiar with).

That would probably mean adding a flag to the return value, here the `Payload`, that would indicate whether to continue or return early. That would allow for logging a warning if we want to provide that option. Another thing this approach would support is catching errors in each transform at that in-between step. Not sure if we need that or not, though.

It might also be more obvious than the continuation passing style that this version follows, since that's not as familiar to a lot of programmers, and is a point where it diverges from how other libraries do it.

AgentRev:
This is well thought-out, I agree with the proposal. It is quite reminiscent of node-based workflows in apps like Blender and DaVinci Resolve, so there is plenty of precedent beyond web dev.

I've worked on microscopy software for many years, and we had a similar pipeline. "Image providers" (analogous to sensor modules) communicated with the drivers of various USB cameras and PCIe capture cards to obtain raw frame buffers, which were then fed thru a hierarchy of transforms, customizable by the researcher. They could even be written in a variety of languages, including Swift, C, Objective-C, and Python.

rfcs/0000_sensor_modules_as_transformation_pipeline.md
tristanls-tbp: I acknowledge the concern.

I acknowledge the alternative implementation, but it's six on the one hand, half a dozen on the other, support for a non-existent use case.

I updated the RFC to not include the early-exit use case. The code is simpler. `TransformMiddleware` and `TransformPipeline` are gone. `SensorModule` calls transforms in a loop. Configuration is less verbose and no longer needs partials.

rfcs/0000_sensor_modules_as_transformation_pipeline.md
tristanls-tbp: Yes, the problem is resetting Monty. The experiment is free to reset itself and the environment as deemed necessary.

rfcs/0000_sensor_modules_as_transformation_pipeline.md
vkakerbeck: Thank you, I like this change! It is more intuitive for me to read and understand.

rfcs/0000_sensor_modules_as_transformation_pipeline.md
```diff
@@ -0,0 +1,823 @@
...
+            rotation=agent.rotation * sensor.rotation,
+        )
+
+    def reset(self: Self) -> None:
+        for transform in self._transforms:
+            transform.reset()
+
+```
+
+It is worth highlighting in this implementation how the type of the `Goal` collection is handled. Internally, each transform invocation receives a payload with a _mutable_ `list[Goal]` and returns a payload with a _mutable_ `list[Goal]`. This way, any transform in the chain can edit the collection of `Goal`s as required by the transform. Initially, the transform pipeline is handed a payload with an empty list `[]` and, at the end, the pipeline returns a payload with a possibly non-empty `list[Goal]` to the sensor module, which is stored at `self._goals`. However, when the `Goal`s finally exit the sensor module via `propose_goals()`, they exit as a read-only `Collection[Goal]`.
```
dschumacher-tbp, line 165: Do we need to commit to a concrete type like `list`? If internal mutability is important, perhaps the abstraction should be `MutableSequence`. This is a "collecting parameter" pattern, where a mutable object accumulates changes/additions within each stage of the pipeline.

Alternately, we could use something closer to an immutable "monadic transformer" pattern. Each stage takes an immutable input type like "Sequence" (or maybe "Collection"). Internally, a mutable copy can be created, then frozen input an output "Sequence" to be passed to the next stage. This also provides an easy transactional (or undoable) "original state", in case the transform wants to revert to the previous state, as it may do if it detects an error. Otherwise there is the risk of partial mutation leaving the object in an inconsistent state.

rfcs/0000_sensor_modules_as_transformation_pipeline.md
```diff
@@ -0,0 +1,823 @@
...
+        return payload.percept
+
+    def update_state(self: Self, agent: AgentState) -> None:
+        """Update information about the sensor's location and rotation."""
+        self._agent_state = agent
+        sensor = agent.sensors[self._sensor_id]
+        self._sensor_state = SensorState(
+            position=agent.position
+            + qt.rotate_vectors(agent.rotation, sensor.position),
+            rotation=agent.rotation * sensor.rotation,
```
dschumacher-tbp, line 156: suggestion: this may be a good place to use `tbp.monty.geometry.Rotation`

dschumacher-tbp (commented):
I love the idea of a configurable transformation pipeline composed of reusable components. Each component could be made more independent and reliable by passing immutable payloads between components, rather than mutating a payload to accumulate results.

rfcs/0000_sensor_modules_as_transformation_pipeline.md
jeremyshoemaker: I also like this simplification. If we need the more complex behaviour we can add it later. This keeps it straightforward for now.

dschumacher-tbp: I also think this is an improvement. If additional flow-control is needed (in future use-cases), control parameters can be added to the payload (as Jeremy mentioned).

rfcs/0000_sensor_modules_as_transformation_pipeline.md
tristanls-tbp: Can you help me understand why we shouldn't commit to a concrete type like `list`? What use case would make us regret this decision? Aside from a list, the only common things are byte arrays, arrays of numeric values, and UserList from collections. None of these will be useful for our case of passing along Goals.

> Mutable sequences
> 
>   Mutable sequences can be changed after they are created. The subscription and slicing notations can be used as the target of assignment and [del](https://docs.python.org/3.8/reference/simple_stmts.html#del) (delete) statements.
> 
>   There are currently two intrinsic mutable sequence types:
> 
>   Lists
> 
>   - The items of a list are arbitrary Python objects. Lists are formed by placing a comma-separated list of expressions in square brackets. (Note that there are no special cases needed to form lists of length 0 or 1.)
>
>   Byte Arrays
> 
>   - A bytearray object is a mutable array. They are created by the built-in [bytearray()](https://docs.python.org/3.8/library/stdtypes.html#bytearray) constructor. Aside from being mutable (and hence unhashable), byte arrays otherwise provide the same interface and functionality as immutable [bytes](https://docs.python.org/3.8/library/stdtypes.html#bytes) objects.
> 
>   The extension module [array](https://docs.python.org/3.8/library/array.html#module-array) provides an additional example of a mutable sequence type, as does the [collections](https://docs.python.org/3.8/library/collections.html#module-collections) module.
>
> -- https://docs.python.org/3.8/reference/datamodel.html#the-standard-type-hierarchy

dschumacher-tbp: I'm biased toward least-required-commitments in defining reusable interfaces. The `MutableSequence` seems like it defines our minimal expectations, without requiring it to be `list`.

However, to be clear, I'd rather be using **immutable** objects like `Sequence` whenever possible.

rfcs/0000_sensor_modules_as_transformation_pipeline.md
```diff
@@ -0,0 +1,823 @@
...
+    goals: list[Goal]
+
+@dataclass
+class TransformContext:
+    rng: np.random.RandomState
+    state: AgentState | None = None
+    motor_only_step: bool = False
+```
+
+Additionally, we have a `tbp.monty.experiment.sensor_module.ExperimentTransform` to support experimental use case where the experiment wants to reset all Monty internals between episodes and epochs. I think a less intrusive approach would be for the experiment to reconstruct Monty between each episode and epoch instead.
```
nielsleadholm, line 82:
```suggestion
Additionally, we have a `tbp.monty.experiment.sensor_module.ExperimentTransform` to support the experimental use case where the experiment wants to reset all Monty internals between episodes and epochs. I think a less intrusive approach would be for the experiment to reconstruct Monty between each episode and epoch instead.
```

nielsleadholm (approved):
Just wanted to voice here on the RFC what I had said on Slack to your earlier posts about this: I really like this approach and support it.

I also like the idea of just rebuilding a Monty instance (with its long term memory) as an alternative to all the existing reset functionality we currently have sprinkled around the code.

PR #1007, opened 2026-06-01
refactor!: remove ambiguity from Resolution2D

jeremyshoemaker:
We noticed that the dimension order in the docstring for the `Resolution2D` tuple was `(width, height)` and the new MuJoCo code was following this convention, but NumPy and other parts of the code assume `(height, width)`. The tuple ordering came from the documentation for MuJoCo cameras, though they don't actually use that argument for anything useful.

This refactor changes Resolution2D into a TypedDict with keys `width` and `height` to remove any ambiguity when configuring and using them.

## Breaking Changes

The type of `Resolution2D` has change from `tuple[int, int]` to a `TypedDict` with `width` and `height` keys. This should only affect code and experiments using the MuJoCo simulator.


src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -192,11 +193,12 @@ def renderer_for_res(self, resolution: Resolution2D) -> Renderer:
         Returns:
             a renderer of the specified resolution
         """
-        if resolution not in self._renderers:
-            self._renderers[resolution] = Renderer(
-                width=resolution[0], height=resolution[1], model=self.model
+        resolution_key = (resolution["height"], resolution["width"])
```
tristanls-tbp, line 196: **question (non-blocking)**: Are you sure this is correct? It used to be (width, height). I'm making this non-blocking since a non-square use case might not have come up in the past, so this might not be an issue even if it technically is backward now.

tristanls-tbp (approved):
Thank you for the update.

**note**: It took me a little bit to get unconfused since in Habitat we have a different `SensorConfig` where resolution is called `Size`, and we also have "resolutions" in form of `tuple[int, int]` throughout the code.

I left one question, but I don't think it's an issue.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: That's a good point, and I don't think I have a test covering this case, so I'll double check it.

jeremyshoemaker: Actually, no it doesn't matter, because this is just the lookup key for the stored renderers, and as long as it's consistent, it will work fine. This is the only place we define **and** access this using a key, so it should stay consistent.

jeremyshoemaker: I think though that I do want to add a test to check that requesting the same resolution renderer results in getting the same renderer back.

jeremyshoemaker:
I refactored the test a little. I realized the one I was adding in NoopAgentTest didn't belong there, because it's testing the simulator's behaviour, so I removed it. The two tests I've added to MuJoCoSimulatorTest cover the case I was checking in the removed test, as well as the new cases I mentioned in my previous comment.

tests/unit/simulators/mujoco/mujoco_simulator_test.py
```diff
@@ -291,3 +294,45 @@ def actuate_look_up(*args, **kwargs):  # noqa: ARG001
...
+    )
+    def test_renderer_for_res_returns_different_renderers(
+        self,
+        width_1,
+        height_1,
+        width_2,
+        height_2,
+    ) -> None:
+        """When different resolutions are requested, renderers should be different."""
+        assume(width_1 != width_2 or height_1 != height_2)
```
tristanls-tbp, line 312: **issue (non-blocking)**: It is better to generate non-conflicting data than to reject data with `assume`, but I'm not gonna hold the PR on that.

tristanls-tbp (approved):
One non-blocking concern about the use of `assume`. But updated tests look good.

tests/unit/simulators/mujoco/mujoco_simulator_test.py
jeremyshoemaker: I considered that, but I think it would be more involved to generate non-conflicting data than to reject the few that conflict. Hypothesis tends to issue warnings when it's having to reject a large number of generated examples, and it doesn't here, so I felt it would be fine.

PR #1008, opened 2026-06-01
chore: remove is_orthogonal check

hlee9212:
As per [Shortcut story](https://app.shortcut.com/thousandbrainsproject/story/5241/optional-make-pc1-dir-and-pc2-dir-orthogonal) and discussed during Monday, June 1st TRM, we will remove these two lines until `principal_curvatures()` is updated.


vkakerbeck (approved):

src/tbp/monty/frameworks/utils/sensor_processing.py
```diff
@@ -155,9 +155,6 @@ def directional_curvature(
             movement_direction does not lie in the plane spanned by pc1_dir
             and pc2_dir.
     """
-    if not is_orthogonal(pc1_dir, pc2_dir):
```
nielsleadholm, line 158: The doc string above needs to be cleaned up "... ValueError: If pc1_dir and pc2_dir are not orthogonal,"

Also, I believe that this means you can remove the `is_orthogonal` function, which is no longer being used anywhere.

nielsleadholm (changes requested):

src/tbp/monty/frameworks/utils/sensor_processing.py
hlee9212: Fixed.

nielsleadholm: Is there a reason "ValueError... movement_direction does not lie in the plane spanned by pc1_dir and pc2_dir." was removed?

hlee9212: No, I was just tired and made a mistake. 😅

nielsleadholm (approved):

PR #1009, opened 2026-06-03
refactor!: learning modules explicitly accept percepts

tristanls-tbp:
While investigating connectivity between Sensor Modules and Learning Modules, it became apparent that the abstract interfaces as defined do not match what the code invokes.

Previously, the code types suggested that a single dictionary was being assembled as learning module input, however, it turns out that it is a list of Messages instead.

This pull request updates the abstract interfaces to match their function.

Additionally, the internals of `_step_learning_modules(...)` is updated to check `if sensory_inputs` instead of `if sensory_inputs is not None`, as internal methods now always return a list instead of list or None.

It is now also clear that `self.buffer.append_input_percepts(percepts)` call expects a single message but is given a list of them. This was already the case, but the corrected types make this error apparent. I decided not to go down the rabbit hole of fixing the buffer append logic at this time, as the only impact I'm aware of is visualization use cases, which work with the code as is and will eventually be replaced when we migrate to emit telemetry inline.

## Breaking Changes
- `LearningModule.matching_step` now requires `percepts: Sequence[Message]`
- `LearningModule.exploratory_step` now requires `percepts: Sequence[Message]`


tristanls-tbp:
@vkakerbeck, I requested your review as I did this investigation while considering the current state of how sensor module outputs are routed to learning modules. Learning modules already accept a list of messages, so if a motion sensor module sends a message that only contains a location, and a feature sensor module sends a message that only contains a feature, the connectivity routing should work as is. It might require smart handling inside the learning module, but I think the connectivity is already in place. The logic is in `_collect_inputs_to_lm` and `_combine_inputs` that the diff touched.

vkakerbeck (approved):
Nice, thanks for clarifying those interfaces. 

Regarding `append_input_percepts` I agree that this is a bit confusing, but I think we currently get the desired behavior. `self.input_percepts` is a list and if we call `append` on another list, like in the case where we get multiple percepts in one step, it gets appended as a sub-list. This way, we still have all the percepts of one step bundled together, and other functions like `get_previous_input_percepts` work correctly (as opposed to if we used `.extend`).

Regarding the connectivity and allowing movement and features to come from different SMs: I agree, from the SM and connectivity side, things should work already now. The places that would need to be changed are the CMP (doesn't allow a feature-only message) and the LM (how it processes them internally, particularly, what it does when receiving only one of the two). In the diagram I drew, I was thinking about how to best formalize that feature and movement data can come from different sources at different times, but that each LM should have both of those input channels. The CMP, as it is currently defined, doesn't capture this idea well. I am not sure about what the best solution would be. But you're probably right that we don't need to split up SMs into two different types or make any other changes to SMs or the Monty routing logic.

PR #1012, opened 2026-06-04
feat!: implement MuJoCo SurfaceAgent

jeremyshoemaker:
Implements a SurfaceAgent for the MuJoCo simulator.

## Breaking Changes
- The `pitch` method on `Embodiment` has changed meaning from pitching the sensor body to pitching the agent itself. A separate `pitch_sensor` method has been added to pitch the sensor body.


tests/unit/simulators/mujoco/agent_actions_test.py
jeremyshoemaker, line 1: The diff here is a little messy because I pulled out the Hypothesis strategies to a separate file when I was going to have separate test classes for both agents, but when I noticed how much duplicated test code there was going to be, I decided on this parameterized approach but didn't move those strategies back.

I think we should probably have a collection of test strategies in one place, and this is the first step in that direction, so I didn't undo that change.

src/tbp/monty/conf/experiment/base_config_10distinctobj_surf_agent_mujoco.yaml
```diff
@@ -0,0 +1,29 @@
...
+defaults:
+  - /monty: evidencegraph_exp1000_emin_t3_tot2500
+  - /monty/motor_system_config: surface_curvature_informed_goal1
+  - /monty/learning_module: evidence_1lm_nn5_dod0025
+  - /monty/sensor_module: camera_surf_delta
+  - /monty/connectivity: 1lm_1sm
+  - /environment: mujoco_ycb_surf_agent
+  - /env_interface: eval_distinctobj_predefined
+  - /env_interface/transform: missing_depthto3d_sensor2_semantic0_clip
+  - /logging: basic_warning_monty_runs
```
tristanls-tbp: **question**: Any specific reason for opting out of wandb here? (`base_config_10distinctobj_surf_agent` is `basic_warning_wandb_monty_runs`).

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -242,20 +245,28 @@ def yaw(self, delta_theta: float) -> None:
         new_rotation = rotation * delta_theta_rot
         self.rotation = new_rotation.as_quat()
 
-    def pitch(self, delta_phi: float, constraint: float) -> None:
+    def pitch(self, delta_phi: float) -> None:
+        """Pitch the embodiment by delta_phi degrees."""
+        delta_phi_rot = Rotation.from_euler("xyz", (delta_phi, 0, 0), degrees=True)
+        rotation = Rotation.from_quat(self.rotation)
+        new_rotation = rotation * delta_phi_rot
+        self.rotation = new_rotation.as_quat()
```
tristanls-tbp: **issue**: Note that this is an invalid assignment in the sense that `as_quat()` returns a numpy array, whereas `self.rotation` is a four-tuple matching `QuaternionWXYZ`.

**note**: Same issue above in `yaw(...)` implementation.

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -369,16 +382,157 @@ def actuate_set_sensor_rotation(self, action: SetSensorRotation) -> None:
...
 
     def actuate_move_forward(self, action: MoveForward) -> None:
-        self._embodiment.move_along_local_axis(distance=-action.distance, axis=Axis.Z)
+        """Moves the agent along its Z axis.
+
+        If a negative distance is supplied, the agent will instead move backwards
+        towards the positive-Z direction.
+        """
+        # The "forward" direction is in the negative-Z direction, so we need to
+        # negate the distance we've been given.
```
tristanls-tbp, line 391: **nitpick**: This feels like it should be in the docstring above instead.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -111,17 +132,14 @@ def __init__(
...
         self.spec = MjSpec()
-        self.model: MjModel = self.spec.compile()
+        self.model = self.spec.compile()
         self.data = MjData(self.model)
         self.data_path = Path(data_path) if data_path else None
         self._raise_actuate_missing = raise_actuate_missing
 
-        self._agent_partials = agents
-        self._agents: dict[AgentID, Agent] = {}
+        self._agent_partials = agents or []
```
tristanls-tbp: **issue (non-blocking)**: This bugs me because `agents` could be `[]` which would be `[] or []` and use the last `[]`, so probably what we mean here is `[] if agents is None else agents`, but they happen to have the same meaning. PEP8 does suggest doing `is` `is not` checks for `None` instead of falsy checks like this. I'm making this non-blocking because it's a style; it is not very clear what way is idiomatic, and our style guide is mum about this.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -95,6 +104,18 @@ class MuJoCoSimulator(SimulatedObjectEnvironment):
...
+    model: MjModel
+    data: MjData
+    data_path: Path | None
+
+    _raise_actuate_missing: bool
+    _agent_partials: Sequence[MuJoCoAgentFactory]
+    _agents: dict[AgentID, Agent]
+    _loaded_custom_types: set[str]
+    _object_count: int
+    _renderers: dict[tuple[int, int], Renderer]
```
tristanls-tbp, line 117: **question (non-blocking)**: I see a start of a convention. Any particular reason for this approach? Should we add it to our style guide?

tests/unit/simulators/mujoco/agent_actions_test.py
```diff
@@ -10,104 +10,158 @@
...
-    Needed because Hypothesis runs all the test cases for a @given test in one run
-    without calling TestCase.setUp and TestCase.tearDown in between.
+def orient_vertical_position(
+    down: float = 0.0, forward: float = 0.0, phi: float = 0.0
+) -> np.ndarray:
+    """Helper to calculate position after an OrientVertical action.
 
-    Yields:
-        the sim object, for use in the `with` statement.
+    Starts from the origin looking down the positive-Z axis.
```
tristanls-tbp: **question**: I'm trying to understand why this docstring states the start is looking down the positive-Z axis, while the `orient_horizontal_position` above states the start is looking down the negative-Z axis. Is this intended?

As far as I can tell, the implementations of OrientHorizontal and OrientVertical interpret forward in the same way, namely, the negative-Z axis. Both actuators do:
```python
self._embodiment.move_along_local_axis(-action.forward_distance, axis=Axis.Z)
```
for the forward distance action component.

tests/unit/simulators/mujoco/agent_actions_test.py
```diff
@@ -10,104 +10,158 @@
...
+@parameterized_class(
+    [{"agent_class": DistantAgent}, {"agent_class": SurfaceAgent}],
+    class_name_func=agent_test_name_func,
+)
+class ActionsTest(TestCase):
+    """This tests the actuate methods on the agents.
+
+    Rather than duplicate the test code for each agent, we run this suite over
+    the agent classes.
+    """
```
tristanls-tbp, line 165: **praise**: Very cool, today I learned. Also, `skip_on_missing_actuate` is extremely clever. I would hesitate a lot if we were that clever in code, but in tests... yay.

tests/unit/simulators/mujoco/agent_actions_test.py
```diff
@@ -293,3 +356,117 @@ def test_multiple_look_down_constrained(self, delta_phi1, delta_phi2, phi_limit)
 
             pitch, _, _ = sensor_rot.as_euler("xyz", degrees=True)
             assert abs(pitch) <= phi_limit + DEFAULT_TOLERANCE
+
+    @skip_on_missing_actuate()
+    @given(
+        delta_theta=st.floats(min_value=-180.0, max_value=180.0),
+        # 10 meters seems like a good upper limit given the size of our
+        # test objects.
```
tristanls-tbp, line 391: **lol**: ... we're only gonna need 10 computers 😬

tests/unit/simulators/mujoco/agent_actions_test.py
```diff
@@ -293,3 +356,117 @@ def test_multiple_look_down_constrained(self, delta_phi1, delta_phi2, phi_limit)
...
+        as we expect it to without over-complicating the test.
+        """
+        action = OrientHorizontal(
+            agent_id=TEST_AGENT_ID,
+            rotation_degrees=delta_theta,
+            forward_distance=forward_distance,
+            left_distance=left_distance,
+        )
+
+        with sim_resetter(self.sim):
```
tristanls-tbp: **issue**: Need `# type: ignore[attr-defined]` on the `self.sim` accesses here and elsewhere. Like you did with `cls.agent_class`. Or, maybe `ActionsTest` can have `sim` property?

tristanls-tbp (commented):
Submitting a review in progress. I'm almost through all of `agent_actions_test.py`, after that, only `strategies.py` remains.

Thank you for the additional docstrings throughout.

**question**: I see some strategy deadlines are set to `None`. I don't have MuJoCo set up on my machine... how much time do these tests add to the unit test runs? 

**todo**: This is a breaking PR. `pitch(...)` signature changed.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I don't remember writing it this way. 🤨 

I'm concerned this got changed with something auto-merged incorrectly. I've been trying out [Mergiraf](https://mergiraf.org/) for conflict resolution and it might have made a mistake. I'm pretty sure I wrote this as the if/else one-liner you suggested.

I'm going to change it back, and stop using that tool for now.

src/tbp/monty/simulators/mujoco/agents.py
jeremyshoemaker: I thought I adjusted all of these. I'll go back over them and make sure it's all correct.

src/tbp/monty/conf/experiment/base_config_10distinctobj_surf_agent_mujoco.yaml
jeremyshoemaker: No reason, I changed it while testing and forgot to change it back.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I've been seeing this in some of the PRs I've reviewed lately, and while at first I didn't like it, I think it's a better way to declare the object fields and their types, than doing it in the constructor. It might make sense to add it to the style guide.

tests/unit/simulators/mujoco/agent_actions_test.py
jeremyshoemaker: It's right on the edge of cleverness that I'm comfortable with. I actually thought you were going to push back on it marking tests to skip, and adding noise to the test output. The other alternative would be to mark the test as passed, but I think that could hide mistakes, and we _are_ in fact skipping the tests, so skipping makes sense.

tests/unit/simulators/mujoco/agent_actions_test.py
jeremyshoemaker: Famous last words, right?

src/tbp/monty/simulators/mujoco/agents.py
jeremyshoemaker: I'm actually a little bothered by how many of these require negating the value that comes from the action. The fact that we're negating it is an implementation detail, which is why I didn't add it to the docstring. If anything, I probably need to add similar comments below where we negate values in ways that seem arbitrary.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I think my main argument against it, is that at first glance it looked like the definition of class fields, rather than instance fields.

tests/unit/simulators/mujoco/agent_actions_test.py
jeremyshoemaker: I wrote it wrong. It's supposed to be negative not positive.

tests/unit/simulators/mujoco/agent_actions_test.py
jeremyshoemaker: `ActionsTest` does have a `sim` property, at the class level, defined in the `setUpClass` method. I'm surprised I'm not getting a warning about it not being defined.

jeremyshoemaker:
> **question**: I see some strategy deadlines are set to `None`. I don't have MuJoCo set up on my machine... how much time do these tests add to the unit test runs?

I ran into issues with Hypothesis failing on one or two of them and the suggestion was to increase the deadline, or set to `None`. I'm probably overdoing it by setting it on all of them, so I can go back and narrow it down to just the ones that need it.

I think it was actually in the GitHub action that it was hitting timeouts, since it was using the CPU for rendering.

> **todo**: This is a breaking PR. `pitch(...)` signature changed.

Fixed. I missed this because I was thinking of `Embodiment` as an internal detail, but it could easily be used by others in their own Agent implementations.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: To be clear, I don't think `agents or None` is good style. I find any use of `or` or `and` for that sort of thing to be kludgy.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: Apparently, [it can be both](https://mypy.readthedocs.io/en/stable/class_basics.html#:~:text=As%20in%20Python%20generally%2C%20a%20variable%20defined%20in%20the%20class%20body%20can%20be%20used%20as%20a%20class%20or%20an%20instance%20variable%2E). To be explicit about it being a class attribute instead of an instance attribute, you can use `ClassVar[T]`, see the updated `ActionsTest` class.

tests/unit/simulators/mujoco/agent_actions_test.py
jeremyshoemaker: I added type declarations to the class to specify that these are class attributes.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I seem to be inconsistent in my use of class attribute vs class field. The documentation says variable, but that seems too generic for these. Apologies for any confusion.

src/tbp/monty/simulators/mujoco/agents.py
tristanls-tbp: **issue (non-blocking)**: Invalid type assignment that's lingering in older code.

jeremyshoemaker:
> I ran into issues with Hypothesis failing on one or two of them and the suggestion was to increase the deadline, or set to `None`. I'm probably overdoing it by setting it on all of them, so I can go back and narrow it down to just the ones that need it.

For posterity, the specific error message I was getting was:

```
Unreliable test timings! On an initial run, this test took 703.80ms, which exceeded the deadline 
of 200.00ms, but on a subsequent run it took 1.87 ms, which did not. If you expect this sort of 
variability in your test timings, consider turning deadlines off for this test by setting 
deadline=None.
```

I think what's happening is that the first test run in each worker, Hypothesis is including the `setUpClass` call in the test run timings, and because there are multiple workers, it varies which one has the failure. If I set the number of workers to one, then the problem seems to go away.

It's inconsistent which ones fail, and how many, so we probably need to keep the `@settings(deadline=None)`. Setting it to None should be safe because these tests will finish, it's just their runtime that's inconsistent.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -95,6 +104,18 @@ class MuJoCoSimulator(SimulatedObjectEnvironment):
     recompile the model and data from whenever an object is added or removed.
     """
 
+    spec: MjSpec
+    model: MjModel
+    data: MjData
+    data_path: Path | None
```
tristanls-tbp: **issue**: I don't see `data_path` used outside of this class.

**suggestion**: Mark "private." (as in `_data_path`)

src/tbp/monty/simulators/mujoco/simulator.py
tristanls-tbp: Doing things this way seems clearer than the alternative. I'll try to update how I write my classes to this.

tests/unit/simulators/mujoco/agent_actions_test.py
```diff
@@ -10,104 +10,162 @@
...
+        pytest.skip(
+            reason=f"{e.agent_class.__name__} missing {e.method_name} method "
+            f"under test."
+        )
 
-@st.composite
-def unit_quaternion(draw) -> QuaternionWXYZ:
-    """Strategy to generate unit quaternions.
+
+def agent_test_name_func(
```
tristanls-tbp: **issue (non-blocking)**: Noisy `_func` suffix. Nothing is lost if this is called `agent_test_name`. While I understand this is provided in `class_name_func`, there's no collision with `class_name` that we're trying to avoid here.

tests/unit/simulators/mujoco/agent_actions_test.py
tristanls-tbp: That's a nice solution 👍. I like the instance and ClassVars declarations on the class.. seems clean.

tests/unit/simulators/mujoco/agent_actions_test.py
```diff
@@ -10,104 +10,162 @@
...
+from tests.unit.simulators.mujoco.strategies import (
+    constrained_angle,
+    position,
+    unit_quaternion,
+    x_rotation_quaterion,
+)
+
+
+@contextmanager
+def sim_resetter(sim: MuJoCoSimulator):
```
tristanls-tbp: **suggestion (non-blocking)**: You could also add a test method wrapper that does this automatically so that you don't have to do it in the test body:

```python
from typing_extensions import Concatenate, ParamSpec

P = ParamSpec("P")


def sim_reset(
    fn: Callable[Concatenate[ActionsTest, P], None],
) -> Callable[Concatenate[ActionsTest, P], None]:
    def wrapper(self: ActionsTest, *args: P.args, **kwargs: P.kwargs) -> None:
        sim = self.sim
        with sim_resetter(sim):
            return fn(self, *args, **kwargs)

    return wrapper
```

This way, test methods that want to use `sim_resetter` could be annotated with `@sim_reset` instead:
```python
    @skip_on_missing_actuate()
    @sim_reset
    @given(
        delta_theta=st.floats(min_value=-180.0, max_value=180.0),
        # 10 meters seems like a good upper limit given the size of our
        # test objects.
        left_distance=st.floats(min_value=0.0, max_value=10.0),
        forward_distance=st.floats(min_value=0.0, max_value=10.0),
    )
    @settings(deadline=None)
    def test_orient_horizontal(
        self, delta_theta: float, left_distance: float, forward_distance: float
    ) -> None:
```

tristanls-tbp: **suggestion**: Add return type `-> Generator[MuJoCoSimulator, None, None]`

tests/unit/simulators/mujoco/agent_actions_test.py
```diff
@@ -10,104 +10,162 @@
...
 
 
-@st.composite
-def position(draw) -> VectorXYZ:
-    x = draw(st.floats(min_value=-10.0, max_value=10.0))
-    y = draw(st.floats(min_value=-10.0, max_value=10.0))
-    z = draw(st.floats(min_value=-10.0, max_value=10.0))
-    return x, y, z
+@contextmanager
+def skip_on_missing_actuate():
```
tristanls-tbp: **issue**: Missing return type.

tests/unit/simulators/mujoco/agent_actions_test.py
```diff
@@ -293,3 +360,117 @@ def test_multiple_look_down_constrained(self, delta_phi1, delta_phi2, phi_limit)
...
+    @skip_on_missing_actuate()
+    @given(
+        delta_theta=st.floats(min_value=-180.0, max_value=180.0),
+        # 10 meters seems like a good upper limit given the size of our
+        # test objects.
+        left_distance=st.floats(min_value=0.0, max_value=10.0),
+        forward_distance=st.floats(min_value=0.0, max_value=10.0),
+    )
+    @settings(deadline=None)
+    def test_orient_horizontal(
```
tristanls-tbp, line 398: **suggestion**: Additional tests that increase confidence that everything is as it should be without duplicating code:
- zero `delta_theta`, `left_distance`, and `forward_distance` should be a no-op
- zero `left_distance` and `forward_distance` should be equivalent to turn right (or left, I don't remember)
- should never leave the xz-plane, no matter the input parameters, i.e. y position being zero is an invariant
- the distance traveled should be sqrt(forward^2, left^2)

tests/unit/simulators/mujoco/agent_actions_test.py
```diff
@@ -293,3 +360,117 @@ def test_multiple_look_down_constrained(self, delta_phi1, delta_phi2, phi_limit)
...
+            np.testing.assert_allclose(agent_pos, expected_pos)
+
+    @skip_on_missing_actuate()
+    @given(
+        delta_phi=st.floats(min_value=-180.0, max_value=180.0),
+        down_distance=st.floats(min_value=0.0, max_value=10.0),
+        forward_distance=st.floats(min_value=0.0, max_value=10.0),
+    )
+    @settings(deadline=None)
+    def test_orient_vertical(
```
tristanls-tbp, line 460: **suggestion**: Additional tests that increase confidence that everything is as it should be without duplicating code:
- zero `delta_theta`, `down_distance`, and `forward_distance` should be a no-op
- zero `down_distance` and `forward_distance` should be equivalent to look up (or down, I don't remember)
- should never leave the ??-plane, no matter the input parameters, one position dimension being zero is an invariant
- the distance traveled should be sqrt(forward^2, down^2)

tests/unit/simulators/mujoco/agent_actions_test.py
```diff
@@ -293,3 +360,117 @@ def test_multiple_look_down_constrained(self, delta_phi1, delta_phi2, phi_limit)
...
+            assert agent_rot.approx_equal(expected_rot)
+            np.testing.assert_allclose(agent_pos, expected_pos)
+
+    @skip_on_missing_actuate()
+    @given(
+        distance=st.floats(min_value=0.0, max_value=10.0),
+        theta=st.floats(min_value=-np.pi, max_value=np.pi),
+    )
+    @settings(deadline=None)
+    def test_move_tangentially(self, distance: float, theta: float) -> None:
```
tristanls-tbp: **suggestion**: Additional tests that increase confidence that everything is as it should be without duplicating code:
- zero `distance` should be a no-op
- orientation (rotation) should be unchanged
- inverse should undo back to start

tristanls-tbp:
Regarding `@settings(deadline=None)`, I'm not worried about the `None` part. What you said is reasonable. I just want to make sure that 16-second tests don't become 1-minute tests, regardless of any deadline settings.

tests/unit/simulators/mujoco/agent_actions_test.py
jeremyshoemaker: Those first two should be covered by Hypothesis. It picks zeros as one of the first examples it tries. I don't know how comparing the values we are given to the final state we calculate wouldn't catch a problem in that edge case. Having a separate test for either seems like overkill.

The other two would definitely be good additions to the asserts in this test.

jeremyshoemaker: The last one is not correct. The forward and left movement don't necessarily create a right triangle. I added a check for the Law of Cosine instead which is the more general equation.

tristanls-tbp (approved):
🙌

PR #1013, opened 2026-06-08
refactor: Explicitly set weights_only on torch.load calls

jeremyshoemaker:
In PyTorch 2.6, the default value for the `weights_only` argument in `torch.load` was changed to True.

This breaks our uses of it, as we are storing more than just weights and primitive Python objects. Originally I added the code block in `src/tbp/monty/__init__.py` to allow-list the modules we're specifically loading, but as I started work on the test migration for MuJoCo, it started wanting me to add internal modules from NumPy to the list.

Since this seems to be getting out of hand, I decided that changing the call sites was the simpler solution.

Technically, this introduces the possibility of RCE if we load untrusted data, but given that this isn't a live service, and we're generally loading data we've saved with `torch.save`, it should be safe.

This should have no effect on the Conda environment and the Habitat version of things, as we're just explicitly setting what was already the default.

### DIFF
--- a/src/tbp/monty/__init__.py
+++ b/src/tbp/monty/__init__.py
@@ -9,24 +9,3 @@
 # https://opensource.org/licenses/MIT.
 
 __version__ = "0.37.0"
-
-import sys
-
-# Things break if we try to import some of these classes in the Conda version
-if sys.version_info >= (3, 10):
-    import torch
-    from torch_geometric.data.data import DataEdgeAttr, DataTensorAttr
-    from torch_geometric.data.storage import GlobalStorage
-
-    from tbp.monty.frameworks.models.object_model import GraphObjectModel
-
-    # PyTorch 2.6 changes the default value of `weights_only` in `torch.load` from False
-    # to True, as a result it now gives an error when we try to load things without
-    # specifying `weights_only=False`.
-    #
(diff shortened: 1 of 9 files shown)


src/tbp/monty/__init__.py
```diff
@@ -9,24 +9,3 @@
...
-    # PyTorch 2.6 changes the default value of `weights_only` in `torch.load` from False
-    # to True, as a result it now gives an error when we try to load things without
-    # specifying `weights_only=False`.
-    #
-    # Rather than change all the call sites, the error message suggests an alternative
-    # of allow-listing the expected global classes we use.
-    # See https://docs.pytorch.org/docs/stable/notes/serialization.html#weights-only-allowlist
-    torch.serialization.add_safe_globals(
-        [GraphObjectModel, DataEdgeAttr, DataTensorAttr, GlobalStorage]
-    )
```
tristanls-tbp, line 32: **praise**: Love to see this disappear.

tristanls-tbp (approved):
Thanks 👍.

src/tbp/monty/__init__.py
jeremyshoemaker: Me too.

PR #1014, opened 2026-06-09
refactor: create MuJoCo version of run_parallel_test.py

jeremyshoemaker:

There are currently 23 test modules that are skipped in the `uv` environment because of the absence of Habitat. We need to migrate all of these to use MuJoCo instead.

As a first step, we're creating a parallel set of tests under `tests/mujoco` that mirror the original tests.

Once there are new versions for each of them, and we're ready to switch over to Habitat, we can move them into their original locations. Hopefully this will result in a useful diff of those tests, and preserve the original history.

> [!note]
> I'm marking this a refactor because I had to change non-test code.


tests/mujoco/integration/frameworks/run_parallel_test.py
```diff
@@ -0,0 +1,203 @@
...
+        shutil.rmtree(cls.output_dir)
+
+    @classmethod
+    def _load_hydra_configs(cls) -> None:
+        def hydra_config(
+            test_name: str, output_dir: Path, model_name_or_path: Path | None = None
+        ) -> DictConfig:
+            overrides = [
+                f"experiment=test/{test_name}",
+                "num_parallel=1",
```
jeremyshoemaker, line 67: This is from the original version of the test. If we're running with `num_parallel=1`, then how are we exercising the code we're testing here?

tests/mujoco/integration/__init__.py
jeremyshoemaker, line 1: It doesn't affect this PR, but I just noticed that this file doesn't match its counterpart in the main test tree. The original has a call to `setup_env`, the absence of which is breaking the `run_test.py` test I'm working on right now in a new branch.

jeremyshoemaker:
I'm not sure why the `test-uv-monty` suite is hanging. At first I was getting errors locally, until I realized when running the full suite I need to exclude the doctests part from trying to load the Habitat code, i.e. the `--ignore` arguments in the workflow definition.

With that sorted out, the tests run fine locally.

tests/mujoco/integration/__init__.py
jeremyshoemaker: I'm going to go ahead and update this now. It _might_ be causing the test issues I'm seeing in GitHub, and if not, we'll need it eventually anyway.

jeremyshoemaker:
It's still hanging, and it seems like it's hanging on the new test from this PR. I'm going to let it run and see if it ever finishes.

I'm starting to wonder if it's something to do with it using a software renderer in the GitHub Action. Locally, the MuJoCo version runs faster than the Habitat version, so I'd be surprised if it's that.

tests/mujoco/integration/frameworks/run_parallel_test.py
tristanls-tbp: The parallel path being tested here is the different execution scripts and the way data is loaded and saved in different directories. Actually, running in parallel is, I think, an incidental confound, so setting `num_parallel=1` was probably done to ensure the data replicated exactly. 

I think `num_parallel=1` might no longer be needed now that we have per-episode replication across serial and parallel runs.

But that's the context as far as I understand.

.github/workflows/monty.yml
```diff
@@ -540,6 +540,7 @@ jobs:
         continue-on-error: true
         run: |
           export MUJOCO_GL=egl
+          export EGL_PLATFORM=headless
```
jeremyshoemaker, line 543: I don't think this is a necessary change, given that MuJoCo simulated tests have been running fine so far. I just want to be explicit about it.

jeremyshoemaker:
Early working on this code, I was wondering _what_ exactly we are testing here, and whether it's a useful test.

I think the issues we uncovered with the MuJoCo simulator's GL context and its interaction with `multiprocessing` on Linux prove out that it's a useful test to have.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -275,6 +290,18 @@ def add_object(
 
         if name in PRIMITIVE_OBJECTS:
             self._add_primitive_object(obj_name, name, position, rotation, scale)
+        elif name in HABITAT_PRIMITIVE_OBJECTS:
+            # Habitat primitive objects are much smaller than the default MuJoCo ones
+            # so we need to adjust the default scale.
+            if scale == (1.0, 1.0, 1.0):
```
jeremyshoemaker: Looking at this again, I'm not sure why I special-cased this. We could always multiply the scale by the scaling factor. Maybe should change the constant name too.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -408,15 +435,31 @@ def _add_primitive_object(
...
+        rotation_quat = rotation_rot.as_quat()
+
         world_body: MjsBody = self.spec.worldbody
-        geom_type = PRIMITIVE_OBJECTS[object_type]
+
+        # TODO: remove try and except clauses when Habitat is removed
+        try:
+            geom_type = PRIMITIVE_OBJECTS[object_type]
+        except KeyError:
+            geom_type = HABITAT_PRIMITIVE_OBJECTS[object_type]
```
jeremyshoemaker, line 451: I went with this pattern because I was getting the KeyError initially, and I don't really want to change that behaviour, so catching the first one and letting the second one raise it if needed seemed like the more direct approach. I don't think that's actually possible though, because these two accesses cover all the possibilities.

Also, it's temporary code.

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
```diff
@@ -55,11 +55,22 @@ def __init__(self, config: DictConfig):
...
+                # MuJoCo agents are configured using a list of partially applied
+                # agent constructors.
+                # TODO: support more than one agent, this assumes there's only one.
+
+                # Introspect the first agent partial to determine what the
+                # original sensor positions were.
+                sensor_cfgs = agents_config[0].keywords["sensor_configs"]
+                self.sensor_pos = np.array(
+                    sensor["position"] for sensor in sensor_cfgs.values()
+                )
```
tristanls-tbp, line 73: **note 😱**: I thought... how hard could this be. So, I looked at what happens to this `self.sensor_pos`. It turns out, this value is **critical** in how **every learning module except the first one** updates its memory from the buffer after every episode. Essentially, `self.sensor_pos` experiment value here is used by every (except the first) learning module to learn **anything**.

Then again... this is the _Supervised_ experiment, so this all might just be the supervision aspect of it?

It's the unassuming lines 160 and 163 here https://github.com/thousandbrainsproject/tbp.monty/blob/8f062c4cd7d3cc87359bf9d70934591908c5d9c4/src/tbp/monty/frameworks/experiments/pretraining_experiments.py#L153-L165

How bad could it be.. those are just buffer _stats_, we probably just log them. Yes... except this one time:

https://github.com/thousandbrainsproject/tbp.monty/blob/8f062c4cd7d3cc87359bf9d70934591908c5d9c4/src/tbp/monty/frameworks/models/buffer.py#L425-L432

That `*_for_graph_update()` got me curious...

https://github.com/thousandbrainsproject/tbp.monty/blob/8f062c4cd7d3cc87359bf9d70934591908c5d9c4/src/tbp/monty/frameworks/models/graph_matching.py#L973-L982

Oh oh.. we are calling `self.graph_memory.update_memory()`. That is starting to sound serious. Let's see if the value we're tracking (`object_location_rel_body`) is used for anything serious in there...

https://github.com/thousandbrainsproject/tbp.monty/blob/8f062c4cd7d3cc87359bf9d70934591908c5d9c4/src/tbp/monty/frameworks/models/graph_matching.py#L1113-L1150

Not liking where this is going... `self._extend_graph()`

https://github.com/thousandbrainsproject/tbp.monty/blob/8f062c4cd7d3cc87359bf9d70934591908c5d9c4/src/tbp/monty/frameworks/models/graph_matching.py#L1344-L1373

🫣 ... `update_model()`:

https://github.com/thousandbrainsproject/tbp.monty/blob/8f062c4cd7d3cc87359bf9d70934591908c5d9c4/src/tbp/monty/frameworks/models/object_model.py#L73-L105

Ok.. phew.. maybe `apply_rf_transform_to_points` doesn't do anything with this...

https://github.com/thousandbrainsproject/tbp.monty/blob/8f062c4cd7d3cc87359bf9d70934591908c5d9c4/src/tbp/monty/frameworks/utils/spatial_arithmetics.py#L522-L570

Well, `ref_point2` pretty much creates `transformed_locations` which are the locations stored in the learning module memory... feel important.

Ok... let's pop back up to 

https://github.com/thousandbrainsproject/tbp.monty/blob/8f062c4cd7d3cc87359bf9d70934591908c5d9c4/src/tbp/monty/frameworks/models/graph_matching.py#L973-L982

Maybe `_update_memory()` sounds more imporant than it is.. oh no..

https://github.com/thousandbrainsproject/tbp.monty/blob/8f062c4cd7d3cc87359bf9d70934591908c5d9c4/src/tbp/monty/frameworks/models/graph_matching.py#L644-L648

... this is how a learning module learns 😟

https://github.com/thousandbrainsproject/tbp.monty/blob/8f062c4cd7d3cc87359bf9d70934591908c5d9c4/src/tbp/monty/frameworks/models/monty_base.py#L396-L401

tristanls-tbp (commented):
Submitting a review in progress, not done yet.. but... there's a story I wanted to share sooner rather than later...

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
```diff
@@ -55,11 +55,22 @@ def __init__(self, config: DictConfig):
...
         super().setup_experiment(config)
         if "agents" in config["environment"]["env_init_args"]:
-            self.sensor_pos = np.array(
-                config["environment"]["env_init_args"]["agents"]["agent_args"][
-                    "positions"
-                ]
-            )
+            agents_config = config["environment"]["env_init_args"]["agents"]
+            # Habitat agents are configured using `agent_args`.
+            # This code has always assumed there is only one.
```
tristanls-tbp: **suggestion (non-blocking)**
```suggestion
            # This code has always assumed there is only one agent.
```

src/tbp/monty/frameworks/run_parallel.py
```diff
@@ -627,7 +627,13 @@ def run_episodes_parallel(
         )
     print(f"Wandb setup took {time.time() - start_time} seconds")
     start_time = time.time()
-    with mp.Pool(num_parallel, maxtasksperchild=1) as p:
+
+    # Create a multiprocessing Context so that we can set the start method to "spawn".
+    # The default on MacOS is "spawn", but on Linux systems it uses "fork" by default.
+    # The "fork" method causes issues with the MuJoCo simulator's GL context, causing
+    # the system to hang when trying to render an image.
+    ctx = mp.get_context("spawn")
```
tristanls-tbp, line 635: **note**: Looks like this can also be done via a side-effect `mp.set_start_method("spawn")`.

src/tbp/monty/simulators/mujoco/simulator.py
tristanls-tbp: **issue**: I agree with you. No need for the special case here.

src/tbp/monty/simulators/mujoco/simulator.py
tristanls-tbp: **note**: Sounds like a reasonable approach to me. Thank you for explaining it.

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
jeremyshoemaker: I just noticed my comment here is inaccurate. It's not that this code _assumes_ there's only one agent. The way we configure Habitat only _allows_ for one agent.

tristanls-tbp (commented):
In progress comments.

src/tbp/monty/frameworks/run_parallel.py
jeremyshoemaker: That method doesn't work very well. It has to be the _very_ first thing that happens in the whole application otherwise you get a "context has already been set" error. There's a way to force it, but I think this approach should be the preferred method. Create a new context so it can be configured as needed and use it like you would the top level `multiprocessing` module.

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
jeremyshoemaker: The "agent" part was implied by the context of the previous line. That said, explicit is better. This is the version I'm planning to change it to.
```python
# The way we configure Habitat agents only allows for one agent.
```

tests/mujoco/integration/frameworks/run_parallel_test.py
```diff
@@ -0,0 +1,203 @@
...
+        cls._load_hydra_configs()
+
+        # Run the training experiments once for this whole test suite.
+        # serial run
+        exp = hydra.utils.instantiate(cls.supervised_pre_training_cfg.experiment)
+        with exp:
+            exp.run()
+
+        # parallel run
+        OmegaConf.clear_resolvers()  # main will re-register resolvers
```
tristanls-tbp: **issue**: I think this is no longer need sinice you updated `register_resolvers()` to skip already registered resolvers.

tests/mujoco/integration/frameworks/run_parallel_test.py
```diff
@@ -0,0 +1,203 @@
...
+        )
+
+    def test_eval_parallel_equals_serial_when_epochs_equals_rotations(self) -> None:
+        # serial run
+        exp = hydra.utils.instantiate(self.eval_cfg.experiment)
+        with exp:
+            exp.run()
+
+        # parallel run
+        OmegaConf.clear_resolvers()  # main will re-register resolvers
```
tristanls-tbp: **issue**: I think this is no longer need sinice you updated `register_resolvers()` to skip already registered resolvers.

tests/mujoco/integration/frameworks/run_parallel_test.py
```diff
@@ -0,0 +1,203 @@
...
+        self.assertTrue(pcsv.equals(scsv))
+
+    def test_eval_parallel_equals_serial_when_epochs_less_than_rotations(self) -> None:
+        # serial run
+        exp = hydra.utils.instantiate(self.eval_lt_cfg.experiment)
+        with exp:
+            exp.run()
+
+        # parallel run
+        OmegaConf.clear_resolvers()  # main will re-register resolvers
```
tristanls-tbp: **issue**: I think this is no longer need sinice you updated `register_resolvers()` to skip already registered resolvers.

tests/mujoco/integration/frameworks/run_parallel_test.py
```diff
@@ -0,0 +1,203 @@
...
+    def test_eval_parallel_equals_serial_when_epochs_greater_than_rotations(
+        self,
+    ) -> None:
+        # serial run
+        exp = hydra.utils.instantiate(self.eval_gt_cfg.experiment)
+        with exp:
+            exp.run()
+
+        # parallel run
+        OmegaConf.clear_resolvers()  # main will re-register resolvers
```
tristanls-tbp: **issue**: I think this is no longer need sinice you updated `register_resolvers()` to skip already registered resolvers.

tests/mujoco/integration/frameworks/run_parallel_test.py
```diff
@@ -0,0 +1,203 @@
...
+        main(self.eval_gt_cfg)
+
+        eval_dir_gt = self.output_dir / "gt"
+        parallel_eval_dir_gt = eval_dir_gt / "test_eval_gt_mujoco"
+
+        scsv_gt = pd.read_csv(eval_dir_gt / "eval_stats.csv")
+        pcsv_gt = pd.read_csv(parallel_eval_dir_gt / "eval_stats.csv")
+
+        drop_cols = ["time", "stepwise_performance", "stepwise_target_object"]
```
tristanls-tbp, line 198: **lol**: I can see the comment for `drop_cols` diminishing as we go down the file 😆

tests/mujoco/integration/frameworks/run_parallel_test.py
tristanls-tbp, line 1: **praise**: Really nice reorganization. Thank you for tidying.

src/tbp/monty/frameworks/experiments/pretraining_experiments.py
tristanls-tbp: **note**: After talking to the team, this is part of supervision. An improvement we could make here is to avoid using `buffer.stats` on the learning module's learning critical path and to ensure this data is passed more directly, not through the logging telemetry code path.

tristanls-tbp (commented):
Thank you. Looks good, and the test updates improve it.

I think `OmegaConf.clear_resolvers()` can be removed thanks to your previous efforts.

jeremyshoemaker:
> I think `OmegaConf.clear_resolvers()` can be removed thanks to your previous efforts.

I was wondering about that too, but hadn't looked into it. I'll give it a try and see if anything breaks.

tristanls-tbp:
**note**: I misread the blame... as you pointed out, Hojae updated the resolver code.

tests/mujoco/integration/frameworks/run_parallel_test.py
jeremyshoemaker: I cleared it up.

I thought about making this a constant on the class and then it was suggested to make it immutable, so I tried a tuple, but then `drop` failed, despite saying that `columns` can be a `Sequence[Hashable]` which a tuple of strings certainly is.

I went back to the original version.

tristanls-tbp (approved):
🙌

PR #1017, opened 2026-06-09
chore: ignore ruff checks RET501 and PLR1711

dschumacher-tbp:
Add the following `ruff` checks to the Monty-specific ignore section of `pyproject.toml`:
- **RET501**: Do not explicitly `return None` in function if it is the only possible return value
- **PLR1711**: Useless `return` statement at end of function

Also, remove the previous `# noqa:` directive for these checks.

More details about these checks can be found at:
- https://docs.astral.sh/ruff/rules/unnecessary-return-none/
- https://docs.astral.sh/ruff/rules/useless-return/


tristanls-tbp (approved):
Thanks for the update.

PR #1025, opened 2026-06-11
test: create MuJoCo version of run_test.py

jeremyshoemaker:

The Habitat version of this test mocks out the internals of Habitat, which is easier to do because it's self-contained in their internal simulator object.

MuJoCo doesn't have an equivalent internal simulator object, but instead we interact with it via several APIs. This makes mocking it all more complicated. Mocking our `MuJoCoSimulator` itself runs into issues because we're pickling the experiment configuration and `Mock`s aren't picklable.

In local testing, it seems like actually running the MuJoCo simulator instead of mocking it is faster than the mocked Habitat version. Roughly 3.5 seconds on average vs 5.5 seconds on average. The difference could be attributed to Rosetta translation of the x86 Conda environment on MacOS, vs the native aarch64 version that MuJoCo uses.

Using the simulator directly keeps the test setup code simpler and easier to follow than mocking it all out would. Also, this is an integration test, so running the full simulator is probably a better test, despite this test only tracking whether certain methods were called in a certain order.

Mocking MontyExperiment would be another option, but that's the code under test, so it doesn't seem appropriate to take that approach.

### DIFF
--- a/tests/mujoco/integration/frameworks/run_test.py
+++ b/tests/mujoco/integration/frameworks/run_test.py
@@ -1,0 +1,83 @@
+# Copyright 2025-2026 Thousand Brains Project
+# Copyright 2022-2024 Numenta Inc.
+#
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+
+import json
+import shutil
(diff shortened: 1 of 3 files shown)


tests/mujoco/integration/frameworks/run_test.py
tristanls-tbp, line 1: **note**: so small..

tristanls-tbp (approved):
Nice, thank you.

PR #1026, opened 2026-06-12
refactor!: split Runtime and Experiment aspects of MotorSystem, et. al.

dschumacher-tbp:
The methods of `MotorSystem`, `MotorPolicySelector`, and `MotorPolicy` are segregated into "Methods that define the algorithm" and "Methods that interact with the experiment". This PR extracts `RuntimeMotorSystem`, `RuntimeMotorPolicySelector`, and `RuntimeMotorPolicy` Protocols for the former, and `ExperimentMotorSystem`, `ExperimentMotorPolicySelector`, and `ExperimentMotorPolicy` Protocols for the latter.

This segregation is in preparation for moving Experiment-related logic from the Motor System into the Experiment.

## Breaking Changes

- `MotorSystem.pre_episode()` is renamed to `ExperimentMotorSystem.reset()`
- `MotorPolicySelector.pre_episode()` is renamed to `ExperimentMotorPolicySelector.reset()`
- `MotorPolicy.pre_episode()` is renamed to `ExperimentMotorPolicy.reset()`
- `MotorSystem.motor_only_step` is moved to `ExperimentMotorSystem.motor_only_step`

## Test Changes

Some tests were constructing a `MotorPolicy` and passing it directly to the `MotorSystem` constructor, which expects a `MotorPolicySelector`. They now construct a `SinglePolicySelector` wrapping the `MotorPolicy`.


src/tbp/monty/frameworks/models/motor_policies.py
```diff
@@ -559,7 +572,7 @@ def _derive_set_agent_pose_from_goal(self, goal: Goal) -> SetAgentPose:
             rotation_quat=target_quat,
         )
 
-    def _reset(self) -> None:
```
tristanls-tbp, line 562: **question**: Why the rename? The policy name is `JumpToGoal`. What state is there other than the jump state?

src/tbp/monty/frameworks/models/motor_system.py
```diff
@@ -48,11 +79,11 @@ def __init__(self, policy_selector: MotorPolicySelector) -> None:
         self._policy_selector = policy_selector
         # For each step, we store the actions produced by the policy and the current
         # motor system state as a (actions, state) tuple.
-        self._action_sequence: list[tuple[list[Action], dict[AgentID, Any] | None]] = []
+        self._action_sequence: list[tuple[Sequence[Action], MotorSystemState]] = []
```
tristanls-tbp: **issue**: Search for the string `action_sequence` in the codebase, and you'll see that there are places that expect the old type signature. For example: `plot_utils_dev.get_action_name()`. So, passing a `Sequence` where a `list` is expected won't do.

**suggestion**: Update the other places if you'd like to keep this change.

src/tbp/monty/frameworks/models/motor_system.py
```diff
@@ -62,18 +93,26 @@ def __init__(self, policy_selector: MotorPolicySelector) -> None:
...
+        self._motor_only_step = value
+
+    @property
+    def action_sequence(self) -> Sequence[tuple[Sequence[Action], MotorSystemState]]:
         return self._action_sequence
 
-    def pre_episode(self) -> None:
-        """Pre episode hook."""
+    def reset(self) -> None:
         # TODO: Passing self to policy pre_episode is a hack. What we should be
```
tristanls-tbp: **issue**: Please update the comment to reflect the code change; You removed `pre_episode`.

src/tbp/monty/frameworks/models/motor_system.py
```diff
@@ -36,7 +39,35 @@ class SurfacePolicyActionDetailsTelemetry:
...
+    """Monty runtime interface to a Motor System."""
+
+    def __call__(
+        self,
+        ctx: RuntimeContext,
+        observations: Observations,
+        proprioceptive_state: ProprioceptiveState,
+        percept: Message,
+        goals: Sequence[Goal],
+    ) -> Sequence[Action]:
```
tristanls-tbp: **issue**: You changed the return value to be `Sequence[Action]`. `MontyBase._actions` expects a `list[Action]`, so giving it a `Sequence` won't do.

tests/integration/frameworks/models/policy_test.py
```diff
@@ -637,9 +638,10 @@ def test_core_following_principal_curvature(self) -> None:
         policy: SurfacePolicyCurvatureInformed = hydra.utils.instantiate(
             self.policy_cfg_fragment
         )
-        motor_system = MotorSystem(policy)
+        policy_selector = SinglePolicySelector(policy)
```
tristanls-tbp, line 641: **praise**: Good catch, thanks for fixing this.

tristanls-tbp (commented):
Thank you. Overall looks good. However, there's one issue with changing the result types.

**issue**: If a method return type changes, then every caller of that method needs to be able to handle the new returned type. I see multiple instances of callers expecting a `list` being handed a `Sequence`.

src/tbp/monty/frameworks/models/motor_policies.py
dschumacher-tbp: answer: to avoid confusion with the new `self.reset()` and also to match
https://github.com/dschumacher-tbp/tbp.monty/blob/58ed9bf3a5857abd766c902b9019b85ecf4a80b6/src/tbp/monty/frameworks/models/motor_policies.py#L838 in `InformedPolicy`.

tristanls-tbp: **note (for context, not asking for a change)**: `InformedPolicy` contains three policies. That's why it has a reset specifically for resetting the jump state, to distinguish it from the other resets in `pre_episode()`. `InformedPolicy` will be removed at some point, so it's not an example to emulate. `JumpToGoal` is the extracted part that only does the jump, so we didn't think it needed a further qualifier.

src/tbp/monty/frameworks/models/salience/motor_policy.py
```diff
@@ -168,7 +168,7 @@ def __call__(
         sensor_pitch = target_pitch_rel_agent - sensor_pitch_rel_agent
 
         # Create actions to return to the the motor system.
-        actions = [
+        actions: list[Action] = [
```
tristanls-tbp, line 171: **issue**: Extraneous type annotation.

tristanls-tbp (commented):
**issue**: I still think you need to update `get_action_name` to accept `MotorSystemState` instead of `dict[str, Any] | None`.

src/tbp/monty/frameworks/models/salience/motor_policy.py
dschumacher-tbp: Type inference doesn't produce the fully-qualified type, so I've added it explicitly, as per:
image

tristanls-tbp: OK. It looks like a difference in our tooling, and it is not extraneous in your environment.

tristanls-tbp (approved):
Looks good.

Thank you for removing some opportunistic changes that would require not insignificant, but seemingly unrelated, refactoring. This helps keep this PR focused.

PR #1027, opened 2026-06-12
refactor!: change Resolution2D and SensorConfig to dataclasses

jeremyshoemaker:
While migrating tests for MuJoCo, I've run into a few instances where the Habitat version of a config relies on default values for some of the sensor configuration.

Rather than make it explicit every time, I'm changing SensorConfig and Resolution2D into dataclasses. This allows us to set defaults, and also improves the way we interact with these objects in code.

I think I was under the impression that Hydra would need extra details to use dataclasses, like specifying a `_target_` but that's not the case.

> [!warning] 
> **Breaking Changes**
> - The type of `SensorConfig` changed from a TypedDict to a dataclass.
> - The type of `Resolution2D` changed from a TypedDict to a dataclass.

### DIFF
--- a/src/tbp/monty/simulators/mujoco/simulator.py
+++ b/src/tbp/monty/simulators/mujoco/simulator.py
@@ -36,7 +36,7 @@ from tbp.monty.frameworks.environments.environment import (
 )
 from tbp.monty.frameworks.models.abstract_monty_classes import Observations
 from tbp.monty.frameworks.models.motor_system_state import ProprioceptiveState
-from tbp.monty.frameworks.sensors import Resolution2D
+from tbp.monty.frameworks.sensors import Resolution2D, SensorConfig, SensorID
 from tbp.monty.geometry import Rotation
 from tbp.monty.math import IDENTITY_QUATERNION, ZERO_VECTOR, QuaternionWXYZ, VectorXYZ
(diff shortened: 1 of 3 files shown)


src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -254,14 +256,14 @@ def _max_sensor_resolution(self) -> Resolution2D:
...
             for p in cast("list[partial[Agent]]", self._agent_partials)
         ]
-        for sensor_cfg in sensor_configs:
-            for sensor in sensor_cfg.values():
-                max_height = max(max_height, sensor["resolution"]["height"])
-                max_width = max(max_width, sensor["resolution"]["width"])
+        for sensor_cfgs in agent_sensor_cfgs:
+            for sensor_cfg in sensor_cfgs.values():
+                max_height = max(max_height, sensor_cfg.resolution.height)
+                max_width = max(max_width, sensor_cfg.resolution.width)
```
jeremyshoemaker, line 266: A lot of this is me clarifying this code a bit. It was confusing before, and when I was writing out the type on line 259, it became clear what the problem was.

The actually necessary changes are on lines 265 and 266.

tristanls-tbp (approved):
Sounds good 👍

PR #1028, opened 2026-06-12
test: create MuJoCo version of base_config_test.py

jeremyshoemaker:

### DIFF
--- a/tests/mujoco/integration/frameworks/experiments/base_config_test.py
+++ b/tests/mujoco/integration/frameworks/experiments/base_config_test.py
@@ -23,8 +23,7 @@ from tests import HYDRA_ROOT
 
 
 class BaseConfigTest(unittest.TestCase):
-    def setUp(self):
-        """Code that gets executed before every test."""
+    def setUp(self) -> None:
         self.output_dir = tempfile.mkdtemp()
 
         with hydra.initialize_config_dir(version_base=None, config_dir=str(HYDRA_ROOT)):
@@ -36,21 +35,15 @@ class BaseConfigTest(unittest.TestCase):
                 ],
             )
 
-    def tearDown(self):
-        """Code that gets executed after every test."""
+    def tearDown(self) -> None:
         shutil.rmtree(self.output_dir)
 
-    def test_can_set_up(self):
-        """Canary for setup_experiment.
-
-        This could be part of the setUp method, but it's easier to debug if
-        something breaks the setup_experiment method if there's a separate test for it.
-        """
+    def test_can_set_up(self) -> None:
         exp = hydra.utils.instantiate(self.base_cfg.experiment)
         with exp:
             pass
 
-    def test_can_run_episode(self):
+    def test_can_run_episode(self) -> None:
         exp = hydra.utils.instantiate(self.base_cfg.experiment)
         with exp:
             exp.experiment_mode = ExperimentMode.TRAIN
@@ -58,22 +51,21 @@ class BaseConfigTest(unittest.TestCase):
             exp.env_interface = exp.train_env_interface
             exp.run_episode()
 
-    def test_can_run_train_epoch(self):
+    def test_can_run_train_epoch(self) -> None:
         exp = hydra.utils.instantiate(self.base_cfg.experiment)
         with exp:
             exp.experiment_mode = ExperimentMode.TRAIN
             exp.model.set_experiment_mode(exp.experiment_mode)
             exp.run_epoch()
 
-    def test_can_run_eval_epoch(self):
+    def test_can_run_eval_epoch(self) -> None:
         exp = hydra.utils.instantiate(self.base_cfg.experiment)
         with exp:
             exp.experiment_mode = ExperimentMode.EVAL
             exp.model.set_experiment_mode(exp.experiment_mode)
             exp.run_epoch()
 
-    def test_observation_unpacking(self):
-        """Make sure this test uses very small n_actions_per_epoch for speed."""
+    def test_observation_unpacking(self) -> None:
         exp = hydra.utils.instantiate(self.base_cfg.experiment)
         with exp:
             monty_module_sids = {s.sensor_module_id for s in exp.model.sensor_modules}
@@ -114,20 +106,19 @@ class BaseConfigTest(unittest.TestCase):
                     "sensor_module must receive depth",
                 )
 
-    def test_can_save_and_load(self):
-        # Make sure deepcopy works (config is serializable)
-        config_1 = OmegaConf.to_object(self.base_cfg)
+    def test_can_save_and_load(self) -> None:
+        config_1: Mapping = OmegaConf.to_object(self.base_cfg)  # type: ignore[assignment,type-arg]
 
         exp = hydra.utils.instantiate(config_1["experiment"])
         with exp:
-            # change something about exp.state that will be saved via save_state_dir
+            # Change something about exp.state that will be saved via save_state_dir.
             new_attr = False
             exp.model.learning_modules[0].test_attr_2 = new_attr
 
             exp.save_state_dir()
             prev_model = exp.model
 
-        config_2: Mapping = OmegaConf.to_object(  # ignore: type[assignment]
+        config_2: Mapping = OmegaConf.to_object(  # type: ignore[assignment,type-arg]
             self.base_cfg
         )
         config_2["experiment"]["config"]["model_name_or_path"] = exp.output_dir
@@ -157,6 +148,9 @@ class BaseConfigTest(unittest.TestCase):
 
     def test_logging_debug_level(self) -> None:
         """Check that logs go to a file, we can load them, and they have basic info."""
+        # TODO: This seems to test the behaviour of the Python `logging` library, which
+        #   we should just assume works as advertised. Change this test to introspect
+        #   the experiment to see if the loggers got configured correctly instead.
         exp = hydra.utils.instantiate(self.base_cfg.experiment)
         with exp:
             # Add some stuff to the logs, verify it shows up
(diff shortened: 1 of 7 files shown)


jeremyshoemaker:
I'm opening this as-is, but I have some thoughts about changing the tests themselves, that I'd like to discuss.

tests/mujoco/integration/frameworks/experiments/base_config_test.py
```diff
@@ -0,0 +1,196 @@
...
+from omegaconf import OmegaConf
+
+from tbp.monty.context import RuntimeContext
+from tbp.monty.frameworks.experiments.mode import ExperimentMode
+from tests import HYDRA_ROOT
+
+
+class BaseConfigTest(unittest.TestCase):
+    def setUp(self):
+        """Code that gets executed before every test."""
```
tristanls-tbp: **issue**: I know this docstring is copy/pasted, but please remove.

tests/mujoco/integration/frameworks/experiments/base_config_test.py
```diff
@@ -0,0 +1,196 @@
...
+
+            logger = logging.getLogger("tbp.monty")
+            logger.debug(debug_message)
+            logger.warning(warning_message)
+
+            with (exp.output_dir / "log.txt").open() as f:
+                log = f.read()
+
+            self.assertTrue(debug_message not in log)
+            self.assertTrue(warning_message in log)
```
jeremyshoemaker, line 193: These two tests seem to me like they're testing Python logging behaviour, which isn't something we should be testing ourselves. 

We _could_ argue that we're testing that our logging configuration code is being tested, but if that's what we want to test I think we should make that more direct. Introspect the configuration after everything is loaded up. We don't need to actually use the Python logger to do that.

tests/mujoco/integration/frameworks/experiments/base_config_test.py
```diff
@@ -0,0 +1,196 @@
...
+            self.base_cfg = hydra.compose(
+                config_name="experiment",
+                overrides=[
+                    "experiment=test/base_config/base_mujoco",
+                    f"experiment.config.logging.output_dir={self.output_dir}",
+                ],
+            )
+
+    def tearDown(self):
+        """Code that gets executed after every test."""
```
tristanls-tbp: **issue**: I know this docstring is copy/pasted, but please remove.

tests/mujoco/integration/frameworks/experiments/detailed_logging_config_test.py
```diff
@@ -0,0 +1,63 @@
...
+        exp = hydra.utils.instantiate(self.cfg.experiment)
+        with exp:
+            exp.model.set_experiment_mode(ExperimentMode.EVAL)
+            exp.run_epoch()
+
+        # Detailed logging handler should create a detailed_run_stats.json file
+        self.assertTrue(
+            (exp.output_dir / "detailed_run_stats.json").exists(),
+            "Expected detailed_run_stats.json file to be created",
+        )
```
jeremyshoemaker, line 61: This also seems like it's testing Python logging behaviour, confirming that it saves files when we configure it to do so.

Looking at the docstring, it sounds like it's trying to test three distinct things, which indicates to me that they should be separate tests.

tests/mujoco/integration/frameworks/experiments/base_config_test.py
```diff
@@ -0,0 +1,196 @@
...
+
+            with (exp.output_dir / "log.txt").open() as f:
+                log = f.read()
+
+            self.assertTrue(info_message in log)
+            self.assertTrue(warning_message in log)
+
+    def test_logging_info_level(self) -> None:
+        """Check that if we set logging level to info, debug logs do not show up."""
+        base_config: Mapping = OmegaConf.to_object(self.base_cfg)
```
tristanls-tbp: **issue**: I'm pretty sure this needs a `# type: ignore[assignment]`

tests/mujoco/integration/frameworks/experiments/base_config_test.py
```diff
@@ -0,0 +1,196 @@
...
+        exp = hydra.utils.instantiate(config_1["experiment"])
+        with exp:
+            # change something about exp.state that will be saved via save_state_dir
+            new_attr = False
+            exp.model.learning_modules[0].test_attr_2 = new_attr
+
+            exp.save_state_dir()
+            prev_model = exp.model
+
+        config_2: Mapping = OmegaConf.to_object(  # ignore: type[assignment]
```
tristanls-tbp: **issue**: This should be `# type: ignore[assignment]`

tests/mujoco/integration/frameworks/experiments/base_config_test.py
```diff
@@ -0,0 +1,196 @@
...
+                )
+                self.assertIn(
+                    "depth",
+                    feature.non_morphological_features,
+                    "sensor_module must receive depth",
+                )
+
+    def test_can_save_and_load(self):
+        # Make sure deepcopy works (config is serializable)
+        config_1 = OmegaConf.to_object(self.base_cfg)
```
tristanls-tbp: **issue**: Type complaints.

```suggestion
        config_1: Mapping = OmegaConf.to_object(self.base_cfg)  # type: ignore[assignment]
```

tests/mujoco/integration/frameworks/experiments/base_config_test.py
tristanls-tbp, line 1: **issue**: All test methods should return `-> None`. I know the original is missing some of these.

tristanls-tbp (commented):
My comments are mostly about types and docstrings.

jeremyshoemaker:
I meant to go through and cleanup the comments and docstrings, and it looks like I forgot to. Probably all the context switching today.

tests/mujoco/integration/frameworks/experiments/base_config_test.py
jeremyshoemaker: Yep, looks like someone messed that up. It's in the original.

tristanls-tbp: It was a team effort 😉 

Screenshot 2026-06-12 at 18 34 53

jeremyshoemaker: I always assume that if I find a mistake, there's good odds that I did it myself months previously. So I definitely included myself in "someone".

tests/mujoco/integration/frameworks/experiments/detailed_logging_config_test.py
jeremyshoemaker: After our discussion on Friday, I updated these tests with TODOs to point out the problems and suggest a solution in each case. Rather than get lost trying to refactor them now, it's something we can address at a later date.

tests/mujoco/integration/frameworks/experiments/base_config_test.py
tristanls-tbp: **note**: I think this is testing our logging configuration code, as you pointed out. Your TODO looks appropriate 👍.

jeremyshoemaker:
Apparently we were wrong about Hydra working with dataclasses, and in hindsight that probably should have been obvious, as how would it know there's a dataclass involved if we don't tell it. 

The reason we didn't catch it is that `DictConfig` lets you access the keys as attributes on the class, which makes it appear as though it's a dataclass when it's not. This falls apart when trying to rely on a dataclass' default values, as I was doing here.

I'm going with the fix of setting the `_target_` for the sensor config correctly so it picks up the defaults. It's less than ideal.

tests/mujoco/integration/frameworks/experiments/detailed_logging_config_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+from tbp.monty.frameworks.experiments.mode import ExperimentMode
+from tests import HYDRA_ROOT
+
+
+class DetailedEvidenceLmLoggingConfigTest(unittest.TestCase):
+    # TODO: This test seems to primarily test our DetailedJSONHandler, and we should
+    #   do that directly instead of testing to see if the Python `logging` library
+    #   did what it is supposed to do.
+    def setUp(self):
```
tristanls-tbp: **issue (type)**: `-> None` missing.

tests/mujoco/integration/frameworks/experiments/detailed_logging_config_test.py
```diff
@@ -0,0 +1,61 @@
...
+            self.cfg = hydra.compose(
+                config_name="experiment",
+                overrides=[
+                    "experiment=test/evidence_lm/base_mujoco",
+                    "logging=detailed_info_evidence_eval_runs",
+                    f"++experiment.config.logging.output_dir={self.output_dir}",
+                ],
+            )
+
+    def tearDown(self):
```
tristanls-tbp: **issue (type)**: `-> None` missing.

tests/mujoco/integration/frameworks/experiments/detailed_logging_config_test.py
```diff
@@ -0,0 +1,61 @@
...
+        shutil.rmtree(self.output_dir)
+
+    def test_can_set_up(self):
```
tristanls-tbp: **issue (type)**: `-> None` missing.

tristanls-tbp (commented):
Looks good, missed a few return types.

tristanls-tbp:
Good finding on dataclasses, thank you for pointing that out and the fix. Maybe Hydra was magic, and it saw what type was expected and instantiated it... was my hope.

jeremyshoemaker:
> Maybe Hydra was magic, and it saw what type was expected and instantiated it... was my hope.

Mine too.

tristanls-tbp (approved):
Thanks!

PR #1030, opened 2026-06-15
refactor!: create MuJoCo version of graph_building_test.py

jeremyshoemaker:

In the new configuration files, the only field that has changed is `/environment` in the defaults list entries. There was some additional type-hinting added to functions called directly from the new test module, due to MyPy's `no-untyped-call` rule. MyPy still returns errors for the test module because we're accessing methods on subclasses of `LearningModule` that don't exist on `LearningModule` itself. Same goes for `SensorModule`.

> [!warning]
> **Breaking Changes**
> - `MuJoCoSimulator` incorrectly returned RGB data for the "rgba" modality. It now returns data with an alpha component set to 255 (full opaque) since that's the only alpha value a camera can capture. The shape of the data has changed from `(height, width, 3)` to `(height, width, 4)`.

### DIFF
--- a/src/tbp/monty/simulators/mujoco/agents.py
+++ b/src/tbp/monty/simulators/mujoco/agents.py
@@ -179,7 +179,14 @@ class Embodiment(Agent):
         for sensor_id, sensor_cfg in self._sensor_configs.items():
             renderer = self.sim.renderer_for_res(sensor_cfg.resolution)
             renderer.update_scene(self.sim.data, camera=f"{self.id}.{sensor_id}")
-            rgba_data = renderer.render()
+            # MuJoCo only renders RGB image data. Monty currently expects RGBA, so
+            # we need to add an alpha channel. We're setting it to a fully opaque
+            # value, since a camera should always return opaque pixel values for
+            # every location.
+            rgb_data = renderer.render()
+            rgba_data = np.dstack(
+                [rgb_data, np.full(rgb_data.shape[:2], 255, dtype=rgb_data.dtype)]
+            )
 
             renderer.enable_depth_rendering()
             depth_data = renderer.render()
--- a/tests/mujoco/integration/frameworks/experiments/graph_building_test.py
+++ b/tests/mujoco/integration/frameworks/experiments/graph_building_test.py
@@ -1,0 +1,257 @@
+# Copyright 2025-2026 Thousand Brains Project
+# Copyright 2022-2024 Numenta Inc.
+#
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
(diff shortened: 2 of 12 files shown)


src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -179,7 +179,14 @@ def observations(self) -> AgentObservations:
...
             renderer.update_scene(self.sim.data, camera=f"{self.id}.{sensor_id}")
-            rgba_data = renderer.render()
+            # MuJoCo only renders RGB image data. Monty currently expects RGBA, so
+            # we need to add an alpha channel. We're setting it to a fully opaque
+            # value, since a camera should always return opaque pixel values for
+            # every location.
+            rgb_data = renderer.render()
+            rgba_data = np.dstack(
+                [rgb_data, np.full(rgb_data.shape[:2], 255, dtype=rgb_data.dtype)]
+            )
```
jeremyshoemaker, line 189: I must admit I had some help with this one. This seems the correct way to do this, and the values returned are what we expect. I confirmed by running one of the experiments I've added that support MuJoCo, and the images render correctly.

tests/mujoco/integration/frameworks/experiments/graph_building_test.py
```diff
@@ -0,0 +1,257 @@
...
+
+        Returns:
+            The experiment.
+        """
+        exp: MontySupervisedObjectPretrainingExperiment = hydra.utils.instantiate(
+            self.sptm_feat_cfg.experiment
+        )
+        with exp:
+            exp.run()
+        return exp
```
jeremyshoemaker, line 140: I'm debating whether these two functions should just return `MontyExperiment`, mostly because the longer more exact type is a little too long for my taste. I also don't think we're using anything not defined in `MontyExperiment`.

tristanls-tbp: **note**: I think this falls under the "return the most precise thing" rule. I think it'd be OK to go to `MontyExperiment` on line 151, for example, but adding a type just for that on line 151 seems extraneous. Line 184 seems fine, for example.

tests/mujoco/integration/frameworks/experiments/graph_building_test.py
tristanls-tbp, line 1: **note**: I'm noticing multiple accesses of attributes that are not defined; however, I prefer the mypy errors to continue instead of suppressing them with `# type: ignore[attr-defined]`. This way, we have a signal to write different tests in the future.

tristanls-tbp (approved):
Looks good, thank you.

tests/mujoco/integration/frameworks/experiments/graph_building_test.py
jeremyshoemaker: That was my thinking as well. I don't want to hide that this is not correct. We can fix it later.

PR #1031, opened 2026-06-15
test: create MuJoCo version of profile_experiment_mixin_test.py

jeremyshoemaker:

### DIFF
--- a/src/tbp/monty/frameworks/experiments/profile.py
+++ b/src/tbp/monty/frameworks/experiments/profile.py
@@ -8,8 +8,11 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
+from __future__ import annotations
+
 import cProfile
 from pathlib import Path
+from typing import Any
 
 import pandas as pd
 import wandb
@@ -19,7 +22,7 @@ from tbp.monty.frameworks.experiments.monty_experiment import MontyExperiment
 __all__ = ["ProfileExperimentMixin"]
 
 
-def make_stats_df(stats):
+def make_stats_df(stats: cProfile.Profile) -> pd.DataFrame | None:
     """Convert cProfile.Profile() stats to dataframe.
 
     Take a cProfile.Profile() object, gather stats, put in dataframe, sort by
@@ -48,7 +51,9 @@ class ProfileExperimentMixin:
         NOTE: make sure this class is leftmost in mixin order.
     """
 
-    def __init_subclass__(cls, **kwargs):
+    profile_dir: Path
+
+    def __init_subclass__(cls, /, **kwargs: Any) -> None:
         """Ensure that the mixin is used in the correct way.
 
         We want to ensure that the mixin is always the leftmost class listed in
@@ -71,11 +76,11 @@ class ProfileExperimentMixin:
                 "of MontyExperiment."
             )
 
-    def make_profile_dir(self):
+    def make_profile_dir(self) -> None:
         self.profile_dir = Path(self.output_dir) / "profile"
         self.profile_dir.mkdir(exist_ok=True, parents=True)
 
-    def setup_experiment(self, config):
+    def setup_experiment(self, config) -> None:
         filename = "profile-setup_experiment.csv"
         pr = cProfile.Profile()
         pr.enable()
@@ -86,7 +91,7 @@ class ProfileExperimentMixin:
         df = make_stats_df(pr)
         df.to_csv(self.profile_dir / filename)
 
-    def run_episode(self):
+    def run_episode(self) -> None:
         mode, epoch, episode = self.get_epoch_state()
         filename = f"profile-{mode}_epoch_{epoch}_episode_{episode}.csv"
         pr = cProfile.Profile()
@@ -96,7 +101,7 @@ class ProfileExperimentMixin:
         df = make_stats_df(pr)
         df.to_csv(self.profile_dir / filename)
 
-    def train(self):
+    def train(self) -> None:
         filename = "profile-train.csv"
         pr = cProfile.Profile()
         pr.enable()
@@ -105,7 +110,7 @@ class ProfileExperimentMixin:
         df = make_stats_df(pr)
         df.to_csv(self.profile_dir / filename)
 
-    def evaluate(self):
+    def evaluate(self) -> None:
         filename = "profile-evaluate.csv"
         pr = cProfile.Profile()
         pr.enable()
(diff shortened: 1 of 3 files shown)


tests/mujoco/integration/frameworks/experiments/profile_experiment_mixin_test.py
```diff
@@ -0,0 +1,123 @@
...
+
+from tbp.monty.frameworks.experiments.mode import ExperimentMode
+from tbp.monty.frameworks.experiments.monty_experiment import MontyExperiment
+from tbp.monty.frameworks.experiments.profile import (
+    ProfileExperimentMixin,
+)
+from tests import HYDRA_ROOT
+
+
+class ProfiledExperiment(ProfileExperimentMixin, MontyExperiment):
```
tristanls-tbp, line 26: **issue (non-blocking)**: Mypy in strict mode complains here. The fix is to ensure that `__init_subclass__` is typed in `ProfileExperimentMixin` implementation:

```python
from typing_extensions import ParamSpec

# ...

P = ParamSpec("P")

class ProfileExperimentMixin:
    """Save cProfile traces for each episode.

    Example:
        class ProfiledExp(ProfileExperimentMixin, SomeExp):
            pass

        my_config["experiment_class"] = ProfiledExp

        NOTE: make sure this class is leftmost in mixin order.
    """

    def __init_subclass__(cls, *args: P.args, **kwargs: P.kwargs) -> None:
        """Ensure that the mixin is used in the correct way.

        # ...
```

tristanls-tbp (approved):

tests/mujoco/integration/frameworks/experiments/profile_experiment_mixin_test.py
jeremyshoemaker: I started looking into that, and didn't get that far. I think that pattern of using `ParamSpec` isn't ingrained in my memory yet, but it's getting there.

One thing to note here, I'm pretty sure `__init_subclass__` only takes keyword arguments. Every example I've seen calls the `super()` version with `**kwargs` but not `*args`. See the [documentation on the method itself](https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__), and [this section of the PEP that added it](https://peps.python.org/pep-0487/#:~:text=this%20PEP%20proposes%20to%20replace) that describes the implementation (I could link directly to the CPython implementation, but this is probably close enough).

I'll update this accordingly.

tristanls-tbp: Mypy will complain if you use `P.kwargs` without also using `P.args`. It is my understanding that every python method accepts `*args` and `**kwargs`, that's a python built-in mechanism. We only write out `*args` or `**kwargs` in a method signature if we want a generic handle onto that mechanism to use inside the method. Sure, `*args` will be empty, but it's still there.

jeremyshoemaker: I think I just rediscovered what I ran into when I tried this. `ParamSpec` is specifically for capturing the parameters of a `Callable` to use elsewhere (see [Python docs on ParamSpec](https://docs.python.org/3/library/typing.html#typing.ParamSpec:~:text=They%20are%20only%20valid%20when%20used%20in%20Concatenate%2C%20or%20as%20the%20first%20argument%20to%20Callable%2C%20or%20as%20parameters%20for%20user%2Ddefined%20Generics%2E)). The canonical example is when you're making a decorator and you want the wrapping function to accept the same arguments as the wrapped function and have it type check correctly.

Because the arguments to `__init_subclass__` come from the the _subclass_ definition, there's nothing to tie into to pull the arguments from.

For context, this is how `__init_subclass__` with keyword arguments generally works.
```python
class Base:
    def __init_subclass__(cls, /, foo, **kwargs):
        super().__init_subclass__(**kwargs)
        cls.foo = foo

class Subclass(Base, foo="bar"):
    pass

assert Subclass.foo == "bar"
```

I think the best we could do here is to use `Any`, which I really don't like.

jeremyshoemaker: I think our comments passed each other.

Mypy does seem to type check it with both the args and kwargs, but based on the documentation, I don't think it's valid. The [typing documentation](https://typing.python.org/en/latest/spec/generics.html#paramspec) seems to agree as well. I don't think using it outside of `Callable`, `Concatenate`, or a Generic will be valid.

If you look at how we used it on [the `sim_reset` decorator I added recently](https://github.com/thousandbrainsproject/tbp.monty/blob/d035bc1b8248035f97ab04b16897132b18c472ac/tests/unit/simulators/mujoco/agent_actions_test.py#L60-L81), it's within the context of a `Concatenate` type.

tristanls-tbp: I see. `ParamSpec` is the wrong solution. `Any` appears to be the thing to do. The only other seems to be `object`, which seems wronger.

jeremyshoemaker: I added some additional type hinting to `profile.py`. I was looking into ways to get rid of the remaining `attr-defined` errors, and from what I can tell, the best way would be to define a Protocol containing the attributes and methods that the mixin depends on, and use that either for the `self` arguments of the methods, or as a base for the Mixin.

See [Mypy docs on mixin classes](https://mypy.readthedocs.io/en/latest/more_types.html#mixin-classes). They only specify the `self` argument typing approach here, but that precludes accessing methods and attributes on the mixin itself.

But I think that Protocol should be defined for MontyExperiment elsewhere.

src/tbp/monty/frameworks/experiments/profile.py
```diff
@@ -80,7 +80,7 @@ def make_profile_dir(self) -> None:
         self.profile_dir = Path(self.output_dir) / "profile"
         self.profile_dir.mkdir(exist_ok=True, parents=True)
 
-    def setup_experiment(self, config) -> None:
+    def setup_experiment(self, config: dict[str, Any]) -> None:
```
tristanls-tbp, line 83: **question**: Is this true? `.setup_experiment` gets called with `DictConfig`

```python
class MontyExperiment:
    """General Monty experiment class used to run sensorimotor experiments.

    This class implements the framework for setting up an environment interface and
    Monty model, the outermost loops for training and evaluating (including run epoch
    and episode).
    """

    def __init__(self, config: DictConfig) -> None:
        """Initialize the experiment based on the provided configuration.

        Args:
        """
        self.config = config
        # ...
```

jeremyshoemaker: I was going off the type signature of `MontyExperiment.setup_experiment`, but looking more closely, it appears that the type signature there is no longer accurate. It should be `DictConfig` there as well. 

I think that's well out of scope here though, so I'll leave `ProfileExperimentMixin` as it is for now and we can change them together at a later date.

PR #1033, opened 2026-06-16
refactor!: create MuJoCo version of graph_learning_test.py

jeremyshoemaker:

Outside of the new `/environment` options, the config changes consist of swapping the `/environment` from a Habitat version to an equivalent MuJoCo version.

I've added type hints to **most** of the functions and methods called from the test, to satisfy Mypy. There a small number remaining that either I couldn't determine the correct type easily, or setting the type would involve more invasive changes. I figured they could be delayed until a latter date.

> [!warning]
> **Breaking Changes**
> - Changed FOV calculation to correctly calculate for zooms

### DIFF
--- a/src/tbp/monty/conf/environment/mujoco_ycb_dist_agent.yaml
+++ b/src/tbp/monty/conf/environment/mujoco_ycb_dist_agent.yaml
@@ -38,6 +38,6 @@ env_init_args:
       position:
         - 0.0
         - 1.5
-        - 0.5
+        - 0.2
   raise_actuate_missing: False
 env_init_func: ${monty.class:tbp.monty.simulators.mujoco.MuJoCoSimulator}
--- a/src/tbp/monty/simulators/mujoco/agents.py
+++ b/src/tbp/monty/simulators/mujoco/agents.py
@@ -66,8 +66,8 @@ def zoom_fovy(base_fovh: float, resolution: Resolution2D, zoom: float) -> float:
     """
     resolution_ratio = resolution.height / resolution.width
     base_fovh_radians = np.deg2rad(base_fovh)
-    zoom_fovy_radians = 2 * np.arctan(
-        resolution_ratio * np.tan(base_fovh_radians / 2.0) / zoom
+    zoom_fovy_radians = 2 * np.arctan2(
+        resolution_ratio * np.tan(base_fovh_radians / 2.0), zoom
     )
     return np.rad2deg(zoom_fovy_radians)
 
--- a/src/tbp/monty/simulators/mujoco/agents.py
+++ b/src/tbp/monty/simulators/mujoco/agents.py
@@ -35,7 +35,7 @@ from tbp.monty.frameworks.models.abstract_monty_classes import (
     SensorObservation,
 )
 from tbp.monty.frameworks.models.motor_system_state import AgentState, SensorState
-from tbp.monty.frameworks.sensors import SensorConfig, SensorID
+from tbp.monty.frameworks.sensors import Resolution2D, SensorConfig, SensorID
 from tbp.monty.geometry import Rotation
 from tbp.monty.math import IDENTITY_QUATERNION, ZERO_VECTOR, QuaternionWXYZ, VectorXYZ
 
@@ -45,9 +45,31 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-# The default field of view value for zoom 1.0
-# Note: this value is the half-FOV rather than the full FOV
-DEFAULT_CAMERA_FOVY: float = 45.0
+# The default horizontal field of view value for zoom 1.0
+# This value needs to stay in sync with the value in the `DepthTo3DLocations` transform.
+# TODO: Make this configurable
+DEFAULT_CAMERA_FOVH: float = 90.0
+
+
+def zoom_fovy(base_fovh: float, resolution: Resolution2D, zoom: float) -> float:
+    """Calculate the vertical field of view for a given zoom level and resolution.
+
+    This helper function takes the base horizontal field of view along with the
+    resolution and requested zoom level and returns the correct vertical field of
+    view angle for that zoom level.
+
+    MuJoCo only allows us to set the vertical field of view for a camera, and
+    DepthTo2DLocations uses horizontal field of view.
+
+    Returns:
+        zoomed vertical field of view angle in degrees.
+    """
+    resolution_ratio = resolution.height / resolution.width
+    base_fovh_radians = np.deg2rad(base_fovh)
+    zoom_fovy_radians = 2 * np.arctan(
+        resolution_ratio * np.tan(base_fovh_radians / 2.0) / zoom
+    )
+    return np.rad2deg(zoom_fovy_radians)
 
 
 class Axis(IntEnum):
@@ -132,7 +154,11 @@ class Embodiment(Agent):
                 pos=sensor_cfg.position,
                 quat=sensor_cfg.rotation,
                 # Camera resolution isn't used in MuJoCo, so we're not setting it.
-                fovy=DEFAULT_CAMERA_FOVY / sensor_cfg.zoom,
+                fovy=zoom_fovy(
+                    base_fovh=DEFAULT_CAMERA_FOVH,
+                    resolution=sensor_cfg.resolution,
+                    zoom=sensor_cfg.zoom,
+                ),
             )
 
     @property
--- a/src/tbp/monty/simulators/mujoco/agents.py
+++ b/src/tbp/monty/simulators/mujoco/agents.py
@@ -59,7 +59,7 @@ def zoom_fovy(base_fovh: float, resolution: Resolution2D, zoom: float) -> float:
     view angle for that zoom level.
 
     MuJoCo only allows us to set the vertical field of view for a camera, and
-    DepthTo2DLocations uses horizontal field of view.
+    DepthTo3DLocations uses horizontal field of view.
 
     Returns:
         zoomed vertical field of view angle in degrees.
--- a/tests/mujoco/integration/frameworks/models/graph_learning_test.py
+++ b/tests/mujoco/integration/frameworks/models/graph_learning_test.py
@@ -26,6 +26,7 @@ from tbp.monty.cmp import Message
 from tbp.monty.context import RuntimeContext
 from tbp.monty.frameworks.actions.actions import Action
 from tbp.monty.frameworks.experiments.mode import ExperimentMode
+from tbp.monty.frameworks.models.abstract_monty_classes import Monty
 from tbp.monty.frameworks.models.feature_location_matching import FeatureGraphLM
 from tbp.monty.frameworks.models.graph_matching import GraphLM
 from tbp.monty.frameworks.utils.logging_utils import (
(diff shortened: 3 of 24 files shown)


tests/mujoco/integration/frameworks/models/graph_learning_test.py
```diff
@@ -0,0 +1,986 @@
...
+        for row in range(eval_stats.shape[0]):
+            detailed_id = train_stats.shape[0] + row - 1
+            self.assertEqual(
+                eval_stats.loc[row]["primary_target_object"],
+                detailed_stats[str(detailed_id)]["LM_0"]["target"]["object"],
+                "targets in eval_stats and detailed stats don't match.",
+            )
+        self.assertEqual(
+            len(detailed_stats["1"]["SM_0"]["processed_observations"]),
+            200,
```
jeremyshoemaker: This is the only line that changed in the tests (outside of the config changes in the `setUp` method).

The previous value here was `73`. I'm not sure the significance of that value to judge whether this is a regression or not.

tests/mujoco/integration/frameworks/models/graph_learning_test.py
```diff
@@ -0,0 +1,986 @@
...
+                "on object",
+            )
+
+    def test_detailed_logging(self) -> None:
+        exp = hydra.utils.instantiate(self.feature_pred_off_object_cfg.experiment)
+        with exp:
+            exp.run()
+
+        # TODO: improve load_stats with a structured return type
+        train_stats, eval_stats, detailed_stats, lm_models = load_stats(
```
tristanls-tbp, line 506: **issue (non-blocking)**: Mypy complains about a call to untyped `load_stats` (first param is untyped).

tests/mujoco/integration/frameworks/models/graph_learning_test.py
tristanls-tbp: **issue**: I think this is a problem. `max_total_steps: 200` in the `test/graph_learning/feature_pred_off_object_mujoco` experiment tells me something ran until it reached the limit without converging first. 

@vkakerbeck can you help interpret what's going on here?

tests/mujoco/integration/frameworks/models/graph_learning_test.py
```diff
@@ -0,0 +1,986 @@
...
+        for observation in observations:
+            obs_to_learn = copy.deepcopy(observation)
+            obs_to_learn.location += offset
+            offset_obs.append(obs_to_learn)
+            graph_lm.exploratory_step(self.ctx, [obs_to_learn])
+
+        graph_lm.detected_object = obj_name
+        graph_lm.detected_rotation_r = None
+        graph_lm.buffer.stats["detected_location_rel_body"] = (
+            graph_lm.buffer.get_current_location(input_channel="first")
```
tristanls-tbp, line 588: **issue (non-blocking)**: Mypy complains about a call to an untyped function.

tests/mujoco/integration/frameworks/models/graph_learning_test.py
```diff
@@ -0,0 +1,986 @@
...
+        graph_lm.fixme_reset_ground_truth(primary_target=self.placeholder_target)
+
+        offset_obs = []
+        for observation in observations:
+            obs_to_learn = copy.deepcopy(observation)
+            obs_to_learn.location += offset
+            offset_obs.append(obs_to_learn)
+            graph_lm.exploratory_step(self.ctx, [obs_to_learn])
+
+        graph_lm.detected_object = obj_name
```
tristanls-tbp, line 585: **issue (non-blocking)**: Mypy complains `detected_object` is `None` type (because it gets assigned `None` inside the LM before ever getting a value).

tristanls-tbp (commented):
Submitting a review in progress. Reviewed everything fully, except the test file. I got up to `test_reverse_sequence_recognition`, from which I'll resume tomorrow.

tests/mujoco/integration/frameworks/models/graph_learning_test.py
jeremyshoemaker: I left this one as-is and added the TODO comment because it's unclear to me what the final argument type should be (it's not used here), and the return type is atrocious. It's something like:

```python
tuple[DataFrame | None, DataFrame | None, DataFrame | None, DataFrame | None]
```

We should probably be returning a dataclass here rather than a tuple. To me, anything beyond a 2-tuple in a return gets unwieldly fast.

tests/mujoco/integration/frameworks/models/graph_learning_test.py
jeremyshoemaker: I left this one as-is for now because supplying at type correctly at the top of the class leads to wanting to type all of the class' instance variables, and that seems out of scope. I don't like the idea of doing only part of the job (though we're kinda doing that elsewhere typing singular methods and functions).

tests/mujoco/integration/frameworks/models/graph_learning_test.py
jeremyshoemaker: I left this one as-is because I couldn't sort out what the return type should be.

tests/mujoco/integration/frameworks/models/graph_learning_test.py
jeremyshoemaker: That was my concern as well, since it felt like we were setting it to the maximum. I think this is part of the broader problem I mentioned to you yesterday with migrating `hierarchy_test.py`. Some of these tests expect specific behaviours that for whatever reason aren't the same with MuJoCo. I'm not sure the best way to shake out what the differences are, and whether we're missing some key element in the MuJoCo simulator that we had in Habitat.

tristanls-tbp (commented):
At this point, I completed the review. The one issue I mentioned before remains. I think it would be helpful to get a researcher to help us understand what's going on with that test failure.

@vkakerbeck, I am adding you as an assignee to see if you can help or recommend someone else who can help us [regarding the test failure issue raised](https://github.com/thousandbrainsproject/tbp.monty/pull/1033#discussion_r3424198543).

tests/mujoco/integration/frameworks/models/graph_learning_test.py
jeremyshoemaker: Just to document what @vkakerbeck and I found this morning. It looks like this is occuring because the field of view of the camera is set wrong. When we were working on the prototype in London, something lead us to believe that the `fovy` value that MuJoCo cameras take was _half_ of the FOV, but it's looking like it's actually the full value.

In the MuJoCo version of the experiment we're getting off the object about twice as fast as in the Habitat version, despite it taking the same amount of steps to rotate 360 degrees (the test just spins the sensor to the left in 5 degree increments).

It wasn't clear what the problem was here until we noticed some distortion while slowly stepping through the experiments side by side in each simulator.

I think the fix is going to be to adjust the `DEFAULT_CAMERA_FOVY` value in the MuJoCo simulator, and adjust the scaling factor for the objects because with the correct FOV they don't look to be the same size as the Habitat ones.

While pairing we were able to get the number in this test down to `142` from `200`, so I'd like to see if we can get it closer to the original.

jeremyshoemaker: So... I'm shocked.

After getting the FOV calculations corrected, I'm getting the **exact same** results as we were getting under Habitat.

I wasn't expecting that to happen, especially given my scaling factor being estimated for the primitive objects.

src/tbp/monty/conf/environment/mujoco_ycb_dist_agent.yaml
```diff
@@ -38,6 +38,6 @@ env_init_args:
       position:
         - 0.0
         - 1.5
-        - 0.5
+        - 0.2
```
jeremyshoemaker, line 41: I'm not sure how this got changed, probably something copied incorrectly a while ago, but the 0.2 value is what Habitat has, and with all the scaling and FOV changes, the MuJoCo version works with the original value.

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -45,9 +45,31 @@
...
+
+def zoom_fovy(base_fovh: float, resolution: Resolution2D, zoom: float) -> float:
+    """Calculate the vertical field of view for a given zoom level and resolution.
+
+    This helper function takes the base horizontal field of view along with the
+    resolution and requested zoom level and returns the correct vertical field of
+    view angle for that zoom level.
+
+    MuJoCo only allows us to set the vertical field of view for a camera, and
+    DepthTo2DLocations uses horizontal field of view.
```
tristanls-tbp:
```suggestion
    DepthTo3DLocations uses horizontal field of view.
```

tristanls-tbp (commented):
Looks good to me. One typo.

I'm waiting to approve until either @scottcanoe or @hlee9212 review the [field of view calculation portion](https://github.com/thousandbrainsproject/tbp.monty/pull/1033/changes#diff-2d79d265d6816cdff3f10336635ef6445a4a9a27c6a31d037c09c3d5d8b7fbe9R67)

tristanls-tbp:
**praise**: Also, I should add, great work figuring this out.

**suggestion (non-blocking)**: This might be a good time to write an explainer in the documentation while you still remember how it all works.

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -45,9 +45,31 @@
...
+    DepthTo3DLocations uses horizontal field of view.
+
+    Returns:
+        zoomed vertical field of view angle in degrees.
+    """
+    resolution_ratio = resolution.height / resolution.width
+    base_fovh_radians = np.deg2rad(base_fovh)
+    zoom_fovy_radians = 2 * np.arctan(
+        resolution_ratio * np.tan(base_fovh_radians / 2.0) / zoom
+    )
```
jeremyshoemaker: I think I tried to use `arctan2` at first, but didn't have something for the second argument. At that point I was missing the `zoom` term, and after fixing it and looking at it again, I think I can change this to use `arctan2` instead.

jeremyshoemaker:
> This might be a good time to write an explainer in the documentation while you still remember how it all works.

Which part are you thinking of? The FOV calculations or something else?

tristanls-tbp:
> Which part are you thinking of? The FOV calculations or something else?

Sorry for the lack of specificity there. I was thinking of the FOV calculations. The whole thread of conversation started with trying to figure out the units, what is being calculated, and whether it worked only for square images, etc. All of that is probably crystal clear in your mind right now, hence the suggestion to capture it.

jeremyshoemaker:
> The whole thread of conversation started with trying to figure out the units, what is being calculated, and whether it worked only for square images, etc. All of that is probably crystal clear in your mind right now, hence the suggestion to capture it.

Documenting it would be a good idea, at least to make sure _I_ understand it.

As for "trying to figure out the units", that was actually for focal length, which I don't think I got an answer for. I don't think we need to change those settings. It was just another parameter that I thought might be causing problems, but I don't think so given the results I've gotten after fixing the field of view calculations.

tristanls-tbp:
> Documenting it would be a good idea, at least to make sure I understand it.

You'll also be teaching me how it works, and anyone else who comes across the repo.

hlee9212 (approved):
I only looked at `zoom_fovy` in `agents.py` and the equation is correct.

vkakerbeck (approved):
Looks good from my side too. Nice debugging, and thanks for the help on calculating the fovy @hlee9212 and @scottcanoe This was a tricky one!

tristanls-tbp (approved):
Thank you for the confirmation @hlee9212. Everything else looks right to me 👍.

PR #1016, opened 2026-06-09
feat!: live plotter prototype

ramyamounir:
This is a breaking change because it eliminates `show_sensor_output` from the experiment configs and replaces it with a Hydra `plotter` package.

PR #1034, opened 2026-06-16
feat: experiment StepHook

tristanls-tbp:
This pull request adds a `StepHook` into the experimental framework. This allows us to use the https://github.com/thousandbrainsproject/tbp.teleop tool to control and visualize an experiment in real time.

This PR supersedes https://github.com/thousandbrainsproject/tbp.monty/pull/1016

https://github.com/user-attachments/assets/e2fa0f3c-0e31-4b46-87db-25b2b85c6e86

### DIFF
--- a/src/tbp/monty/frameworks/experiments/hooks.py
+++ b/src/tbp/monty/frameworks/experiments/hooks.py
@@ -36,16 +36,26 @@ class StepHook(Protocol):
     ) -> list[Action]:
         """Execute the step hook.
 
+        The step hook is used to customize the step behavior on behalf of the
+        experiment. The hook occurs after the Monty model output actions intended for
+        the environment, but before the environment is stepped.
+
+        Having a hook at this point in the execution loop allows for visualization of
+        observations, resultant actions, as well as any internal state of the Monty
+        model available to the experiment. Additionally, the hook can be used to
+        augment, override, modify, delete, or otherwise customize the actions intended
+        for the environment.
+
         Args:
             ctx: The runtime context.
             monty: The Monty model.
             supervised_lm_ids: The list of supervised learning module IDs.
             step: The current step.
-            observations: The observations.
-            actions: The actions to take.
+            observations: The observations provided to the Monty model.
+            actions: The actions output by the Monty model intended for the environment.
 
         Returns:
-            The actions to take.
+            The actions to take by the environment.
         """
         ...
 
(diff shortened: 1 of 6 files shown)


jeremyshoemaker (commented):
I don't think this should be named for the specific tool we're making it for. I'm trying to think of a good alternative.

It's potentially overriding the action, but that's not all it's doing. Maybe just `ActionHook`? As you said in the engineering meeting earlier, this is experiment code that we eventually want to pull out of the main Monty repo, so maybe it doesn't matter.

tristanls-tbp:
@jeremyshoemaker, what do you think of `StepHook`?

jeremyshoemaker:
> @jeremyshoemaker, what do you think of `StepHook`?

I was thinking of something like that, I just didn't include it in the comment. Seems fine to me.

tristanls-tbp:
@jeremyshoemaker I updated name of this thing to `StepHook` throughout.

src/tbp/monty/frameworks/experiments/hooks.py
```diff
@@ -0,0 +1,72 @@
...
+        Args:
+            ctx: The runtime context.
+            monty: The Monty model.
+            supervised_lm_ids: The list of supervised learning module IDs.
+            step: The current step.
+            observations: The observations.
+            actions: The actions to take.
+
+        Returns:
+            The actions to take.
```
jeremyshoemaker: I think the documentation could use a little work here. The `actions` coming in isn't _exactly_ the actions to take, but rather the actions that Monty is planning/planned to take. The "Returns" could use some guidance that the hook should return the `actions` list that came in if it doesn't want to override the actions that Monty planned.

jeremyshoemaker (approved):
I'm gonna go ahead and approve this. I have some comments about the documentation on the StepHook protocol that I didn't think of previously, but they're not blocking.

ramyamounir (approved):
Thanks for putting this together, looks good to me.

src/tbp/monty/frameworks/experiments/hooks.py
tristanls-tbp: Thank you. [I elaborated the documentation](https://github.com/thousandbrainsproject/tbp.monty/pull/1034/changes/b86561c6c2274fc363a0c3798fa862ee2cb791d8).

jeremyshoemaker:
The changes to the docs look much better. I'd have liked to have the "return the actions you got to not change what happens" be more explicitly called out but at the end of the day, this is in experiment code, so I'm fine with it. And the no-op version kinda gives an example to go off of.

PR #1035, opened 2026-06-17
refactor!: split Runtime and Experiment aspects of Monty

dschumacher-tbp:
The methods of `Monty` are segregated into "Methods that define the algorithm" and "Methods that interact with the experiment". This PR extracts a `RuntimeMonty` Protocol for the former, and an `ExperimentMonty` Protocol for the latter.

This segregation is in preparation for moving Experiment-related logic from Monty into the Experiment.

## Breaking Changes

- `Monty.pre_episode()` is split into `ExperimentMonty.reset()` and `ExperimentMonty.fixme_set_ground_truth()`
- `MontyForGraphMatching` (and its subclasses) use `ExperimentMonty.fixme_set_ground_truth()` to pass additional information into the model
- `Monty.post_episode()` becomes `ExperimentMonty.update_ltm()`

### DIFF
--- a/src/tbp/monty/experiment/monty.py
+++ b/src/tbp/monty/experiment/monty.py
@@ -1,0 +1,58 @@
+# Copyright 2026 Thousand Brains Project
+#
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+from __future__ import annotations
+
+from typing import Any, Protocol
+
+from tbp.monty.frameworks.environments.environment import SemanticID
+from tbp.monty.frameworks.experiments.mode import ExperimentMode
+
+__all__ = [
+    "ExperimentMonty",
+]
+
+
+class ExperimentMonty(Protocol):
+    """Experiment interface to Monty model."""
+
+    def reset(self) -> None:
+        """Reset the internal state of this Monty model."""
+        ...
+
+    def fixme_set_ground_truth(
+        self,
+        primary_target: dict[str, Any] | None = None,
+        semantic_id_to_label: dict[SemanticID, str] | None = None,
+    ) -> None:
+        """Provide ground truth from experiment supervision.
+
+        Args:
+            primary_target: Optional primary target to recognize.
+            semantic_id_to_label: Optional mapping from IDs to labels.
+        """
+        ...
+
+    def update_ltm(self) -> None:
+        """Transfer short-term buffer to long-term memory."""
+        ...
+
+    def set_experiment_mode(self, mode: ExperimentMode) -> None:
+        """Set the experiment mode.
+
+        Update state variables based on which method (train or evaluate) is being
+        called at the experiment level.
+
+        Args:
+            mode: The experiment mode.
+        """
+        ...
+
+    def is_done(self) -> bool:
+        """Return `True` if the model has reached a terminal condition."""
+        ...
--- a/src/tbp/monty/frameworks/models/abstract_monty_classes.py
+++ b/src/tbp/monty/frameworks/models/abstract_monty_classes.py
@@ -71,7 +71,66 @@ class Observations(Dict[AgentID, AgentObservations]):
     pass
 
 
-class Monty(ExperimentMonty, Snapshotable, metaclass=abc.ABCMeta):
+class RuntimeMonty(Protocol):
+    """Runtime interface to Monty."""
+
+    def step(
+        self,
+        ctx: RuntimeContext,
+        observations: Observations,
+        proprioceptive_state: ProprioceptiveState,
+    ) -> list[Action]:
+        """Take a matching, exploratory, or custom user-defined step.
+
+        Step taken depends on the value of self.step_type.
+
+        Args:
+            ctx: The runtime context.
+            observations: The observations from the environment.
+            proprioceptive_state: The proprioceptive state from the environment.
+
+        Returns:
+            The actions to take.
+        """
+        ...
+
+    def motor_only_step(
+        self,
+        ctx: RuntimeContext,
+        observations: Observations,
+        proprioceptive_state: ProprioceptiveState,
+    ) -> list[Action]:
+        """Take a step of the sensors and motor system only.
+
+        This skips stepping the learning modules.
+
+        Args:
+            ctx: The runtime context.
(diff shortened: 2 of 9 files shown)


src/tbp/monty/experiment/monty.py
```diff
@@ -0,0 +1,58 @@
...
+        primary_target: dict[str, Any] | None = None,
+        semantic_id_to_label: dict[SemanticID, str] | None = None,
+    ) -> None:
+        """Provide ground truth from experiment supervision.
+
+        Args:
+            primary_target: Optional primary target to recognize.
+            semantic_id_to_label: Optional mapping from IDs to labels.
+        """
+        ...
```
jeremyshoemaker, line 39: Maybe we should add a note to the docstring about why this method has `fixme` in the name. I don't know if it's worth it though since all of this is temporary (how long it's temporary is the question).

src/tbp/monty/experiment/monty.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def update_ltm(self) -> None:
+        """Transfer short-term buffer to long-term memory."""
+        ...
+
+    def set_experiment_mode(self, mode: ExperimentMode) -> None:
```
jeremyshoemaker, line 45: Given the temporary nature of all of this, I'm not sure if it's worth the effort, but one thing we've been trying to do is migrate away from the `get_*` and `set_*` pattern to using `@property` instead to define a getter and setter property.

Changing it might be more trouble than it's worth. One interesting thing here is that we're defining a setter for this field in the protocol, but not the getter. That seems... odd.

src/tbp/monty/frameworks/models/abstract_monty_classes.py
```diff
@@ -69,7 +71,66 @@ class Observations(Dict[AgentID, AgentObservations]):
...
+        """
+        ...
+
+    def aggregate_sensory_inputs(
+        self,
+        ctx: RuntimeContext,
+        observations: Observations,
+        proprioceptive_state: ProprioceptiveState,
+    ) -> None:
+        """Receive data from environment, organize on a per sensor module basis.
```
jeremyshoemaker, line 123:
```suggestion
        """Receive data from environment, organized on a per sensor module basis.
```

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
```diff
@@ -49,47 +52,57 @@ class MontyForNoResetEvidenceGraphMatching(MontyForEvidenceGraphMatching):
...
+
         # keep target up-to-date for logging
         self.primary_target = primary_target
         self.semantic_id_to_label = semantic_id_to_label
         for lm in self.learning_modules:
             lm.primary_target = primary_target["object"]
             lm.primary_target_rotation_quat = primary_target["quat_rotation"]
 
-        # reset LMs and SMs buffers to save memory
-        self._reset_modules_buffers()
```
jeremyshoemaker, line 91: Why does `fixme_set_ground_truth` not need to call this anymore? I'm not saying it's wrong, I'm just not clear on why this changed.

Is it because we're usually calling `reset` right before calling this method?

I'm wondering if it might make more sense to change the call sites to only call this method and put this back. It seems brittle to have to remember to call `reset` before `fixme_set_ground_truth`. Maybe the ordering is important?

All that said, I think this is only used in test code, so maybe it doesn't matter that much. Especially given that it's a "FIXME" method.

tests/integration/frameworks/models/graph_learning_test.py
```diff
@@ -953,10 +953,11 @@ def test_5lm_feature_experiment(self):
             for episode_num in range(tm.num_episodes):
                 exp.pre_episode()
                 # Normally the experiment `pre_episode` method would call the model
-                # `pre_episode` method, but it expects to feed data from an environment
+                # `reset` method, but it expects to feed data from an environment
                 # interface to the model, and we aren't using that, so we call it again
                 # with the correct target value.
-                monty.pre_episode(self.placeholder_target)
+                monty.reset()
+                monty.fixme_set_ground_truth(self.placeholder_target)
```
jeremyshoemaker, line 960: I need to make a note to myself to also change this in #1033 because I made the copy before this change.

This kinda highlights my concern about test drift we were all discussing the other day.

jeremyshoemaker (approved):
This all looks good to me. I don't have any real concerns. If it was more long-term code I might have some comments about choice of names for the two protocols, but I think eventually `RuntimeMonty` will just get folded into whatever Protocol that the `Monty` ABC turns into.

I'll let you respond to any of my comments before I hit the merge button.

src/tbp/monty/experiment/monty.py
dschumacher-tbp: The `fixme_` prefix is meant to be a marker of deprecated code, so I don't think more explanation is needed.

src/tbp/monty/experiment/monty.py
dschumacher-tbp: I'm not sure refactoring to properties is within the scope of splitting the episode logic into runtime and experiment affordances.

src/tbp/monty/frameworks/models/abstract_monty_classes.py
dschumacher-tbp: If I understand the code properly, it is "organizing" the environment data on a per SM basis.

jeremyshoemaker: Ah, yeah I think you're right.

src/tbp/monty/frameworks/models/no_reset_evidence_matching.py
dschumacher-tbp: The idea was to separate the experiment-triggered reset logic from the ground-truth hints formerly in `pre_episode`. This logic moved to `reset` instead of `fixme_set_ground_truth`, with the goal of removing the fixme code sooner.

tests/integration/frameworks/models/graph_learning_test.py
dschumacher-tbp: I highlighted this in the "Breaking Changes" part of the PR description.

jeremyshoemaker: Yep, it's just something I would have missed, so making this comment here links this PR to mine and hopefully helps me remember to fix it.

PR #1039, opened 2026-06-19
test: create MuJoCo version of hierarchy_test.py

jeremyshoemaker:

There's one substantive change to the tests here, and that's adding "cubeSolid" to the list of objects to one of the checks in the second test. There was a TODO about making it so that it would recognize "cubeSolid" and when I tried it both in the original and in the MuJoCo version, the test passed, so I think we can make the change and remove the TODO.

There's also a bug fix that I didn't catch until this test hit the code path that would trigger it. I was accidentally passing a generator to `np.array` instead of the collection it was supposed to generate. I was under the impression that this would work as originally written, but I might be misremembering something. What ends up happening is we get an `np.array` with a single element that is the unrealized generator, which obviously isn't what we want.

### DIFF
--- a/src/tbp/monty/conf/environment/test_mujoco_2lm_dist_transforms.yaml
+++ b/src/tbp/monty/conf/environment/test_mujoco_2lm_dist_transforms.yaml
@@ -1,0 +1,31 @@
+# @package experiment.config.environment
+
+env_init_args:
+  agents:
+    - _target_: tbp.monty.simulators.mujoco.agents.DistantAgent
+      _partial_: true
+      agent_id: agent_id_0
+      sensor_configs:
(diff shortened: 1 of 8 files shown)


tristanls-tbp (commented):
**issue**: Test is missing return and method argument types throughout.

Aside from that, the test looks like a faithful copy-and-paste 🙂.

jeremyshoemaker:
> **issue**: Test is missing return and method argument types throughout.

I was wondering if I'd forgotten to do that for this one. I was halfway through another one when I thought of it and wanted to finish up before switching back.

tristanls-tbp (approved):
Thanks!

PR #1040, opened 2026-06-19
test: create MuJoCo version of no_reset_evidence_lm_test.py

jeremyshoemaker:

No substantive changes here. I moved the `tearDown` method to the top of the class because I prefer the setup and teardown code to be adjacent to each other. I also removed some of the redundancy in the test docstring. The original version repeated itself and also described _how_ the code does what it does which is prone to drift when the code changes.

### DIFF
--- a/tests/mujoco/integration/frameworks/models/no_reset_evidence_lm_test.py
+++ b/tests/mujoco/integration/frameworks/models/no_reset_evidence_lm_test.py
@@ -1,0 +1,126 @@
+# Copyright 2025-2026 Thousand Brains Project
+#
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+from __future__ import annotations
+
+import copy
+import shutil
+import tempfile
+from typing import Any
+
+import hydra
+import numpy as np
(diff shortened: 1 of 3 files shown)


tests/mujoco/integration/frameworks/models/no_reset_evidence_lm_test.py
```diff
@@ -0,0 +1,126 @@
...
+                ],
+            )
+
+    def tearDown(self) -> None:
+        shutil.rmtree(self.output_dir)
+
+    def assert_dicts_equal(
+        self,
+        d1: dict[Any, npt.NDArray[np.float64]],
+        d2: dict[Any, npt.NDArray[np.float64]],
```
tristanls-tbp: **issue**: `dict[str, ...`. There's no need for this method to expect non-string dictionary keys.

**note**: Thank you for cleaning up the value types.

tristanls-tbp (commented):
Aside from one typing issue I think is worth fixing, it looks good 👍.

tests/mujoco/integration/frameworks/models/no_reset_evidence_lm_test.py
jeremyshoemaker: Didn't even see the Any. I was focused on sorting out the array type.

tristanls-tbp (approved):

PR #1049, opened 2026-06-19
test: create MuJoCo version of policy_test.py

jeremyshoemaker:

I've split the single test class in `policy_test.py` into two separate tests, along the lines of the original comments that divided things up.

There were no changes to the tests required to make these pass, which I find to be _very_ surprising. A lot of the tests here are dependent on specific observations and specific sequences of actions, and we're getting almost the exact same results.

I did need to adjust the near Z clipping plane for `test_surface_policy_orientation` because when an agent like a SurfaceAgent gets too close to an object at an oblique angle, the nearer side of the surface polygon can end up clipped. In this case the near side of the cube as getting clipped so much that Monty thought it was off-object.

### DIFF
--- a/src/tbp/monty/cmp.py
+++ b/src/tbp/monty/cmp.py
@@ -46,7 +46,7 @@ class Message:
     def __init__(
         self,
         # TODO: Consider changing to VectorXYZ
-        location: npt.NDArray[np.float64],
+        location: npt.NDArray[np.float64] | None,
         morphological_features: dict[str, Any],
         non_morphological_features: dict[str, Any],
         confidence: float,
@@ -236,7 +236,7 @@ class Goal(Message):
 
     def __init__(
         self,
-        location: np.ndarray | None,
+        location: npt.NDArray[np.float64] | None,
         morphological_features: dict[str, Any] | None,
         non_morphological_features: dict[str, Any] | None,
         confidence: float,
--- a/src/tbp/monty/cmp.py
+++ b/src/tbp/monty/cmp.py
@@ -9,9 +9,10 @@
 # https://opensource.org/licenses/MIT.
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, Literal
 
 import numpy as np
+import numpy.typing as npt
 
 from tbp.monty.frameworks.models.buffer import BufferEncoder
 
@@ -44,13 +45,14 @@ class Message:
 
     def __init__(
         self,
-        location,
-        morphological_features,
-        non_morphological_features,
-        confidence,
-        use_state,
-        sender_id,
-        sender_type,
+        # TODO: Consider changing to VectorXYZ
+        location: npt.NDArray[np.float64],
+        morphological_features: dict[str, Any],
+        non_morphological_features: dict[str, Any],
+        confidence: float,
+        use_state: bool,
+        sender_id: str,
+        sender_type: Literal["SM", "LM"],
     ):
         """Initialize a message."""
         self.location = location
--- a/src/tbp/monty/simulators/mujoco/simulator.py
+++ b/src/tbp/monty/simulators/mujoco/simulator.py
@@ -195,6 +195,10 @@ class MuJoCoSimulator(SimulatedObjectEnvironment):
         g = self.spec.visual.global_
         g.offheight = render_resolution.height
         g.offwidth = render_resolution.width
+        # Set the near Z clipping plane a little closer to the camera so we don't
+        # clip objects when an agent is very close to an object, e.g. with the
+        # SurfaceAgent.
+        self.spec.visual.map.znear = 0.001
 
     def _configure_lights(self) -> None:
         """Configure the lights as needed.
--- a/src/tbp/monty/simulators/mujoco/simulator.py
+++ b/src/tbp/monty/simulators/mujoco/simulator.py
@@ -195,11 +195,15 @@ class MuJoCoSimulator(SimulatedObjectEnvironment):
         g = self.spec.visual.global_
         g.offheight = render_resolution.height
         g.offwidth = render_resolution.width
-        # Set the extent to 1.0, as the znear and zfar values are multiplied
-        # by the extent to get the values for near and far clipping, and this
-        # results in near clipping of objects that are close to the camera,
-        # e.g. when using SurfaceAgent.
-        self.spec.stat.extent = 1.0
+        # Habitat used a znear and zfar of 0.01 and 1000.0 respectively. MuJoCo
+        # lets use set these values directly, but it multiplies them by an `extent`
+        # value, which defaults to 2.0. Trying to set this to 1.0 and then setting
+        # the Z clipping values to match Habitat, we see a reduction in depth accuracy.
+        # By leaving the `extent` alone and setting values that will result in matching
+        # the Habitat values, we get better depth values, and avoid near clipping
+        # issues we were getting with the default znear of `0.01`.
+        self.spec.visual.map.znear = 0.005
+        self.spec.visual.map.zfar = 500.0
 
     def _configure_lights(self) -> None:
         """Configure the lights as needed.
--- a/src/tbp/monty/simulators/mujoco/simulator.py
+++ b/src/tbp/monty/simulators/mujoco/simulator.py
@@ -195,10 +195,11 @@ class MuJoCoSimulator(SimulatedObjectEnvironment):
         g = self.spec.visual.global_
         g.offheight = render_resolution.height
         g.offwidth = render_resolution.width
-        # Set the near Z clipping plane a little closer to the camera so we don't
-        # clip objects when an agent is very close to an object, e.g. with the
-        # SurfaceAgent.
-        self.spec.visual.map.znear = 0.001
+        # Set the extent to 1.0, as the znear and zfar values are multiplied
+        # by the extent to get the values for near and far clipping, and this
+        # results in near clipping of objects that are close to the camera,
+        # e.g. when using SurfaceAgent.
+        self.spec.stat.extent = 1.0
 
     def _configure_lights(self) -> None:
         """Configure the lights as needed.
--- a/tests/mujoco/integration/frameworks/models/policy_test.py
+++ b/tests/mujoco/integration/frameworks/models/policy_test.py
@@ -1,0 +1,1115 @@
+# Copyright 2025-2026 Thousand Brains Project
+# Copyright 2022-2024 Numenta Inc.
+#
+# Copyright may exist in Contributors' modifications
+# and/or contributions to the work.
+#
+# Use of this source code is governed by the MIT
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+from __future__ import annotations
+
+import copy
+import shutil
+import tempfile
+import unittest
+from typing import Any, MutableMapping
+
+import hydra
+import numpy as np
+import quaternion as qt
+from omegaconf import DictConfig
+
+from tbp.monty.cmp import Goal, Message
+from tbp.monty.context import RuntimeContext
+from tbp.monty.frameworks.actions.actions import (
+    Action,
+    LookDown,
+    LookUp,
+    MoveForward,
+    MoveTangentially,
+    OrientHorizontal,
+    OrientVertical,
+    TurnLeft,
+    TurnRight,
+)
+from tbp.monty.frameworks.agents import AgentID
+from tbp.monty.frameworks.experiments.mode import ExperimentMode
+from tbp.monty.frameworks.experiments.monty_experiment import MontyExperiment
+from tbp.monty.frameworks.models.abstract_monty_classes import LearningModule
+from tbp.monty.frameworks.models.evidence_matching.learning_module import (
+    EvidenceGraphLM,
+)
+from tbp.monty.frameworks.models.goal_generation import (
+    EvidenceGoalGenerator,
+)
+from tbp.monty.frameworks.models.motor_policies import (
+    SurfacePolicyCurvatureInformed,
+)
+from tbp.monty.frameworks.models.motor_policy_selectors import SinglePolicySelector
+from tbp.monty.frameworks.models.motor_system import MotorSystem
+from tbp.monty.frameworks.models.motor_system_state import (
+    AgentState,
+    MotorSystemState,
+    ProprioceptiveState,
+)
+from tbp.monty.geometry import Rotation
+from tbp.monty.math import EulerAnglesXYZ, VectorXYZ
+from tests import HYDRA_ROOT
+
+
+class BasicPolicyTest(unittest.TestCase):
+    def setUp(self) -> None:
+        self.output_dir = tempfile.mkdtemp()
+
+        def hydra_config(test_name: str) -> DictConfig:
+            return hydra.compose(
+                config_name="experiment",
+                overrides=[
+                    f"experiment=test/policy/{test_name}",
+                    f"experiment.config.logging.output_dir={self.output_dir}",
+                ],
+            )
+
+        with hydra.initialize_config_dir(version_base=None, config_dir=str(HYDRA_ROOT)):
+            self.base_dist_cfg = hydra_config("base_dist_mujoco")
+            self.base_surf_cfg = hydra_config("base_surf_mujoco")
+            self.spiral_cfg = hydra_config("spiral_mujoco")
+            self.curve_informed_cfg = hydra_config("curve_informed_mujoco")
+            self.surf_hypo_driven_cfg = hydra_config("surf_hypo_driven_mujoco")
+            self.dist_hypo_driven_cfg = hydra_config("dist_hypo_driven_mujoco")
+            self.dist_hypo_driven_multi_lm_cfg = hydra_config(
+                "dist_hypo_driven_multi_lm_mujoco"
+            )
+
+    def tearDown(self) -> None:
+        shutil.rmtree(self.output_dir)
+
+    def test_can_run_informed_policy(self) -> None:
+        exp = hydra.utils.instantiate(self.base_dist_cfg.experiment)
+        with exp:
+            exp.run()
+
+    def test_can_run_spiral_policy(self) -> None:
+        exp = hydra.utils.instantiate(self.spiral_cfg.experiment)
+        with exp:
+            # TODO: test that no two locations are the same
+            exp.run()
+
+    def test_can_run_dist_agent_hypo_driven_policy(self) -> None:
+        exp = hydra.utils.instantiate(self.dist_hypo_driven_cfg.experiment)
+        with exp:
+            exp.run()
+
+    def test_can_run_surface_policy(self) -> None:
+        exp = hydra.utils.instantiate(self.base_surf_cfg.experiment)
+        with exp:
+            exp.run()
+
+    def test_can_run_curv_informed_policy(self) -> None:
+        exp = hydra.utils.instantiate(self.curve_informed_cfg.experiment)
+        with exp:
+            exp.run()
+
+    def test_can_run_surf_agent_hypo_driven_policy(self) -> None:
+        exp = hydra.utils.instantiate(self.surf_hypo_driven_cfg.experiment)
+        with exp:
+            exp.run()
+
+    def test_can_run_multi_lm_dist_agent_hypo_driven_policy(self) -> None:
+        exp = hydra.utils.instantiate(self.dist_hypo_driven_multi_lm_cfg.experiment)
+        with exp:
+            exp.run()
+
+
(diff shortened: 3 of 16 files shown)


jeremyshoemaker:
I cleaned up most of the typing issues, but there are some that will require changes elsewhere in the codebase that feel out of scope.

src/tbp/monty/cmp.py
```diff
@@ -44,13 +45,14 @@ class Message:
...
-        morphological_features,
-        non_morphological_features,
-        confidence,
-        use_state,
-        sender_id,
-        sender_type,
+        # TODO: Consider changing to VectorXYZ
+        location: npt.NDArray[np.float64],
+        morphological_features: dict[str, Any],
+        non_morphological_features: dict[str, Any],
```
jeremyshoemaker, line 51: Several places here, I ran into needing to type these values. They have more structure than this type suggests, but we'd need to use a TypedDict or dataclass to clear it up more. I'd vote for the latter. Definitely out of scope.

tests/mujoco/integration/frameworks/models/policy_test.py
```diff
@@ -0,0 +1,1115 @@
...
+        policy_selector = SinglePolicySelector(policy)
+        motor_system = MotorSystem(policy_selector)
+        policy.max_pc_bias_steps = 2
+        policy.reset(motor_system)
+
+        rng = np.random.RandomState(123)
+        ctx = RuntimeContext(rng)
+
+        # Initialize motor-system state
+        proprioceptive_state = MotorSystemState(
```
jeremyshoemaker: This was originally `ProprioceptiveState` but the `tangential_direction` method below expects a `MotorSystemState`. The underlying type seems to be the same, so I'm not sure what the difference between these two types is.

tests/mujoco/integration/frameworks/models/policy_test.py
```diff
@@ -0,0 +1,1115 @@
...
+        )
+        assert policy.continuous_pc_steps == 0, "Should have reset continuous counter"
+        assert policy.following_pc_counter == 2, (
+            "Should have not changed following_pc_counter"
+        )
+        assert policy.using_pc_guide is False, "Should not be using PC guide"
+        assert policy.prev_angle is None, "Should have reset prev_angle"
+        assert policy.pc_is_z_defined is False, "Should have reset z-defined flag"
+
+    def core_evaluate_compute_goal_for_target_loc(
```
jeremyshoemaker, line 906: The return type for this method is complicated. It's pulling values from a few places, including e.g. the aforementioned `morphological_features` that needs a clearer type. I left it as-is for now, but don't want to silence it.

jeremyshoemaker:
On the z clipping issue. I just went to go see what Habitat actually sets, and it seems to also use `0.01`, so there must be some differences in the two renderers. It might be that Habitat renders any polygons that are visible completely, where the MuJoCo one definitely clips the surface at the near plane.

Here's an example, the first one is MuJoCo and you can see the clipping in the top left, and the second one is Habitat.

image
image

These are both from the view finder. The patch views were clipping enough that only the bottom right of the surface is visible, thus making Monty think it was off object.

jeremyshoemaker:
I've been thinking about this z-clipping issue and whether changing the near plane is a good solution. One of the issues that can arise when setting a nearer clipping plane is a loss of depth precision due to the ratio of `far / near` increasing. I don't think that will be an issue here, because MuJoCo sets the far plane at 50.0, while Habitat sets it at 1000.0.

The real issue I see is that this solution only kicks the can down the road. The only reason this came up is because of a specific test case where the camera gets too close to the edge of the cube. If we got closer, we'd eventually end up in the same situation. It could also happen if the angle of the plane were closer to parallel with the camera's Z axis.

I'm wondering if we instead need a different solution for this problem. If the camera was on-object, and we MoveForward toward it, and end up off-object, I think we should do something to try to correct the situation. Undo the move, or move back some distance until we're back on-object. I think we're guaranteed to have that happen since we started from a position where we were on-object.

tristanls-tbp:
> I think we're guaranteed to have that happen since we started from a position where we were on-object.

Perhaps not relevant today, but in general, this is false if the object is moving independently of the agent.

jeremyshoemaker:
> Perhaps not relevant today, but in general, this is false if the object is moving independently of the agent.

That's true, but that's a much more complicated problem anyway.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -195,6 +195,10 @@ def _configure_spec_settings(self) -> None:
         g = self.spec.visual.global_
         g.offheight = render_resolution.height
         g.offwidth = render_resolution.width
+        # Set the near Z clipping plane a little closer to the camera so we don't
+        # clip objects when an agent is very close to an object, e.g. with the
+        # SurfaceAgent.
+        self.spec.visual.map.znear = 0.001
```
tristanls-tbp: **issue**: As far as I understand, in Habitat, the `0.01` means `0.01 meters` and [is not a multiplier of anything](https://github.com/facebookresearch/habitat-sim/blob/57ee4941dc4765240f0f91f70b2c97a919bf9038/src/esp/sensor/CameraSensor.cpp#L51). In MuJoCo, `znear` is a multiplier of `extent`:

Image

To match `0.01 meters`, maybe you could specify `extent = 1.0` to disable dynamic scaling, then set `znear = 0.01`?

tristanls-tbp (commented):
Submitting a review in progress. Will return to it later on.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I'm not sure how I missed that, and [it took some digging to find that the default value](https://github.com/google-deepmind/mujoco/blob/57008291227bbed526cd95844effca3792971ab3/src/engine/engine_init.c#L114) for `extent` is 2.0 which is really not what we want.

Setting it to 1.0 fixes the issue I was seeing.

jeremyshoemaker: While that _does_ fix the clipping issue, I'm wondering if it might negatively affect depth values. I had to loosen the tolerance on a depth check in a test to get it to pass again.

tristanls-tbp: **suggestion**: _NOTE: I'm just pattern matching, my mental model for this is poor_. So... [this reddit thread talks about "z-fighting"](https://www.reddit.com/r/gamedev/comments/puiqxa/what_purpose_does_a_nearclip_plane_serve_exactly/)... which makes me think that `znear`, `zfar`, and `extent` are all related. So perhaps with the `znear` matching Habitat now that `extent = 1.0`, perhaps `zfar` needs to be adjusted to match Habitat ([`1000` I think?](https://github.com/facebookresearch/habitat-sim/blob/57ee4941dc4765240f0f91f70b2c97a919bf9038/src/esp/gfx/PbrIBLHelper.cpp#L41))

jeremyshoemaker: I should have mentioned in my previous comment that I tried setting the `zfar` to `1000.0` to see if it would help with the loss of precision in the test, but that didn't fix it. I don't have a similar test for Habitat to see if it it more accurate.

I don't think we have to worry about z-fighting yet, since our scenes are close by and don't involve a lot of overlapping objects.

jeremyshoemaker: I'm not even sure the test change is a "loss of precision" of the type that would cause issues with z-depth.

src/tbp/monty/cmp.py
```diff
@@ -44,13 +45,14 @@ class Message:
...
         self,
-        location,
-        morphological_features,
-        non_morphological_features,
-        confidence,
-        use_state,
-        sender_id,
-        sender_type,
+        # TODO: Consider changing to VectorXYZ
+        location: npt.NDArray[np.float64],
```
tristanls-tbp: **issue**: `location` can be `None`. https://github.com/thousandbrainsproject/tbp.monty/blob/00e779edf69b6f4f9e9bd02204f6f8741b5eb344/src/tbp/monty/frameworks/models/evidence_matching/learning_module.py#L520

tristanls-tbp: **suggestion**: Update `Goal.location` to match `npt.NDArray[np.float64]` instead of `np.ndarray`.

tests/mujoco/integration/frameworks/models/policy_test.py
```diff
@@ -0,0 +1,1115 @@
...
+            [[0, 1, 0], [0, 0, 1], [1, 0, 0]]
+        )
+
+        self.fake_percept_advanced_pc = [
+            Message(**default_percept_args),
+            Message(**fp_1_backtrack_pc),
+            Message(**fp_2_corrupt_z),
+        ]
+
+    def initialize_lm_with_gsg(self) -> tuple[EvidenceGraphLM, dict[str, int | float]]:
```
tristanls-tbp, line 218: **note**: I'm aware this is a copy-and-paste. I'm just noticing that this method is being used exactly once, and out of all the `gsg_args` that are being returned, the only use is:
```python
surface_displacement = gsg_args["desired_object_distance"] * 1.5
```

tests/mujoco/integration/frameworks/models/policy_test.py
```diff
@@ -0,0 +1,1115 @@
...
+                if action_sequence:
+                    actions = action_sequence[-1][0]
+                    if actions:
+                        last_action = actions[0]
+
+                if step == 3:
+                    stored_action = last_action
+                    assert not exp.model.learning_modules[
+                        0
+                    ].buffer.get_last_obs_processed(), "Should be off object"
```
tristanls-tbp, line 343: **issue (non-blocking)**: Type error (LearningModule doesn't have `buffer`), but we are reaching deep into assumptions here. Best left for future us.

tests/mujoco/integration/frameworks/models/policy_test.py
```diff
@@ -0,0 +1,1115 @@
...
+
+                if step == 4:
+                    should_have_moved_back = (
+                        "Should have moved back by reversing last movement"
+                    )
+                    self.assertIsInstance(
+                        last_action, type(stored_action), should_have_moved_back
+                    )
+                    if isinstance(stored_action, (LookDown, LookUp)):
```
tristanls-tbp, line 352: **issue**: Lots of type errors in this if/elif chain that might actually be fixable and appropriate for the test context. Below is a suggestion, the pattern of which can be applied down the if/elif chain.

**suggestion*::
```suggestion
                    if isinstance(stored_action, (LookDown, LookUp)):
                        assert isinstance(last_action, (LookDown, LookUp))
```

tests/mujoco/integration/frameworks/models/policy_test.py
```diff
@@ -0,0 +1,1115 @@
...
+                            should_have_moved_back,
+                        )
+                    assert exp.model.learning_modules[
+                        0
+                    ].buffer.get_last_obs_processed(), "Should be back on object"
+                    break  # Don't go into exploratory mode
+
+                step += 1
+
+    def test_surface_policy_moves_back_to_object(self) -> None:
```
tristanls-tbp, line 431: **praise**: Amazing this passed. Good job recreating the conditions so precisely.

tests/mujoco/integration/frameworks/models/policy_test.py
```diff
@@ -0,0 +1,1115 @@
...
+                step += 1
+
+            # Most recently observed surface normal sent to the learning module
+            current_pose = exp.model.learning_modules[0].buffer.get_current_pose(
+                input_channel="first"
+            )
+
+            # Rotate vector representing agent's pointing direction by the agent's
+            # current orientation
+            agent_rot = Rotation.from_quat(
```
tristanls-tbp, line 618: **praise**: Thanks for using `Rotation`.

tests/mujoco/integration/frameworks/models/policy_test.py
tristanls-tbp: Thank you for catching and correcting the discrepancy.

`MotorSystemState` was meant to hold extra properties like `motor_only_step: bool`, but that never worked out. We should probably refactor things at some point, but for now I'm assuming that refactor is out of scope, hence the ask below.

**issue**: Please rename `proprioceptive_state` variable to `motor_system_state` here for coherence. I'm assuming that updating `tangential_direction(...)` signature is out of scope, hence I'm not suggesting going down that rabbit hole.

tests/mujoco/integration/frameworks/models/policy_test.py
```diff
@@ -0,0 +1,1115 @@
...
+        # into taking PC steps when testing this
+        initial_min_general_steps = 1
+        policy.min_general_steps = initial_min_general_steps
+        policy.reset(motor_system)
+
+        rng = np.random.RandomState(123)
+        ctx = RuntimeContext(rng)
+
+        # Initialize motor system state
+        proprioceptive_state = ProprioceptiveState(
```
tristanls-tbp: **issue**: This should be `MotorSystemState` for the same reason as above. Also, please rename `proprioceptive_state` to `motor_system_state` for the same reason.

tests/mujoco/integration/frameworks/models/policy_test.py
tristanls-tbp: **issue**: I think it's this
```python
-> tuple[npt.NDArray[np.float64], npt.NDArray[np.float64], VectorXYZ, npt.NDArray[np.float64]]
```

It looks complicated, but all four values are vectors. The `morphological_features` bit extracts a surface normal vector.

**note**: Technically, it's this
```python
-> tuple[npt.NDArray[np.float64] | None, npt.NDArray[np.float64], VectorXYZ, npt.NDArray[np.float64]]
```
because `Goal.location` can be `None` (first in tuple). But we can cheat and know it's not `None` here because `_compute_goal_for_target_loc` won't return a `Goal` with `None` location.

tests/unit/simulators/mujoco/noop_agent_test.py
```diff
@@ -88,7 +88,7 @@ def test_agent_observation_single_sensor(self) -> None:
             assert depth.shape == (64, 64)
             assert rgba.shape == (64, 64, 4)
             # assert that the min depth is the near surface of the cube
-            assert np.allclose(depth.min(), 4.0)
+            assert np.allclose(depth.min(), 4.0, atol=1e-5)
```
tristanls-tbp: **todo**: Could you share the failure case without `atol` so we can see the exact problem?

tristanls-tbp (commented):
I reviewed the whole PR.

tests/mujoco/integration/frameworks/models/policy_test.py
jeremyshoemaker: Yeah, I've seen that a lot, where we depend on things that `LearningModule` doesn't define but subclasses do. I've left those as a later problem.

tests/mujoco/integration/frameworks/models/policy_test.py
jeremyshoemaker: I'll take a look at it. I think there are a lot of places where the type checker should be able to infer narrowed types and doesn't, which is annoying to say the least. This might not be one of them though given that you're suggesting asserting on `last_action` and not `stored_action`.

tests/mujoco/integration/frameworks/models/policy_test.py
jeremyshoemaker: It was originally using Habitat's utilities so I had to replace it with _something_.

tests/unit/simulators/mujoco/noop_agent_test.py
jeremyshoemaker: So the difference is in `depth.min()`. 
- In the new version with `extent` set to `1.0` the value is `4.000041`
- In the version with `extent` at the default and `znear` set to `0.001` the value is `3.9999769`
- In the original version with no changes to `extent` or `znear` the value was `4.000002`. This is the version I wrote the test with originally.

The absolute difference between these values and the test value of `4.0` are:
- `0.000041`
- `0.0000231`
- `0.000002`

The last two are close enough with the default `atol` of `1e-8`, but the first one isn't. Maybe I'm not setting the right values for `atol` and `rtol` (which I'm leaving at the default).

With the `extent` set to `1.0` and setting `zfar` to `1000.0` to try to match what Habitat uses, we get `4.000044` which is roughly the same degree of accuracy we get with the default `zfar` of `50.0`.

---

This continued to bother me while I was writing this comment, mostly because I'm wondering if the loss of precision is causing issues in the test I'm currently working on, and so I decided to test it with the default `extent` of `2.0` and set the `znear` and `zfar` to values that would end up matching what Habitat uses, i.e. `znear` set to `0.005` and `zfar` set to `500.0`.

With those values, these policy tests pass, and the `NoopAgentTest` passes, as that `depth.min()` value becomes `3.9999924` which is a difference of `0.0000076`. That's much closer than even the original, so I feel like this leads to better precision values.

I need to check if it fixes the issues I'm seeing in the test I'm currently working on, but I think this is the correct solution here.

jeremyshoemaker: It doesn't fix the issues I'm seeing with the `evidence_lm_test` but I still think it's the correct solution.

tests/mujoco/integration/frameworks/models/policy_test.py
jeremyshoemaker: The returned `lm` value is being used in three places.

tristanls-tbp: **note**: I don't see how it matters that the returned value is used in three places when the method is _called once_. I can inline the method where `lm` is assigned and still use `lm` in three places. It feels like some anticipated reuse did not materialize.

jeremyshoemaker: Ah, I thought you were suggesting deleting it and just inlining the value for `surface_displacement`.

tests/unit/simulators/mujoco/noop_agent_test.py
tristanls-tbp: **note**: Oh, interesting. `extent: 2.0, znear: 0.005, zfar: 500.0` sounds ok to me 👍.

tests/mujoco/integration/frameworks/models/policy_test.py
tristanls-tbp: Sorry, I intend my labels to be interpreted per https://conventionalcomments.org/ ...

**note:** Notes are always non-blocking and simply highlight something the reader should take note of.
vs.
**suggestion**: Suggestions propose improvements to the current subject. It’s important to be explicit and clear on _what_ is being suggested and _why_ it is an improvement. Consider using patches and the _blocking_ or _non-blocking_ [decorations](https://conventionalcomments.org/#decorations) to further communicate your intent.

jeremyshoemaker: Yeah, I know. 😄

I don't think I wrote this code, but I might have written code this same way if it makes the test itself more concise. Making a method even if it's only used once is still naming a concept and that might be worth it.

tristanls-tbp: 6️⃣ 7️⃣
🫲 🫱

jeremyshoemaker:
@tristanls-tbp I think I addressed all your feedback, so this is ready for review again.

tests/mujoco/integration/frameworks/models/policy_test.py
```diff
@@ -0,0 +1,1128 @@
...
+        # As the agent faces "forward" along the negative z-axis, we use this vector
+        # to visualize its orientation
+        agent_direction = resulting_rot.apply(np.array([0, 0, -1]))
+
+        return (
+            motor_goal.location,
+            motor_goal.morphological_features["pose_vectors"][0],
+            target_loc,
+            agent_direction,
+        )
```
tristanls-tbp, line 1009: **issue**: Mypy type complaints.

**suggestion**: Amazingly, mypy stops complaining if we do

```suggestion
        assert motor_goal.location is not None
        return (
            motor_goal.location,
            motor_goal.morphological_features["pose_vectors"][0],
            target_loc,
            agent_direction,
        )
```

tristanls-tbp (commented):
Thank you, one more type-related chore, and I'm good to go.

tests/mujoco/integration/frameworks/models/policy_test.py
jeremyshoemaker: Ah, I missed that one too. PyCharm had a big block of yellow on the return value because to the type mismatch. Fixed.

tristanls-tbp (approved):
Ship it!

PR #1050, opened 2026-06-22
docs: Add and update SM & LM future work pages

vkakerbeck:
Add new pages for SM and LM related future work items. Also updates some existing text and items.


docs/future-work/learning-module-improvements/better-model-surface-representation.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Better Model Surface Representation
+description: Figure out and test a way to represent and objects surface better in our models or matching algorithm.
```
nielsleadholm:
```suggestion
description: Figure out and test a way to represent an object's surface better in our models and/or matching algorithm.
```

docs/future-work/learning-module-improvements/better-model-surface-representation.md
```diff
@@ -0,0 +1,22 @@
...
+
+![Example of problems with the naive approach of search. Sparsity in the model will lead to some search locations not having a nearby point stored in the model (red). If we increase the search radius, many points that should be off the object will be considered on object instead.](../../figures/future-work/search_naive.png)
+
+This problem is partially addressed by the custom distance metric we use in the EvidenceGraphLM, which takes into account the direction of the sensed surface normal when searching for nearest neighbors in the model (read more [here](../../how-monty-works/learning-module/evidence-based-learning-module.md#search-radius-and-distance-weighting)). However, it only works up to a certain degree of sparsity in the models. It also is only an approximation of surface representation (only during inference, using spherical radius without distinction between going into vs. out of the surface, basic heuristic to account for curvature). There might be small tweaks that could be made to it (account for curvature better, test parameters more) or a larger rethinking of representing surface in the models.
+
```
nielsleadholm, line 21: To help inspire further work, maybe we could add something like this? (+/- another drawing)

"For example, the existing method uses the surface curvature sensed at a point to adjust the search radius. However, in the absence of learning a dense model, there is no form of learning through repeated sensation that a surface might extend farther or less in different directions from a given point. This issue is apparent where there is asymmetry in the surface characteristics around a sensed point: in one direction, the curvature might continue smoothly for some time, while in the other, it might suddenly change (e.g. due to a sharp corner). An improved representation of surface should help us learn such surface characteristics, while avoiding an overly dense representation.

This might imply an inversion of how search operates - rather than the sensation determining the extent of the matching radius, learned characteristics of points and their surrounding surface could possibly dictate this."

docs/future-work/learning-module-improvements/better-model-surface-representation.md
```diff
@@ -0,0 +1,22 @@
...
+![Example of problems with the naive approach of search. Sparsity in the model will lead to some search locations not having a nearby point stored in the model (red). If we increase the search radius, many points that should be off the object will be considered on object instead.](../../figures/future-work/search_naive.png)
+
+This problem is partially addressed by the custom distance metric we use in the EvidenceGraphLM, which takes into account the direction of the sensed surface normal when searching for nearest neighbors in the model (read more [here](../../how-monty-works/learning-module/evidence-based-learning-module.md#search-radius-and-distance-weighting)). However, it only works up to a certain degree of sparsity in the models. It also is only an approximation of surface representation (only during inference, using spherical radius without distinction between going into vs. out of the surface, basic heuristic to account for curvature). There might be small tweaks that could be made to it (account for curvature better, test parameters more) or a larger rethinking of representing surface in the models.
```
nielsleadholm:
```suggestion
This problem is partially addressed by the custom distance metric we use in the EvidenceGraphLM, which takes into account the direction of the sensed surface normal, as well as the amount of curvature, when searching for nearest neighbors in the model (read more [here](../../how-monty-works/learning-module/evidence-based-learning-module.md#search-radius-and-distance-weighting)). However, it only works up to a certain degree of sparsity in the models. It is also only an approximation of surface representation (only during inference, using spherical radius without distinction between going into vs. out of the surface, basic heuristic to account for curvature). There might be small tweaks that could be made to it (account for curvature better, test parameters more) or a larger rethinking of representing surface in the models.
```

docs/future-work/learning-module-improvements/better-model-surface-representation.md
```diff
@@ -0,0 +1,22 @@
...
+
+![How we currently account for surface when searching the models. Depending on the amount of curvature sensed, the search radius will be narrowed more or less in the direction of the sensed surface normal.](../../figures/future-work/search_incl_surface_rep.png)
```
nielsleadholm: Is it worth briefly mentioning the heuristic we have around point normal directions around 180 to distinguish two sides of a thin surface? It might be useful context for anyone coming up with an alternative formulation of surfaces.

docs/future-work/learning-module-improvements/better-model-surface-representation.md
```diff
@@ -0,0 +1,22 @@
...
+output-type: experiments, analysis, PR
+skills: python, research, monty
+contributor: 
+status: open
+---
+
+> [!NOTE]
+> This item relates to [using models with fewer points](use-models-with-fewer-points.md), which would be one of the ways to measure success here.
+
+The naive way of testing whether a sensor is still on the object if a given hypothesis is correct is to apply the movement in the object's RF and searching in the radius of the new location whether there are points stored nearby in the model. However, this disregards the concept of surface, where we would want to have very little tolerance when the movement is going off the surface (i.e. in the direction of the surface normal) but allow for more tolerance if the next point along the surface is a bit further away. If we don't allow this second tolerance, we add a requirement to learn dense models where even a plain, even surface is represented with many points in the model because otherwise we won't find near enough neighbors.
```
nielsleadholm:
```suggestion
The naive way of testing whether a sensor is still on the object if a given hypothesis is correct is to apply the movement in the object's RF and, searching in the radius of the new location, determine whether there are points stored nearby in the model. However, this disregards the concept of surface, where we would want to have very little tolerance when the movement is going off the surface (i.e. in the direction of the surface normal) but allow for more tolerance if the next point along the surface is a bit further away. If we don't allow this second tolerance, we are forced to learn dense models where even a plain, even surface is represented with many points in the model because otherwise we won't find near enough neighbors.
```

docs/future-work/learning-module-improvements/deal-with-incomplete-models.md
```diff
@@ -2,10 +2,17 @@
...
+
+![Problem with the current algorithm and incomplete models. Assume we are trying to infer a cube.
+There is no mechanism to distinguish between the scenario on the left (learned a partial model of the cube and sampling a new point on it) and the one on the right (sampling a point that is not on the cube). (The one in the middle relates to [surface representations to allow for sparser models](better-model-surface-representation.md).) In the case on the left, we would want to keep the cube hypothesis and extend the model for the cube with new points. In the case on the right, we would want to say that we are not on a cube anymore. Currently, this is what happens in both cases.](../../figures/future-work/incomplete_models.png)
+
+It's unclear how this would best be solved. A possibility would be to use heuristics of how an object's surface might continue based on the already stored surface models. One could also store points in the model that represent the end of an object (although this seems to already be encoded in the existing surface normals).
```
nielsleadholm: Maybe we could also add something like the below?

"Another alternative might be leveraging gradual, unsupervised learning in the context of an ever-changing environment. For example, if moving to a new location results in a sensed feature that was not predicted, this results in a prediction error (suggesting we have moved off of the object). However, the object's reference frame might slightly increment a weight that captures whether the feature there is actually part of the reference frame. After only a couple exposures, this weight would be insufficient to result in the model predicting that feature - i.e. it is still not considered part of the object model. However, given sufficient, repeated exposures, this weight would become strong enough that the feature *is* now a permanent part of the model, and it is predicted at that location. Given the gradual nature of this weight increase, spurious associations of objects (e.g. two objects that happen to be next to one another at a point in time) would not be learned.

Such an approach relates to the unsupervised learning mechanism we have implemented in the constrained object models (link), as well as the idea of a "permanence" associated with a binary synaptic connection in biological neurons (link to Numenta papers). This could also be complemented with different learning modules having different learning rates (link) for their feature-location binding, such that some (including the equivalent of the hippocampus) have a lower threshold for expanding their existing models. Ideally, this would move Monty even further in the direction of not having a separation between learning and inference."

docs/future-work/learning-module-improvements/deal-with-incomplete-models.md
```diff
@@ -2,10 +2,17 @@
...
 improved-metric: learning
 output-type: prototype, monty-feature, PR
 skills: python, research, monty
 contributor: 
 status: open
 ---
+
+As described in [this document](https://www.overleaf.com/read/qxchttxzpfnd#ea9111) (section 3.3.2), Monty's object models have no concept of whether the absence of a point in the model means that the object doesn't exist there or the sensor hasn't explored this area yet.
+
+![Problem with the current algorithm and incomplete models. Assume we are trying to infer a cube.
```
nielsleadholm: When I look at the preview, I'm not seeing any of this text. Sorry if this is a known issue with the preview (but not with the actual build once merged) that I've forgotten about. 

Image

docs/future-work/learning-module-improvements/learning-and-forgetting-speed-parameter.md
```diff
@@ -0,0 +1,15 @@
...
+rfc: required
+estimated-scope: unknown
+improved-metric: learning, compositional
+output-type: RFC, prototype, monty-feature, PR
+skills: python, research, monty
+contributor: 
+status: open
+---
+
+At the moment, all our learning modules learn at the same speed (instantaneously integrating information into their models). However, as Monty becomes more hierarchical, it will likely be necessary to have adjustable learning speeds where some LMs learn slower and others faster. For instance, we want to be able to quickly learn a representation of the current environment to be able to interact with it (and also quickly forget it again). This is similar to what happens in the Hippocampus and is what Monty can currently do (without the forgetting part). However, we will also need more stable models that don't change quickly and are remembered for a long time. This is important since higher-level LMs rely on the output of lower-level LMs. If that output is changing faster than the higher level models can adapt, they will fail to learn useful compositional models.
```
nielsleadholm:
```suggestion
At the moment, all our learning modules learn at the same speed (instantaneously integrating information into their models). However, as Monty becomes more hierarchical, it will likely be necessary to have adjustable learning speeds where some LMs learn slower and others faster. For instance, we want to be able to quickly learn a representation of the current environment to be able to interact with it (and also quickly forget it again). This is similar to what happens in the hippocampus and is what Monty can currently do (without the forgetting part). However, we will also need more stable models that don't change quickly and are remembered for a long time. This is important since higher-level LMs rely on the output of lower-level LMs. If that output is changing faster than the higher level models can adapt, they will fail to learn useful compositional models.
```

docs/future-work/learning-module-improvements/model-repeating-patterns.md
```diff
@@ -0,0 +1,19 @@
...
+rfc: required
+estimated-scope: unknown
+improved-metric: learning, numsteps, features-and-morphology
+output-type: RFC, experiments, analysis, PR
+skills: python, research, monty
+contributor: 
+status: open
+---
+
+Repeating patterns are somewhere between structured objects that require a reference frame (learned in LMs) and primitive features that can be extracted by sensor modules directly. Sensor modules can extract features that go beyond simple color values, such as textures (like explored in the LBP work mentioned [here](../sensor-module-improvements/extract-better-features.md)). However, they don't learn structured models like LMs do and would therefor struggle with more structured patterns.
```
nielsleadholm: A note on the use of therefore (TLDR: therefore is generally what you need, not therefor): https://www.grammarly.com/blog/commonly-confused-words/therefore-vs-therefor/

```suggestion
Repeating patterns are somewhere between structured objects that require a reference frame (learned in LMs) and primitive features that can be extracted by sensor modules directly. Sensor modules can extract features that go beyond simple color values, such as textures (explored in the LBP work mentioned [here](../sensor-module-improvements/extract-better-features.md)). However, they don't learn structured models like LMs do and would therefore struggle with more structured patterns. 
```

docs/future-work/learning-module-improvements/model-repeating-patterns.md
```diff
@@ -0,0 +1,19 @@
...
+
+Think of patterns on bed sheets or on a carpet, where you have an object or structured pattern that repeats many times. At the moment, we would need to learn or infer the pattern over and over again for each location where it applies. We could have a lower-level LM recognize the repeating aspect of the pattern and assign it to the different locations on the object where it repeats. However, the lower-level LM would have no concept of the pattern repeating and would have to infer it every time again (unless the higher LM helps with top-down connections). 
+
+A better representation could be that the models RF wraps around itself. So when incoming movement information would take the LM outside of the model's RF, it instead integrates the movement in such a way that it predicts to come out on the other side of the pattern again. Similar to how ins some video games you can move off on the left side of the screen and reappear on the right side again. This would implement a toroidal space where you never move out of the model's RF.
```
nielsleadholm:
```suggestion
A better representation could be that the models RF wraps around itself. When incoming movement information would take the LM outside of the model's RF, it instead integrates the movement in such a way that it predicts coming out on the other side of the pattern again. This is similar to how in some video games you can move off on the left side of the screen and reappear on the right side again. This would implement a toroidal space where you never move out of the model's RF, related to how grid cells represent space.
```

docs/future-work/learning-module-improvements/unsupervised-learning-in-a-hierarchy.md
```diff
@@ -0,0 +1,19 @@
...
+improved-metric: learning, compositional
+output-type: experiments, analysis, PR
+skills: python, research, monty
+contributor: 
+status: open
+---
+
+Currently, training a hierarchical setup of Monty involves a complicated procedure of supervising the lower-level LMs to learn the child objects, followed by supervising the higher-level LM to learn the parent object. This is complex and unrealistic to scale. It also requires labeled datasets where child objects can be learned in isolation first.
+
+A first step for this item would be to test what kind of models are learned if there is no supervision. The `max_model_size` parameter of the [constrained object models](../../how-monty-works/learning-module/object-models.md#object-models) can be leveraged to encourage the lower level LMs to learn smaller sub-components.
```
nielsleadholm, line 15: Maybe add something like, "In addition, models which learn and forget at different rates (link) could gradually disentangle objects from one another by only storing information that consistently appears together (i.e. the features of a child object at their relative arrangements)."

docs/future-work/sensor-module-improvements/one-d-sensor-module.md
```diff
@@ -0,0 +1,20 @@
...
+
+We want to extend this concept to even more extreme distortions by projecting movement into 1D space. One concrete example we imagine is recognizing cables in various configurations. The movement in this case should be mapped onto one dimension: moving up or moving down the cable. It can be estimated using local features (curvature of the cable). In addition to transforming the incoming movement, Monty will also need to leverage a specific policy that lets it's sensor follow this principal curvature direction. Just like with the 2D SM, we want to move smoothly along the surface, we need to move smoothly along the cable instead of taking random jumps around it for this to work.
```
nielsleadholm:
```suggestion
We want to extend this concept to even more extreme distortions by projecting movement into 1D space. One concrete example we imagine is recognizing cables or strings in various configurations. The movement in this case should be mapped onto one dimension: moving up or moving down the cable. It can be estimated using local features (curvature of the cable). In addition to transforming the incoming movement, Monty will also need to leverage a specific policy that lets its sensor follow this principal curvature direction. Just like with the 2D SM, we need to move smoothly along the cable instead of taking random jumps around it for this to work - note this is no different from a human, who would not be able to locate their position on a tangled cable without slowly moving a sensor (eye or finger) along it. 
```

docs/future-work/learning-module-improvements/hybrid-node-and-edge-matching.md
```diff
@@ -0,0 +1,21 @@
...
+rfc: optional
+estimated-scope: large
+improved-metric: numsteps, pose, scale
+output-type: experiments, analysis, PR
+skills: python, research, monty
+contributor: 
+status: open
+---
+
+When developing Monty, we explored several different learning module approaches. There were three major LM versions we implemented (you can still run them all in Monty). The first LM implementation (`DisplacementGraphLM`) used edges and movements to recognize objects, while the later two used nodes and locations. The difference between the two node-based LMs was that the first version (`FeatureGraphLM`) would delete a hypothesis after one inconsistent observation, while the second version (`EvidenceGraphLM`) would associate a continuous evidence score with each hypothesis which made it more robust to noise. Between those two, the evidence-based approach was clearly favored. However, when moving from edge- to node-based matching we lost some desirable properties, such as very efficient pose recognition as well as scale recognition. For a full comparison, see the table below or read more [here](../../how-monty-works/learning-module.md).
```
nielsleadholm:
```suggestion
When developing Monty, we explored several different learning module approaches. There were three major LM versions we implemented (you can still run them all in Monty). The first LM implementation (`DisplacementGraphLM`) used edges and movements to recognize objects, while the latter two used nodes and locations. The difference between the two node-based LMs was that the first version (`FeatureGraphLM`) would delete a hypothesis after one inconsistent observation, while the second version (`EvidenceGraphLM`) would associate a continuous evidence score with each hypothesis which made it more robust to noise. Between those two, the evidence-based approach was clearly favored. However, when moving from edge- to node-based matching we lost some desirable properties, such as very efficient pose recognition as well as scale recognition. For a full comparison, see the table below or read more [here](../../how-monty-works/learning-module.md).
```

docs/future-work/learning-module-improvements/learning-and-forgetting-speed-parameter.md
```diff
@@ -0,0 +1,15 @@
...
+
+The idea here is to leverage the existing [constrained object models](../../how-monty-works/learning-module/object-models.md#object-models) and add a learning speed parameter into it. The concrete way of how this parameter could work would need to be outlined in an [RFC](../../contributing/request-for-comments-rfc.md). Ideally, it should also incorporate forgetting.
```
nielsleadholm:
```suggestion
The idea here is to leverage the existing [constrained object models](../../how-monty-works/learning-module/object-models.md#object-models) and add a learning speed parameter into it. The concrete way in which this parameter could work would need to be outlined in an [RFC](../../contributing/request-for-comments-rfc.md). Ideally, it should also incorporate forgetting.
```

docs/future-work/learning-module-improvements/better-model-surface-representation.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Better Model Surface Representation
+description: Figure out and test a way to represent and objects surface better in our models or matching algorithm.
+rfc: optional
+estimated-scope: unknown
+improved-metric: speed, pose, accuracy, noise
```
nielsleadholm, line 6: Should we add an improved metric for `model-sparsity`? This could fit with the fact that the Use fewer model points Future Work item refers to "Update learning algorithm, sensor module, and policy to learn sparser models." - i.e. this is not something that we can achieve through just a single change.

docs/future-work/learning-module-improvements/better-model-surface-representation.md
```diff
@@ -0,0 +1,22 @@
...
+output-type: experiments, analysis, PR
```
nielsleadholm, line 7: Maybe add publication? I think a genuine improvement here (and several of the other Future Work items you updated in this PR) would be a significant step if done, and worthy of making into a paper.

nielsleadholm (commented):
Nice updates! Just some misc. suggestions.

docs/future-work/learning-module-improvements/better-model-surface-representation.md
vkakerbeck: Is there any advantage of model sparsity that speed doesn't already capture?

docs/future-work/learning-module-improvements/better-model-surface-representation.md
vkakerbeck: I'm not sure this one would be. I think it might end up being a relatively small change. Maybe we add that once it is scoped more? But its a good point, I will go through all the items again and check if publication fits on some of them as well.

docs/future-work/learning-module-improvements/better-model-surface-representation.md
vkakerbeck: I'm not sure I follow this. Wouldn't we store sudden changes in curvature in the model already and then just have to use our algorithm to make use of this assumption? What would we additionally store in the model about the surface?

docs/future-work/learning-module-improvements/deal-with-incomplete-models.md
vkakerbeck: Oh good catch! That might be because it can't deal with the link inside the caption (since the other captions render correctly in the preview). I removed that and will see if it fixes it.

docs/future-work/learning-module-improvements/better-model-surface-representation.md
```diff
@@ -13,10 +13,10 @@ status: open
...
 
-![How we currently account for surface when searching the models. Depending on the amount of curvature sensed, the search radius will be narrowed more or less in the direction of the sensed surface normal.](../../figures/future-work/search_incl_surface_rep.png)
\ No newline at end of file
+![How we currently account for surface when searching the models. Depending on the amount of curvature sensed, the search radius will be narrowed more or less in the direction of the sensed surface normal. Points in the radius that have a surface normal that points in the opposite direction of the hypothesized one with be ignored in the evidence calculation to account for thin surface where points on both side of a surface will end up in the radius.](../../figures/future-work/search_incl_surface_rep.png)
```
nielsleadholm: A couple suggestions / typos but for some reason I can't do an inline suggestion:

--> Points in the radius with a surface normal pointing in the opposite direction of the hypothesized normal will be ignored in the evidence calculation. This accounts for thin surfaces where points on both side of a surface will end up in the radius.

docs/future-work/learning-module-improvements/better-model-surface-representation.md
nielsleadholm: I guess my point is that the concept of a surface is really being captured by the single point we are sensing right now, and heuristics based on the curvature sensed there. The learned points themselves are not doing anything to capture a notion of an extending surface. I'm not sure exactly what the solution would look like, but it feels like that kind of inversion could be useful. Let me know if you want to Huddle, might be easier to describe.

docs/future-work/learning-module-improvements/better-model-surface-representation.md
nielsleadholm: Yeah that's fair.

docs/future-work/learning-module-improvements/deal-with-incomplete-models.md
nielsleadholm: Just fyi I think it might still be hidden?

vkakerbeck: yeah I just puched another fix. I think it was because there was a newline in the caption. But I decided to just move this long block out of the caption, that way we can also keep the link

docs/future-work/learning-module-improvements/better-model-surface-representation.md
vkakerbeck: Ah okay, I see :) I was never able to figure out a way to make this work (like how can we adjust the search radius based on points that we first have to search for) but I think there actually is a way. I pushed a simplified version of your paragraphs (I am still not sure I understand the middle part). Happy to huddle if you want me to understand and add those parts too

vkakerbeck:
Thanks for the thorough review. I think I made all the changes & also confirmed that the caption now renders 👍

docs/future-work/learning-module-improvements/better-model-surface-representation.md
nielsleadholm: Ok nice, yeah I think that's captured it well. And yeah I'm also not sure how it would work, but it feels like in biology (and maybe with the appropriate hardware?) it is possible, because it's basically just looking at the intersection of two spaces. Thanks for the update!

docs/future-work/learning-module-improvements/deal-with-incomplete-models.md
```diff
@@ -2,10 +2,21 @@
...
 status: open
 ---
+
+As described in [this document](https://www.overleaf.com/read/qxchttxzpfnd#ea9111) (section 3.3.2), Monty's object models have no concept of whether the absence of a point in the model means that the object doesn't exist there or the sensor hasn't explored this area yet.
+
+![Inability to distinguish between an incomplete model and actually reaching the end of an object.](../../figures/future-work/incomplete_models.png)
+The figure above illustrated the problem with the current algorithm and incomplete models. Assume we are trying to infer a cube. There is no mechanism to distinguish between the scenario on the left (learned a partial model of the cube and sampling a new point on it) and the one on the right (sampling a point that is not on the cube). (The one in the middle relates to [surface representations to allow for sparser models](../learning-module-improvements/better-model-surface-representation.md).) In the case on the left, we would want to keep the cube hypothesis and extend the model for the cube with new points. In the case on the right, we would want to say that we are not on a cube anymore. Currently, this is what happens in both cases.
```
nielsleadholm:
```suggestion
The figure above illustrates the problem with the current algorithm and incomplete models. Assume we are trying to infer a cube. There is no mechanism to distinguish between the scenario on the left (learned a partial model of the cube and sampling a new point on it) and the one on the right (sampling a point that is not on the cube). (The one in the middle relates to [surface representations to allow for sparser models](../learning-module-improvements/better-model-surface-representation.md).) In the case on the left, we would want to keep the cube hypothesis and extend the model for the cube with new points. In the case on the right, we would want to say that we are not on a cube anymore. Currently, this is what happens in both cases.
```

nielsleadholm (approved):
Looks great!

PR #1052, opened 2026-06-22
test: create MuJoCo version of sensor_module_test.py

jeremyshoemaker:

The only real change here outside of which configs the tests load is replacing the `while` loop with a simpler `for` loop in `test_can_set_up`.

That said, I'm not sure that test is doing anything useful, and the same goes for `test_feature_change_sm`. In the latter case, it's an unfinished test. These both take some non-zero amount of time to run and don't really test any results of the experiments they run.


tristanls-tbp (approved):
👍

PR #1053, opened 2026-06-22
chore: update the readme docs script...

brothman-tbp:
Explicitly set slug for doc creation, and to normalize titles into slugs for categories to match the way readme automatically creates slugs, to avoid duplicate discrepancy errors we've been getting that have been changing our URLs unexpectedly.

As readme slugs are case sensitive, also change the 'CMP' file to 'cmp' to match the readme_slug.


tools/github_readme_sync/hierarchy.py
vkakerbeck, line 1: Why this new line?

tools/github_readme_sync/readme.py
```diff
@@ -571,4 +594,4 @@ def replace_match(match):
 class DocumentNotFound(RuntimeError):
     """Raised when a document is not found."""
 
-    pass
+    pass
```
vkakerbeck: same here

tools/github_readme_sync/upload.py
```diff
@@ -131,4 +131,4 @@ def load_doc(file_path: str, category_slug: str, child: dict):
 
     with file_path.open(encoding="utf-8") as file:
         body = file.read()
-        return process_markdown(body, child["slug"])
+        return process_markdown(body, child["slug"])
```
vkakerbeck: and here

tools/github_readme_sync/readme.py
```diff
@@ -293,6 +315,7 @@ def create_or_update_doc(
             "hidden": doc.get("hidden", False),
             "order": order,
             "parentDoc": parent_id,
+            "slug": doc["slug"],
```
vkakerbeck, line 318: What's the reason that this isn't also readme-slug? Is this the simplification you found where we can just send out slug to readme?

vkakerbeck (commented):
There are a couple on newlines at then end of files that seem unrelated to the PR. Can we remove those?

tools/github_readme_sync/readme.py
brothman-tbp: Yep, this is the simplification where we can send our own slug to ReadMe, so then the ReadMe slug will always be our own slug (only for docs, not categories)

tools/github_readme_sync/hierarchy.py
brothman-tbp: It seems like the issue was actually not having a new line at the end, added a new line to the end of those files and it fixed the problem.

tools/github_readme_sync/readme.py
brothman-tbp: It seems like the issue was actually not having a new line at the end, added a new line to the end of those files and it fixed the problem.

tools/github_readme_sync/upload.py
brothman-tbp: It seems like the issue was actually not having a new line at the end, added a new line to the end of those files and it fixed the problem.

vkakerbeck (approved):
LGTM

PR #1054, opened 2026-06-23
refactor!: deprecate the live plotter and point to `tbp.teleop`

ramyamounir:
This PR deprecates and removes the existing live plotter mechanism that uses `show_sensor_output`. It is now being replaced by `tbp.teleop`. I added documentation for the teleop repo.

This is a **breaking change** because the `show_sensor_output` will not be read by the experiment if used in any user custom configuration.


tristanls-tbp:
I added @jeremyshoemaker @nielsleadholm and @vkakerbeck to not review the details but to weigh in on the overall Monty-as-a-product question of removing the live plotter from Monty at this time.

@jeremyshoemaker pointed out that `tbp.teleop` does not support the upcoming `uv` environment that he's working in, so he's still relying on LivePlotter.

jeremyshoemaker:
I _can_ work without it, as I can pull and render image data in the PyCharm debugger, but I've been using the live plotter (when it works) to run whole experiments, instead of stepping through one slowly, to compare what MuJoCo is doing vs what Habitat is doing.

To be clear, it's quite likely that getting `tbp.teleop` to work with the `uv` version is relatively painless. I don't expect any problems there, I just haven't had a chance to look into it yet.

tristanls-tbp:
I should add that I think we should promote `tbp.teleop`, but leave the live plotter in place for now, at least until we emit equivalent telemetry so that something else can visualize read-only live views.

vkakerbeck:
I don't have a strong opinion. I think right now the show_sensor_output flag is still the most convenient way to quickly look at what Monty is doing (not needing to find and set up a separate repo etc.) If it isn't hindering any of our current platform work, I would tend towards leaving it in for now. We could add a note to it somewhere to make it more explicit that this feature isn't actively maintained/tested and is planned to be deprecated.

nielsleadholm:
> I don't have a strong opinion. I think right now the show_sensor_output flag is still the most convenient way to quickly look at what Monty is doing (not needing to find and set up a separate repo etc.) If it isn't hindering any of our current platform work, I would tend towards leaving it in for now. We could add a note to it somewhere to make it more explicit that this feature isn't actively maintained/tested and is planned to be deprecated.

Yeah this seems like a good approach given there is still some utility.

nielsleadholm:
Just a note that even if we are holding off on removing LivePlotter, we should still add the new sections to the tutorial that discuss teleop (`docs/how-to-use-monty/logging-and-analysis.md`).

ramyamounir:
Thank you all for your comments, *I will close this PR*, and open a new one with just the docs to promote `tbp.teleop`, since there is a consensus on doing that. I do not have a strong opinion, but I want to explain why I opened this PR and why I am still leaning towards removing the live plotter (after Jeremy switches to using `tbp.teleop`).
- The live plotter is fully **untested** (we do not have unit/integration tests for it), **unmaintained** (it currently breaks with some experiments), **redundant** (given `tbp.teleop`) and we do not plan on improving or adding new features to it.
- If the convenience of starting a live visualization inside `tbp.monty` outweighs the engineering benefits of decoupling it from the Monty runtime and in to a separate repo, then we should have added a few unit tests for `tbp.teleop` and kept it inside `tbp.monty`, as I originally [proposed](https://github.com/thousandbrainsproject/tbp.monty/pull/1016).
- The real utility I see here is that Jeremy (and possibly others?) is still relying on it, so removing it would add the overhead time of setting up `tbp.teleop` which would delay the progress on the simulator work. Otherwise, I think removing it would simplify the experiment class, decouple visualization tools from the monty runtime and possibly allow us to get rid of matplotlib dependencies in `tbp.monty`.

jeremyshoemaker:
I'm not sure if it makes sense to handle this in your new PR, but I don't think the live plotter is working at all right now. This line is calling a method that was renamed, but this invocation was missed. https://github.com/thousandbrainsproject/tbp.monty/blob/main/src/tbp/monty/frameworks/utils/live_plotter.py#L169

I may throw it in one of my test migration PRs if someone else doesn't get to it first.

ramyamounir:
@jeremyshoemaker Yes, I noticed that before I created this PR and almost opened another PR to correct it but then realized that it is superseded by this PR which deletes the whole live plotter file. I will open another PR to fix this right away.

[PR opened](https://github.com/thousandbrainsproject/tbp.monty/pull/1059)

ramyamounir:
Superseded by [PR#1060](https://github.com/thousandbrainsproject/tbp.monty/pull/1060)

tristanls-tbp:
> If the convenience of starting a live visualization inside tbp.monty outweighs the engineering benefits of decoupling it from the Monty runtime and in to a separate repo, then we should have added a few unit tests for tbp.teleop and kept it inside tbp.monty, as I originally https://github.com/thousandbrainsproject/tbp.monty/pull/1016.

@ramyamounir, I disagree with this assessment and its conclusion. On the one hand (live plotter), we don't have to touch an existing convenience that sometimes works, which we'll eventually remove. On the other hand (tbp.plot), we would be adding brand-new, untested code that will be expected to be maintained forever. I'm not saying the first option is good, but I don't think the second option is better.

PR #1055, opened 2026-06-23
chore: Future work widget updates

vkakerbeck:
A couple of updates to our future work table widget:
- remove unused tag option
- display all columns by default (since otherwise the values are not visible anywhere)
- reorder columns
- Color the entries in status column

### DIFF
--- a/tools/future_work_widget/app/css/future-work-widget.css
+++ b/tools/future_work_widget/app/css/future-work-widget.css
@@ -34,14 +34,70 @@ body {
 }
 
 .badge-status {
-  padding: 2px 4px;
-  border-radius: 4px;
   margin-bottom: 2px;
-  font-size: 0.7rem;
-  display: inline-block;
-  background-color: #ffffff;
-  color: #666;
-  border: 1px solid #999;
+}
+
+.badge.badge-status-open {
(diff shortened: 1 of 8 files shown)


tools/future_work_widget/app/css/future-work-widget.css
tristanls-tbp, line 1: **note**: I didn't observe any status badge color differences that are implied by hover states included in these changes.

tristanls-tbp (commented):
Looks good, however, I think you forgot to delete `docs/snippets/future-work-tags.md`.

tools/future_work_widget/app/css/future-work-widget.css
vkakerbeck: Do you mean you don't see it in the docs preview? As far as I understand, the changes need to be deployed first and we can't see them in the preview. This is how it looks locally:
 

https://github.com/user-attachments/assets/5637cdd3-b55b-4fe2-8416-e547777bedca

vkakerbeck:
Good catch on the tags snippet! I updated that and also deleted the reference to it in the docs

tristanls-tbp (approved):
Ah yes, I forgot that can't see the updates in the preview 👍.

PR #1056, opened 2026-06-24
chore: Render future work metadata on pages

vkakerbeck:
So far, metadata associated with future work pages is only shown in the interactive table. This can be a bit inconvenient, since once one is on a specific future work item's page, the useful metadata is not visible anymore. This PR renders the metadata at the top of future work pages.

Disclaimer: This was 100% written by LLM. I am not very familiar with the internals of the readme sync tool and wouldn't have the expertise to make this change myself. I reviewed all the code and made adjustments, but I want to be transparent about this. If there is a better way to go about this change, let me know.

Since the auto-rendered preview here will not show changes done on the tool, I manually rendered a version based on my local env here: https://docs.thousandbrains.org/v0.0-vclay-metadata-tests/docs/extract-better-features Some example screenshots below:

Screenshot 2026-06-24 at 3 24 12 PM
Screenshot 2026-06-24 at 3 24 22 PM
Screenshot 2026-06-24 at 3 25 56 PM
Screenshot 2026-06-24 at 3 26 14 PM


tristanls-tbp (commented):
Since this was written by an LLM, I haven't reviewed it myself yet.

Here is the initial LLM feedback

Findings (2, both low severity — no correctness bugs)

  - tools/github_readme_sync/future_work_metadata.py:124,152 — a contributor with no status renders the avatar under a `<strong>Status:</strong>` label.
  _field_cell("status", …) always calls _status_cell_content, which appends contributor avatars even when status is absent, then wraps the result in
  _label_cell("Status", …). Verified by execution: input {"contributor": "x"} produces `<strong>Status:</strong> `. Latent today — 0 of 67
  future-work docs with a contributor lack a status — but it will mislabel the day someone adds a contributor before setting status.
  - tools/github_readme_sync/future_work_metadata.py:40,42,48 — badge color palette is duplicated from future-work-widget.css. DEFAULT_BADGE_STYLE,
  SCOPE_STYLES, and STATUS_STYLES hardcode the same hex values already defined as .badge-status-* / .badge-size-* rules in
  tools/future_work_widget/app/css/future-work-widget.css (e.g. open #cce5ff/#004085, in-progress #2f2b5c, small `#f0f0f0/#666`). Two sources of truth: a
  future change to the widget's status/scope colors will silently leave the rendered-on-page metadata block out of sync. (Likely a deliberate tradeoff,
  since [block:html] can't reference external CSS classes — worth at least a comment noting the two must be kept in lockstep.)

vkakerbeck:
I addressed both of those by
- adding a comment to keep the color values in sync (since, as it mentions, we can't reference the external CSS class)
- updating code to label the field Contributor if no Status is specified. However, this isn't really a case that should happen. Whenever there is a contributor, the item should have a status (or generally every item should have a status). We could also have an assert for this instead.

tools/github_readme_sync/tests/md_test.py
```diff
@@ -126,3 +126,30 @@ def test_description_field_in_frontmatter_is_processed(self):
...
+        result = process_markdown(body, slug)
+
+        self.assertEqual(result["estimated-scope"], "large")
+        self.assertEqual(result["improved-metric"], "learning")
+        self.assertEqual(result["output-type"], "prototype, monty-feature, PR")
+        self.assertEqual(result["skills"], "python, research, monty")
+        self.assertEqual(result["contributor"], "vkakerbeck")
+        self.assertEqual(result["status"], "open")
+        self.assertEqual(result["rfc"], "required")
+        self.assertEqual(result["body"].strip(), "Body content here.")
```
tristanls-tbp: **issue (non-blocking)**: Extraneous assertion. Don't need to test the body when checking for metadata fields. Other tests check body content.

tools/github_readme_sync/tests/future_work_metadata_test.py
```diff
@@ -0,0 +1,123 @@
+# Copyright 2025-2026 Thousand Brains Project
```
tristanls-tbp: **issue**: 2026.

tools/github_readme_sync/future_work_metadata.py
```diff
@@ -0,0 +1,243 @@
+# Copyright 2025-2026 Thousand Brains Project
```
tristanls-tbp: **issue**: 2026.

tools/github_readme_sync/future_work_metadata.py
```diff
@@ -0,0 +1,243 @@
...
+    "padding:2px 4px;border-radius:4px;font-size:0.85em;"
+    "display:inline-block;margin:2px 4px 2px 0;"
+)
+# Inline badge colors mirror tools/future_work_widget/app/css/future-work-widget.css
+# (.badge-status-*, .badge-size-*). ReadMe [block:html] cannot reference external
+# stylesheets, so keep these hex values in lockstep with the widget CSS.
+DEFAULT_BADGE_STYLE = f"{BADGE_STYLE}background-color:#e8e8f0;color:#2f2b5c;"
+SKILLS_BADGE_STYLE = DEFAULT_BADGE_STYLE
+SCOPE_STYLES = {
+    "small": f"{BADGE_STYLE}background-color:#f0f0f0;color:#666;",
```
tristanls-tbp: **issue**: Please use `#666666` here for the same reason we have `#ffffff` everywhere instead of `#fff`.

tools/github_readme_sync/future_work_metadata.py
```diff
@@ -0,0 +1,243 @@
...
+GITHUB_AVATAR_URL = "https://github.com"
+GITHUB_USERNAME_PATTERN = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9-]{0,38}$")
+
+BADGE_STYLE = (
+    "padding:2px 4px;border-radius:4px;font-size:0.85em;"
+    "display:inline-block;margin:2px 4px 2px 0;"
+)
+# Inline badge colors mirror tools/future_work_widget/app/css/future-work-widget.css
+# (.badge-status-*, .badge-size-*). ReadMe [block:html] cannot reference external
+# stylesheets, so keep these hex values in lockstep with the widget CSS.
```
tristanls-tbp: **issue**: I doubt anyone will read and obey this comment (at least a human). The best way to keep these in sync is to have a test that reads both and ensures they match.

tools/github_readme_sync/tests/readme_test.py
```diff
@@ -357,6 +357,39 @@ def test_description_field_is_sent_to_readme_as_excerpt(self, mock_post, mock_ge
...
+            category_id="category-id",
+            doc=doc,
+            parent_id="parent-doc-id",
+            file_path="docs/future-work/learning-module-improvements",
+        )
+
+        actual_body = mock_post.call_args[0][1]["body"]
+        self.assertIn("Scope", actual_body)
+        self.assertIn("[block:html]", actual_body)
+        self.assertIn("/docs/future-work-widget-metadata", actual_body)
```
tristanls-tbp: **issue**: These tests are insufficient. The test could pass if the string appears anywhere in the body, but that is not the desired outcome.

**suggestion**: The tool is deterministic, the expected output is known exactly, that's what should be compared.

tristanls-tbp (commented):
Submitting a review in progress.

tools/github_readme_sync/future_work_metadata.py
```diff
@@ -124,27 +124,38 @@ def _badges_html(values: list[str], style: str = DEFAULT_BADGE_STYLE) -> str:
...
+    avatars = _contributor_avatars_html(fields)
     if avatars:
-        parts.append(" ".join(avatars))
+        parts.append(avatars)
+
+    if not parts:
+        return ""
 
-    return " ".join(parts)
+    label = "Status" if has_status else "Contributor"
```
tristanls-tbp: **issue**: This method is called inside `_field_cell` where it is gated by a condition of `if key == "status"`, the name of this method is `_status_field_cell`. The fact that this sometimes returns "contributor" instead of "status" feels off.

tools/github_readme_sync/readme.py
```diff
@@ -305,7 +309,7 @@ def replace_match(match):
     def create_or_update_doc(
         self, order: int, category_id: str, doc: dict, parent_id: str, file_path: str
     ) -> tuple[str, bool]:
-        markdown = self.process_markdown(doc["body"], file_path, doc["slug"])
+        markdown = self.process_markdown(doc["body"], file_path, doc["slug"], doc=doc)
```
tristanls-tbp: **issue**: Passing in the full `doc` defeats the purpose of extracting `body` and `slug` in the first place.

tools/github_readme_sync/readme.py
```diff
@@ -341,7 +345,11 @@ def create_or_update_doc(
 
         return doc_id, created
 
-    def process_markdown(self, body: str, file_path: str, slug: str) -> str:
+    def process_markdown(
+        self, body: str, file_path: str, slug: str, doc: dict | None = None
```
tristanls-tbp: **issue**: It is impossible for `doc` to be `None`. This method is only called with `body` coming from `doc["body"]` and `slug` coming from `doc["slug"]`.

tristanls-tbp:
**question**: An honest question since the more I read, the more complexity I see. Why so much HTML and styling? Why not start with a simple list of data? We can always iterate on the look. 

```markdown
> **Status**: paused
> **Scope**: large
> **Output**: prototype analysis monty-feature PR
> **Skills**: python research monty
> **Metric**: accuracy deformations noise generalization
> **RFC**: required
```
which renders something like...

> **Status**: paused
> **Scope**: large
> **Output**: prototype analysis monty-feature PR
> **Skills**: python research monty
> **Metric**: accuracy deformations noise generalization
> **RFC**: required

The reason I'm picking on HTML and styling because I see hardcoded two columns, which make me wonder how does this look on mobile. I'm also seeing a desire to have consistent styling but the code doesn't guarantee any of it. I'm thinking an alternative implemenation would be to query the future work tool for a single row... 🤔 ...

tristanls-tbp:
**note**: If you want to match the style of the widget, another thing we could do instead of this is to update the Future Work Widget to accept more configuration parameters so that it would display a per-page configured widget, something like: https://futureworkwidget.thousandbrains.ai/index.html?columns=status%2Cestimated-scope%2Coutput-type%2Cskills%2Cimproved-metric%2Crfc&q=Use+models

Screenshot 2026-06-25 at 19 22 29

On a page, it would render something like this...

Screenshot 2026-06-25 at 19 32 53

vkakerbeck:
Yeah those could be options, but I much prefer the two-column look. Integrating the table seems off since it shows the headers as if you could sort the columns, but there is only one, and the markdown version won't render contributors etc. The way the columns are set up, they should adjust to narrow layouts as well (hard to test right now with the general mobile issues of our docs, but see the screenshot below).

Screenshot 2026-06-26 at 9 37 07 AM

I don't think it is super important that the styling exactly matches the table; I just want it to look nice.
I'll have a look at addressing the PR comments on Monday.

tools/github_readme_sync/future_work_metadata.py
vkakerbeck: I don't really care that these colors are in sync. I added it because of the previous PR review comment. The colors are mostly useful when there are many rows in a table, to find things at a glance. In the one page view teh colors don't mean much and I doubt someone will remember them from the table view and compare. I could remove colors here completely if that addresses your concerns. I just thought it looked a bit nicer to keep them.

tools/github_readme_sync/future_work_metadata.py
vkakerbeck: This was also added bc of the previous review comment. I don't really think it will ever happen that we specify a contributor but no status. Like I mentioned above, I could just add an asset for that case instead.

vkakerbeck: I removed this logic here and instead added a check that status is always specified if a contributor is specified.

tools/github_readme_sync/future_work_metadata.py
vkakerbeck: I added the test as you suggested. However if you think it is too complicated we could also remove a lot of teh formatting html and colors and just display the data as a comma separated list of text like this:
Screenshot 2026-06-30 at 3 26 20 PM
Personally, I don't think it looks as nice, but it does simplify things.

tristanls-tbp:
@jeremyshoemaker, would you please take a look at this PR and see if it can be merged?

tools/future_work_widget/tests/validator_test.py
```diff
@@ -252,11 +253,32 @@ def test_github_username_validation(self):
...
+        with self.assertRaises(ValidationError) as cm:
+            FutureWorkRecord.model_validate(record)
+
+        errors = cm.exception.errors()
+        self.assertTrue(
+            any(
+                "contributor is specified, status must also be specified" in e["msg"]
+                for e in errors
+            )
+        )
```
jeremyshoemaker, line 282: This is following the pattern set in the rest of this test suite and I don't think you need to change it, but I'm not sure I like this approach.

We should be able to assert that the length of `errors` is 1 and then assert on the contents of that single error's message rather than just looking through all of the errors for the one we want.

This has the potential to hide issues where the record fails to validate for multiple reasons when it shouldn't.

tools/github_readme_sync/tests/future_work_metadata_test.py
```diff
@@ -0,0 +1,181 @@
...
+    / "css"
+    / "future-work-widget.css"
+)
+
+
+def _normalize_hex(color: str) -> str:
+    color = color.strip().lower()
+    if re.fullmatch(r"#[0-9a-f]{3}", color):
+        color = "#" + "".join(channel * 2 for channel in color[1:])
+    return color
```
jeremyshoemaker: This could use a docstring. It took me a minute to figure out what this is doing. It's taking short 3 digit CSS color codes and expanding them to their canonical 6 digit form, e.g. `#abc` -> `#aabbcc`.

tools/github_readme_sync/future_work_metadata.py
```diff
@@ -0,0 +1,232 @@
...
+RIGHT_COLUMN_KEYS = ("skills", "improved-metric", "rfc")
+
+
+def is_future_work_doc_path(file_path: str) -> bool:
+    """Check whether a document path is within the future work docs section.
+
+    Returns:
+        True when the path is under the future work section.
+    """
+    return FUTURE_WORK_SECTION in Path(file_path).as_posix().split("/")
```
jeremyshoemaker: This is more platform specific than it needs to be. A better approach is to use `Path(file_path).parts` here, as that returns a tuple of strings of the parts of the path, and lets `Path` handle what the separators should be, regardless of platform.

tools/github_readme_sync/tests/future_work_metadata_test.py
```diff
@@ -0,0 +1,181 @@
...
+        self.assertLess(result.index("Status"), result.index("Scope"))
+        self.assertLess(result.index("Scope"), result.index("Output"))
+        self.assertLess(result.index("Skills"), result.index("Metric"))
+        self.assertLess(result.index("Metric"), result.index("RFC"))
+        self.assertLess(result.index("Status"), result.index("Skills"))
+        self.assertLess(result.index("Scope"), result.index("Metric"))
+        self.assertLess(result.index("Output"), result.index("RFC"))
+        self.assertIn(METADATA_DOC_URL, result)
+        self.assertIn("[block:html]", result)
+        self.assertIn("[/block]", result)
```
jeremyshoemaker: This seems like it could be error prone. Things like "open" might be too vague.

Maybe we should have an "expected" rendered string value that we compare to directly. I'm thinking of what might happen if one of these asserts fails. Trying to fix it, we'd probably want to see the rendered content anyway, so a direct comparison might be better and more explicit about what we're looking for.

That string is pretty big though. :thinking: I'm not sure what a better solution is here.

tools/github_readme_sync/tests/future_work_metadata_test.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def test_render_future_work_metadata_wraps_multiple_badges_inline(self):
+        doc = {"output-type": "prototype, monty-feature, PR"}
+        result = render_future_work_metadata(doc)
+
+        self.assertIn(">prototype</span> <span", result)
+        self.assertIn(">monty-feature</span> <span", result)
+        self.assertNotIn(">prototype</span><br>", result)
```
jeremyshoemaker: I can see what this is testing for, but it seems very brittle. Like if the amount of whitespace between tags changes.

tools/github_readme_sync/tests/future_work_metadata_test.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def test_render_future_work_metadata_status_colors(self):
+        doc = {"status": "in-progress"}
+        result = render_future_work_metadata(doc)
+
+        self.assertIn("#2f2b5c", result)
```
jeremyshoemaker: Maybe this color string should be a constant? It's no different than putting an arbitrary number in a test. There's a similar color string in one of the tests above that might make sense to pull out.

tools/github_readme_sync/future_work_metadata.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+
+    if key == "improved-metric":
+        return _label_cell("Metric", _badges_html(_split_values(fields[key])))
+
+    if key == "rfc":
+        rfc = str(fields[key]).strip()
+        if rfc.lower().startswith(("http://", "https://")):
+            rfc_content = (
+                f'<a href="{html.escape(rfc)}" target="_blank" '
```
jeremyshoemaker: Aren't we sanitizing the HTML in `ReadMe.sanitize_html`? In general, this sort of sanitization should happen on the whole HTML output at the end, not small snippets deep inside. It's too easy to add another interpolated string where malicious content could get rendered and forgetting to escape it.

We might want a test to ensure that anything from this metadata gets properly sanitized by `ReadMe` to make sure we aren't missing it.

tools/github_readme_sync/readme.py
```diff
@@ -375,6 +380,19 @@ def sanitize_html(self, body: str) -> str:
...
+
+        Returns:
+            The document body, with metadata prepended when applicable.
+        """
+        if not is_future_work_doc_path(file_path):
+            return body
+        metadata_html = render_future_work_metadata(doc)
+        if not metadata_html:
+            return body
+        return f"{metadata_html}\n\n{body.lstrip()}"
```
jeremyshoemaker: Following on the conversation about sanitization in `future_work_metadata`, I think this should be where we call `sanitize_html` on the value. I don't think it gets sanitized in `ReadMe` otherwise.

tools/github_readme_sync/future_work_metadata.py
```diff
@@ -0,0 +1,232 @@
...
+import html
+import json
+import re
+from pathlib import Path
+from typing import Any
+
+FUTURE_WORK_SECTION = "future-work"
+METADATA_DOC_URL = "/docs/future-work-widget-metadata"
+
+FUTURE_WORK_METADATA_KEYS = frozenset(
```
jeremyshoemaker: Unless we're using this value as a key in a dictionary, we probably don't _need_ to use `frozenset`.

jeremyshoemaker (changes requested):
I left some comments. Some of these things may not matter since this isn't part of Monty itself, I mostly mention them to raise awareness.

The santization one should be fixed just to be safe.

Also, the `md_test.py` and `readme_test.py` files need their copyright comments updated.

tools/future_work_widget/tests/validator_test.py
vkakerbeck: Good point. I changed this, let me know if this is what you were thinking.

tools/github_readme_sync/tests/future_work_metadata_test.py
vkakerbeck: I decided to just remove all the coloring and styling since this significantly simplifies the code and tests (related thread here https://github.com/thousandbrainsproject/tbp.monty/pull/1056#discussion_r3477562577)

tools/github_readme_sync/tests/future_work_metadata_test.py
vkakerbeck: Yeah, I think I could use an expected string like this and assert that it matches:
```
EXPECTED_ALL_FIELDS_RENDERED = """\
[block:html]
{
  "html": "<div style=\\"border:1px solid #ddd;border-radius:8px;padding:12px 16px;margin-bottom:16px;\\"><div style=\\"display:flex;gap:32px;\\"><div style=\\"flex:1;min-width:0;\\"><div style=\\"margin-bottom:8px;\\"><strong>Status:</strong> open <img src=\\"https://github.com/vkakerbeck.png\\" alt=\\"vkakerbeck\\" title=\\"vkakerbeck\\" style=\\"width:24px;height:24px;border-radius:50%;vertical-align:middle;margin-right:4px;\\" /></div><div style=\\"margin-bottom:8px;\\"><strong>Scope:</strong> large</div><div style=\\"margin-bottom:8px;\\"><strong>Output:</strong> prototype, monty-feature, PR</div></div><div style=\\"flex:1;min-width:0;\\"><div style=\\"margin-bottom:8px;\\"><strong>Skills:</strong> python, research, monty</div><div style=\\"margin-bottom:8px;\\"><strong>Metric:</strong> learning</div><div style=\\"margin-bottom:8px;\\"><strong>RFC:</strong> required</div></div></div><div style=\\"width:100%;margin-top:8px;font-size:0.9em;\\">For details on what these values mean, see <a href=\\"/docs/future-work-widget-metadata\\">here</a>.</div></div>"
}
[/block]"""
```

Do you think that would be better?

tools/github_readme_sync/tests/future_work_metadata_test.py
vkakerbeck: removed the badges styling in general

tools/github_readme_sync/readme.py
vkakerbeck: Okay, that makes sense. I am not very familiar with how this is normally done but I tried addressing your comment in this commit: https://github.com/thousandbrainsproject/tbp.monty/pull/1056/changes/6352d8ac1fbd40f66f48cb23c3713cf63d69bf8f Let me know if this is kind of what you were thinking

tools/future_work_widget/tests/validator_test.py
jeremyshoemaker: That's closer to what I had in mind. I was going to suggest removing the "Value error" part, but then I realized that's not the assert failure message, that's the actual contents of `error[0]["msg"]`. :raised_eyebrow:

tools/github_readme_sync/readme.py
```diff
@@ -59,6 +63,15 @@
 ALLOWED_CSS_PROPERTIES = {"width", "height"}
 
 
+def _wrap_readme_html_block(html_content: str) -> str:
+    """Wrap sanitized HTML in a ReadMe HTML block.
+
+    Returns:
+        ReadMe markdown HTML block containing the content.
+    """
+    return f"[block:html]\n{json.dumps({'html': html_content}, indent=2)}\n[/block]"
```
jeremyshoemaker: This is very specific to the `ReadMe` class, and we have similar "static" helper methods on it, so I think this should be moved into that class.

tools/github_readme_sync/tests/future_work_metadata_test.py
jeremyshoemaker: That's more in line with what I was thinking, and I think what @tristanls-tbp was suggesting above with:

> The tool is deterministic, the expected output is known exactly, that's what should be compared.

That said, I'm not sure I like that really long line of text. I'm not sure what the best option here is.

I had totally forgotten about escaping newlines in multiline strings using `\`, i.e. what's on the first line of that multiline string, so maybe we could break up the long line and use that to remove the added newlines?

```python
EXPECTED_ALL_FIELDS_RENDERED = """\
[block:html]
{
  "html": "<div style=\\"border:1px solid #ddd;border-radius:8px;padding:12px 16px;margin-bottom:16px;\\">\
  <div style=\\"display:flex;gap:32px;\\">\
  <div style=\\"flex:1;min-width:0;\\">\
  <div style=\\"margin-bottom:8px;\\"><strong>Status:</strong> open <img src=\\"https://github.com/vkakerbeck.png\\" alt=\\"vkakerbeck\\" title=\\"vkakerbeck\\" style=\\"width:24px;height:24px;border-radius:50%;vertical-align:middle;margin-right:4px;\\" /></div>\
  <div style=\\"margin-bottom:8px;\\"><strong>Scope:</strong> large</div><div style=\\"margin-bottom:8px;\\"><strong>Output:</strong> prototype, monty-feature, PR</div></div>/
  <div style=\\"flex:1;min-width:0;\\"><div style=\\"margin-bottom:8px;\\"><strong>Skills:</strong> python, research, monty</div>\
  <div style=\\"margin-bottom:8px;\\"><strong>Metric:</strong> learning</div><div style=\\"margin-bottom:8px;\\"><strong>RFC:</strong> required</div></div></div>\
  <div style=\\"width:100%;margin-top:8px;font-size:0.9em;\\">For details on what these values mean, see <a href=\\"/docs/future-work-widget-metadata\\">here</a>.</div></div>"
}
[/block]"""
```

Those breaks might not be exactly in the best places, but you get the idea.

jeremyshoemaker (changes requested):
Just a couple more comments.

tools/github_readme_sync/tests/future_work_metadata_test.py
vkakerbeck: okay, I added this changed test with predefined expected HTML & line breaks. I also removed some tests in that file that became redundant given the simplifications made before and the fixed html check. lmk what you think

tools/github_readme_sync/tests/future_work_metadata_test.py
```diff
@@ -12,11 +12,36 @@
...
+    "<strong>Metric:</strong> learning</div>"
+    '<div style="min-width:0;overflow-wrap:break-word;"><strong>RFC:</strong> '
+    '<a href="https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/'
+    '0015_future_work.md" target="_blank" rel="noopener noreferrer">RFC</a></div>'
+    "</div>"
+    '<div style="margin-top:8px;font-size:0.9em;overflow-wrap:break-word;">'
+    "For details on what these values mean, see "
+    '<a href="/docs/future-work-widget-metadata">here</a>.</div></div>'
+)
+
```
jeremyshoemaker, line 44: That works too. It also solves the problem of not being able to indent a multi-line string (though there is [`dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) which is a topic for another day).

tools/github_readme_sync/tests/future_work_metadata_test.py
```diff
@@ -0,0 +1,80 @@
...
+                "https://github.com/thousandbrainsproject/tbp.monty/blob/main/rfcs/"
+                "0015_future_work.md"
+            ),
+        }
+
+        self.assertEqual(render_future_work_metadata(doc), EXPECTED_ALL_FIELDS_RENDERED)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
jeremyshoemaker: Sorry, I didn't catch this earlier in my earlier reviews.

We had a discussion a while back about whether to include this `unittest.main` block or not. Since we generally expect people to use `pytest` to run the test suite, we decided to not add these since they imply we also support running the tests with `unittest` which none of us do regularly (as far as I know). So taking these out makes it less likely that people will try.

tools/github_readme_sync/tests/md_test.py
```diff
@@ -126,3 +126,29 @@ def test_description_field_in_frontmatter_is_processed(self):
...
+improved-metric: learning
+output-type: prototype, monty-feature, PR
+skills: python, research, monty
+contributor: vkakerbeck
+status: open
+rfc: required
+---
+
+Body content here.
+"""
```
jeremyshoemaker: Since I mentioned it in a previous comment, maybe this would be a good place to suggest this. Take it or leave it. I'm just mentioning it as a possibility. If you do change it, I'd probably go and change the rest of the text strings in this file to match.

You can make this look better using [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) from the standard library. The way that would look is something like this:

```suggestion
        body = dedent("""\
        ---
        title: Deal with Incomplete Models
        description: Example description
        estimated-scope: large
        improved-metric: learning
        output-type: prototype, monty-feature, PR
        skills: python, research, monty
        contributor: vkakerbeck
        status: open
        rfc: required
        ---

        Body content here.
        """)
```

The way this works is that `dedent` looks at the first line of text and any whitespace it starts with gets removed from it and all the following lines. The first line is setting where the first column in the text should be. 

Note that to make it work, we have to escape the first newline we're adding to the multi-line string with a `\` which is why the string starts with `"""\`. That prevents `dedent` from thinking the first line has no leading whitespace.

tools/github_readme_sync/future_work_metadata.py
```diff
@@ -0,0 +1,174 @@
...
+        doc: Processed markdown document containing optional metadata fields.
+
+    Returns:
+        HTML for the metadata block, or an empty string when no metadata is
+        present. Callers must sanitize this HTML before publishing it.
+    """
+    fields = {
+        key: doc[key]
+        for key in FUTURE_WORK_METADATA_KEYS
+        if key in doc and _has_metadata_value(doc[key])
```
jeremyshoemaker: Since `_has_metadata_value` accepts `None`, this could be simplified.

```suggestion
        if _has_metadata_value(doc.get(key))
```

Using `dict.get()` is usually the better option because you can provide a default for when it's not present and not worry about KeyErrors. The default "default" is `None`.

tools/github_readme_sync/md.py
```diff
@@ -23,6 +25,10 @@ def process_markdown(body: str, slug: str) -> dict:
     if "description" in frontmatter:
         doc["description"] = frontmatter.get("description", "")
 
+    for key in FUTURE_WORK_METADATA_KEYS:
+        if key in frontmatter and frontmatter[key] is not None:
```
jeremyshoemaker: This can be simplified to:

```suggestion
        if frontmatter.get(key):
```

The only way that might be a problem is if we don't want empty strings to be skipped, in which case, keeping the `is not None` will solve that.

tools/github_readme_sync/md.py
```diff
@@ -10,6 +10,8 @@
 
 import yaml
 
+from tools.github_readme_sync.future_work_metadata import FUTURE_WORK_METADATA_KEYS
+
 
 def process_markdown(body: str, slug: str) -> dict:
```
jeremyshoemaker, line 16: I feel like this whole function could use a rewrite, since we're accessing multiple keys in `frontmatter` in different ways. I'm guessing it just built up this way over time and we haven't had a chance to refactor it.

All that said, this is in a tool, so I don't know how much we should actually worry about it.

tools/github_readme_sync/readme.py
```diff
@@ -510,6 +534,14 @@ def delete_version(self):
     def _should_ignore_video(self, identifier: str, ignore_list: list[str]) -> bool:
         return identifier in ignore_list
 
+    def _wrap_html_block(self, html_content: str) -> str:
+        """Wrap HTML in a ReadMe HTML block.
+
+        Returns:
+            ReadMe markdown HTML block containing the content.
+        """
+        return self._create_video_block("html", {"html": html_content})
```
jeremyshoemaker: Maybe this helper method should just be renamed to `_create_block` or `_wrap_block`. If that were changed, I don't think that this `_wrap_html_block` needs to be a separate method. I'd probably just inline it where it's called since it's only called once.

jeremyshoemaker (commented):
I have a few more comments, it should be the last of them. I think looking at it with fresh eyes made me see more things.

tools/github_readme_sync/tests/md_test.py
vkakerbeck: oh yes, that does look a lot nicer

tools/github_readme_sync/md.py
vkakerbeck: Yes that would probably be good to refactor at some point. Do you want that to be part of this PR?

tools/github_readme_sync/readme.py
vkakerbeck: Good point. I renamed `_create_video_block` to `_create_readme_block` and just call that everywhere

vkakerbeck:
I think I addressed all your remaining comments. But due to the readme API update I'm currently not able to verify that everything still looks the same when rendered so I'd hold off with this PR until github.com/thousandbrainsproject/tbp.monty/pull/1103 is merged

tools/github_readme_sync/md.py
jeremyshoemaker: I wouldn't worry about it. I realize now that comment didn't actually go anywhere.

jeremyshoemaker (approved):
This looks good to me now. I'll go ahead and approve.

Let me know if you need me to look at any further changes that come out of updating it once that other PR is merged.

PR #1059, opened 2026-06-25
fix: stale function name in the live plotter

ramyamounir:
This is to correct the stale function name in the live plotter after [this refactor](https://github.com/thousandbrainsproject/tbp.monty/pull/1034) where `EvidenceGraphLM.get_evidence_for_each_graph()` was renamed to `EvidenceGraphLM.evidence_for_each_graph()`


jeremyshoemaker (approved):
Oh, that reformatting explains why when I tried to make a link to this line from my fixed branch, the line number was off by one.

Looks good to me.

tristanls-tbp:
Ah, thank you for the cleanup.

PR #1061, opened 2026-06-25
test: create MuJoCo version of evidence_lm_test.py

jeremyshoemaker:

There's not a lot of substantive code changes to the new version of `evidence_lm_test.py`. Aside from the added return types, which was the only thing MyPy was flagging, these are the changes:

```diff
11,19d10
<
< import pytest
<
< from tests import HYDRA_ROOT
<
< pytest.importorskip(
<     "habitat_sim",
<     reason="Habitat Sim optional dependency not installed.",
< )
30a22
> from tests import HYDRA_ROOT
48c40
<                 f"experiment=test/evidence_lm/{test_name}",
---
>                 f"experiment=test/evidence_lm/{test_name}_mujoco",
118c110
<             exp.env._env.remove_all_objects()
---
>             exp.env.remove_all_objects()
468c460
<             exp.env._env.remove_all_objects()
---
>             exp.env.remove_all_objects()
```

On the configuration side of things, the majority of these are the same except for the swapped `/environment` configuration.

There are two exceptions. First, `fixed_actions_evidence_mujoco`, where I had to change the `max_match_distance` from 0.001 to 0.0001 to prevent the LM from thinking the surface of the cube was the same as the surface of the capsule side. I think this is because it seems like the depth values coming from MuJoCo are slightly shallower in curvature than the values from Habitat. This could be caused by differences in the number of polygons the models use, combined with the low resolution of the images.

Second, the `no_features_mujoco` configuration is using an alternative agent configuration that offsets the starting Z position of the agent by 1cm. This small change makes the test behave the same as the Habitat version. This seems to me like a chaos theory type result, where the test is very sensitive to initial conditions.

Finally, there are a couple of changes to the `MuJoCoSimulator` here that are somewhat incidental. 

I've added a scaling adjustment to the capsule to better match the dimensions of the Habitat version. Previously the MuJoCo version was more elongated than the Habitat version, but now it's a much closer match, almost pixel to pixel. 

The other change was to the type of the agent partials. This came from an optimization I added but since removed to pre-create the renderers when creating the agents rather than on demand.  I thought that the renderer creation was affecting the first episode run time, but that ended up being debugger pauses instead. Trying to access the `keywords` property on the agent_partials with the original type leads to mismatched types since `Callable` doesn't have the attributes that `functools.partial`s have, e.g. the `keywords` property. Elsewhere, I `cast` the variable to the `partial` type, but that didn't work in the other place I was using this. The better solution is a Protocol that defines the specific `__call__` signature that we use, as well as any properties or methods of `partial` that we use.

The reason I added it is gone, but it lets me remove the `cast` and I think it's a better approach overall.


src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -158,6 +170,7 @@ def __init__(
         self.data = MjData(self.model)
         self._data_path = Path(data_path) if data_path else None
         self._raise_actuate_missing = raise_actuate_missing
+        self._renderers = {}
```
jeremyshoemaker, line 173: This movement of where this is defined came out of the aforementioned removed optimization to pre-create the renderers when creating the agents.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -227,8 +239,9 @@ def _configure_lights(self) -> None:
         self.spec.visual.headlight.specular = (0.0, 0.0, 0.0)
         # Add a directional light on the "front" side of the object.
         self.spec.worldbody.add_light(
-            pos=(0, 0, 0.2),
+            pos=(0, 0.0, 0.2),
```
jeremyshoemaker, line 242: I was trying to adjust the position of this light upwards to match Habitat's lighting better, but I don't think this field does anything. When I changed it back, I didn't match the original. I'll probably just go and change the first parameter to also have a decimal point.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -227,8 +239,9 @@ def _configure_lights(self) -> None:
...
             diffuse=(0.6, 0.6, 0.6),
+            specular=(0.1, 0.1, 0.1),
```
jeremyshoemaker, line 244: Trying to reduce a bright specular highlight in the middle of the initial images.

tests/unit/simulators/mujoco/noop_agent_test.py
```diff
@@ -96,7 +96,7 @@ def test_agent_observation_single_sensor(self) -> None:
             assert rgba.min() == 0
             assert rgba.max() == 255
 
-    def test_agent_observation_multiple_resolutions(self):
+    def test_agent_observation_multiple_resolutions(self) -> None:
```
jeremyshoemaker, line 99: Incidental type hint from when I was messing with the `test_agent_observation_single_sensor` test when I thought I needed to adjust the depth values to account for the distance between the near plane and the camera position. MuJoCo already does that adjustment for us.

This whole line of thought came out of misreading a comment on an issue in their project, but looking at the rendering code, it's clear they handle this for us. And that single sensor test proves it for us. I'm glad I wrote it.

tristanls-tbp:
I'm adding @vkakerbeck to approve/disapprove the test configuration changes highlighted in the description.

src/tbp/monty/simulators/mujoco/simulator.py
tristanls-tbp: **issue**: I don't understand this explanation. As far as I can tell, this movement is a no-op. `self._create_agents()` is the only method call between this and the previous location, and I don't see anything use it there.

tests/mujoco/integration/frameworks/models/evidence_matching/evidence_lm_test.py
```diff
@@ -0,0 +1,599 @@
...
+from omegaconf import DictConfig
+
+from tbp.monty.frameworks.experiments.mode import ExperimentMode
+from tests import HYDRA_ROOT
+from tests.unit.resources.unit_test_utils import BaseGraphTest
+
+
+class EvidenceLMTest(BaseGraphTest):
+    def setUp(self) -> None:
+        """Code that gets executed before every test."""
```
tristanls-tbp: **issue**: Extraneous comment.

tests/mujoco/integration/frameworks/models/evidence_matching/evidence_lm_test.py
```diff
@@ -0,0 +1,599 @@
...
+            self.five_lm_three_done_cfg = hydra_config("five_lm_three_done")
+            self.five_lm_off_object_cfg = hydra_config("five_lm_off_object")
+            self.five_lm_no_threading_cfg = hydra_config("five_lm_no_threading")
+            self.five_lm_maxnn1 = hydra_config("five_lm_maxnn1")
+            self.five_lm_bounded = hydra_config("five_lm_bounded")
+            self.noise_mixin_cfg = hydra_config("noise_mixin")
+            self.noisy_sensor_cfg = hydra_config("noisy_sensor")
+
+    def tearDown(self) -> None:
+        """Code that gets executed after every test."""
```
tristanls-tbp: **issue**: Extraneous docstring.

tests/mujoco/integration/frameworks/models/evidence_matching/evidence_lm_test.py
```diff
@@ -0,0 +1,599 @@
...
+                f"{key} should be stored in buffer when using DETAILED logging.",
+            )
+        self.assertGreater(
+            len(exp.model.learning_modules[0].buffer.stats["possible_matches"]),
+            1,
+            "When using detailed logging we should store matches at every steps.",
+        )
+
+    def test_pre_episode_raises_error_when_no_object_is_present(self) -> None:
+        """Test that pre_episode raises an error when no object is present."""
```
tristanls-tbp: **issue**: Extraneous docstring (test name tells me this already).

tests/mujoco/integration/frameworks/models/evidence_matching/evidence_lm_test.py
```diff
@@ -0,0 +1,599 @@
...
+            exp.logger_handler.pre_eval(exp.logger_args)
+            exp.model.set_experiment_mode(exp.experiment_mode)
+            for _ in range(exp.n_eval_epochs):
+                exp.run_epoch()
+            exp.logger_handler.post_eval(exp.logger_args)
+            eval_stats = pd.read_csv(output_dir / "eval_stats.csv")
+            self.check_eval_results(eval_stats, num_lms=5)
+
+            for lm in exp.model.learning_modules:
+                lm.max_match_distance = 0.01
```
tristanls-tbp, line 360: **note**: Huh... I guess we just change this sometimes...

tests/mujoco/integration/frameworks/models/evidence_matching/evidence_lm_test.py
```diff
@@ -0,0 +1,599 @@
...
+            self.assertLess(
+                eval_stats["num_steps"][5 + lm_id],
+                13,
+                f"LM {lm_id} in episode 1 should have less than 13 steps"
+                f" since it was off the object for longer than other LMs.",
+            )
+
+    def test_5lms_pre_episode_raises_error_when_no_object_is_present(self) -> None:
+        """Test that pre_episode raises an error when no object is present."""
+        pprint("...parsing experiment...")
```
tristanls-tbp: **issue**: Extraneous docstring and `pprint`.

tristanls-tbp (commented):
Thanks for getting this working. As I mentioned in a comment, I'm looking for approval from a researcher saying these changes to get the tests passing are acceptable.

For me, a few extraneous docstrings for the most part, otherwise, it looks good.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I was moving things around when I was adding code to initialize the renders that an agent will use in `_create_agents()`. I removed that code after I realized it wasn't needed. This change didn't get undone.

I can undo the change, but it doesn't hurt anything, and it might actually make more sense here.

tests/mujoco/integration/frameworks/models/evidence_matching/evidence_lm_test.py
jeremyshoemaker: These have been some of the first things I've changed with these, not sure how I missed them in this one.

tests/mujoco/integration/frameworks/models/evidence_matching/evidence_lm_test.py
jeremyshoemaker: I'm surprised that wasn't removed when we were updating these for Hydra. Must have slipped through.

src/tbp/monty/simulators/mujoco/simulator.py
tristanls-tbp: I'm ok with the change if it's a no-op. I was just making sure I understood correctly. Maybe it should have been a **question**.

tests/mujoco/integration/frameworks/models/evidence_matching/evidence_lm_test.py
```diff
@@ -0,0 +1,594 @@
...
+        for lm_id in [0, 2, 3, 4]:
+            self.assertLess(
+                eval_stats["num_steps"][5 + lm_id],
+                13,
+                f"LM {lm_id} in episode 1 should have less than 13 steps"
+                f" since it was off the object for longer than other LMs.",
+            )
+
+    def test_5lms_pre_episode_raises_error_when_no_object_is_present(self) -> None:
+        """Test that pre_episode raises an error when no object is present."""
```
tristanls-tbp: **issue (non-blocking)**: This docstring is still extraneous because it just restates the test name (aside from the pprint you removed). I am making this non-blocking because the rest of the changes look good to me, and I don't want to block this PR if Viviane approves it tomorrow.

tristanls-tbp (approved):

vkakerbeck (commented):
Those config changes, together with the explanation you give, seem reasonable to me. The only thing I want to clarify is what you mean by "This could be caused by differences in the number of polygons the models use, combined with the low resolution of the images."
What resolution are you referring to here? Don't the sensor patches still have a 64x64 resolution?

jeremyshoemaker:
> What resolution are you referring to here? Don't the sensor patches still have a 64x64 resolution?

The sensor resolution _is_ the same, I just mentioned it because I don't know exactly how Habitat and MuJoCo determine depth. Like is it just a sample of the center point of each pixel? Is it some averaging of the surrounding area? I really don't know how the point clouds are determined in the LMs to calculate surface normals and and principal curvatures, so I don't know if it matters.

The thing that makes me wonder is because the RGB rendering in MuJoCo is using anti-aliasing and the Habitat version is not. I don't think that will significantly impact the results in most cases, if at all, but it does make me wonder how they determine depth.

I actually think differences in the number of polygons the capsule meshes use would have much more of an effect here, and I really doubt they are the same between the two simulators. Depending on how many facets the cylinder part has, or how many triangles the rounded caps have could throw off the principal curvature calculations just enough to end up with different results.

vkakerbeck:
Ah okay, thanks for clarifying!

vkakerbeck (approved):

jeremyshoemaker:
Looks like I slipped in a change that's causing some test failures in the `test-uv-monty` suite, so I'll need to fix that.

PR #1063, opened 2026-06-25
chore: updating readme script to validate slug and file path accuracy

brothman-tbp:
- Updated the `hierarchy.md` file to fix broken relative links when users would click into that document on the Github UI.
- Updated our links checker to ensure that all slugs are lowercase, and that we confirm the relative file path is also accurate.


tools/github_readme_sync/hierarchy.py
```diff
@@ -116,10 +116,15 @@ def check_hierarchy_file(folder: str):
             parent_stack.append(new_doc)
 
             slug_path = folder.joinpath(*[el["slug"] for el in parent_stack])
-            errors = sanity_check(slug_path.with_suffix(".md"))
+            errors = sanity_check_slugs(slug_path.with_suffix(".md"))
             if errors:
                 link_check_errors.extend(errors)
 
+            file_path = folder.joinpath(extract_file_path(line))
+            errors2 = sanity_check_file_path(file_path)
```
vkakerbeck: Maybe you could give this a more expressive name? Like `file_path_errors`?

vkakerbeck (commented):
Nice updates! Just one comment inline. On a high-level, could you also update the corresponding docs (https://docs.thousandbrains.org/docs/documentation#creating-a-document) to describe the rules that are being enforced here (i.e. slugs need to be all lower-case etc.)?

brothman-tbp:
> Nice updates! Just one comment inline. On a high-level, could you also update the corresponding docs (https://docs.thousandbrains.org/docs/documentation#creating-a-document) to describe the rules that are being enforced here (i.e. slugs need to be all lower-case etc.)?

Great suggestion! That will be the next PR - I already have it written, but when I tried to upload it in this one, the check links function broke (because it uses the current code for the tests, not this update), and then the example markdown (to demonstrate how slugs and relative links work in hierarchy.md) was checked as a link and the tests failed.

tools/github_readme_sync/hierarchy.py
brothman-tbp: Done, thank you!

vkakerbeck (approved):
Ah okay, that makes sense

PR #1064, opened 2026-06-25
test: create MuJoCo version of motor_policy_test.py

jeremyshoemaker:

I had to change this test in a few places, I'll annotate the test itself with explanations. I also noticed the import for `ExperimentMode` was an indirect import, which was flagged because it's correctly not in the `__all__` where it was imported from.


tests/mujoco/integration/frameworks/models/salience/motor_policy_test.py
```diff
@@ -0,0 +1,184 @@
...
+    policy (which is actually necessary) will work correctly in that case.
+    """
+
+    env: ClassVar[SimulatedObjectEnvironment]
+    env_interface: ClassVar[Interface]
+    motor_policy: ClassVar[MotorPolicy]
+    view_finder_res: ClassVar[Resolution2D]
+
+    observations: Observations
+    proprioceptive_state: ProprioceptiveState
```
jeremyshoemaker, line 76: These two were originally class variables, but they can't be because ClassVars can't be assigned to through an instance. It's only something a type checker will flag. It also means I had to move their initialization to the `setUp` method.

tests/mujoco/integration/frameworks/models/salience/motor_policy_test.py
```diff
@@ -0,0 +1,184 @@
...
+                        resolution=cls.view_finder_res,
+                        zoom=1.0,
+                    ),
+                },
+            ),
+        ]
+
+        cls.env = MuJoCoSimulator(agents=agent_partials)
+
+        transforms = [
```
jeremyshoemaker, line 102: This had `MissingToMaxDepth` which for MuJoCo does nothing because it returns very large values for the background void instead of the 0 that Habitat does.

tests/mujoco/integration/frameworks/models/salience/motor_policy_test.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+        policy_result = self.motor_policy(
+            ctx=MagicMock(),
+            observations=self.observations,
+            state=self.proprioceptive_state,
+            percept=MagicMock(),
+            goal=goal,
+        )
+        # We're storing both of these values not only for the assertions below, but
+        # for the next iteration of the Hypothesis test.
```
jeremyshoemaker, line 177: I'm adding this comment because I was confused when, because of the ClassVar thing mentioned previously, I tried to change these to local variables and started getting failures.

It took me a while to realize we were using the updated values in the next iteration of the test, and so I think we should definitely call that out here.

tests/mujoco/integration/frameworks/models/salience/motor_policy_test.py
```diff
@@ -0,0 +1,184 @@
...
+            experiment_mode=ExperimentMode.EVAL,
+            rng=np.random.default_rng(42),
+            seed=42,
+        )
+
+        # Since we aren't going through the full experiment process, we don't call
+        # pre_epoch on the environment interface we built. This means we never add
+        # an object to the scene using the object_init_sampler, which is why we mocked
+        # it. We need to manually add the object to the scene for the test.
+        cls.env.add_object(name="cubeSolid", position=CUBE_LOCATION)
```
jeremyshoemaker, line 133: This comment spells it out pretty clearly, but it wasn't obvious that the original test was relying on the `HabitatEnvironment` constructor that adds the provided `objects` to the scene.

In a real experiment, that object is never used, because it gets removed when `pre_epoch` is called. In fact, all of our experiment configurations set it to a "coneSolid" that is never used.

tests/mujoco/integration/frameworks/models/salience/motor_policy_test.py
```diff
@@ -0,0 +1,184 @@
...
+            use_state=True,
+            sender_id="view_finder",
+            sender_type="SM",
+            goal_tolerances=None,
+            info=None,
+        )
+        policy_result = self.motor_policy(
+            ctx=MagicMock(),
+            observations=self.observations,
+            state=self.proprioceptive_state,
```
tristanls-tbp: **issue**: Should be `MotorSystemState` here.

We briefly talked about this at the ops meeting. But here's the context for why this should be `MotorSystemState(self.proprioceptive_state)`:

https://github.com/thousandbrainsproject/tbp.monty/blob/4ef826f450f35979f875fa8c484d05883ec54ef1/src/tbp/monty/frameworks/models/motor_system.py#L127-L138

The test code uses the motor policy, as the motor system (via the policy selector) would.

The original intent of the `MotorSystemState` wrapper was for the motor system to provide additional state to the policy, for example, something like:
```python
motor_system_state = MotorSystemState(
    motor_only_step=self._motor_only_step,
    proprioceptive_state=proprioceptive_state,
)
```
But it didn't work out that way. Nevertheless, the expected type is `MotorSystemState`, so it is appropriate to do here what the motor system does: `MotorSystemState(self.proprioceptive_state)`.

tests/mujoco/integration/frameworks/models/salience/motor_policy_test.py
```diff
@@ -0,0 +1,184 @@
...
+    def setUpClass(cls) -> None:
+        cls.view_finder_res = Resolution2D(64, 64)
+        agent_partials = [
+            partial(
+                DistantAgent,
+                agent_id=AGENT_ID,
+                position=(0.0, 1.5, 0.2),
+                sensor_configs={
+                    SensorID("patch"): SensorConfig(
+                        resolution=Resolution2D(width=64, height=64),
```
tristanls-tbp: **nitpick**: Inconsistent `Resolution2D` use convention between here and 8 lines up (`cls.view_finder_res`).

tests/mujoco/integration/frameworks/models/salience/motor_policy_test.py
tristanls-tbp: **praise**: Thank you for adding this. I was present when we wrote this test and I too got confused reading it. This helps 👍.

tristanls-tbp (commented):
Everything looks good to me, just the one `MotorSystemState` comment, hopefully with informative context this time.

tests/mujoco/integration/frameworks/models/salience/motor_policy_test.py
jeremyshoemaker: That was how the test originally did it, and I didn't catch it.

jeremyshoemaker: Oh, the reason it's this way is because `view_finder_res` is used in the tests, and this one isn't. I'm not sure it makes sense though. I'll refactor it.

jeremyshoemaker: I went back and forth, but decided to add a class variable for this one too, so we can reuse it in the `DepthTo3DLocation` transform.

tests/mujoco/integration/frameworks/models/salience/motor_policy_test.py
```diff
@@ -0,0 +1,188 @@
...
+    env_interface: ClassVar[Interface]
+    motor_policy: ClassVar[MotorPolicy]
+    view_finder_res: ClassVar[Resolution2D]
+
+    observations: Observations
+    proprioceptive_state: ProprioceptiveState
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.patch_res = Resolution2D(64, 64)
```
tristanls-tbp, line 80: **todo**: I think `patch_res` needs to get declared as a `ClassVar` above. I'm approving conditional on this todo.

tristanls-tbp (approved):
Approving with one **todo**.

PR #1066, opened 2026-06-26
feat: implement semantic sensors in MuJoCo

jeremyshoemaker:

This was originally just going to create a MuJoCo version of `get_good_view_test.py`, but in order to do that, I had to implement semantic sensors in the MuJoCo simulator. Adding the `semantic` config option to `SensorConfig` also meant I had to go back and edit the previous MuJoCo environment configurations to add the `_target_` to the sensor configs, so they could pick up the default value for it.

No significant changes were made to the test itself.
```diff
10,16d9
< import pytest
<
< pytest.importorskip(
<     "habitat_sim",
<     reason="Habitat Sim optional dependency not installed.",
< )
<
33a27
> from tbp.monty.frameworks.sensors import SensorID
41c35
<             f"experiment=test/integration/positioning_procedures/get_good_view/{test_name}",
---
>             f"experiment=test/integration/positioning_procedures/get_good_view/{test_name}_mujoco",
76a71
>                 assert exp.env_interface is not None  # for type narrowing
78c73
<                 view = observation[agent_id]["view_finder"]
---
>                 view = observation[agent_id][SensorID("view_finder")]
128a124
>                 assert exp.env_interface is not None  # for type narrowing
130c126
<                 view = observation[agent_id]["view_finder"]
---
>                 view = observation[agent_id][SensorID("view_finder")]
```


src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -328,16 +326,18 @@ def add_object(
...
 
         self._recompile()
 
-        # Using the object count for the semantic_id will give a distinct
-        # value for each added object, and _might_ map to MuJoCo's internal
-        # object IDs if we need to use those.
+        self._object_count += 1
+        if semantic_id is None:
+            semantic_id = SemanticID(self._object_count)
+        object_id = ObjectID(self.model.geom(obj_name).id)
```
jeremyshoemaker: Now that I'm pulling the object ID from the simulator model, I think this should also be used for the default semantic ID. That probably means that `_object_count` is not needed anymore, but I'll need to confirm that.

jeremyshoemaker: The updated version more closely matches the Habitat approach of using the type ID of the object added. I'm not sure if this change constitutes a breaking change though, since the values we're returning will change. An argument for would be that the value changed so it's a different API behaviour. An argument against would be that SemanticIDs are meant to be opaque so the underlying value changing shouldn't matter.

jeremyshoemaker:
When I was writing this, I was thinking the test I migrated here was covering the semantic sensor behaviour enough, but I think I'm gonna add some unit tests for it to be sure it matches our expectations.

tristanls-tbp:
**note**: I'm happy you're continually updating this PR, but I converted it to a draft so I can review it when it's ready.

src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -57,6 +57,8 @@ class MontyExperiment:
     and episode).
     """
 
+    env_interface: Interface | None
```
tristanls-tbp, line 60: **note**: There's always the option below if the names start to feel too generic:
```python
from tbp.monty.experiment import environment
# ...
env_interface: environment.Interface | None
```

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -328,16 +335,23 @@ def add_object(
...
+        self._object_count += 1
+
+        added_obj = self.model.geom(obj_name)
+        # MuJoCo gives us a NumPy array of 1 element for some reason, so we use the
+        # first element for the type.
+        added_obj_type = added_obj.type[0]
+
+        if semantic_id is None:
+            semantic_id = SemanticID(added_obj_type)
+        object_id = ObjectID(self.model.geom(obj_name).id)
```
tristanls-tbp: **question**: Is there a reason this does not use `added_obj`? As far as I can tell, there are no writes between the initialization of `added_obj` and here.

tests/unit/simulators/mujoco/agent_actions_test.py
```diff
@@ -46,8 +46,8 @@
 from tbp.monty.simulators.mujoco.agents import Axis, DistantAgent, SurfaceAgent
 from tbp.monty.simulators.mujoco.simulator import ActuateMethodMissing
 from tests.unit.simulators.mujoco.noop_agent_test import (
-    TEST_AGENT_ID,
-    TEST_SENSOR_ID,
+    AGENT_ID,
+    PATCH_SENSOR_ID,
     default_agent_args,
```
tristanls-tbp: **question**: This seems like a peculiar place to import from. Any reason these are coming from `noop_agent_test`, which is not a test library module but a specific test. 

**note**: It feels like if we're gonna import these, they should be somewhere more... central? `tests.unit.simulators` maybe?

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -328,16 +335,23 @@ def add_object(
...
         else:
             self._add_custom_object(obj_name, name, position, rotation, scale)
-        self._object_count += 1
 
         self._recompile()
+        self._object_count += 1
+
+        added_obj = self.model.geom(obj_name)
+        # MuJoCo gives us a NumPy array of 1 element for some reason, so we use the
+        # first element for the type.
```
tristanls-tbp: **note**: The reason is that in `added_obj = self.model.geom(obj_name)`, `added_obj` is the returned accessor for arrays in the `MjModel` object `self.model` associated with the geom named `obj_name`. I think this is a multi-item general interface that MuJoCo reuses for a single item here.

So `added_obj.type` accesses the `type` array, which is a writeable NumPy array.

See https://mujoco.readthedocs.io/en/latest/python.html#named-access

**suggestion**: Update comment and link to https://mujoco.readthedocs.io/en/latest/python.html#named-access

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -141,6 +151,7 @@ class MuJoCoSimulator(SimulatedObjectEnvironment):
     model: MjModel
     data: MjData
 
+    id_to_semantic_id: dict[ObjectID, SemanticID]
```
tristanls-tbp, line 153: **nitpick**: It seems to me like this should be with other "public" props above.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -328,16 +335,23 @@ def add_object(
...
+
+        added_obj = self.model.geom(obj_name)
+        # MuJoCo gives us a NumPy array of 1 element for some reason, so we use the
+        # first element for the type.
+        added_obj_type = added_obj.type[0]
+
+        if semantic_id is None:
+            semantic_id = SemanticID(added_obj_type)
+        object_id = ObjectID(self.model.geom(obj_name).id)
+        self.id_to_semantic_id[object_id] = semantic_id
```
tristanls-tbp, line 350: **question**: Why is this being built here?

The reason I ask is that https://github.com/thousandbrainsproject/tbp.monty/blob/196b73323d14b38490a6b382357c43a1b36c07b6/src/tbp/monty/experiment/environment.py#L291-L312 builds the semantic mapping between `ObjectID` and `SemanticID`.

tristanls-tbp (commented):
Submitting comments from a review in progress.

src/tbp/monty/frameworks/experiments/monty_experiment.py
jeremyshoemaker: My thinking is unless it's _really_ ambiguous, using the unqualified name is fine, especially when using a tool that can tell you what it is pretty quickly, e.g. via tooltip, or quickly take you to the definition if need be.
image
The name of the variable is `env_interface` which gives an indication of what it is. If it was just `interface` then I'd be concerned with ambiguity.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: Because I wrote this line before I introduced `added_obj`. I definitely don't need to be looking it up twice.

tests/unit/simulators/mujoco/agent_actions_test.py
jeremyshoemaker: I wasn't quite sure where a good central place for these would be back when I added this test. We create these constants across a lot of tests, not just the simulator tests, and I think it would make sense to decide on a central place for them all.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: Yeah, not sure how that extra line got there.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: Looking at `create_semantic_mapping`, that appears to be creating a mapping of labels to semantic IDs, e.g. `{"cubeSolid": 1, "cylinderSolid": 2}` and it's mirror. This also explains the thing I was commenting about the other day where in a multi object experiment, several objects in the scene end up with the same semantic ID because they're the same type of object, e.g. "cubeSolid".

The `id_to_semantic_id` mapping in the `MuJoCoSimulator` is mapping the object ID — technically they're MuJoCo's internal IDs for the objects but we're using those as object IDs — to semantic IDs, so that when we get the semantic data from MuJoCo that uses the internal IDs, we can map them to the semantic IDs that Monty expects. See the changes to `Embodiment.observations`.

It's public because the `Embodiment` needs access to it to perform the translation.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I meant to come back and clean that up, and forgot to leave a TODO for myself. I don't like handwavy explanations.

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -218,10 +218,31 @@ def observations(self) -> AgentObservations:
             depth_data = renderer.render()
             renderer.disable_depth_rendering()
 
-            obs[sensor_id] = SensorObservation(
-                depth=depth_data,
-                rgba=rgba_data,
-            )
+            if sensor_cfg.semantic:
```
tristanls-tbp: **note**: I wonder if it'd read better with early exit instead:

```python
if not sensor_cfg.semantic:
  return obs[sensor_id] = SensorObservation(
    depth=depth_data,
    rgba=rgba_data,
  )
```

src/tbp/monty/simulators/mujoco/agents.py
```diff
@@ -218,10 +218,31 @@ def observations(self) -> AgentObservations:
...
+                renderer.enable_segmentation_rendering()
+                semantic_data = renderer.render()
+                renderer.disable_segmentation_rendering()
```
tristanls-tbp: **note**: This is starting to smell... I think if we add one more modality, this whole method will need to be refactored into something more structured.

tests/unit/simulators/mujoco/agent_actions_test.py
tristanls-tbp: **suggestion**: How do you feel about having `default_agent_args` accept agent id and patch sensor id, something like:

```python
def default_agent_args(
    agent_id: AgentID = AgentID("agent_id_0"), 
    patch_sensor_id: SensorID = SensorID("patch"),
) -> dict[str, Any]:
    # ...
```

Here are my reasons:
- I feel uncomfortable having to import `AGENT_ID` from somewhere else when it is straightforward to declare in the test file I'm writing/reading.
- I think the only reason we import `AGENT_ID` and `VIEW_FINDER_SENSOR_ID` is because those are hardcoded in `default_agent_args()`. Having `default_agent_args()` accept params would mitigate that.

With the `default_agent_args()`, one could use them if one does not care about agent id or patch sensor id via:
```python
            agents=[
                partial(
                    cls.agent_class,
                    **default_agent_args(),
                )
            ]
```
whereas, if we actually care for test reasons, then we have the option:
```python
AGENT_ID = AgentID("potato")
PATCH_SENSOR_ID = SensorID("tomato")

# ...

agents=[
  partial(
    DistantAgent,
    **default_agent_args(AGENT_ID, PATCH_SENSOR_ID),
  )
]
```
As to where... how about `tests/agents.py`:

```python
from tests.agents import default_agent_args
```

tests/unit/simulators/mujoco/noop_agent_test.py
```diff
@@ -96,6 +99,72 @@ def test_agent_observation_single_sensor(self) -> None:
...
             assert rgba.max() == 255
 
+    def test_agent_observation_semantic_default_ids(self) -> None:
+        """Test that the semantic sensor default ids match expectations.
+
+        We want the semantic sensor to behave similar to how the Habitat version does
+        so we need to confirm that the values returned map correctly.
+        """
+        agent_args = default_agent_args()
+        agent_args.update(
```
tristanls-tbp: **suggestion (non-blocking)**: For my taste, this `.update(...` of `default_agent_args()` is more code than writing out args directly where they're used:

```python
sim = MuJoCoSimulator(
    agents=[
        partial(
            NoopAgent,
            agent_id=AGENT_ID,
            sensor_configs={
                PATCH_SENSOR_ID: SensorConfig(
                    position=ZERO_VECTOR,
                    rotation=IDENTITY_QUATERNION,
                    resolution=DEFAULT_RESOLUTION,
                    zoom=1.0,
                    semantic=True,
                )
            }
        )
    ]
)
```

tests/unit/simulators/mujoco/noop_agent_test.py
```diff
@@ -96,6 +99,72 @@ def test_agent_observation_single_sensor(self) -> None:
...
+            assert unique_ids == {0, mjtGeom.mjGEOM_BOX, mjtGeom.mjGEOM_SPHERE}
+
+    def test_agent_observation_semantic_custom_ids(self) -> None:
+        """Test that the semantic sensor with custom ids match expectations.
+
+        If we give semantic ids to `add_object` we want to make sure that the
+        returned semantic sensor values use those IDs and not the defaults.
+        """
+        agent_args = default_agent_args()
+        agent_args.update(
```
tristanls-tbp: **suggestion (non-blocking)**: Same this-is-more-code-than-direct-args-in-constructor comment as in the other test.

tests/unit/simulators/mujoco/noop_agent_test.py
```diff
@@ -96,6 +99,72 @@ def test_agent_observation_single_sensor(self) -> None:
             assert rgba.min() == 0
             assert rgba.max() == 255
 
+    def test_agent_observation_semantic_default_ids(self) -> None:
```
tristanls-tbp, line 92: **note**: This is nicely written, and it looks like this could easily become a property test across all default IDs.

tristanls-tbp (commented):
Thank you for answering the questions. I have one major suggestion for your consideration.

src/tbp/monty/simulators/mujoco/agents.py
jeremyshoemaker: Yep, I'm not sure why this didn't occur to me. I usually prefer early return.

src/tbp/monty/simulators/mujoco/agents.py
jeremyshoemaker: I agree. Definitely something to keep an eye on.

tests/unit/simulators/mujoco/noop_agent_test.py
jeremyshoemaker: I didn't realize it was that similar.

tests/unit/simulators/mujoco/agent_actions_test.py
jeremyshoemaker: I ended up refactoring `default_agent_args` out of existence. When I stopped using it as the basis for `agent_args`, it ended up only being called in 3 places, so I inlined it. Then I realized that I wasn't using the defaults of `SensorConfig` or the agent's position and rotation, so that shortened it more.

src/tbp/monty/simulators/mujoco/agents.py
jeremyshoemaker: It's a loop, so we can't exit early, but we can continue early.

tests/unit/simulators/mujoco/noop_agent_test.py
jeremyshoemaker: I thought about this quite a bit. The problem is that when we're adding objects, we're using names like "box" or "sphere" and when we're checking the results, we're using MuJoCo's enum for the various types which maps to numbers. To make this a property test, I'd need to add a mapping between those two to pull values from. 

Also, at that point we'd be testing more of the internals of MuJoCo. Once we test with a single pair of objects and get the results we expect, I think we should be able to trust that MuJoCo will do the rest of them correctly. Otherwise, the problem is with them.

tristanls-tbp (approved):
Thank you for the updates.

PR #1069, opened 2026-06-29
refactor!: remove Habitat importskip from burst_sampling_test.py

jeremyshoemaker:

This test doesn't use any Habitat features, so it doesn't need to have the `pytest.importskip`. I think it used to build a full experiment and run it, but during the Implementation Project that added it, that was removed and it became more of a unit test suite that doesn't use a simulator.

The majority of changes here are type annotations. One in particular came up with the `BurstSamplingHypothesesUpdater` constructor. The `initial_possible_poses` type was `Literal["uniform", "informed"] | list[Rotation]` but every call site and configuration that creates instances of this class pass it a list of numbers. In fact, the call to `get_initial_possible_poses` in the constructor **assumes** that its argument is **not** a `Rotation` object and tries to create one from the angles passed in, which fails if it's already a `Rotation`. I changed the type to match what the values actually are.

> [!warning]
> **Breaking Changes**
> - The `initial_possible_poses` argument to `BurstSamplingHypothesesUpdater` has been changed to match its actual usage.


tests/unit/frameworks/models/evidence_matching/burst_sampling_test.py
```diff
@@ -119,8 +110,9 @@ def test_update_hypotheses_ids_map_correctly(self) -> None:
             evidence_update_threshold=0,
         )
 
-        assert_array_equal(result.possible, np.array([True, False]))
-        assert_array_equal(result.evidence, np.array([1, 2]))
+        assert result is not None  # for type narrowing
+        np.testing.assert_array_equal(result.possible, np.array([True, False]))
```
jeremyshoemaker, line 114: This change happened because `assert_array_equal` from `numpy.ma.testutils` is untyped, and probably the wrong place to import this from.

tristanls-tbp:
**note**: Marking this as a breaking change refactor since the type of an argument in the src/tbp/monty code has changed.

tests/unit/frameworks/models/evidence_matching/burst_sampling_test.py
tristanls-tbp: **praise**: Nice update, thank you.

tristanls-tbp (approved):
Thank you, looks like an improvement all around.

PR #1070, opened 2026-06-29
refactor!: compositional benchmark configs

hlee9212:
This PR consolidates the compositional "logos on objects" benchmark. 

> [!warning]
> **Breaking Changes**
> - removes `infer_comp_*` experiment configurations.
> - removes/replaces the old compositional pretraining configuration names.

I have verified that the new `base_infer_objects_with_stickers_comp_models` have the exact same performance as the old `infer_comp_lvl3_with_comp_models` after running pretraining & inference in AWS. 

CleanShot 2026-06-29 at 13 19 18

(Question to @ramyamounir : I wasn't sure if `tbp.benchmarks` could handle name change from `infer_comp_lvl3_with_comp_models` to `base_infer_objects_with_stickers_comp_models`?)

(Also tagging/adding Ramy into Reviewers as Niels suggested)


benchmarks/logos_on_objects.csv
```diff
@@ -1,5 +1,2 @@
 Experiment,Correct Child or Parent (%)|align right,Used MLH (%)|align right,Num Match Steps|align right,Rotation Error (degrees)|align right,Avg Prediction Error|align right,Run Time (mins)|align right,Num Episodes|hidden
-infer_comp_lvl1_with_monolithic_models,89.29,84.52,415,56.28,0.25,38,84
-infer_comp_lvl1_with_comp_models,86.90,29.76,34,58.33,0.29,4,84
-infer_comp_lvl2_with_comp_models,86.67,32.38,40,42.34,0.29,12,210
-infer_comp_lvl3_with_comp_models,72.00,42.57,34,49.35,0.30,17,350
+base_infer_objects_with_stickers_comp_models,72.00,42.57,34,49.35,0.30,19,350
```
nielsleadholm, line 3: Please also update the benchmark to include the `base_infer_objects_with_stickers_monolithic_models`.

We should hopefully still see a big difference in the number of Used MLH and Match Steps vs. the compositional Monty.

tests/conf/snapshots/supervised_pre_training_objects_with_stickers_monolithic_models.yaml
```diff
@@ -7,7 +7,7 @@ experiment:
       monty_args:
         num_exploratory_steps: 1000
         min_eval_steps: 3
-        min_train_steps: 100
+        min_train_steps: 3
```
nielsleadholm, line 10: What was the reason for this changing?

nielsleadholm (commented):
Nice, looks great. Just a few minor comments.

nielsleadholm:
@ramyamounir re. reviewing - please don't worry about that in this PR (although feel free to take a quick look to familiarize yourself with the changes) - it's when we add in the 2D SM and burst sampling that I would appreciate a second review from yourself. 

But useful to get your thoughts on Hojae's question about `tbp.benchmarks`

ramyamounir (approved):
Looks great! I really like those changes, makes it a lot easier to understand.

`tbp.benchmarks` cannot handle this change because it matches on the experiment name between sources (repo, wandb). You could rename the wandb experiment name in the CSV file if you want to run `tbp.benchmarks` but I don't think this is necessary for just one experiment.

I assume we cannot show that the monolithic inference is the same because we changed monolithic from lvl1 to lvl3, correct? Just thought I'd point it out for others because it wasn't clear to me at first.

Questions:
- Did we remove the burst sampling configs because we are planning to add it to all the new compositional/monolithic configs in a follow-up PR?
- Will there be a follow-up PR to update the pretrained compositional models to V4?

tests/conf/snapshots/supervised_pre_training_objects_with_stickers_monolithic_models.yaml
ramyamounir: I think git is incorrectly doing a diff between lvl2 and monolithic (probably because the new monolithic snapshot is closer to the object list of lvl2 snapshot than the lvl1 original monolithic snapshot). The original monolithic configs has `min_train_steps: 3`.

Image

nielsleadholm: Nice, thanks!

nielsleadholm:
Thanks Ramy:

- "I assume we cannot show that the monolithic inference is the same because we changed monolithic from lvl1 to lvl3, correct?" --> yes that's correct
- "Did we remove the burst sampling configs because we are planning to add it to all the new compositional/monolithic configs in a follow-up PR?" --> yes exactly, I thought it was easiest to tie the refactor to an existing benchmark result (the level 3 comp), but yes we'll update all of them to use bursting in a follow-up PR.
- Good catch about updating the comp pretraining models to v4. I think that would be worth doing in this PR @hlee9212 .

jeremyshoemaker:
Marking this as a breaking change because we're removing configurations.

@hlee9212 do you mind adding a section to the description about it, to help us with the release notes when it comes up in whatever release this ends up in?

I've been using this formatting lately, which makes it stand out nicely. Feel free to rephrase or add to it as needed.

```markdown
> [!warning]
> **Breaking Changes**
> - removes `infer_comp_*` experiment configurations.
```

benchmarks/logos_on_objects.csv
hlee9212: I added the line for `base_infer_objects_with_stickers_monolithic_models`.

hlee9212:
Hi @nielsleadholm, I made the couple of changes requested:

1. I updated `logos_on_objects.csv` to include the `base_infer_objects_with_stickers_monolithic_models` results.
2. I copied the models within our AWS and also uploaded to our bucket and updated the link in our docs to `pretrained_compositional_objects_v4`.

nielsleadholm:
Great thanks, when you say "copied the models", these are the newly trained models correct? (note that the monolithic model is different from what it was in version 3).

hlee9212:
> Great thanks, when you say "copied the models", these are the newly trained models correct? (note that the monolithic model is different from what it was in version 3).

Yes, I meant I copied the newly trained models from /home/hlee to /mnt/results/TBP

nielsleadholm (approved):
Great work!

PR #1065, opened 2026-06-26
chore: updating documentation to explain how we validate slug and file path accuracy

brothman-tbp:
I'll be able to remove the   from this section in the next PR, once the new constants to ignore are merged.

```markdown Markdown
# category-slug: Category Title
- [document-slug](category-slug/document-slug.md) <!-- check-links-ignore --> 
  - [subdocument-slug](category-slug/document-slug/subdocument-slug.md) <!-- check-links-ignore --> 
```

I have to leave the  for the tutorial path example, because those are a real pathway, and we don't want to ignore the actual docs on that pathway.

```markdown Markdown
# how-to-use-monty: How to Use Monty
- [tutorials](how-to-use-monty/tutorials.md) <!-- check-links-ignore --> 
  - [running-your-first-experiment](how-to-use-monty/tutorials/running-your-first-experiment.md) <!-- check-links-ignore -->
```

PR #1071, opened 2026-06-29
docs: remove unnecessary ignore link checker to clean up documentation

brothman-tbp:
In the last PR (#1065), I had to include the ignore link checker for the "document-slug" because I needed the update to the constants.py to be merged first (https://github.com/thousandbrainsproject/tbp.monty/pull/1065/changes#diff-9bf9ff52bb73ac6fb8db51704efe020ae281fbc1ea819d762aa21bd8e008a576).

Now that constant.py ignores "document-slug", I can remove the   

I have to leave    for the [tutorials](how-to-use-monty/tutorials.md ) section, because that's a real pathway that would be expressed in hierarchy.md, that I wanted to show for educational purposes and we don't want to ignore that pathway when checking the real docs.

### DIFF
--- a/docs/contributing/documentation.md
+++ b/docs/contributing/documentation.md
@@ -84,8 +84,8 @@ Each entry has the following format:
 
 ```markdown Markdown
 # category-slug: Category Title
-- [document-slug](category-slug/document-slug.md) <!-- check-links-ignore --> 
-  - [subdocument-slug](category-slug/document-slug/subdocument-slug.md) <!-- check-links-ignore --> 
+- [document-slug](category-slug/document-slug.md) 
+  - [subdocument-slug](category-slug/document-slug/subdocument-slug.md) 
 ```
 
 There are two important parts:


vkakerbeck:
Could you give a bit more context? Why is this unnecessary now? And if it is, do we still need the check for "check-links-ignore" in the readme sync tool?

jeremyshoemaker:
> Could you give a bit more context?

@brothman-tbp, from a triage perspective, this extra context probably should be in the description of this PR. It's a small change here, but it's not clear _why_ this change is being made. Having that context up front makes it easier to review.

brothman-tbp:
> > Could you give a bit more context?
> 
> @brothman-tbp, from a triage perspective, this extra context probably should be in the description of this PR. It's a small change here, but it's not clear _why_ this change is being made. Having that context up front makes it easier to review.

@jeremyshoemaker Thank you for explaining. That makes sense. I updated the description of this PR to give the context. 

@vkakerbeck  I updated the description with the context (like we discussed offline), and why we can remove "check-links-ignore" for the document-slug example but still need it for the the tutorial example.

vkakerbeck (approved):
Thanks for the explanation (in our meeting and PR description), that all makes sense and looks good

PR #1073, opened 2026-06-30
fix: pop step_hook from config to avoid error when saving live plotter

pablogarciaamolina:
A configured step_hook is currently stored on self.config in MontyExperiment. When save_state_dir runs at the end of an episode, that config is serialized with torch.save. When using tbp.teleop, the hook holds an unpicklable object, so the save raises `TypeError: cannot pickle 'FigureCanvasQTAgg' object`, causing the run to crash.

As @ramyamounir suggested, this changes the hook to be popped out of self.config instead, so the live hook is held only as self._step_hook and never reaches the serialized config.


tristanls-tbp (approved):
Thanks!

PR #1074, opened 2026-06-30
test: create MuJoCo version of conf_test.py

jeremyshoemaker:

We're also updating the `update_snapshots.py` script to support generating either Habitat (the default) or MuJoCo snapshots. This involved changes to how we handle removed or renamed configs so that running it for one simulator's snapshots won't delete the snapshots for the other simulator.

Minimal differences to the MuJoCo version of the test.
```diff
10,18d9
< import pytest
<
< from tests import HYDRA_ROOT
<
< pytest.importorskip(
<     "habitat_sim",
<     reason="Habitat Sim optional dependency not installed.",
< )
<
24a16,17
> from tests import HYDRA_ROOT
>
26c19
<     Path(__file__).parents[2] / "src" / "tbp" / "monty" / "conf" / "experiment"
---
>     Path(__file__).parents[3] / "src" / "tbp" / "monty" / "conf" / "experiment"
31,33c24,26
<     # Exclude MuJoCo experiments
<     # TODO: Revert once we convert to MuJoCo
<     if not x.stem.endswith("mujoco")
---
>     # Only check MuJoCo experiments
>     # TODO: Merge with original once we convert to MuJoCo
>     if x.stem.endswith("mujoco")
35c28
< EXPERIMENT_SNAPSHOTS_DIR = Path(__file__).parent / "snapshots"
---
> EXPERIMENT_SNAPSHOTS_DIR = Path(__file__).parents[2] / "conf" / "snapshots"
41,43c34,36
<     # Exclude MuJoCo tutorials
<     # TODO: Revert once we convert to MuJoCo
<     if not x.stem.endswith("mujoco")
---
>     # Only check MuJoCo tutorials
>     # TODO: Merge with original once we convert to MuJoCo
>     if x.stem.endswith("mujoco")
45c38
< TUTORIAL_SNAPSHOTS_DIR = Path(__file__).parent / "snapshots" / "tutorial"
---
> TUTORIAL_SNAPSHOTS_DIR = EXPERIMENT_SNAPSHOTS_DIR / "tutorial"
```


src/tbp/monty/conf/experiment/base_config_10distinctobj_noop_agent_mujoco.yaml
```diff
@@ -2,7 +2,7 @@
 
 defaults:
   - /monty: evidencegraph_exp1000_emin_t3_tot2500
-  - /monty/motor_system_config: informed_5_goal1
+  - /monty/motor_system_config: distant_5
```
jeremyshoemaker, line 5: This is a change I already noticed elsewhere, but this test is forcing me to fix it here.

jeremyshoemaker:
I just noticed this test doesn't catch when you haven't generated a snapshot for a config, because it uses the snapshots to decide what to check, not the source configs. Was that a conscious choice?

I noticed because I generated snapshots but forgot to add them the PR but the `uv` tests pass.

jeremyshoemaker:
> I just noticed this test doesn't catch when you haven't generated a snapshot for a config, because it uses the snapshots to decide what to check, not the source configs. Was that a conscious choice?

Per our discussions, I updated the tests to fail if the snapshot file is missing. The previous behaviour was generating them if they were missing.

tests/conf/conf_test.py
```diff
@@ -74,17 +74,15 @@ def test_experiment(self, experiment: str):
...
             except FileNotFoundError:
-                snapshot_config_yaml = None
-            if snapshot_config_yaml is not None:
-                _assert_config_matches_snapshot(
-                    current_config_yaml, snapshot_config_yaml, experiment
-                )
-            else:
-                with snapshot_path.open("w") as f:
-                    f.write(current_config_yaml)
+                pytest.fail(f"Missing snapshot file for '{experiment}'")
```
jeremyshoemaker: Rather than rely on the `FileNotFoundError` to tell the user that the file is missing, I'm using `fail` to give a more useful message about what the problem is.

tests/conf/conf_test.py
```diff
@@ -42,7 +42,7 @@
     # TODO: Revert once we convert to MuJoCo
     if not x.stem.endswith("mujoco")
 ]
-TUTORIAL_SNAPSHOTS_DIR = Path(__file__).parent / "snapshots" / "tutorial"
+TUTORIAL_SNAPSHOTS_DIR = EXPERIMENT_SNAPSHOTS_DIR / "tutorial"
```
jeremyshoemaker, line 45: This is a change I made in the MuJoCo version that I'm backporting to the original since I'm making other changes here now.

src/tbp/monty/conf/update_snapshots.py
```diff
@@ -38,18 +40,21 @@ def update_snapshots(
         override_prefix: Prefix for the override value
             (e.g. "tutorial/" or "evidence_lm/").
         snapshots_dir: The directory to write the snapshots to.
+        generate_mujoco: Whether to we're generating MuJoCo or Habitat snapshots.
```
tristanls-tbp:
```suggestion
        generate_mujoco: Whether we're generating MuJoCo or Habitat snapshots.
```

src/tbp/monty/conf/update_snapshots.py
tristanls-tbp, line 1: **todo**: Please update the documentation that snapshot errors point to in `tests/conf/README.md`

tristanls-tbp (approved):
I'm approving conditional on the **todo**.

PR #1075, opened 2026-06-30
test: add zoom test for MuJoCo

jeremyshoemaker:

Looking at the HabitatSim integration and unit tests, they seem to cover much of what we cover in the MuJoCo unit tests, except they're for HabitatSim.

One thing they test that we aren't currently testing in MuJoCo is the zoom functionality. Since we're implementing that ourselves in Embodiment by modifying the FOV of the camera, we should add unit tests to confirm that it correctly scales the resulting images.

The approach we're using is a property test based on the percentage of the image covered by the object. Because of pixelation of the images, it's hard to get an actual value consistently close enough to the calculated expected value.


tests/unit/simulators/mujoco/noop_agent_test.py
```diff
@@ -193,3 +195,67 @@ def test_agent_observation_multiple_resolutions(self) -> None:
...
+            semantic = obs[PATCH_SENSOR_ID]["semantic"]
+
+            covered_pixels = semantic.sum()
+            actual_ratio = covered_pixels / total_pixels
+            # In one case, the object becomes large enough that it takes up the
+            # entire image and the expected ratio ends up larger than 1.0, so we
+            # need to clamp it to 1.0.
+            expected_ratio = min(
+                1.0,
+                (zoom_1x_covered_pixels / total_pixels) * zoom**2,
```
tristanls-tbp: **nitpick**: While mathematically true, from my perspective it would be clearer as:
```python
(zoom_1x_covered_pixels * zoom**2) / total_pixels
```
I found the fact that `total_pixels` was being multiplied somewhat confusing.

tests/unit/simulators/mujoco/noop_agent_test.py
```diff
@@ -193,3 +195,67 @@ def test_agent_observation_multiple_resolutions(self) -> None:
...
             )
             assert view_finder_rgba.shape == (256, 256, 4)
+
+    @given(
+        # Using integers because arbitrary floats make the calculated ratio
+        # too imprecise for a testable property due to the pixelation.
+        zoom=st.integers(min_value=1, max_value=10),
+    )
+    @settings(deadline=None)
+    def test_agent_observation_zoom(self, zoom) -> None:
```
tristanls-tbp, line 205: **note**: Another property that comes to mind, instead of checking for a specific expected value, is to do a monotonicity test. i.e., pixel coverage at `zoom=1` < pixel coverage at `zoom=2` < pixel coverage at `zoom=3`, and so on.

pseudocode:
```
zoom = 1
zoom_delta = 1
previous_coverage = 0
while zoom_1x_covered_pixels / total_pixels < 1.0:
    coverage = current_coverage(zoom)
    assert coverage > previous_coverage
    previous_coverage = coverage
    zoom += zoom_delta
```

tristanls-tbp (approved):
Thanks, looks good.

tests/unit/simulators/mujoco/noop_agent_test.py
jeremyshoemaker: That test wouldn't catch the bug I had in the original FOV zoom calculation where I was doing `base_fov / zoom`. If I change the code back to that and run this test as written, it catches it immediately when `zoom=2` where the expectation is a coverage of 4.9% instead of the 7.2% we end up with.

tests/unit/simulators/mujoco/noop_agent_test.py
jeremyshoemaker: You're right, that's clearer and gets rid of the need for parentheses.

I think I was thinking of $(\frac{a}{b}) \cdot z^2$ instead of $\frac{a \cdot z^2}{b}$

PR #1076, opened 2026-07-01
docs: Additional future work page updates

vkakerbeck:
This PR includes the remaining future work page updates I wanted to make (all except LM and SM related ones made in a previous PR).

PR #1080, opened 2026-07-02
docs: Update future work tags

vkakerbeck:
This PR mostly updates the `monty` skills tag to distinguish between `monty-beginner` and `monty-advanced`. A corresponding explanation is added to the docs as well (see copy below). While going through I also fixed a few minor other metadata fields, such as missing status. Finally, I removed the `monty` tag for the skills snippet.

> **monty-beginner** are tasks that requires understanding one component of Monty to be able to modify it and experiment with it. They also require knowing how to run experiments with Monty and update experiment configs. These are good tasks to get started with Monty and get an initial understanding of how Monty works. Having a monty-beginner task tag doesn't mean that this item is easy or small in scope. It only indicates how much understanding of Monty's components and the system as a whole is required.
>
> **monty-advanced** are tasks that require an understanding of Monty as a whole. They require understanding several Monty components and how they interact with each other. Some tasks might not involve changing several Monty components, but they require an in-depth understanding of one (such as the algorithm within LMs) and how changes to it affect the whole system. They also require an understanding of Monty's design philosophy and the theory behind it to know how to make larger structural changes to Monty. We recommend to start with a beginner task to learn about Monty and to stay in communication with our team for the advanced tasks.

Note: I built this PR off of this one: https://github.com/thousandbrainsproject/tbp.monty/pull/1076 to include the added pages there so best to review after that one is merged.


docs/contributing/documentation/future-work-widget-metadata.md
```diff
@@ -83,7 +83,9 @@ Skills is a comma separated list of skills that will be needed to complete this.
 > [!NOTE] Notes on Some of the Fields
 > **research** means the ability to run controlled experiments, interpret results, investigate unexpected behavior, come up with novel solutions, draw from existing literature, and communicate the results
 > 
-> **monty** refers to knowledge of how Monty works, how it is structured, and usually more detailed knowledge of the part of its algorithm that the specific future work item touches. This is usually required for tasks that involve innovative changes to the existing Monty algorithm or larger structural changes to the code base.
+> **monty-beginner** are tasks that requires understanding one component of Monty to be able to modify it and experiment with it. They also require knowing how to run experiments with Monty and update experiment configs. These are good tasks to get started with Monty and get an initial understanding of how Monty works. Having a monty-beginner task tag doesn't mean that this item is easy or small in scope. It only indicates how much understanding of Monty's components and the system as a whole is required.
```
nielsleadholm:
```suggestion
> **monty-beginner** are tasks that require understanding one component of Monty to be able to modify it and experiment with it. They also require knowing how to run experiments with Monty and update experiment configs. These are good tasks to get started with Monty and get an initial understanding of how Monty works. Having a monty-beginner task tag doesn't mean that this item is easy or small in scope. It only indicates how much understanding of Monty's components and the system as a whole is required.
```

docs/contributing/documentation/future-work-widget-metadata.md
```diff
@@ -83,7 +83,9 @@ Skills is a comma separated list of skills that will be needed to complete this.
...
+>
+> **monty-advanced** are tasks that require an understanding of Monty as a whole. They require understanding several Monty components and how they interact with each other. Some tasks might not involve changing several Monty components, but they require an in-depth understanding of one (such as the algorithm within LMs) and how changes to it affect the whole system. They also require an understanding of Monty's design philosophy and the theory behind it to know how to make larger structural changes to Monty. We recommend to start with a beginner task to learn about Monty and to stay in communication with our team for the advanced tasks.
```
nielsleadholm:
```suggestion
> **monty-advanced** are tasks that require an understanding of Monty as a whole. They require understanding several Monty components and how they interact with each other. Some tasks might not involve changing several Monty components, but they require an in-depth understanding of one (such as the algorithm within LMs) and how changes to it affect the whole system. They also require an understanding of Monty's design philosophy and the theory behind it to know how to make larger structural changes to Monty. We recommend starting with a beginner task to learn about Monty and to stay in communication with our team for the advanced tasks.
```

docs/future-work/framework-improvements/make-configs-easier-to-use.md
```diff
@@ -5,7 +5,7 @@ rfc: optional
 estimated-scope: large
 improved-metric: infrastructure, learning-experience, platform
 output-type: PR
-skills: python, refactoring, monty
+skills: python, refactoring
```
nielsleadholm: Was this intentional? Wouldn't it be good to understand Monty to make the configs easier? I feel like there's a risk that someone will make non-helpful improvements to the configs if they don't understand the underlying system (e.g. that Monty is meant to be separate from the experimental framework, or that the MotorSystem is currently shared by all LMs).

docs/future-work/learning-module-improvements/contributing-learning-modules.md
```diff
@@ -1,5 +1,12 @@
 ---
 title: Contributing Learning Modules
+description: Test different algorithms or representations inside of LMs.
+rfc: optional
+estimated-scope: unknown
+improved-metric: speed, pose, accuracy, noise
+output-type: experiments, analysis, PR
+skills: python, research, monty-beginner
```
nielsleadholm: I think I'd suggest monty-advanced for this one, since a new LM will have significant impacts throughout the system, and someone is likely to come up with a poor custom LM if they don't deeply understand all of Monty's components.

docs/future-work/learning-module-improvements/support-scale-invariance.md
```diff
@@ -5,7 +5,7 @@ rfc: required
 estimated-scope: large
 improved-metric: scale
 output-type: theory, prototype, experiments, analysis
-skills: python, research, monty, literature-review
+skills: python, research, monty-beginner, literature-review
```
nielsleadholm: I think I'd also suggest monty-advanced for this, since it will likely require an intersection with things like the SM and motor system (e.g. transforming the size of displacements coming in and out of the system, and using model-free signals from the SM to estimate scale).

nielsleadholm (commented):
Nice, this is a helpful change. Just a few minor suggestions.

docs/future-work/framework-improvements/make-configs-easier-to-use.md
vkakerbeck: Is it possible to make these changes without refactoring Monty itself? I can add it back in

docs/future-work/learning-module-improvements/support-scale-invariance.md
vkakerbeck: The way this task is described it only involves changes to the internals of the LM. I.e. hypotheses about scale. The caling of the displacement by hypothesis happens inside the LM. And the hypotheses aren't informed by model-free signals, they are just hypotheses like rotation. Also inductive biases about scale would be stored inside the LM and informing it by edges in the graph would also be inside the LM.

But since it would require a deep understanding of the LM algorithm I could still mark it as advanced

docs/future-work/framework-improvements/make-configs-easier-to-use.md
vkakerbeck: I guess I wouldn't be sure whether to make it beginner or advanced if adding back. Depending on what type of config change is made, it could require understanding one, many or no components of Monty. I don't want to scare people away from this item by marking it advanced, when there are also basic improvements that could be made.

nielsleadholm: Hmm right, could we put both? And then in the description write something kind of like we both wrote, e.g.:

"Depending on what type of config change you want to make, it may require understanding one, many or no components of Monty. For example, we are moving towards separating Monty (as a sensorimotor system) from the experimental framework, and the `MotorSystem` is currently shared by all LMs - factors that would influence the structuring of configs. On the other hand, you may identify a simple change that would make the configs easier to understand. To reflect this, we include the tags `monty-begginer` and `monty-advanced` in the description."

But yeah also happy to just leave it.

docs/future-work/learning-module-improvements/support-scale-invariance.md
nielsleadholm: Yeah personally I think that would be a helpful heads-up to someone working on this.

For example, I interpret the following as mostly being about model-free methods (and thus at least related to SM processing):
- "Heuristics based on low-level sensory input (e.g. inferred distance) that are used to rapidly propose the most probable scales."
- ... "if an enormous coffee mug was on display in an art installation, the inferred distance from perceived depth, together with the size of eye movements, could suggest that - whatever the object - features are separated on the scale of meters. This low-level information would inform testing objects on a large scale"). 

The above could be implemented via the hybrid edge matching (but that requires storing and matching those displacements), or it could be a pure model-free signal that tells the LM the range of scales that is appropriate for the object we're inferring (e.g. movement was 10cm - assuming that was within an object, which is a reasonable model-free assumption, then object scale is probably in the range of 10-50cm in size). 

The other thing I was alluding to is something we don't currently discuss in the text itself, but maybe we should add - I think hypothesis-testing and other model-based LM policies may break with scale changes unless the scale is accounted for in the output. If the mug is huge relative to the learned model, the LM needs to make sure the output reflects this. 

My point is I feel like there is a reasonable amount of non-trivial interaction between the LM and SMs and motor-system for scale to work as expected. If you don't understand how the SM plays a role in the measured displacements, or how the MotorSystem acts on goals, then there is a good chance that things will start acting strange.

vkakerbeck: okay, I updated it to advanced

docs/future-work/framework-improvements/make-configs-easier-to-use.md
vkakerbeck: good idea, I added both

nielsleadholm (approved):
LGTM

PR #1083, opened 2026-07-04
feat: unsupervised wandb logging

haoyucn:
## Summary

Adds WandB logging for unsupervised learning benchmark metrics currently reported by `print_unsupervised_stats`.

Logged metrics include:

- `unsupervised/correct_first_epoch_percent`
- `unsupervised/correct_after_first_epoch_percent`
- `unsupervised/mean_objects_per_graph`
- `unsupervised/mean_graphs_per_object`
- `unsupervised/run_time_minutes`
- `unsupervised/episode_run_time_seconds`

## Notes

This only reads existing BASIC logging data inside the WandB handler and does not change learning behavior, matching logic, graph construction, or experiment logic.

## Validation

- Ran `ruff format` on `wandb_handlers.py`
- Ran `ruff check` on `wandb_handlers.py`
- Ran a lightweight mocked WandB handler test using synthetic BASIC logging data

I was not able to run the full Habitat-based tutorial locally because `habitat_sim` is not installed in my Windows environment.


tristanls-tbp:
Hi @haoyucn, thank you for the pull request.

Please do not include your RFC proposal in this pull request. RFC proposals follow a separate process, and we won't be able to merge this pull request with the RFC proposal included.

tristanls-tbp:
I requested your review, @vkakerbeck, to confirm this is an appropriate set of statistics to send to WandB. Please reassign to someone else if needed.

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -239,7 +242,133 @@ def report_episode(
         basic_logs = data["BASIC"]
         mode_key = f"{mode}_overall_stats"
         stats = basic_logs.get(mode_key, {})
-        wandb.log(stats[episode], step=episode, commit=False)
+
+        wandb_stats = dict(stats[episode])
+        wandb_stats.update(
+            self.get_unsupervised_benchmark_stats(basic_logs, episode, mode)
+        )
```
vkakerbeck: Does this work even if this is not an unsupervised learning experiment? (i.e. if we are just evaluating or doing supervised training). What happens in those cases? It seems like this function should only be called if we are running the unsupervised learning experiments.

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -239,7 +242,133 @@ def report_episode(
...
+
+        wandb.log(wandb_stats, step=episode, commit=False)
+
+    def get_unsupervised_benchmark_stats(self, basic_logs, episode, mode):
+        """Log benchmark-style unsupervised learning stats to WandB.
+
+        These stats mirror the metrics reported by print_unsupervised_stats:
+        first-epoch correctness, post-first-epoch correctness, graph/object
+        association stats, runtime, and average episode runtime.
+
```
vkakerbeck: runtimes are already logged to wandb for all experiments. They don't need to be logged here again

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -239,7 +242,133 @@ def report_episode(
...
+        This method only reads existing BASIC logging data and does not change
+        experiment behavior.
```
vkakerbeck: I don't think this needs to be stated here. It shoul be assumed that a logger doesn't change experiment behavior

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -239,7 +242,133 @@ def report_episode(
...
+            "time",
+            "lm_id",
+            "episode",
+        }
+        if not required_columns.issubset(stats_df.columns):
+            return {}
+
+        # The current unsupervised benchmarks use one LM. If multiple LMs are present,
+        # use LM_0 to avoid counting one episode multiple times.
+        lm_stats = stats_df[stats_df["lm_id"] == "LM_0"].copy()
```
vkakerbeck: Other wandb logging takes into account multiple LMs already so it might lead to unexpected results if this one just looks at the first one without telling the user. Is it not possible to calculate per-lm stats like in the evaluation stats? Or otherwise the column names should indicate that this statistic is only for LM 0

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -239,7 +242,133 @@ def report_episode(
...
+
+        epoch_len = self.infer_epoch_len(lm_stats)
+        if epoch_len is not None:
+            first_epoch_stats = lm_stats.iloc[:epoch_len]
+            later_epoch_stats = lm_stats.iloc[epoch_len:]
+
+            first_epoch_new = first_epoch_stats[
+                first_epoch_stats["primary_performance"] == "no_match"
+            ]
+            unsupervised_stats["unsupervised/correct_first_epoch_percent"] = (
```
vkakerbeck: To be consistent with other logging column names, we should call this `percent_correct_first_epoch`

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -239,7 +242,133 @@ def report_episode(
...
+                len(first_epoch_new) / len(first_epoch_stats) * 100
+            )
+
+            if len(later_epoch_stats) > 0:
+                later_correct = later_epoch_stats[
+                    later_epoch_stats["primary_performance"].isin(
+                        ["correct", "correct_mlh"]
+                    )
+                ]
+                unsupervised_stats["unsupervised/correct_after_first_epoch_percent"] = (
```
vkakerbeck: To be consistent with other logging column names, we should call this `percent_correct_after_first_epoch`

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -239,7 +242,133 @@ def report_episode(
...
+                    len(later_correct) / len(later_epoch_stats) * 100
+                )
+
+        run_time_seconds = pd.to_numeric(lm_stats["time"]).sum()
+        unsupervised_stats["unsupervised/run_time_minutes"] = run_time_seconds / 60
+        unsupervised_stats["unsupervised/episode_run_time_seconds"] = (
+            run_time_seconds / len(lm_stats)
+        )
```
vkakerbeck: This is not specific to unsupervised learning and should already be logged by default

src/tbp/monty/frameworks/loggers/wandb_handlers.py
```diff
@@ -239,7 +242,133 @@ def report_episode(
...
+
+        return unsupervised_stats
+
+    def infer_epoch_len(self, stats):
+        """Infer epoch length from the first repeated target object.
```
vkakerbeck: This seems like a pretty specific assumption. It holds for our current unsupervised learning experiments and I am not sure how to better get the epoch length here, but I think if we use this, we should make sure that these stats are only added to wandb if we are running those specific unsupervised learning experiments (using the `Predefined` `object_init_sampler`)

vkakerbeck (commented):
Thank you for making this PR @haoyucn This will be very useful to have.

I had a couple of comments around the logged variable names and when they are being logged.

@tristanls-tbp I didn't do a thorough review of the logger integration itself and mostly focused on the high-level data being logged.

I don't think this PR should include the RFC file (I don't think we need an RFC for this in general, but if we wanted to have one it should be a separate PR, which goes through our [RFC process](https://docs.thousandbrains.org/docs/request-for-comments-rfc))

src/tbp/monty/frameworks/loggers/wandb_handlers.py
haoyucn: Thanks, that makes sense. I agree that inferring `epoch_len` inside the logger is too assumption-heavy.

I’m planning to revise the design as follows:

1. Keep `BasicWandbChartStatsHandler` unchanged.
2. Refactor the existing `print_unsupervised_stats(stats, epoch_len)` logic so the metric computation is reusable, e.g. by adding a helper that returns the benchmark values instead of only printing them.
3. Have both `print_unsupervised_stats(...)` and WandB logging use that same helper, so the benchmark computation has one source of truth.
4. Do not log runtime again, since that is already handled by the existing logging.

The only remaining question is where `epoch_len` should come from during WandB logging, since `print_unsupervised_stats(stats, epoch_len)` currently takes it explicitly. Is there an existing config/experiment field I should read from, or would you prefer passing `epoch_len` explicitly through the logger config?

vkakerbeck: The simplest thing that comes to mind right now is to look at the value `n_train_epochs` in the experiment configs and divide the overall number of episodes by that number.

PR #1084, opened 2026-07-04
docs: Updates to future work items pt 3 (Niels)

nielsleadholm:
This PR makes several additional changes the Future Work pages. Some points to highlight for your context @vkakerbeck beyond what we've already discussed:
- I updated the description of the Motor System section of Future Work to make it clear that this includes policies and goal generation (which it already did in practice, and which I think isn't a bad thing - otherwise almost everything falls into the already massive LM bucket). We may just want to move Learn Hot Spots on Objects if you agree with this change.
- Added a brief description the hybrid edge-matching approach based on the research meeting discussion the other day (I think this is a helpful clarification for those unfamiliar with the implementation).
- I added an item about moving all of our benchmark configs to use constrained object models; it wasn't one we specifically discussed in the scope of the updates, but I think it should be a useful one.
- Let me know what you think of the bio framing of the "CMA neurons" and alternative methods for tessellating space. My thinking is that these will be exciting for anyone interested in working on more bio-like algorithms, without being a distraction.
- I provisionally removed the item about a model-free surface agent policy that does a spiral-scan. Re-reading this, I'm not sure we'd really want to implement this; in the medium term, I think it would be better to move all of our learning to the salience-->uniform sampling policy that is now described (i.e. rely on cross-modality transfer); if we wanted to improve learning specifically with the surface agent, then I think it would be better to explore model-based methods that move to regions which have received limited exploration. Let me know what you think.
- As per our discussions, I've removed the automated benchmarks item - I added some more context about this to our Shortcut document if you're curious.


docs/future-work/framework-improvements/generalize-dimenensionality-of-monty-computations.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Generalize Dimensionality of Monty's Computations
+description: We are developing LMs that support 1D and 2D reference frames, but our computations all assume more expensive 3D representations.
+rfc: required
+estimated-scope: unknown
+improved-metric: speed
+output-type: analysis, PR
+skills: python
```
vkakerbeck:
```suggestion
skills: python, spatial-transformations
```

docs/future-work/framework-improvements/improve-terminology-including-directory-and-experiment-names.md
```diff
@@ -0,0 +1,56 @@
+---
+title: Improve Terminology Including Directory and Experiment Names
+description: There are several instances where terminology in the code could be clarified, including in the directories organizing tbp.monty, and our experiment configs.
+rfc: required
```
vkakerbeck: I think this could be optional depending on the scope of the change

docs/future-work/framework-improvements/improve-terminology-including-directory-and-experiment-names.md
```diff
@@ -0,0 +1,56 @@
...
+### Policies, Goals, and Actions
+One concrete example is how we refer to goals, actions, and policies. Currently we have "Goal Generators" in SMs and LMs, and "Policies" in the Motor System. We also have "Goals" that are passed to the Motor System, but in our research meetings, we now refer to Goal States vs. Target Poses. This latter distinction has helped clarify when something is about changing an object's state, vs. specifying a spatial target. A potential resolution might be to remove the term "Generator", and use something like:
+- Goal-State Policies:
+    - Exist in LMs.
+    - Produce Goal States, which only go to other LMs, and specify the desired *state* of an object.
+- Target-Pose Policies:
+    - Can exist in SMs or LMs; as such, the one form of policy that can be either model-free (if within an SM) or model-based (if within an LM).
+    - The equivalent of the current Goal Generators (hypothesis testing policy and SM salience policy).
+    - Pass a Target Pose to the Motor System (i.e., specify a location and orientation, but not an object state).
+- Action Policies
```
vkakerbeck, line 26: I think this might be problematic when we need policies that coordinate a series of state and pose changes. I would probably leave this section less specified and just add a note that one should have a discussion with our team on how to best change this

docs/future-work/learning-module-improvements/evaluate-alternative-tessellation-methods-for-3d-space.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Evaluate Alternative Tessellation Methods of 3D Space
+description: The constrained object models use cubic voxels to tile space, but alternatives may improve spatial precision while enabling computationally efficient search algorithms.
+rfc: optional
+estimated-scope: large
+improved-metric: accuracy, pose, speed
```
vkakerbeck, line 6: How do you think this would affect accuracy and pose? To me this sounds like mostly a speed improvement

docs/future-work/learning-module-improvements/evaluate-alternative-tessellation-methods-for-3d-space.md
```diff
@@ -0,0 +1,34 @@
...
+contributor: 
+status: open
+---
+
+Our [constrained object models](../../how-monty-works/learning-module/object-models.md#object-models) represent space during learning as tessellated cubes (voxels). The presence of features in each of these voxels then informs the construction of a point cloud that we later use for K-D tree search.
+
+It may be that querying these voxels directly, for example with a hash table, could represent a more efficient method of search than our current K-D tree method ([see this related item for further discussion](../framework-improvements/find-faster-alternative-to-kdtree-search.md)). While our past investigations have indicated that K-D tree search is quite efficient, this may differ with hardware appropriately matched to a specific algorithm. Moreover, K-D trees have the disadvantage that, during significant updates to the point cloud (i.e. learning), the tree must generally be reconstructed. A long-term aim however is that Monty can continuously learn in an online manner, prompting the desire to avoid such a rebuilding process. Direct voxel-querying with methods such as hash tables may be an approach to avoid this issue.
```
vkakerbeck: Based on the experiments I ran, using hashing magnified the issue (i.e. the main reason why we didn't switch to it was because construction took too long. But querying was actually faster than KD-trees.)

docs/future-work/learning-module-improvements/evaluate-alternative-tessellation-methods-for-3d-space.md
```diff
@@ -0,0 +1,34 @@
...
+
+A problem arises, however, with simply using the existing voxel-based representations. In particular, cubes are a poor way of tessellating 3D space when concerned with cell membership: points that fall just within the corner of a cube are still considered a member of that cube, yet are more distant from the voxel center than points that fall just within the face-center of a cube. This is what is known as *anisotropy*, or a lack of uniformity in all directions. In Monty, this has the practical risk of biasing which voxel a point belongs to, depending on which direction a sensor is moving. This could impact the precision with which the location and pose of an object is resolved, additionally harming classification accuracy. This issue will be more pronounced as our models become sparser (as voxels will cover a larger proportion of space), yet [sparser models are an important long term aim of ours](./use-models-with-fewer-points.md).
```
vkakerbeck: Not sure how to incorporate these but just a few notes:
- making voxels larger isn't the only way that we can make the models sparser
- The hypotheses testing is still done in continuous space, using the graph built from points in voxels. Locations in the graph correspond to average locations stored in a voxel, not the voxels center. It can still be skewed by voxel borders but I don't think it is as bad as you describe and it might not be an issue at all

docs/future-work/learning-module-improvements/evaluate-alternative-tessellation-methods-for-3d-space.md
vkakerbeck, line 1: I'm generally unsure whether we should add this item. It seems more like a curiosity than a crucial step on our roadmap. I think if we keep it, we should add a note at the beginning that a first step would be to evaluate the effect that anisotropy has on performance in our current implementation. It might not be an important issue at all. And if it is just about speed, then we still have the item on finding a faster alternative to KD Tree search.

docs/future-work/learning-module-improvements/improve-bounded-evidence-performance.md
```diff
@@ -10,25 +10,37 @@ contributor: ramyamounir, scottcanoe
...
 This behavior becomes problematic when new hypotheses are sampled later in time.
 Newly sampled hypotheses are disadvantaged and may fail to out-compete existing ones, even if they are more consistent with recent observations.
+A concrete outcome that we could measure for is that the most likely hypothesis (MLH) that Monty infers approaches the "theoretical limit" (see `TheoreticalLimitLMLoggingMixin`), i.e. the error associated with the best hypothesis that Monty has initialized. Unbound evidence currently means that an older, "good enough" hypothesis will often remain the MLH even where the theoretical limit is lower following [sampling of new hypotheses](https://docs.thousandbrains.org/docs/evidence-based-learning-module#burst-sampling). Note that standard measures of rotation error fail to account for objects that have symmetry, which Monty detects. As such, an additional change may be to modify the theoretical limit to use alternative measures of rotation error that are not biased by symmetry, such as the Chamfer Distance metric used in our [Thousand Brains systems paper](https://arxiv.org/abs/2507.04494).
```
vkakerbeck: How is a different rotation error measure important for bounded evidence?

docs/future-work/learning-module-improvements/improve-bounded-evidence-performance.md
```diff
@@ -10,25 +10,37 @@ contributor: ramyamounir, scottcanoe
...
 Bounded scores are easier to work with as they allow downstream components to rely on known numerical ranges, simplifying threshold selection and enabling more stable decision rules for deleting and resampling hypotheses.
 
+## Potential Solutions
+
 Bounding evidence through methods such as [exponential moving averages](https://youtu.be/A1cOwvZpgjU?si=1oyk6-BTij6TaRdG&t=1668), weighted averages, or normalization by hypothesis age can keep scores within a bounded range.
 However, these techniques compress historical information and cause memory to fade over time.
 Monty loses the influence of important past features, such as a mug handle, as a result, recognition can become less accurate in long sequences.
 
-A promising direction is to pair bounded evidence with [saliency driven saccades](https://docs.thousandbrains.org/docs/implement-efficient-saccades-driven-by-model-free-and-model-based-signals) or other attention mechanisms.
+A promising direction is to pair bounded evidence with saliency-driven saccades or other attention mechanisms, which [we now have a working version of](https://docs.thousandbrains.org/docs/salience-sm#saliencestrategy).
```
vkakerbeck: Have we tested setting past and present weight to add up to 1 since we introduced the saliency strategy?

docs/future-work/learning-module-improvements/use-constrained-grid-object-models-for-all-benchmark-experiments.md
```diff
@@ -0,0 +1,23 @@
...
+
+This update could work in concert with [an improved exploration policy for the distant agent](../motor-system-improvements/model-free-distant-agent-exploration-policy.md). Early tests with salience-based learning policies suggest that these naturally support sampling in a scale-invariant way. In particular, this policy will perform smaller saccades to explore a smaller object such as a dice, and larger saccades to explore an object like a cereal box; this is in contrast to the naive scan and surface-agent learning policies, which will sample many more points for a larger object.
+
+Finally, this change could enable a significant simplification of the Monty code-base by removing the `FeatureGraphLM` and `DisplacementGraphLM` classes. Even if some elements of the latter are cherry-picked for the purpose of a [hybrid node and edge-matching LM](./hybrid-node-and-edge-matching.md), this would still enable the removal of >1k lines of complex code and dozens of associated unit tests.
```
vkakerbeck: This would also involve rewriting a lot of the tests to use the EvidenceLM since otherwise we will lose significant test coverage. Not sure this is a positive we should mention here...

docs/future-work/motor-system-improvements/bottom-up-exploration-policy-for-surface-agent.md
vkakerbeck, line 1: This might become more relevant for continual learning scenarios that we want to move towards more. When we want to learn and infer with a surface agent, we might want to switch to a more learning-focused policy mid-experiment, and currently, we don't have a good one. But yes, we could revive this at that point.

docs/future-work/motor-system-improvements/model-free-distant-agent-exploration-policy.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Model-Free Distant Agent Exploration Policy
+description: A policy for the distant agent to rapidly visit salient regions, then systematically explore.
+rfc: optional
+estimated-scope: medium
+improved-metric: learning, accuracy, speed, numsteps
```
vkakerbeck, line 6: Would this impact speed in any other way than reducing num_steps? If not I would remove the speed tag

docs/future-work/motor-system-improvements/model-free-distant-agent-exploration-policy.md
```diff
@@ -0,0 +1,23 @@
...
+
+The proposal is to implement a decay in the salience map such that we initially saccade to salient regions, but that these slowly become less interesting. As the salience decays, the policy will converge to a pure inhibition of return strategy. This would capture the intuitive practice of initially looking at interesting things, then exploring more uniformly, and this could continue until a model-based goal is received to move to a particular location.
+
+This policy could be used at both learning and inference, giving us sparse, but still generalizable (cross-modality) models. In particular, one limitation of a salience-only strategy during learning and inference is that other modalities which do not follow the same policy at inference (such as the surface agent) will visit under-represented regions. At the same time, the policy proposed here could ensure that salient regions are more strongly represented at both learning and inference. This is particularly important in naturalistic settings, where sensory inputs are often sparse (e.g. objects scattered around a room).
```
vkakerbeck: Would salient regions actually be represented more during learning with this strategy? It sounds more like it guarantees that we visit the most salient regions first, but then we might sample an equal (or even larger) amount of non-salient locations on top of that.

I am not sure I understand what you mean with "sensory inputs are often sparse". Do you mean we get few observations on each object?

docs/future-work/motor-system-improvements/model-free-distant-agent-exploration-policy.md
vkakerbeck: Finishing reading the description, did you add speed to refer to faster inference because of sparser learned models? I think in this case it is fine to leave in

docs/future-work/sensor-module-improvements/additional-methods-for-depth-perception.md
```diff
@@ -0,0 +1,76 @@
...
+
+For this task, one could test existing computer vision techniques that estimate depth based, for example, on parallax. We are particularly interested in techniques that are motivated by how biological systems solve this.
+
+To assist with this task, a [video overview of depth processing in the brain and computer vision systems, and their relevance to Monty, is available here](https://www.youtube.com/watch?v=6xr42m5vjbM). Below is a discussion summarizing some of the content of this video, as well as how these ideas might be combined to improve depth estimation in Monty.
```
vkakerbeck: I would put the link on its own line so the video is rendered in the docs

docs/future-work/sensor-module-improvements/universal-sm-for-morphological-feature-extraction.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Universal SM for Morphological Feature Extraction
+description: Develop a sensor module that automatically extracts CMP-compliant morphological features, independent of input modality and dimensionality.
+rfc: required
+estimated-scope: large
+improved-metric: features-and-morphology, transfer, deformations, generalization, real-world
+output-type: prototype, monty-feature, PR, paper
```
vkakerbeck:
```suggestion
output-type: prototype, monty-feature, PR, publication
```

docs/future-work/sensor-module-improvements/universal-sm-for-morphological-feature-extraction.md
```diff
@@ -0,0 +1,19 @@
...
+
+Importantly, the "space" in "local changes across space" would ideally include 1D, 2D, and 3D space. As such, progress here might support other Future Work items, such as feature extraction in [a 1D sensor module](one-d-sensor-module.md). Also note that extracting *movement* in an appropriate dimensionality is a separate, but related, problem to extracting morphological features.
+
+A proposed implementation could be tested by verifying that it recovers the morphological features that our current SMs already extract in [3D](https://docs.thousandbrains.org/docs/sm-for-rgbd-data) and [2D](https://docs.thousandbrains.org/docs/two-d-sensor-module) space given an RGB-D input, as well as the features extracted by the pipeline we developed for [ultrasound](https://github.com/thousandbrainsproject/tbp.ultrasound_perception).
```
vkakerbeck: I'm not sure this will happen. Why would we expect it to learn our hand-crafted features?

docs/future-work/voting-improvements/clean-up-and-simplify-voting.md
```diff
@@ -12,4 +12,8 @@ status: open
...
+
+Voting on rotations before sharing hypothesis locations may also improve accuracy. The intuition is that with our current voting approach, all of the votes exist in a single, shared space after being transformed by the orientations and relative sensor displacements. This means that votes can still align with hypotheses even where their object orientations are inconsistent. In other words, the vote could have a spurious neighbor in the receiving LMs hypothesis space, and thereby - incorrectly - provide positive evidence during the voting process. This is much like if we allowed voting across reference frames (a mug vs. a banana model) - the sheer number of points means that there is a chance of a collision occurring. This is particularly risky where thousands of votes are involved, which will increase in likelihood as the scale of Monty grows to more LMs.
+
+Finally, the existing voting algorithm requires the construction of a K-D tree search whenever voting occurs. Through appropriate formulation of the voting process, including filtering by orientations, it may be possible to eliminate this requirement, and thereby reduce the computational cost of voting.
```
vkakerbeck: How would filtering by orientation help with this? Wouldn't that just reduce the number of votes?

docs/future-work/motor-system-improvements.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Motor System Improvements
+title: Motor System and Policy Improvements
```
vkakerbeck, line 2: We could consider splitting out this topic into
- Motor System (goal state selection, subcortical policies)
- Model-based Policies (LM GSG)
- Model-free Policies (SM GSG)

vkakerbeck (commented):
Nice, thanks for all the updates and additions! I left a few comments and questions, but overall great changes :) 

On some items, you still need to update the skill `monty` to `monty-beginner` or `monty-advanced`

vkakerbeck:
Oh one other thing I just saw when going through our list: We wanted to delete this page: https://docs.thousandbrains.org/docs/contributing-learning-modules after adding the more specific page on different space representation

docs/future-work/framework-improvements/improve-terminology-including-directory-and-experiment-names.md
nielsleadholm: Good point, rather than remove the example (which I think can be helpful), I've added the below. Let me know what you think:

"However, even the above suggestion isn't without its problems. For example, there are likely to be policies that must coordiante a mixture of Goal States and Target Poses ("Mixed" Policies?). This example emphasizes that many such terminology changes will benefit from the RFC process."

docs/future-work/learning-module-improvements/evaluate-alternative-tessellation-methods-for-3d-space.md
nielsleadholm: That's around this point "This could impact the precision with which the location and pose of an object is resolved, additionally harming classification accuracy. This issue will be more pronounced as our models become sparser", and related discussion of anisotropy. Let me know if it would be helpful to clarify more.

docs/future-work/learning-module-improvements/evaluate-alternative-tessellation-methods-for-3d-space.md
nielsleadholm: Ok interesting, I wonder whether that would depend on the hashing method. A straightforward look-up table based on checking whether we are in a voxel or not shouldn't have must cost with updating, unless we're constantly changing the memory allocated. I'll add a caveat to the description.

docs/future-work/motor-system-improvements.md
nielsleadholm: Do you think it's confusing the way it is with this change? I think for many people, the connection between Motor Systems and Policies would be intuitive. We could break it down, but at the moment there are only 9 items in this section, whereas there are 30 in the `learning-module-improvements` section. If we want it to be more granular, then I think we should also break down that section, but I'm not sure if that's necessary with the look-up tool. Basically I just wanted to make sure the heading was more accurate, i.e. these items aren't only about Motor System.

docs/future-work/learning-module-improvements/evaluate-alternative-tessellation-methods-for-3d-space.md
nielsleadholm: Good point, about sparsity and voxel size, I've updated the wording.

And yeah re. the second point, this is what the "A complementary benefit could also emerge even with the existing conversion we perform from a voxel grid to a point cloud. In particular". I've updated the description to include locations in the averaging comment.

docs/future-work/learning-module-improvements/evaluate-alternative-tessellation-methods-for-3d-space.md
nielsleadholm: Hmmm right, I've updated the description - I think the entry point is clearer now, while still providing some useful information for anyone wanting to work on this.

docs/future-work/learning-module-improvements/improve-bounded-evidence-performance.md
nielsleadholm: No, I've clarified this in the text

docs/future-work/learning-module-improvements/improve-bounded-evidence-performance.md
nielsleadholm: Yeah fair point, if there is symmetry, then we don't really need to worry about the theoretical limit. Updated.

docs/future-work/learning-module-improvements/use-constrained-grid-object-models-for-all-benchmark-experiments.md
nielsleadholm: Yeah I should have specified the unit tests point more clearly, I've updated the text.

docs/future-work/motor-system-improvements/model-free-distant-agent-exploration-policy.md
nielsleadholm: Yeah exactly, I updated the description to make it clearer

docs/future-work/motor-system-improvements/model-free-distant-agent-exploration-policy.md
nielsleadholm: I think it would, as we're basically guaranteed to visit the salient regions, whereas particular points in non-salient regions will be sampled subject to more random interactions between the IoR and the points that were previously visited. But yes that doesn't mean there are *more* salient points. I've updated "strongly" to "robustly", which I think captures this. 

Oh yes that was a weird choice of words on my part 😅 Updated

docs/future-work/sensor-module-improvements/universal-sm-for-morphological-feature-extraction.md
nielsleadholm: Not the same, but similar. I've updated the wording.

docs/future-work/voting-improvements/clean-up-and-simplify-voting.md
nielsleadholm: Hmm yeah not sure what I meant there. I've updated the wording to refer to the possibility of using model points, as I recently mentioned in the context of voting with symmetry in mind (although this may not be possible without affecting accuracy negatively).

nielsleadholm:
Great, thanks for the useful comments. Those should be addressed now.

And re. the concrete LM approach, from our last Slack discussion on this, I believe we decided to try a rename rather than deletion. I've gone ahead and modified it based on that, which I think is helpful, but let me know what you think.

docs/future-work/learning-module-improvements/evaluate-alternative-tessellation-methods-for-3d-space.md
vkakerbeck: When I tried this, I tried locality sensitive hashing (this library: https://github.com/kayzhu/lshash ) Similar to KD Tree search, I think many use cases of it focus on higher-dimensional data (whereas we just do 3D) and so we don't get as many of the benefits.
I also tried just rounding x,y,z values and using them as indices: https://github.com/thousandbrainsproject/nupic.monty/blob/7a028bbdd68c70fec8d65487e49ccffdf56bd4eb/projects/speedup/KNNSearch.ipynb I'd have to dig things back up but I think there the issue is more precision and memory requirements but I am not sure

docs/future-work/motor-system-improvements.md
vkakerbeck: Yes, I think for now it is fine to leave as it is

vkakerbeck (approved):
Nice, thanks for the updates.

PR #1041, opened 2026-06-19
feat!: vocus2 benchmark configs and new pretrained models

scottcanoe:
This PR adds two new benchmark configs
 - supervised_pretraining_10distinctobj_dist_agent_vocus2
 - randrot_noise_10distinctobj_dist_agent_vocus2

The addition of a a new group of pretrained models necessitates an upgrade from YCB v12 -> v13, so this PR makes the necessary changes.

## Benchmarks

There are no changes to existing benchmarks, though the benchmark tool shows one slight change in a 10simobj experiment since it looks like the output precision has changed (0 decimals to 2 decimals in rotation error).
### base_config_10distinctobj_dist_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100 | 100 | 0 | ⬜ |
| used_mlh (%) | 0 | 0 | 0 | ⬜ |
| match_steps | 28 | 28 | 0 | ⬜ |
| rotation_error (deg) | 10.29 | 10.29 | 0 | ⬜ |
| runtime (min) | 3 | 2 | -1 | ✅ |
| episode_runtime (sec) | 5 | 5 | 0 | ⬜ |
| num_episodes | 140 | 140 | 0 | ⬜ |

### base_config_10distinctobj_surf_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100 | 100 | 0 | ⬜ |
| used_mlh (%) | 0 | 0 | 0 | ⬜ |
| match_steps | 28 | 28 | 0 | ⬜ |
| rotation_error (deg) | 3.64 | 3.64 | 0 | ⬜ |
| runtime (min) | 4 | 4 | 0 | ⬜ |
| episode_runtime (sec) | 12 | 11 | -1 | ✅ |
| num_episodes | 140 | 140 | 0 | ⬜ |

### randrot_noise_10distinctobj_dist_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 99 | 99 | 0 | ⬜ |
| used_mlh (%) | 2 | 2 | 0 | ⬜ |
| match_steps | 39 | 39 | 0 | ⬜ |
| rotation_error (deg) | 15.85 | 15.85 | 0 | ⬜ |
| runtime (min) | 2 | 2 | 0 | ⬜ |
[...]

PR #1086, opened 2026-07-07
chore: update ycb_10objs.csv randrot_noise_10simobj_dist_agent

jeremyshoemaker:
The results of the `randrot_noise_10simobj_dist_agent` slightly changed as a result of #1041. This updates the CSV to match so it doesn't get flagged as a change in benchmark results anymore.


tristanls-tbp (approved):

scottcanoe (approved):
I don't remember making this change, so I think it was probably a rounding/string-formatting quirk in tbp.benchmarks.

jeremyshoemaker:
> I don't remember making this change, so I think it was probably a rounding/string-formatting quirk in tbp.benchmarks.

Maybe I've cited the wrong PR as the source. 🤷‍♂️

PR #1087, opened 2026-07-07
test: fix GetGoodViewTest to confirm agent moves

jeremyshoemaker:
While investigating some code coverage differences between the MuJoCo tests and the Habitat tests, I noticed that the `GetGoodViewTest.test_agent_too_far_away` test doesn't actually test what it is supposed to test. As written, the agent doesn't actually move, and most of the positioning procedure is skipped.

First, I updated the test to assert that the agent position has changed to make the test verify that it moves. Then with a failing test, I updated the starting position of the agent to make it have to move closer to the object, making the test pass.

There are some additional changes that I'll note with comments where they appear. I think there were a few mistakes in the configurations and the tests that I'm attempting to fix here, but don't by themselves fix the original problem.

### DIFF
--- a/src/tbp/monty/conf/environment/habitat_dist_agent_posz05.yaml
+++ b/src/tbp/monty/conf/environment/habitat_dist_agent_posz05.yaml
@@ -1,0 +1,52 @@
+# @package experiment.config.environment
+
+env_init_args:
+  objects:
+    - name: coneSolid
+      position:
+        - 0.0
+        - 1.5
+        - -0.1
+  scene_id: null
+  seed: 42
+  data_path: null
+  agents:
+    agent_args:
+      agent_id: agent_id_0
+      sensor_ids:
+        - patch
+        - view_finder
+      height: 0.0
+      position:
+        - 0.0
+        - 1.5
+        - 0.5
+      resolutions:
+        - - 64
+          - 64
+        - - 64
+          - 64
+      positions:
+        - - 0.0
+          - 0.0
+          - 0.0
+        - - 0.0
+          - 0.0
+          - 0.0
+      rotations:
+        - - 1.0
+          - 0.0
+          - 0.0
+          - 0.0
+        - - 1.0
+          - 0.0
+          - 0.0
+          - 0.0
+      semantics:
+        - false
+        - false
+      zooms:
+        - 10.0
+        - 1.0
+    agent_type: ${monty.class:tbp.monty.simulators.habitat.MultiSensorAgent}
+env_init_func: ${monty.class:tbp.monty.simulators.habitat.environment.HabitatEnvironment}
--- a/src/tbp/monty/conf/environment/mujoco_dist_agent_posz05.yaml
+++ b/src/tbp/monty/conf/environment/mujoco_dist_agent_posz05.yaml
@@ -1,0 +1,25 @@
+# @package experiment.config.environment
+
+env_init_args:
+  agents:
+    - _target_: tbp.monty.simulators.mujoco.agents.DistantAgent
+      _partial_: true
+      agent_id: agent_id_0
+      sensor_configs:
+        patch:
+          _target_: tbp.monty.frameworks.sensors.SensorConfig
+          resolution:
+            width: 64
+            height: 64
+          zoom: 10.0
+        view_finder:
+          _target_: tbp.monty.frameworks.sensors.SensorConfig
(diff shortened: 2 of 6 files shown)


src/tbp/monty/conf/environment/habitat_dist_agent_posz05.yaml
```diff
@@ -0,0 +1,52 @@
...
+    agent_args:
+      agent_id: agent_id_0
+      sensor_ids:
+        - patch
+        - view_finder
+      height: 0.0
+      position:
+        - 0.0
+        - 1.5
+        - 0.5
```
jeremyshoemaker, line 23: This is the only difference in this file from the base one, changing this from 0.2 to 0.5.

src/tbp/monty/conf/environment/mujoco_dist_agent_posz05.yaml
```diff
@@ -0,0 +1,25 @@
...
+        view_finder:
+          _target_: tbp.monty.frameworks.sensors.SensorConfig
+          resolution:
+            width: 64
+            height: 64
+          zoom: 1.0
+      position:
+        - 0.0
+        - 1.5
+        - 0.5
```
jeremyshoemaker, line 24: This is the only difference in this file from the base one, changing this from 0.2 to 0.5.

src/tbp/monty/conf/experiment/test/integration/positioning_procedures/get_good_view/dist_agent_too_far_away.yaml
```diff
@@ -6,10 +6,10 @@ defaults:
   - /monty/learning_module: displacement_1lm
   - /monty/sensor_module: camera_dist
   - /monty/connectivity: 1lm_1sm
-  - /environment: habitat_dist_agent
+  - /environment: habitat_dist_agent_posz05
   - /env_interface: test_train_2obj_predefined_eval_1obj_cubeSolid_predefined
-  - /env_interface/positioning_procedures_train: getgoodview_mop1_viewfinder_patch
-  - /env_interface/positioning_procedures_eval: getgoodview_mop1_viewfinder_patch
+  - /env_interface/positioning_procedures_train: getgoodview_viewfinder_patch
+  - /env_interface/positioning_procedures_eval: getgoodview_viewfinder_patch
```
jeremyshoemaker, line 12: I think this test isn't supposed to use the multiple objects configurations. It works with the original configs, but I don't think that was the intent.

tests/integration/positioning_procedures/get_good_view_test.py
jeremyshoemaker, line 120: I was going to say I thought this was a mistake, but looking at the code more closely here in the diff, I see why it's here. I'll be adding it back.

jeremyshoemaker: To be clear, I had removed the `exp.train()` call in the second test because I thought it was a mistake. I didn't notice that we're running the steps of the experiment below this in EVAL mode. It passes either way so I'm not sure it makes a difference.

scottcanoe (approved):
It all makes sense to me. But just to be clear, changing the initial z-position from 0.2 to 0.5 was needed to make sure the agent actually moves?

jeremyshoemaker:
> It all makes sense to me. But just to be clear, changing the initial z-position from 0.2 to 0.5 was needed to make sure the agent actually moves?

From what I could tell looking at the rendered images at the beginning of the experiment, the cube is taking up the full frame of the view finder, so the coverage is 100% which is above the threshold, and so the procedure gets skipped. I've seen this in other experiments using the primitive objects, where they're bigger than the YCB objects. I think it's like 10cm per side for the cube.

scottcanoe:
> From what I could tell looking at the rendered images at the beginning of the experiment, the cube is taking up the full frame of the view finder, so the coverage is 100% which is above the threshold, and so the procedure gets skipped. I've seen this in other experiments using the primitive objects, where they're bigger than the YCB objects. I think it's like 10cm per side for the cube.

OK great, nice catch.

jeremyshoemaker:
Oh, that was the thing I forgot I'd already looked up. In the `multi_object_target_not_visible` test, the objects during the eval phase are set back along the negative Z axis at -0.2, which makes it further away, and that's why that test wasn't failing when I add the assert.

PR #1090, opened 2026-07-08
test: create MuJoCo version of tutorials_test.py

jeremyshoemaker:

Initially I wasn't going to migrate this one until we were ready to switch over the tutorials, but those so rarely change that it's fine to do it now, especially because we don't have to update the tutorials themselves until later.

This also adds a script for generating the YCB dataset with the metadata files, and adds a `test-uv-tutorials` action.

I updated the `TutorialTest`s to not run a full training pass, since we don't do the same with the evaluation, but instead limit it to a small number of steps.


tests/mujoco/tutorials/tutorials_test.py
```diff
@@ -0,0 +1,159 @@
...
+    """Tests of the configurations that we have in our tutorial documentation.
+
+    The purpose of these tests is to ensure that the tutorial configs load properly
+    and run to completion without error so that any users of Monty aren't surprised
+    by broken tutorials. We aren't actually testing the results against any expected
+    values, though we could do that if we decide we need to.
+    """
+
+    # TODO - Find a way to make these test configs be the source of truth for the
+    #   documentation rather than having to manually keep them in sync.
```
jeremyshoemaker, line 35: This is probably the biggest actual change, and it's just adding the docstring and TODO comment.

jeremyshoemaker:
Well, something isn't right because the tests are failing here. One is obvious because I haven't split out the tutorial test from the main set like we do for the Habitat versions, and it doesn't have the datasets for some of these tests. But other tests in that set are failing too, so I need to sort that out.

tristanls-tbp:
**note**: Based on your comment and the failing tests, I assume I should return to this PR later. In the meantime, I am marking it as a draft until it's ready to review.

**suggestion**: You need to download the dataset onto the test machine. See the existing tutorial test job: https://github.com/thousandbrainsproject/tbp.monty/blob/d2c72dd7b0830f9d6f8bad764de9b3cf8d0d7966/.github/workflows/monty.yml#L475

jeremyshoemaker:
Oh yeah, that's right, this uses the YCB set, and not just the Omniglot set, which explains the other test failures. I have the former locally but not the latter, which meant the Omniglot ones failed locally and I knew they probably just worked, but I forgot about the YCB set and that the tutorial test job had a step to download them.

jeremyshoemaker:
I think this is ready for review.

.github/workflows/monty.yml
```diff
@@ -480,7 +480,7 @@ jobs:
           source activate tbp.monty
           set -e
           python -m habitat_sim.utils.datasets_download --uids ycb --data-path ~/tbp/data/habitat
-          git clone https://github.com/brendenlake/omniglot.git ~/tbp/data/omniglot
+          git clone --depth 1 https://github.com/brendenlake/omniglot.git ~/tbp/data/omniglot
```
jeremyshoemaker, line 483: This probably doesn't make a big difference this repo, but setting `--depth 1` makes Git only clone the most recent commit, since we don't need the rest of the history in this context.

tests/mujoco/tutorials/tutorials_test.py
jeremyshoemaker: Technically no longer true now that I added the step limits to the training experiments.

pyproject.toml
jeremyshoemaker, line 1: There are some white space changes because apparently the most recent PyCharm formats TOML files now. I looked into disabling it and it seems involved, and I don't think these changes are too out of the ordinary.

.github/workflows/tools.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           (conda env list | grep tbp.monty) && conda remove --name tbp.monty --all --yes || true
           conda env create
           source activate tbp.monty
-          pip install -e .[dev,github_readme_sync_tool,print_version_tool,generate_api_docs_tool,future_work_widget_tool]
+          pip install -e .[dev,github_readme_sync_tool,print_version_tool,generate_api_docs_tool,future_work_widget_tool,ycb_downloader_tool]
```
jeremyshoemaker: The `ycb_downloader_tool` doesn't have any doctests, but the tests need to be able to import the module to learn that. Adding the dependency here is better than doing a one-off exclusion of the downloader module.

.github/workflows/monty.yml
```diff
@@ -497,6 +497,69 @@ jobs:
           mkdir -p test_results/pytest
           xvfb-run pytest tests/tutorials --junitxml=test_results/pytest/results.xml --verbose -n 1
 
+  test_uv_tutorials_monty:
```
jeremyshoemaker, line 500: With the performance improvement of not running full training experiments, this doesn't need to be a separate step, but I'm leaving it as one for now.

Before that change, I was seeing runtimes around 30 minutes, and when I tried to switch to using the GPU runner it made it worse, running closer to 1 hour and 30 minutes. I'm not sure if that was due to a misconfiguration (I did remove the environment variables that switch MuJoCo to using the software rendering to get it to use the GPU), but fixing the runtimes of the tests was an easier fix.

.github/workflows/monty.yml
```diff
@@ -497,6 +497,69 @@ jobs:
...
+    name: test-uv-tutorials-monty
+    runs-on:
+      group: tbp.monty
+      labels: tbp-linux-x64-ubuntu2204-8core
+    timeout-minutes: 120
+    needs:
+      - install_monty
```
tristanls-tbp: **question**: Why does this job need `install_monty`? I think the `Create uv environment` step obsoletes all of `install_monty`.

.github/workflows/monty.yml
```diff
@@ -497,6 +497,69 @@ jobs:
...
+      - should_run_monty
+      - test_monty
```
tristanls-tbp: **issue**: This should be `test_uv_monty`.

.github/workflows/monty.yml
```diff
@@ -497,6 +497,69 @@ jobs:
...
+          set -e
+          mkdir -p ~/tbp/data/mujoco/objects/ycb
+          curl -L https://tbp-data-public-5e789bd48e75350c.s3.us-east-2.amazonaws.com/tbp.monty/ycb_objects_1.0.tgz | tar -xzf - -C ~/tbp/data/mujoco/objects/ycb
+          git clone --depth 1 https://github.com/brendenlake/omniglot.git ~/tbp/data/omniglot
+          unzip -qq ~/tbp/data/omniglot/python/images_background.zip -d ~/tbp/data/omniglot/python
+          unzip -qq ~/tbp/data/omniglot/python/strokes_background.zip -d ~/tbp/data/omniglot/python
+          mkdir -p ~/tbp/data/worldimages
+          curl -L https://tbp-data-public-5e789bd48e75350c.s3.us-east-2.amazonaws.com/tbp.monty/worldimages.tgz | tar -xzf - -C ~/tbp/data
+          mkdir -p ~/tbp/results/monty/pretrained_models/pretrained_ycb_v13
+          curl -L https://tbp-pretrained-models-public-c9c24aef2e49b897.s3.us-east-2.amazonaws.com/tbp.monty/pretrained_ycb_v13.tgz | tar -xzf - -C ~/tbp/results/monty/pretrained_models
```
tristanls-tbp, line 534: **issue**: To avoid this brittle duplication, I think this should be an action in `.github/actions`.

tristanls-tbp (commented):
Submitting comments for a review in progress.

.github/workflows/monty.yml
jeremyshoemaker: I've been meaning to change that for the `test-uv-monty` workflow as well and keep forgetting to.

.github/workflows/monty.yml
jeremyshoemaker: I did think of that, but I also thought that when the Habitat version goes away, or rather, this one becomes the `test-tutorial` action, then there isn't any duplication.

tools/ycb_downloader/__init__.py
tristanls-tbp, line 1: **question**: Is there anything preventing this tool from being in its own repo?

**note**: It is my understanding that the sole purpose of this script was one-time use to create a tarball that we now host ourselves.

**suggestion**: Move into its own repo if you want to hang onto this.

.github/workflows/tools.yml
tristanls-tbp: **issue**: I think `ycb_downloader_tool` shouldn't be part of the `tbp.monty` repo.

.github/workflows/monty.yml
jeremyshoemaker: Actually, now that I look at it, I know why that hasn't happened. The `test-uv-monty` workflow doesn't wait for `install-monty`, it waits for the various checks that themselves wait for `install-monty`.

I noticed the `test-tutorials-monty` job isn't waiting for those checks, was that on purpose? Should they wait for them?

jeremyshoemaker: Actually, never mind. Waiting on `test-monty` makes it wait on those checks indirectly.

tristanls-tbp: `test_tutorials_monty` does wait for `install_monty`. The reason is that the first step reads its output.

```yaml
  test_tutorials_monty:
    name: test-tutorials-monty
    runs-on:
      group: tbp.monty
      labels: tbp-linux-x64-ubuntu-nvidia-4core-t4gpu
    timeout-minutes: 120
    needs:
      - install_monty
      - should_run_monty
      - test_monty
    # ...
```

The reason `test_tutorials_monty` waits for `test_monty` is so that we don't bother running more involved tutorial testing if unit tests fail.

src/tbp/monty/conf/experiment/tutorial/surf_agent_2obj_unsupervised_mujoco.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+defaults:
+  # This configuration will take 1000 exploratory steps (`exp1000`) after recognizing the object to
+  # collect more information about it. It also requires at least 100 train steps (`t100`) before
+  # recognizing an object to avoid early misclassifications when we have few objects in
+  # memory.
+  - /monty: graph_exp1000_e3_t100_tot2500
+  - /monty/motor_system_config: surface
+  - /monty/learning_module: evidence_tutorial_surf_agent_2obj_unsupervised
+  # We will use the default surface sensor modules (one habitat surface patch, one logging view finder)
```
tristanls-tbp: **issue**: "habitat surface patch"

**suggestion**: Delete parentheses and their contents.

tests/mujoco/tutorials/tutorials_test.py
```diff
@@ -0,0 +1,184 @@
...
+    by broken tutorials. We aren't actually testing the results against any expected
+    values, though we could do that if we decide we need to.
+
+    Note: since we aren't checking the results of these experiments, we don't
+    need to run the full training and inference. Limiting the number of steps cuts
+    down on the runtime of these tests. If we start checking the results, we'll
+    need to change these.
+    """
+
+    # TODO - Find a way to make these test configs be the source of truth for the
```
tristanls-tbp: **issue**: Our TODO convention is `TODO: ` not `TODO - `.

tests/mujoco/tutorials/tutorials_test.py
```diff
@@ -0,0 +1,184 @@
...
+    #   documentation rather than having to manually keep them in sync.
+
+    def test_first_experiment(self) -> None:
+        with hydra.initialize_config_dir(version_base=None, config_dir=str(HYDRA_ROOT)):
+            config = hydra.compose(
+                config_name="experiment",
+                overrides=[
+                    "experiment=tutorial/first_experiment_mujoco",
```
tristanls-tbp: **suggestion (non-blocking)**: I feel that at least the first experiment should run through all of the steps. i.e., do run the whole thing. I would hate for it to fail on step 4 when someone tries to run Monty for the first time.

tristanls-tbp (commented):
This is the rest of my review.

.github/workflows/monty.yml
tristanls-tbp: **issue (non-blocking)**: Ok, changing this to non-blocking as it will be a you-problem when someone forgets to update it 🙂.

.github/workflows/tools.yml
jeremyshoemaker: Per our discussion, I'm going to move this to the `tbp.cli` repo.

jeremyshoemaker:
I think I addressed all your feedback. Let me know if I missed anything.

tristanls-tbp (approved):
Thank you for the updates. 👍

PR #1092, opened 2026-07-08
refactor: isolate common MotorPolicySelector init/reset code

dschumacher-tbp:
Put common init/reset code into `_init<ClassName>()` methods and extract some typed properties.

### DIFF
--- a/src/tbp/monty/frameworks/models/motor_policy_selectors.py
+++ b/src/tbp/monty/frameworks/models/motor_policy_selectors.py
@@ -88,17 +88,26 @@ class MotorPolicySelector(
 
 
 class SinglePolicySelector(MotorPolicySelector):
+    _policy: MotorPolicy
+
+    # TODO: Get rid of this once we have another path for telemetry.
+    _selected_goals: list[Goal | None]
+
     def __init__(self, policy: MotorPolicy):
         self._policy = policy
-        # TODO: Get rid of this once we have another path for telemetry.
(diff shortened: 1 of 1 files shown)


tristanls-tbp (commented):
**question**: Why is this necessary? For example, in `SinglePolicySelector`, you are creating more lines of code, where the existing code already works. There is no hierarchy that we have in policy selectors that requires this pained `_init_ClassName` pattern.

The `TODO` you created is the state of the code before this PR.

**suggestion**: Close this PR, there's nothing to do here.

dschumacher-tbp:
I suppose we can extract the properties when `reset()` is removed, if that's still desirable.

dschumacher-tbp:
Closed. No action.

tristanls-tbp:
**question**: What do you mean by extracting the properties? What properties? And what does extraction mean? As far as I can tell, all the properties touched by this PR will be retained when `reset()` is removed.

dschumacher-tbp:
I was just referring to "lifting" typed properties to the class level, per the pattern we're encouraging...

```python
class DistantPolicySelector(MotorPolicySelector):
    # policies
    _jump_to_goal: JumpToGoal
    _look_at_goal: LookAtGoal
    _default: MotorPolicy

    # state
    _is_jumping: bool

    # telemetry
    _selected_policies: list[MotorPolicy]
    _selected_goals: list[Goal | None]

    def __init__(
...
```

PR #1093, opened 2026-07-08
chore: reduce max_eval_steps in compositional inference benchmark experiments

hlee9212:
This PR addresses the issue of >4 hour runtime when evaluating `base_infer_objects_with_stickers_monolithic_models` on AWS. 

We are fixing this issue by reducing `max_eval_steps` from 500 (default, previous) to 100 steps. This results in a runtime just under 1 hour. 

For fair comparison, this change has been also applied to `base_infer_objects_with_stickers_comp_models`.


tristanls-tbp:
**suggestion (non-blocking)**: I think this is a `chore:`. This is a configuration change, so it does not feel like a code bug to me. Technically, it could also be a `refactor:` because our configs are in `src/tbp/monty`.

nielsleadholm (commented):
LGTM, just if you can please add a brief comment in the benchmarks readme document highlighting that, for these benchmarks, we run with fewer total steps (100 rather than our typical 500), as the monolithic models have many time-outs, which makes the benchmarks prohibitively slow to run otherwise.

nielsleadholm (approved):

PR #1098, opened 2026-07-13
chore: update sanity checks to fix one broken test when run locally on macs

brothman-tbp:
Locally we would get an error (below) because on Macs, temp directories have that capital 'T' that fail our new checker for only lower case docs paths. So I updated our checker to only test for the relative path, which should always be lower case and ignores the temp directory path.

```
if link_check_errors:
            for error in link_check_errors:
                logger.error(error)
>           sys.exit(1)
E           SystemExit: 1

tools/github_readme_sync/hierarchy.py:131: SystemExit
-------------------------------------------------------- Captured log call --------------------------------------------------------
ERROR    tools.github_readme_sync.hierarchy:hierarchy.py:130 File /var/folders/sm/klbtjmn17vxfl10knt_t13480000gn/T/tmpimsmvk24/category-1/doc-1.md does not exist based on slugs capitalization - check what's in the [] in hierarchy.md to make sure everything is lowercase
===================================================== short test summary info =====================================================
FAILED tools/github_readme_sync/tests/hierarchy_test.py::TestHierarchyFile::test_check_hierarchy_file_success - SystemExit: 1
```

### DIFF
--- a/tools/github_readme_sync/hierarchy.py
+++ b/tools/github_readme_sync/hierarchy.py
@@ -121,7 +121,7 @@ def check_hierarchy_file(folder: str):
                 link_check_errors.extend(errors)
 
             file_path = folder.joinpath(extract_file_path(line))
-            file_path_errors = sanity_check_file_path(file_path)
+            file_path_errors = sanity_check_file_path(file_path, folder)
             if file_path_errors:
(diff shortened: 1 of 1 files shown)


vkakerbeck (approved):
LGTM. Thanks for the quick fix

PR #1099, opened 2026-07-14
refactor: code updates to support newer library versions

jeremyshoemaker:
As part of the migration to MuJoCo, we're also updating Python and a few of our dependencies. For a while, we've been targeting 3.10. One issue we've run into is the newer version of NumPy has an issue on Apple Silicon where it logs a lot of spurious floating point warnings. Newer versions don't have this issue but require a newer Python version.

We decided to try to upgrade all the way to 3.14 and with the fixes in this PR, the result of changes to how strict NumPy is, the test suite passed, but trying to run an experiment, we ran into an error with Hydra being incompatible with Python's newer argparse library. They have a fix, but it hasn't been released yet.

As a result, we're now targeting 3.13 for now. This PR contains all the changes needed to get the tests to pass under Python 3.13 and newer library versions, while not breaking the current environment under Conda using Habitat.

<details>
<summary>Benchmarks Results (no significant changes)</summary>

| rotation_error (deg) | 10.29 | 10.29 | 0 | ⬜ |
| runtime (min) | 2 | — | — | — |
| episode_runtime (sec) | 5 | 5 | 0 | ⬜ |
| num_episodes | 140 | — | — | — |

| rotation_error (deg) | 3.64 | 3.64 | 0 | ⬜ |
| runtime (min) | 4 | — | — | — |
| episode_runtime (sec) | 11 | 12 | 1 | ❌ |
| num_episodes | 140 | — | — | — |

| percent_correct (%) | 99 | 99 | 0 | ⬜ |
| used_mlh (%) | 2 | 2 | 0 | ⬜ |
| match_steps | 39 | 39 | 0 | ⬜ |
| rotation_error (deg) | 15.85 | 15.85 | 0 | ⬜ |
| runtime (min) | 2 | — | — | — |
| episode_runtime (sec) | 6 | 7 | 1 | ❌ |
| num_episodes | 100 | — | — | — |

| match_steps | 29 | 29 | 0 | ⬜ |
| rotation_error (deg) | 9.82 | 9.82 | 0 | ⬜ |
| runtime (min) | 2 | — | — | — |
| episode_runtime (sec) | 7 | 7 | 0 | ⬜ |
| num_episodes | 100 | — | — | — |

### randrot_noise_10distinctobj_vocus_on_vocusm

| percent_correct (%) | — | — | — | — |
| used_mlh (%) | — | — | — | — |
| match_steps | — | — | — | — |
| rotation_error (deg) | — | — | — | — |
| runtime (min) | — | — | — | — |
| episode_runtime (sec) | — | — | — | — |
| num_episodes | — | — | — | — |

| match_steps | 30 | 30 | 0 | ⬜ |
| rotation_error (deg) | 12.85 | 12.85 | 0 | ⬜ |
| runtime (min) | 4 | — | — | — |
| episode_runtime (sec) | 21 | 25 | 4 | ❌ |
| num_episodes | 100 | — | — | — |

| rotation_error (deg) | 5.22 | 5.22 | 0 | ⬜ |
| runtime (min) | 3 | — | — | — |
| episode_runtime (sec) | 13 | 13 | 0 | ⬜ |
| num_episodes | 100 | — | — | — |

| match_steps | 39 | 39 | 0 | ⬜ |
| rotation_error (deg) | 53.2 | 53.2 | 0 | ⬜ |
| runtime (min) | 4 | — | — | — |
| episode_runtime (sec) | 21 | 21 | 0 | ⬜ |
| num_episodes | 100 | — | — | — |

### base_10simobj_surf_agent

| percent_correct (%) | 98.57 | 98.57 | 0 | ⬜ |
| used_mlh (%) | 2.14 | 2.14 | 0 | ⬜ |
| match_steps | 49 | 49 | 0 | ⬜ |
| rotation_error (deg) | 2.39 | 2.39 | 0 | ⬜ |
| runtime (min) | 5 | — | — | — |
| episode_runtime (sec) | 16 | 16 | 0 | ⬜ |
| num_episodes | 140 | — | — | — |

| percent_correct (%) | 97 | 97 | 0 | ⬜ |
| used_mlh (%) | 26 | 26 | 0 | ⬜ |
| match_steps | 177 | 177 | 0 | ⬜ |
| rotation_error (deg) | 20.04 | 20.04 | 0 | ⬜ |
| runtime (min) | 6 | — | — | — |
| episode_runtime (sec) | 30 | 25 | -5 | ✅ |
| num_episodes | 100 | — | — | — |

| percent_correct (%) | 98 | 98 | 0 | ⬜ |
| used_mlh (%) | 27 | 27 | 0 | ⬜ |
| match_steps | 166 | 166 | 0 | ⬜ |
| rotation_error (deg) | 16.89 | 16.89 | 0 | ⬜ |
| runtime (min) | 16 | — | — | — |
| episode_runtime (sec) | 116 | 125 | 9 | ❌ |
| num_episodes | 100 | — | — | — |

| percent_correct (%) | 68 | 68 | 0 | ⬜ |
| used_mlh (%) | 68 | 68 | 0 | ⬜ |
| match_steps | 14 | 14 | 0 | ⬜ |
| rotation_error (deg) | 92.93 | 92.93 | 0 | ⬜ |
| runtime (min) | 5 | — | — | — |
| episode_runtime (sec) | 7 | 7 | 0 | ⬜ |
| num_episodes | 100 | — | — | — |

| percent_correct (%) | 48.57 | 48.57 | 0 | ⬜ |
| used_mlh (%) | 1.43 | 1.43 | 0 | ⬜ |
| match_steps | 32 | 32 | 0 | ⬜ |
| rotation_error (deg) | 37.74 | 37.74 | 0 | ⬜ |
| runtime (min) | 32 | — | — | — |
| episode_runtime (sec) | 1 | 2 | 1 | ❌ |
| num_episodes | 140 | — | — | — |

### base_77obj_dist_agent

| percent_correct (%) | 98.27 | 98.27 | 0 | ⬜ |
| used_mlh (%) | 8.66 | 8.66 | 0 | ⬜ |
| match_steps | 75 | 75 | 0 | ⬜ |
| rotation_error (deg) | 13.68 | 13.68 | 0 | ⬜ |
| runtime (min) | 12 | — | — | — |
| episode_runtime (sec) | 24 | 22 | -2 | ✅ |
| num_episodes | 231 | — | — | — |

### base_77obj_surf_agent

| used_mlh (%) | 4.76 | 4.76 | 0 | ⬜ |
| match_steps | 43 | 43 | 0 | ⬜ |
| rotation_error (deg) | 2.27 | 2.27 | 0 | ⬜ |
| runtime (min) | 12 | — | — | — |
| episode_runtime (sec) | 24 | 28 | 4 | ❌ |
| num_episodes | 231 | — | — | — |

| percent_correct (%) | 96.97 | 96.97 | 0 | ⬜ |
| used_mlh (%) | 16.88 | 16.88 | 0 | ⬜ |
| match_steps | 123 | 123 | 0 | ⬜ |
| rotation_error (deg) | 24.63 | 24.63 | 0 | ⬜ |
| runtime (min) | 20 | — | — | — |
| episode_runtime (sec) | 52 | 55 | 3 | ❌ |
| num_episodes | 231 | — | — | — |

| percent_correct (%) | 99.57 | 99.57 | 0 | ⬜ |
| used_mlh (%) | 15.58 | 15.58 | 0 | ⬜ |
| match_steps | 93 | 93 | 0 | ⬜ |
| rotation_error (deg) | 21.91 | 21.91 | 0 | ⬜ |
| runtime (min) | 35 | — | — | — |
| episode_runtime (sec) | 95 | 78 | -17 | ✅ |
| num_episodes | 231 | — | — | — |

| percent_correct (%) | 92.21 | 92.21 | 0 | ⬜ |
| match_steps | 38 | 38 | 0 | ⬜ |
| rotation_error (deg) | 47.4 | 47.4 | 0 | ⬜ |
| runtime (min) | 10 | — | — | — |
| episode_runtime (sec) | 99 | 85 | -14 | ✅ |
| num_episodes | 77 | — | — | — |

| percent_correct (%) | 98 | 98 | 0 | ⬜ |
| match_steps | 98 | 98 | 0 | ⬜ |
| runtime (min) | 10 | — | — | — |
| episode_runtime (sec) | 4 | 5 | 1 | ❌ |
| num_episodes | 100 | — | — | — |

| match_steps | 98 | 98 | 0 | ⬜ |
| runtime (min) | 23 | — | — | — |
| episode_runtime (sec) | 12 | 10 | -2 | ✅ |
| num_episodes | 100 | — | — | — |

### randrot_noise_sim_on_scan_monty_world

| percent_correct (%) | 90 | 90 | 0 | ⬜ |
| used_mlh (%) | 85 | 85 | 0 | ⬜ |
| match_steps | 444 | 444 | 0 | ⬜ |
| runtime (min) | 27 | — | — | — |
| episode_runtime (sec) | 174 | 168 | -6 | ✅ |
| num_episodes | 120 | — | — | — |

### world_image_on_scanned_model

| percent_correct (%) | 66.67 | 66.67 | 0 | ⬜ |
| used_mlh (%) | 62.5 | 62.5 | 0 | ⬜ |
| match_steps | 374 | 374 | 0 | ⬜ |
| runtime (min) | 10 | — | — | — |
| episode_runtime (sec) | 11 | 10 | -1 | ✅ |
| num_episodes | 48 | — | — | — |

### dark_world_image_on_scanned_model

| percent_correct (%) | 29.17 | 29.17 | 0 | ⬜ |
| used_mlh (%) | 18.75 | 18.75 | 0 | ⬜ |
| match_steps | 242 | 242 | 0 | ⬜ |
| runtime (min) | 8 | — | — | — |
| episode_runtime (sec) | 9 | 9 | 0 | ⬜ |
| num_episodes | 48 | — | — | — |

### bright_world_image_on_scanned_model

| percent_correct (%) | 37.5 | 37.5 | 0 | ⬜ |
| used_mlh (%) | 64.58 | 64.58 | 0 | ⬜ |
| match_steps | 413 | 413 | 0 | ⬜ |
| runtime (min) | 12 | — | — | — |
| episode_runtime (sec) | 14 | 12 | -2 | ✅ |
| num_episodes | 48 | — | — | — |

### hand_intrusion_world_image_on_scanned_model

| percent_correct (%) | 43.75 | 43.75 | 0 | ⬜ |
| used_mlh (%) | 18.75 | 18.75 | 0 | ⬜ |
| match_steps | 272 | 272 | 0 | ⬜ |
| runtime (min) | 9 | — | — | — |
| episode_runtime (sec) | 10 | 9 | -1 | ✅ |
| num_episodes | 48 | — | — | — |

### multi_object_world_image_on_scanned_model

| percent_correct (%) | 25 | 25 | 0 | ⬜ |
| used_mlh (%) | 2.08 | 2.08 | 0 | ⬜ |
| match_steps | 158 | 158 | 0 | ⬜ |
| runtime (min) | 6 | — | — | — |
| episode_runtime (sec) | 7 | 6 | -1 | ✅ |
| num_episodes | 48 | — | — | — |

### base_infer_objects_with_stickers_monolithic_models

| correct_child_or_parent (%) | 78.86 | — | — | — |
| used_mlh (%) | 92.57 | 92.57 | 0 | ⬜ |
| match_steps | 97 | 97 | 0 | ⬜ |
| rotation_error (deg) | 39.88 | 39.88 | 0 | ⬜ |
| avg_prediction_error | 0.2 | — | — | — |
| runtime (min) | 60 | — | — | — |
| num_episodes | 350 | — | — | — |

### base_infer_objects_with_stickers_comp_models

| correct_child_or_parent (%) | 70.86 | — | — | — |
| used_mlh (%) | 43.14 | 43.14 | 0 | ⬜ |
| match_steps | 32 | 32 | 0 | ⬜ |
| rotation_error (deg) | 49.49 | 49.49 | 0 | ⬜ |
| avg_prediction_error | 0.3 | — | — | — |
| runtime (min) | 16 | — | — | — |
| num_episodes | 350 | — | — | — |

</details>


.github/workflows/monty.yml
```diff
@@ -15,9 +15,6 @@ on:
     types:
       - checks_requested
 
-env:
-  UV_PYTHON: 3.10.17 # version of Python to use in `uv` environment
```
jeremyshoemaker, line 19: I was able to simplify how the `uv` environment is set up, and part of that is [using a `.python-version` file, as suggested by the documentation](https://docs.astral.sh/uv/concepts/python-versions/#python-version-files). That means we don't need this anymore.

src/tbp/monty/frameworks/models/salience/sensor_module.py
```diff
@@ -158,7 +158,7 @@ def _normalize_salience(self, weighted_salience: np.ndarray) -> np.ndarray:
         min_ = weighted_salience.min()
         max_ = weighted_salience.max()
         scale = max_ - min_
-        if np.isclose(scale, 0):
+        if np.allclose(scale, 0):
```
jeremyshoemaker: ~In some instances, maybe all of them, `scale` can be an array, and newer NumPy won't silently convert a single element 1D array into a scalar.~

**Edit:** This whole statement is actually wrong. I'd forgotten what originally lead me to this by the time I wrote this. See the replies below.

src/tbp/monty/frameworks/models/buffer.py
```diff
@@ -342,26 +342,22 @@ def get_all_features_on_object(self):
             f"Observed {np.array(self.locations).shape} locations, "
             f"{len(global_on_object_ids)} global on object"
         )
-        for input_channel in self.features:
+        for channel_name, channel in self.features.items():
             # Here we want to make sure the input-specific observation was on the object
-            channel_off_object_ids = np.where(
-                self.features[input_channel]["on_object"] is False
-            )[0]
+            channel_off_object_ids = np.where(channel["on_object"] == 0)[0]
```
jeremyshoemaker: This line is the actual functional change here.

I'm pretty sure the original version of this was a bug, `is` will always compare objects not values, and a NumPy array will never be the same object as `False`, so this would always return `False`, and newer versions of NumPy won't convert a scalar to an array inside `np.where`. Looking at where `channel["on_object"]` comes from, the values can only be 0 or 1, so the correct test is to look for 0's.

Everything else in this file is a refactor to use `items()` to pull off both the key and value from the dictionary in the `for` loop instead of using the key to look up the value over and over again.

tests/unit/frameworks/utils/logging_utils_test.py
```diff
@@ -37,7 +37,7 @@ def test_min_error_from_list(self):
         The target rotation is 12°, so the closest in the list [0°, 10°, 20°] is 10°,
         with a 2° error.
         """
-        rotations_list = Rotation.from_euler("z", [0, 10, 20], degrees=True)
+        rotations_list = Rotation.from_euler("z", 10, degrees=True)
```
jeremyshoemaker: The underlying SciPy `Rotation.from_euler` method has changed to not allow passing in more values than described by the first argument.

I'm not certain this is correct, or what the original behaviour was doing. 10 is the correct value to make the test pass and matches what the docstring says, but I don't see how 10 was the Z-axis value in the original.

src/tbp/monty/frameworks/utils/sensor_processing.py
```diff
@@ -424,6 +424,16 @@ def surface_normal_total_least_squares(
         try:
             # Find eigenvector of M with the minimum eigenvalue
             eig_val, eig_vec = np.linalg.eig(m_mat)
+            # Newer versions of `np.linalg.eig` always return complex values instead
+            # of converting them to floats when they have a zero imaginary component,
+            # so we need to explicitly convert to a real. We're assuming that the
+            # eigenvectors and eigenvalues have no imaginary components.
+            # TODO: change this comment to a mathematical argument for why
+            #  this is a safe assumption.
```
jeremyshoemaker: I'd like to be able to explain why we can make this assumption about the eigenvalues and eigenvectors being real values.

src/tbp/monty/frameworks/utils/sensor_processing.py
```diff
@@ -682,9 +698,16 @@ def principal_curvatures(
...
             eigval, eigvec = np.linalg.eig(m)
-            idx = eigval.argsort()[::-1]
-            eigval_sorted = eigval[idx]
-            eigvec_sorted = eigvec[:, idx]
+            # Newer versions of `np.linalg.eig` always return complex values instead
+            # of converting them to floats when they have a zero imaginary component.
+            # We're assuming that the eigenvectors and eigenvalues have no imaginary
+            # components.
+            # TODO: change this comment to a mathematical argument for why
+            #  this is a safe assumption.
```
jeremyshoemaker: Same with this code, if the mathematical justification happens to be different than the previous change.

src/tbp/monty/geometry.py
```diff
@@ -239,9 +239,14 @@ def inv(self) -> Rotation:
...
         Examples:
+            TODO: The use of `bool()` in the example below is to work around newer
+              versions of NumPy changing the representation of `np.True_`, in a
+              backwards-compatible way. Remove it and update the test once we fully
+              upgrade to Python 3.13.
+
             >>> fwd = Rotation.from_euler("y", -np.pi/6)  # yaw right 30°
             >>> inv = fwd.inv()
-            >>> inv.inv().approx_equal(fwd)
+            >>> bool(inv.inv().approx_equal(fwd))
```
jeremyshoemaker, line 249: See this [NumPy issue](https://github.com/numpy/numpy/issues/26774) for details.

src/tbp/monty/simulators/habitat/simulator.py
```diff
@@ -180,7 +180,7 @@ def __init__(
                 # json files with the attributes of each object in the dataset.
                 # The json file name is in this format:
                 # "{object_name}.object_config.json".
-                # See https://aihabitat.org/docs/habitat-sim/attributesJSON.html#objectattributes # noqa: E501
+                # See https://aihabitat.org/docs/habitat-sim/attributesJSON.html#objectattributes
```
jeremyshoemaker, line 183: I ran the newer version of `ruff` on the full code base, similar to how we currently run the check in our Actions, to make sure there weren't any new rules flagging things and it flagged this line as an [unused-noqa](https://docs.astral.sh/ruff/rules/unused-noqa/).

The [line-too-long](https://docs.astral.sh/ruff/rules/line-too-long/) rule ignores lines that end in a URL so long as the URL starts within the allowed width.

I'm not sure why the older version of `ruff` doesn't flag this, as neither of these rules appears to be new. But the older version also doesn't complain when I remove this.

jeremyshoemaker:
@scottcanoe, @hlee9212, I assigned this to both of you since part of this involves a change to something salience related, and you both seem to have a good foundation in using NumPy, and can verify I'm not breaking anything.

tests/unit/frameworks/utils/logging_utils_test.py
hlee9212: **issue**: This should be updated to:

```python
rotation_list = Rotation.from_euler("z", [[0], [10], [20]], degrees=True)
```

We cannot just change to 10 because the original test is testing whether we can get the minimum error between 12 deg (target) and [0 deg, 10 deg, 20 deg]. If we just pass 10, then we are not testing whether we are calculating min error. 

Source: I verified in REPL using the new environment.

jeremyshoemaker: Oh, okay, now I get what the original was doing. Thanks!

jeremyshoemaker: That also makes the original error I was seeing make sense:
```
ValueError: Expected last dimension of `angles` to match number of sequence axes specified, got 3.
```

src/tbp/monty/frameworks/models/salience/sensor_module.py
hlee9212: **issue**: I don't think it is necessary to change from `np.isclose()` to `np.allclose()`, because `scale` is always scalar and not array. This is because we are taking `.min()` and `.max()` (line 158 and 159), which should always give scalar regardless of what `weighted_salience` is.

jeremyshoemaker: Without that change, the `SalienceSMTest` fails with this error, from this line:
```
ValueError: The truth value of an empty array is ambiguous. Use `array.size > 0` to check that an array is not empty.
```

I don't think the change it is suggesting is the right one. When I wrote my original comment, I'd forgotten the details. I should have gone back and pulled this out up front.

jeremyshoemaker: Oh, it's because `scale` is a Mock, and then calling `np.isclose(scale, 0)` returns an `np.array` of `[]`, which has an ambiguous boolean value.

jeremyshoemaker: So, this is the train of thought I went through to get where it is now. `np.isclose` will return a scalar if both values are scalars, but since in this test case `scale` is a Mock, it returns an `np.array` instead.

I'm less happy with the solution of switching to `np.allclose` now that I'm looking at it again, but I'm not sure how easy it would be to change the value in the test so that it's not a Mock. It's pretty deep down from where the original Mock comes from.

scottcanoe: I'm assuming the failing test was calling `SalienceSM._normalize_salience` indirectly? As in, the whole pipeline runs on mocked input, and the results of that stuff is being sent to `_normalize_salience`?

I'll not hold up the PR for it, but I'll see if I can reproduce the failure if I have a minute later today.

src/tbp/monty/frameworks/models/buffer.py
scottcanoe: Yup, looks like a big to me too. Nice catch.

src/tbp/monty/frameworks/utils/sensor_processing.py
scottcanoe: As would I. It looks like a couple of these functions are untested. Maybe one day I'll try to understand how this construction works.

Side note: I saw something I didn't like in this file, so I'm pushing a commit. Basically, `surface_normal_ordinary_least_squares` and `surface_normal_total_least_squares` are both copying their input point cloud, but `center_neighbors` is actually the place where input is modified. Luckily, `center_neighbors` is only being called by those two functions right now. But it seemed prudent to relocate the copy into the place where the mutation actually happens. I'll push it after this review.

tests/unit/frameworks/utils/logging_utils_test.py
```diff
@@ -37,7 +37,7 @@ def test_min_error_from_list(self):
         The target rotation is 12°, so the closest in the list [0°, 10°, 20°] is 10°,
         with a 2° error.
         """
-        rotations_list = Rotation.from_euler("z", [0, 10, 20], degrees=True)
```
scottcanoe, line 40: Is this due to a scipy API change?

scottcanoe (commented):
Cool. Looking forward to switching over to 3.13 so I can write my first match statement.

scottcanoe (approved):
@jeremyshoemaker I didn't push my commit because I wasn't sure about the etiquette. I'll just open a small PR next week and assign you. Hopefully it's still fresh in our heads on Monday.

tests/unit/frameworks/utils/logging_utils_test.py
jeremyshoemaker: I haven't looking through the SciPy release notes to find out, but I suspect it ultimately comes from NumPy's changes to their strictness about mixing up scalars and single element arrays.

src/tbp/monty/frameworks/models/salience/sensor_module.py
jeremyshoemaker: Yeah, it's calling it indirectly. The test I was using to investigate it is `SalienceSMTest.test_step_returns_no_percept`, which is just calling step with a mocked observation. That observation makes more Mocks as things are accessed and called on it, ultimately ending up in this line of code with all the values and arguments being various Mocks.

tests/unit/frameworks/utils/logging_utils_test.py
jeremyshoemaker: Hojae and I discussed this change in a [resolved comment thread](https://github.com/thousandbrainsproject/tbp.monty/pull/1099#pullrequestreview-4698679159).

jeremyshoemaker:
> @jeremyshoemaker I didn't push my commit because I wasn't sure about the etiquette. I'll just open a small PR next week and assign you. Hopefully it's still fresh in our heads on Monday.

If there's anyone on the team whose branches you could push to without completely throwing them off, it's me.

It sounds like the change you want to make is out of scope of this PR though.

tests/unit/frameworks/utils/logging_utils_test.py
jeremyshoemaker: It looks like it's new in 1.17.0.

> The resulting shapes for `transform.Rotation.from_euler` / `from_davenport` have changed to make them consistent with broadcasting rules. Angle inputs to Euler angles must now strictly match the number of provided axes in the last dimension. 
> ~ [SciPy 1.17.0 release notes](https://docs.scipy.org/doc/scipy/release/1.17.0-notes.html#scipy-spatial-improvements)

src/tbp/monty/frameworks/models/buffer.py
hlee9212: Very nice catch. 

**suggestion (non-blocking)**: It seems like `channel_off_object_ids` is just used to print as part of logger.debug(...). The following is a bit more concise and readable to me, but not a blocker.

```python
channel_on_object_count = np.count_nonzero(channel["on_object"] == 1)
logger.debug(
    f"{channel_name} has {channel_on_object_count} on object observations"
)
```

src/tbp/monty/frameworks/models/salience/sensor_module.py
hlee9212: I see. It seems like the better fix is passing a concrete empty salience instead of `MagicMock()`. 

I think line 45 in `sensor_module_test.py` (the `mocked_object_observation():`) may be updated as follows:

```python
@pytest.fixture
def mocked_object_observation():
    empty_observation = OnObjectObservation(center_location=None, locations=np.empty((0, 3)), salience=np.array([]),
    with patch(
        "tbp.monty.frameworks.models.salience.sensor_module.on_object_observation",
        return_value=empty_observation
    ):
        yield
```

We'll also need to update `setUp()` with something like:
```python
def setUp(self) -> None:
    empty_salience = np.array([]))
    self.sensor_module = SalienceSM(
        # change salience_strategy and return_inhibitor as:
       salience_strategy = MagicMock(return_value=empty_salience),
       return_inhibitor = MagicMock(return_value=empty_salience),
      # keep the other arguments
    )
    self.observation = SensorObservation(
        rgba = np.zeros((1, 1, 4), dtype=np.uint8),
        depth = np.zeros((1, 1)),
```

Then I think we can replace `SalienceSMTest.test_step_returns_no_percept` with actual empty observation instead of MagicMock, I think: 
```python
self.assertIsNone(self.sensor_module.step(self.ctx, self.observation)) # <- replace MagicMock()
```

---
**suggestion (if-minor)**: My preference is to keep `np.isclose(scale, 0)` and make the above change. However, if the above change itself fails tests or brings up new issues, I'm okay if someone wants to make a follow-up PR.

src/tbp/monty/frameworks/utils/sensor_processing.py
hlee9212: :D Well I'm not opposed to introducing one of my absolute favorite theorems into the Monty codebase...

[Spectral Theorem](https://people.math.harvard.edu/~knill/teaching/math22b2019/handouts/lecture17.pdf) states that **any real, symmetric matrix** (or Hermitian) can be diagonalized by an orthogonal matrix, i.e. it has real eigenvalues and orthonormal eigenvectors. (This is one of the most important and useful results out there)

So let's check if `m_mat` is real and symmetric. The code says:
```python
m_mat = 1 / n_points * np.matmul(x_mat.T, x_mat) - np.matmul(p_mean, p_mean.T)
```

In LaTeX, that's `m_mat` = $\frac{1}{n} X^\mathrm{T}X - pp^\mathrm{T}$. 

Is `m_mat` real? Yes, because `x_mat` is coming from our point cloud which are in real 3D space, and `p_mean` is some average of `x_mat`, so `p_mean` is also real.

Is `m_mat` symmetric? Yes, because $X^\mathrm{T}X$ is symmetric and $pp^\mathrm{T}$ is symmetric and the difference between symmetric matrices is another symmetric matrix. 

So, by spectral theorem a real symmetric matrix has real eigenvalues and real eigenvectors. (And this doesn't matter to us, but our eigenvectors are also orthonormal, which is an amazing result, but I digress). 

---
**suggestion for TODO**: 
```
# `m_mat` is real and symmetric because both `x_mat.T @ x_mat`
# and `p_mean @ p_mean.T` are real symmetric matrices. By the
# spectral theorem, its eigenvalues are real and it admits an
# orthonormal basis of real eigenvectors, so discarding the zero
# imaginary components is safe.
```

**suggestion for code**: My preferred code block would be:
```python
try:
    # `m_mat` must be real and symmetric. Both `x_mat.T @ x_mat` and
    # `p_mean @ p_mean.T` are symmetric by construction.
    if not np.allclose(m_mat, m_mat.T):
        raise np.linalg.LinAlgError("Surface-normal covariance matrix is not symmetric")

    # `eigh` is specialized for real symmetric matrices. Its eigenvalues
    # are real and returned in ascending order.
    eig_val, eig_vec = np.linalg.eigh(m_mat)

    # The eigenvector with the smallest eigenvalue is the direction of
    # minimum point-cloud variation: the normal to the best-fit plane.
    n_dir = eig_vec[:, 0] 
    valid_sn = True
```
While the `np.real(..)` technically works, it's just very "code smelly". I realize Spectral Theorem is not necessarily common knowledge outside of deep learning (which relies on linear algebra), but common enough that I'd gotten interview questions about it. 

I'll leave it up to you. It's not required for supporting newer library, and the above will just be in the backlog of things I want to make improvements to `sensor_processing.py` file / MISC Monty codebase improvement.

src/tbp/monty/frameworks/utils/sensor_processing.py
hlee9212: Yes, the justification is slightly different here because `m` is **not** a symmetric matrix in this case. 

This one is more complicated. The code is constructing $M = G^{-1}B$ (where G is `guv` and B is `buv`). This is not a symmetric matrix of it's own (you can check by looking at the transpose), but it is **similar** to a symmetric matrix. 

By "similar" in linear algebra, it means it can be written as a linear transformation of a real symmetric matrix $S= G^{-\frac{1}{2}} B G^{-\frac{1}{2}}$ (and we can do this because G is a real symmetric positive definite matrix... which means it has positive eigenvalues and therefore we can take square root of the matrix - I know, the justification goes deep). 

So in this case, because M is similar to a symmetric matrix S, we can apply Spectral Theorem again to S to justify that we have real eigenvalues and eigenvectors. 

Equivalently, this is not solving the usual eigendecomposition problem you have seen in linear algebra, i.e.:
$$Bv = \lambda v$$
But something called **generalized** decomposition problem:
$$Bv = \lambda G v$$ 

(The usual eigendecomposition we know is just where G = Identity matrix) 

This is why I suggested we use [`scipy.linalg.eigh`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.eigh.html) because it can do generalized eigendecomposition unlike Numpy. 

---
Honestly I'm not sure what the "concise" comment would be. 😅 Below is my preferred approach.

```python
# Solve the symmetric-definite generalized eigenvalue problem:
#
#     buv @ v = k * guv @ v
#
# `buv` is real symmetric and `guv` is real symmetric positive
# definite, so the generalized spectral theorem guarantees real
# eigenvalues and a complete basis of real eigenvectors.
eigval, eigvec = scipy.linalg.eigh(buv, guv)

# `eigh` returns eigenvalues in ascending order. Preserve the existing
# convention of ordering principal curvatures from largest to smallest.
eigval_sorted = eigval[::-1]
eigvec_sorted = eigvec[:, ::-1]
```

hlee9212 (approved):
Submitting Review/Approving via PyCharm because the GitHub website keeps dying on me. Hope this works.

Overall I gave some alternative code suggestions related to `sensor_processing.py`, but I don't think these are blockers. 

Thanks for updating these!

hlee9212:
@scottcanoe This is a side note but I think in all experiments we always use TLS estimation for surface normal, so `surface_normal_ordinary_least_squares()` is technically dead code (and so is the Naive estimation method). 

The reason why I haven't gone to delete this is that while TLS is the superior estimation method for most patches we encounter in Monty, technically OLS can have it's place (I think when there is a lot of depth noise) occasionally. 

I don't know, we can also delete the dead code for now and then just later revive it whenever we find ourselves in a need to improve surface normal accuracy for different cases. I'll try to remember to ask in TRM whether it may be alright to delete OLS/Naive estimation methods for now.

scottcanoe:
@hlee9212 Sounds good, thanks for the context. Always happy to delete dead code. Even if we do that though, I wouldn't be opposed to merging my changes before doing so in case the resurrection does take place.

src/tbp/monty/frameworks/models/buffer.py
jeremyshoemaker: Oh, that would explain why it hasn't broken anything. I guess we haven't looked closely at the output of this.

I do like that simplification though. It saves us creating an array that we don't really need.

src/tbp/monty/frameworks/utils/sensor_processing.py
jeremyshoemaker: I prefer this because it's getting to the heart of why this is a safe operation. Thanks for the detailed explanation, it's exactly what I was looking for.

jeremyshoemaker:
@hlee9212 Do you want to give this another look now that I've updated it with changes addressing your feedback?

src/tbp/monty/frameworks/utils/sensor_processing.py
```diff
@@ -421,26 +426,35 @@ def surface_normal_total_least_squares(
         p_mean = 1 / n_points * np.mean(x_mat, axis=0, keepdims=True).T
         m_mat = 1 / n_points * np.matmul(x_mat.T, x_mat) - np.matmul(p_mean, p_mean.T)
 
+        # The matrix `m_mat` must be real and symmetric as both `x_mat.T @ x_mat`
+        # and `p_mean @ p_mean.T` are symmetric by construction. This follows
+        # from the spectral theorem.
+        if not np.allclose(m_mat, m_mat.T):
+            logger.debug("Surface normal covariance matrix is note symmetric.")
+            return n_dir, valid_sn
```
jeremyshoemaker, line 435: I get linting errors if I try to use the `raise` approach you suggested, so I refactored this a little.

hlee9212: **typo**: "note" should be "not". :)

jeremyshoemaker: 🤦 Good catch.

src/tbp/monty/frameworks/utils/sensor_processing.py
```diff
@@ -681,10 +701,20 @@ def principal_curvatures(
...
-            eigvec_sorted = eigvec[:, idx]
+
+            # Solve the symmetric-definite generalized eigenvalue problem:
+            #
+            #     buv @ v = k * guv @ v
+            #
+            # `buv` is real symmetric and `guv` is real symmetric positive
+            # definite, so the generalized spectral theorem guarantees real
+            # eigenvalues and a complete basis of real eigenvectors.
+            eigval, eigvec = np.linalg.eigh(m)
```
hlee9212: **issue**: This needs to be updated to scipy.linalg.eigh(buv, guv).

`m` here is **not** symmetric (may try to edit this comment to see if I can explain better).

hlee9212 (changes requested):
One blocking comment, but otherwise looking good.

src/tbp/monty/frameworks/utils/sensor_processing.py
jeremyshoemaker: Thanks. I see I didn't catch that in your suggestion.

jeremyshoemaker: Fixed

hlee9212: Okay, so I've written [in previous comment](https://github.com/thousandbrainsproject/tbp.monty/pull/1099#discussion_r3607480322) that `m` is not a symmetric matrix. Note that a symmetric matrix is one where $A = A^T$. 

Some facts:
- `buv` (B) and `guv` (G) are symmetric matrices.
- `m = np.linalg.inv(guv).dot(buv)` is not a symmetric matrix (multiplying two symmetric matrices do not end up in a symmetric matrix). This is equivalent to $M = G^{-1}B$
- `m` is [similar](https://en.wikipedia.org/wiki/Matrix_similarity) to a symmetric matrix. This is a very specific definition of "similar" that I'm using here. 
- Solving for eigenvalues of `m` is equivalent to solving for the [**generalized** eigendecomposition problem](https://en.wikipedia.org/wiki/Eigendecomposition_of_a_matrix#Generalized_eigenvalue_problem). In my previous comment I said generalized eigendecomp solves:

$$Bv = \lambda Gv$$

Multiply both sides by $G^{-1}$:

$$G^{-1}Bv = \lambda v$$

We just said that $M = G^{-1}B$, so the above is $Mv = \lambda v$, which is equivalent to using `np.linalg.eig(m)`, but **not** eigh because M is not symmetric.

The `scipy.linalg.eigh(buv, guv)` allows us to solve the general eigendecomposition problem - the API allows us to pass in both arguments, unlike `np.linalg.eig` or `np.linalg.eigh`.

jeremyshoemaker: Thanks for the explanation. I'm pretty clear on the math, I just keep flubbing copying what you actually wrote originally.

hlee9212: Was just about give approval but then noticed it was:

`eigval, eigvec = scipy.linalg.eigh(m)`

Should be:

`eigval, eigvec = scipy.linalg.eigh(buv, guv)`

We don't need to define `m = np.linalg.inv(guv) @ buv` at all, actually.

jeremyshoemaker: It _should_ be correct this time, unless I still missed something.

hlee9212 (approved):
Looks great! :D

PR #1104, opened 2026-07-16
fix: outline the view finder for 3-channel RGB images, not only RGBA

foxalike-dev:
## Summary

`add_patch_outline_to_view_finder` hard-codes a 4-element RGBA colour `[0, 0, 255, 0]` when drawing the patch outline on the view-finder image. When the view finder is a 3-channel RGB image, this raises:

```
ValueError: could not broadcast input array from shape (4,) into shape (..., 3)
```

It surfaces whenever the live plotter / teleop viewer is run on a data source whose view finder is RGB rather than RGBA (e.g. real RGB-D frames).

## Change

Size the outline colour to the image's channel count (`[0, 0, 255, 0][:n_channels]`):

- 3-channel RGB → `[0, 0, 255]` (blue outline, no crash)
- 4-channel RGBA → `[0, 0, 255, 0]` (byte-for-byte unchanged)

Added `tests/unit/frameworks/utils/plot_utils_test.py` covering both cases.

## Testing

- `pytest tests/unit/frameworks/utils/plot_utils_test.py` — 2 passed
- `ruff check` / `ruff format --check` — clean


jeremyshoemaker (approved):
Thanks for submitting this fix. This looks good to me.

PR #1068, opened 2026-06-27
refactor!: create Monty from dehydrated configs

dschumacher-tbp:
As part of the epoch/episode removal, we want to re-create Monty from dehydrated configs whenever the Experiment chooses to do so. This PR introduces an `MontyExperiment._recreation_config` property that is used to inject the dehydrated Monty config. The easiest way to properly set this property is to create the Experiment using `tbp.monty.hydra.hydrate_experiment()` instead of `hydra.utils.instantiate()`. The tests have all been updated to use this method.

The initial Monty instance is created using `with exp:` on the Experiment. This eventually calls `MontyExperiment.setup_experiment()`, which uses `_recreation_monty_factory()` to instantiate the model. The `_recreation_monty_factory()` performs a bespoke construction processed based on the (now removed) `init_model()` algorithm, and sets the `model` property of the `MontyExperiment`.

## Breaking Changes
- `MontyExperiment._recreation_config` must be set to the dehydrated Monty config
- `MontyExperiment` instances should be created using `tbp.monty.hydra.hydrate_experiment()` instead of `hydra.utils.instantiate()`.

### base_config_10distinctobj_dist_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100 | 100 | 0 | ⬜ |
| used_mlh (%) | 0 | 0 | 0 | ⬜ |
| match_steps | 28 | 28 | 0 | ⬜ |
| rotation_error (deg) | 10.29 | 10.29 | 0 | ⬜ |
| runtime (min) | 3 | 2 | -1 | ✅ |
| episode_runtime (sec) | 5 | 5 | 0 | ⬜ |
| num_episodes | 140 | 140 | 0 | ⬜ |

### base_config_10distinctobj_surf_agent

| Metric | Baseline | Proposed | Δ | Result |
| --- | --- | --- | --- | --- |
| percent_correct (%) | 100 | 100 | 0 | ⬜ |
| used_mlh (%) | 0 | 0 | 0 | ⬜ |
| match_steps | 28 | 28 | 0 | ⬜ |
| rotation_error (deg) | 3.64 | 3.64 | 0 | ⬜ |
| runtime (min) | 4 | 4 | 0 | ⬜ |
| episode_runtime (sec) | 12 | 11 | -1 | ✅ |
| num_episodes | 140 | 140 | 0 | ⬜ |

### randrot_noise_10distinctobj_dist_agent

[...]

PR #1106, opened 2026-07-21
fix: `GraphLearningTest::test_5lm_feature_experiment` in both modes

dschumacher-tbp:
This training/evaluation strategy in this test had numerous incompatibilities with `Experiment._recreation_mode = True`. The fake observations didn't align with the model configuration and the learning-module wrappers injected into Monty were destroyed by `Experiment.pre_episode()`.

With these changes, the test to passes with `Experiment._recreation_mode = False` (the default), and `Experiment._recreation_mode = True`.


src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -408,7 +408,7 @@ def get_epoch_state(self):
     def _create_monty(self) -> None:
         """Create a Monty model from dehydrated config.
 
-        **WARNING:** `self._recreation_config` must be initialized
+        **WARNING:** `self._monty_cfg` must be initialized
         with the dehydrated config before calling this method.
```
jeremyshoemaker, line 412: Maybe we should check for this and raise an error if `self._monty_config` is none. This is an invariant that must be true, otherwise we'll end up with an error somewhere else that will be more confusing.

Actually, it looks like it will come from the first line:
```
TypeError: 'NoneType' object is not iterable
```

It could just be a ValueError with an appropriate message, e.g. the contents of this comment. I think this is more of a coding error, so making a custom exception for it is probably not needed.

tests/integration/frameworks/models/graph_learning_test.py
```diff
@@ -618,31 +628,25 @@ def get_gm_with_fake_object(self):
...
-                    "patch": {
-                        "hsv": [0.1, 1, 1],
-                        "principal_curvatures_log": [1, 1],
-                    }
-                },
-            )
+            # wrap already-configured `graph_lm`
+            graph_lm = graph_lms[lm]
             object_obs = []
+            sender_id = f"patch_{lm}"
```
jeremyshoemaker: This is probably better with `enumerate()`. Could change to this and cleanup the rest of the loop code.
```python
for lm_num, graph_lm in enumerate(graph_lms):
    object_obs = []
    sender_id = f"patch_{lm_num}"
```
Maybe that should be `lm_id`? I can't remember how we identify these.

tests/integration/frameworks/models/graph_learning_test.py
```diff
@@ -938,48 +942,50 @@ def test_5lm_feature_experiment(self):
...
             exp.post_epoch()
 
         output_dir = Path(exp.output_dir)
         eval_stats = pd.read_csv(output_dir / "eval_stats.csv")
-        # Just testing 1 episode here. Somehow the second rotation doesn't get
-        # recognized. Probably just some parameter setting due to flaws in old
-        # LM but didn't want to dig too deep into that for now.
         self.check_multilm_eval_results(
-            eval_stats, num_lms=5, min_done=3, num_episodes=1
+            eval_stats, num_lms=5, min_done=3, num_episodes=2
```
jeremyshoemaker: Now that this works with both episodes, I think it would be better if you restored this line above, and move it up a line outside the `with exp`...
```python
objects = [self.fake_obs_learn, self.fake_obs_house_3d]
```
...and then change this to set `num_epsiodes=len(objects)`.

I spent a lot of time trying to figure out why this was 2 and not the default of 3, and thought there might be an issues elsewhere, before realizing we only have 2 episodes, which comes from how many objects we have.

jeremyshoemaker (changes requested):
The broad strokes look fine. I just have a few requests plus, as we discussed, updating the MuJoCo version of this test.

src/tbp/monty/frameworks/experiments/monty_experiment.py
dschumacher-tbp: We discussed this in [PR #1068](https://github.com/thousandbrainsproject/tbp.monty/pull/1068) and decided to "let it crash". I'm just fixing the comment to match the renamed property.

tests/integration/frameworks/models/graph_learning_test.py
```diff
@@ -957,7 +960,7 @@ def test_5lm_feature_experiment(self):
             # FIXME: snapshot must be initialized after training
             exp._snapshot_monty()
 
-            for episode_num in range(tm0.num_episodes):
+            for episode_num in range(num_episodes):
```
jeremyshoemaker, line 963: I was trying to come up with a way to improve this loop. This is a step in the right direction.

src/tbp/monty/frameworks/experiments/monty_experiment.py
jeremyshoemaker: I must still have mixed feelings about it, if I'm bringing it up again. I know I get frustrated when I run into this type of error, what is essentially a null pointer exception, because then I have to figure out where the null came from.

I guess the place it gets raised is close enough to this comment to probably make it okay. I think if it was me, I'd add the check, but I'm not going to press you on it.

tests/integration/frameworks/models/graph_learning_test.py
```diff
@@ -936,50 +941,55 @@ def test_5lm_displacement_experiment(self):
...
 
-            for episode_num in range(tm.num_episodes):
+            for episode_num in range(num_episodes):
                 exp.pre_episode()
-                # Normally the experiment `pre_episode` method would call the model
-                # `reset` method, but it expects to feed data from an environment
-                # interface to the model, and we aren't using that, so we call it again
-                # with the correct target value.
-                monty.reset()
+                # When `exp._recreation_mode = True`, `exp.model` is replaced
```
jeremyshoemaker, line 965: This is more of a nitpick. I'd probably add "so we need to fetch it again" to the end of this sentence. I'm might be overly verbose with my comments though.

jeremyshoemaker (approved):
At this point, it looks good to me. I added some comments, but if you're ready to merge this, let me know and I'll do that.

PR #1107, opened 2026-07-21
feat: simple overall stats logging in run_parallel

jeremyshoemaker:
There's nothing fancy here. It's just the simplest thing that works.

There _is_ a bug fix here as well for a situation in which `log_parallel_wandb` is `False`. In that case `wandb_run` is never defined. I also renamed it to be clear that it was WandB related.

### DIFF
--- a/src/tbp/monty/frameworks/run_parallel.py
+++ b/src/tbp/monty/frameworks/run_parallel.py
@@ -468,7 +468,7 @@ def print_benchmark_stats(overall_stats: dict) -> None:
         "overall/num_episodes",
     ]
     benchmark_stats = [overall_stats[k] for k in benchmark_keys]
-    comma_keys = ",".join([k[8:] for k in benchmark_keys])
+    comma_keys = ",".join([k[len("overall/"):] for k in benchmark_keys])
     comma_stats = ",".join([str(s) for s in benchmark_stats])
     print(f"Benchmark Stats\n{comma_keys}\n{comma_stats}")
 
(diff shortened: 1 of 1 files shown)


jeremyshoemaker:
The more I poke at this section of code, the more I see things that really need to be decoupled from `log_parallel_wandb`. I could simplify this code a lot, but that one config option has deep tendrils into the rest of the system, see `single_evaluate`.

src/tbp/monty/frameworks/run_parallel.py
```diff
@@ -456,6 +456,23 @@ def get_overall_stats(stats):
...
+        "overall/percent_correct_child_or_parent",
+        "overall/avg_rotation_error",
+        "overall/avg_num_monty_matching_steps",
+        "overall/run_time",
+        "overall/avg_episode_run_time",
+        "overall/percent_used_mlh_after_timeout",
+        "overall/num_episodes",
+    ]
+    benchmark_stats = [overall_stats[k] for k in benchmark_keys]
+    comma_keys = ",".join([k[8:] for k in benchmark_keys])
```
tristanls-tbp: **issue**: magic number

**suggestion**:
```suggestion
    comma_keys = ",".join([k[len("overall/"):] for k in benchmark_keys])
```

tristanls-tbp (commented):
One magic number issue. Other than that, I tried to break it and couldn't find a counterexample 👍.

tristanls-tbp (approved):

PR #1108, opened 2026-07-21
refactor!: Update compositional configs to use 3LM Heterarchy architecture

hlee9212:
## Summary of Breaking Changes

Major update to Monty architecture for compositionality benchmark experiments. Previously, Monty used 2 LMs in compositional benchmark experiments; with this PR, Monty uses 3 LMs. 

## Recommended File Order for Review

I suggest reviewing files in order of pretraining sequence and then inference. 

1. supervised_pre_training_objects_with_stickers_3d_children
2. supervised_pre_training_objects_with_stickers_2d_children
3. supervised_pre_training_objects_with_stickers_monolithic_models
4. supervised_pre_training_objects_with_stickers_comp_models
5. base_infer_objects_with_stickers_monolithic_models
6. base_infer_objects_with_stickers_comp_models

Many thanks to @ramyamounir for providing configs [here](https://github.com/ramyamounir/tbp.monty/commit/5fc780b8a1cfecb9952d1af0589ca337e19b7a2c) if anyone is wondering why particular feature weights or tolerances changed).

Actually @ramyamounir, I would greatly appreciate if you may be able to take a look at the updated configs when you have time? We've been getting pretty low performance (as per this [Shortcut](https://app.shortcut.com/thousandbrainsproject/story/6280/update-compositional-benchmark-configs)) and some fresh, expert pair of eyeballs would be great. (I'll also come back to it tomorrow, right now my eyes might be just glossing over some super obvious errors 😅). 

## Useful Screenshots

1. Diff showing that the inference configs between monolithic and comp are identical except for connectivity 
CleanShot 2026-07-21 at 14 30 30

1. Diff showing that the pretraining configs between monolithic and comp are nearly identical as well
CleanShot 2026-07-21 at 14 32 55


ramyamounir:
Thanks for tagging me for review, @hlee9212. I haven't looked through the changes yet, but I'm looking at the list of experiments that you've listed in the summary and I'm wondering if we even need `supervised_pre_training_objects_with_stickers_monolithic_models` at all. 

I'm thinking it would be fine to load the compositional models that include all 3 input channels in the pretrained graph and just configure the inference experiment to be monolithic. So, both inference experiments use the same compositional pretrained models, the only difference would be that the monolithic inference just doesn't get messages from LMs, but the compositional inference does. If the HL-LM doesn't get messages on the LM input channels (different connectivity config file), it just won't add to the evidence from these LM input channels. What do you think?

tristanls-tbp:
**note**: I assigned this PR to @ramyamounir per @hlee9212's review request assignment already. @jeremyshoemaker, I assigned it to you to review with the "what about MuJoCo" framing in mind.

ramyamounir:
@hlee9212 I thought it would be easier to just run the configs and debug where we don't get good performance or where things seem off. I was able to learn good child objects and then when I started training the compositional model I noticed that we learned just one surface (see image below for the SM channel). This is because of [this](https://github.com/thousandbrainsproject/feat.compositional_configs/blob/da4a15386387bc7ae6f3fb0fa125dd17e38e3bb6/src/tbp/monty/conf/experiment/supervised_pre_training_objects_with_stickers_comp_models.yaml#L15-L20) in the configs (probably leftover from my configs, sorry :sweat_smile: )

cylinder_tbp_horz_patch2


I think we would want to learn the full compositional object, especially if we do the inference on all rotations. Can you update the configs to learn the full objects and try again? Also let's just do the monolithic inference on pretrained compositional models, I don't think we need to train special monolithic models unless I'm missing something.

src/tbp/monty/conf/monty/learning_module/evidence_3lm_heterarchy.yaml
```diff
@@ -0,0 +1,83 @@
...
+  object_evidence_threshold: 5
+  # gsg:
+  #   _target_: tbp.monty.frameworks.models.goal_generation.EvidenceGoalGenerator
+  #   goal_tolerances:
+  #     location: 0.15
+  #   elapsed_steps_factor: 10
+  #   min_post_goal_success_steps: 20
+  #   x_percent_scale_factor: 0.75
+  #   desired_object_distance: 0.03
+  #   wait_growth_multiplier: 1
```
jeremyshoemaker, line 31: If we don't need this commented out config section, we should probably remove it. Same with the one below.

jeremyshoemaker (commented):
From the perspective of MuJoCo specific issues, I don't see any problems with this PR. None of the configs being changed have equivalent MuJoCo versions yet, and I don't expect those to be done by others _just_ yet.

I did notice a handful of these config files are missing a final newline. That's probably not going to break anything, but consistency would be nice.

Image

I also commented on some commented out configuration that should probably be removed.

hlee9212:
Hi @ramyamounir, thanks so much! 🙏

> I think we would want to learn the full compositional object, especially if we do the inference on all rotations. Can you update the configs to learn the full objects and try again? Also let's just do the monolithic inference on pretrained compositional models, I don't think we need to train special monolithic models unless I'm missing something.

Thanks! Yeah, I had this hunch yesterday night and I think it'll very likely explain the poor performance. :) And sounds good about the monolithic inference on pretrained comp models! I'll ping you once I finish pretraining again and run inference. 

Few other questions:
1. The current `min_lms_match` is 4, and I was wondering if you ever played around with any less value? I think this is good to ensure all the LMS come to some conclusion, but may take a long time. Well, we'll see as I'm re-doing the pretraining with all 14 rotations. 
2. Actually, why is `min_lms_match` set to 4 and not 3? 😅 My dumb self has all this time knew there were 3 LMs (like `evidence_3lm_heterarchy`) and was like....yeah 3 equals 4. 😂
3. Can I ask if you think using `naive_scan_1` and the transforms (like GaussianBlurRGB) in your [train_heterarch.yaml](https://github.com/ramyamounir/tbp.monty/blob/5fc780b8a1cfecb9952d1af0589ca337e19b7a2c/src/tbp/monty/conf/experiment/train_heterarch.yaml) are necessary? Looking at this again, since we are training the HL-LM, the motor policy could be `naive_scan_5` and we don't need to use the transforms for 2D SM. (I realize with `tbp.teleop` it probably didn't matter which motor policy to use? but curious about the transforms).

ramyamounir:
Sounds good, thank you!

> why is `min_lms_match` set to 4 and not 3?

Yeah, nice catch. I think 3 would be equally good. The goal here was to make sure the episode didn't end prematurely and I remember thinking this needs to be 3 or more, so I just set it to 4 to be on the safe side. I don't think it matters much when doing supervised pretraining, but let's set it to 3 because that makes more sense.

> Can I ask if you think using naive_scan_1 and the transforms (like GaussianBlurRGB) in your [train_heterarch.yaml](https://github.com/ramyamounir/tbp.monty/blob/5fc780b8a1cfecb9952d1af0589ca337e19b7a2c/src/tbp/monty/conf/experiment/train_heterarch.yaml) are necessary?

Good point, I'm not sure if we need the Gaussian blur for inferring the logo when training the HL-LM compositional model. We can test without it and see. But eitherway, I think it would be better to move these transforms to a new file under `conf/env_interface/transforms` if using different ones from the `missing_depthto3d_sensor4`. This would be better than overriding in the experiment file.

Re. the agent step size, I think this is one of the open questions. One argument for a smaller step size is that we believe the 2D inference works better this way and it gives the LM enough time to infer the logo before moving out. This gives us better compositional models. But it slows down the training considerably and the graphs aren't good at the moment anyway because the spiral policy still ends up moving out of the logo area. So I think it's fine to switch to `naive_scan_5`, as it should not affect the monolithic inference too much. Once we have Scott's policy work integrated, we will not need the small step size, so this choice is temporary anyway.

hlee9212:
Closing this PR as it will be superseded with other PRs related to compositional configs.

PR #1110, opened 2026-07-22
fix: normalize eigenvectors in principal_curvatures()

hlee9212:
Hi Jeremy, I just realized that I forgot to add normalization after we updated to `scipy.linalg.eigh(buv, guv)` in #1099. 

`scipy.linalg.eigh(buv, guv)` returns guv-normalized eigenvector*, unlike the Euclidean-unit eigenvectors previously returned by `np.linalg.eig(m)`. The fix normalizes each eigenvector to Euclidean unit length. Otherwise I get `ValueError("The pc1_dir and pc2_dir must be unit vectors.")` in some compositional training experiment that uses `TwoDSensroModule`. Thank goodness we added this check back then. 😂

Note on *: By `guv` (G)-normalized, I mean satisfying this equation: $v^T Gv = 1$ instead of $v^T v = 1$, where the latter is the usual Euclidean norm.

### DIFF
--- a/src/tbp/monty/frameworks/utils/sensor_processing.py
+++ b/src/tbp/monty/frameworks/utils/sensor_processing.py
@@ -709,6 +709,8 @@ def principal_curvatures(
             # definite, so the generalized spectral theorem guarantees real
             # eigenvalues and a complete basis of real eigenvectors.
             eigval, eigvec = scipy.linalg.eigh(buv, guv)
+            eigvec[:, 0] = normalize(eigvec[:, 0])
+            eigvec[:, 1] = normalize(eigvec[:, 1])
 
             # `eigh` returns eigenvalues in ascending order. Preserve the existing
             # convention of ordering principal curvatures from largest to smallest.


jeremyshoemaker (commented):
This sounds reasonable. Is there a test we could add to catch things like this in a future refactor?

hlee9212:
I remember we added `sensor_processing_test.py` together during IP for `TwoDSensorModule`... but that was the first time we added unit tests for the file `sensor_processing.py` (and only for the new functions that were added, not existing ones like `principal_curvatures()` function). I was just thinking perhaps adding more unit tests (especially around sensor_processing.py and spatial_arithmetics.py) might be a good project during the retreat. 😂 I definitely wanted to do this since our IP work.

A test for `principal_curvatures()` would check whether the output `pc1_dir` and `pc2_dir` are (1) unit vectors, and (2) orthogonal. (Or combined together, whether it is orthonormal - we may have an `is_orthonormal()` somewhere?). The only complication is that actually our `pc1_dir` and `pc2_dir` are not **always** orthogonal, as per the TODO comment: `# TODO: sometimes dir1 and dir2 are not orthogonal, why?`

I have figured out why this is the case but do need to set aside time to test just because all experiments that I know of use `principal_curvatures()`. Alas, that is not this PR, though.

jeremyshoemaker (approved):

PR #1111, opened 2026-07-23
test: `GraphLearningTest::test_time_out` passes in both modes

dschumacher-tbp:
`tests/integration/frameworks/models/graph_learning_test.py::GraphLearningTest::test_time_out` was modifying Learning Modules after intialization. This works using the "reset" strategy, but not using "recreation". In "recreation" mode, the configuration structure is patched instead. Then the `Experiment.pre_episode()` method can re-create the Learning Modules with their modified configurations.

### DIFF
--- a/tests/integration/frameworks/models/graph_learning_test.py
+++ b/tests/integration/frameworks/models/graph_learning_test.py
@@ -325,7 +325,7 @@ class GraphLearningTest(BaseGraphTest):
                     graph_2 = graph_memory_2[obj_name][input_channel]
                     self.check_graphs_equal(graph_1, graph_2)
 
-    def test_time_out(self):
+    def test_time_out(self) -> None:
         """Test time_out and pose_time_out detection and logging.
 
         # TODO: This test is a little shaky and should be improved.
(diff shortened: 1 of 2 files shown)


tests/integration/frameworks/models/graph_learning_test.py
```diff
@@ -342,15 +342,26 @@ def test_time_out(self):
             for e in range(6):
                 if e % 2 == 0:
                     exp.pre_epoch()
-                if e == 2:
+                if e >= 2:
```
jeremyshoemaker: I think we should rename this variable. I had to look around to figure out the context, and it would just be quicker if it was something explicit like `episode`.

A more important question though, as I stare at this some more, are we okay with this `if` block running in the >= 4 cases? The previous version wasn't, and this one is.

:thinking: I guess it's because the original version was relying on the LMs not being replaced, so the previously set `max_match_distance` would still be there.

tests/integration/frameworks/models/graph_learning_test.py
```diff
@@ -342,15 +342,26 @@ def test_time_out(self):
...
                     # Set max steps low & raise mmd to get pose time outs
                     exp.max_train_steps = 3
-                    exp.model.learning_modules[0].max_match_distance = 0.1
-                if e == 4:
+                    if exp._recreation_mode:
+                        lm_cfg = exp._monty_cfg["learning_modules"]["learning_module_0"]
+                        lm_cfg["max_match_distance"] = 0.1
```
jeremyshoemaker, line 350: I appreciate that you didn't make this a one-liner. We've done that far too often elsewhere, and it makes it harder to read.

tests/integration/frameworks/models/graph_learning_test.py
```diff
@@ -371,11 +382,6 @@ def test_time_out(self):
             "pose time out episode should have more than one possible pose.",
         )
         for episode in [4, 5]:
-            self.assertEqual(
-                train_stats["primary_performance"][episode],
-                "time_out",
-                "time out not recognized/logged correctly",
-            )
```
jeremyshoemaker, line 378: :laughing: We had a duplicate assert?! I'm sure that's not the only time this has happened. Thanks for cleaning it up.

jeremyshoemaker (commented):
Looks good to me, but could you verify my understanding of one of the changes is correct? I think it is, but I want to be sure.

jeremyshoemaker (approved):
I'm gonna go ahead and approve this to unblock you, since I'm pretty sure my understanding is correct.

@tristanls-tbp can merge it for you when it's ready.

jeremyshoemaker:
I'd do it right now, but he's merging the release PR and I don't want to mess that up.

tests/integration/frameworks/models/graph_learning_test.py
dschumacher-tbp: The `>=` is **required** in the "recreate" case, and only _harmless_ in the "reset" case.

tristanls-tbp:
I enabled auto-merge since this has @jeremyshoemaker's approval.

PR #1113, opened 2026-07-23
chore: create MuJoCo versions of ycb_10objs benchmarks

jeremyshoemaker:
Created experiment configurations for the benchmarks listed in `benchmarks/ycb_10objs.csv`. We decided to exclude the `base_10multi_distinctobj_dist_agent` for now, as it uses distractor objects that we haven't implemented in MuJoCo and we've decided we don't want to implement them at this time.

I've run all of these locally, and the run to completion with reasonable overall stats. The numbers look good even with these still running on Habitat trained models.

### DIFF
--- a/src/tbp/monty/conf/experiment/base_10simobj_surf_agent_mujoco.yaml
+++ b/src/tbp/monty/conf/experiment/base_10simobj_surf_agent_mujoco.yaml
@@ -1,0 +1,29 @@
+# @package _global_
+
+defaults:
+  - /monty: evidencegraph_exp1000_emin_t3_tot2500
+  - /monty/motor_system_config: surface_curvature_informed_goal1
+  - /monty/learning_module: evidence_1lm_nn5_dod0025
+  - /monty/sensor_module: camera_surf_delta
+  - /monty/connectivity: 1lm_1sm
+  - /environment: mujoco_ycb_surf_agent
+  - /env_interface: eval_simobj_predefined
+  - /env_interface/transform: missing_depthto3d_sensor2_semantic0_clip
+  - /logging: basic_warning_wandb_monty_runs
+
+experiment:
+  _target_: tbp.monty.frameworks.experiments.object_recognition_experiments.MontyObjectRecognitionExperiment
+  config:
+    show_sensor_output: false
+    max_train_steps: 1000
+    max_eval_steps: 500
+    max_total_steps: 5000
+    n_train_epochs: 3
+    model_name_or_path: ${constants.pretrained_dir}/surf_agent_1lm_10similarobj/pretrained/
+    n_eval_epochs: ${constants.rotations_all_count}
+    min_lms_match: 1
+    seed: 42
+    supervised_lm_ids: []
+    python_log_level: DEBUG
+    logging:
+      run_name: base_10simobj_surf_agent_mujoco
--- a/src/tbp/monty/conf/experiment/randrot_10distinctobj_surf_agent_mujoco.yaml
+++ b/src/tbp/monty/conf/experiment/randrot_10distinctobj_surf_agent_mujoco.yaml
@@ -2,14 +2,13 @@
 
 defaults:
   - /monty: evidencegraph_exp1000_emin_t3_tot2500
-  - /monty/motor_system_config: distant_5
-  - /monty/learning_module: evidence_1lm_nn5_dod003
-  - /monty/sensor_module: camera_dist_delta_salience
+  - /monty/motor_system_config: surface_curvature_informed_goal1
+  - /monty/learning_module: evidence_1lm_nn5_dod0025
+  - /monty/sensor_module: camera_surf_delta
   - /monty/connectivity: 1lm_1sm
-  - /environment: mujoco_ycb_dist_agent
-  - /env_interface: eval_distinctobj_predefined
-  - /env_interface/positioning_procedures_eval: getgoodview_viewfinder_patch
-  - /env_interface/transform: missing_depthto3d_sensor2_semantic0
+  - /environment: mujoco_ycb_surf_agent
+  - /env_interface: eval_distinctobj_random
+  - /env_interface/transform: missing_depthto3d_sensor2_semantic0_clip
   - /logging: basic_warning_wandb_monty_runs
 
 experiment:
@@ -18,13 +17,13 @@ experiment:
     show_sensor_output: false
     max_train_steps: 1000
     max_eval_steps: 500
-    max_total_steps: 6000
+    max_total_steps: 5000
     n_train_epochs: 3
     model_name_or_path: ${constants.pretrained_dir}/surf_agent_1lm_10distinctobj/pretrained/
-    n_eval_epochs: ${constants.rotations_all_count}
+    n_eval_epochs: 10
     min_lms_match: 1
     seed: 42
     supervised_lm_ids: []
     python_log_level: DEBUG
     logging:
-      run_name: base_config_10distinctobj_dist_agent_mujoco
+      run_name: randrot_10distinctobj_surf_agent_mujoco
--- a/src/tbp/monty/conf/experiment/randrot_noise_10distinctobj_dist_agent_mujoco.yaml
+++ b/src/tbp/monty/conf/experiment/randrot_noise_10distinctobj_dist_agent_mujoco.yaml
@@ -3,11 +3,11 @@
 defaults:
   - /monty: evidencegraph_exp1000_emin_t3_tot2500
   - /monty/motor_system_config: distant_5
-  - /monty/learning_module: evidence_1lm_nn5_dod003
-  - /monty/sensor_module: camera_dist_delta_salience
+  - /monty/learning_module: evidence_1lm_nn10_dod003
+  - /monty/sensor_module: camera_dist_noise_raw1_salience
   - /monty/connectivity: 1lm_1sm
   - /environment: mujoco_ycb_dist_agent
-  - /env_interface: eval_distinctobj_predefined
+  - /env_interface: eval_distinctobj_random
   - /env_interface/positioning_procedures_eval: getgoodview_viewfinder_patch
   - /env_interface/transform: missing_depthto3d_sensor2_semantic0
   - /logging: basic_warning_wandb_monty_runs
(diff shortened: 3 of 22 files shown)


src/tbp/monty/conf/experiment/base_10simobj_surf_agent_mujoco.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+experiment:
+  _target_: tbp.monty.frameworks.experiments.object_recognition_experiments.MontyObjectRecognitionExperiment
+  config:
+    show_sensor_output: false
+    max_train_steps: 1000
+    max_eval_steps: 500
+    max_total_steps: 5000
+    n_train_epochs: 3
+    model_name_or_path: ${constants.pretrained_dir}/surf_agent_1lm_10similarobj/pretrained/
```
jeremyshoemaker, line 22: I want to point out that I haven't changed these for any of these configurations for two reasons.

1. This should be where they are when we're done, by putting the new models there when we train them.
2. I can change which models I'm running on by changing the environment variable used in `constants.pretrained_dir` to switch if I need to.

Hard-coding it by putting `mujoco` in the path would make that harder to do.

src/tbp/monty/conf/experiment/base_config_10distinctobj_dist_agent_mujoco.yaml
```diff
@@ -10,7 +10,7 @@ defaults:
   - /env_interface: eval_distinctobj_predefined
   - /env_interface/positioning_procedures_eval: getgoodview_viewfinder_patch
   - /env_interface/transform: missing_depthto3d_sensor2_semantic0
-  - /logging: basic_info_monty_runs
+  - /logging: basic_warning_wandb_monty_runs
```
jeremyshoemaker, line 13: This is me correcting a mistake from when I originally added this months ago. I think I changed it for testing and forgot to change it back.

src/tbp/monty/conf/experiment/randrot_10distinctobj_surf_agent_mujoco.yaml
```diff
@@ -0,0 +1,29 @@
+# @package _global_
+
+defaults:
+  - /monty: evidencegraph_exp1000_emin_t3_tot2500
+  - /monty/motor_system_config: surface_curvature_informed_goal1
+  - /monty/learning_module: evidence_1lm_nn5_dod0025
+  - /monty/sensor_module: camera_surf_delta
```
tristanls-tbp, line 7: **question**: The Habitat experiment has this as `camera_surf_noise_raw1`. What is the reason for the change here?

tristanls-tbp (commented):
Thank you for putting these together.

**suggestion (non-blocking)**: You could put all of these in `experiment/mujoco` instead. They'd be accessible from the command line via `experiment=mujoco/base_10...`. This way, the namespace isn't included in every file.

jeremyshoemaker:
I meant to include this to help with review. It's a diff of the MuJoCo versions against the Habitat versions.

<details>
<summary>Config diffs</summary>

```diff
=== diff: base_config_10distinctobj_surf_agent.yaml <-> base_config_10distinctobj_surf_agent_mujoco.yaml ===
-   - /environment: habitat_ycb_surf_agent
+   - /environment: mujoco_ycb_surf_agent
-       run_name: base_config_10distinctobj_surf_agent
+       run_name: base_config_10distinctobj_surf_agent_mujoco

=== diff: randrot_noise_10simobj_surf_agent.yaml <-> randrot_noise_10simobj_surf_agent_mujoco.yaml ===
-   - /environment: habitat_ycb_surf_agent
+   - /environment: mujoco_ycb_surf_agent
-       run_name: randrot_noise_10simobj_surf_agent
+       run_name: randrot_noise_10simobj_surf_agent_mujoco

=== diff: randrot_noise_10distinctobj_dist_on_distm.yaml <-> randrot_noise_10distinctobj_dist_on_distm_mujoco.yaml ===
-   - /environment: habitat_ycb_dist_agent_semantics0
+   - /environment: mujoco_ycb_dist_agent
-       run_name: randrot_noise_10distinctobj_dist_on_distm
+       run_name: randrot_noise_10distinctobj_dist_on_distm_mujoco

=== diff: base_config_10distinctobj_dist_agent.yaml <-> base_config_10distinctobj_dist_agent_mujoco.yaml ===
-   - /environment: habitat_ycb_dist_agent_semantics0
+   - /environment: mujoco_ycb_dist_agent
-       run_name: base_config_10distinctobj_dist_agent
+       run_name: base_config_10distinctobj_dist_agent_mujoco

=== diff: randrot_noise_10simobj_dist_agent.yaml <-> randrot_noise_10simobj_dist_agent_mujoco.yaml ===
-   - /environment: habitat_ycb_dist_agent_semantics0
+   - /environment: mujoco_ycb_dist_agent
-       run_name: randrot_noise_10simobj_dist_agent
+       run_name: randrot_noise_10simobj_dist_agent_mujoco

=== diff: base_10simobj_surf_agent.yaml <-> base_10simobj_surf_agent_mujoco.yaml ===
-   - /environment: habitat_ycb_surf_agent
+   - /environment: mujoco_ycb_surf_agent
-       run_name: base_10simobj_surf_agent
+       run_name: base_10simobj_surf_agent_mujoco

=== diff: randrot_10distinctobj_surf_agent.yaml <-> randrot_10distinctobj_surf_agent_mujoco.yaml ===
-   - /environment: habitat_ycb_surf_agent
+   - /environment: mujoco_ycb_surf_agent
-       run_name: randrot_10distinctobj_surf_agent
+       run_name: randrot_10distinctobj_surf_agent_mujoco

=== diff: randrot_noise_10distinctobj_5lms_dist_agent.yaml <-> randrot_noise_10distinctobj_5lms_dist_agent_mujoco.yaml ===
-   - /environment: habitat_dist_agent_sensors5
+   - /environment: mujoco_dist_agent_sensors5

=== diff: randrot_noise_10distinctobj_surf_agent.yaml <-> randrot_noise_10distinctobj_surf_agent_mujoco.yaml ===
-   - /environment: habitat_ycb_surf_agent
+   - /environment: mujoco_ycb_surf_agent
-       run_name: randrot_noise_10distinctobj_surf_agent
+       run_name: randrot_noise_10distinctobj_surf_agent_mujoco

=== diff: randrot_noise_10distinctobj_vocus_on_vocusm.yaml <-> randrot_noise_10distinctobj_vocus_on_vocusm_mujoco.yaml ===
-   - /environment: habitat_ycb_dist_agent_semantics0_256
+   - /environment: mujoco_ycb_dist_agent_256
-       run_name: randrot_noise_10distinctobj_vocus_on_vocusm
+       run_name: randrot_noise_10distinctobj_vocus_on_vocusm_mujoco

=== diff: randrot_noise_10distinctobj_dist_agent.yaml <-> randrot_noise_10distinctobj_dist_agent_mujoco.yaml ===
-   - /environment: habitat_ycb_dist_agent_semantics0
+   - /environment: mujoco_ycb_dist_agent
-       run_name: randrot_noise_10distinctobj_dist_agent
+       run_name: randrot_noise_10distinctobj_dist_agent_mujoco

=== diff: randomrot_rawnoise_10distinctobj_surf_agent.yaml <-> randomrot_rawnoise_10distinctobj_surf_agent_mujoco.yaml ===
-   - /environment: habitat_surf_rawnoise_agent
+   - /environment: mujoco_surf_rawnoise_agent
-       run_name: randomrot_rawnoise_10distinctobj_surf_agent
+       run_name: randomrot_rawnoise_10distinctobj_surf_agent_mujoco
```

</details>

src/tbp/monty/conf/experiment/randrot_10distinctobj_surf_agent_mujoco.yaml
jeremyshoemaker: I'll take a look. It didn't show up in my diffs, so I didn't see it. I definitely didn't deliberately change it.

jeremyshoemaker:
> **suggestion (non-blocking)**: You could put all of these in `experiment/mujoco` instead. They'd be accessible from the command line via `experiment=mujoco/base_10...`. This way, the namespace isn't included in every file.

That's probably a good idea. This was following the pattern I started with the `base_config_10distinctobj_dist_agent` and surface agent ones I added back when I added  `DistantAgent` and `SurfaceAgent`. That was before I decided to do the `tests/mujoco` thing. 

The difference there was I needed to keep track of where the original file was relative to the new one, and making a matching tree made sense to me then. These are all at one level, adjacent to their originals. It also makes it easier to eyeball which ones have a matching partner and which ones don't.

But if I change it, I'll need to update the `conf_test`, because it's looking for `mujoco` at the end of the file name stem.

src/tbp/monty/conf/experiment/randrot_10distinctobj_surf_agent_mujoco.yaml
jeremyshoemaker: I'm not sure which one you're looking at, but this looks to match the original to me. https://github.com/thousandbrainsproject/tbp.monty/blob/29df8e4dbfe8bf44802a7f1e2951e93e1b4052b2/src/tbp/monty/conf/experiment/randrot_10distinctobj_surf_agent.yaml#L7

src/tbp/monty/conf/experiment/randrot_noise_10distinctobj_dist_agent_mujoco.yaml
```diff
@@ -0,0 +1,30 @@
+# @package _global_
+
+defaults:
+  - /monty: evidencegraph_exp1000_emin_t3_tot2500
+  - /monty/motor_system_config: distant_5
+  - /monty/learning_module: evidence_1lm_nn10_dod003
+  - /monty/sensor_module: camera_dist_noise_raw1_salience
+  - /monty/connectivity: 1lm_1sm
+  - /environment: mujoco_ycb_dist_agent
```
jeremyshoemaker, line 9: I meant to point this out earlier too. In the Habitat version, this is called `habitat_ycb_dist_agent_semantic0`. This seems to be one of only two places we use `semantic0` in the filename for an environment config. 

Compare this to `habitat_dist_agent` and `habitat_ycb_surf_agent`, both of which set `semantics` to `False` but don't have `semantic0` in their filename. I consider this to be the default, so putting it in the filename seems redundant, when we don't do it for other default values.

src/tbp/monty/conf/experiment/randrot_10distinctobj_surf_agent_mujoco.yaml
tristanls-tbp: I'm not sure what I was looking at either. Looking now I see what you see.

tristanls-tbp (approved):
Thank you for clarifying. Like I mentioned, it looks like I was seeing the wrong thing. Changes look good to me.

src/tbp/monty/conf/experiment/randrot_10distinctobj_surf_agent_mujoco.yaml
jeremyshoemaker: Some of the names are very close to each other. I've definitely made similar mistakes while working on this, which is why I made sure to diff against the originals using a script.

PR #1115, opened 2026-07-23
test: all `GraphLearningTest` cases pass in both modes

dschumacher-tbp:
`GraphLearningTest::test_detailed_logging()` failed when `_recreation_mode = True` because it relied on accumulating `possible_matches` across episodes. I adjusted the test to compare against `detailed_stats['1']['LM_0']['individual_ts_reached_at_step']` instead, which works in both modes.

With this change, all the tests in `tests/integration/frameworks/models/graph_learning_test.py` pass in both modes.

### DIFF
--- a/tests/integration/frameworks/models/graph_learning_test.py
+++ b/tests/integration/frameworks/models/graph_learning_test.py
@@ -509,7 +509,7 @@ class GraphLearningTest(BaseGraphTest):
                 "on object",
             )
 
-    def test_detailed_logging(self):
+    def test_detailed_logging(self) -> None:
         exp = instantiate_experiment(self.feature_pred_off_object_cfg.experiment)
         with exp:
             exp.run()
@@ -547,7 +547,7 @@ class GraphLearningTest(BaseGraphTest):
             "even those off the object.",
         )
         self.assertEqual(
-            len(detailed_stats["1"]["LM_0"]["possible_matches"]),
+            detailed_stats["1"]["LM_0"]["individual_ts_reached_at_step"],
             train_stats.loc[1]["monty_matching_steps"],
             "matching steps in detailed stats don't match with those in train stats.",
         )
--- a/tests/mujoco/integration/frameworks/models/graph_learning_test.py
+++ b/tests/mujoco/integration/frameworks/models/graph_learning_test.py
@@ -508,7 +508,6 @@ class GraphLearningTest(BaseGraphTest):
         with exp:
             exp.run()
 
-        # TODO: improve load_stats with a structured return type
         train_stats, eval_stats, detailed_stats, lm_models = load_stats(
             exp.output_dir,
             load_train=True,
@@ -542,7 +541,7 @@ class GraphLearningTest(BaseGraphTest):
             "even those off the object.",
         )
         self.assertEqual(
-            len(detailed_stats["1"]["LM_0"]["possible_matches"]),
+            detailed_stats["1"]["LM_0"]["individual_ts_reached_at_step"],
             train_stats.loc[1]["monty_matching_steps"],
             "matching steps in detailed stats don't match with those in train stats.",
         )
--- a/tests/mujoco/integration/frameworks/models/graph_learning_test.py
+++ b/tests/mujoco/integration/frameworks/models/graph_learning_test.py
@@ -508,6 +508,7 @@ class GraphLearningTest(BaseGraphTest):
         with exp:
             exp.run()
 
+        # TODO: improve load_stats with a structured return type
         train_stats, eval_stats, detailed_stats, lm_models = load_stats(
             exp.output_dir,
             load_train=True,


tristanls-tbp (commented):
I'm glad you got this working.

Could you please elaborate on why these two lines are equivalent in the context of this test? I don't understand why this is a valid change.

I started looking into it, and there are many paths to explore, so it is not straightforward for me to reconstruct.

First thing I don't understand is why the length of possible matches in episode 1 should equal `train_stats`' `monty_matching_steps`. Perhaps once you explain that, the rest will fall into place?

dschumacher-tbp:
This test seems to rely on accumulation of telemetry in the model, and I know we want to extract the telemetry from the model, where it can't be corrupted by the re-creation process.

When in "reset" mode (where the test passes), the `detailed_stats['1']['LM_0']['possible_matches']` looks like:
```
[['new_object0'],['new_object0'],['new_object0'],['new_object0'],['new_object0'],['new_object0'],['new_object0'],['new_object0'],['new_object0'],['new_object0'],['new_object0'],['new_object0'],['new_object0']]
```
but in the "recreate" mode, the stats look like:
```
[['new_object0']]
```
which appears to be because the possible matches are not being accumulated for each step.

The test seems to be trying to confirm the number of steps taken, expecting them to be `train_stats.loc[1]["monty_matching_steps"]`, which is 13.

The current code seems to infer that from the `len()` of `possible_matches`. However, the `individual_ts_reached_at_step` seems to record that information directly, as in `FeatureAtLocationBuffer`:
```python
    def set_individual_ts(self, object_id, pose):
        """Update self.stats with the individual LMs terminal state."""
        # Only log first time terminal condition is met
        if self.stats["individual_ts_reached_at_step"] is None:
            self.stats["individual_ts_reached_at_step"] = self.get_num_matching_steps()
            self.stats["individual_ts_object"] = object_id
            self.stats["individual_ts_pose"] = pose
            self.stats["individual_ts_rot"] = self.stats["detected_rotation_quat"]
            # If no symmetry was detected, this will be None.
            self.stats["symmetric_rotations_ts"] = self.stats["symmetric_rotations"]
```

tristanls-tbp:
Hi Dale, thanks for the additional detail, this is as far as I've gotten myself as well.

**question**: What I'm looking for is an explanation of why we expected the length of the detailed stats to match the count of matching steps before. The assertion text states that it is a problem when the matching steps in the detailed stats don't match those in the train stats: "matching steps in detailed stats don't match with those in train stats." I don't understand why the `possible_matches` aren't being accumulated now. Once I understand the previous and current meaning, this will be good to go, but right now I can't tell if the test was asserting something important.

Here's some of the tracing I did for myself, in case it's useful.

A `GraphLM.matching_step(...)` calls `self._compute_possible_matches(...)`, which in turn calls `self._update_possible_matches(...)`. An `EvidenceGraphLM` implements that by setting `self.possible_matches = self._threshold_possible_matches()`. This is significant because `self.possible_matches` is what is returned by `self.get_possible_matches()`, which is what is recorded by `self.collect_stats_to_save()` as `stats["possible_matches"]` later on in `GraphLM.matching_step(...)` (where we started) via `stats = self.collect_stats_to_save()`. So, a lot of round trips, but there's a relationship here of `GraphLM.matching_step(...)` creating possible matches and saving them via `self.buffer.update_stats(stats, ...)`

```python
    def matching_step(
        self,
        ctx: RuntimeContext,
        percepts: Sequence[Message],
    ) -> None:
        """Update the possible matches given an observation."""
        first_movement_detected = self._agent_moved_since_reset()
        buffer_data = self._add_displacements(percepts)
        self.buffer.append(buffer_data)
        self.buffer.append_input_percepts(percepts)

        if first_movement_detected:
            logger.debug("performing matching step.")
        else:
            logger.debug("we have not moved yet.")

        # *** possible matches computed here ***
        self._compute_possible_matches(
            ctx, percepts, first_movement_detected=first_movement_detected
        )

        if len(self.get_possible_matches()) == 0:
            self.set_individual_ts(terminal_state="no_match")

        if self.gsg is not None:
            self.gsg.step(ctx, percepts)

        # *** possible matches are read here ***
        stats = self.collect_stats_to_save()
        # *** possible matches are stored here ***
        self.buffer.update_stats(stats, append=self.has_detailed_logger)
```

Where `self.buffer.update_stats(...)` is implemented as
```python
    def update_stats(self, stats, update_time=True, append=True, init_list=True):
        """Update statistics for this step in the episode."""
        for stat in stats:
            if stat in self.stats and append:
                self.stats[stat].append(copy.deepcopy(stats[stat]))
            elif init_list:
                self.stats[stat] = [copy.deepcopy(stats[stat])]
            else:
                self.stats[stat] = copy.deepcopy(stats[stat])
        if update_time:
            self.stats["time"].append(time.time() - self.start_time)
```

Hopefully, this example trace helps. However, having traced it, I'm still not clear why a feature-flag change would affect it. Could it be that `stats` need to be recreated as well?

dschumacher-tbp:
I've been tracing failures in the Benchmark Experiments, and I'm beginning to think the root-cause of this issue may be related. I'm planning to pursue this idea further to see if this theory provides better answers.

dschumacher-tbp:
This work was superseded by [refactor: don't recreate Monty the first time](https://github.com/thousandbrainsproject/tbp.monty/commit/c87f72b2185b3a4ddc8bdfe3f26ff0cf3ab0f938) ([#1117](https://github.com/thousandbrainsproject/tbp.monty/pull/1117))

PR #1118, opened 2026-07-30
fix: Update directory and add symlink command for compositional dataset

nielsleadholm:
If you follow the instructions as currently written, the compositional object directory is not identified during an experiment run.

Updates the description so that dataset is unpacked in the versioned directory, and provides instructions for a symlink. 

@scottcanoe let me know if this fits with the way you planned out the updated directory structures, or if we should structure the symlink differently. 

[Edit: for context, here is the relevant earlier PR

https://github.com/thousandbrainsproject/tbp.monty/pull/874

]


scottcanoe (approved):
Thanks @nielsleadholm, apologies for the mistake. This is how I have it laid out on my machine and is consistent with the configs.

nielsleadholm:
Fyi @hlee9212 and @ramyamounir , depending on how you've downloaded the datasets and unpacked the directories, you may want to modify your setup to reflect the updated guide.

PR #1119, opened 2026-07-30
docs: update explanation of some of the results

scottcanoe:
This PR updates a discussion section regarding the comparison between surface agent and distant agent benchmark results.


docs/overview/benchmark-experiments.md
```diff
@@ -73,11 +73,11 @@ The following results are obtained from experiments on the entire YCB dataset (7
...
 
-- **Why does the distant agent do worse than the surface agent?**
-  The distant agent has limited capabilities to move along the object. In particular, the distant agent currently uses an almost random policy which is not as efficient and informative as the surface agent which follows the principal curvatures of the object. Note however that both the distant and surface agent can now move around the object using the hypothesis-testing action policy, and so the difference in performance between the two is not as great as it previously was.
-
 - **Why is the distant agent on the distant agent models worse than on the surface agent model?**
-  As you can see in the figure above, the models learned with distant agent have several blind spots and unevenly sampled areas. When we test random rotations we may see the object from views that are underrepresented in the object model. If we use a 10% threshold instead of 20% we can actually get a little better performance with the distant agent since we allow it to converge faster. This may be because it gets less time to move into badly represented areas and because it reaches the time-out condition less often.
```
scottcanoe, line 80: The explanation I've written sums up my understanding of the surface/distant-agent disparities with a focus on learned models. I think the deleted threshold text refers to inference parameters, but I'm not sure which. Is the threshold text still relevant? If I should add it back in, can you clarify when thresholds we're talking about?

nielsleadholm: Is there actually an appreciable difference between `randrot_noise_10distinctobj_dist_agent` and `randrot_noise_10distinctobj_dist_on_distm`? Couldn't we just get rid of most of these paragraphs?

Also I think "Why do the distant agent experiments take longer and have more episodes where the most likely hypothesis is used?" can be removed. 

For what it's worth, we already discuss differences in the density of the models under "For experiments with multiple sensors and learning modules we currently only have a setup with the distant agent so we also have to train the models with the distant agent..."

scottcanoe: You're right, there's no appreciable difference. The differences are a bit more pronounced on the 77-object dataset, but everything is performing about the same on the 10-distinctobj dataset that these benchmarks use. I was loading in old context probably.

Seems like maybe we just remove this bullet point entirely @nielsleadholm ?

nielsleadholm: Yeah that's true, I think it might be worth still highlighting the difference in the 77-object ones. This difference is probably because, as good as the saccade policy is, it still can't move around to the other side of objects without an LM goal being invoked. It's also potentially related to the modality transfer (surface rather than distant based models, when doing inference with distant).

scottcanoe: > ...I think it might be worth still highlighting the difference in the 77-object ones.

Sorry, probably should have been more clear on this, but we don't have dist on dist (or vocus on vocus) for 77-object benchmark experiments. My comments regarding 77-object experiments just come from experience/RP work, so I'm inclined to leave that discussion out of the benchmark docs so we don't confuse anyone. It seems like a potentially useful discussion, but I'm not sure where it'd go.

nielsleadholm: Ah I see, I was referring to the greater matching steps on the 77 object condition in the benchmarks for distant-on-surface, but yeah we can also just remove all of this discussion, the difference isn't that marked.

PR #1128, opened 2026-08-03
chore: remove base_10multi_distinctobj_dist_agent benchmark

tristanls-tbp:
Per team discussion, we are removing the `base_10multi_distinctobj_dist_agent` benchmark that uses distractor objects. If/when we reintroduce this benchmark in MuJoCo, we will reimplement the distractor setup mechanism.


jeremyshoemaker (approved):
Looks good to me.

nielsleadholm (approved):
LGTM. Just noting that with this removed, there is a fair amount of code around things like `source_object_list`, `add_distractor_objects`, and `primary_target` vs `stepwise_target` that could be removed.

tristanls-tbp:
Sounds good @nielsleadholm, I created #1130 to clean up `add_distractor_objects`, but it is not clear to me how to remove the rest. To me, they seem to be used in other contexts.

Screenshot 2026-08-04 at 10 33 09

Screenshot 2026-08-04 at 10 34 03

nielsleadholm:
So looking at this PR (https://github.com/thousandbrainsproject/nupic.monty/pull/298) might be helpful to identify some other code that could be removable, for example:

- `_bounding_corners`
- `non_conflicting_vector`
- `check_viewpoint_collision`
- `find_non_colliding_positions`

Re. the others I mentioned
- Before, there was no `stepwise_target`, and `primary_target` existed (and should still exist), but it was just called `target`
- Re. `create_semantic_mapping` and `source_object_list` - before this functionality, the semantic ID was always returned as 1; it just told us whether we were on the object or not, but it couldn't distinguish between different objects (which is fine when there's only ever one object).

nielsleadholm:
It's quite a lot of code, let me know if you'd want to do it as a pair programming exercise at some point.

tristanls-tbp:
@nielsleadholm I'll add it to the scope of #1130 if that's alright, it will expand quite a bit.

PR #1127, opened 2026-08-03
refactor!: min_lms_match is now MatchCriterion

tristanls-tbp:
#1122 surfaced a need for more elaborate experiment matching criteria. This pull request replaces the no longer fit-for-purpose `min_lms_match` with a `MatchCriterion` protocol that allows us to configure experiments with different kinds of match criteria.

This is a minimal pull request, so it does not repair how the experiment currently wires up the match criterion. That is, we still reach into Monty and set a property on it after initialization. The only change to the wiring mechanism was making the property "private" to highlight the inappropriate access.

Screenshot 2026-08-03 at 15 03 22

## Review Notes

The main/relevant changes are in one new file and two modified ones:
```
new file:   src/tbp/monty/experiment/match_criteria.py
modified:   src/tbp/monty/frameworks/experiments/monty_experiment.py
modified:   src/tbp/monty/frameworks/models/graph_matching.py
```

The file count is large because every configuration changed. The config changes are of the shape:

```diff
- min_lms_match: 1
+ match_criterion:
+   _target_: tbp.monty.experiment.match_criteria.AnyLMsMatch
+   count: 1
```

## Breaking Changes

`experiment.min_lms_match` configuration is replaced by `experiment.match_criterion`.

## Benchmarks

<details>
<summary>No changes</summary>

| Experiment | Metric | Baseline | Proposed | Change |
| --- | --- | --- | --- | --- |
| base_config_10distinctobj_dist_agent | percent_correct (%) | 100 | 100 | 0 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 28 | 28 | 0 |
|  | rotation_error (deg) | 10.29 | 10.29 | 0 |
|  | runtime (min) | 2 | 3 | 1 |
|  | episode_runtime (sec) | 5 | 5 | 0 |
|  | num_episodes | 140 | 140 | 0 |
| base_config_10distinctobj_surf_agent | percent_correct (%) | 100 | 100 | 0 |
|  | used_mlh (%) | 0 | 0 | 0 |
|  | match_steps | 28 | 28 | 0 |
|  | rotation_error (deg) | 3.64 | 3.64 | 0 |
|  | runtime (min) | 4 | 4 | 0 |
|  | episode_runtime (sec) | 11 | 12 | 1 |
|  | num_episodes | 140 | 140 | 0 |
| randrot_noise_10distinctobj_dist_agent | percent_correct (%) | 99 | 99 | 0 |
[...]

PR #1126, opened 2026-08-03
fix!: change eigen* calculations to fix benchmarks

jeremyshoemaker:
In PR #1099 we updated a few places we use NumPy to make it compatible with newer versions of NumPy where `np.linalg.eig` always returns complex values even when the values are real.

The method we took of changing to using `eigh` ended up changing benchmark results unexpectedly. Even after fixing an issue with PR #1110, we're still getting unexpected results.

This changes things back to the original approach I was taking in #1099 of calculating the eigenvectors and eigenvalues using `np.linalg.eig` and taking the real values off the complex results. In both places, we know that the values have to be real, so this should be a safe assumption.

> [!warning]
> **Breaking Changes**
> The `principal_curvatures` and `surface_normal_total_least_squares` functions were erroneously returning different results in certain situations. This fix corrects them both.

PR #1134, opened 2026-08-06
chore: version 0.45.0

tristanls-tbp:
Minor release version due to 0 Major version and breaking changes since 0.44.1.

Screenshot 2026-08-06 at 09 09 50

## fix!: change eigen* calculations to fix benchmarks (#1126)

The `principal_curvatures` and `surface_normal_total_least_squares` functions were erroneously returning different results in certain situations. This fix corrects them both.

## refactor!: min_lms_match is now MatchCriterion (#1127)

`experiment.min_lms_match` configuration is replaced by `experiment.match_criterion`.

### DIFF
--- a/src/tbp/monty/__init__.py
+++ b/src/tbp/monty/__init__.py
@@ -8,4 +8,4 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
-__version__ = "0.44.1"
+__version__ = "0.45.0"


jeremyshoemaker (approved):
I was on the fence about whether that `fix` should be a breaking change. The original change wasn't marked that way because we didn't know it was changing things, but technically this is, so I ultimately went with marking it as such. It really only affects anyone using the versions in between.

🤔 Maybe we should list the affected releases in the release notes? It's any releases between #1099 and this one.

tristanls-tbp:
@jeremyshoemaker, what I've done in the past is yank or mark the releases themselves to highlight the bug. In this case, I'd just update the release notes for those releases with:

## Known Bugs

### <title>

\<Description\>

---

I can do that. Looks like it's gonna be 0.44.0 and 0.44.1. Do you have a title and description I could use?

jeremyshoemaker:
The problem is I don't know the math details of _why_ it was returning different results, aside from us switching from `eig` to `eigh`. We expected it to be the same, and it wasn't, but I don't know why. That's why the breaking changes notice on that PR is as vague as it is.

@hlee9212 might be able to give you a succinct summary, but maybe we don't need to go into those details.

tristanls-tbp:
I went with 

## Known Bugs

### unintentionally introduced minor deviations in some numerical calculations

In PR #1099, we updated a few places that use NumPy to make them compatible with newer versions of NumPy, where `np.linalg.eig` always returns complex values even when the input is real. This resulted in minor deviations in numerical calculations, causing benchmark results to no longer exactly match previous results.

Fixed as of [v0.45.0](https://github.com/thousandbrainsproject/tbp.monty/releases/tag/v0.45.0)

PR #1135, opened 2026-08-07
docs: Fix symlink version number + add instructions for updating

nielsleadholm:
Fixes the version number for the symlink instructions for the compositional dataset (v1.2), and adds instructions for updating the link when it already exists.

### DIFF
--- a/docs/overview/benchmark-experiments.md
+++ b/docs/overview/benchmark-experiments.md
@@ -174,10 +174,18 @@ Note: To obtain these results, pretraining was run without parallelization acros
 > unzip compositional_objects_1.2.zip
 > ```
 >
-> Then create a symlink so that the experiment configs (which look for `~/tbp/data/habitat/objects/compositional_objects`) can find the versioned dataset folder:
+> Then create a symlink so that the experiment configs (which look for `~/tbp/data/habitat/objects/compositional_objects`) can find the versioned dataset folder. If this is the first time you are doing this, you can use:
 >
 > ```plaintext
 > mkdir -p ~/tbp/data/habitat/objects/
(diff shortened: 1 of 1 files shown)


docs/overview/benchmark-experiments.md
```diff
@@ -174,12 +174,20 @@ Note: To obtain these results, pretraining was run without parallelization acros
...
 > ```
 >
-> Then create a symlink so that the experiment configs (which look for `~/tbp/data/habitat/objects/compositional_objects`) can find the versioned dataset folder:
+> Then create a symlink so that the experiment configs (which look for `~/tbp/data/habitat/objects/compositional_objects`) can find the versioned dataset folder. If this is the first time you are doing this, you can use:
 >
 > ```plaintext
 > mkdir -p ~/tbp/data/habitat/objects/
 >
-> ln -s ~/tbp/data/habitat/versioned_data/compositional_objects_1.1 ~/tbp/data/habitat/objects/compositional_objects
+> ln -s ~/tbp/data/habitat/versioned_data/compositional_objects_1.2 ~/tbp/data/habitat/objects/compositional_objects
```
vkakerbeck, line 182: just fyi, if there are different options of commands, you an also render it as tabs like this: https://github.com/thousandbrainsproject/tbp.monty/blob/9b435f78fb715c43e7d27a9bb63bf0e311b22423/docs/how-to-use-monty/getting-started.md?plain=1#L74 (looks like this: https://docs.thousandbrains.org/docs/getting-started#22-setup-monty)

hlee9212 (approved):
LGTM, thanks!

PR #1051, opened 2026-06-22
feat!: SM always sends up movement regardless of feature change

ramyamounir:
## Breaking Changes
- **CMP Messages:** `Message` argument `use_state` is renamed to `pass_message`, and a new required argument `process_features_in_lm` is added.
- **Updater Protocol:** `HypothesesUpdater.update_hypotheses` is split into `displace_hypotheses` + `update_evidence`.
- **Displacer Protocol:** `HypothesesDisplacer.displace_hypotheses_and_compute_evidence` is split into `displace_hypotheses` + `compute_evidence`.
- **Buffer:** `FeatureAtLocationBuffer.locations` is now a single `(N, 3)` array instead of a per-channel dict. 
- **Learning Module:** `EvidenceGraphLM.get_current_mlh()` is now private (`_get_current_mlh()`).
- **Other:** SMs now always send movement on non-motor-only steps and use `process_features_in_lm` to signal significant feature change; LMs displace hypotheses on location-only steps.

Currently, the SM would filter messages to only send a CMP message to the LM if there is a significant feature change. The filter sets `use_state=False` on the message, and the `Monty` class does not route it to the LM. Ideally, we want all LMs to know the agent's location at any step (to store co-located features from any channel using a single shared location), and this coupling between feature change and movement is a blocker.

This PR:
- ensures that an SM always sends all movements (except for `motor_only_steps`) to update the LMs' sense of sensors' location.
- adds a new message flag `process_features_in_lm` to indicate whether this message contains significant feature change and whether it should be used for a full matching step in the LM.
- simplifies the `FeaturesAtLocationsBuffer`. The buffer locations are now an array of shape (N, 3). These do not condition on the input channel anymore. The buffer locations only store locations where the LM received any messages with `process_features_in_lm==True`, and not every agent location.
- adds a new `_displace_hypotheses` to the LMs which allows for displacing hypotheses on location-only messages.
- ensures graph building now use the new shared buffer locations for all the input channel graphs.
- splits the displacer function to separate `displace_hypotheses` and `compute_evidence`. This way we can displace the hypotheses on location only steps. This change touches code in the LMs, updaters and displacers.
- adds a few more buffer and feature change filter unit tests.

This change allows us to build heterarchy compositional models where the 2D SM-LM inputs are colocated with the 3D SM-LM inputs in the HL-LM 3D space, as shown below.

012_cylinder_tbp_horz_child_objects

---

**This change may require pretraining the models, not just inference benchmarks.**

PR #1136, opened 2026-08-09
feat!: 3LM Heterarchy compositional configs

hlee9212:
## Breaking Changes
- Completely replaces the 2LM Compositional Configs to use 3LM Heterarchy architecture. 

## Summary of Changes

- Introduces 3LM versions of compositional configs using heterarchy architecture (previously 2LM-hierarchical). 
- Update to docs to point to `pretrained_compositional_objects_v6`. 
- Updates `logos_on_objects.csv` benchmarks. 

## Further Changes Made from PR Comments
- `min_train_steps = 3` (from 100) for compositional pretraining. This increases the number of points sent by LL-LM to HL-LM (and we have a few points from 2D LL-LM as well, which we didn't have before). 
- `num_voxels_per_dim = 100` for all LMs. This overall reduces the number of points per object.
- Removed `object_evidence_threshold: 5`. Does not affect performance. 

### base_infer_objects_with_stickers_monolithic_models

| correct_child_or_parent (%) | 72.56 | 59.02 | -13.54 | ❌ |
| correct_parent (%) | 41.73 | 36.84 | -4.89 | ❌ |
| used_mlh (%) | 50.38 | 55.64 | 5.26 | ❌ |
| match_steps | 95 | 90 | -5 | ✅ |
| rotation_error (deg) | 50.15 | 35.97 | -14.18 | ✅ |
| avg_prediction_error | 0.26 | 0.47 | 0.21 | ❌ |
| runtime (min) | 41 | 40 | -1 | ✅ |
| num_episodes | 266 | 266 | 0 | ⬜ |

### base_infer_objects_with_stickers_comp_models

| correct_child_or_parent (%) | 68.05 | 58.27 | -9.78 | ❌ |
| correct_parent (%) | 34.59 | 35.34 | 0.75 | ✅ |
| used_mlh (%) | 38.72 | 56.02 | 17.3 | ❌ |
| match_steps | 88 | 90 | 2 | ❌ |
| rotation_error (deg) | 53.16 | 41.68 | -11.48 | ✅ |
| avg_prediction_error | 0.29 | 0.47 | 0.18 | ❌ |
| runtime (min) | 35 | 41 | 6 | ❌ |
| num_episodes | 266 | 266 | 0 | ⬜ |

## Learned Models 
A reference of the learned models in `pretrained_compositional_objects_v6` for reference. 

CleanShot 2026-08-24 at 13 38 24
Note: Monolithic has the same points as compositional `patch_2` (but none from LL-LMs, as expected). 
image

<details>
<summary>Old Description</summary>
## Additional Notes
- Instead of overriding existing `supervised_pre_training_objects_with_stickers_*` and `base_infer_objects_with_stickers_*`, I created new files with `_3lm` suffix because I think it will actually be easier to compare than using Github's diff. 
- A follow-up PR is planned to remove the existing 2LM architecture and rename to rename `_3lm`, completing the replacement. It will also remove `correct_child_or_parent` metric. 

### Diagram of 3LM Architecture
image

### Plots of Learned Models
image
image
- The learned models in monolithic HL-LM is identical to the 2LM version. Recall monolithic only has input from `patch_2`. When comparing just points from patch_2, the models between comp. and monolithic are identical. Again, comp models have slightly more points, but these are coming from 3D LL-LM.
- With our current policy, not all objects have points from LL-LMs. Only 10/19 objects have points from 3D LL-LM (notably disk and cylinder objects do not have points from 3D LL-LM, and I think we have observed before that these are hard to distinguish because top of cylinder looks like a disk).
- No objects have have points from learning_module_1 (2D LL-LM), but I think the attention-related policies will improve this.
</details>


nielsleadholm:
Thanks very much @hlee9212 , great to see the configs reach this point! 

Can you please remind me, with this final "shift" (adding bursting, 2DSM, and 3LM architecture), did you have a check for the monolithic objects showing that the results here are identical to before these changes, if GSG in LL-LMs are deactivated in both?

hlee9212:
> Can you please remind me, with this final "shift" (adding bursting, 2DSM, and 3LM architecture), did you have a check for the monolithic objects showing that the results here are identical to before these changes, if GSG in LL-LMs are deactivated in both?

Hi Niels, the 3LM monolithic has the same performance as 2LM monolithic (as in, identical HL-LM accuracy) when:
1. GSG deactivate in all LMs for 3LM and 2LM architectures
2. No burst sampling in all LMs for 3LM and 2LM architecture
3. The 2D LL-LM is not contributing to matching steps in 3LM architecture

The Shortcut story was [here](https://app.shortcut.com/thousandbrainsproject/story/7211/run-inference-in-3lm-versions-of-monolithic-and-comp-models).

ramyamounir:
Thanks @hlee9212! I haven't reviewed the configs yet, but I decided to start from inspecting the pretrained models first, specifically the comp models in `learning_module_2`. I wouldn't expect these models to have nodes from the 2D channel because of the policy, but I would expect them to have received input from the 3D LL-LM (`learning_module_0`). Many of them only have `patch_2` channels as you can see below. Some have input from the 3D LL-LM, which means that it's not a connectivity between LMs issue. This may not necessarily be an issue with the configs, but looking into it may lead to something bigger that we've missed.

image

My thinking is that, for some objects, the 3D LL-LM is unable to narrow down to one child object and reach tc. For example, when looking at the disk, it may think that it could also be the top of the cylinder. I'm seeing this issue on the disk, sphere and cylinder objects in the pretrained models. I think the sphere is easily distinguishable from the other objects, so I will investigate further with teleop. Have you tried learning a compositional model with teleop interactive policy to see if in theory it can learn a model with both LL-LM input channels?

ramyamounir (commented):
The new configs look great to me! I just left some non-blocking comments [here](https://app.shortcut.com/thousandbrainsproject/story/7017/review-3lm-heterarchy-configs#activity-7465)

src/tbp/monty/conf/constants/defaults.yaml
```diff
@@ -33,7 +33,8 @@ pretrained_dir: ${path.expanduser:${oc.env:MONTY_MODELS}/pretrained_ycb_v13}
 # v3 : Trained after fixing the monolithic experiments connectivity issues (PR#942)
 # v4 : Trained after compositional benchmark config refactor (PR#1070)
 # v5 : Updated pretrained models for compositional_objects_1.2 (symmetrically ambiguous objects removed), and fix to monolithic supervised IDs
-compositional_pretrained_dir: ${path.expanduser:${oc.env:MONTY_MODELS}/pretrained_compositional_objects_v5}
+# v6 : Updated pretrained models for the current compositional benchmark release
```
nielsleadholm: Can you expand on this a bit> "Current compositional benchmark release" won't be particularly meaningful in the future, but something like adding 2D SM and burst sampling, and using 3LMs total, would be.

src/tbp/monty/conf/env_interface/transform/missing_depthto3d_sensor4.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - 64
+  - - 64
+    - 64
+  - - 64
+    - 64
+  - - 64
+    - 64
+  world_coord: true
+  zooms:
+  - 30.0
```
nielsleadholm: As Ramy pointed out, this order of zoom values does not match the configuration we use elsewhere. We should remove this config if it is not being used, or update otherwise.

src/tbp/monty/conf/environment/habitat_dist_agent_sensors3.yaml
```diff
@@ -0,0 +1,76 @@
...
+  seed: 42
+  data_path: ${path.expanduser:${oc.env:MONTY_DATA}/habitat/objects/compositional_objects}
+  agents:
+    agent_args:
+      agent_id: agent_id_0
+      sensor_ids:
+      - patch_0
+      - patch_1
+      - patch_2
+      - view_finder
```
nielsleadholm, line 20: Can I check what sensors3 vs sensor4 refers to? I'm assuming it's the three different patches, plus the view-finder? If so however, the configs seem to be inconsistent about which naming convention they use.

src/tbp/monty/conf/experiment/supervised_pre_training_objects_with_stickers_2d_children_3lm.yaml
```diff
@@ -0,0 +1,32 @@
+# @package _global_
+
+defaults:
+  - /monty: evidencegraph_exp1000_e3_t3_tot2500
+  - /monty/motor_system_config: naive_scan_1
+  - /monty/learning_module: evidence_3lm_heterarchy
+  - /monty/sensor_module: 3sm_2d3d_dist
+  - /monty/connectivity: 3lm_3sm_heterarchy
+  - /environment: habitat_dist_agent_sensors3
+  - /env_interface: train_objects_with_stickers_2d_children_predefined_3lm
```
nielsleadholm, line 10: Nitpick but in the config `train_objects_with_stickers_3d_children_predefined`, you don't make any mention of the number of LMs. I think this is better, since as far as I can see, the `env_interface` doesn't really care about the number of LMs.

src/tbp/monty/conf/monty/learning_module/evidence_3lm_heterarchy.yaml
```diff
@@ -0,0 +1,78 @@
...
+    elapsed_steps_factor: 10
+    min_post_goal_success_steps: 5
+    x_percent_scale_factor: 0.75
+    desired_object_distance: 0.03
+
+learning_module_1:
+  _target_: tbp.monty.frameworks.models.evidence_matching.learning_module.EvidenceGraphLM
+  feature_weights:
+    patch_1:
+      hsv: ${np.array:[1, 0.5, 0.5]}
```
nielsleadholm, line 32: For my understanding, can you clarify the origin of using different feature weights here for LM-0?

src/tbp/monty/conf/monty/sensor_module/3sm_2d3d_dist.yaml
```diff
@@ -0,0 +1,90 @@
...
+    coherence_threshold: 0.5
+    radius: 14.0
+    sigma_r: 7.0
+    depth_edge_threshold: 0.01
+    max_center_offset: null
+  noise_params: null
+  features:
+    - pose_vectors
+    - pose_fully_defined
+    - principal_curvatures
```
nielsleadholm, line 46: Is `principal_curvatures` really in the 2D SM config?

tests/conf/snapshots/supervised_pre_training_objects_with_stickers_2d_children_3lm.yaml
```diff
@@ -0,0 +1,464 @@
...
+              pose_vectors: ${np.array:[0, 2, 1]}
+          tolerances:
+            patch_1:
+              hsv: ${np.array:[0.1, 0.2, 0.2]}
+          max_match_distance: 0.001
+          x_percent_threshold: 20
+          evidence_threshold_config: all
+          max_graph_size: 0.1
+          num_model_voxels_per_dim: 100
+          object_evidence_threshold: 5
```
nielsleadholm: Is there a reason only LM-1 in all of the configs has `object_evidence_threshold` specified? And how did you arrive at a value of 5?

tests/conf/snapshots/supervised_pre_training_objects_with_stickers_monolithic_models_3lm.yaml
```diff
@@ -0,0 +1,452 @@
+experiment:
+  config:
+    do_eval: false
+    do_train: true
+    monty_config:
+      monty_class: ${monty.class:tbp.monty.frameworks.models.evidence_matching.model.MontyForEvidenceGraphMatching}
+      monty_args:
+        num_exploratory_steps: 1000
+        min_eval_steps: 3
+        min_train_steps: 3
```
nielsleadholm, line 10: As Ramy also pointed out, it seems `min_train_steps` is different from the compositional config (which is set to 100)?

nielsleadholm (commented):
Great effort Hojae. Overall looks good, just a few small pieces that could do with clarifying. Most of these were brought up by Ramy in the story he linked.

nielsleadholm:
Hey @hlee9212 , I believe after Ramy's recent PR, you may need to re-train the models, as this affects the location representations of the LL-LM channels: https://github.com/thousandbrainsproject/tbp.monty/pull/1051

Regardless, it would alsoo be worth addressing some of the comments from Ramy and myself. One point that you and I ummed and ahhed on but on further discussion with Ramy would be worth changing - I think for all of these configs, it would be worth setting `num_model_voxels_per_dim: 100` rather than 200. Any deviation from a simple number in configs i) adds cognitive overhead ("Why is it this number? Was there something special found?"), and ii) feels like over-engineering for a particular task. I think just setting it to 100 would be easier, even if it results in a marginal loss in performance.

src/tbp/monty/conf/constants/defaults.yaml
hlee9212: Good catch. I noticed the last commit I had to fix this wasn't pushed and the old vague description was used. I'll use TODO in the future so Pycharm notifies me whenever a new TODO is added. Fixed.

src/tbp/monty/conf/env_interface/transform/missing_depthto3d_sensor4.yaml
hlee9212: Deleted.

src/tbp/monty/conf/environment/habitat_dist_agent_sensors3.yaml
hlee9212: Tagging @tristanls-tbp as he may remember why this convention was used. I called this `_sensors3.yaml` because there was already a `habitat_dist_agent_sensors2.yaml` and `habitat_dist_agent_sensors5.yaml` that has 2 (or 5) patch sensors + 1 viewfinder. But it seems it seems configs under transforms, we already had `missing_depthto3d_sensor2_semantic0.yaml` where the 2 refers to the patch + viewfinder (i.e. the inconsistencies already existed, and I followed the convention for each folder). 

I think fixing the file names for several of these would be a separate PR.

src/tbp/monty/conf/experiment/supervised_pre_training_objects_with_stickers_2d_children_3lm.yaml
hlee9212: Yes, I created a `_3lm` version because I didn't want to mess up the existing 2LM config that uses `train_objects_with_stickers_3d_children_predefined`. I'll change it back to this in follow-up PR where I completely remove the 2LM configs and get rid of the `_3lm` suffix.

tests/conf/snapshots/supervised_pre_training_objects_with_stickers_monolithic_models_3lm.yaml
hlee9212: Here's the reply I wrote to Ramy:

> This is because the 2LM version used evidencegraph_exp1000_e3_t3_tot2500 for monolithic and evidencegraph_exp1000_e3_t100_tot2500 for comp. Again, because we were debugging a lot on 2LM vs. 3LM differences in performance, this setting has been copied from 2LM version. I did fix this before, but changed it back when I had to really dig down to find the cause of performance differences.

---
I'll update to use min_train_steps = 3 for both monolithic and comp and check performance. I don't see a reason for performance difference at this point.

tests/conf/snapshots/supervised_pre_training_objects_with_stickers_2d_children_3lm.yaml
hlee9212: I got the value of 5 from @ramyamounir's compositional configs (see `evidence_3lm_heterarch.yaml`): https://github.com/ramyamounir/tbp.monty/commit/5fc780b8a1cfecb9952d1af0589ca337e19b7a2c#diff-ba36787e050114cbed0c6915987f3ff3ab905e835627a7d10b5dd3398a14e0e3

@ramyamounir do you remember if this was a setting specific to be used with `tbp.teleop`? My memory was that you set this lower (from 20) so LL-LM can have a match to send up to HL-LM (from a TRM long time ago.. 😅).

src/tbp/monty/conf/monty/sensor_module/3sm_2d3d_dist.yaml
hlee9212: Yes, and this is a required feature to pass because 2D SM need principal_curvatures (k1, k2) to perform arc-length correction.

tests/conf/snapshots/supervised_pre_training_objects_with_stickers_2d_children_3lm.yaml
ramyamounir: Yeah, this was before we had the sending of LM messages gated by reaching terminal conditions. Before we made this change, we would keep sending LM messages above `object_evidence_threshold`, so I raised it for the teleop demo. We shouldn't need this now because it has to reach terminal conditions before sending anyway, I think we can go back to the default `1` by removing it from the configs.

Sorry, looks like I had a few hacks in my configs back then :sweat_smile: should have documented them better.

hlee9212: Thanks @ramyamounir! 🙏 No worries. I'll remove the `object_evidence_threshold`.

src/tbp/monty/conf/environment/habitat_dist_agent_sensors3.yaml
nielsleadholm: Ok thanks, yeah something we can consider cleaning up / clarifying at some point.

src/tbp/monty/conf/monty/sensor_module/3sm_2d3d_dist.yaml
nielsleadholm: Ah right, thanks.

tests/conf/snapshots/supervised_pre_training_objects_with_stickers_monolithic_models_3lm.yaml
nielsleadholm: Perfect, thanks Hojae.

src/tbp/monty/conf/experiment/supervised_pre_training_objects_with_stickers_2d_children_3lm.yaml
hlee9212: Marking resolved as I captured this in [Shortcut](https://app.shortcut.com/thousandbrainsproject/story/7212/follow-up-pr-remove-correct-child-or-parent-etc) for follow-up PR.

tests/conf/snapshots/supervised_pre_training_objects_with_stickers_2d_children_3lm.yaml
hlee9212: Marking this resolved as I tested removing `object_evidence_threshold` and did not see any changes in performance as expected.

src/tbp/monty/conf/monty/learning_module/evidence_3lm_heterarchy.yaml
hlee9212: Sorry about the late reply on this one, was trying to dig several things. 

1. I presume you were asking about LM1 (i.e. 2D LL-LM), and not LM0? 
2. I presume you meant why is this different from our usual `hsv: ${np.array: [2, 0.5, 0.5]}` and not why is it different from `feature_weights: {}` in LM0? 

Let me know if I am making the right assumptions. 😅

If yes:
The `hsv: ${np.array: [1, 0.5, 0.5]}` was the setting I used when testing TwoDSensorModule. I gave this setting to Ramy when he was working on debugging compositional training. Scott made the change to `hsv: ${np.array: [2, 0.5, 0.5]}` afterwards during his IP, and I found the [Shortcut comment by Ramy](https://app.shortcut.com/thousandbrainsproject/story/6288/update-the-hierarchical-monty-to-use-the-2d-sm-lm-for-learning-the-stickers#activity-6665) that it should have been updated (the "Yes, I set these configs up before Scott changed them.").

tristanls-tbp:
Hi, I noticed that this PR is marked with `!` indicating a breaking change. Could you please explain what the breaking change is in the description under a `## Breaking Changes` header or something similar?

hlee9212:
> Hi, I noticed that this PR is marked with ! indicating a breaking change. Could you please explain what the breaking change is in the description under a ## Breaking Changes header or something similar?

Hi @tristanls-tbp, I added a `## Breaking Changes` above. Apologies, I thought `## Summary of Changes` I had was sufficient.

src/tbp/monty/conf/monty/learning_module/evidence_3lm_heterarchy.yaml
nielsleadholm: Thanks, yes sorry meant LM1. Ok, in that case it would be worth changing this to ${np.array: [2, 0.5, 0.5]} so that we are using this everywhere and don't have an inconsistency.

nielsleadholm: Yes sorry I meant LM1. Ok thanks, per the discussion on Shortcut, we can leave this for now. In the future, we may want to make this more principled and simplify everything to be the same, unless we have a good reason to deviate.

nielsleadholm (approved):
LGTM!

PR #1137, opened 2026-08-10
feat!: structured telemetry MVP

AgentRev:
## Summary
Introduces a structured telemetry module MVP, addressing Injection <span>#</span>81 "Use structured logging to emit telemetry inline", relying on pub/sub mechanics provided by Python's `logging` module. The schemas are stored in the internal dict of `logging.LogRecord` objects via `logger.log()`'s `extra` param, as documented [here](https://docs.python.org/3/library/logging.html#logging.Logger.debug).

See this forum topic for more details: https://forum.thousandbrains.org/t/333-telemetry-collection-is-configurable-down-to-the-python-module-level/1076/35

### Core Module (`tbp.monty.telemetry`)
* **API (`__init__.py` via `telemeter.py`)**: Added `getTelemeter(__name__)` helper for clean instantiation across modules, and a factory for telemetry publishers.
* **Publishers (`publishers.py`)**: Added `Logger` subclass `TelemetryPublisher`, obtainable via `telemetry.getTelemeter(...)`, isolated from root logger.
* **Schemas (`schemas.py`)**: Added `TelemetrySchema` and `TelemetryEvent` Pydantic models for structured schemas. Can be subclassed, or used as-is on the spot.

* **Unit tests (`telemetry_test.py`)**: Validates all functionality of schemas, events, and publishers.

###  Integration Examples
* **`MontyBase`**: Emits `GlobalMatchingStep` telemetry events on each global matching step execution.
* **`GraphMemory`**: Emits `NewGraphAdded` telemetry events when new graphs are stored in memory.

## Notes
* Schemas can be processed on emission in an event-like fashion, by implementing a custom [`logging.Handler`](https://docs.python.org/3/library/logging.html#handler-objects), with an `emit(record)` method that calls `record.__dict__.get("telemetry_schema")`, then assigning the handler to the appropriate namespace via `telemeter.addHander(...)`.
  
  At present, there is no wrapper provided; see `TelemetryPublisherTest` for an example. However, a wrapper is planned in the near future via "subscribers", i.e. managed `LogHandler` subclasses within the telemetry module.

## Usage Example
```python
from tbp.monty import telemetry
from tbp.monty.telemetry.schemas import TelemetryEvent

telemeter = telemetry.getTelemeter(__name__)
telemeter.info(TelemetryEvent(kind="CustomEvent", your_key="your_value", ...))
```


AgentRev:
@tristanls-tbp Two things to settle before I can mark this as ready:

- Which experiment would you like me to provide a telemetry config example for? (and/or should there be a default telemetry config for all experiments?)
- Did you expect my `TelemetrySubscriber` to be part of this first PR, or is it fine without?

tristanls-tbp:
Thank you, @AgentRev, for putting this together.

The PR is fine without the `TelemetrySubscriber`.

As for the telemetry config, I think all experiments should be updated eventually 🙂. For the first PR, how about the tutorial experiments? Here's the small change that @jeremyshoemaker made to load the logging configuration: https://github.com/thousandbrainsproject/tbp.monty/pull/987.

src/tbp/monty/frameworks/telemetry/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+CRITICAL = logging.CRITICAL
+FATAL = logging.FATAL
+ERROR = logging.ERROR
+WARNING = logging.WARNING
+INFO = logging.INFO
+DEBUG = logging.DEBUG
+NOTSET = logging.NOTSET
+
+
+def getTelemeter(*args, **kwargs):  # noqa: N802 - lowercase
```
jeremyshoemaker: I think we should comment here on _why_ this is not following the typical Python style, because it's not immediately obvious.

It looks like you're following the lead of the `logging` module, and I don't know why it never stood out to me before that `getLogger` also breaks with the Python style.

Also, I don't know if it's relevant to your design, but if you're following `logging`'s lead, then this comment in the documentation of `getLogger` seems like it could lead people to make incorrect assumptions here.

> All calls to this function with a given name return the same logger instance. This means that logger instances never need to be passed between different parts of an application.
> ~ [Python Documentation - logging module](https://docs.python.org/3/library/logging.html#logging.getLogger)

In all my time working with Python logging, I'm not sure if I've ever relied on this behaviour because I usually pass it the module `__name__`, which is going to be unique, so maybe it's not anything to worry about. I just thought I'd point it out.

tristanls-tbp: **question**: What is the `- lowercase` annotation doing? As far as I remember, we don't usually explain rule numbers in our `noqa` directives.

src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -287,6 +294,9 @@ def init_loggers(self, logging_config: dict[str, Any]) -> None:
         for handler in logger.handlers:
             logger.removeHandler(handler)
 
+        for handler in telemeter.event_logger.handlers:
```
tristanls-tbp: **issue**: It looks like `TelemetryPublisher` is missing a `handlers` accessor and a `removeHandler` method. We shouldn't be reaching through `event_logger` here. The API should mirror the `logger` API.

src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -320,6 +330,8 @@ def init_loggers(self, logging_config: dict[str, Any]) -> None:
             "incremental": True,
         }
         final_config.update(logging_config)
+        if telemetry_config:
```
tristanls-tbp: **issue**: I find the mix of default `{}`, and `telemetry_config` being optional (could be `None`), and this conditional check confusing.

The problem is:
| value | `if telemetry_config:` | `if telemetry_config is not None:`
|---|---|---
|`None`|False|False
|`{}`|False|**True**

But, the only place where the input for `init_loggers` is generated uses `self.config.get("telemetry", {})`, which will set the default `{}`. 

**suggestion**: I think this will be easier to parse if `telemetry_config` is a required parameter in `init_loggers` and not allowed to be `None`. That way, we don't have to worry about the `None` case.

src/tbp/monty/frameworks/telemetry/schemas.py
```diff
@@ -0,0 +1,51 @@
...
+    @field_validator("kind")
+    @classmethod
+    def validate_kind(cls, value):
+        return value or cls.__name__  # schema class name fallback
+
+
+class TelemetryEvent(TelemetrySchema):
+    """Base model class for telemetry events; carries instantaneous data changes."""
+
+    values: Mapping = {}
```
tristanls-tbp: **issue**: I think this approach forces knowledge of `values` into any callsite, e.g.,:
```python
        telemeter.info(
            TelemetryEvent(kind="NewGraphAdded", values={"graph_id": graph_id})
        )
```

**suggestion**: Remove and add `model_config = ConfigDict(extra="allow")`. This way, the callsites look like:
```python
        telemeter.info(
            TelemetryEvent(kind="NewGraphAdded", graph_id=graph_id)
        )
```
With this suggestion, a future
```python
class NewGraphAdded(TelemetryEvent):
    graph_id: str
```
becomes equivalent without a need to rewrite the callsite, since:
```python
        telemeter.info(
            TelemetryEvent(kind="NewGraphAdded", graph_id=graph_id)
        )
```
and
```python
        telemeter.info(NewGraphAdded(graph_id=graph_id))
```
result in the same behavior.

src/tbp/monty/frameworks/telemetry/schemas.py
```diff
@@ -0,0 +1,51 @@
...
+    kind: Annotated[str, Field(validate_default=True)] = ""
+    """Schema identifier used by telemetry loggers as the log message in text sinks.
+    It can also be used for event filtering by subscribed handlers.
+    If empty, defaults to schema class name."""
+
+    timestamp: float = Field(default_factory=time.time, kw_only=True)
+    """Unix time in seconds when the schema was instantiated."""
+
+    origin: str = Field(default_factory=lambda: sys._getframe(2).f_globals["__name__"])
+    """Name of module where schema was instantiated. Auto-populated, but overridable."""
```
tristanls-tbp: **question**: What is the use case for including these fields (`timestamp`, `origin`)?

src/tbp/monty/frameworks/telemetry/__init__.py
tristanls-tbp: **suggestion**: It seems like memoizing the telemeter so that all calls return the same instance for a given name is something to copy from `logging` implementation.

tests/integration/frameworks/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,197 @@
...
+
+        for log_func, _, event in events:
+            log_func(event)
+
+        self.assertEqual(len(self.handler.records), len(events))
+
+        for record, (_, level, event) in zip(self.handler.records, events):
+            self.assertEqual(record.levelno, level)
+            self.assertEqual(record.msg, event.kind)
+            self.assertIs(record.__dict__.get("telemetry_schema"), event)
```
tristanls-tbp: **issue**: Please put unit tests in `tests/unit` with other unit tests. This will accomplish two things. It'll separate fast tests from slow ones. Also, it will greatly reduce the MuJoCo duplicate integration test scope, which seems to have just one method for now.

tests/integration/frameworks/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+class TelemetryPublisherTest(unittest.TestCase):
+    """Unit tests for TelemetryPublisher."""
+
+    def setUp(self):
+        self.handler = TelemetryLogHandler()
+        self.publisher = TelemetryPublisher(__name__)
+        self.publisher.event_logger.addHandler(self.handler)
+        self.publisher.event_logger.setLevel(logging.DEBUG)
```
tristanls-tbp: **issue**: This demonstrates insufficient API for `TelemetryPublisher`. It needs `addHandler`, `setLevel`, etc. Nothing should be reaching through publisher internals (`*.event_logger.*`) like this.

src/tbp/monty/telemetry/publishers.py
```diff
@@ -0,0 +1,69 @@
...
+        """Emits a structured telemetry event at the specified log level.
+
+        Args:
+            level: The log level.
+            event: The event instance.
+        """
+        self.event_logger.log(
+            level=level,
+            msg=event.kind,
+            extra={"telemetry_schema": event},
```
tristanls-tbp: **question**: I assume you're doing the right thing by anticipating a future use case, but for now, I don't see a reason to place the `event` inside the `telemetry_schema` property rather than using `extra={**event}` (not sure the literal code works, but trying to communicate spreading the keys and values into `extra`). Could you elaborate on the design behind this approach? `message` already contains the name of the kind for any consumer to dispatch on.

tests/integration/frameworks/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,197 @@
...
+                    f"experiment.config.logging.output_dir={self.output_dir}",
+                    "+telemetry=info",
+                ],
+            )
+
+    def tearDown(self):
+        """Clean up temporary directory."""
+        shutil.rmtree(self.output_dir)
+
+    def test_global_matching_step_telemetry(self):
```
tristanls-tbp: **praise**: I'm glad you're thinking about proving that telemetry works for the experiments end-to-end.

**note**: I'm wondering about the test's brittleness and value, though. What is being tested here? Configuration? End-to-end assembly? Nevertheless, I don't have a better idea yet, so I'll think some more on it for now.

tristanls-tbp (commented):
Exciting to see! Thank you for putting together telemetry @AgentRev !

**issue**: Now that this is being added, could you please move it from `src/tbp/monty/frameworks/telemetry` to `src/tbp/monty/telemetry`. Similarly for the test. We are slowly abandoning the `frameworks` folder.

src/tbp/monty/frameworks/experiments/monty_experiment.py
AgentRev: My goal here was to try and keep the PR non-breaking, so I first wrote `get("telemetry", {})` and `telemetry_config: dict[str, Any] = {}`. Ruff didn't like that, so I made it optional and added the `if telemetry_config:`, but didn't think to remove the first `{}`. Since you want it to be required, I'll clean that up and change the PR title to `feat!`.

src/tbp/monty/frameworks/experiments/monty_experiment.py
AgentRev: This was just a stop-gap solution to palliate the temporary absence of my subscriber implementation, which already abstracts handler management away; cleanup becomes the responsibility of the subscribers' creators rather than the parent experiment's. I think the lifetime of a subscriber should not be decided by the experiment, as its creator might very well be outside the experiment.

Ideally, I think only the internals of publishers and subscribers should be aware about the existence of telemetry logger instances and their full list of handlers, as those are just implementation details. I don't think they should provide outside access to them, and as such, I don't think the logger API should be mirrored for that portion.

I'm aware of the irony that these 2 lines here represent the polar opposite of my own opinion, but yeah, it's temporary. I can put in a `clear_handlers` method to encapsulate them if you wish, but I'd most likely remove it in the next phase.

All open to debate, of course. 🙂

src/tbp/monty/frameworks/telemetry/__init__.py
AgentRev: I thought I would write the reason as to why this particular `noqa` is here, since it's not immediately evident what this rule number does, and that it is seldom ever encountered elsewhere. I can remove it if you want.

src/tbp/monty/telemetry/publishers.py
AgentRev: It's to enable the `isinstance` calls I used with derived schema classes in the graph matching telemetry, which allows the schema's type hints to work end-to-end. If you gut the schema to stuff its values inside the record, you lose the type hints.

Also, records contain a bunch of other stuff at the same level where the contents of `extra` end up: https://docs.python.org/3/library/logging.html#logrecord-attributes

Although, to think of it, it should possible with `extra={**event}` to regain type hints on the other end by doing something like:
```
if record.msg == DerivedTelemetryEvent.__name__:
    derived_event = DerivedTelemetryEvent.model_validate(record.__dict__, from_attributes=True)
```

If I were to pick one of either, I'd probably keep `extra={"telemetry_schema": event}`, but it's not a strong preference. I'll let you call it.

src/tbp/monty/frameworks/telemetry/schemas.py
AgentRev: Just in case, I guess; currently unused. I ended up realizing that log records already provide nearly-identical equivalents via their `created` and `name` attributes. I will remove.

tests/integration/frameworks/telemetry/telemetry_test.py
AgentRev: The `setLevel` was a stop-gap for the lack of Hydra in this test, which resulted in all telemetry loggers sitting at `NOTSET` level. I don't want to add a `setLevel` to the publisher, since Hydra already takes care of log levels. I'll add Hydra here with a `"telemetry=debug"` override and remove the `setLevel`.

As for the handler, see my reply to your other related comment: https://github.com/thousandbrainsproject/tbp.monty/pull/1137/changes#r3875885519

tests/integration/frameworks/telemetry/telemetry_test.py
AgentRev: It's testing the sum of all parts, end-to-end. I couldn't think of a "better" way either, if there is one.

Perhaps this test is not necessary if I move the Hydra part to `TelemetryPublisherTest`.

src/tbp/monty/frameworks/telemetry/__init__.py
jeremyshoemaker: This is an instance of a comment saying _what_ and not _why_. 

It's related to [my comment on this line of code](https://github.com/thousandbrainsproject/tbp.monty/pull/1137#discussion_r3875811476) above. The [N802 rule](https://docs.astral.sh/ruff/rules/invalid-function-name/) is actually called "invalid-function-name" -- I don't think invalid is the right word here 🤷 -- not "lowercase". It's complaining because the function name is not `get_telemetry` and less about the fact that there's a `T` in it.

But the _reason_ we're ignoring it here is because we're following the pattern that `logging` sets.

src/tbp/monty/frameworks/experiments/monty_experiment.py
AgentRev: For this PR, since there isn't an immediate need for anything handler-related AFAIK, I'm thinking about removing everything involving `event_logger.handlers`, and make `event_logger` private (`_event_logger`). Anyway, I realized these 2 lines here in particular are not even needed at all regardless, since `logger` here is `telemetry.tbp.monty`, which exists only to act as the global log level and isn't supposed to emit or handle anything.

About the publisher API, I suggest we postpone discussion until we move on to the next phase.

src/tbp/monty/frameworks/telemetry/schemas.py
jeremyshoemaker: I hadn't noticed this when I skimmed this before, but what is the `kind` field being used for? I ask because I'm always suspicious of things that look like we're using strings to name things like classes and functions without a good reason for the indirection. It tends to break the ability of tooling to recognize the link between the thing being named and the name in a string.

AgentRev: @jeremyshoemaker `kind` was first suggested here by Tristan: https://forum.thousandbrains.org/t/333-telemetry-collection-is-configurable-down-to-the-python-module-level/1076/13

It is meant to be an identifier that tells subscribers (handlers) what the event contains. The docstring I put for now is:
> Schema identifier used by telemetry loggers as the log message in text sinks. It can also be used for event filtering by subscribed handlers. If empty, defaults to schema class name.

In this PR, there are no custom schema classes, it's only a generic `TelemetryEvent`, so `kind` contains a custom value chosen by the researcher (see Usage Example in PR description).

In my  [`telemetry_graph_loggers_wip`](https://github.com/AgentRev/tbp.monty/tree/telemetry_graph_loggers_wip) branch however, I do have custom schemas ([example](https://github.com/AgentRev/tbp.monty/blob/fe55e1a9407d093624fc9be6a24007d66938e9f0/src/tbp/monty/frameworks/telemetry/episode/schemas.py#L95)), where `kind` is not currently used anywhere beyond the `msg` arg of [`Logger.log`](https://docs.python.org/3/library/logging.html#logging.Logger.log). Instead, subscribers do an `isinstance` check on the `telemetry_schema` value from the log record, then dispatch it where it needs to go ([example](https://github.com/AgentRev/tbp.monty/blob/fe55e1a9407d093624fc9be6a24007d66938e9f0/src/tbp/monty/frameworks/telemetry/episode/subscribers.py#L577-L580)).

Ultimately, the necessity of  `kind` boils down to whether generic schemas should be supported or not, and also Tristan's future answer to [this comment](https://github.com/thousandbrainsproject/tbp.monty/pull/1137#discussion_r3878632793).

---

> I'm always suspicious of things that look like we're using strings to name things like classes and functions without a good reason for the indirection.  It tends to break the ability of tooling to recognize the link between the thing being named and the name in a string.

Indeed, I'm always careful about that. In addition, if I can't avoid putting a class / module name in a string, I write `Thing.__name__` rather than hardcoding `"Thing"`. ([example](https://github.com/AgentRev/tbp.monty/blob/fe55e1a9407d093624fc9be6a24007d66938e9f0/src/tbp/monty/frameworks/telemetry/episode/subscribers.py#L544-L545))

jeremyshoemaker: That makes sense. Thanks for the detailed reply.

src/tbp/monty/frameworks/experiments/monty_experiment.py
tristanls-tbp: > My goal here was to try and keep the PR non-breaking

No need. Now (before 1.0) is the time to break things.

src/tbp/monty/frameworks/telemetry/schemas.py
tristanls-tbp: I don't know whether named schemas will be relevant for `TelemetryEvent`s. But I think they are relevant for snapshots, since we probably don't want to introspect a snapshot and decide what to do based on each field's contents. A blob of bytes can be anything. I think, instead, we want to just do the right thing for the right snapshot, hence a well-known field like `kind`.

src/tbp/monty/telemetry/publishers.py
tristanls-tbp: What is the need for type hints for consumers of the telemetry stream?

src/tbp/monty/frameworks/experiments/monty_experiment.py
tristanls-tbp: I think my main point here is that `telemeter` should look like `logger` from the outside. This is why we are violating the naming convention so that `getTelemeter` matches `getLogger`. This is to strongly suggest that the user interact with a `telemeter` the same way they would with a `logger`. Why? Because everyone already knows how to use a `logger`.

tests/integration/frameworks/telemetry/telemetry_test.py
tristanls-tbp: > I don't want to add a `setLevel` to the publisher, 

Publisher API should match the `logger` API. Even if you find a way not to add `setLevel`, for any interaction you do with `publisher.event_loger.something` I will come back and point out that the publisher is missing `publisher.something` API.

src/tbp/monty/frameworks/experiments/monty_experiment.py
AgentRev: I guess there are 2 paths:

- Telemetry is a pub/sub framework with a restricted `logging`-like API, or
- Telemetry is a set of `Logger`- and `Handler`-derived classes with extra features.

It's now clear the second one better matches your expectations. Instead of making `TelemetryPublisher` redundantly rewrap existing `Logger` methods, directly making it a `Logger` subclass would avoid the issue altogether. What led me to take the first path up until now is snapshots; more specifically, having two separate loggers under a single abstract roof. It's an implementation detail that diverges from `Logger`, which could confuse users wanting to add snapshot handlers for example. If we instead go all-in on the second path, I think explicitly providing access the snapshot logger is warranted.

Would it be okay if snapshots also have their own publisher class once introduced? Via something like `snapshotter = telemetry.getSnapshotter(...)` and `snapshotter.emit()` or along those lines. That way, the telemeter and the snapshotter would both be independent `Logger` instances with all the parent methods available, and `telemetry.snapshot(...)` could actually still internally target the other logger.

src/tbp/monty/telemetry/publishers.py
AgentRev: Mostly to make Cmd+Space correctly list all the fields of the derived event, and have all associated types resolved and validated by the IDE. This only applies to derived events, like [this one](https://github.com/AgentRev/tbp.monty/blob/fe55e1a9407d093624fc9be6a24007d66938e9f0/src/tbp/monty/frameworks/telemetry/episode/schemas.py#L95-L141), and provides no benefit for generic `TelemetryEvent` schemas. However, usage of generic events should probably be minimized, to instead prioritize derived events, as they are easier to track and refactor than surprise bags of key-value pairs.

src/tbp/monty/frameworks/experiments/monty_experiment.py
tristanls-tbp: I think your characterization of the two paths and my preference are spot on.

The way I've been thinking about snapshots and the general implementation details of which method maps to which inner logger has been settled in my mind, but perhaps I missed something. I imagined everything going to the one `f"telemetry.{name}"` logger, except `.snapshot(...)` method, which would go to `f"snapshots.{name}"` logger. This way, everything would map straightforwardly, except for the snapshot route, which was the primary reason for the wrapper class to exist.

However, your idea of `snapshotter = telemetry.getSnapshotter(...)` sounds fine to me, especially if it's easier.

If 
```python
logger = getLogger(__name__)
telemeter = telemetry.getTelemeter(__name__)
snapshotter = telemetry.getSnapshotter(__name__)
```
ever becomes too much boilerplate, we can always fix up the user interface later to make it more convenient.

AgentRev: My comment was a bit oddly written, I've edited it for clarity. `.snapshot(...)` can still exist.

tristanls-tbp: I agree. That's what I meant when I said we can always fix up the user interface later.

It can always end up something like:

```python
class Telemeter(...):

    _snapshotter: Snapshotter

    def __init__(self, name) -> ...:
       self._snapshotter = telemetry.getSnapshotter(f"snapshots.{name}")
       ...
   
    def snapshot(self, ...) -> ...:
        self._snapshotter.debug(...) # or whatever method is appropriate
```

src/tbp/monty/telemetry/factories.py
```diff
@@ -0,0 +1,46 @@
...
+            cls._logger_class = logger_class
+            logging.setLoggerClass(cls)
+
+    def __new__(cls, name, *args, **kwargs) -> logging.Logger:
+        """Instantiates and returns a new logger; subclass varies with prefix."""
+        # TODO telemetry: add "snapshots." here in the future
+        if name.startswith("telemetry."):
+            return TelemetryPublisher(name, *args, **kwargs)
+
+        return cls._logger_class(name, *args, **kwargs)
```
AgentRev: Switching the publisher to a `Logger` subclass introduced the problem that those cannot not be instantiated directly ([as per Python docs](https://docs.python.org/3/library/logging.html#logger-objects)). I created this factory class to address that. `setup()` here is called by `getTelemeter` before it calls `getLogger`. This ensures that the latter calls the `__new__` here when a new logger is needed, which returns one of `TelemetryPublisher` subtype when then name is prefixed by `telemetry.`. `getTelemeter` then casts it to a `TelemetryPublisher` via an `isinstance` check before returning it. It's a bit of a strange mechanic, but it appears to work quite well. Let me know if you have any concern with this approach.

src/tbp/monty/telemetry/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+    from tbp.monty import telemetry
+    from tbp.monty.telemetry.schemas import TelemetryEvent
+
+    telemeter = telemetry.getTelemeter(__name__)
+    telemeter.info(TelemetryEvent(kind="CustomEvent", values={"key": "value"}))
+"""
+
+from tbp.monty.telemetry._api import getTelemeter
+
+__all__ = ["getTelemeter"]
```
AgentRev: I moved `getTelemeter` to an `_api.py`, since having it contained in `__init__.py` was messing up import suggestions of some telemetry classes in IDEs. I've seen this pattern in some other projects ([example](https://github.com/pytorch/pytorch/tree/main/torch/distributed/tensor)). Let me know if you'd like it done differently.

src/tbp/monty/telemetry/publishers.py
AgentRev: I've left `telemetry_schema` as is for now since we didn't reach a conclusion, let me know if that's okay.

src/tbp/monty/telemetry/factories.py
AgentRev, line 46: Switching the publisher to a `Logger` subclass introduced the problem that those cannot not be instantiated directly ([as per Python docs](https://docs.python.org/3/library/logging.html#logger-objects)). I created this factory class to address that. `setup()` here is called by `getTelemeter` before it calls `getLogger`. This ensures that the latter calls the `__new__` here when a new logger is needed, which returns one of `TelemetryPublisher` subtype when then name is prefixed by `telemetry.`. `getTelemeter` then casts it to a `TelemetryPublisher` via an `isinstance` check before returning it. It's a bit of a strange mechanic, but it appears to work quite well. Let me know if you have any concern with this approach.

src/tbp/monty/telemetry/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+    from tbp.monty.telemetry.schemas import TelemetryEvent
+
+    telemeter = telemetry.getTelemeter(__name__)
+    telemeter.info(TelemetryEvent(kind="CustomEvent", your_key="your_value", ...))
+    telemeter.debug(TelemetryEvent(kind="DebugEvent", ...))
+"""
+
+from tbp.monty.telemetry._api import getTelemeter
+
+__all__ = ["getTelemeter"]
```
AgentRev, line 48: I moved `getTelemeter` to an `_api.py`, since having it contained in `__init__.py` was messing up import suggestions of some telemetry classes in IDEs. I've seen this pattern in some other projects ([example](https://github.com/pytorch/pytorch/tree/main/torch/distributed/tensor)). Could also be named `helpers.py`. Let me know if you'd like it done differently.

src/tbp/monty/telemetry/publishers.py
AgentRev: I left `telemetry_schema` as is for now since we didn't reach a conclusion yet.

src/tbp/monty/telemetry/telemeter.py
tristanls-tbp, line 1: If this can't be in `__init__.py`, then please rename to `telemeter.py`. This way, we don't add another convention to follow.

src/tbp/monty/telemetry/_api.py
```diff
@@ -0,0 +1,51 @@
...
+
+from __future__ import annotations
+
+import logging
+
+from tbp.monty.telemetry.factories import _TelemetryLoggerFactory
+from tbp.monty.telemetry.publishers import TelemetryPublisher
+
+
+def getTelemeter(  # noqa: N802 - needed to mirror getLogger camelCase
```
tristanls-tbp: **issue**: Please leave it as `# noqa: N802` without extra words. I appreciate the desire for clarity, but this now marks the beginning of a new convention in which an author must decide whether to include comments next to directives. An easy way to avoid adding that cognitive load is to not include any comments here. Otherwise, do I say `- because I wanted to have a descriptive test name` on every `# noqa: E501`? I feel the rationale is already better explained and demonstrated in the docstring. Also, my IDE does this when I hover over `N802`:

Image

src/tbp/monty/telemetry/__init__.py
tristanls-tbp: **note**: I'm curious about how the import suggestions were messed up. But, I prefer not to have things in `__init__.py`, so 👍.

src/tbp/monty/telemetry/factories.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @classmethod
+    def setup(cls):
+        """Ensures this factory is designated as the main logger class."""
+        logger_class = logging.getLoggerClass()
+        if not issubclass(logger_class, cls):
+            cls._logger_class = logger_class
+            logging.setLoggerClass(cls)
+
+    def __new__(cls, name, *args, **kwargs) -> logging.Logger:
```
tristanls-tbp, line 40: **issue**: Maybe I'm reading the wrong documentation, but `logging.setLoggerClass()` docs that I read specifically say to implement `__init__` method:

> Set the class to be used when instantiating a logger. The class should define `__init__()` such that only a name argument is required, and the `__init__()` should call `Logger.__init__()`

tristanls-tbp (commented):
Thank you for the updates @AgentRev. I haven't had a chance to review everything yet, but submitting a review-in-progress in case it's useful to keep things going.

src/tbp/monty/telemetry/__init__.py
AgentRev: If I wrote `from tbp.monty.telemetry.publishers import TelemetryPublisher` in this `__init__.py`, and then I tried using `TelemetryPublisher` in another file without properly importing it, PyCharm suggested adding `from tbp.monty.telemetry import TelemetryPublisher` (missing the `.publishers`), which is incorrect.

After removing that import from `__init__.py`, PyCharm's suggestion changed to the expected `from tbp.monty.telemetry.publishers import TelemetryPublisher`. I'm not entirely sure why; it threw me for a loop a little bit.

src/tbp/monty/telemetry/factories.py
AgentRev: I assume they used that wording for reasons of simplicity. Under the hood inside `logging`, loggers are instantiated via `rv = (self.loggerClass or _loggerClass)(name)`. It doesn't explicitly rely on an `__init__()`.

This is what I'm leaning on:

>When you call a class like `obj = MyClass(x)`, Python performs two sequential steps under the hood:
>
>   1. `__new__` runs first: It takes the class `cls` as an argument, allocates memory, and returns the blank instance.
>   2. `__init__` runs second: If (and only if) `__new__` returns an instance of that class, Python passes that instance into `__init__` as self, along with any other arguments.
>
> `__init__` is only triggered automatically if the object returned by `__new__` is an instance of the class that was originally called (or a subclass of it).

In this case, we have `class _TelemetryLoggerFactory(logging.Logger)` (i.e. `class B(A)`), with its `__new__` returning an initialized instance of either `logging.Logger` (i.e. `class A`) or `TelemetryPublisher` (i.e. `class C(A)`).

Since `A` and `C(A)` are neither instances nor subclasses of `B(A)`, the conditions are not met for the second step to occur, and thus Python will not attempt to call `B.__init__()`.

So, at the end of the day, once `_TelemetryLoggerFactory` is assigned via `setLoggerClass`, the instance that comes out of `rv = (self.loggerClass or _loggerClass)(name)` will always be a vanilla `logging.Logger` or `TelemetryPublisher`, never a `_TelemetryLoggerFactory` itself.

Long story short, `__new__` is kind of an "override" for `__init__`. Using it this way is a niche but valid mechanic.

tristanls-tbp: It doesn't seem I need a factory at all if `getTelemeter` implementation swaps out the logger class momentarily:
```python
    if not name.startswith("telemetry."):
        name = f"telemetry.{name}"

    logger_class = logging.getLoggerClass()
    logging.setLoggerClass(TelemetryPublisher)
    telemeter = logging.getLogger(name)
    logging.setLoggerClass(logger_class)

    if not isinstance(telemeter, TelemetryPublisher):
        raise TypeError(
            f"Logger factory failed to return a telemetry publisher for '{name}'"
        )

    return telemeter
```

AgentRev: While that was my initial version, ultimately I realized it is not thread-safe. If you have a thread A calling `getTelemeter` getting preempted right after `setLoggerClass(TelemetryPublisher)`, and thread B calls `getLogger` at this instant, it will incorrectly receive a telemeter. Or, B calling `getLoggerClass` after A's first `setLoggerClass`, then A calls its second `setLoggerClass`, followed by B, causing a loss of the original logger class.

The odds are very low, but non-zero. I guess it boils down to whether or not such edge cases are acceptable.

Upon investigation after your comment however, I noticed it is possible to avoid those edge cases by protecting the class juggling with `logging._lock`, although this is technically a private variable.

    with logging._lock:  # threading.RLock
        logger_class = logging.getLoggerClass()
        try:
            logging.setLoggerClass(TelemetryPublisher)
            telemeter = logging.getLogger(name)
        finally:
            logging.setLoggerClass(logger_class)

Would this be fine with you?

tristanls-tbp: Good call on thread safety. I'm spoiled by Rust telling me when I'm wrong. 

With thread safety in mind, I now appreciate the factory method. 

What do you think about having this in `__init__.py`:

```python
import logging

from tbp.monty.telemetry.publishers import TelemetryPublisher as _TelemetryPublisher
from tbp.monty.telemetry.telemeter import getTelemeter

__all__ = ["getTelemeter"]

class _TelemetryLoggerFactory(logging.Logger):
    def __new__(cls, name, *args, **kwargs) -> logging.Logger:  # type: ignore[misc]
        if name.startswith("telemetry."):
            return _TelemetryPublisher(name, *args, **kwargs)
        return logging.getLoggerClass()(name, *args, **kwargs)


logging.Logger.manager.setLoggerClass(_TelemetryLoggerFactory)
```

This is your factory, but it is installed on the `Logger.manager` on import of telemetry (hence `__init__.py`). To avoid IDE confusion I did `from blah import X as _X` pattern, does that solve the IDE import suggestion problem? 

Having it in the `Logger.manager` makes sure that `__new__` runs inside the critical section:

```python
class Manager(object):
    def getLogger(self, name):
        """
        Get a logger with the specified name (channel name), creating it
        if it doesn't yet exist. This name is a dot-separated hierarchical
        name, such as "a", "a.b", "a.b.c" or similar.

        If a PlaceHolder existed for the specified name [i.e. the logger
        didn't exist but a child of it did], replace it with the created
        logger and fix up the parent/child references which pointed to the
        placeholder to now point to the logger.
        """
        rv = None
        if not isinstance(name, str):
            raise TypeError('A logger name must be a string')
        _acquireLock()
        try:
            if name in self.loggerDict:
                rv = self.loggerDict[name]
                if isinstance(rv, PlaceHolder):
                    ph = rv
                    rv = (self.loggerClass or _loggerClass)(name)
                    rv.manager = self
                    self.loggerDict[name] = rv
                    self._fixupChildren(ph, rv)
                    self._fixupParents(rv)
            else:
                rv = (self.loggerClass or _loggerClass)(name)
                rv.manager = self
                self.loggerDict[name] = rv
                self._fixupParents(rv)
        finally:
            _releaseLock()
        return rv
```

AgentRev: Almost perfect, just an _itsy-bitsy_ more:

```py
import logging
from typing import ClassVar

from tbp.monty.telemetry.publishers import TelemetryPublisher as _TelemetryPublisher
from tbp.monty.telemetry.telemeter import getTelemeter

__all__ = ["getTelemeter"]


class _TelemetryLoggerFactory(logging.Logger):
    _manager_logger_class: ClassVar = logging.Logger.manager.loggerClass

    def __new__(cls, name, *args, **kwargs) -> logging.Logger:  # type: ignore[misc]
        if name.startswith("telemetry."):
            return _TelemetryPublisher(name, *args, **kwargs)

        logger_class = cls._manager_logger_class or logging.getLoggerClass()
        return logger_class(name, *args, **kwargs)


logging.Logger.manager.setLoggerClass(_TelemetryLoggerFactory)
```

This ensures the preservation of a pre-existing `logging.Logger.manager.loggerClass` 🙂

tristanls-tbp: 👍 

Feels like we need some docstrings and a comment on `logging.Logger.manager.setLoggerClass(_TelemetryLoggerFactory)`. We are being very clever and will wonder what is going on here in the future.

**suggestion**: Suggested doctring for `_TelemetryLoggerFactory`:

```python
      """Routes ``telemetry.*`` names to become ``TelemetryPublisher``s.

      ``logging`` notes that:

      > Loggers should _NEVER_ be instantiated directly, but always through the
      > module-level function ``logging.getLogger(name)``.

      ``TelemetryPublisher`` is a subclass of ``logging.Logger``, so it cannot be
      instantiated directly either. This is why ``TelemetryPublisher`` is not instantiated
      in ``getTelemeter`` and why this factory exists.

      ``logging`` has the ``loggerClass`` hook it invokes to create a logger it
      has not seen before, in `logging.Manager.getLogger`::

          rv = (self.loggerClass or _loggerClass)(name)

      This class is installed as that hook and dispatches on the name prefix.
      ``telemetry.*`` gets a `TelemetryPublisher`, everything else gets whatever logger
      class the process would otherwise have used.

      Despite subclassing ``logging.Logger``, this class is never itself instantiated.
      Python calls ``__init__`` only when ``__new__`` returns an instance of the class
      being constructed. This ``__new__`` always returns something else. This routing
      mechanism depends on this behavior, hence the ``# type: ignore[misc]`` on the
      ``__new__`` method.

      Additionally, this factory is installed on the ``Logger.manager`` rather
      than on the module-level ``logging`` global. This is because ``logging`` allows
      libraries to extend the logger class by subclassing the current logger class and
      calling ``logging.setLoggerClass``::

          class LibLogger(logging.getLoggerClass()): ...
          logging.setLoggerClass(LibLogger)

      If this factory was installed globally, it would become the base class of the next
      library to do that, and its ``__new__`` would prevent the library's subclass from
      being instantiated. Installing it on the manager avoids that problem. The manager
      logger class is never returned by ``logging.getLoggerClass()``, so it cannot be
      inherited by accident, and the global logger class remains free for its intended
      purpose. ``Manager.getLogger`` consults the manager logger class first, so
      telemetry names are still always caught.

      Another reason for placing this factory on the manager is thread-safety.
      ``Manager.getLogger`` invokes the factory while holding the ``logging`` module
      lock, so the dispatch happens inside the same critical section that guards the
      logger dictionary. If we did not do this, setting the logger class around a
      ``getLogger`` call and restoring it afterwards would instead leave a window
      in which another thread creating an unrelated logger could receive a
      ``TelemetryPublisher``.

      The non-telemetry branch resolves its class on every call, so a logger class
      installed after this module is imported is still honored.
      ``_manager_logger_class`` preserves any manager-level class installed
      before it, which ``logging`` would have given precedence over the global.
      """
```

**suggestion**: Suggested comment above `logging.Logger.manager.setLoggerClass(_TelemetryLoggerFactory)` code line:

```python
# See the _TelemetryLoggerFactory docstring for why we install on Logger.manager
# instead of logging itself.
logging.Logger.manager.setLoggerClass(_TelemetryLoggerFactory)
```

AgentRev: How about slightly less verbose:

    """Routes ``telemetry.*`` names to ``TelemetryPublisher``-typed loggers.

    ``logging`` notes that "Loggers should _NEVER_ be instantiated directly, but always
    through the module-level function ``logging.getLogger(name)``."

    Installed as ``Logger.manager.loggerClass``, this class is invoked by
    ``Manager.getLogger`` to construct loggers it hasn't seen: ``rv = (self.loggerClass
    or _loggerClass)(name)``. Names not prefixed with ``telemetry.*`` are passed through
    to whatever logger class the process would otherwise have used.

    A factory is needed here because ``TelemetryPublisher`` subclasses
    ``logging.Logger`` and so, per ``logging``'s rules, must never be instantiated
    directly. This factory class is itself never instantiated either: its ``__init__``
    never runs, as its ``__new__`` always returns another object (hence the ``# type:
    ignore[misc]``).

    It is installed on the manager rather than globally via ``logging.setLoggerClass``
    so it can't be picked up as a base class by libraries that do ``class
    LibLogger(logging.getLoggerClass()): ...``, which would block their subclass from
    being instantiated. It is decoupled from ``logging.getLoggerClass()``, which avoids
    the possibility of accidental inheritance, while still being checked first by
    ``Manager.getLogger``, so telemetry names are still always caught.

    This placement also gives thread-safety for free: ``Manager.getLogger`` calls the
    factory while holding the logging lock, so dispatch happens atomically with the
    logger dictionary, unlike swapping the global logger class around a ``getLogger()``
    call, which would race with other threads.

    The non-telemetry branch resolves the fallback class on every call (so a class
    installed later is honored), falling back to ``_manager_logger_class`` to preserve
    any manager-level class set before this one.
    """

Also note that `>` blocks are not part of Google-style docstrings, so their rendering is hit-or-miss.

tristanls-tbp: Thank you, yes, that's a better docstring 🙂.

src/tbp/monty/telemetry/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+
+    The non-telemetry branch resolves the fallback class on every call (so a class
+    installed later is honored), falling back to ``_manager_logger_class`` to preserve
+    any manager-level class set before this one.
+    """
+
+    _manager_logger_class: ClassVar = logging.Logger.manager.loggerClass
+
+    def __new__(cls, name, *args, **kwargs) -> logging.Logger:  # type: ignore[misc]
+        # TODO telemetry: add "snapshots." here in the future
```
tristanls-tbp, line 92: **issue**: Please don't namespace the TODO. In the history of this codebase, none of the TODO annotation conventions survived 🙂. e.g., `TODO M:`, `TODO S:`, `TODO H:`, `TODO(username):`, etc.

src/tbp/monty/frameworks/experiments/monty_experiment.py
```diff
@@ -52,6 +53,7 @@
 __all__ = ["MontyExperiment"]
 
 logger = logging.getLogger("tbp.monty")
+telemeter = telemetry.getTelemeter("tbp.monty")
```
tristanls-tbp, line 56: **issue**: Unused.

src/tbp/monty/telemetry/publishers.py
```diff
@@ -0,0 +1,114 @@
...
+    def __init__(self, *args, **kwargs):
+        """Initializes the logger; do not instantiate this class outside its module."""
+        super().__init__(*args, **kwargs)
+        self.propagate = False  # do not propagate to root logger
+
+        # Prevent logging.lastResort from printing to stderr
+        if not self.hasHandlers():
+            self.addHandler(logging.NullHandler())
+
+    def log(self, level: int, event: TelemetryEvent, *args, **kwargs):
```
tristanls-tbp, line 43: **issue**: Incorrect method signature here and in the rest of the methods.

**suggestion**:

```suggestion
    def log(self, level: int, msg: object, *args, **kwargs):
```

If we want a type hint, I think we'd need a new method:
```python
    def emit(self, level: int, event: TelemetryEvent, *args, **kwargs):
        """Emits a structured telemetry event at the specified log level.

        Args:
            level: The log level.
            event: The event instance.
            *args: Passed forward to ``Logger.log`` method.
            **kwargs: Passed forward to ``Logger.log`` method.
        """
        self.log(level, event, *args, **kwargs)
```

src/tbp/monty/telemetry/publishers.py
```diff
@@ -0,0 +1,114 @@
...
+
+    @deprecated("Unsupported")
+    def critical(self, *args) -> NoReturn:
+        """Unsupported; ``CRITICAL`` level is reserved for telemetry silencing."""
+        raise NotImplementedError
+
+    @deprecated("Unsupported")
+    def fatal(self, *args) -> NoReturn:
+        """Unsupported; ``CRITICAL`` level is reserved for telemetry silencing."""
+        raise NotImplementedError
```
tristanls-tbp, line 114: **issue**: This doesn't look right to me. If someone explicitly outputs critical or fatal telemetry, there is no reason to prevent them from doing so. I'm guessing this comes from an example that I gave where I showed that setting the log level to `CRITICAL` would turn things off. This is more of a snapshot thing. Snapshots SHALL only be emitted at `DEBUG` level, so any non-`DEBUG` snapshot log level will, in effect, turn them off. But `CRITICAL` is not intended as a special sentinel value.

src/tbp/monty/telemetry/publishers.py
```diff
@@ -0,0 +1,114 @@
...
+            event: The event instance.
+            *args: Passed forward to ``Logger.log`` method.
+            **kwargs: Passed forward to ``Logger.log`` method.
+        """
+        self.log(logging.ERROR, event, *args, **kwargs)
+
+    @deprecated("Unsupported")
+    def exception(self, *args) -> NoReturn:
+        """Unsupported; telemetry does not handle exception logging."""
+        raise NotImplementedError
```
tristanls-tbp, line 104: **question**: Why not?

src/tbp/monty/conf/telemetry/none.yaml
tristanls-tbp, line 1: **issue**: In another comment, I explained that `CRITICAL` is not intended as a special sentinel value. 

**suggestion**: Rename to `critical.yaml`.

tristanls-tbp (commented):
Thanks! Looking very close now 🙂. 

I added some inline comments, but also noticed that there's no default telemetry handler and formatter.

This is a draft only, but take a look at this patch file to see what I mean.

[edit: initial patch had `handlers.py`, but that was accidental and since removed]

```patch
diff --git a/src/tbp/monty/conf/experiment/base_config_10distinctobj_dist_agent.yaml b/src/tbp/monty/conf/experiment/base_config_10distinctobj_dist_agent.yaml
index 55d66728..9fb9f622 100644
--- a/src/tbp/monty/conf/experiment/base_config_10distinctobj_dist_agent.yaml
+++ b/src/tbp/monty/conf/experiment/base_config_10distinctobj_dist_agent.yaml
@@ -12,6 +12,7 @@ defaults:
   - /env_interface/transform: missing_depthto3d_sensor2_semantic0
   - /termination: any1_objectrecognition_t1000_e500_tot6000
   - /logging: basic_warning_wandb_monty_runs
+  - /telemetry: info
 
 experiment:
   _target_: tbp.monty.frameworks.experiments.object_recognition_experiments.MontyObjectRecognitionExperiment
diff --git a/src/tbp/monty/conf/telemetry/info.yaml b/src/tbp/monty/conf/telemetry/info.yaml
index f4747eaf..2af5bcbe 100644
--- a/src/tbp/monty/conf/telemetry/info.yaml
+++ b/src/tbp/monty/conf/telemetry/info.yaml
@@ -1,5 +1,16 @@
 # @package experiment.config.telemetry
 
+formatters:
+  telemetry_formatter:
+    class: tbp.monty.telemetry.formatters.JsonFormatter
+handlers:
+  telemetry_console:
+    class: logging.StreamHandler
+    formatter: telemetry_formatter
+    level: INFO
+    stream: ext://sys.stdout
 loggers:
   telemetry.tbp.monty:
     level: INFO
+    handlers:
+      - telemetry_console
diff --git a/src/tbp/monty/frameworks/experiments/monty_experiment.py b/src/tbp/monty/frameworks/experiments/monty_experiment.py
index 6fdb3a0a..60e067e4 100644
--- a/src/tbp/monty/frameworks/experiments/monty_experiment.py
+++ b/src/tbp/monty/frameworks/experiments/monty_experiment.py
@@ -14,12 +14,12 @@ import logging
 import logging.config
 import pprint
 from pathlib import Path
-from typing import Any, Literal
+from typing import Any, Literal, cast
 
 import hydra
 import numpy as np
 import torch
-from omegaconf import DictConfig
+from omegaconf import DictConfig, OmegaConf
 from typing_extensions import Self
 
 from tbp.monty import telemetry
@@ -272,6 +272,13 @@ class MontyExperiment:
             logging_config: Logging configuration.
             telemetry_config: Telemetry configuration.
         """
+        if telemetry_config:
+            telemetry_config["version"] = 1
+            telemetry_config["disable_existing_loggers"] = False
+            logging.config.dictConfig(
+                cast("dict", OmegaConf.to_container(telemetry_config, resolve=True))
+            )
+
         # Unpack individual logging arguments
         self.python_log_level = logging_config["python_log_level"]
         self.log_to_file = logging_config["python_log_to_file"]
@@ -321,7 +328,6 @@ class MontyExperiment:
             "incremental": True,
         }
         final_config.update(logging_config)
-        final_config.update(telemetry_config)
         logging.config.dictConfig(final_config)
 
     def init_monty_data_loggers(self, logging_config: dict[str, Any]) -> None:
diff --git a/src/tbp/monty/telemetry/formatters.py b/src/tbp/monty/telemetry/formatters.py
new file mode 100644
index 00000000..0a0dc4ce
--- /dev/null
+++ b/src/tbp/monty/telemetry/formatters.py
@@ -0,0 +1,28 @@
+# Copyright 2026 Thousand Brains Project
+#
+#
+
+import json
+import logging
+from typing import cast
+
+from tbp.monty.telemetry.schemas import TelemetryEvent
+
+
+class JsonFormatter(logging.Formatter):
+    def format(self, record: logging.LogRecord) -> str:
+        event = cast("TelemetryEvent", record.__dict__["telemetry_schema"])
+        return json.dumps(
+            {
+                "level": record.levelname,
+                "name": record.name,
+                "funcName": record.funcName,
+                "lineno": record.lineno,
+                **event.model_dump(mode="json"),
+            }
+        )
diff --git a/src/tbp/monty/telemetry/publishers.py b/src/tbp/monty/telemetry/publishers.py
index 451559d2..a57e82a6 100644
--- a/src/tbp/monty/telemetry/publishers.py
+++ b/src/tbp/monty/telemetry/publishers.py
@@ -40,25 +40,25 @@ class TelemetryPublisher(logging.Logger):
         if not self.hasHandlers():
             self.addHandler(logging.NullHandler())
 
-    def log(self, level: int, event: TelemetryEvent, *args, **kwargs):
+    def log(self, level: int, msg: object, *args, **kwargs):
         """Emits a structured telemetry event at the specified log level.
 
         Args:
             level: The log level.
-            event: The event instance.
+            msg: The event instance.
             *args: Passed forward to ``Logger.log`` method.
             **kwargs: Passed forward to ``Logger.log`` method.
 
         Raises:
-            TypeError: If the event is not a `TelemetryEvent`.
+            TypeError: If the ``msg`` is not a `TelemetryEvent`.
         """
-        if not isinstance(event, TelemetryEvent):
+        if not isinstance(msg, TelemetryEvent):
             raise TypeError("This logger is only for telemetry events")
 
-        kwargs.setdefault("extra", {})["telemetry_schema"] = event
-        super().log(level, event.kind, *args, **kwargs)
+        kwargs.setdefault("extra", {})["telemetry_schema"] = msg
+        super().log(level, msg.kind, *args, **kwargs)
 
-    def debug(self, event: TelemetryEvent, *args, **kwargs):
+    def debug(self, msg: object, *args, **kwargs):
         """Emits a structured telemetry event at ``DEBUG`` log level.
 
         Args:
@@ -66,7 +66,8 @@ class TelemetryPublisher(logging.Logger):
             *args: Passed forward to ``Logger.log`` method.
             **kwargs: Passed forward to ``Logger.log`` method.
         """
-        self.log(logging.DEBUG, event, *args, **kwargs)
+        kwargs.setdefault("stacklevel", 3)  # skip 2 frames to get to the caller
+        self.log(logging.DEBUG, msg, *args, **kwargs)
 
     def info(self, event: TelemetryEvent, *args, **kwargs):
         """Emits a structured telemetry event at ``INFO`` log level.
@@ -76,6 +77,7 @@ class TelemetryPublisher(logging.Logger):
             *args: Passed forward to ``Logger.log`` method.
             **kwargs: Passed forward to ``Logger.log`` method.
         """
+        kwargs.setdefault("stacklevel", 3)  # skip 2 frames to get to the caller
         self.log(logging.INFO, event, *args, **kwargs)
 
     def warning(self, event: TelemetryEvent, *args, **kwargs):
@@ -86,6 +88,7 @@ class TelemetryPublisher(logging.Logger):
             *args: Passed forward to ``Logger.log`` method.
             **kwargs: Passed forward to ``Logger.log`` method.
         """
+        kwargs.setdefault("stacklevel", 3)  # skip 2 frames to get to the caller
         self.log(logging.WARNING, event, *args, **kwargs)
 
     def error(self, event: TelemetryEvent, *args, **kwargs):
@@ -96,10 +99,11 @@ class TelemetryPublisher(logging.Logger):
             *args: Passed forward to ``Logger.log`` method.
             **kwargs: Passed forward to ``Logger.log`` method.
         """
+        kwargs.setdefault("stacklevel", 3)  # skip 2 frames to get to the caller
         self.log(logging.ERROR, event, *args, **kwargs)
 
     @deprecated("Unsupported")
-    def exception(self, *args) -> NoReturn:
+    def exception(self, msg: object, *args, **kwargs) -> NoReturn:
         """Unsupported; telemetry does not handle exception logging."""
         raise NotImplementedError
 ```

To see the telemetry output with the patch above I've been running
```
WANDB_MODE=disabled python run_parallel.py experiment=base_config_10distinctobj_dist_agent
```

**note**: I'm still questioning the `"telemetry_schema"` nesting. With `msg` being `kind`, the nesting seems unnecessary. Although it did make it easy to write the `JsonFormatter`.

File: src/tbp/monty/frameworks/experiments/monty_experiment.py (hunk 16 above)
Comment (AgentRev) [reply src/tbp/monty/frameworks/experiments/monty_experiment.py]: Right, I forgot that one

File: src/tbp/monty/telemetry/__init__.py (hunk 15 above)
Comment (AgentRev) [reply src/tbp/monty/telemetry/__init__.py]: Yeah it doesn't belong here,  my intent with the "telemetry" keyword was to track related TODOs outside of the module, but here we are actually inside.

Although, there is value in having the username in TODOs, at my current workplace we have a strict `TODO [username]: [description] [Jira issue number]` policy, but that wouldn't make sense for a non-maintainer.

File: src/tbp/monty/telemetry/publishers.py (hunk 17 above)
Comment (AgentRev) [reply src/tbp/monty/telemetry/publishers.py]: The signatures I put did make me frown a bit. However, is it better to have "correct" `debug`, `info`, etc. signatures that don't show any IDE warning, just for the sake of making them look superficially correct, or put a `TelemetryEvent` type hint with probably `# type: ignore[override]` to strongly signal that it expects only that type? Since we are purposefully violating Liskov principle by raising a `TypeError` for all of those methods, we might as well be as explicit about it as possible. I've changed it to what you suggested for now, but let me know what you think.

**Comment by AgentRev** [comment]
> I added some inline comments, but also noticed that there's no default telemetry handler and formatter.
> 
> This is a draft only, but take a look at this patch file to see what I mean.

Would you want those sections to be added to all of the other level YAMLs? Maybe moved to their own YAML and have them imported via `defaults:`?

> **note**: I'm still questioning the `"telemetry_schema"` nesting. With `msg` being `kind`, the nesting seems unnecessary. Although it did make it easy to write the `JsonFormatter`.

I found a very simple solution; the point of `msg` being of `object` type is so that we can pass whatever we want to it. So, I made the schema the `msg` itself, adding to it an `__str__` method that returns `kind`. This eliminates any need to juggle with `extra` and `"telemetry_schema"`, also allowing the `msg` parameter to remain identical across the whole call chain.

File: src/tbp/monty/telemetry/__init__.py (hunk 15 above)
Comment (tristanls-tbp) [reply src/tbp/monty/telemetry/__init__.py]: I tried adding `username` for a bit, but Jeremy pointed out that git blame tells you who added it. So, that was that.

File: src/tbp/monty/telemetry/publishers.py [hunk 20]
```diff
@@ -0,0 +1,192 @@
...
+
+from __future__ import annotations
+
+import logging
+
+from typing_extensions import deprecated, override
+
+from tbp.monty.telemetry.schemas import TelemetryEvent
+
+# ruff: noqa: DOC502
```
tristanls-tbp: **issue**: Looks like a leftover.

src/tbp/monty/telemetry/publishers.py
```diff
@@ -0,0 +1,192 @@
...
+            msg: The `TelemetryEvent` instance.
+            *args: Passed forward to ``Logger.log`` method.
+            **kwargs: Passed forward to ``Logger.log`` method.
+
+        Raises:
+            TypeError: If ``msg`` is not a `TelemetryEvent`.
+        """
+        self._log_event(level, msg, *args, **kwargs)
+
+    # Type-hinted equivalent of `log` for convenience
```
tristanls-tbp: **suggestion**: This looks like it belongs in the docstring 🙂.

src/tbp/monty/telemetry/publishers.py
```diff
@@ -0,0 +1,192 @@
...
+        """Emits a telemetry event at the specified log level.
+
+        Args:
+            level: The log level.
+            msg: The `TelemetryEvent` instance.
+            *args: Passed forward to ``Logger.log`` method.
+            **kwargs: Passed forward to ``Logger.log`` method.
+
+        Raises:
+            TypeError: If ``msg`` is not a `TelemetryEvent`.
```
tristanls-tbp: **issue**: Per DOC502, we don't document raises that aren't explicitly raised, here and below.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+pytest.importorskip(
+    "habitat_sim",
+    reason="Habitat Sim optional dependency not installed.",
+)
+
+
+class TelemetryLogHandler(logging.Handler):
+    """Logging handler that collects log records for telemetry assertions."""
+
+    def __init__(self):
```
tristanls-tbp: **issue**: We use explicit `-> None` returns. Otherwise, type checking ignores the method.

**note**: We are not consistent with this, and I forget sometimes too. But please use explicit `-> None` return.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+        super().__init__()
+        self.records: list[logging.LogRecord] = []
+
+    def emit(self, record: logging.LogRecord):
```
tristanls-tbp: **issue**: We use explicit `-> None` returns. Otherwise, type checking ignores the method.

**note**: We are not consistent with this, and I forget sometimes too. But please use explicit `-> None` return.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+        self.records.append(record)
+
+
+class LoggerTestCase(unittest.TestCase):
+    def setUp(self):
```
tristanls-tbp: **issue**: We use explicit `-> None` returns. Otherwise, type checking ignores the method.

**note**: We are not consistent with this, and I forget sometimes too. But please use explicit `-> None` return.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+    reason="Habitat Sim optional dependency not installed.",
+)
+
+
+class TelemetryLogHandler(logging.Handler):
+    """Logging handler that collects log records for telemetry assertions."""
+
+    def __init__(self):
+        super().__init__()
+        self.records: list[logging.LogRecord] = []
```
tristanls-tbp: **todo**: We settled on declaring instance properties at the class level, like below:

```suggestion
    records: list[logging.LogRecord]

    def __init__(self):
        super().__init__()
        self.records = []
```

It makes for more readable docstrings (when there are any): https://docs.thousandbrains.org/docs/code-style-guide#class-attribute-instance-property-and-constant-docstrings

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+class LoggerTestCase(unittest.TestCase):
+    def setUp(self):
+        # Clean up all loggers
+        logging.Logger.manager.loggerDict.clear()
+
+
+class TelemetryEventTest(LoggerTestCase):
+    """Unit tests for `TelemetryEvent`."""
+
+    def test_telemetry_event_defaults(self):
```
tristanls-tbp:
```suggestion
    def test_telemetry_event_defaults(self) -> None:
```

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.records: list[logging.LogRecord] = []
+
+    def emit(self, record: logging.LogRecord):
+        self.records.append(record)
+
+
+class LoggerTestCase(unittest.TestCase):
```
tristanls-tbp: **issue**: Per the style guide, please don't use inheritance for functionality reuse. It becomes unreadable quickly.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+    def setUp(self):
+        # Clean up all loggers
+        logging.Logger.manager.loggerDict.clear()
+
+
+class TelemetryEventTest(LoggerTestCase):
+    """Unit tests for `TelemetryEvent`."""
```
tristanls-tbp: **issue**: Extraneous docstring.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def test_telemetry_event_defaults(self):
+        """Verify `TelemetryEvent` kind fallback and values mapping."""
+        event = TelemetryEvent(kind="")
+        self.assertEqual(event.kind, event.__class__.__name__)
+
+    def test_telemetry_event_custom_fields(self):
+        """Verify `TelemetryEvent` kind when explicitly specified."""
```
tristanls-tbp: **issue**: The test name and the docstring seem to describe two different things.

**suggestion**:

```suggestion
    def test_custom_fields_override_kind(self) -> None:
```

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+    def setUp(self):
+        # Clean up all loggers
+        logging.Logger.manager.loggerDict.clear()
+
+
+class TelemetryEventTest(LoggerTestCase):
+    """Unit tests for `TelemetryEvent`."""
+
+    def test_telemetry_event_defaults(self):
+        """Verify `TelemetryEvent` kind fallback and values mapping."""
```
tristanls-tbp:
```suggestion
    def test_kind_defaults_to_class_name(self) -> None:
```

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+        self.records.append(record)
+
+
+class LoggerTestCase(unittest.TestCase):
+    def setUp(self):
+        # Clean up all loggers
+        logging.Logger.manager.loggerDict.clear()
+
+
+class TelemetryEventTest(LoggerTestCase):
```
tristanls-tbp: **note**: The test names don't need `telemetry_event` in their names, as we are in the `TelemetryEventTest`.

This way, we have `TelemetryEventTest::test_kind_defaults_to_class_name` instead of `TelemetryEventTest:test_telemetry_event_defaults`.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+        """Verify `TelemetryEvent` kind when explicitly specified."""
+        kind = "NewGraphAdded"
+        graph_id = "new_object0"
+        event = TelemetryEvent(kind=kind, graph_id=graph_id)
+        self.assertEqual(event.kind, kind)
+        self.assertEqual(event.graph_id, graph_id)
+
+
+class JsonFormatterTest(LoggerTestCase):
+    """Unit tests for `JsonFormatter`."""
```
tristanls-tbp: **issue**: Extraneous docstring.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+            event = TelemetryEvent(kind="StreamEvent", graph_id="obj_1")
+            self.telemeter.info(event)
+            output = stream.getvalue()
+            data = json.loads(output)
+
+            self.assertEqual(data["level"], "INFO")
+            self.assertEqual(data["kind"], "StreamEvent")
+            self.assertEqual(data["graph_id"], "obj_1")
+        finally:
+            self.telemeter.removeHandler(handler)
```
tristanls-tbp: **issue**: This feels like we're testing whether `logging` works, which we don't need to test. 

**note**: Also, without this test, `setUp` is not needed.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+class JsonFormatterTest(LoggerTestCase):
+    """Unit tests for `JsonFormatter`."""
+
+    def setUp(self):
+        super().setUp()
+        telemetry.getTelemeter("tbp.monty").setLevel(logging.NOTSET)
+        self.telemeter = telemetry.getTelemeter(monty_base.__name__)
+        self.telemeter.setLevel(logging.NOTSET)
+
+    def test_format(self):
```
tristanls-tbp: **suggestion (non-blocking)**: Recently, we turn to `hypothesis` to generate many test scenarios instead of hard-coding single example cases. But, I'm not blocking this PR on that as this is a minimal introduction of the telemetry concept.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+
+            self.assertEqual(data["level"], "INFO")
+            self.assertEqual(data["kind"], "StreamEvent")
+            self.assertEqual(data["graph_id"], "obj_1")
+        finally:
+            self.telemeter.removeHandler(handler)
+
+
+class TelemetryPublisherTest(LoggerTestCase):
+    """Unit tests for `TelemetryPublisher`."""
```
tristanls-tbp: **issue**: Extraneous docstring.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+        super().setUp()
+        telemetry.getTelemeter("tbp.monty").setLevel(logging.NOTSET)
+        self.handler = TelemetryLogHandler()
+        self.telemeter = telemetry.getTelemeter(monty_base.__name__)
+        self.telemeter.setLevel(logging.NOTSET)
+
+    def tearDown(self):
+        self.telemeter.removeHandler(self.handler)
+
+    def test_params(self):
```
tristanls-tbp: **issue**: This seems like a `getTelemeter` test, more than a `TelemetryPublisher` test.

**suggestion**:
```python
class GetTelemeterTest(unittest.TestCase):
    def test_returns_telemetry_publisher_with_telemetry_prefixed_name(self) -> None:
        ...
```

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+        finally:
+            self.telemeter.removeHandler(handler)
+
+
+class TelemetryPublisherTest(LoggerTestCase):
+    """Unit tests for `TelemetryPublisher`."""
+
+    def setUp(self):
+        super().setUp()
+        telemetry.getTelemeter("tbp.monty").setLevel(logging.NOTSET)
```
tristanls-tbp: **question**: Why is this needed?

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+        self.handler = TelemetryLogHandler()
+        self.telemeter = telemetry.getTelemeter(monty_base.__name__)
```
tristanls-tbp: **question**: What is special about `monty_base`? Why is this not a random string?

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+            ),
+        ]
+
+        for log_func, _, event in events:
+            log_func(event, stack_info=True)
+
+        self.assertEqual(len(self.handler.records), len(events))
+
+        for record, (_, level, event) in zip(self.handler.records, events):
+            print(record.stack_info)
```
tristanls-tbp: **todo**: Looks like a leftover.

tests/unit/telemetry/telemetry_test.py
tristanls-tbp: **suggestion (non-blocking)**: Also could test that the name doesn't get double-prefixed if it already starts with `telemetry.`.

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+        self.telemeter.removeHandler(self.handler)
+
+    def test_params(self):
+        """Verify logger params."""
+        self.assertIsInstance(self.telemeter, TelemetryPublisher)
+        self.assertEqual(self.telemeter.name, f"telemetry.{monty_base.__name__}")
+        self.assertFalse(self.telemeter.propagate)
+
+    def test_events(self):
+        """Verify emitting events produces log records with telemetry schemas."""
```
tristanls-tbp: **suggestion**:

```suggestion
    def test_logs_log_records_with_telemetry_event_as_the_message(self):
```

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+
+        for log_func, _, event in events:
+            log_func(event, stack_info=True)
+
+        self.assertEqual(len(self.handler.records), len(events))
+
+        for record, (_, level, event) in zip(self.handler.records, events):
+            print(record.stack_info)
+            self.assertEqual(record.levelno, level)
+            self.assertEqual(str(record.msg), event.kind)
```
tristanls-tbp: **issue**: This assertion looks like it belongs in the`TelemetryEvent` and not the publisher. Looks like it should be moved to a new `TelemetryEventTest::test_str_returns_event_kind`

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+from tbp.monty.hydra import instantiate_experiment
+from tbp.monty.telemetry.formatters import JsonFormatter
+from tbp.monty.telemetry.publishers import TelemetryPublisher
+from tbp.monty.telemetry.schemas import TelemetryEvent
+from tests import HYDRA_ROOT
+
+pytest.importorskip(
+    "habitat_sim",
+    reason="Habitat Sim optional dependency not installed.",
+)
```
tristanls-tbp: **question**: What here depends on habitat?

tests/unit/telemetry/telemetry_test.py
```diff
@@ -0,0 +1,222 @@
...
+            self.telemeter.addHandler(self.handler)
+            self.telemeter.error(TelemetryEvent(kind="TestEvent"))
+            self.assertEqual(len(self.handler.records), 1)
+
+    def test_critical_config(self):
+        """Verify behavior of Hydra config ``telemetry=critical``."""
+        with self._instantiate_experiment("critical"):
+            self.telemeter.addHandler(self.handler)
+            self.telemeter.critical(TelemetryEvent(kind="TestEvent"))
+            self.assertEqual(len(self.handler.records), 1)
```
tristanls-tbp: **issue**: These don't look like publisher tests, but some sort of configuration tests.

**question**: What is being tested here? `self.telemeter` is a `logging.Logger`. `self.handler` is a `logging.Handler`.

**note**: I'd rather not start introducing tests to every configuration package we create. That'd be too much for not enough return. If we mess up our YAML, we'll notice.

tristanls-tbp (commented):
Thank you. Looks great.

Few documentation notes on the main code.

Also, thank you for authoring the tests. This is the first time I've read through the tests, which is why there are many comments, mostly to match our conventions and inquiries.

PR #1141, opened 2026-08-18
chore: Remove redundant additions and subtractions

nielsleadholm:
While reading the `apply_rf_transform_to_points` method during the hackathon, I noticed that we have some redundant operations. Substituting in variables, you can see that the original lines are equivalent to:

```
transformed_locations = location_offset + ref_point2
= rotated_locs + ref_point_offset + ref_point2
= rotated_locs + (ref_point1 - ref_point2) + ref_point2
```

The last two terms cancel out, and it can just be:
```
transformed_locations = rotated_locs + ref_point1
```

I noticed this because it also fit better with my mental model of what is actually necessary. Hopefully this makes the function easier to understand for newcomers.

### DIFF
--- a/src/tbp/monty/frameworks/utils/spatial_arithmetics.py
+++ b/src/tbp/monty/frameworks/utils/spatial_arithmetics.py
@@ -562,7 +562,7 @@ def apply_rf_transform_to_points(
     # Apply detected rotation to new locations
     rotated_locs = object_rotation.inv().apply(zero_centered_locs)
 
-    # Undo the 0 centering to bring back into model reference frame
+    # Undo the 0 centering and align the new points with the model's reference frame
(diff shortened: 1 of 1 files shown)


vkakerbeck:
This is probably just personal preference but to me it was more intuitive to understand before. I get that these cancel out and are equivalent, but the reason I had written them out is because it made it easier to parse for my brain. For example, with the new version, I would be confused why we undo the 0 centering by subtracting ref_point1 when had applied it using ref_point2. It would at least be useful to add a comment to explain this.

nielsleadholm:
Ok interesting, I've updated the comment to what I think is a clearer description of what's actually happening, let me know what you think.

vkakerbeck (approved):
yes that works, thanks!

PR #1142, opened 2026-08-20
fix: Add check for pre-existing lights in Mujoco

nielsleadholm:
## The Problem 

While working with our compositional dataset (which is now supported in Mujoco thanks to Scott) in teleop, I noticed that the marble texture was very washed out and over-saturated. Looking at the standard YCB objects in Mujoco, these are also very "bright", most noticeable with lighter-textured objects. See examples below:

bright_a

bright_b

## The Reason

The Mujoco simulator call to `_recompile` happens multiple times, including when an object is added. This in turn means that the `_configure_lights` addition of a directional light gets called multiple times. 

### The Solution

The solution I've added is to check whether this light has already been initialized. With this change in place, the objects look like we would expect.

after_a

after_b


jeremyshoemaker:
Re-implemented by #1143

PR #1143, opened 2026-08-20
fix: don't add extra lights to MuJoCo scenes

jeremyshoemaker:
In the `_recompile` method we're unconditionally adding lights to the scene, so every time we add an object, we add another light to the scene, causing lighting issues.

There are only two places that we need to recreate them, both when we've got a fresh new MjSpec object, so this moves that initialization into a method called from both those places.

### DIFF
--- a/src/tbp/monty/simulators/mujoco/simulator.py
+++ b/src/tbp/monty/simulators/mujoco/simulator.py
@@ -187,24 +187,31 @@ class MuJoCoSimulator(SimulatedObjectEnvironment):
         self._renderers = {}
         self._agent_partials = [] if agents is None else agents
         self._agents = {}
-        self._create_agents()
 
         # Track how many objects we add to the environment.
         # This is used to give added objects unique names.
         self._object_count = 0
 
+        self._initialize_new_spec()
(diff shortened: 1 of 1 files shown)


src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -187,33 +187,40 @@ def __init__(
...
+
     def _configure_spec_settings(self) -> None:
         """Set all the relevant global settings on the spec object."""
         self.spec.option.gravity = (0.0, 0.0, 0.0)
         # Configure the maximum rendering resolution for the off-screen buffer.
         # Start with a default resolution in case we don't have agents and therefore
         # sensors to query, e.g. in tests.
         render_resolution = DEFAULT_RESOLUTION
-        if self._agents:
+        if self._agent_partials:
```
jeremyshoemaker, line 223: This is a minor unrelated bug that this change exposes. We're examining the partials in the `_max_sensor_resolution` method, so we should condition this call on whether we have those, not the agents themselves.

nielsleadholm (approved):
LGTM!

PR #1144, opened 2026-08-20
feat: MuJoCo custom object scaling support

jeremyshoemaker:
The `numenta_lab` objects use custom per object scaling stored in the JSON config files that Habitat uses. We need to support that in MuJoCo to correctly scale the objects when we use them in experiments.


jeremyshoemaker:
I feel like there's some awkwardness here around `ModelMetadata` and `ObjectMetadata`. I think there's a larger refactor needed here around pulling the object loading out into a different module, especially if we later incorporate loading other types of object formats or even whole scenes, and I don't think now is the time for it.

jeremyshoemaker:
I’m going to move this to draft. I had some thoughts about this over the weekend that I want to address. Specifically, I think the scale argument to `add_object` should scale whatever values are in the metadata file, not replace them. The metadata values define what the default scale should be, so scaling relative to it makes more sense.

jeremyshoemaker:
Okay, I've fixed the issue I was concerned about, and also found a bug in the original version.

I was using the scale from the metadata to determine the key to store the results in to prevent duplicate loading, but a caller would never know that value. They only know the value they asked for in the `add_object` call. 

So given that...
- the metadata file `scale` is redefining what 1.0 scale means for an object
- the `scale` argument to `add_object` is supposed to scale the object relative to whatever its default scale is

...I've changed the keys and mesh names to use the requested scale, not the calculated scale.

tests/unit/simulators/mujoco/mujoco_simulator_test.py
```diff
@@ -145,10 +145,70 @@ def test_custom_object_adding(self) -> None:
                 mesh.refquat, [np.sin(np.pi / 4), np.cos(np.pi / 4), 0.0, 0.0]
             )
 
+    def test_custom_object_scaling_in_call(self):
+        """Test adding a custom object with a scaling factor."""
```
tristanls-tbp: **nitpick**: It feels like the docstring is repeating what's already contained in the test name. Probably fine to remove.

tests/unit/simulators/mujoco/mujoco_simulator_test.py
```diff
@@ -145,10 +145,70 @@ def test_custom_object_adding(self) -> None:
...
+
+            geom = sim.spec.geom("valid_object_0")
+            assert geom.type == mjtGeom.mjGEOM_MESH
+            assert geom.meshname == "valid_object_2.0_2.0_2.0_mesh"
+
+            mesh = sim.spec.mesh(geom.meshname)
+            assert np.allclose(mesh.scale, [2.0, 2.0, 2.0])
+
+    def test_custom_object_scaling_from_metadata(self) -> None:
+        """Test adding a custom object with a scaling factor defined in metadata."""
```
tristanls-tbp: **nitpick**: It feels like the docstring is repeating what's already contained in the test name. Probably fine to remove.

src/tbp/monty/simulators/mujoco/objects.py
```diff
@@ -32,6 +32,9 @@ class ObjectMetadata:
     normals. The model compiler normalizes the quaternion automatically.
     """
 
+    scale: VectorXYZ = (1.0, 1.0, 1.0)
```
tristanls-tbp, line 35: **suggestion (non-blocking)**: Since `ZERO_VECTOR` already exists, a matching `ONES_VECTOR` might improve readability.

src/tbp/monty/simulators/mujoco/objects.py
```diff
@@ -32,6 +32,9 @@ class ObjectMetadata:
...
+    """Scaling factor for the model in each direction."""
```
tristanls-tbp, line 36: **suggestion**: I just noticed the trailing docstrings after attribute names in this class. They're very helpful, but [our style guide](https://docs.thousandbrains.org/docs/code-style-guide#code-docstrings), which points to [Google's class conventions](https://google.github.io/styleguide/pyguide.html#384-classes), places them in the `Attributes:` section of the class docstring.

Renders as

Image

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -91,6 +93,18 @@
 # configurations, e.g. in tests.
 DEFAULT_RESOLUTION = Resolution2D(width=64, height=64)
 
+# Key for tracking what custom objects we've already loaded meshes for.
+# Includes scale since MuJoCo stores scale on the mesh, and not the geom.
```
tristanls-tbp: **question**: This is helpful context. Any specific reason why this is a code comment and not a docstring?

**suggestion**: Move to a docstring.

**note**: I'm also noticing that the other constant docstrings are similarly written as code comments.

**suggestion (non-blocking)**: They should be similarly updated.

src/tbp/monty/simulators/mujoco/simulator.py
```diff
@@ -384,29 +398,23 @@ def _add_custom_object(
...
-            raise NotImplementedError(
-                "Custom objects do not currently support "
-                "'scale' other than (1.0, 1.0, 1.0)."
-            )
-
-        if object_type not in self._loaded_custom_types:
-            self._load_custom_object(object_type)
+        object_lookup_key: LoadedObjectKey = (object_type, scale)
+        if object_lookup_key not in self._loaded_custom_types:
+            object_lookup_key = self._load_custom_object(object_type, scale)
```
tristanls-tbp: **issue**: I'm confused by this API. I expected:

```python
object_lookup_key: LoadedObjectKey = (object_type, scale)
if object_lookup_key not in self._loaded_custom_types:
    self._load_custom_object(object_lookup_key)
```

You should be able to:
```python
object_type, scale = object_lookup_key
```
inside `self._load_custom_object(...)`

tests/unit/simulators/mujoco/mujoco_simulator_test.py
```diff
@@ -145,10 +145,70 @@ def test_custom_object_adding(self) -> None:
...
+        scale_z=st.floats(min_value=0.0001, max_value=2.0),
+    )
+    def test_custom_object_scaling_with_both(
+        self, scale_x: float, scale_y: float, scale_z: float
+    ) -> None:
+        """Test scaling a custom object when a scale is defined in both places.
+
+        We want to make sure that we use the scale argument to scale an object based
+        on its default scale, so if that default scale isn't 1.0 then we need to make
+        sure we scale that instead of just replacing it with the arugment scale.
```
tristanls-tbp:
```suggestion
        sure we scale that instead of just replacing it with the argument scale.
```

tristanls-tbp (commented):
Thank you for getting scaling into MuJoCo. I have one substantive comment and a few documentation notes that surfaced during my review.

src/tbp/monty/simulators/mujoco/simulator.py
tristanls-tbp: **suggestion (non-blocking)**: `object_lookup_key` -> `object_key`? "lookup" is implied in dictionary keys.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: I'm not in the habit of putting docstring style documentation on things other than functions or classes. I don't think Python actually sees them as docstrings. It's just other tooling that makes that connection. 

As you pointed out elsewhere with the `ObjectMetadata` attributes, I'm even inconsistent with it, but PyCharm's syntax highlighting sees them as two different syntactic elements:

image

The docstring on the class is highlighted like a comment, and the docstring on the property is highlighted like a multi-line string, which is what happens at the top level here.

...

I decided to look it up because it's been bothering me for a while, and the [PEP for docstrings](https://peps.python.org/pep-0257/) says that:

> String literals occurring elsewhere in Python code may also act as documentation. They are not recognized by the Python bytecode compiler and are not accessible as runtime object attributes (i.e. not assigned to `__doc__`), but two types of extra docstrings may be extracted by software tools:
> - String literals occurring immediately after a simple assignment at the top level of a module, class, or __init__ method are called “attribute docstrings”.
> - String literals occurring immediately after another docstring are called “additional docstrings”.

It also doesn't look like things get included in the documentation if they aren't "exported" in the `__all__` which this value isn't.

All of this is to say, I think I'll start following this approach going forward. Call me out if I revert to this style out of habit in the future.

src/tbp/monty/simulators/mujoco/simulator.py
jeremyshoemaker: Oh, yeah that's left over from when I was trying to store the values using the calculated scale, which was wrong, so it returned the key. I'll fix it.

src/tbp/monty/simulators/mujoco/objects.py
jeremyshoemaker: I've been thinking of doing that for a while, so now that you've nudged me, I'll just do it. Maybe as a separate PR though, because I want to change other places we initialize vectors like this, and that seems out of scope of this PR.

src/tbp/monty/simulators/mujoco/objects.py
jeremyshoemaker: I'll go ahead and change it, though I wonder if it would be better to use this docstring style instead especially in light of the PEP quote from [my other comment](https://github.com/thousandbrainsproject/tbp.monty/pull/1144#discussion_r3854094479), specifically the part about "attribute docstrings".

I usually prefer to keep documentation as close to the thing being documented as possible, for ease of maintenance.

tristanls-tbp: We talked about this and will document attributes, properties, and constants as docstrings, following their definitions.

tristanls-tbp (approved):
Thanks!

PR #1148, opened 2026-08-25
chore: remove stale 2LM compositional configs after 3LM migration

hlee9212:
## Summary

Clean up the compositional-object configuration and documentation following the move to the 3LM heterarchy.

- Update the compositional benchmark documentation.
- Remove the `correct_child_or_parent` metric & `consistent_child_obj` classification (and `parent_to_child_mapping`)
- Remove 2LM configs and replace with 3LM ones (and removing the `_3lm` suffixes).
- Remove unused 2LM configuration files, e.g. `evidence_2lm_*.yaml`. 
- **Note**: The 2LM configs used in hierarchy_test.py (e.g. `2lm_2sm.yaml` for connectivity) have been moved to `conf/experiment/test/hierarchy`. 

## Remaining Work (potentially another PR?)

There are a few 2LM configs left that are used in these hierarchy integration tests:
- `tests/integration/frameworks/models/hierarchy_test.py`
- `tests/mujoco/integration/frameworks/models/hierarchy_test.py`

I'd figure I'll tag @jeremyshoemaker since he might be most familiar with these tests (but they are likely all the way back from Brighton hackathon). I'm a bit hesitant to update these right now based on several notes in the `hierarchy_test.py` files:

```python
    def test_two_lm_heterarchy_experiment(self):
        """Test two LMs stacked on top of each other.

        LM0 receives input from SM0
        LM1 receives input from SM1 and LM0

        LM0 can store smaller models at a higher resolution and receives higher
        frequency input from SM0.
        LM1 can store larger models and a lower resolution and receives lower frequency
        input from SM1. It also receives input from LM0 once this one has a high
        confidence hypothesis.

        What happens in this experiment:
        Episodes 0-3: Both LMs have no_match and add a new model to memory.
        Episode 4: Both LMs recognize object 0 correctly and update their models.
        Episode 5: LM0 recognizes cubeSolid (new_object0) and updates its memory. LM1
            reaches a time out and does not update its memory (but has correct mlh).
        Evaluation:
            In each episode LM0 first recognizes the correct object. Since LM1 gets such
            low frequency input and stores very few points in its models it reaches
            no_match.

        NOTE: LM1 usually reaches no_match even if it knows about the object already. I
        think this is because for the first few observations it does not store features
        from LM0 yet. This would be different with a longer exploration phase that
        builds a full model of the object.

        NOTE: This test tests a lot of different things. We could split it up into many
        separate tests and test each aspect independently. However, this would increase
        computational cost since for many tests (like extending a graph correctly or
        getting the LM input) several episodes need to be run first (to build up graphs
        from which the object can be recognized in the first place). We could use mock
        data and test the LM in isolation like we already do in some places but we
        would still want to test the whole pipeline at least once. So why not make use
        of this longer run if we already have it? Maybe in the future we want to change
        this but this is my current reasoning.
        """
```

and

```python
    def test_semisupervised_stacked_lms_experiment(self):
        """Test two LMs stacked on top of each other with semisupervised learning.

        First, both LMs learn with supervision. Then, we remove the supervision for LM_0
        and only supervised LM_1 to learn a compositional object.
        Last, we load the learned models and evaluate on them.

        NOTE: It's not ideal to test compositional scenarios since we don't have access
        to compositional objects in the unit tests. So we are pretending that the cube
        and capsule are compositional and extending the graphs with the classification
        from LM_0 of the same object.

        NOTE: This test also implicitly tests a bunch of other things, such as:
        - Extending grid object models after loading them
        - Extending a graph with a new input channel
        - logging prediction errors
        """
```

I think I first want to have a clearer understanding of what integration tests should test for and what the intents were for these tests in particular? I'd also appreciate any concrete ideas on what to test for 3LM compositional configs, if any. 

@nielsleadholm Would you be okay if these are done in another PR? I'm happy to work on it here, but also want to be considerate of your/Scott's work on the compositional benefits figure.


jeremyshoemaker:
> I'd figure I'll tag `@jeremyshoemaker` since he might be most familiar with these tests (but they are likely all the way back from Brighton hackathon).

Yeah, those come from the implementation project that came out of that Brighton hackathon. I haven't looked at them in a while so I don't know that I have that much of the context still, though I did recently look at them for the MuJoCo migration. I can't remember if that was one of the ones that took some extra work to get working, I'd have to check the PR.

If I had the time to go through all of the integration tests, I think it would be a useful exercise to identity what each test is testing to see if we can break them apart into unit tests, or simplify them in any way. But that's a problem for another time.

src/tbp/monty/conf/experiment/supervised_pre_training_objects_with_stickers_2d_children.yaml
```diff
@@ -19,14 +19,14 @@ experiment:
     max_train_steps: 1000
     max_eval_steps: 500
     max_total_steps: 6000
-    n_train_epochs: 3
-    n_eval_epochs: 3
-    model_name_or_path: ${constants.compositional_pretrained_dir}/supervised_pre_training_objects_with_stickers_3d_children/pretrained/
+    n_train_epochs: 7
+    n_eval_epochs: 7
+    model_name_or_path: ${path.expanduser:${oc.env:MONTY_MODELS}/my_trained_models/supervised_pre_training_objects_with_stickers_3d_children/pretrained/}
```
nielsleadholm, line 24: These shouldn't be using "my_trained_models" right?

src/tbp/monty/conf/experiment/supervised_pre_training_objects_with_stickers_comp_models.yaml
```diff
@@ -21,12 +21,12 @@ experiment:
     max_total_steps: 6000
     n_train_epochs: ${constants.rotations_all_count}
     n_eval_epochs: 3
-    model_name_or_path: ${constants.compositional_pretrained_dir}/supervised_pre_training_objects_with_stickers_2d_children/pretrained/
+    model_name_or_path: ${path.expanduser:${oc.env:MONTY_MODELS}/my_trained_models/supervised_pre_training_objects_with_stickers_2d_children/pretrained/}
```
nielsleadholm, line 24: Again re. "my_trained_models".

src/tbp/monty/conf/experiment/supervised_pre_training_objects_with_stickers_monolithic_models.yaml
```diff
@@ -21,12 +21,12 @@ experiment:
     max_total_steps: 6000
     n_train_epochs: ${constants.rotations_all_count}
     n_eval_epochs: 3
-    model_name_or_path: ${constants.compositional_pretrained_dir}/supervised_pre_training_objects_with_stickers_2d_children/pretrained/
+    model_name_or_path: ${path.expanduser:${oc.env:MONTY_MODELS}/my_trained_models/supervised_pre_training_objects_with_stickers_2d_children/pretrained/}
```
nielsleadholm, line 24: Note

nielsleadholm (commented):
Thanks Hojae, just a few things to note as well as the in-line comments about my_trained_models.

- It should be possible to remove all of the `parent_to_child_mapping` definitions. I'm fine with that being a follow-up PR, but we should make sure to not lose track of it.
- That's fine to leave the 2LM configs that are needed for the integration tests, and I wouldn't spend time on changing them. The one thing that you could do is make sure these configs (environment etc) have something about "test_setup" or the like in their name, so it's clear that this is the only place we use them.

jeremyshoemaker:
> The one thing that you could do is make sure these configs (environment etc) have something about "test_setup" or the like in their name.

I'm pretty sure the top level configs themselves are already in the `test` subdirectory of the `src/tbp/monty/conf` tree, so changing the names shouldn't be needed. As for any of the component parts that are only used by the tests, we could change the names of those, but I'm also pretty sure we won't use them by mistake.

nielsleadholm:
@jeremyshoemaker ok great, yeah basically as long as their use-case is clear.

src/tbp/monty/conf/experiment/supervised_pre_training_objects_with_stickers_2d_children.yaml
hlee9212: Using "my_trained_models" is correct because `silent_info_train.yaml` saves the model (i.e. output_dir) to include "my_trained_models". Since the 2d_children needs to load from the pretrained 3d_children model, we need my_trained_models.

```bash
# silent_info_train.yaml
# @package experiment.config.logging

monty_log_level: SILENT
monty_handlers: []
wandb_handlers: []
python_log_level: INFO
python_log_to_file: true
python_log_to_stderr: true
output_dir: ${path.expanduser:${oc.env:MONTY_MODELS}/my_trained_models} <-- where model is being saved to
resume_wandb_run: false
wandb_id:
  _target_: wandb.util.generate_id
wandb_group: debugging
```

And in case you were were wondering why we saved to `$MONTY_MODELS/my_trained_models`, it's because of our AWS setup: https://docs.google.com/document/d/1I1edeUZdLlioBrYOKKhYzDQJJaEA42Hk/edit

src/tbp/monty/conf/experiment/supervised_pre_training_objects_with_stickers_comp_models.yaml
hlee9212: Same response as before. I'm trying to minimize having the user (or us in AWS) to pass custom `logging.output_dir` and `model_name_or_path` and just make the experiments work without passing any overrides. 

In the previous 2LM configs:
- 2D-child training loaded the published 3D-child model from `${constants.compositional_pretrained_dir}`.
- Both compositional and monolithic parent training loaded the published 2D-child model from the same shared directory.
- Their newly trained outputs were saved privately under `my_trained_models`.

Therefore, retraining a lower-level LM did not automatically feed it into the next stage. When I was making changes to the 3d_children or 2d_children configs & retraining (which was all the time), then I had to explicitly overwrite `model_name_or_path` every single time.

src/tbp/monty/conf/experiment/supervised_pre_training_objects_with_stickers_monolithic_models.yaml
hlee9212: See above.

hlee9212:
@jeremyshoemaker and @nielsleadholm:

> As for any of the component parts that are only used by the tests, we could change the names of those, but I'm also pretty sure we won't use them by mistake.

I moved the component parts (e.g. 2lm_2sm.yaml from connectivity or evidence_2lm.yaml from learning_module) to `conf/experiment/test`. 

---
@nielsleadholm:

I removed `parent_to_child_mapping` - good point. This also reminded me to remove the `consistent_child_obj` since we don't need that anymore. Lots of snapshot updates because this removes `parent_to_child_mapping: null` in every snapshot, I think.

src/tbp/monty/conf/experiment/supervised_pre_training_objects_with_stickers_comp_models.yaml
nielsleadholm: Ok thanks, that makes sense.

tests/unit/frameworks/run_parallel_test.py
```diff
@@ -44,17 +43,17 @@ def test_accepts_supported_episode_metrics(self):
         overall_stats = get_overall_stats(stats)
 
         self.assertEqual(overall_stats["overall/percent_correct"], 100.0)
+        self.assertNotIn("overall/percent_consistent_child_obj", overall_stats)
```
nielsleadholm: Non-blocking - what's the reason for adding this? Shouldn't it be sufficient that we've just removed all the relevant code?

nielsleadholm (approved):
Great, thanks for those updates Hojae. Overall looks good, just a non-blocking comment.

tests/unit/frameworks/run_parallel_test.py
hlee9212: My bad, I only had to to check that I did indeed remove all instances of consistent_child_obj. Removed.

tests/unit/frameworks/run_parallel_test.py
```diff
@@ -12,18 +12,20 @@
...
+    per_lm_stats,
+)
 
 
 class PerLMStatsTest(unittest.TestCase):
     def test_reconstructs_named_accuracy(self):
         eval_stats = pd.DataFrame(
             {
-                "lm_id": ["LM_0"] * 4 + ["LM_1"] * 4,
+                "lm_id": ["LM_0"] * 3 + ["LM_1"] * 4,
```
jeremyshoemaker: I know why you changed this, since you removed an item from the list below, but does this make sense from an experiment setup perspective? Would we get only 3 results from LM_0 when we get 4 from LM_1?

Should we add back a result to the list to make it match what a real experiment would do?

I've been sitting here trying to figure out _why_ this changed, and it took me too long to sort out, so I'm trying to save our future selves from having the same question of why these are different.

Maybe this test really should have pulled out the 4 as a variable like `episode_count` or something.

jeremyshoemaker (commented):
Just one not too serious question, before I approve.

tests/unit/frameworks/run_parallel_test.py
hlee9212: Yeah I was waffling back and forth about this. I think I'll it back in and also just make a clearer DataFrame that we would see in eval_stats.csv.

Update: Made a clearer DataFrame.

jeremyshoemaker (approved):
Thanks! That definitely makes more sense compared to an actual experiment.

PR #1150, opened 2026-08-26
refactor: introduce ONES_VECTOR constant

jeremyshoemaker:
We have ZERO_VECTOR and IDENTITY_QUATERNION constants defined in `tbp.monty.math`, and I've been thinking of adding one for a vector of all ones, but hadn't really felt the need until @tristanls-tbp suggested it in #1144.


tristanls-tbp (approved):

tristanls-tbp:
**note**: could be a `feat:`, as it introduces a new surface to the public API.

jeremyshoemaker:
> could be a feat:, as it introduces a new surface to the public API.

I thought about that a bit, and went back and forth. The values aren't changing, so I went with `refactor`.

tristanls-tbp:
If the values changed, that would indicate a breaking change (`!`). An orthogonal concern.

PR #1151, opened 2026-08-26
test: make uv tests failable

jeremyshoemaker:
I noticed that while we made the `uv` tests required, they aren't failing when they should. This is because we forgot to remove the `continue-on-errors: true` option from the workflow configuration.


tristanls-tbp:
**note**: If this ends up as only a GitHub Actions change, then this should be `ci:`. `fix:` is reserved for `tbp/monty/src`. But we'll see if other changes are needed.

jeremyshoemaker:
> **note**: If this ends up as only a GitHub Actions change, then this should be `ci:`. `fix:` is reserved for `tbp/monty/src`. But we'll see if other changes are needed.

I already knew there would be other changes, which is why I went with `fix:`.

tests/unit/simulators/mujoco/object_metadata_decoder_test.py
```diff
@@ -69,15 +69,6 @@ def test_invalid_json_raises_error(self):
...
-            load_object_metadata(self.missing_field_json, "test_object")
-
     def test_metadata_extra_fields(self):
         with pytest.raises(InvalidObjectMetadata, match=r"extra fields: {'extra'}"):
             load_object_metadata(self.extra_field_json, "test_object")
-
-    def test_metadata_missing_extra_fields(self):
-        match_str = r"missing fields: {'refquat'} and extra fields: {'extra'}"
-        with pytest.raises(InvalidObjectMetadata, match=match_str):
-            load_object_metadata(self.missing_extra_field_json, "test_object")
```
jeremyshoemaker, line 83: Even I forget to run the tests locally every time.

tests/mujoco/integration/frameworks/models/evidence_matching/evidence_lm_test.py
```diff
@@ -127,11 +127,7 @@ def test_moving_off_object(self) -> None:
             exp.run()
 
         self.assertEqual(
-            len(
-                exp.model.learning_modules[0].buffer.get_all_locations_on_object(
-                    input_channel="patch"
-                )
-            ),
+            len(exp.model.learning_modules[0].buffer.get_all_locations_on_object()),
```
jeremyshoemaker, line 130: This is mirroring a change in the original version of this test that was missed because the test didn't appear to fail.

tests/unit/simulators/mujoco/object_metadata_decoder_test.py
jeremyshoemaker: To be clear, these are being deleted because we aren't checking for missing fields anymore.

jeremyshoemaker:
> I already knew there would be other changes, which is why I went with `fix:`.

Actually, maybe it's not a fix? The changes I needed to make are in the tests, and regenerating the config snapshots, but not the main part of Monty.

tristanls-tbp (approved):

PR #1152, opened 2026-08-26
docs: class attribute, instance property, and constant docstrings style guide

tristanls-tbp:
After going back and forth a few times about how to write docstrings for class attributes, instance properties, and constants, we reached a consensus in engineering.

Why? 

For constants, the docstring is preferable to an inline comment, since an inline comment does not appear in the documentation.

For class attributes and instance properties, the general heuristic is to keep together things that are coupled rather than split them across different locations. And, with docstrings following the attribute or property definition, one does not need to repeat the name of the attribute/property as a string. Instead of:
```python
class MyClass:
    """...

    ...

    Attributes:
        potato: ...
        tomato: ...
    """

    potato: str
    tomato: str
```
We have:
```python
class MyClass:
    """...
    
    ...
    """

    potato: str
    """..."""
    tomato: str
    """..."""
```
The documentation output is the same; attribute/property docstrings are closer to the definitions; and accidental typos are eliminated.

I'm requesting a review from everyone so that you have a chance to read and comment on this suggested style update.


docs/contributing/style-guide/code-style-guide.md
```diff
@@ -16,7 +16,36 @@ A convenient way to ensure your code is formatted correctly is using the [ruff f
...
+### Class Attribute, Instance Property, and Constant Docstrings
+
+Diverging from the Google Style docstring convention, we place docstrings for class attributes, instance properties, and constants after their declarations.
+
+```python
+MY_CONSTANT = 17
+"""This is my constant docstring."""
+
+class MyClass:
+
```
dschumacher-tbp, line 37: suggestion: add a docstring for the class to your example so it's a better template.

dschumacher-tbp (approved):
I like it.

hlee9212 (approved):
I'm on board with the idea.

ramyamounir (approved):
👍

docs/contributing/style-guide/code-style-guide.md
tristanls-tbp: https://github.com/thousandbrainsproject/tbp.monty/pull/1152/changes/c63e0b0e67b26c2d08d9faf8a7c4e05c354fb6e4

scottcanoe (approved):
I've not done it like this before, but it's fine by me. Is there a ruff check for this style?

docs/contributing/style-guide/code-style-guide.md
```diff
@@ -16,7 +16,37 @@ A convenient way to ensure your code is formatted correctly is using the [ruff f
...
+	my_property: str
+	"""This is my instance property docstring."""
+
+	_my_private_ish_property: str
+	"""This is my "private" instance property docstring."""
+
+	def __init__(self, my_property: str):
+		"""Initialize MyClass.
+
+		Arguments:
```
AgentRev: nitpick: Monty primarily uses `Args:` here, like most Google Style codebases

docs/contributing/style-guide/code-style-guide.md
```diff
@@ -16,7 +16,37 @@ A convenient way to ensure your code is formatted correctly is using the [ruff f
...
 ## Code Docstrings
 
-We adopted the Google Style for docstrings. For more details, see the [Google Python Style Guide - 3.8 Comments and Docstrings](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings).
+We adopted the Google Style for docstrings, with modifications. For more details on the Google Style docstrings, see the [Google Python Style Guide - 3.8 Comments and Docstrings](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings).
+
+### Class Attribute, Instance Property, and Constant Docstrings
+
+Diverging from the Google Style docstring convention, we place docstrings for class attributes, instance properties, and constants after their declarations.
+
+```python
```
AgentRev, line 25: suggestion: Maybe a module docstring here too?

AgentRev (commented):
Good idea!

vkakerbeck (approved):
LGTM. The only thing that is unclear to me is why we are diverging from the standard Google Style. I am sure you discussed good reasons, but maybe you could share them here (on line 23? or in the PR description?) as well. Is it so that IDE hovering displays those doc strings?

nielsleadholm (approved):
LGTM, but yeah like Viviane it would be interesting to know a bit more about what behind the decision.

tristanls-tbp:
@vkakerbeck @nielsleadholm I updated the description.

docs/contributing/style-guide/code-style-guide.md
tristanls-tbp: Oh, good catch @AgentRev, thanks!

docs/contributing/style-guide/code-style-guide.md
tristanls-tbp: Ok, done.

docs/contributing/style-guide/code-style-guide.md
```diff
@@ -16,7 +16,39 @@ A convenient way to ensure your code is formatted correctly is using the [ruff f
 
 ## Code Docstrings
 
-We adopted the Google Style for docstrings. For more details, see the [Google Python Style Guide - 3.8 Comments and Docstrings](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings).
+We adopted the Google Style for docstrings, with modifications. For more details on the Google Style docstrings, see the [Google Python Style Guide - 3.8 Comments and Docstrings](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings).
+
+### Class Attribute, Instance Property, and Constant Docstrings
+
+Diverging from the Google Style docstring convention, we place docstrings for class attributes, instance properties, and constants after their declarations.
```
jeremyshoemaker, line 23: Earlier, I thought about commenting here to say maybe we should include the _why_ of this decision, but then didn't. 

Given everyone else's comments asking about it, I think it might make sense to add it here. I'm trying to think of a succinct explanation based on our discussions.

For everyone's context, I was originally against this style because the Python interpreter doesn't process these as docstrings in the way it does on class, module, and function objects. When the interpreter sees a docstring on a class, module, or function, it gets stored in the `__doc__` attribute on the object, so you can access it at runtime if you want.

This is why in IDEs like PyCharm, it highlights these two kinds of "docstrings" differently. The "real" ones look like comments, but the others get highlighted as regular strings. VSCode seems to highlight them both in the same style, but both see them as documentation and provide it in tooltips and other ways.

But then I looked up [PEP-257](https://peps.python.org/pep-0257/) which discusses docstring conventions, and it has this to say:

> String literals occurring elsewhere in Python code may also act as documentation. They are not recognized by the Python bytecode compiler and are not accessible as runtime object attributes (i.e. not assigned to `__doc__`), but two types of extra docstrings may be extracted by software tools:
> - String literals occurring immediately after a simple assignment at the top level of a module, class, or `__init__` method are called “attribute docstrings”.
> - String literals occurring immediately after another docstring are called “additional docstrings”.
>
> ~ [PEP-257 - Docstring Conventions](https://peps.python.org/pep-0257/)

Given that, I don't really have the objection I had before, which leads to the other reasons I now think this is a better approach:

1. The top level module attributes aren't as easily documentable without using this. There's probably a way to do it in the module docstring, but that seems problematic and leads to my second point.
2. Using a docstring right after the attribute puts the documentation as close as possible to the thing being documented. Having to put "Attributes:" in a class docstring some distance from the attributes themselves makes it more likely that they won't be updated when the attributes change.

AgentRev (approved):

AgentRev:
Side-note, I've been using [format-docstring](https://github.com/jsh9/format-docstring) lately to tidy up my docstrings, it works pretty well. They added Google Style support a couple months ago. The arguments I use for its External Tools entry in PyCharm with Monty are:

```
--line-length 88 --docstring-style google --fix-rst-backticks false --include-arg-types false --include-arg-defaults false --include-return-and-yield-types false "$FilePath$"
```

>  Is there a ruff check for this style?

@scottcanoe Apart from the usual Ruff check for Google Style, this bespoke convention is not enforceable by Ruff. It would require something like a pre-commit hook script using Abstract Syntax Tree.

docs/contributing/style-guide/code-style-guide.md
tristanls-tbp: Attempted to address in https://github.com/thousandbrainsproject/tbp.monty/pull/1152/commits/3a111ffd112678077b4f23d502b3ec030f87e6c9

brothman-tbp:
LGTM

brothman-tbp (approved):

docs/contributing/style-guide/code-style-guide.md
```diff
@@ -26,30 +26,44 @@ Diverging from the Google Style docstring convention, we place docstrings for cl
...
-			my_property: My argument to initialize with.
+			my_property: This is a method argument description.
 		"""
 		self.my_property = my_property
 		self._my_private_ish_property = "shhh"
 ```
 
+Using a docstring right after the attribute, property, or constant definition puts the documentation as close as possible to what is being documented. Having to put `Attributes: ` in a class docstring some distance from the attributes/properties (as per the Google Style convention) makes it more likely that they won't be updated when names change.
+
+Note that while docstrings after a class, module, or function are stored on the `__doc__` attribute of the object, docstrings placed after attribute, property, or constant definition do not.
```
jeremyshoemaker: Minor grammatical fix for subject-verb agreement and plurality. The sentence starts with "docstrings... are stored" but ends with "docstrings... definition do not" which doesn't match.

```suggestion
Note that while docstrings after a class, module, or function are stored on the `__doc__` attribute of the object, docstrings placed after attribute, property, or constant definitions are not.
```

jeremyshoemaker (approved):
One small grammatical error, but otherwise this looks great.

PR #1157, opened 2026-08-27
docs: Add comment to cmp change to displacements

nielsleadholm:
After yesterday's research meeting, I felt like it would be worth adding a short clarifying comment to this Future Work item. Otherwise it's a bit confusing why it even exists.


docs/future-work/voting-improvements/can-we-change-the-cmp-to-use-displacements-instead-of-locations.md
```diff
@@ -11,7 +11,7 @@ status: open
 ---
 
 > [!NOTE]
-> Although, there is reasonable evidence that the brain may be using movements instead of locations, it is unclear whether there are any computational benefits to doing this. Since we haven't identified concrete benefits, the `improved-metric` field is left open. A first step to tackling this task would be to think through any functional implications of this change.
+> Although there is reasonable evidence that the brain may be using movements instead of locations, it is unclear whether there are any computational benefits to doing this. As such, the `improved-metric` field is left open. A first step to tackling this task would be to think through any functional implications of this change. In the meantime, we include this task because choosing the right assumptions about a system is key to developing long-term, successful solutions. Using displacements rather than locations would have impacts both for how voting and motor actions are implemented. Continuing to use locations, and diverging from a more brain-like approach, may therefore lead us into a "local minimum" style solution for many elements in Monty.
```
vkakerbeck:
```suggestion
> Although there is reasonable evidence that the brain may be using movements instead of locations, it is unclear whether there are any computational benefits to doing this. As such, the `improved-metric` field is left open. A first step to tackling this task would be to think through any functional implications of this change. In the meantime, we include this task because choosing the right assumptions about a system is key to developing long-term, successful solutions. Using displacements rather than locations affect how both voting and motor actions can be implemented. Continuing to use locations, and diverging from a more brain-like approach, may therefore lead us into a "local minimum" style solution for some elements in Monty.
```

vkakerbeck (commented):
Thanks for making that update! Just a slight wording suggestion

vkakerbeck (approved):

PR #1158, opened 2026-08-27
ci: trigger triage on synchronize event

tristanls-tbp:
For reasons(?) the triage job is not always triggered despite being set for `opened`, `edited`, `labeled`, `unlabeled`. So, we'll also trying `synchronize`.

Screenshot 2026-08-27 at 09 25 14


jeremyshoemaker (approved):
I believe this is the correct solution. The reason it wasn't working on `edited` is because that one only triggers when the PR itself is edited, or the target branch (they use the term "base branch") changes. I _think_ they mean if you change the branch that the PR is targeting, and not that the target branch itself was updated with new commits, but it's unclear.

> The title or body of a pull request was edited, or the base branch of a pull request was changed.
> ~ [GitHub documentation](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request)

The `synchronize` action type says it triggers when the pull request's head branch is updated, which is what I would call the source branch, i.e. the one with your changes on your fork.

> A pull request's head branch was updated. For example, the head branch was updated from the base branch or new commits were pushed to the head branch.
> ~ [GitHub documentation](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=edited#pull_request)

PR #1164, opened 2026-08-31
refactor!: remove unused `MontyGeneralizationExperiment` experiment

ramyamounir:

## Breaking Changes
- Remove `MontyGeneralizationExperiment` experiment.

The `MontyGeneralizationExperiment` is designed to remove the primary target object being inferred to evaluate Monty's generalization ability to find the next best object, i.e., most similar object in the learned models. This experiment is untested, and not used in any benchmarks (deadcode). It also gets in the way of an in-progress IP that may end up redesigning the pre/post-episode pattern being overridden in this class.


nielsleadholm (approved):
Nice, thanks for cleaning this up. For posterity, it's also worth noting that the same question that this experiment explores can be easily tested via configs or analysis (e.g. training and inference pipeline, or just looking at the distribution of hypotheses that Monty has during inference), so having a dedicated experiment class is unnecessary.

PR #1167, opened 2026-08-31
ci: 'reads as generated' label and generic LLM comment

tristanls-tbp:
This pull request provides for a quick mechanism for Maintainers to mark an issue or a pull request with a `reads as generated` label, which will result in an automated comment indicating the contribution will not be reviewed further, with a reference to our [Large Language Model Use](https://docs.thousandbrains.org/docs/large-language-model-use) guidelines.


hlee9212 (approved):
LGTM.

docs/contributing/large-language-model-use.md
```diff
@@ -6,7 +6,11 @@ We're really excited to receive contributions from the wider community - one of
...
+As such, the following outlines several concerning attributes that are associated with LLM-generated contributions, and why these can be problematic. While we do not institute any strict guidance, please note that if we feel that the use of AI/LLMs has resulted in a contribution that matches these concerning patterns, Maintainers reserve the right to not review your contribution. We do not want to discourage contributions, but this balancing act on our part is an unfortunate side-effect of the small size of the Maintainers team vs. the speed at which LLM content can be generated. If a Maintainer interprets your contribution as not meeting the guidance below, you will see a `reads as generated` label applied to your contribution along with the following comment:
```
nielsleadholm:
```suggestion
As such, the following outlines several concerning attributes that are associated with LLM-generated contributions, and why these can be problematic. While we do not institute any strict guidance, please note that if we feel that the use of AI/LLMs has resulted in a contribution that matches these concerning patterns, Maintainers reserve the right to not review your contribution. We do not want to discourage contributions, but this balancing act on our part is an unfortunate side-effect of the small size of the Maintainers team vs. the speed at which LLM content can be generated. If a Maintainer interprets a PR contribution as not meeting the guidance below, you will see a `reads as generated` label applied to your contribution along with the following comment:
```

Unless we plan on having something similar on the Discourse forums, it would be worth clarifying this relates to PRs.

nielsleadholm (approved):
Nice, I like the idea of having a dedicated label. Just a non-blocking suggestion.

docs/contributing/large-language-model-use.md
```diff
@@ -6,7 +6,11 @@ We're really excited to receive contributions from the wider community - one of
...
+
+```
+Thank you for your contribution. While this may not have been your intention, this contribution reads as generated by an LLM to a degree that Maintainers will not review it further. We might be wrong, and if this is the case, our apologies, and please let us know we made a mistake. For our general guidance on the use of AI/LLMs, please see [Large Language Model Use](https://docs.thousandbrains.org/docs/large-language-model-use).
+```
```
jeremyshoemaker: This should probably be a block quote instead of a code block. Looking at the preview, it doesn't wrap nicely as a code block. Perhaps it renders better as a code block on the site?

There's also what feels like a comma-splice to me. 
> We might be wrong, and if this is the case, our apologies, and please let us know we made a mistake.

I'd probably split this sentence up.

```suggestion
> Thank you for your contribution. While this may not have been your intention, this contribution reads as generated by an LLM to a degree that Maintainers will not review it further. We might be wrong. If this is the case, our apologies, and please let us know we made a mistake. For our general guidance on the use of AI/LLMs, please see [Large Language Model Use](https://docs.thousandbrains.org/docs/large-language-model-use).
```

.github/workflows/triage.yml
```diff
@@ -46,6 +49,16 @@ jobs:
...
+      - name: reads as generated label is present
+        if: ${{ contains(github.event.pull_request.labels.*.name, 'reads as generated') }}
+        uses: ./tbp.monty/.github/actions/pin_comment
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          issue_number: ${{ github.event.pull_request.number }}
+          comment_id: reads-as-generated
+          working_directory: tbp.monty
+          message: |
+            Thank you for your contribution. While this may not have been your intention, this contribution reads as generated by an LLM to a degree that Maintainers will not review it further. We might be wrong, and if this is the case, our apologies, and please let us know we made a mistake. For our general guidance on the use of AI/LLMs, please see [Large Language Model Use](https://docs.thousandbrains.org/docs/large-language-model-use).
```
jeremyshoemaker: If you make the comma-splice change I mentioned below, it also needs to be updated here.

```suggestion
            Thank you for your contribution. While this may not have been your intention, this contribution reads as generated by an LLM to a degree that Maintainers will not review it further. We might be wrong. If this is the case, our apologies, and please let us know we made a mistake. For our general guidance on the use of AI/LLMs, please see [Large Language Model Use](https://docs.thousandbrains.org/docs/large-language-model-use).
```

docs/contributing/large-language-model-use.md
jeremyshoemaker: I'm not sure how GitHub will treat multiple suggestions on a single line of text, so you might need to do this by hand if you decide to do so. This is also sort of outside the specific scope of this PR's changes.

I would probably strike "please note that" from this sentence as it doesn't really add anything to the message. This is part of a broader thing I've been taught, to avoid using "please" in technical writing. 

This reads to me like we're politely asking someone to... continue reading? 🤨 My thinking is that if we have something that we really want people to pay attention to, or "note", then it should be in a callout box set apart from the rest of the text, not hidden in the middle of the paragraph.

```suggestion
As such, the following outlines several concerning attributes that are associated with LLM-generated contributions, and why these can be problematic. While we do not institute any strict guidance, if we feel that the use of AI/LLMs has resulted in a contribution that matches these concerning patterns, Maintainers reserve the right to not review your contribution. We do not want to discourage contributions, but this balancing act on our part is an unfortunate side-effect of the small size of the Maintainers team vs. the speed at which LLM content can be generated. If a Maintainer interprets your contribution as not meeting the guidance below, you will see a `reads as generated` label applied to your contribution along with the following comment:
```

There's a similar thing in the last line of this file that starts with "Please note that". I don't think it's necessary for the sentence really.

tristanls-tbp: Ah, valid point. In light of that, I moved the labeling prose to the code section in b74a323a12d53418ca4ee673f2789a9d03bc99b0

jeremyshoemaker (approved):
None of my suggestions here are blocking.

docs/contributing/large-language-model-use.md
tristanls-tbp: Good point on the block quote. I originally had it as a code block for easy copy-and-paste, but since I added automation, it makes more sense as a block quote now 👍.

jeremyshoemaker:
I spent some time looking at other projects' policies in this area, as I feel like we could maybe be more explicit, but I didn't really come away from it with any concrete suggestions. 

I looked at [Python](https://devguide.python.org/getting-started/ai-tools/index.html), [Rust](https://forge.rust-lang.org/policies/llm-usage.html), and [Zig](https://ziglang.org/code-of-conduct/) as just examples of ones across the spectrum. I mention them if anyone wants to take a look at how some other projects are approaching this.

docs/contributing/large-language-model-use.md
tristanls-tbp: I hear you, but I'm going to avoid discussing and changing this in this PR 🙂.

.github/workflows/triage.yml
tristanls-tbp: Updated in 44a0f8a5ee6de4c5cc5fa316a58b363e13304894

PR #1174, opened 2026-09-02
fix!: correctly merge new and previous pose vectors

tristanls-tbp:
This pull request aims to fix a bug discovered by @jeremyshoemaker during the SciPy upgrade. SciPy started raising exceptions because it observed rotation matrices with invalid determinants. 

The essence of the problem is this line:
https://github.com/thousandbrainsproject/tbp.monty/blob/25232e7762ab0d083a2d95f15ed566bdbe1214c9/src/tbp/monty/frameworks/models/object_model.py#L855

This is not a valid way of averaging rotation matrices (element-wise), as it does not preserve orthonormality. Take note of the care that `pose_vector_mean(...)` takes to ensure that pose vectors retain desired properties. It throws away up to half of the pose vectors to ensure that averaging makes sense and ensures orthonormality, etc.
https://github.com/thousandbrainsproject/tbp.monty/blob/25232e7762ab0d083a2d95f15ed566bdbe1214c9/src/tbp/monty/frameworks/utils/object_model_utils.py#L315

The fix is to replace the faulty merge of the new and previous pose vectors with one that maintains the result's orthonormality. Hence, the `pose_vector_merge` function that uses `Rotation.mean()`, which, in turn, ensures the result's orthonormality.

### DIFF
--- a/src/tbp/monty/frameworks/models/object_model.py
+++ b/src/tbp/monty/frameworks/models/object_model.py
@@ -32,6 +32,7 @@ from tbp.monty.frameworks.utils.object_model_utils import (
     get_values_from_dense_last_dim,
     increment_sparse_tensor_by_count,
     pose_vector_mean,
+    pose_vector_merge,
     remove_close_points,
     torch_graph_to_numpy,
 )
@@ -841,18 +842,27 @@ class GridObjectModel(GraphObjectModel):
                 num_old_obs = num_obs_in_voxel - num_new_obs
 
                 if feature == "pose_vectors":
-                    if avg_feat is None:
-                        avg_feat = previous_average
-                    elif use_cds_to_update is False:
-                        avg_feat[3:] = previous_average[3:]
-                elif feature == "object_id" and avg_feat != previous_average:
-                    # TODO: Figure out a more nuanced way to take into account past obs
-                    if num_old_obs > num_new_obs:
-                        avg_feat = previous_average
-                    else:
-                        previous_average = avg_feat
-                # NOTE: could weight these
-                avg_feat = (avg_feat + previous_average) / 2
+                    avg_feat = (
+                        previous_average
+                        if avg_feat is None
+                        else pose_vector_merge(
+                            avg_feat,
+                            previous_average,
+                            use_cds_to_update,
+                            num_new_obs,
+                            num_old_obs,
+                        )
+                    )
+                else:
+                    if feature == "object_id" and avg_feat != previous_average:
+                        # TODO: Figure out a more nuanced way to take into account
+                        # past obs
+                        if num_old_obs > num_new_obs:
+                            avg_feat = previous_average
+                        else:
+                            previous_average = avg_feat
+                    # NOTE: could weight these
+                    avg_feat = (avg_feat + previous_average) / 2
             target_ids = target_fm[feature]
             new_feature_avg[target_ids[0] : target_ids[1]] = avg_feat
         return new_feature_avg
--- a/src/tbp/monty/frameworks/models/object_model.py
+++ b/src/tbp/monty/frameworks/models/object_model.py
@@ -842,17 +842,16 @@ class GridObjectModel(GraphObjectModel):
                 num_old_obs = num_obs_in_voxel - num_new_obs
 
                 if feature == "pose_vectors":
-                    avg_feat = (
-                        previous_average
-                        if avg_feat is None
-                        else pose_vector_merge(
+                    if avg_feat is None:
+                        avg_feat = previous_average
+                    else:
+                        avg_feat = pose_vector_merge(
                             avg_feat,
                             previous_average,
                             use_cds_to_update,
                             num_new_obs,
                             num_old_obs,
                         )
-                    )
                 else:
                     if feature == "object_id" and avg_feat != previous_average:
                         # TODO: Figure out a more nuanced way to take into account
--- a/src/tbp/monty/frameworks/models/object_model.py
+++ b/src/tbp/monty/frameworks/models/object_model.py
@@ -813,7 +813,7 @@ class GridObjectModel(GraphObjectModel):
             pdefined_ids = obs_fm["pose_fully_defined"]
             pose_vecs = new_features_in_voxel[:, pv_ids[0] : pv_ids[1]]
             pdefined = new_features_in_voxel[:, pdefined_ids[0] : pdefined_ids[1]]
-            pv_mean, use_cds_to_update = pose_vector_mean(pose_vecs, pdefined)
+            pv_mean, _ = pose_vector_mean(pose_vecs, pdefined)
         for feature in obs_fm:
             ids = obs_fm[feature]
             feats = new_features_in_voxel[:, ids[0] : ids[1]]
@@ -848,7 +848,6 @@ class GridObjectModel(GraphObjectModel):
                         avg_feat = pose_vector_merge(
                             avg_feat,
                             previous_average,
-                            use_cds_to_update,
                             num_new_obs,
                             num_old_obs,
                         )
--- a/src/tbp/monty/frameworks/models/object_model.py
+++ b/src/tbp/monty/frameworks/models/object_model.py
@@ -852,14 +852,14 @@ class GridObjectModel(GraphObjectModel):
                             num_new_obs,
                             num_old_obs,
                         )
-                else:
-                    if feature == "object_id" and avg_feat != previous_average:
+                elif feature == "object_id" and avg_feat != previous_average:
                         # TODO: Figure out a more nuanced way to take into account
                         # past obs
                         if num_old_obs > num_new_obs:
                             avg_feat = previous_average
                         else:
                             previous_average = avg_feat
+                else:
                     # NOTE: could weight these
                     avg_feat = (avg_feat + previous_average) / 2
             target_ids = target_fm[feature]
--- a/src/tbp/monty/frameworks/models/object_model.py
+++ b/src/tbp/monty/frameworks/models/object_model.py
@@ -853,12 +853,12 @@ class GridObjectModel(GraphObjectModel):
                             num_old_obs,
                         )
                 elif feature == "object_id":
-                        # TODO: Figure out a more nuanced way to take into account
-                        # past obs
-                        if num_old_obs > num_new_obs:
-                            avg_feat = previous_average
-                        else:
-                            previous_average = avg_feat
+                    # TODO: Figure out a more nuanced way to take into account
+                    # past obs
+                    if num_old_obs > num_new_obs:
+                        avg_feat = previous_average
+                    else:
+                        previous_average = avg_feat
                 else:
                     # NOTE: could weight these
                     avg_feat = (avg_feat + previous_average) / 2
--- a/src/tbp/monty/frameworks/models/object_model.py
+++ b/src/tbp/monty/frameworks/models/object_model.py
@@ -852,7 +852,7 @@ class GridObjectModel(GraphObjectModel):
                             num_new_obs,
                             num_old_obs,
                         )
-                elif feature == "object_id" and avg_feat != previous_average:
+                elif feature == "object_id":
                         # TODO: Figure out a more nuanced way to take into account
                         # past obs
                         if num_old_obs > num_new_obs:
--- a/src/tbp/monty/frameworks/utils/object_model_utils.py
+++ b/src/tbp/monty/frameworks/utils/object_model_utils.py
@@ -322,8 +322,8 @@ def orthonormal_pose_vectors(
 ) -> npt.NDArray[np.float64]:
     """Build a right-handed orthonormal pose from two rough directions.
 
-    The surface normal is used as a give, up to normalization. The curvature direction
-    is orthogonlized agaist it, so the returned pose vectors are always a valid
+    The surface normal is used as a given, up to normalization. The curvature direction
+    is orthogonalized against it, so the returned pose vectors are always a valid
     rotation matrix.
 
     Args:
@@ -357,7 +357,7 @@ def pose_vector_merge(
 
     Surface normals observed from opposite sides of a surface are not two estimates
     of the same direction, so they are not averaged. The side with more observations
-    is kept, the same way `pose_vector_mean` discards the minority side withing a
+    is kept, the same way `pose_vector_mean` discards the minority side within a
     single batch of observations.
 
     Curvature directions are ambiguous in direction, so the new one is flipped to agree
--- a/src/tbp/monty/frameworks/utils/object_model_utils.py
+++ b/src/tbp/monty/frameworks/utils/object_model_utils.py
@@ -441,7 +441,7 @@ def pose_vector_mean(pose_vecs, pose_fully_defined):
         # )
         # Just take 1st one. Shouldn't matter since cd should not be used anyways if
         # not pose_fully_defined. Only has a small effect on sampled possible poses.
-        pv_means = np.hstack([norm_mean, cds1[0], cds2[0]])
+        pv_means = orthonormal_pose_vectors(norm_mean, cds1[0])
         use_cds_to_update = False
     else:
         # Find cds pointing in opposing directions and invert them. This is needed
@@ -449,13 +449,8 @@ def pose_vector_mean(pose_vecs, pose_fully_defined):
         # equivalent. If we average over opposing directions, we will get noise.
         cd1_dirs = get_right_hand_angle(cds1, cds2[0], norm_mean) < 0
         cds1[cd1_dirs] = -cds1[cd1_dirs]
-        cd1_mean = normalize(np.mean(cds1, axis=0))
-        # Get the second cd by calculating a vector orthogonal to cd1 and surface normal
-        cd2_mean = normalize(np.cross(norm_mean, cd1_mean))
-        if get_right_hand_angle(cd1_mean, cd2_mean, norm_mean) < 0:
-            cd2_mean = -cd2_mean
+        pv_means = orthonormal_pose_vectors(norm_mean, np.mean(cds1, axis=0))
         use_cds_to_update = True
-        pv_means = np.hstack([norm_mean, cd1_mean, cd2_mean])
 
     assert not np.any(np.isnan(pv_means)), "NaN in pose vector mean"
     return pv_means, use_cds_to_update
--- a/src/tbp/monty/frameworks/utils/object_model_utils.py
+++ b/src/tbp/monty/frameworks/utils/object_model_utils.py
@@ -11,13 +11,17 @@
 import logging
 
 import numpy as np
+import numpy.typing as npt
 import torch
 
 from tbp.monty.frameworks.utils.spatial_arithmetics import (
+    TangentFrame,
     get_angle,
     get_right_hand_angle,
     normalize,
+    project_onto_tangent_plane,
 )
+from tbp.monty.math import DEFAULT_TOLERANCE
 
 logger = logging.getLogger(__name__)
 
@@ -310,6 +314,88 @@ def get_cubic_patches(arr_shape, centers, size):
     mask = np.any((new_centers < 0) | (new_centers >= np.array(arr_shape[:3])), axis=-1)
 
     return new_centers, mask
+
+
+def orthonormal_pose_vectors(
+    surface_normal: npt.NDArray[np.float64],
+    curvature_direction: npt.NDArray[np.float64],
+) -> npt.NDArray[np.float64]:
+    """Build a right-handed orthonormal pose from two rough directions.
+
+    The surface normal is used as a give, up to normalization. The curvature direction
+    is orthogonlized agaist it, so the returned pose vectors are always a valid
+    rotation matrix.
+
+    Args:
+        surface_normal: Surface normal direction. Does not need to be unit length.
+        curvature_direction: First curvature direction. Does not need to be unit length
+            or orthogonal to the surface normal.
+
+    Returns:
+        Flat array of nine elements holding the surface normal and the two curvature
+        directions.
+    """
+    normal = normalize(surface_normal)
+    tangent = project_onto_tangent_plane(curvature_direction, normal)
+    if np.linalg.norm(tangent) < DEFAULT_TOLERANCE:
+        # The curvature direction holds no information perpendicular to the surface
+        # normal, so fall back to an arbitrary direction in the tangent plane. The
+        # curvature directions are not used for matching in this case anyway.
+        tangent = TangentFrame(normal).basis_u
+    cd1 = normalize(tangent)
+    return np.hstack([normal, cd1, np.cross(normal, cd1)])
+
+
+def pose_vector_merge(
+    new_pose_vecs: npt.NDArray[np.float64],
+    previous_pose_vecs: npt.NDArray[np.float64],
+    use_cds_to_update: bool,
+    num_new_obs: int,
+    num_previous_obs: int,
+) -> npt.NDArray[np.float64]:
+    """Merge newly observed pose vectors into previous ones.
+
+    Surface normals observed from opposite sides of a surface are not two estimates
+    of the same direction, so they are not averaged. The side with more observations
+    is kept, the same way `pose_vector_mean` discards the minority side withing a
+    single batch of observations.
+
+    Curvature directions are ambiguous in direction, so the new one is flipped to agree
+    with the stored one before they are averaged.
+
+    Note that the returned pose vectors could be weighted by the number of observations,
+    but this is currently not implemented.
+
+    Args:
+        new_pose_vecs: Flat array of nine elements holding the pose vectors averaged
+            over the new observations.
+        previous_pose_vecs: Flat array of nine elements holding the averaged previous
+            pose vectors.
+        use_cds_to_update: Whether the new curvature directions are meaningful.
+        num_new_obs: Number of new observations.
+        num_previous_obs: Number of previous observations.
+
+    Returns:
+        Flat array of nine elements holding the merged pose vectors.
+    """
+    new_normal = new_pose_vecs[:3]
+    previous_normal = previous_pose_vecs[:3]
+    if np.dot(new_normal, previous_normal) < 0:
+        majority = (
+            new_pose_vecs if num_new_obs > num_previous_obs else previous_pose_vecs
+        )
+        return orthonormal_pose_vectors(majority[:3], majority[3:6])
+
+    if use_cds_to_update:
+        new_cd1 = new_pose_vecs[3:6]
+        previous_cd1 = previous_pose_vecs[3:6]
+        if np.dot(new_cd1, previous_cd1) < 0:
+            new_cd1 = -new_cd1
+        curvature_direction = new_cd1 + previous_cd1
+    else:
+        curvature_direction = previous_pose_vecs[3:6]
+
+    return orthonormal_pose_vectors(new_normal + previous_normal, curvature_direction)
 
 
 def pose_vector_mean(pose_vecs, pose_fully_defined):
--- a/src/tbp/monty/frameworks/utils/object_model_utils.py
+++ b/src/tbp/monty/frameworks/utils/object_model_utils.py
@@ -358,7 +358,7 @@ def pose_vector_merge(
     Surface normals observed from opposite sides of a surface are not two estimates
     of the same direction, so they are not averaged. The side with more observations
     is kept, the same way `pose_vector_mean` discards the minority side within a
-    single batch of observations.
+    single episode of observations.
 
     Curvature directions are ambiguous in direction, so the new one is flipped to agree
     with the stored one before they are averaged.
--- a/src/tbp/monty/frameworks/utils/object_model_utils.py
+++ b/src/tbp/monty/frameworks/utils/object_model_utils.py
@@ -21,6 +21,7 @@ from tbp.monty.frameworks.utils.spatial_arithmetics import (
     normalize,
     project_onto_tangent_plane,
 )
+from tbp.monty.geometry import Rotation
 from tbp.monty.math import DEFAULT_TOLERANCE
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


src/tbp/monty/frameworks/models/object_model.py
```diff
@@ -841,18 +842,27 @@ def _get_new_voxel_features(
...
+                        previous_average
+                        if avg_feat is None
+                        else pose_vector_merge(
+                            avg_feat,
+                            previous_average,
+                            use_cds_to_update,
+                            num_new_obs,
+                            num_old_obs,
+                        )
+                    )
```
jeremyshoemaker: I'm not totally opposed to this construction, but I think it would be clearer to most people if this was a normal `if/else` block. The way it is now somewhat buries the branching here.

```suggestion
                    if avg_feat is None:
                        avg_feat = previous_average
                    else:
                        avg_feat = pose_vector_merge(
                            avg_feat,
                            previous_average,
                            use_cds_to_update,
                            num_new_obs,
                            num_old_obs,
                        )
```

Note: I haven't checked this change for style or correctness, since I wrote it in the browser.

tests/unit/frameworks/utils/object_model_utils_test.py
```diff
@@ -0,0 +1,186 @@
...
+        In other words, that the surface normal and the two principal curvatures form an
+        orthonormal, right-handed basis.
+
+        Args:
+            pose_vecs: Flat array of nine elements holding the pose vectors.
+            msg: Assertion message.
+        """
+        matrix = np.array(pose_vecs).reshape((3, 3))
+        self.assertAlmostEqual(np.linalg.det(matrix), 1.0, places=6, msg=msg)
+        np.testing.assert_allclose(matrix @ matrix.T, np.eye(3), atol=1e-6, err_msg=msg)
```
jeremyshoemaker: I think in other PRs, @hlee9212 and I decided to go with `np.identity` instead of `np.eye`. Though as I look at the docs again now, there does seem to be _some_ difference between those two functions, with `eye` having more options that we aren't using here.

I personally don't like the name `np.eye` because it's not naming the thing it is. Instead it's playing a cute game with how "eye" and "I" are homonyms. While `np.identity` is longer, it's also clearer.

```suggestion
        np.testing.assert_allclose(matrix @ matrix.T, np.identity(3), atol=1e-6, err_msg=msg)
```

tests/unit/frameworks/utils/object_model_utils_test.py
```diff
@@ -0,0 +1,186 @@
...
+        pose_vecs = orthonormal_pose_vectors(self.normal, np.array([1.0, 0.0, 0.7]))
+        self.assert_is_rotation(pose_vecs, "Pose vectors are not a rotation")
+        np.testing.assert_allclose(pose_vecs[:3], self.normal, atol=DEFAULT_TOLERANCE)
+
+    def test_orthonormal_pose_vectors_falls_back_to_arbitrary_direction_if_curvature_direction_is_parallel_to_surface_normal(  # noqa: E501
+        self,
+    ) -> None:
+        pose_vecs = orthonormal_pose_vectors(self.normal, self.normal * 2.0)
+        self.assert_is_rotation(
+            pose_vecs, "Degenerate curvature direction did not fall back"
```
jeremyshoemaker: After I was using this term with you, I looked up what "degenerate" means for matrices, and it has a very specific meaning which might not cover this.

tests/unit/frameworks/utils/object_model_utils_test.py
```diff
@@ -0,0 +1,186 @@
...
+            observations = np.array(
+                [
+                    pose_frame(self.normal + rng.normal(0, 0.3, 3), rng.normal(size=3))
+                    for _ in range(6)
+                ]
+            )
+            pv_mean, _ = pose_vector_mean(observations, np.ones((6, 1)))
+            self.assert_is_rotation(
+                pv_mean, "Mean of spread observations is not a rotation"
+            )
```
jeremyshoemaker, line 106: Unless I'm missing something, this test looks like it's re-implementing what `hypothesis` gives us. It might be simpler to use that instead.

jeremyshoemaker (approved):
This _seems_ reasonable to me, but I can't speak to the math details or the object model parts of it, so my approval will need to get seconded by someone who is more familiar with this than I.

tests/unit/frameworks/utils/object_model_utils_test.py
tristanls-tbp: I hear you, but this is describing a vector, not a matrix. Nevertheless, I'll change it to "Parallel" since that's what the test is actually testing.

tests/unit/frameworks/utils/object_model_utils_test.py
tristanls-tbp: It's definitely not simpler 😄... but I'm looking into it.

jeremyshoemaker: Well, I did say it _might_ be simpler. 😁

tristanls-tbp: So far I got this to replace the `observations = np.array(...)`

```python
@st.composite
def observations_batch(
    draw: st.DrawFn, normal: npt.NDArray[np.float64], batch_size: int = 6
) -> npt.NDArray[np.float64]:
    observations = np.array(batch_size)
    for i in range(batch_size):
        perturbation = draw(
            hnp.arrays(
                dtype=np.float64,
                shape=(3),
                elements=st.floats(min_value=-1.0, max_value=1.0),
            )
        )
        tangent_seed = draw(
            hnp.arrays(
                dtype=np.float64,
                shape=(3),
                elements=st.floats(min_value=-2.0, max_value=2.0),
            )
        )
        observations[i] = pose_frame(normal + perturbation, tangent_seed)
    return observations
```

jeremyshoemaker: If it's too complicated, don't worry about it. The test needs to be easy to read, and if it makes it more complicated, then it's not really helping.

tristanls-tbp: I think it'll be ok, but it's gonna take some time. I'll invest more in this once we have confirmation that the fix actually works 😬.

src/tbp/monty/frameworks/utils/object_model_utils.py
```diff
@@ -312,6 +316,88 @@ def get_cubic_patches(arr_shape, centers, size):
...
+
+    Returns:
+        Flat array of nine elements holding the merged pose vectors.
+    """
+    new_normal = new_pose_vecs[:3]
+    previous_normal = previous_pose_vecs[:3]
+    if np.dot(new_normal, previous_normal) < 0:
+        majority = (
+            new_pose_vecs if num_new_obs > num_previous_obs else previous_pose_vecs
+        )
```
hlee9212, line 386: **issue (non-blocking)**: A bit confused by this based on the docstring... maybe the "withing a single batch of observations" in docstring is doing the heavy lifting or I'm not understanding properly... (probably latter). 

Let's say we have theses two batches of observations: 
- New: 6 that support direction in +A and 4 that supports direction in -A.
- Previous: 8 that support in direction of -A. 

But `num_new_obs` is 10 compared to 8 in `num_previous_obs`, so the majority will be using the `new_pose_vecs` which only has 6 observations, even though there are more observations for the other side? Do we actually want to track the support count for each side separately? 

(I assume this was done previously and we are trying to preserve behavior?)

src/tbp/monty/frameworks/utils/object_model_utils.py
```diff
@@ -312,6 +316,88 @@ def get_cubic_patches(arr_shape, centers, size):
...
+    if use_cds_to_update:
+        new_cd1 = new_pose_vecs[3:6]
+        previous_cd1 = previous_pose_vecs[3:6]
+        if np.dot(new_cd1, previous_cd1) < 0:
+            new_cd1 = -new_cd1
+        curvature_direction = new_cd1 + previous_cd1
+    else:
+        curvature_direction = previous_pose_vecs[3:6]
+
+    return orthonormal_pose_vectors(new_normal + previous_normal, curvature_direction)
```
hlee9212, line 398: **note (non-blocking)**: I know the doc says "pose vectors could be weighted by the number of observations, but this is currently not implemented" but if anyone is implementing in the future:

```python
surface_normal = num_new_obs * new_normal + num_previous_obs * previous_normal
return orthonormal_pose_vectors(surface_normal, curvature_direction)
```

src/tbp/monty/frameworks/utils/object_model_utils.py
```diff
@@ -312,6 +316,88 @@ def get_cubic_patches(arr_shape, centers, size):
...
+    previous_pose_vecs: npt.NDArray[np.float64],
+    use_cds_to_update: bool,
+    num_new_obs: int,
+    num_previous_obs: int,
+) -> npt.NDArray[np.float64]:
+    """Merge newly observed pose vectors into previous ones.
+
+    Surface normals observed from opposite sides of a surface are not two estimates
+    of the same direction, so they are not averaged. The side with more observations
+    is kept, the same way `pose_vector_mean` discards the minority side withing a
```
hlee9212: **typo**: withing --> within

src/tbp/monty/frameworks/utils/object_model_utils.py
```diff
@@ -312,6 +316,88 @@ def get_cubic_patches(arr_shape, centers, size):
...
 
 
+def orthonormal_pose_vectors(
+    surface_normal: npt.NDArray[np.float64],
+    curvature_direction: npt.NDArray[np.float64],
+) -> npt.NDArray[np.float64]:
+    """Build a right-handed orthonormal pose from two rough directions.
+
+    The surface normal is used as a give, up to normalization. The curvature direction
+    is orthogonlized agaist it, so the returned pose vectors are always a valid
```
hlee9212: **typo**: orthogonlized --> orthogonalized 
**typo**: agaist --> against

src/tbp/monty/frameworks/utils/object_model_utils.py
```diff
@@ -312,6 +316,88 @@ def get_cubic_patches(arr_shape, centers, size):
     return new_centers, mask
 
 
+def orthonormal_pose_vectors(
+    surface_normal: npt.NDArray[np.float64],
+    curvature_direction: npt.NDArray[np.float64],
+) -> npt.NDArray[np.float64]:
+    """Build a right-handed orthonormal pose from two rough directions.
+
+    The surface normal is used as a give, up to normalization. The curvature direction
```
hlee9212: **typo**: give --> given

src/tbp/monty/frameworks/utils/object_model_utils.py
```diff
@@ -312,6 +316,88 @@ def get_cubic_patches(arr_shape, centers, size):
...
+    Returns:
+        Flat array of nine elements holding the merged pose vectors.
+    """
+    new_normal = new_pose_vecs[:3]
+    previous_normal = previous_pose_vecs[:3]
+    if np.dot(new_normal, previous_normal) < 0:
+        majority = (
+            new_pose_vecs if num_new_obs > num_previous_obs else previous_pose_vecs
+        )
+        return orthonormal_pose_vectors(majority[:3], majority[3:6])
```
hlee9212, line 387: **issue**: I think this return bypasses `use_cds_to_update`. If the new batch wins by count but its curvature directions are undefined, then this stores the new batch's arbitrary curvature direction. I think we need to select the "winning" normal here but retain previous curvature direction when `use_cds_to_update` is false.

tests/unit/frameworks/utils/object_model_utils_test.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Args:
+            pose_vecs: Flat array of nine elements holding the pose vectors.
+            msg: Assertion message.
+        """
+        matrix = np.array(pose_vecs).reshape((3, 3))
+        self.assertAlmostEqual(np.linalg.det(matrix), 1.0, places=6, msg=msg)
+        np.testing.assert_allclose(
+            matrix @ matrix.T, np.identity(3), atol=1e-6, err_msg=msg
+        )
```
hlee9212: **nitpick**: Unless there is a specific reason not to use `DEFAULT_TOLERANCE`, it would be nice to make this consistent:

```python
np.testing.assert_allclose(
    matrix @ matrix.T,
    np.identity(3),
    rtol=0,
    atol=DEFAULT_TOLERANCE,
   err_msg=msg,
)
np.testing.assert_allclose(
    np.linalg.det(matrix),
    1.0,
    rtol=0,
    atol=DEFAULT_TOLERANCE,
    err_msg=msg,
)
```

hlee9212 (changes requested):

hlee9212:
Hi @tristanls-tbp, overall the math in `orthogonal_pose_vectors()` is correct (i.e. Gram-Schmidt will give us valid Rotation matrices). I think I only had one blocking issue, the rest are just minor notes or typos. The overall filtering/grouping logic was harder to understand than math IMHO 😂, but I can somewhat see why we did what we did over just calling [scipy.spatial.transforms.Rotation.mean()](https://docs.scipy.org/doc/scipy-1.18.0/reference/generated/scipy.spatial.transform.Rotation.mean.html#scipy.spatial.transform.Rotation.mean), like:

```python
merged = Rotation.from_matrix(
    np.stack([new_matrix, previous_matrix])
).mean(
    weights=[num_new_obs, num_previous_obs]
)
```

(In my mind the fix was just using the above line after some filtering because the method will automatically give a valid Rotation matrix with the desired `weights` mechanism built-in, but I'm sure there was a good explanation back when this was implemented / we wanted to take the surface normal more priority than treating all axes equally.)

jeremyshoemaker:
I hacked something together — using a library called `histoprint` — to get an idea of the distribution of pose determinants to see if this actually fixes things. I ran it under Conda/Habitat to:
- verify that we were always getting negative determinants and just not failing on them because of that version of SciPy was being more lenient
- confirm that this change fixes the distribution

These are samples from both runs of `surf_agent_unsupervised_10simobj`:

### Without the fix
```
                 Histogram of hypothesis pose determinants (log scale)
 -1.000 _                                                                  
 -0.867 _══════
 -0.733 _
 -0.600 _
 -0.467 _
 -0.333 _
 -0.200 _
 -0.067 _
  0.067 _═════════════════════════════════════════
  0.200 _════════════════════════
  0.333 _═════════════════
  0.467 _═════════════════
  0.600 _════════════════════
  0.733 _══════════════════════════
  0.867 _═══════════════════════════════════
  1.000 _═══════════════════════════════════════════════════════════════════════════════
```


### With the fix
```
                 Histogram of hypothesis pose determinants (log scale)
 -1.000 _ 
 -0.867 _
 -0.733 _
 -0.600 _
 -0.467 _
 -0.333 _
 -0.200 _
 -0.067 _
  0.067 _
  0.200 _
  0.333 _
  0.467 _
  0.600 _
  0.733 _
  0.867 _
  1.000 _═══════════════════════════════════════════════════════════════════════════════
```

These samples are not necessary from the same part of the run, but in the first case, we very quickly get values all across the range from 0 to 1, and then get a few negative values. I'm guessing there are probably quite a few around 0 that we don't really see here.

In the second case, we get this graph through the whole run without change, so I think we can safely say this fixes that problem.

Whether we're fixing it correctly or not is another question.

---

P.S. Those em-dashes are mine. I'm taking them back.

tests/unit/frameworks/utils/object_model_utils_test.py
tristanls-tbp: No specific reason, missed this one, thank you.

src/tbp/monty/frameworks/utils/object_model_utils.py
tristanls-tbp: As far as I can tell, your description describes what the algorithm does.

If the new average direction (+A, with +A\*6 and -A\*4 data) is different from the previous average direction (-A, with -A\*8 data). We will go with +A.

For what it's worth, the "batch" is an entire episode's worth of buffer. Update happens in `_get_new_voxel_features`, called by `_update_grids`, called by `update_model`, called by `_extend_graph`, called by `update_memory`, called by `_update_memory`, called by `update_ltm_from_stm`:
```python
    def update_ltm_from_stm(self) -> None:
        """If training, update memory from buffer."""
        if self.mode is ExperimentMode.TRAIN and len(self.buffer) > 0:
            logger.info(f"\n---Updating memory of {self.learning_module_id}---")
            self._update_memory()
```

So, in practice, if I understand correctly, we update the model based on all observations in the episode.

tristanls-tbp: I'll update the docstring to state "within a single episode of observations." to at least clarify that part.

src/tbp/monty/frameworks/utils/object_model_utils.py
tristanls-tbp: Since this is an early return for the case where surface normals point in opposite directions, I'm not sure I understand why we should assume that curvature directions coupled to the "wrong" direction would happen to be correct for the new "correct" direction?

src/tbp/monty/frameworks/models/object_model.py
```diff
@@ -843,16 +844,24 @@ def _get_new_voxel_features(
...
+                else:
+                    if feature == "object_id" and avg_feat != previous_average:
+                        # TODO: Figure out a more nuanced way to take into account
+                        # past obs
+                        if num_old_obs > num_new_obs:
+                            avg_feat = previous_average
+                        else:
+                            previous_average = avg_feat
+                    # NOTE: could weight these
+                    avg_feat = (avg_feat + previous_average) / 2
```
nielsleadholm, line 864: You didn't introduce this, but I'm just noticing that this line is a no-op in the case that `feature == "object_id" `, as that proceeding block will always set `avg_feat == previous_average`. Instead this line could follow an `else`.

nielsleadholm (commented):
Overall this looks like a reasonable and correct fix, my main, high-level comment is whether we could do something simpler. Conceptually, this would be to tile the `previous_average` by `num_old_obs`, stack this with the new `pose_vecs`, and then pass the whole thing to the existing `pose_vector_mean`. We would do the same treatment for `pdefined`. This would avoid introducing a new function (`pose_vector_merge`), and it should operate as a weighted averaging, which is what we want.

If we're worried about the computational overhead of the tiling, I think a small tweak to `pose_vector_mean` with weights would keep the same results but avoid this. 

This would still keep the fixes from `pv_means = orthonormal_pose_vectors` within `pose_vector_mean`.

src/tbp/monty/frameworks/utils/object_model_utils.py
hlee9212: Thanks for the explanation. I guess I'm wondering if this is more "is this the way we want to do it?" but I also think it is out of scope for this PR to go into that.

src/tbp/monty/frameworks/utils/object_model_utils.py
hlee9212: That's fair. Perhaps instead of retaining previous curvature, we could try to ensure that the merged node remains `pose_fully_defined=False` so the arbitrary curvature direction is not used for matching if `use_cds_to_update` is false? 

I assume "merging" of previously True value with a new False value would store 0.5 (which would be True in `bool`), which may contribute to allow arbitrary curvature direction to contribute to matching. 

(Apologies if this is already being done and I didn't catch it.)

hlee9212:
This is just a response to @jeremyshoemaker about:

> Whether we're fixing it correctly or not is another question.

I spent some time thinking about this. The way we are doing now at a high-level is:
- Surface normal gets priority
- cd1 adapts to normal
- cd2 follows from cross product

It's a bit different from `scipy.Rotation.mean()` which does a little bit different optimization where surface normal, cd1, and cd2 are treated with equal importance. Our method is valid, albeit a bit unconventional and bespoke - but prioritizing surface normal is justifiable. 

One example where our method vs. `Rotation.mean()` may differ:
- If we have R1 = Identity matrix and R2 = 90 rotated around arbitrary axis, then `Rotation.mean()` will produce 45 deg (i.e. what someone may assume what "mean" between 0 and 90 deg is), but our method will not (but it will be close to 45 deg). I think this is okay, as long as it was intended and clearly justified/documented. 

All in all, my guess is that it's probably likely that we need to do something similar to what Ramy & Dale are doing with his current IP to "clean up" terminal state but equivalent to all instances related to pose in Monty. For example, I found using the flattened (9,) vector representation of pose here inconsistent with the np.vstack-ed `pose_vectors` in CMP (and also tbp.Rotation), but I think it is out of scope and didn't comment on this previously.

src/tbp/monty/frameworks/utils/object_model_utils.py
ramyamounir: > I'm not sure I understand why we should assume that curvature directions coupled to the "wrong" direction would happen to be correct for the new "correct" direction?

The curvature direction is coupled to the tangent plane not the surface normal direction, and this tangent plane is the same for both surface normal directions (`n` and `-n` in the figure below). So the stored cd1 is still correct when the normal flips; cd2 is the only thing that changes, and Gram-Schmidt recomputes it anyway. Even if the dot product between the old surface normal and the new surface normal is not exactly `-1` (resulting in different tangent planes), the projection extracts whatever component of the stored cd1 lies in the winning normal's tangent plane.

I agree with @hlee9212's raised issue, but maybe I'm missing something.

Image

hlee9212: @ramyamounir I'm a bit confused by the diagram because it feels like $n$ and $-n$ should have the same cd1 (so I'm not sure why it says new cd1: undefined, arbitrary in the plane) 😅

Anyway, this was the original fix I was thinking of:
```python
if np.dot(new_normal, previous_normal) < 0:
    majority = (
        new_pose_vecs
        if num_new_obs > num_previous_obs
        else previous_pose_vecs
    )
    curvature_direction = (
        majority[3:6]
        if use_cds_to_update
        else previous_pose_vecs[3:6]
    )
    return orthonormal_pose_vectors(
        majority[:3],
        curvature_direction,
    )
```
But I guess now I'm more concerned about whether `pose_fully_defined` is merged correctly because if we do:
```python
avg_feat = (avg_feat + previous_average) / 2
```
For a boolean then this could do something like (False + True) / 2 = 0.5 (which becomes True converted back to boolean). 

But I'm beginning to think this is another PR. We probably haven't been merging `pose_fully_defined` properly, it just that it hasn't been raised as an issue.

ramyamounir: > I'm a bit confused by the diagram because it feels like `n` and `−n` should have the same cd1 (so I'm not sure why it says new cd1: undefined, arbitrary in the plane)

Yes, the point of the diagram is to argue that `n` and `-n` should have the same `cd1`. The issue is that if the new observations win by majority and they happen to have undefined `cd1`, then we end up replacing the good `cd1` with some arbitrary/undefined `cd1`. It's one example of what could happen, I should have explained it better. But also even if the new observations have defined `cd1`, we would still average (or flip the sign and average) with the stored `cd1`. Either way, I don't think we should replace the stored `cd1`.

> I'm more concerned about whether `pose_fully_defined` is merged correctly

I agree. I think updating `cd1` may have issues with keeping `pose_fully_defined` in sync. I'm not really sure if this is pre-existing or in-scope here. Ideally, we should have test coverage for this.

tests/unit/frameworks/utils/object_model_utils_test.py
```diff
@@ -0,0 +1,194 @@
...
+        tangent_seed: Direction to derive the first curvature direction from.
+
+    Returns:
+        Flat array of nine elements holding the surface normal and the two curvature
+        directions.
+    """
+    normal = normalize(normal)
+    tangent = project_onto_tangent_plane(tangent_seed, normal)
+    cd1 = normalize(tangent)
+    return np.hstack([normal, cd1, np.cross(normal, cd1)])
```
ramyamounir: This looks like it's re-implementing to the `orthonormal_pose_vectors` function. Just for my understanding, why aren't we importing and using the function we already have?

ramyamounir (changes requested):
Nice fix! I am not an expert in this part of the code, but I did leave a couple of comments on an issue @hlee9212 raised and a function in one of the test files. Otherwise, LGTM.

tristanls-tbp:
After sitting with this PR and another discussion with @jeremyshoemaker, I want to raise another issue for clarification.

It seems to me that the average calculation and the problems it introduces stem from the fact that we have multiple observations within a single voxel and are unwilling to forgo extracting pose vectors from it.

Some illustrative examples I have in mind:

_A sphere fully encapsulated within a voxel_

Screenshot 2026-09-08 at 13 33 44

If we gather an episode's worth of data for this voxel, it seems to me that the voxel should not have pose vectors associated with it. The average of all pose vectors should cancel out across the entire voxel.

_A plane fully encapsulated within a voxel_

Screenshot 2026-09-08 at 13 35 31

If we gather an episode's worth of data for this voxel, again, it seems to me that the voxel should not have pose vectors associated with it. Granted, the distribution of surface normals in this example would be bimodal, but for the entire voxel, it seems to me that the average of all surface normals (and curvatures) should cancel out.

_A volume with one surface uniformly intersecting a voxel_

Screenshot 2026-09-08 at 13 46 27

This is an easy one; here, the average of the surface normals/pose vectors seems meaningful.

_A volume with one surface intersecting a voxel in a way leading to different pose vector readings_

Screenshot 2026-09-08 at 13 48 04

This is not as straightforward as the previous example, but averaging the pose vectors here also seems meaningful to me. It captures something true about the voxel.

_A plane fully bisecting a voxel_

Screenshot 2026-09-08 at 13 49 43

The bimodal distribution of surface normals here appears to provide some true information about the voxel. Picking an arbitrary side does not lose all true information, but it does misrepresent the voxel in a way. On the other hand, anything like an average would cancel out for the voxel as a whole. So, a question in this case would be: is this voxel equivalent to the fully encapsulated sphere, or the fully encapsulated plane, or somehow different?

If we decide that a plane fully bisecting a voxel carries different information than a fully encapsulated plane, then how would we tell those measurements apart?

This suggests an "easy" (but perhaps degenerate) proposal: don't tell them apart, and have them cancel out (i.e., don't determine pose vectors for voxel) rather than pick a side.

In summary, must we try this hard to extract pose vectors for every voxel, or perhaps we should more readily declare a lack of pose and curvature when faced with contradictory data?

tests/unit/frameworks/utils/object_model_utils_test.py
tristanls-tbp: Good catch, thanks. No meaningful reason; it is just my unfamiliarity with the formula that prevents recognition of the obvious substitution.

src/tbp/monty/frameworks/utils/object_model_utils.py
tristanls-tbp: I don't believe curvatures remain the same if the object is nonplanar. For example, a blob:

Screenshot 2026-09-08 at 17 03 02

src/tbp/monty/frameworks/models/object_model.py
tristanls-tbp: Thank you, good find. I also removed the extraneous check `avg_feat != previous_average` as that branch arm is also a no-op if `avg_feat == previous_average`. This improves the overall readability, too.

```python
                if feature == "pose_vectors":
                    if avg_feat is None:
                        avg_feat = previous_average
                    else:
                        avg_feat = pose_vector_merge(
                            avg_feat,
                            previous_average,
                            use_cds_to_update,
                            num_new_obs,
                            num_old_obs,
                        )
                elif feature == "object_id":
                    # TODO: Figure out a more nuanced way to take into account
                    # past obs
                    if num_old_obs > num_new_obs:
                        avg_feat = previous_average
                    else:
                        previous_average = avg_feat
                else:
                    # NOTE: could weight these
                    avg_feat = (avg_feat + previous_average) / 2
```

tests/unit/frameworks/utils/object_model_utils_test.py
tristanls-tbp: Deleted and using `orthonormal_pose_vectors` as of https://github.com/thousandbrainsproject/tbp.monty/pull/1174/commits/34e440a35d6a71eb994ebe8d5605263b8cee55a8

src/tbp/monty/frameworks/utils/object_model_utils.py
ramyamounir: If my understanding is correct, this logic was introduced to handle a thin surface passing through the voxel. In this case, you get the two opposing surface normal directions and they could cancel each other if we didn't do the majority trick. But this whole logic doesn't work if the voxel size is too big (as in the figure you shared). Not only does the `cd1` change, but we can't use majority to get the surface normal anymore. My thinking is that this whole logic is built on the assumption that the voxel is correctly sized yet we still get a negative dot product, which means that this must be because of the thin surface with two faces passing through a single voxel.

The problem you shared is a different problem that relates more to the problem of how to scale the voxels, similar to the other problem of how to size the attention voxels dynamically (the one you identified in Rome).

tristanls-tbp:
In today's (September 9, 2026) research meeting, we highlighted the open problem of how to correctly merge pose vectors inside a voxel. For the purpose of this pull request, we settled on using `Rotation.mean()`, which is weighted and ensures orthonormality of the result. It is no worse than the bug `(next + prev) / 2`, and it leaves open the question of what we should be doing in the long term. 

[Latest commits](https://github.com/thousandbrainsproject/tbp.monty/pull/1174/changes/a2197275db6e724671bb145e0aa1e7253a6dfe67..238ea440658cb92a39c83e83fc27670e63d92a0b) implement our decision from the research meeting.

Please take another look @hlee9212 @nielsleadholm @ramyamounir .

**question**: Should I retain `use_cds_to_update` and do this inside `pose_vector_merge`?:
```python
    if not use_cds_to_update:
        new_pose_vecs[3:] = previous_pose_vecs[3:]
    return (
        Rotation.from_matrix(
            np.stack([new_pose_vecs.reshape(3, 3), previous_pose_vecs.reshape(3, 3)])
        )
        .mean(weights=[num_new_obs, num_previous_obs])
        .as_matrix()
        .flatten()
    )
```

hlee9212 (approved):
LGTM. 

@tristanls-tbp Regarding your question about retaining `use_cds_to_update`: I'd say let's hold off for now. Copying previous curvature directions may likely make new_pose_vecs not be orthonormal (and can cause the determinant to be less than 0), and I don't want to shove that into `Rotation.from_matrix()`. Just my preference.

nielsleadholm (commented):
Nice, thanks for the update and for sparking the discussion on this topic. 

Overall LGTM, my main comment now is that this will likely change the learned compositional models (which use `GridObjectModel`), so I would suggest re-training those to check the impact on benchmark results, and any necessary update there.

I'm not sure if we want to consider it in scope of this PR, but @hlee9212 also pointed out that we may be facing a similar issue with merging `pose_fully_defined` ("For a boolean then this could do something like (False + True) / 2 = 0.5 (which becomes True converted back to boolean)."). It's probably a one-line change + a unit test though, so maybe worth fixing here? 

Re. your question, I agree with Hojae. For what it's worth, `pose_fully_defined` should generally be False at these object locations regardless, so it shouldn't really matter what we store for the curvature directions. This approach reduces making the code complex and hard to follow.

ramyamounir (commented):
Great, thanks @tristanls-tbp. I agree with @hlee9212 that surgically replacing the `cd1` like that without re-applying Gram-Schmidt may result in a rotation with `<1` determinant. I think `Rotation.mean` expects two proper rotations. 

I think the only blocking issue is retraining the models for LMs that use `GridObjectModel`, as Niels mentioned.

tristanls-tbp:
@ramyamounir @nielsleadholm, regarding blocking this PR on retraining compositional models and then running benchmarks to detect changes in the trained models. @hlee9212 investigated this and determined there are no changes between `main` and this branch. However, there _are_ discrepancies between `main` and the benchmarks stored in the CSVs. Per @hlee9212 , this discrepancy was introduced in #1180.

As such, I see no reason to block this update further. I assume we're not going to revert #1180, right? Also, updating the CSVs seems like it should have been done in the scope of #1180 (which introduced the changes), so remediation of that also does not seem in scope nor a blocker to this PR.

I am asking for a re-review and to keep in mind that we will be retraining at least two more times shortly. I propose we merge, then do the CSV benchmark update work on `main` (a PR that only updates CSVs). And move forward from here. 

The alternative is to wait until benchmarks are updated on `main`, and then redo training and benchmarks for this PR. However, it is my understanding that we are not going to do this because #1194 will retrain and redo the benchmarks again. If that's the case, we should merge this.

ramyamounir (approved):
Thanks @tristanls-tbp and @hlee9212. I agree, this can be merged and we can do a follow-up retraining of models and benchmark updates after.

tristanls-tbp:
Thanks @ramyamounir.

Since @nielsleadholm is currently unavailable, I'm removing him from the assignees so that I can merge this PR to unblock work tomorrow. I am assuming he'd be ok with this course of action. If I'm wrong, since we aren't making any benchmark number updates with this PR, it should be straightforward to revert.

PR #1179, opened 2026-09-03
refactor!: remove unused `max_total_steps` from monty args

ramyamounir:

## Breaking Changes
- Removed `max_total_steps` from `MontyBase` constructor and configs

Note that we have another `max_total_steps` under the experiment configs, that one is used. The duplicate in the Monty classes is unused and is removed in this PR.


hlee9212 (approved):
LGTM.

PR #1182, opened 2026-09-04
refactor!: update monty args filenames to remove `_tot2500` suffix

ramyamounir:

## Breaking Changes
- Update monty args config filenames to remove the `_tot*` suffix

Note that this is a follow-up to [PR#1179](https://github.com/thousandbrainsproject/tbp.monty/pull/1179) where the unused `max_total_steps` arg was removed from Monty classes.


hlee9212 (approved):
LGTM.

PR #1186, opened 2026-09-09
refactor!: make `RecognitionCounter` properties read-only and provide defaults

dschumacher-tbp:

## Breaking Changes
- `RecognitionCounter` is no longer a `dataclass`, so initialization is easier.


jeremyshoemaker (commented):
You don't have to throw out the baby with the bathwater. You can provide a custom constructor for a `dataclass` and keep the rest of the machinery if you want to.

dschumacher-tbp:
Can a `dataclass` define read-only properties?

jeremyshoemaker:
> Can a `dataclass` define read-only properties?

You can pass `frozen=True` to the `@dataclass` decorator, e.g. `@dataclass(frozen=True)` to get that behaviour. Do note that in Python you really can't make things truly immutable, see this [note from the docs](https://docs.python.org/3/library/dataclasses.html#frozen-instances), especially the part about `__init__` needing to use `object.__setattr__` to set values on frozen dataclasses.

I'm not sure how much trying to make it read-only is worth the effort and pain.

dschumacher-tbp:
Ok. We can keep "`dataclass` with custom `__init__`" as an alternative where read-only isn't desired. For our present use-case, I think this implementation is simple and clear.

jeremyshoemaker (approved):

tristanls-tbp:
> - `RecognitionCounter` is no longer a `dataclass`, so initialization is easier.

Could you please explain how initialization is easier? What can you do now that you couldn't do before?

jeremyshoemaker:
> Could you please explain how initialization is easier? What can you do now that you couldn't do before?

The default constructor that `@dataclass` provides doesn't allow for leaving out arguments that have a default, so you always have to specify all of them.

jeremyshoemaker:
Actually, that might not be true.

jeremyshoemaker:
I could have sworn that was the case, but from testing it just now, that doesn't seem to be the case, so I'm not sure why this is needed.

jeremyshoemaker (changes requested):
Unapproving for the moment.

dschumacher-tbp:
I'm running similar tests. I could have sworn Ramy and I ran into a situation where all the parameters had to be provided, but that doesn't seem to be a problem now. If we don't need this, I'll close this PR. The read-only aspect was much less important, but seemed like a good way to communicate our intent.

jeremyshoemaker:
I'm surprised as anyone, and I spent a long time just now looking at the docs trying to determine if there was an option or something to support that. I should have just tested it.

dschumacher-tbp:
After further testing, this PR appear unnecessary, therefore it is being closed with no further action required. The default behavior of `dataclass` covers the main issue (allowing default initializers). The read-only aspect is much less important.

PR #1176, opened 2026-09-03
docs: add future works item on hierarchical percept filtering

ramyamounir:

PR #1189, opened 2026-09-10
ci: update GitHub `actions/*` to latest versions

jeremyshoemaker:
The current versions of the actions we're using are on a deprecated version of Node, and so we've been getting warnings in the action logs about them for some time now.

This is as good a time as any to upgrade to the latest versions if we can.


jeremyshoemaker:
I had a feeling these might fail. I saw the change note about this one in particular and wondered if we might run into it.

> Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.

I'm not sure how I feel about changing it to allow checking it out given the security implications. I'll take a look at the workflow to see if we can adjust it.

jeremyshoemaker:
So, I added exceptions to the `docs_preview` and `future_work_widget_validate` workflows because I _think_ it is safe to do so given how we're using the PR version of the code in both of those actions.

That said, I don't know why it's currently failing. Because those checks are using the workflows from `main` they shouldn't be using the updated versions that require adding an exception. And looking at the output, it looks like they're still using `checkout@v4`, but the output of the run looks like it _is_ using the new version because it fails with the error.

tristanls-tbp:
@jeremyshoemaker could you split out the changes into ones that don't create problems so that those are not held back by a security discussion?

jeremyshoemaker:
Actually, looking back to previous PRs with docs changes, e.g. #1176, #1157, etc., this has been broken for some time. It looks like [they updated `checkout@v4` to backport the security fix](https://github.com/actions/checkout/releases/tag/v4.4.0).

The change I made would fix the problem, if we're okay with it. I took it out and it's still failing.

jeremyshoemaker:
Here's a failing example from #1157 which is a merged PR. https://github.com/thousandbrainsproject/tbp.monty/actions/runs/33169193014

.github/workflows/docs_preview.yml
jeremyshoemaker, line 34: I reverted the change to update this action. That doesn't seem to matter though given that they backported the security fix, but I'm not changing it from what it currently is.

.github/workflows/future_work_widget_validate.yml
jeremyshoemaker, line 60: I reverted the change to update this action. That doesn't seem to matter though given that they backported the security fix, but I'm not changing it from what it currently is.

.github/workflows/monty_api_docs.yml
```diff
@@ -46,21 +46,21 @@ jobs:
...
+        uses: actions/download-artifact@v8
         with:
           name: sphinx-html-${{ github.event.workflow_run.head_sha }}
           run-id: ${{ github.event.workflow_run.id }}
           path: tbp.monty/sphinx_html
           github-token: ${{ github.token }}
       - name: Upload GitHub Pages artifact
         if: ${{ steps.should_run_monty_api_docs.outputs.should_run_monty == 'true'}}
-        uses: actions/upload-pages-artifact@v3
+        uses: actions/download-artifact@v8
```
tristanls-tbp: **issue**: I think this is a typo? This should continue to be `upload-pages-artifact@...`, no?

tristanls-tbp (changes requested):
Thank you for making the updates. These version numbers went up quickly 😬.

I think there's one unintentional change. Otherwise, things look good.

.github/workflows/monty_api_docs.yml
jeremyshoemaker: Yep that's a typo.

jeremyshoemaker:
Fixed the typo, and per our discussion, I'm bumping the `actions/checkout` versions back to `v7` on the ones that fail since GitHub back-ported the security fix to v4 and it's broke there too.

tristanls-tbp (approved):

PR #1191, opened 2026-09-11
fix: remove max_total_steps in expanded compositional configs

hlee9212:
This PR removes obsolete config parameter `max_total_steps`.


ramyamounir (approved):
Thanks, LGTM.

PR #1193, opened 2026-09-11
chore: create MuJoCo versions of training experiments

jeremyshoemaker:
Created MuJoCo versions of training experiments. I ran all of these all locally, and they seem to generate model files correctly. I don't have a good criteria for determining if they're working correctly, but I'm certain the configurations are correct. We won't know for sure until we train some models and run some inference experiments on them.

Diff against the original versions.
```diff
=== diff: only_surf_agent_training_10obj.yaml <-> only_surf_agent_training_10obj_mujoco.yaml ===
-   - /environment: habitat_ycb_surf_agent
+   - /environment: mujoco_ycb_surf_agent

=== diff: only_surf_agent_training_allobj.yaml <-> only_surf_agent_training_allobj_mujoco.yaml ===
-   - /environment: habitat_ycb_surf_agent
+   - /environment: mujoco_ycb_surf_agent

=== diff: only_surf_agent_training_10simobj.yaml <-> only_surf_agent_training_10simobj_mujoco.yaml ===
-   - /environment: habitat_ycb_surf_agent
+   - /environment: mujoco_ycb_surf_agent

=== diff: only_surf_agent_training_numenta_lab_obj.yaml <-> only_surf_agent_training_numenta_lab_obj_mujoco.yaml ===
-   - /environment: habitat_numentalab_surf_agent
+   - /environment: mujoco_numentalab_surf_agent

=== diff: supervised_pre_training_5lms_all_objects.yaml <-> supervised_pre_training_5lms_all_objects_mujoco.yaml ===
-   - /environment: habitat_ycb_dist_agent_sensors5_semantics0
+   - /environment: mujoco_ycb_dist_agent_sensors5

=== diff: supervised_pre_training_5lms.yaml <-> supervised_pre_training_5lms_mujoco.yaml ===
-   - /environment: habitat_ycb_dist_agent_sensors5_semantics0
+   - /environment: mujoco_ycb_dist_agent_sensors5

=== diff: supervised_pre_training_10distinctobj_vocus2.yaml <-> supervised_pre_training_10distinctobj_vocus2_mujoco.yaml ===
-   - /environment: habitat_ycb_dist_agent_semantics0_256
+   - /environment: mujoco_ycb_dist_agent_256

=== diff: supervised_pre_training_objects_with_stickers_comp_models.yaml <-> supervised_pre_training_objects_with_stickers_comp_models_mujoco.yaml ===
-   - /environment: habitat_dist_agent_sensors3
+   - /environment: mujoco_dist_agent_sensors3

=== diff: supervised_pre_training_objects_with_stickers_monolithic_models.yaml <-> supervised_pre_training_objects_with_stickers_monolithic_models_mujoco.yaml ===
-   - /environment: habitat_dist_agent_sensors3
+   - /environment: mujoco_dist_agent_sensors3

=== diff: supervised_pre_training_objects_with_stickers_2d_children.yaml <-> supervised_pre_training_objects_with_stickers_2d_children_mujoco.yaml ===
-   - /environment: habitat_dist_agent_sensors3
+   - /environment: mujoco_dist_agent_sensors3

=== diff: supervised_pre_training_objects_with_stickers_3d_children.yaml <-> supervised_pre_training_objects_with_stickers_3d_children_mujoco.yaml ===
-   - /environment: habitat_dist_agent_sensors3
+   - /environment: mujoco_dist_agent_sensors3
```


src/tbp/monty/conf/experiment/only_surf_agent_training_numenta_lab_obj_mujoco.yaml
```diff
@@ -29,4 +29,4 @@ experiment:
     seed: 42
     supervised_lm_ids: all
     logging:
-      run_name: surf_agent_1lm_numenta_lab_obj_mujoco
+      run_name: surf_agent_1lm_numenta_lab_obj
```
jeremyshoemaker, line 32: I discussed this briefly with @hlee9212 in chat. My plan has been to leave the pretraining models in the same paths so that I didn't have to change the inference experiments to load from different paths.

This makes it easier to switch between running against MuJoCo trained models and Habitat trained ones by just changing the MONTY_MODELS environment variable, and then put the MuJoCo trained models in a separate directory from the Habitat ones, while keeping the layout the same.

When I wrote this training config in #1153, I added the `_mujoco` suffix out of habit, so I'm fixing that here.

tristanls-tbp (commented):
Hey, since #1197 these are now the "old" way of specifying configs. Could you add the `/termination` to `defaults` instead of the inline `match_criterion` and `recognition_policy`?

tristanls-tbp:
@jeremyshoemaker uv don't like you. Oh, it looks like mujoco snapshots need to be updated.

jeremyshoemaker:
> @jeremyshoemaker uv don't like you. Oh, it looks like mujoco snapshots need to be updated.

Yep, that one gets me every time.

tristanls-tbp (approved):
👍

PR #1180, opened 2026-09-03
feat: add expanded compositional pretraining configs

hlee9212:
## Summary of Changes
- Adds four new pretraining configs, `supervised_pre_training_expanded_3d_children`, `_2d_children`, `_comp_models`, and `_monolithic_models`. These are separate from the `supervised_pre_training_objects_with_stickers...` for now; they may stay separated if we keep some inference configs for "small" dataset. 
- Adds `skip_on_object_not_visible` config to enable training 2D stickers of different sizes. 
- Updates how numeric object ID are calculated to prevent collision. 

Note: As mentioned before, links to pretrained models will be added or updated in docs after/during PR for inference.

PR #1194, opened 2026-09-12
feat!: add expanded compositional inference configs

hlee9212:
## Breaking Changes
- Officially name compositional dataset as COWS (Compositional Objects With Stickers).
- Updates models from pretrained_compositional_objects_v6 to pretrained_compositional_objects_v7.
- Updates dataset to compositional_objects_1.4 and compositional_objects_1.4_random_locs.
- Updates benchmarks for COWS_small (existing compositional inference experiments).
- Adds 6 new benchmarks for COWS_large. 

### 1. Performance Change in base `cows_small` experiments

There are two reasons for why performance on these changes:
- #1180 where we updated the numeric value of object IDs.
- Changing dataset from `compositional_objects_1.3` to `compositional_objects_1.4`. 

@nielsleadholm I can dig through old WandB/tbp.benchmarks outputs if you'd like to see the effect of the first reason - I believe the effect was small if I remember correctly. 

### 2. **Why we are skipping `compositional_objects_1.3`**

I originally prepared the PR to update to compositional_objects_1.3, then updated everything again when we decided later to use darker marble objects. I thought the change in objects was significant to warrant a version update and not to delete the existing compositional_objects_1.3 in AWS and EFS where `feat.comp_benefits_figures` might still be using. For users, they probably won't recognize we skipped a version. I figure there might also be a chance we might return to the lighter marble if we are stress testing segmentation algorithm, in which case I'd rather have `compositional_objects_1.3` handy.


docs/overview/benchmark-experiments.md
```diff
@@ -119,111 +119,130 @@ These experiments are currently run without multiprocessing (using `run.py`).
...
+> [!WARNING]
+> These benchmarks track ongoing compositional research and are not currently expected to have good performance.
 
-Note: These experiments use 100 inference steps rather than the typical 500. Monolithic models often time out, making the benchmarks prohibitively slow at the larger step limit.
+#### COWS Small
 
-> [!WARNING]
->
-> These benchmarks are not currently expected to have good performance and are used to track our research progress for compositional datasets.
+`cows_small` contains the original 19 objects, including the five plain objects and objects with TBP or Numenta logo stickers. Its 2D child LM learns the two logos.
```
nielsleadholm: I would avoid referring to "original 19", and just spell it out a bit, as most people coming across this won't have any context about original vs. new.

docs/overview/benchmark-experiments.md
```diff
@@ -119,111 +119,130 @@ These experiments are currently run without multiprocessing (using `run.py`).
...
->
-> These benchmarks are not currently expected to have good performance and are used to track our research progress for compositional datasets.
+`cows_small` contains the original 19 objects, including the five plain objects and objects with TBP or Numenta logo stickers. Its 2D child LM learns the two logos. 
+
+!table[../../benchmarks/cows_small.csv]
 
-Note: To obtain these results, pretraining was run without parallelization across episodes, inference was run with parallelization.
+#### COWS Large
+
+`cows_large` contains 119 objects: the original 19 plus 100 additional compositional objects. Its 2D child LM learns seven stickers: the two logos plus square, circle, triangle, star, and heart.
```
nielsleadholm: Again, original could instead be - the objects in cows_small.

docs/overview/benchmark-experiments.md
```diff
@@ -119,111 +119,130 @@ These experiments are currently run without multiprocessing (using `run.py`).
...
-> | compositional_objects_1.2 | zip | [compositional_objects_1.2.zip](https://tbp-data-public-5e789bd48e75350c.s3.us-east-2.amazonaws.com/tbp.monty/compositional_objects_1.2.zip) |
-> 
+> | compositional_objects_1.4 | tgz | [compositional_objects_1.4.tgz](https://tbp-data-public-5e789bd48e75350c.s3.us-east-2.amazonaws.com/tbp.monty/compositional_objects_1.4.tgz) |
+> | compositional_objects_1.4 | zip | [compositional_objects_1.4.zip](https://tbp-data-public-5e789bd48e75350c.s3.us-east-2.amazonaws.com/tbp.monty/compositional_objects_1.4.zip) |
+> | compositional_objects_1.4_random_locs | tgz | [compositional_objects_1.4_random_locs.tgz](https://tbp-data-public-5e789bd48e75350c.s3.us-east-2.amazonaws.com/tbp.monty/compositional_objects_1.4_random_locs.tgz) |
+> | compositional_objects_1.4_random_locs | zip | [compositional_objects_1.4_random_locs.zip](https://tbp-data-public-5e789bd48e75350c.s3.us-east-2.amazonaws.com/tbp.monty/compositional_objects_1.4_random_locs.zip) |
+>
+> Both COWS Small and COWS Large use the standard dataset. The random-location dataset is only required for the large `random_locs` experiments.
```
nielsleadholm: The term "standard" vs large/small might cause confusion. It might be clearer to say `compositional_objects_x.x` vs `compositional_objects_x.x_random_locs`

docs/overview/benchmark-experiments.md
```diff
@@ -119,111 +119,130 @@ These experiments are currently run without multiprocessing (using `run.py`).
...
-> curl -O https://tbp-data-public-5e789bd48e75350c.s3.us-east-2.amazonaws.com/tbp.monty/compositional_objects_1.2.zip
-> 
-> unzip compositional_objects_1.2.zip
+>
+> curl -O https://tbp-data-public-5e789bd48e75350c.s3.us-east-2.amazonaws.com/tbp.monty/compositional_objects_1.4.zip
+>
+> unzip compositional_objects_1.4.zip
 > ```
 >
 > Then create a symlink so that the experiment configs (which look for `~/tbp/data/habitat/objects/compositional_objects`) can find the versioned dataset folder. If this is the first time you are doing this, you can use:
```
nielsleadholm: Does the user also need to create a symlink for `compositional_objects_1.4_random_locs`? More generally, do the symlink instructions even apply any more? It looks like the version number might now be hard-coded as part of the constants?

docs/overview/benchmark-experiments.md
```diff
@@ -119,111 +119,130 @@ These experiments are currently run without multiprocessing (using `run.py`).
...
+> ln -s ~/tbp/data/habitat/versioned_data/compositional_objects_1.4 ~/tbp/data/habitat/objects/compositional_objects
 > ```
 >
-> To generate the pretrained models, run the following experiments in order:
-> ```
-> python run.py experiment=supervised_pre_training_objects_with_stickers_3d_children
-> python run.py experiment=supervised_pre_training_objects_with_stickers_2d_children
-> python run.py experiment=supervised_pre_training_objects_with_stickers_comp_models
-> python run.py experiment=supervised_pre_training_objects_with_stickers_monolithic_models
+> To generate the pretrained models, run the following experiments in order. train the shared five 3D children first, then the size-specific 2D children, then each parent model. Both parent architectures load the corresponding 2D-child checkpoint, which already contains the shared 3D children. Intermediate checkpoints are read from `${MONTY_MODELS}/my_trained_models/<experiment>/pretrained/`; keep the output directories consistent with these dependencies.
```
nielsleadholm:
```suggestion
> To generate the pretrained models, run the following experiments in order. Train the shared five 3D children first, then the size-specific 2D children, then each parent model. Both parent architectures load the corresponding 2D-child checkpoint, which already contains the shared 3D children. Intermediate checkpoints are read from `${MONTY_MODELS}/my_trained_models/<experiment>/pretrained/`; keep the output directories consistent with these dependencies.
```

benchmarks/cows_large.csv
```diff
@@ -0,0 +1,7 @@
+Experiment,Correct (%)|align right,Used MLH (%)|align right,Num Match Steps|align right,Rotation Error (degrees)|align right,Avg Prediction Error|align right,Run Time (mins)|align right,Num Episodes|hidden
+base_infer_cows_large_monolithic,6.72,88.52,92,80.31,0.42,30,357
+base_infer_cows_large_compositional,4.76,84.31,91,83.42,0.44,27,357
+randrot_noise_infer_cows_large_monolithic,3.08,74.51,83,129.73,0.77,30,357
+randrot_noise_infer_cows_large_compositional,2.80,74.23,83,122.43,0.78,32,357
+base_infer_cows_large_monolithic_random_locs,2.67,86.33,90,114.25,0.44,24,300
```
nielsleadholm, line 6: At some point it would be worth standardizing our inference experiment names a bit more. E.g. `random_locs_infer_cows_large_monolithic` might be more consistent with how we specify others like `randrot_noise_infer_cows_large_monolithic`. But I think we can leave it for now.

src/tbp/monty/conf/README.md
```diff
@@ -29,17 +29,21 @@ To generate models for the YCB experiments, run the following pretraining:
 
 All of the above can be run at the same time, in parallel.
 
-#### Objects with Logos Experiments
+#### COWS Experiments
 
-The compositional benchmark uses objects with logo stickers as a single baseline condition. It includes flat objects, curved objects, and rotated logo stickers.
+Compositional Objects With Stickers (COWS) has small (19 objects) and large (119 objects) variants. Both use dataset 1.4 and share the five plain 3D children.
```
nielsleadholm: "dataset 1.4" looks like it would be at risk of going stale. I would either use the full name (so that others will find it with control replace), or remove the version number reference here.

src/tbp/monty/conf/experiment/base_infer_cows_large_compositional_mujoco.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - /termination: named_lm2_objectrecognition_t1000_e100_tot6000
+  - /logging: basic_warning_wandb_evidence_eval_runs
+
+experiment:
+  _target_: tbp.monty.frameworks.experiments.object_recognition_experiments.MontyObjectRecognitionExperiment
+  config:
+    show_sensor_output: false
+    n_train_epochs: 3
+    n_eval_epochs: ${constants.rotations_3_count}
+    model_name_or_path: ${constants.compositional_pretrained_dir}/supervised_pre_training_cows_large_compositional/pretrained/
```
nielsleadholm, line 22: Shouldn't this use the `supervised_pre_training_cows_large_compositional_mujoco` models?

src/tbp/monty/conf/experiment/base_infer_cows_large_compositional_random_locs_mujoco.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - /termination: named_lm2_objectrecognition_t1000_e100_tot6000
+  - /logging: basic_warning_wandb_evidence_eval_runs
+
+experiment:
+  _target_: tbp.monty.frameworks.experiments.object_recognition_experiments.MontyObjectRecognitionExperiment
+  config:
+    show_sensor_output: false
+    n_train_epochs: 3
+    n_eval_epochs: ${constants.rotations_3_count}
+    model_name_or_path: ${constants.compositional_pretrained_dir}/supervised_pre_training_cows_large_compositional/pretrained/
```
nielsleadholm, line 22: Again, I think this should be `supervised_pre_training_cows_large_compositional_mujoco`?

src/tbp/monty/conf/experiment/base_infer_cows_large_monolithic_mujoco.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - /termination: named_lm2_objectrecognition_t1000_e100_tot6000
+  - /logging: basic_warning_wandb_evidence_eval_runs
+
+experiment:
+  _target_: tbp.monty.frameworks.experiments.object_recognition_experiments.MontyObjectRecognitionExperiment
+  config:
+    show_sensor_output: false
+    n_train_epochs: 3
+    n_eval_epochs: ${constants.rotations_3_count}
+    model_name_or_path: ${constants.compositional_pretrained_dir}/supervised_pre_training_cows_large_monolithic/pretrained/
```
nielsleadholm, line 22: Please check all of these unless I'm mistaken.

src/tbp/monty/conf/experiment/base_infer_cows_large_monolithic_random_locs_mujoco.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - /termination: named_lm2_objectrecognition_t1000_e100_tot6000
+  - /logging: basic_warning_wandb_evidence_eval_runs
+
+experiment:
+  _target_: tbp.monty.frameworks.experiments.object_recognition_experiments.MontyObjectRecognitionExperiment
+  config:
+    show_sensor_output: false
+    n_train_epochs: 3
+    n_eval_epochs: ${constants.rotations_3_count}
+    model_name_or_path: ${constants.compositional_pretrained_dir}/supervised_pre_training_cows_large_monolithic/pretrained/
```
nielsleadholm, line 22: ^

nielsleadholm (commented):
Overall looks good to me, just some high level comments:

1. For posterity, can you please comment in the PR description on:
- The reason for the performance change in base cows_small (presumably because of retraining on dark marble?)
- Why we skip version 1.3 for compositional_objects

2. I left a couple comments about the absence of mujoco pretrained models in the configs. If this is because we don't have these yet (e.g. you're waiting on Jeremy for something), then it would be worth commenting on this in the PR description, and making sure we have a story to track adding these. Otherwise we should include them now.

docs/overview/benchmark-experiments.md
hlee9212: Updated in 7fa09ff0.

docs/overview/benchmark-experiments.md
hlee9212: Updated in e5b92985.

docs/overview/benchmark-experiments.md
hlee9212: Updated in 308cfd0e.

docs/overview/benchmark-experiments.md
hlee9212: Good catch - I did explicitly set data_path in COWS experiment configs since I think the overall consensus is to move toward explicit paths in configs ([Jeremy Slack Post](https://thousandbrainsproject.slack.com/archives/C0842EPEMHV/p1789066265613369)). I'll only remove for the COWS section since the YCB experiments still use symlink.

hlee9212: Updated in dcb1ac93.

benchmarks/cows_large.csv
hlee9212: Agreed. Jeremy and I also want to keep run_name and file name consistent - like for pretraining we have something like `only_surf_something_something.yaml` but the run_name is `surf_something_something`.

src/tbp/monty/conf/README.md
hlee9212: Updated in e0929714.

src/tbp/monty/conf/experiment/base_infer_cows_large_compositional_random_locs_mujoco.yaml
hlee9212: The `_mujoco` experiment runs inference in MuJoCo, but currently loads the existing Habitat-trained v7 models. I _think_ Jeremy’s migration convention keeps checkpoint subdirectory names unchanged and uses MONTY_MODELS to select the model collection. I've made some other fixes to keep aligned with his updates so far, but the checkpoint path here should remain unsuffixed. (Same for the below comments)

hlee9212:
> I left a couple comments about the absence of mujoco pretrained models in the configs. If this is because we don't have these yet (e.g. you're waiting on Jeremy for something), then it would be worth commenting on this in the PR description, and making sure we have a story to track adding these. Otherwise we should include them now.

@nielsleadholm Not sure if I saw specific comments, but updated to add MuJoCo versions of `cows_small` experiments: just checked the files changed and saw that I only added MuJoCo versions to `cows_large` only. 
CleanShot 2026-09-21 at 12 43 11 PM (Updated in 7e89878b)

Also updated PR description.

benchmarks/cows_large.csv
nielsleadholm: Ok nice, yeah that would also be good to improve - or maybe even obviate, so run name is typically taken from the config unless the user specifically wants to append additional metadata like `run_name_append = debug_with_high_threshold`?

docs/overview/benchmark-experiments.md
```diff
@@ -138,13 +138,13 @@ The accuracy, Used MLH, average rotation error, and average prediction error in
 
 #### COWS Small
 
-`cows_small` contains the original 19 objects, including the five plain objects and objects with TBP or Numenta logo stickers. Its 2D child LM learns the two logos. 
+`cows_small` evaluates recognition 19 objects:five plain 3D objects (disk, cube, cylinder, sphere, and mug) and 14 compositional objects with either TBP or Numenta logos. The two logos are learned separately by the 2D child LM.
```
nielsleadholm: Can't edit inline:

--> *on 19 objects

nielsleadholm (commented):
Thanks for the update Hojae, including the description of the PR. 

Was it deliberate that you removed `_mujoco` from the `_run_name`s? Is that what you were partly referring to in terms of aligning with Jeremy? 

Re. the loaded pre-trained model, I would just coordinate with Jeremy so that we make sure that, after the switch-over, we are definitely loading Mujoco models, and not old Habitat models.

hlee9212:
Hi @nielsleadholm,

> Was it deliberate that you removed _mujoco from the _run_names? Is that what you were partly referring to in terms of aligning with Jeremy?

Yep, I deliberately removed the _mujoco from the run_names following the pattern of configs Jeremy has been writing. 

> Re. the loaded pre-trained model, I would just coordinate with Jeremy so that we make sure that, after the switch-over, we are definitely loading Mujoco models, and not old Habitat models.

Sounds good. I _think_ his initial plan was to make sure (1) MuJoCo inference works on Habitat pretrained, (2) MuJoCo pretrained works on Habitat Inference (this was one suggestion I gave him when he asked how to "evaluate" if the pretrained models are good, besides just checking the objects themselves), and (3) MuJoCo pretrained + MuJoCo inference. I think once we completely switch over to MuJoCo we may not have any `_mujoco` suffix (assuming we will no longer use Habitat)?

docs/overview/benchmark-experiments.md
hlee9212: Fixed.

nielsleadholm (approved):
Great, thanks Hojae.

Yeah the plan should be after we make the switch, that basically all Habitat and related items will be removed. So as long as we point to pretrained models that have been fully trained on a Mujoco pipeline, we should be good. 

Nice job getting all of this across the line, it was a major series of updates!

src/tbp/monty/conf/experiment/base_infer_cows_large_compositional_random_locs_mujoco.yaml
jeremyshoemaker: Rather than repeat things, I'll point you to [this comment](https://github.com/thousandbrainsproject/tbp.monty/pull/1205#issuecomment-5760667467) on another PR where I discussed this.

PR #1195, opened 2026-09-13
perf: limit BLAS/torch threads in parallel episode workers

foxalike-dev:
`run_episodes_parallel` spawns `num_parallel` worker processes, but nothing limits the BLAS/torch thread pools inside them: every worker starts a pool sized to the whole machine, so N workers oversubscribe it N times.

This PR caps each worker at one intra-op thread. The BLAS thread variables (OMP/MKL/OPENBLAS/VECLIB/NUMEXPR) are exported before the pool is spawned so the workers inherit them (values the user already set are left alone), and the pool initializer calls `torch.set_num_threads`. `MONTY_THREADS_PER_PROCESS` overrides the default of 1.

Measured on `base_config_10distinctobj_dist_agent` (140 episodes, pretrained YCB models, 10-core Apple M5):
- `num_parallel=8`: wall time 142 s -> 127 s, mean per-episode time 0.74 s -> 0.66 s, load-average peak 19 -> 12
- `num_parallel=4`: 120 s -> 119 s, per-episode time unchanged (fewer workers than cores, so little oversubscription)
- all four runs give identical per-episode results (performance, steps, MLH rotation)

The gain grows with the compute per episode: in my own longer-episode multi-LM configs a single process is ~17% faster with one thread than with the default pool. One thing I noticed while profiling, not touched here: on this benchmark most of the wall time is the parent's setup plus the per-episode process turnover (`maxtasksperchild=1` re-imports the framework and reloads the pretrained models for every episode); the episodes themselves are roughly a tenth of it.

Relates to the future-work item "Profile and Speed up Monty". Four unit tests added in `tests/unit/frameworks/run_parallel_test.py`; ruff check/format clean.


tristanls-tbp:
Thank you, @foxalike-dev.

FYI, @jeremyshoemaker @scottcanoe, I assigned this to myself for review, but let me know if you want to look into it instead.

tristanls-tbp (commented):
Thank you @foxalike-dev for spotting and implementing this performance improvement.

I think we can reduce what's necessary to implement your idea.

1. Keep `_export_thread_limits` as is.

2. Instead of an environment variable `MONTY_THREADS_PER_PROCESS`, create a configuration variable that is a sibling to `num_parallel` (see `src/tbp/monty/conf/experiment.yaml`). Suggest calling it `threads_per_process`. Then use that config variable as input to `_export_thread_limits` like so:

```python
_export_thread_limits(num_threads)
with ctx.Pool(# ... nothing needs to change here
```

So, one method, one line before `with ctx.Pool(...)` should cover it. No need for initializer, as the environment variables should fulfill that function.

**suggestion**: `_export_thread_limits` -> `_export_thread_limits_env`.

foxalike-dev:
Thanks for the review. Implemented as suggested:

- `_export_thread_limits` kept, renamed to `_export_thread_limits_env`.
- `MONTY_THREADS_PER_PROCESS` replaced by a `threads_per_process` config variable next to `num_parallel` in `conf/experiment.yaml`, passed into `run_episodes_parallel` and used in the one line before the pool is created.
- Pool initializer and `torch.set_num_threads` dropped. I checked that the exported variables cover torch as well: in a spawned worker `torch.get_num_threads()` is 1 once the parent has exported `OMP_NUM_THREADS=1`.

The branch is now +54 lines instead of +87, with two unit tests instead of four, and is rebased onto main (a17e7261). Re-ran `base_config_10distinctobj_dist_agent` at `num_parallel=8`: 140/140 correct, mean matching steps 28.5, mean rotation error 6.68 degrees, same as before the change.

PR #1200, opened 2026-09-16
refactor!: split Recognition Policies into composable parts

dschumacher-tbp:
This is part of the Epic: [Configurable Recognition Policies control Experiment episode boundaries](https://app.shortcut.com/thousandbrainsproject/epic/7943)

## Breaking Changes
- `model.is_done` check removed from `MaximumSteps`, use `AnyPolicy` with `MontyIsDone` instead.
- `max_{train,eval}_steps` check removed from `MinimumLMs`, use `AnyPolicy` with `MaximumSteps` instead.
- `model.is_done`  and `max_total_steps` checks removed from `NaiveScan`, use `AnyPolicy` with `MontyIsDone` and `MaxTotalSteps` instead.
- `MaximumSteps` uses `model.matching_steps` and `model.is_exploring` instead of `count.step`.

WIP tracked in [Move terminal-condition checking from Monty into Experiment.](https://app.shortcut.com/thousandbrainsproject/story/7945/move-terminal-condition-checking-from-monty-into-experiment)


tests/unit/experiments/recognition_policy_test.py
```diff
@@ -173,108 +173,45 @@ def test_raises_value_error_if_max_eval_steps_is_not_positive(
...
-            None
-        ] * num_pending
+        match_options = list(RecognitionConclusion)
+        conclusions: list[RecognitionConclusion | None] = [
+            random.choice(match_options)  # noqa: S311
+            for _ in range(num_concluded)
+        ]
+        for _ in range(num_pending):
+            i = random.randint(0, len(conclusions))  # noqa: S311
+            conclusions.insert(i, None)
```
ramyamounir: I would avoid using `random` to generate random data that's not under the control of the Hypothesis library. See the [comment here](https://github.com/thousandbrainsproject/tbp.monty/pull/951#discussion_r3161861166). I don't know if this change is necessary at this point. If you think it is, is there a way to generate this `list[RecognitionConclusion]` using `st.composite` or some other way using Hypothesis?

tests/unit/experiments/recognition_policy_test.py
```diff
@@ -173,108 +173,45 @@ def test_raises_value_error_if_max_eval_steps_is_not_positive(
...
         is_done = step >= 307
         self.assertEqual(result.is_done, is_done)
 
-    @given(is_done=st.booleans(), step=st.integers(min_value=0, max_value=306))
-    def test_defers_to_model_before_step_limit(self, is_done: bool, step: int) -> None:
-        model = _model_is_done(is_done)
-        policy = NaiveScan(max_total_steps=500, fixed_amount=5)
-        count = RecognitionCounter(step)
+    @given(fixed_amount=st.integers(min_value=1, max_value=100))
+    def test_step_limit_less_than_10000(self, fixed_amount: int) -> None:
```
ramyamounir, line 233: Can you provide some documentation about why the `step_limit` is never expected to be higher than `10000` regardless of the `fixed_amount` value?

src/tbp/monty/experiment/recognition_policy.py
```diff
@@ -120,153 +145,86 @@ def __init__(self: Self, max_train_steps: int, max_eval_steps: int) -> None:
...
-            raise ValueError("max_total_steps must be positive")
-
         if fixed_amount <= 0:
             raise ValueError("fixed_amount must be positive")
-        k = math.ceil(90 / fixed_amount)  # arm length when angular extent >= 90
-        max_scan_steps = k * (k - 1) + 1  # 0 when k <= 1 (i.e.: fixed_amount >= 90)
 
-        self._step_limit = min(max_total_steps, max_scan_steps)
+        k = math.ceil(90 / fixed_amount)  # arm length when angular extent >= 90
+        self._step_limit = k * (k - 1) + 1  # 0 when k <= 1 (i.e.: fixed_amount >= 90)
```
ramyamounir: When `fixed_amount>=90`, `k=1` because of the `math.ceil`. So the smallest possible `_step_limit` becomes `1` step. 

I don't know if `self._step_limit` can ever be `0`. I know you didn't add this comment in this PR, but it's probably better to fix it here while we're adjusting the policy.

src/tbp/monty/experiment/recognition_policy.py
```diff
@@ -120,153 +145,86 @@ def __init__(self: Self, max_train_steps: int, max_eval_steps: int) -> None:
...
         Args:
             min_lms: The number of Learning Modules that must reach a conclusion for
                 the policy to be satisfied.
-            max_train_steps: The maximum steps to take in training mode.
-            max_eval_steps: The maximum steps to take in evaluation mode.
 
         Raises:
-            ValueError: If `min_lms`, `max_train_steps`, or `max_eval_steps`
-                are not positive.
+            ValueError: If `min_lms`is not positive.
```
ramyamounir:
```suggestion
            ValueError: If `min_lms` is not positive.
```

src/tbp/monty/experiment/recognition_policy.py
```diff
@@ -86,13 +86,38 @@ def __call__(
...
+            raise ValueError("max_total_steps must be positive")
+        self._max_total_steps = max_total_steps
+
+    def __call__(
+        self: Self,
+        model: MontyBase,  # noqa: ARG002
+        count: RecognitionCounter,
+    ) -> RecognitionResult:
+        # Even if many exploratory steps have not sent information to learning
+        # modules (so is_done remains False), eventually terminate exploration
```
ramyamounir: I think this comment assumes that the `MontyIsDone` policy is still part of `MaxTotalSteps`. Maybe it's best to just remove it, the policy is self-explanatory.

ramyamounir (changes requested):
Really nice, I like how this simplifies the policies and the tests! I left some minor comments.

In general `max_{train,eval}_steps` should be compared against the model's `matching_steps`, not the `RecognitionCounter.step`. So I have a suspicion that `MaximumSteps` is incorrect and will need to be changed to the behavior inside the `ObjectRecognition` policy [here](https://github.com/thousandbrainsproject/tbp.monty/blob/8cf3b455b0254100084926738fc393d3daace8c9/src/tbp/monty/experiment/recognition_policy.py?plain=1#L283). I looked at where the policy `MaximumSteps` is being used and it's only inside `test_any1_maximumsteps_t10_e5.yaml`, which is only referenced some basic tests in `test/run` and `test/run_mujoco` [here](https://github.com/thousandbrainsproject/tbp.monty/blob/8cf3b455b0254100084926738fc393d3daace8c9/tests/integration/frameworks/run_test.py?plain=1#L98-L105). Nothing to change in this PR, but just something to think about for the next refactoring.

src/tbp/monty/experiment/recognition_policy.py
dschumacher-tbp: I've updated the documentation as follows:
```python
        k = math.ceil(90 / fixed_amount)  # (>=90)->1, 10->9, 5->18, 1->90
        self._step_limit = k * (k - 1) + 1  # 1->1, 9->73, 18->307, 90->8011
```

This should also explain the `10000` and `307` constants in the tests.

src/tbp/monty/experiment/recognition_policy.py
dschumacher-tbp: Removed.

src/tbp/monty/experiment/recognition_policy.py
dschumacher-tbp: Fixed.

tests/unit/experiments/recognition_policy_test.py
dschumacher-tbp: See the updated documentation in `recognition_policy.py:NaiveScan`.

tests/unit/experiments/recognition_policy_test.py
tristanls-tbp: given a generator of `RecognitionConclusion` such as `recognition_conclusion`, one can do:
```python
@st.composite
def recognition_conclusions(
    ...
) -> list[RecognitionConclusion]:
    return draw(st.lists(recognition_conclusion(...)))

dschumacher-tbp:
I'm still considering how best to fix `MinimumLMsTest`, so not quite ready to merge yet.

As for `MaximumSteps`, I simply refactored the existing implementation, which used `count.step`. Clearly, this will not work for `ObjectRecognition`, but there are already several reasons that one can't be decomposed (yet). We'll have to address the actual use-case(s) in a future PR.

ramyamounir:
> As for MaximumSteps, I simply refactored the existing implementation, which used count.step.

I was making a point that this was incorrect even before this IP (not something we introduced). This incorrect matching of `max_steps` against the experiment step counter is also not used in the code except in some very simple test, that's why it was not an issue. We can fix `MaximumSteps` and use for object recognition instead of creating a new one.

tests/unit/experiments/recognition_policy_test.py
dschumacher-tbp: With help from @tristanls-tbp, I've created a Hypothesis Strategy to generate the list of conclusions.

ramyamounir (approved):
Thank you for all the changes, LGTM.

dschumacher-tbp:
@ramyamounir Thanks for carrying this over the finish line.

PR #1202, opened 2026-09-17
refactor: clip coherence to [0, 1]

scottcanoe:
**Breaking Changes: None**

Hypothesis found a failing test for [`StructuredTensorTest.test_coherence_in_unit_interval`](https://github.com/thousandbrainsproject/tbp.monty/blob/f55675b3fdd09bae35589d9efba143871b6701c4/tests/unit/frameworks/utils/edge_detection_test.py#L333). That test looks like this:
```python
    @given(t=structure_tensors())
    @example(t=StructureTensor(xx=4.0, yy=0.0, xy=0.0))
    def test_coherence_in_unit_interval(self, t):
        assert 0.0 <= t.coherence <= 1.0
```

In the failing example, `t.coherence` was `1.0000000000000002`. Briefly, `np.linalg.eigh` returned a very small but negative eigenvalue (`-1.0658141036401503e-14`) which caused the coherence calculation `(lambda_max - lambda_min) / (lambda_max + lambda_min)` to go slightly over `1.0`. [Here's the full coherence calculation code](https://github.com/thousandbrainsproject/tbp.monty/blob/f55675b3fdd09bae35589d9efba143871b6701c4/src/tbp/monty/frameworks/utils/edge_detection.py#L124), with my edit:
```python
    @property
    def coherence(self) -> float:
        """Edge quality in [0, 1]: 1 means perfectly oriented, 0 means isotropic."""
        lambda_min, lambda_max = self.eigenvalues
        lambda_diff = lambda_max - lambda_min
        if lambda_diff < DEFAULT_TOLERANCE:
            return 0.0
        return np.clip(lambda_diff / (lambda_max + lambda_min), 0.0, 1.0)  # <-- We don't currently clip.
```

There isn't a real issue here. The eigenmatrix was basically singular, so we just got a bit of floating point runoff in the wrong (i.e., <0) direction. Therefore, I think it's probably fine to just clip the coherence to [0, 1], which is the range the docstring says to expect.

This is my first time looking at this code though. I'm assigning @hlee9212 and @jeremyshoemaker since I believe they're the experts here.


hlee9212 (approved):
Thanks for catching this, Scott. Yes, `1.0000000000000002` looks like just one floating-point step above. I didn't think about clipping because structure tensors are supposed to be positive semi-definite matrices (i.e. eigenvalues >= 0), but alas, floating point rounding....

jeremyshoemaker (approved):
Seems reasonable to me.

PR #1203, opened 2026-09-18
chore: let docs cleanup succeed when a preview branch is already absent

brothman-tbp:
The cleanup_docs_preview job was failing when it tried to delete a version that didn’t exist (https://github.com/thousandbrainsproject/tbp.monty/actions/runs/35155958458/job/104995307521). There’s nothing to clean up in that case, so the job should finish successfully.

This PR makes three small changes:

- `req.py`: Return normally when a DELETE request gets a 404. This applies to all callers of the shared deletion helper (currently, that's `delete_version`, `delete_category` and `delete_doc` in `readme.py`).
- `readme.py`: Update the `delete_version` log message to explain it either successfully deleted the version or the version was already absent.
- `req_test.py`: Add a test confirming that a DELETE returning 404 completes without raising an exception.

Other HTTP errors still raise `ReadMeRequestError` in `req.delete`, so other failures remain visible.


tools/github_readme_sync/readme.py
```diff
@@ -676,7 +676,10 @@ def replace_image(match):
 
     def delete_version(self):
         delete(f"{API_PREFIX}/branches/{self.version}")
-        logger.info(f"{GREEN}Successfully deleted version {self.version}{RESET}")
+        logger.info(
+            f"{GREEN}Successfully deleted version {self.version} "
+            f"or confirmed it was already absent{RESET}"
```
tristanls-tbp, line 681: **suggestion (non-blocking)**:

```suggestion
            f"{GREEN}Successfully deleted version {self.version} or not found"
```

tristanls-tbp (approved):
Thank you, looks good 👍.

PR #1205, opened 2026-09-18
chore: create MuJoCo versions of ycb_unsupervised experiments

jeremyshoemaker:
This creates MuJoCo versions of the experiments present in `ycb_unsupervised.csv`. This work uncovered the bug fixed in #1174. After updating this branch to include that one, we no longer get the error.

Diff against Habitat versions:

```diff
=== diff: surf_agent_unsupervised_10distinctobj_noise.yaml <-> surf_agent_unsupervised_10distinctobj_noise_mujoco.yaml ===
-   - /environment: habitat_ycb_surf_agent
+   - /environment: mujoco_ycb_surf_agent
-       run_name: surf_agent_unsupervised_10distinctobj_noise
+       run_name: surf_agent_unsupervised_10distinctobj_noise_mujoco

=== diff: surf_agent_unsupervised_10simobj.yaml <-> surf_agent_unsupervised_10simobj_mujoco.yaml ===
-   - /environment: habitat_ycb_surf_agent
+   - /environment: mujoco_ycb_surf_agent
-       run_name: surf_agent_unsupervised_10simobj
+       run_name: surf_agent_unsupervised_10simobj_mujoco

=== diff: surf_agent_unsupervised_10distinctobj.yaml <-> surf_agent_unsupervised_10distinctobj_mujoco.yaml ===
-   - /environment: habitat_ycb_surf_agent
+   - /environment: mujoco_ycb_surf_agent
-       run_name: surf_agent_unsupervised_10distinctobj
+       run_name: surf_agent_unsupervised_10distinctobj_mujoco
```


src/tbp/monty/conf/experiment/surf_agent_unsupervised_10distinctobj_mujoco.yaml
```diff
@@ -0,0 +1,26 @@
...
+  config:
+    show_sensor_output: false
+    n_train_epochs: 10
+    n_eval_epochs: ${constants.rotations_all_count}
+    model_name_or_path: ""
+    seed: 42
+    supervised_lm_ids: []
+    python_log_level: DEBUG
+    logging:
+      run_name: surf_agent_unsupervised_10distinctobj_mujoco
```
tristanls-tbp, line 26: **question**: Did you want to not include `_mujoco` in these?

tristanls-tbp (commented):
Looks good, except I think you want run names to omit `_mujoco` suffix. Please confirm.

jeremyshoemaker:
> Looks good, except I think you want run names to omit `_mujoco` suffix. Please confirm.

My approach has evolved as I've gone along. I was initially adding the suffix to experiments so that the output would be in different directories. Then when I got to the training experiments, I realized if I added the suffix there, then I'd have to go back and change all the inference experiments to update the model paths. 

It seemed easier to have the training experiments use the same paths, without the suffix, than to go back and change the rest only to change them again later when we move the MuJoCo versions in place of the Habitat versions. My plan is that when we remove the Habitat ones, all of the MuJoCo ones will have the suffix removed, and from a Git history perspective it will look like the MuJoCo ones are just updates of the Habitat versions.

So, currently the approach is:
- Inference experiment run names get the `_mujoco` suffix.
- Training experiment run names do not.

jeremyshoemaker:
Another benefit of this approach is that with the training experiments not having the suffix, we can run the MuJoCo inference experiments against Habitat models or MuJoCo models by just adjusting the MONTY_MODELS variable.

tristanls-tbp:
@jeremyshoemaker, understood on all accounts.

Are these not **training** experiments?

```yaml
  - /env_interface: train_distinctobj_random
  - /env_interface: train_simobj_random
```

You said training experiments do not get the `_mujoco` suffix.

jeremyshoemaker:
> @jeremyshoemaker, understood on all accounts.
> 
> Are these not **training** experiments?
> 
> ```yaml
>   - /env_interface: train_distinctobj_random
>   - /env_interface: train_simobj_random
> ```
> 
> You said training experiments do not get the `_mujoco` suffix.

I hadn't noticed that they were training. Now that I'm looking at this some more, I'm wondering if we need to do these at all. They don't appear in the `run_all_benchmarks.sh` script. Are we still using them? I pulled them from `ycb_unsupervised.csv` without checking to see if we regularly run them.

tristanls-tbp:
@ramyamounir any thoughts on these experiments and @jeremyshoemaker 's observations above? https://github.com/thousandbrainsproject/tbp.monty/pull/1205#issuecomment-5763464658

ramyamounir:
These are the unsupervised learning experiments reported in this [benchmark table](https://docs.thousandbrains.org/docs/benchmark-experiments#unsupervised-learning). We don't run them as part of the `run_all_benchmarks` script because they need some post processing to get the values shown in the columns, but we do need Mujoco configs for them.

Re. the `_mujoco` suffix. These experiments do not store pretrained models that we later use during inference. So I think it's fair to treat them as inference experiments, i.e., use the `_mujoco` suffix. Note that these experiments use `MontyObjectRecognitionExperiment`, they just don't start from pretrained models, they learn their models during the run as a demo of continual learning.

tristanls-tbp:
Great, thank you @ramyamounir.

tristanls-tbp (approved):
Thanks for all the clarifications 🙂.

jeremyshoemaker:
Thanks @ramyamounir

